PR #57, opened 2017-02-19
[ServiceSDK] Add IDisposable to base classes

yfakariya:
This PR added `IDisposable` and `protected virtual void Dispose(bool)` to base classes of service SDK to improve API usability and users can use them more stable easily.

### Benefit

* API users should remember cleanup by existence of IDisposable.
    * It is more common for .NET developers to call Dispose() to clean up resources even if they expose CloseAsync(). For example, System.IO.Stream classes still expose Dispose while they expose asynchronous IOs.
    * Many code analysis tools warns about forget of Dispose() call but do not for CloseAsync().

### Changes summary

* I added IDisposable and `protected virtual void Dispose(bool)` to following classes:
    * `ServiceClient`
    * `JobClient`
    * `RegistryManager`
* I integrated `CloseAsync` and `Dispose(bool)` implementation to avoid code duplication.
* I added `IHttpClientHelper.Dispose()` to `AmqpServiceClient.CloseAsync()`. They should be equal effectively.

### Background story

* I faced `FeebackReceiver` unresponsiveness last week because we had forgotten to call `CloseAsync()`. In the investigation of this work, we noticed that non public derived classes implements `IDisiposable` but public base classes do not.
* In addition, it is common to wrap external access including service SDK for unit testing and implements `IDisposable` in it, so implementing `IDisposable` is more natural than only implements `CloseAsync()`.

### DIFF
--- a/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
+++ b/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
@@ -53,9 +53,9 @@ namespace Microsoft.Azure.Devices
                 DefaultOperationTimeout,
                 client => {});
         }
-                
+
         internal AmqpServiceClient(IotHubConnectionString iotHubConnectionString, bool useWebSocketOnly, IHttpClientHelper httpClientHelper) : base()
-        {            
+        {
             this.httpClientHelper = httpClientHelper;
         }
 
@@ -101,7 +101,9 @@ namespace Microsoft.Azure.Devices
 
         public async override Task CloseAsync()
         {
+            await this.faultTolerantSendingLink.CloseAsync();
             await this.feedbackReceiver.CloseAsync();
+            await this.fileNotificationReceiver.CloseAsync();
             await this.iotHubConnection.CloseAsync();
         }
 
@@ -218,6 +220,7 @@ namespace Microsoft.Azure.Devices
             if (disposing)
             {
                 this.faultTolerantSendingLink.Dispose();
+                this.fileNotificationReceiver.Dispose();
                 this.feedbackReceiver.Dispose();
                 this.iotHubConnection.Dispose();
                 this.httpClientHelper.Dispose();
--- a/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
+++ b/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
@@ -101,10 +101,8 @@ namespace Microsoft.Azure.Devices
 
         public async override Task CloseAsync()
         {
-            this.faultTolerantSendingLink.Dispose();
             await this.feedbackReceiver.CloseAsync();
             await this.iotHubConnection.CloseAsync();
-            this.httpClientHelper.Dispose();
         }
 
         public async override Task SendAsync(string deviceId, Message message)
@@ -219,6 +217,7 @@ namespace Microsoft.Azure.Devices
         {
             if (disposing)
             {
+                this.faultTolerantSendingLink.Dispose();
                 this.feedbackReceiver.Dispose();
                 this.iotHubConnection.Dispose();
                 this.httpClientHelper.Dispose();
--- a/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
+++ b/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
@@ -219,8 +219,6 @@ namespace Microsoft.Azure.Devices
         {
             if (disposing)
             {
-                // This is more efficient than async and wait in base.Dispose(bool).
-                this.faultTolerantSendingLink.Dispose();
                 this.feedbackReceiver.Dispose();
                 this.iotHubConnection.Dispose();
                 this.httpClientHelper.Dispose();
--- a/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
+++ b/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
@@ -59,6 +59,15 @@ namespace Microsoft.Azure.Devices
             this.httpClientHelper = httpClientHelper;
         }
 
+        internal AmqpServiceClient(IotHubConnection iotHubConnection, IHttpClientHelper httpClientHelper)
+        {
+            this.iotHubConnection = iotHubConnection;
+            this.faultTolerantSendingLink = new FaultTolerantAmqpObject<SendingAmqpLink>( this.CreateSendingLinkAsync, iotHubConnection.CloseLink );
+            this.feedbackReceiver = new AmqpFeedbackReceiver(iotHubConnection);
+            this.fileNotificationReceiver = new AmqpFileNotificationReceiver(iotHubConnection);
+            this.httpClientHelper = httpClientHelper;
+        }
+
         public TimeSpan OpenTimeout
         {
             get
--- a/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
+++ b/service/Microsoft.Azure.Devices/AmqpServiceClient.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Azure.Devices
     using Microsoft.Azure.Devices.Common.Data;
     using Microsoft.Azure.Devices.Common.Exceptions;
 
-    sealed class AmqpServiceClient : ServiceClient, IDisposable
+    sealed class AmqpServiceClient : ServiceClient
     {
         static readonly TimeSpan DefaultOperationTimeout = TimeSpan.FromSeconds(100);
         const string StatisticsUriFormat = "/statistics/service?" + ClientApiVersionHelper.ApiVersionQueryString;
@@ -101,8 +101,10 @@ namespace Microsoft.Azure.Devices
 
         public async override Task CloseAsync()
         {
+            this.faultTolerantSendingLink.Dispose();
             await this.feedbackReceiver.CloseAsync();
             await this.iotHubConnection.CloseAsync();
+            this.httpClientHelper.Dispose();
         }
 
         public async override Task SendAsync(string deviceId, Message message)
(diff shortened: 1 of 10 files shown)


yfakariya:
I've fixed (and squashed) latest commit since I found a mistake.

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
```diff
@@ -101,8 +101,10 @@ public SendingAmqpLink SendingLink
 
         public async override Task CloseAsync()
         {
+            this.faultTolerantSendingLink.Dispose();
```
rajeevmv: >this.faultTolerantSendingLink.Dispose(); [](start = 12, length = 40)

Disposing the sending link in not graceful. Also, it is not needed. Closing the Amqp connection will close all links and sessions inside it.

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
```diff
@@ -213,11 +215,16 @@ Task<SendingAmqpLink> CreateSendingLinkAsync(TimeSpan timeout)
...
         /// <inheritdoc/>
-        public void Dispose()
+        protected override void Dispose(bool disposing)
         {
-            this.faultTolerantSendingLink.Dispose();
-            this.feedbackReceiver.Dispose();
-            this.httpClientHelper.Dispose();
+            if (disposing)
+            {
+                // This is more efficient than async and wait in base.Dispose(bool).
```
rajeevmv: >// This is more efficient than async and wait in base.Dispose(bool). [](start = 16, length = 68)

This is confusing. The base class dispose method is never used.

service/Microsoft.Azure.Devices/ServiceClient.cs
```diff
@@ -44,6 +45,26 @@ public static ServiceClient CreateFromConnectionString(string connectionString)
...
+        /// </summary>
+        /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if ( disposing )
+            {
+                // Call CloseAsync() synchronously avoiding AggregateException.
+                this.CloseAsync().GetAwaiter().GetResult();
+            }
+        }
```
rajeevmv: Remove this implementation. It is not used.

service/Microsoft.Azure.Devices/RegistryManager.cs
```diff
@@ -26,6 +26,26 @@ public static RegistryManager CreateFromConnectionString(string connectionString
...
+        /// </summary>
+        /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
+        protected virtual void Dispose( bool disposing )
+        {
+            if (disposing)
+            {
+                // Call CloseAsync() synchronously avoiding AggregateException.
+                this.CloseAsync().GetAwaiter().GetResult();
+            }
+        }
```
rajeevmv: Remove this implementation. It is never used

service/Microsoft.Azure.Devices/JobClient/JobClient.cs
```diff
@@ -25,6 +25,26 @@ public static JobClient CreateFromConnectionString(string connectionString)
...
+        /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
+        protected virtual void Dispose( bool disposing )
+        {
+            if (disposing)
+            {
+                // Call CloseAsync() synchronously avoiding AggregateException.
+                this.CloseAsync().GetAwaiter().GetResult();
+            }
+        }
+
```
rajeevmv: Remove this implementation. Let the derived  concrete implementation deal with this.

rajeevmv:
:clock1:

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
yfakariya: Do you mean that this line should be removed because the link will be shutdown in `iotHubConnection.Dispoise()` ?

service/Microsoft.Azure.Devices/ServiceClient.cs
yfakariya: I think these base implementations are required because they are `public` (they can have another implementation than genuine Service SDK) and fallback implementation should exist. What do you think about this?

yfakariya:
@rajeevmv Thank you for reviewing, but I have 3 questions for your comments. I will fix them after the answers.

The 3rd question is: what do you mean the `1clock` emoji? Thank you!

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
rajeevmv: Yes.

---
In reply to: [103077496](https://github.com/Azure/azure-iot-sdk-csharp/pull/57#discussion_r103077496) [](ancestors = 103077496)

service/Microsoft.Azure.Devices/ServiceClient.cs
rajeevmv: Since this is an abstract class, you can declare it as protected abstract void Dispose(bool disposing);

---
In reply to: [103077630](https://github.com/Azure/azure-iot-sdk-csharp/pull/57#discussion_r103077630) [](ancestors = 103077630)

rajeevmv:
Sorry, I don't understand the 3rd question. I am not sure what you mean by '1 clock emoji'.

---
In reply to: [282479826](https://github.com/Azure/azure-iot-sdk-csharp/pull/57#issuecomment-282479826) [](ancestors = 282479826)

service/Microsoft.Azure.Devices/ServiceClient.cs
yfakariya: You are right. But adding abstract methods to public class causes breaking change (It may hurt people who already have derived classes such as hand made mock objects.)
Would you tell me why you think the base impls should be removed?

yfakariya:
I mean that I don't understand what 🕐 is :)

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
yfakariya: I got it, thanks.

service/Microsoft.Azure.Devices/ServiceClient.cs
rajeevmv: In general, that is a valid concern. However, in this case, ServiceClient has an internal parameterless ctor preventing mock objects. 

The real issue is that I believe that dispose should not attempt a graceful close mechanism like CloseAsync(). It should be ungraceful and run to completion without throwing any exceptions. Calling a CloseAsync method which is overridden by derived classes does not allow this.

---
In reply to: [103446925](https://github.com/Azure/azure-iot-sdk-csharp/pull/57#discussion_r103446925) [](ancestors = 103446925)

rajeevmv:
I still don't get it. Where did I refer to  '🕐 ' ?

---
In reply to: [283032466](https://github.com/Azure/azure-iot-sdk-csharp/pull/57#issuecomment-283032466) [](ancestors = 283032466)

service/Microsoft.Azure.Devices/ServiceClient.cs
yfakariya: Thank you for explanation. I understand it.

yfakariya:
This comment https://github.com/Azure/azure-iot-sdk-csharp/pull/57#issuecomment-282353680 
I'll not care about this since I recognize it was not important comment.

yfakariya:
@rajeevmv PTAL

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
```diff
@@ -213,11 +215,14 @@ Task<SendingAmqpLink> CreateSendingLinkAsync(TimeSpan timeout)
...
+        protected override void Dispose(bool disposing)
         {
-            this.faultTolerantSendingLink.Dispose();
-            this.feedbackReceiver.Dispose();
-            this.httpClientHelper.Dispose();
+            if (disposing)
+            {
+                this.feedbackReceiver.Dispose();
+                this.iotHubConnection.Dispose();
+                this.httpClientHelper.Dispose();
```
rajeevmv: I would add this.faultTolerantSendingLink.Dispose() here.

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
```diff
@@ -101,8 +101,10 @@ public SendingAmqpLink SendingLink
 
         public async override Task CloseAsync()
```
rajeevmv: >CloseAsync() [](start = 35, length = 12)

I would remove both changes to this method. The dispose method will take care of all disposing actions.

service/Microsoft.Azure.Devices/HttpRegistryManager.cs
```diff
@@ -71,11 +71,7 @@ public override Task OpenAsync()
 
         public override Task CloseAsync()
         {
-            if (this.httpClientHelper != null)
-            {
-                this.httpClientHelper.Dispose();
-                this.httpClientHelper = null;
-            }
+            this.Dispose();
```
rajeevmv: > this.Dispose(); [](start = 11, length = 16)

This is no longer required. Since dispose will take care of it.

rajeevmv:
:clock1:

rajeevmv:
One other thing: Can you add unit test cases to exercise the IDisposable methods that were added?

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
yfakariya: Sorry, I removed it from here rather than `CloseAsync()`

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
yfakariya: Do you mean separating roles of "CloseAsync()` and `Dispose()`? (Do you want to remove `Dispose()` s from `CloseAsync()`?

rajeevmv: Yes. Since Dispose() is guaranteed to run at the end, no need to invoke it from CloseAsync()

---
In reply to: [104151627](https://github.com/Azure/azure-iot-sdk-csharp/pull/57#discussion_r104151627) [](ancestors = 104151627)

yfakariya:
@rajeevmv I revised `Dispose()` and `CloseAsync()`, separate them, and added tests for them. PTAL.

jasmineymlo:
merged with commit 0b6b1cd0283b6c15b3419f6d004b21b54a1d72c5.
@yfakariya Thanks for providing the PR.

PR #59, opened 2017-02-22
Update a few things that were out of date

arturl:
Fixes:
 - UWP now supports MQTT, so removed a statement that says it doesn't;
 - A recommended workaround is no longer needed since the underlying issue with the NuGet package has been fixed.


tameraw:
@arturl - Thanks for the updates!

Issue #76, opened 2017-03-26
Azure.Devices.Client deprecated dotnetty signed nuget dependency


krish-gh:
Azure.Devices.Client nuget has dependencies on dotnetty signed nugets which are deprecated and has different public token in the dlls than the mainstream ones. Since Azure.Devices.Client is also used in https://github.com/Azure/azure-iot-protocol-gateway, it's makes it impossible to update dotnetty nugets there because of difference of publictoken. Protocol Gateway needs to move to dotnetty mainstream nugets and this issue is preventing that.

Raised https://github.com/Azure/azure-iot-sdk-csharp/pull/77

krish-gh:
I created a Pull request #77 since there is an urgency in a customer engagement. Otherwise, I have to create a private nuget from the fork.

jasmineymlo:
fixed in commit cfc7dc89f358d44576701e35e5cdeb95bd7899cc

PR #77, opened 2017-03-26
changed deprecated dotnetty nuget dependency to mainstream

krish-gh:
Azure.Devices.Client nuget has dependencies on dotnetty signed nugets which are deprecated and has different public token in the dlls than the mainstream ones. Since Azure.Devices.Client is also used in https://github.com/Azure/azure-iot-protocol-gateway, it's makes it impossible to update dotnetty nugets there because of difference of publictoken. Protocol Gateway needs to move to dotnetty mainstream nugets and this issue is preventing that.

#76

### DIFF
--- a/device/Microsoft.Azure.Devices.Client/Microsoft.Azure.Devices.Client.csproj
+++ b/device/Microsoft.Azure.Devices.Client/Microsoft.Azure.Devices.Client.csproj
@@ -59,34 +59,7 @@
     <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
   <Choose>
-    <When Condition="'$(DotNettyPath)' == ''">
-      <ItemGroup>
-        <Reference Include="DotNetty.Buffers, Version=0.3.2.0, Culture=neutral, PublicKeyToken=e7a0210a354f294a, processorArchitecture=MSIL">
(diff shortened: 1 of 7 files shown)


device/Microsoft.Azure.Devices.Client/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -59,34 +59,28 @@
...
-        </Reference>
-        <Reference Include="DotNetty.Transport, Version=0.3.2.0, Culture=neutral, PublicKeyToken=e7a0210a354f294a, processorArchitecture=MSIL">
-          <HintPath>$(SolutionDir)\packages\DotNetty.Transport-signed.0.3.2\lib\net45\DotNetty.Transport.dll</HintPath>
-          <Private>True</Private>
-        </Reference>
-      </ItemGroup>
-    </When>
+    <When Condition="'$(DotNettyPath)' == ''" >
+	<ItemGroup>
+    <Reference Include="DotNetty.Buffers, Version=0.0.0.0, Culture=neutral, PublicKeyToken=bc13ca065fa06c29, processorArchitecture=MSIL">
```
jasmineymlo: The assembly versions of the DotNetty nuget packages are wrong.

jasmineymlo: @krish-gh, 
Thanks for the PR.  Due to the issue in DotNetty package versioning in v0.4.3, we will be upgrading to the next available DotNetty release (v0.4.4) when it becomes available.

PR #84, opened 2017-04-12
Update nuspec to fix deprecated DotNetty libs

K47S:
DotNetty.Codecs.Mqtt and DotNetty.Handlers don't exist in the 0.4.4 version with -signed


jasmineymlo:
We have released a new version which fixed the issue: Microsoft Azure Devices Client 1.2.8.

Thanks.

PR #109, opened 2017-05-16
Pretty print twin JSON

avranju:
Before pic:

image

After pic:

image


tameraw:
@avranju - Thanks for providing this PR!

PR #131, opened 2017-07-10
Add ConnectionStatusChangeReason to ConnectionEventArgs

jasmineymlo:
Add ConnectionStatusChangeReason field in ConnectionEventArgs for population when the information is available from transports (mqtt or amqp).

### DIFF
--- a/device/Microsoft.Azure.Devices.Client/ConnectionEventArgs.cs
+++ b/device/Microsoft.Azure.Devices.Client/ConnectionEventArgs.cs
@@ -13,5 +13,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
         public string ConnectionKey { get; set; }
 
         public ConnectionStatus ConnectionStatus { get; set; }
+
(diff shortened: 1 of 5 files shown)


e2e/Microsoft.Azure.Devices.E2ETests/MethodE2ETests.cs
```diff
@@ -68,7 +68,7 @@ static public void ClassCleanup()
             await sendMethodAndRespondWithObseletedSetMethodHandler(Client.TransportType.Mqtt_WebSocket_Only);
         }
 
-        [Ignore]
+        //[Ignore]
```
anthonyvercolano: >       //[Ignore] [](start = 1, length = 17)

Should this just be deleted?

anthonyvercolano:
:shipit:

e2e/Microsoft.Azure.Devices.E2ETests/MethodE2ETests.cs
jasmineymlo: removed change

PR #132, opened 2017-07-10
Enable Twin and Direct Methods on AMQP

jasmineymlo:
Remove WIP compile flag to enable Twin and Direct Methods on AMQP by default


anthonyvercolano:
:shipit:

PR #133, opened 2017-07-11
Device Explorer new C2D listening Tab

Niyo:
# Device Explorer
I implemented a new Tab for the Device Explorer.
You can receive cloud to device messages in this tab and also complete/abondon them.

## Scenario
You can now send messages to a device in the already existing "Messages to Device", then receive those messages with the newly implemented tab and complete/abondon them. Then you will see (if feedback receive is activated) the feedback in the "Messages to Device" tab.

This is very useful when simulating devices.

## Changes
- new C2D listening Tab
- improvements in `updateDeviceIdsComboBoxes` method
- added Microsoft.Azure.Devices.Client nuget package for C2D receive

I tried to not change the existing coding style with my changes.

**Please tell me if I need to change something, before this gets approved.**

## Screenshots
image

### DIFF
--- a/tools/DeviceExplorer/DeviceExplorer/CommandAction.cs
+++ b/tools/DeviceExplorer/DeviceExplorer/CommandAction.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace DeviceExplorer
+{
+    /// <summary>
+    /// Actions for C2D messages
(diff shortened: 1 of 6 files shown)


tameraw:
@Niyo - Thanks for providing this PR. We do not believe this is the right place for the feature at the moment. Currently the goal of this tool is not simulate a device, but rather debug and test against actual device code (using the samples from the different repos). If this objective changes in the future, we'll definitely consider adding a similar feature to the one you proposed here. Regards.

PR #135, opened 2017-07-12
implement twin links recovery and fix multiple calls to SafeAddClosed

jasmineymlo:
- implement twin amqp links recovery
- rename ConnectionKeys to ConnectionType
- fix multiple calls to SafeAddClosed for twin links


tameraw:
:shipit:

PR #136, opened 2017-07-14
fix amqp twin connection prefix

jasmineymlo:
- remove trailing slash on twin amqp connection path
- add null checking on incoming property for twin messages


anthonyvercolano:
:shipit:

tameraw:
:shipit:

PR #160, opened 2017-09-18
Adding DevDocs

ryan-z-johnson:


service/DevDocs/SendAsyncTimeout.md
```diff
@@ -0,0 +1,36 @@
...
+## References
+
+##Exposed API
+
+```csharp
+
+sealed class AmqpServiceClient : ServiceClient
+{
+    ...
+    public async override Task SendAsync(string deviceId, Message message, TimeSpan timeout)
```
CIPop: Should we pass a `CancellationToken` instead of a `TimeSpan` here?

service/DevDocs/SendAsyncTimeout.md
```diff
@@ -0,0 +1,36 @@
...
+public async override Task SendAsync(string deviceId, Message message, TimeSpan timeout)
+```
+
+### Result
+
+```csharp
+
+```
+
+### Excepetions
```
CIPop: Spelling: exceptions

service/DevDocs/SendAsyncTimeout.md
```diff
@@ -0,0 +1,36 @@
+# SendAsyncTimeout Requirements
+
+## Overview
+
+DeviceClient class is used by the device client callback delegate to prepare the callback results.
```
CIPop: Is this correct? The code below is referring to the AmqpServiceClient.

service/DevDocs/SendAsyncTimeout.md
```diff
@@ -0,0 +1,36 @@
...
+
+## References
+
+##Exposed API
```
CIPop: Looking at the MD preview, it seems like this isn't transforming in the correct header. I think we need to add maybe `###`?

service/DevDocs/SendAsyncTimeout.md
```diff
@@ -0,0 +1,36 @@
...
+
+```csharp
+
+sealed class AmqpServiceClient : ServiceClient
```
CIPop: Please specify public|internal|private.

service/DevDocs/SendAsyncTimeout.md
```diff
@@ -0,0 +1,36 @@
...
+{
+    ...
+    public async override Task SendAsync(string deviceId, Message message, TimeSpan timeout)
+}
+```
+
+### SendAsyncTimeout Thumbprint
+
+```csharp
+public async override Task SendAsync(string deviceId, Message message, TimeSpan timeout)
```
CIPop: Same as above: should we have a `CancellationToken` instead of `TimeSpan`?

service/DevDocs/SendAsyncTimeout.md
```diff
@@ -0,0 +1,36 @@
...
+```csharp
+
+sealed class AmqpServiceClient : ServiceClient
+{
+    ...
+    public async override Task SendAsync(string deviceId, Message message, TimeSpan timeout)
+}
+```
+
+### SendAsyncTimeout Thumbprint
```
CIPop: Not sure if "Thumbprint" is the right word here. Other reqs only mention the actual API name (in this case `SendAsync`).

service/DevDocs/SendAsyncTimeout.md
ryan-z-johnson: Copy and paste shrapnel. Will be fixed with next iteration

service/DevDocs/SendAsyncTimeout.md
ryan-z-johnson: The AMQP send message we call already takes a Timespan for timeout. Adding the cancellation token may make it harder than need be.

service/DevDocs/SendAsyncTimeout.md
ryan-z-johnson: The AMQP send message we call already takes a Timespan for timeout. Adding the cancellation token may make it harder than need be.

service/DevDocs/SendAsyncTimeout.md
```diff
@@ -0,0 +1,36 @@
...
+
+```csharp
+
+```
+
+### Excepetions
+
+TimeoutException - Response not recieved before timeout was reached.
+
+**SRS_SENDASYNCTIMEOUT_36_1480581: [** Method should throw an error if timeout reached before response is recieved **]**
```
tameraw: >response [](start = 94, length = 8)

What response?

service/DevDocs/SendAsyncTimeout.md
CIPop: I noticed that the `ServiceClient` object is using `CancellationToken` but not TimeSpan so it would be nice to be consistent.
That said, I do see your point: the AMQP library offers the timeout parameter and no cancellationToken. Unless there's a way to cancel that operation whenever the token is cancelled, we could use the TimeSpan in this case.

service/DevDocs/SendAsyncTimeout.md
ryan-z-johnson: good question. Removing.

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
```diff
@@ -127,15 +127,26 @@ public SendingAmqpLink SendingLink
...
                 {
                     SendingAmqpLink sendingLink = await this.GetSendingLinkAsync();
-                    outcome = await sendingLink.SendMessageAsync(amqpMessage, IotHubConnection.GetNextDeliveryTag(ref this.sendingDeliveryTag), AmqpConstants.NullBinary, this.OperationTimeout);
+                    if (timeout != null)
+                    {
+                        outcome = await sendingLink.SendMessageAsync(amqpMessage, IotHubConnection.GetNextDeliveryTag(ref this.sendingDeliveryTag), AmqpConstants.NullBinary, (TimeSpan)timeout);                        
+                    }
+                    else
+                    {
+                        outcome = await sendingLink.SendMessageAsync(amqpMessage, IotHubConnection.GetNextDeliveryTag(ref this.sendingDeliveryTag), AmqpConstants.NullBinary, this.OperationTimeout);
```
CIPop: @ryan-z-johnson please add another test that pins down this behavior. (i.e. set `this.OperationTimeout` to something small. Calling `SendMessageAsync` should have the same behavior as calling it with a non-null timeout).

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
```diff
@@ -153,7 +164,7 @@ public SendingAmqpLink SendingLink
                 throw AmqpErrorMapper.GetExceptionFromOutcome(outcome);
             }
         }
-
+
```
CIPop: Nit: please remove the extra white-space on this empty line.

service/Microsoft.Azure.Devices/ServiceClient.cs
```diff
@@ -88,8 +88,9 @@ public static ServiceClient CreateFromConnectionString(string connectionString,
         /// </summary>
         /// <param name="deviceId">The device identifier for the target device</param>
         /// <param name="message">The message containing the notification</param>
+        /// <param name="timeout">The operation timeout</param>
```
CIPop: For AMQP this is actually the operation timeout override. If not used, the `OperationTimeout` is used instead. Docs should point that out.

service/DevDocs/ServiceClient.md
```diff
@@ -0,0 +1,30 @@
+# ServiceClient Requirements
+
+## Overview
+
+ServiceClient class is used by the Service client callback delegate to prepare the callback results.
+
+### References
```
CIPop: @tameraw @jasmineymlo - I see 3 dev-docs having this References but none actually filling it. What should go here? (We can still have a template checked in containing it, explaining what each of these fields mean.)

Maybe we can add a link to the AMQP SendAsync API that we're passing this through? If not, we can just delete it.

e2e/Microsoft.Azure.Devices.E2ETests/MessageE2ETests.cs
```diff
@@ -566,6 +565,35 @@ public static void ClassCleanup()
...
+        [TestCategory("Message-E2E")]
+        [TestCategory("Timeout")]
+        public async Task Message_TimeOutReachedResponse()
+        {
+            Exception error = null;
+            try
+            {
+                await SlowResponseToMessage();
+            }
+            catch (AggregateException e)
```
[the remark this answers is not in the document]
ryan-z-johnson: Because of mixing async and non-async as you pointed out below.

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
ryan-z-johnson: This.OperationTimeout is not settable, so won't be able to test changing that to something small without changing the source code. I have added a test for the SendMessageAsync with a null timeout. @CIPop

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
CIPop: Nit: @ryan-z-johnson you may have forgotten to remove this ws :) - or GH has a bug(?)

e2e/Microsoft.Azure.Devices.E2ETests/MessageE2ETests.cs
```diff
@@ -566,6 +565,42 @@ public static void ClassCleanup()
...
+        [ExpectedException(typeof(TimeoutException))]
+        public async Task Message_TimeOutReachedResponse()
+        {
+            await FastTimeout();
+        }
+
+        [TestMethod]
+        [TestCategory("Message-E2E")]
+        [TestCategory("Timeout")]
+        // Should not have any exceptions thrown.
```
CIPop: @ryan-z-johnson, shouldn't this timeout after [100 seconds](https://github.com/Azure/azure-iot-sdk-csharp/blob/9272857f1065585f2c4804dd5836b0ea95584ee9/service/Microsoft.Azure.Devices/AmqpServiceClient.cs#L20)?

CIPop: Or is it because the server responds well within that time?

service/DevDocs/ServiceClient.md
tameraw: Could probably point a link to the public spec.

---
In reply to: [140533683](https://github.com/Azure/azure-iot-sdk-csharp/pull/160#discussion_r140533683) [](ancestors = 140533683)

service/DevDocs/ServiceClient.md
```diff
@@ -0,0 +1,30 @@
...
+}
+```
+
+### SendAsync
+
+```csharp
+public async override Task SendAsync(string deviceId, Message message, TimeSpan? timeout = null)
+```
+
+**SRS_ServiceClient_36_1480581: [** Method should throw an error if timeout reached before response is recieved **]**
```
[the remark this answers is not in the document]
tameraw: That's correct. That's the requirement.
It should then show up in the code & the test. And that's how we'd trace that requirements are implemented and tested.

---
In reply to: [140531965](https://github.com/Azure/azure-iot-sdk-csharp/pull/160#discussion_r140531965) [](ancestors = 140531965)

service/Microsoft.Azure.Devices/AmqpServiceClient.cs
ryan-z-johnson: I think it is a bug. The only extra line I have in that area is one line between the two functions

e2e/Microsoft.Azure.Devices.E2ETests/MessageE2ETests.cs
ryan-z-johnson: Responds well within that time.

PR #166, opened 2017-09-21
Provisioning - build setup

jasmineymlo:
To set up build scripts and directory structures for Provisioning clients

### DIFF
--- a/device/build/build.cmd
+++ b/device/build/build.cmd
@@ -35,7 +35,7 @@ rem // default build options
 set build-clean=0
 set build-config=Release
 set build-platform="Any CPU"
-set wip-use-c2d-amqp-methods=0
+set wip-provisioning=0
 
 :args-loop
 if "%1" equ "" goto args-done
@@ -43,7 +43,7 @@ if "%1" equ "-c" goto arg-build-clean
 if "%1" equ "--clean" goto arg-build-clean
 if "%1" equ "--config" goto arg-build-config
 if "%1" equ "--platform" goto arg-build-platform
-if "%1" equ "--wip-use-c2d-amqp-methods" goto arg-wip-use-c2d-amqp-methods
+if "%1" equ "--wip-provisioning" goto arg-wip-provisioning
 call :usage && exit /b 1
 
 :arg-build-clean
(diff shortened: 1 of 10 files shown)


device/build/build.cmd
```diff
@@ -35,15 +35,15 @@ rem // default build options
 set build-clean=0
 set build-config=Release
 set build-platform="Any CPU"
-set wip-use-c2d-amqp-methods=0
```
ryan-z-johnson: Not sure why we are removing all the use-c2d-amqp-methods calls.

e2e/build/build.cmd
```diff
@@ -35,15 +35,15 @@ rem // default build options
 set build-clean=0
 set build-config=Release
 set build-platform="Any CPU"
-set wip-use-c2d-amqp-methods=0
```
ryan-z-johnson: same here, why remove all use-c2d-amqp-methods calls

device/build/build.cmd
jasmineymlo: methods over amqp is released.  all corresponding compile flags are removed and I am just cleaning them up.

provisioning/device/build/build.cmd
```diff
@@ -0,0 +1,134 @@
...
+
+set current-path=%~dp0
+rem // remove trailing slash
+set current-path=%current-path:~0,-1%
+
+set build-root=%current-path%\..\..\..
+rem // resolve to fully qualified path
+for %%i in ("%build-root%") do set build-root=%%~fi
+
+rem ensure nuget.exe exists
```
CIPop: For the new builds we may want to switch to use the dotnet tools instead of VS.
dotnet should perform the nuget tasks as well.

That said, we should merge this to get the folder layout started.

provisioning/device/build/build.cmd
```diff
@@ -0,0 +1,134 @@
...
+shift
+goto args-loop
+
+:args-done
+
+rem -----------------------------------------------------------------------------
+rem -- build csharp provisioning client
+rem -----------------------------------------------------------------------------
+
+call nuget restore -config "%current-path%\NuGet.Config" "%build-root%\provisioning\device\provisioning_deviceclient.sln"
```
CIPop: Nit: we'll need to rethink the bat file layout - there is a lot of code duplication currently. (Maybe switching to powershell would allow us to use both Windows and Linux builds?)

PR #173, opened 2017-09-28
Adding certificate validation tool.

CIPop:
Provisioning Service certificate validation tool for .Net.
I wasn't able to find a BouncyCastle crypto lib released for NetStandard2.0 (@DMattoon, do you know of anything that we could use instead?)

### DIFF
--- a/provisioning/service/samples/GroupCertificateVerificationSample/GroupCertificateVerificationSample.csproj
+++ b/provisioning/service/samples/GroupCertificateVerificationSample/GroupCertificateVerificationSample.csproj
@@ -1,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BouncyCastle.Crypto.dll" Version="1.8.1" />
+  </ItemGroup>
+
+</Project>
--- a/provisioning/service/samples/GroupCertificateVerificationSample/GroupCertificateVerificationSample.csproj
+++ b/provisioning/service/samples/GroupCertificateVerificationSample/GroupCertificateVerificationSample.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="BouncyCastle.Crypto.dll" Version="1.8.1" />
+    <PackageReference Include="BouncyCastle" Version="1.8.1" />
   </ItemGroup>
 
 </Project>
--- a/provisioning/service/samples/GroupCertificateVerificationSample/Program.cs
+++ b/provisioning/service/samples/GroupCertificateVerificationSample/Program.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Azure.Devices.Provisioning.Samples
+{
+    public class Program
+    {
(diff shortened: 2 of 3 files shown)


provisioning/service/samples/GroupCertificateVerificationSample/VerificationCertificateGenerator.cs
```diff
@@ -0,0 +1,106 @@
...
+                if (key != null && key.Key.IsPrivate)
+                {
+                    privateParameter = key.Key as ECPrivateKeyParameters;
+                }
+            }
+
+            return privateParameter;
+        }
+
+        private static X509Certificate2 GenerateAndSignCertificate(
```
simonporter: A new requirement I know but we should actually break this into generating a CSR and consuming it. This is how it really should happen and makes this extensible to using an actual CSR service.

provisioning/service/samples/GroupCertificateVerificationSample/GroupCertificateVerificationSample.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BouncyCastle.Crypto.dll" Version="1.8.1" />
```
alextolp: >Include [](start = 22, length = 7)

nit: not sure if we should use the one you listed or this one: [https://www.nuget.org/packages/BouncyCastle/](https://www.nuget.org/packages/BouncyCastle/)

the owner of this pkg is listed as Bouncy Castle Project Contributors

CIPop: Thanks @alextolp: I was actually looking into this right now. Changed to the official one.

provisioning/service/samples/GroupCertificateVerificationSample/VerificationCertificateGenerator.cs
CIPop: While that may be the way a real CA or Intermediate CA would do it, this is just providing a way for the developer to bootstrap their development process. 

On Windows, the methods I'm aware of are via a [wizard](https://technet.microsoft.com/en-us/library/cc754490(v=ws.11).aspx) or via [certreq](https://technet.microsoft.com/en-us/library/dn296456(v=ws.11).aspx).

simonporter: This serves the dev bootstrap scenario well, we can produce CSR based samples as a more formal way of doing this.

provisioning/service/samples/GroupCertificateVerificationSample/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+using System.IO;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Azure.Devices.Provisioning.Samples
+{
+    public class Program
+    {
+        private const string VerificationCertificatePath = "verificationCertificate.cer";
```
jasmineymlo: Please add <> in the verification cert path to indicate that users have to replace with their own values.  Our other samples have something like:
private const string DeviceConnectionString = "<replace>";

CIPop:
Thanks for reviewing @alextolp @jasmineymlo @simonporter !

provisioning/service/samples/GroupCertificateVerificationSample/VerificationCertificateGenerator.cs
CIPop: I'll start an internal thread to have this added to the documentation on OSs that we support.

CIPop: @yzhong94 PTAL

Issue #181, opened 2017-10-09
Enable FXCop for existing projects


CIPop:
We would like to enable the following for Hub projects:
- [ ] Treat warnings as errors (`<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` )
- [ ] FXCop: (`<Features Condition=" '$(Configuration)' == 'Debug' ">IOperation</Features>` and `<PackageReference Condition=" '$(Configuration)' == 'Debug' " Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.3.0-beta1" />`)
- [ ] SDL and CodeScanning Plugins to VSTS

This requires investigating and fixing all existing warnings.

sharmasejal:
All our projects now have FXCop enabled.

az-iot-builder-01:
@CIPop, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #182, opened 2017-10-09
Enabling FXCop for new projects.

CIPop:
Enabling FXCop for all new projects. Old projects have warnings and cannot have FXCop enabled easily at this moment (tracked by #181).


jasmineymlo:
Question:  Are we not enabling fxcop in the shared project?

CIPop:
> Question: Are we not enabling fxcop in the shared project?

We should but it will take quite some time to address all issues. Starting with the new projects and tracking remaining projects with #181.

PR #191, opened 2017-10-16
WIP: HTTP device client implementation.

CIPop:
Work in progress, missing Unit and E2E tests.

* API changes: 
  * Transport is particular to the Provisioning service
  * Renamed: ProvisioningSecurityClient is used only by the Provisioning service
* ProvisioningTransportClient + HttpTransportHandler
* ProvisioningSecurityClient + TPM and TPM Emulator


provisioning/transport/http/src/ApiVersionDelegatingHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Http
+{
+    internal class ApiVersionDelegatingHandler : DelegatingHandler
+    {
+        private const string ApiVersion = "2017-08-31-preview";
```
jasmineymlo: api version should be common to protocols.  We have a common file for that....TODO for cleanup

provisioning/transport/http/src/Generated/DeviceProvisioningServiceRuntimeClient.cs
```diff
@@ -0,0 +1,247 @@
+// Copyright (c) Microsoft. All rights reserved.
```
jasmineymlo: Do we keep the "Generated" folder?

provisioning/transport/http/src/Generated/Models/TpmAttestation.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Http.Models
```
jasmineymlo: >Http [](start = 64, length = 4)

should the attestation class be under non transport specific namespace?

CIPop: This is HTTP/JSON specific and so AMQP/MQTT shouldn't require this.

provisioning/transport/http/src/Generated/DeviceProvisioningServiceRuntimeClient.cs
CIPop: This code was generated initially then I've fixed all FXCop issues and formatted using our coding standard.
Unless there's a good reason, we could keep this as-is in case we need to pick-up & regen when the service changes in the future.

CIPop:
Thanks @ryan-z-johnson , @jasmineymlo  for the review!

PR #202, opened 2017-10-23
Add ability to append user agent string from DeviceClient

damonbarry:
Consumers of the device SDK need to be able to append the user agent string that is sent to IoT Hub. This change adds the property `DeviceClient.ProductInfo` for this purpose. `DeviceClient` stills sends its own name/version at the beginning of the user agent string, but then it:
1. appends a comment that provides some more detail: "(runtime; operating system; architecture)"
2. appends the value of `ProductInfo` to the end.

The user agent string is intended to conform to User-Agent syntax per [RFC 7231 section 5.5.3](https://tools.ietf.org/html/rfc7231#section-5.5.3), although the client doesn't do any validation--that should happen in the service.

Additionally, there were some inconsistencies in user-agent string handling that are fixed here. Specifically:
- The version of the Microsoft.Azure.Devices assembly (service SDK) reported in the user-agent string is 1.3.2; updated it to report the current version (1.4.0).
- The device SDK user-agent string was generated in several different places (especially in the AMQP case); consolidated into a single location (`ProductInfo.ToString()`).
- In the device SDK's AMQP transport, there are #ifdef sections that grabbed different version strings for UWP, PCL, etc. Since the bump_version script only differentiates between (1) PCL and (2) everything else, I removed the UWP-specific handling.
- In the device SDK's HTTP transport, we apparently weren't sending the User-Agent header at all in the UWP and PCL cases, but we are for the other transports. Since I can't think of a reason why we would support User-Agent in UWP and PCL for AMQP and MQTT but not HTTP, I removed the #ifdef.

Other changes:
- Removed some unnecessary constructor overrides for `HttpTransportHandler`. They appear to be the earliest versions of the `HttpTransportHandler` ctor--before some newer features were added. It was easier to remove them than to retrofit them with the new `ProductInfo` argument.
- Removed what appears to be an unused (and out-of-sync) copy of `HttpClientHelper.cs`.


device/Microsoft.Azure.Devices.Client/ProductInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    class ProductInfo
+    {
+        public string Extra { get; set; } = "";
+
+        public override string ToString()
+        {
+            string userAgent = "Microsoft.Azure.Devices.Client/";
+#if PCL
+            // DO NOT EDIT the following line; it is updated by the bump_version script (https://github.com/Azure/iot-sdks-internals/blob/master/release/csharp/inputs.js)
```
damonbarry: I'll update the bump_version script to search/replace in this file instead of previous locations.

damonbarry:
BTW, besides the new unit tests, I also tested this manually for .NET Framework (all protocols), UWP (all protocols), and .NET Standard (.NET Core, MQTT only). In all cases I've confirmed from the service side that the expected user-agent strings are coming through.

device/Microsoft.Azure.Devices.Client/ProductInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+    class ProductInfo
+    {
+        public string Extra { get; set; } = "";
+
+        public override string ToString()
+        {
+            const string name = "Microsoft.Azure.Devices.Client";
+#if PCL
+            // DO NOT EDIT the following line; it is updated by the bump_version script (https://github.com/Azure/iot-sdks-internals/blob/master/release/csharp/inputs.js)
+            const string version = "1.5.1"; // PCLAssemblyVersion
```
damonbarry: I'll update the bump_version script to search/replace in the updated file location, using an updated regex.

device/Microsoft.Azure.Devices.Client/ProductInfo.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    class ProductInfo
```
CIPop: Nit: please add `internal`.

device/Microsoft.Azure.Devices.Client/ProductInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        public string Extra { get; set; } = "";
+
+        public override string ToString()
+        {
+            const string name = "Microsoft.Azure.Devices.Client";
```
CIPop: Nit: for new code we're following https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md 
`private const string Name = ...`

CIPop: Same with other places.

damonbarry:
I just squashed + rebased on the latest master... @CIPop @jasmineymlo

device/tests/Microsoft.Azure.Devices.Client.Test/DeviceClientTests.cs
```diff
@@ -1147,5 +1147,15 @@ public void DeviceClient_OnConnectionClosed_RecoverThrowsException_InvokeHandler
             Assert.AreEqual(ConnectionStatus.Disabled, status);
             Assert.AreEqual(ConnectionStatusChangeReason.Client_Close, statusChangeReason);
         }
+
+        [TestMethod]
+        [TestCategory("DeviceClient")]
+        public void DeviceClient_ProductInfo_stores_product_info()
```
CIPop: Nit:
- Naming should be of form: <Unit>_<Action>_<Expected result>. E.g. `ProductInfo_StoresProductInfo_Ok`.
- Each word should be capitalized.

device/Microsoft.Azure.Devices.Client/ProductInfo.cs
damonbarry: Will do. In this case `name` is a variable in a method, so `private` doesn't make sense, but I get what you're saying.

device/tests/Microsoft.Azure.Devices.Client.Test/DeviceClientTests.cs
damonbarry: Will do. I find this test naming scheme bewildering...is there a doc somewhere that explains it?

device/Microsoft.Azure.Devices.Client/ProductInfo.cs
damonbarry: Will do

device/Microsoft.Azure.Devices.Client/ProductInfo.cs
CIPop: Ah, sorry... missed that.

CIPop:
@damonbarry you don't need to close this one - if you have more changes, just push to the same branch a new check-in. That way we get to review only the diff.
When you're done with the PR, please squash before submitting to the CI system.

PR #204, opened 2017-10-24
draft for amqp dps for api review

jasmineymlo:


provisioning/transport/amqp/src/AmqpClientSession.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public bool IsSessionClosed => _isSessionClosed;
+
+        public async Task OpenAsync(TimeSpan timeout)
+        {
+            // Create the Session
+            var amqpTestLinkFactory = new AmqpLinkFactory();
+            amqpTestLinkFactory.LinkCreated += this.OnLinkCreated;
+            this.AmqpSession = new AmqpSession(this.amqpConnection.AmqpConnection, this.AmqpSessionSettings, amqpTestLinkFactory);
+            this.amqpConnection.AmqpConnection.AddSession(this.AmqpSession, new ushort?());
```
CIPop: I believe that `default(ushort?)` is `null`. Since we're not passing an actual value, we could just pass null for the channel?

provisioning/transport/amqp/src/AmqpClientSession.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                await session.CloseAsync(timeout).ConfigureAwait(false);
+            }
+        }
+
+        public AmqpClientLink CreateSendingLink(Address address)
+        {
+            this.SendingLink = CreateLink();
+            this.SendingLink.AmqpLinkSettings.SettleType = SettleMode.SettleOnSend;
+            this.SendingLink.AmqpLinkSettings.Role = false; // i.e. IsReceiver = false
```
CIPop: Not sure I understand the comment here. Do you mean to say Role == false means that this link is a Sender?

provisioning/transport/amqp/src/AmqpClientSession.cs
```diff
@@ -0,0 +1,94 @@
...
+            };
+
+            return this.SendingLink;
+        }
+
+        public AmqpClientLink CreateReceivingLink(Address address)
+        {
+            this.ReceivingLink = CreateLink();
+            this.ReceivingLink.AmqpLinkSettings.SettleType = SettleMode.SettleOnDispose;
+            this.ReceivingLink.AmqpLinkSettings.Role = true; // i.e. IsReceiver = true
```
CIPop: Same comment as above.

provisioning/transport/amqp/src/AmqpLinkFactory.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.Amqp;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    class AmqpLinkFactory : ILinkFactory
+    {
+        public IAsyncResult BeginOpenLink(AmqpLink link, TimeSpan timeout, AsyncCallback callback, object state)
```
CIPop: Not used anywhere - I think we can delete.

provisioning/transport/amqp/src/AmqpLinkFactory.cs
```diff
@@ -0,0 +1,56 @@
...
+            }
+            else
+            {
+                link = new SendingAmqpLink(session, settings);
+            }
+            this.OnLinkCreated(link);
+            return link;
+        }
+
+        public void EndOpenLink(IAsyncResult result)
```
CIPop: Not used.

provisioning/transport/amqp/src/LinkCreatedEventArgs.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Microsoft.Azure.Amqp;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    public class LinkCreatedEventArgs : EventArgs
```
CIPop: I'm not sure but it looks like this isn't needed?

provisioning/transport/amqp/src/AmqpTransportClient.cs
```diff
@@ -1,25 +1,84 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.Azure.Devices.Provisioning.Client;
+using System;
 using Microsoft.Azure.Devices.Shared;
 using System.Threading.Tasks;
 using System.Threading;
+using Microsoft.Azure.Devices.Provisioning.Client.Transport.Http.Models;
```
CIPop: Nit: sort `using` statements.

provisioning/transport/amqp/src/AmqpTransportClient.cs
```diff
@@ -1,25 +1,84 @@
...
 
-        public override Task<ProvisioningRegistrationResult> RegisterAsync(string globalDeviceEndpoint, string idScope, ProvisioningSecurityClient securityClient, CancellationToken cancellationToken)
+        public override async Task<ProvisioningRegistrationResult> RegisterAsync(
+            string globalDeviceEndpoint, 
+            string idScope, 
+            ProvisioningSecurityClient securityClient, 
+            CancellationToken cancellationToken)
         {
-            throw new System.NotImplementedException();
+            AmqpClientConnection client;
```
CIPop: I know this is work-in-progress but have you considered how multiplexing using a single client across multiple `ProvisioningClient`s would work?
Is it possible with the current structure? (I believe we may need an implicit or explicit OpenAsync that will create the first link?)

provisioning/transport/amqp/src/AmqpTransportHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+using Microsoft.Azure.Amqp.Transport;
+using Microsoft.Azure.Devices.Provisioning.Client.Transport.Http.Models;
+using Microsoft.Azure.Devices.Shared;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    internal static class AmqpTransportHandler
+    {
+        public const string clientVersion = "test-client-version";
```
CIPop: TODO: let's have this in common - I think the HTTP one exposes at least the apiVersion.

provisioning/transport/amqp/src/AmqpTransportHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+using Microsoft.Azure.Amqp;
+using Microsoft.Azure.Amqp.Sasl;
+using Microsoft.Azure.Amqp.Transport;
+using Microsoft.Azure.Devices.Provisioning.Client.Transport.Http.Models;
+using Microsoft.Azure.Devices.Shared;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    internal static class AmqpTransportHandler
```
CIPop: You may want to consider a template class where AmqpTransportHandler implements the high-level logic and we have TPM and X509 separate implementations.

provisioning/transport/amqp/src/AmqpTransportHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+            var settings = new AmqpSettings();
+            byte[] ekBuffer = await securityClient.GetEndorsementKeyAsync().ConfigureAwait(false);
+            byte[] srkBuffer = await securityClient.GetStorageRootKeyAsync().ConfigureAwait(false);
+
+            string ek = Convert.ToBase64String(ekBuffer);
+            string srk = Convert.ToBase64String(srkBuffer);
+
+            TpmNetworkCredential credential = new TpmNetworkCredential(ek, srk);
+
+                var saslProvider = new SaslTransportProvider();
```
CIPop: Nit: spacing (this portion is indented).

provisioning/transport/amqp/src/AmqpTransportHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+        else
+            {
+                amqpCloudConnection = await CreateConnection(
+                    new Uri("amqps://" + deviceEndpoint + ":" + AmqpConstants.DefaultSecurePort), 
+                    securityClient, 
+                    linkEndpoint).ConfigureAwait(false);
+            }
+
+            await amqpCloudConnection.OpenAsync(TimeSpan.FromMinutes(1), useWebSocket, clientCert).ConfigureAwait(false);
```
CIPop: Please extract the TimeSpan as a const. We should consider exposing these timings on the AmqpTransportHandler public API.

provisioning/transport/amqp/src/AmqpTransportHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+            amqpSendingLink.AddApiVersion(apiVersion);
+
+            await amqpSendingLink.OpenAsync(TimeSpan.FromMinutes(1)).ConfigureAwait(false);
+
+            return amqpCloudConnection;
+        }
+
+        public static async Task<RegistrationOperationStatus> RegisterDeviceAsync(AmqpClientConnection client, string correlationId)
+        {
+            var amqpMessage = AmqpMessage.Create(new MemoryStream(Encoding.ASCII.GetBytes(DeviceOperations.Register)), false);
```
CIPop: Seems like overhead that we need to start the message with the "iotdps-register" string then add the same string within the properties.

provisioning/transport/amqp/src/AmqpTransportHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+            amqpMessage.Properties.CorrelationId = correlationId;
+            amqpMessage.ApplicationProperties.Map[MessageApplicationPropertyNames.OperationType] = DeviceOperations.Register;
+            amqpMessage.ApplicationProperties.Map[MessageApplicationPropertyNames.ForceRegistration] = false;
+            var outcome = await client.AmqpSession.SendingLink.SendMessageAsync(amqpMessage, new ArraySegment<byte>(Guid.NewGuid().ToByteArray()), TimeSpan.FromMinutes(1)).ConfigureAwait(false);
```
CIPop: From the AMQP source code this function requires 2 ArraySegment structs: deliveryTag and txnId. Are we using an older version?

provisioning/transport/amqp/src/CrytoKeyGenerator.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    public static class CryptoKeyGenerator
+    {
+        // Function to Generate a 64 bits Key.
+        public static string GenerateKey(int maxNumber)
+        {
+            var builder = new StringBuilder();
+
+            //// TODO v-renesc 2016-08-15 Discuss with Russell... this code uses a non-threadsafe and cryptographically weak random number generator.
```
CIPop: Not used anyway - please remove.

provisioning/transport/amqp/src/LinkCreatedEventArgs.cs
CIPop: Same as above - not sure this is used anywhere.

provisioning/transport/amqp/src/Microsoft.Azure.Devices.Provisioning.Transport.Amqp.csproj
```diff
@@ -13,6 +21,7 @@
 
   <ItemGroup>
     <!-- FXCop -->
+    <PackageReference Include="Microsoft.Azure.Amqp" Version="2.0.6" />
```
CIPop: Please add a TODO: + issue that we should be switching to the latest Amqp version (right now it's not working for the client so we can't since we can expect both Provisioning and IoT Hub client in the same project).

provisioning/transport/amqp/src/ParameterNames.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    internal static class ParameterNames
+    {
+        public const string ForceRegistration = "forceRegistration";
```
CIPop: This is used in a single other place - any reason for this being declared in a different class?

provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+                case SaslChallengeAction.Final:
+                    SaveEncodedNonceSegment(challenge);
+                    SendLastResponse();
+                    break;
+            }
+        }
+
+        private void SendInterimResponse()
+        {
+            var response = new SaslResponse {Response = new ArraySegment<byte>(new byte[1] {0})};
```
CIPop: Please add a constant if this single byte is never changed or read otherwise this is going to use a lot of memory.

provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+            Negotiator.WriteFrame(response, true);
+        }
+
+        private async void SendLastResponse()
+        {
+            //Notes: The _encodedNonce from service would soon change to non base64 string
+            byte [] sasBytes = await _security.SignAsync(Convert.FromBase64String(_encodedNonceStringBuilder.ToString())).ConfigureAwait(false);
+            string sas = Convert.ToBase64String(sasBytes);
+
+            var responseBuffer = new byte [sas.Length * sizeof(char) + 1];
```
CIPop: sizeof(char) is 2 but you are encoding to UTF8 below which is 1 byte / character

provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+            responseBuffer[0] = 0x0;
+            Buffer.BlockCopy(Encoding.UTF8.GetBytes(sas), 0, responseBuffer, 1, sas.Length);
+
+            var response = new SaslResponse { Response = new ArraySegment<byte>(responseBuffer) };
+            Negotiator.WriteFrame(response, true);
+
```
CIPop: Nit: extra NL.

provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+            }
+
+            throw new AmqpException(AmqpErrorCode.InvalidField,
+                $"Sasl challenge control byte contains invalid data: {challengeBytes[0]:X}.");
+        }
+
+        public override void OnResponse(SaslResponse response)
+        {
+            //This method is only implemented on the server
+            var exception = new AmqpException(AmqpErrorCode.NotImplemented,
```
CIPop: Debug.Fail instead of throwing. Since it's not user actionable, I would hope the device doesn't crash because this got called for some reason by AMQP.

provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+               "OnResponse is not implemented on client");
+            throw exception;
+        }
+
+        protected override void OnStart(SaslInit init, bool isClient)
+        {
+            if (isClient)
```
CIPop: We should assert that isClient == true?

provisioning/transport/amqp/src/StringFormattingExtensions.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Globalization;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    public static class StringFormattingExtensions
```
CIPop: Not used anywhere - please remove.

provisioning/transport/amqp/src/AmqpLinkFactory.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                throw new ArgumentNullException(nameof(link));
+            }
+
+            if (timeout.TotalMilliseconds > 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(timeout));
+            }
+
+            return TaskConstants.Completed;
```
CIPop: Task already has a [CompletedTask](https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.completedtask(v=vs.110).aspx). Please use that instead.

provisioning/transport/amqp/src/TaskConstants.cs
```diff
@@ -0,0 +1,91 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    public static class TaskConstants
```
CIPop: Not used except for the CompletedTask. See the other comment.
Please remove.

provisioning/transport/amqp/src/ClientWebSocketTransport.cs
```diff
@@ -0,0 +1,335 @@
...
+
+            return true;
+        }
+
+        protected override bool CloseInternal()
+        {
+            var webSocketState = this.webSocket.State;
+            if (webSocketState != WebSocketState.Closed && webSocketState != WebSocketState.Aborted)
+            {
+                this.CloseInternalAsync(CloseTimeout).Fork();
```
CIPop: What does Fork() do here?

provisioning/transport/amqp/src/TaskHelpers.cs
```diff
@@ -0,0 +1,89 @@
...
+
+using System;
+using System.Diagnostics.Contracts;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    public static class TaskHelpers
```
CIPop: Not sure if we need this - see other comment regarding Fork().
If we don't we should remove it.

provisioning/transport/amqp/src/AmqpTransportHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+        private static async Task<AmqpSettings> CreateAmqpSettings(ProvisioningSecurityClientSasToken securityClient, string linkendpoint)
+        {
+            var settings = new AmqpSettings();
+            byte[] ekBuffer = await securityClient.GetEndorsementKeyAsync().ConfigureAwait(false);
+            byte[] srkBuffer = await securityClient.GetStorageRootKeyAsync().ConfigureAwait(false);
+
+            string ek = Convert.ToBase64String(ekBuffer);
+            string srk = Convert.ToBase64String(srkBuffer);
+
+            TpmNetworkCredential credential = new TpmNetworkCredential(ek, srk);
```
CIPop: `credential` doesn't seem to be referenced.

provisioning/transport/amqp/src/TpmNetworkCredential.cs
```diff
@@ -0,0 +1,16 @@
+using System.Net;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    public class TpmNetworkCredential : NetworkCredential
```
CIPop: Used by an unreferenced variable. If not needed, please remove.

provisioning/transport/amqp/src/ClientWebSocketTransport.cs
```diff
@@ -0,0 +1,335 @@
...
+                    this.Abort();
+                }
+            }
+        }
+
+        public override bool ReadAsync(TransportAsyncCallbackArgs args)
+        {
+            this.ThrowIfNotOpen();
+
+            Utils.ValidateBufferBounds(args.Buffer, args.Offset, args.Count);
```
CIPop: This is an internal function - we should assert instead of throwing using this validation util.
We should also try to use ArraySegment whenever possible to avoid needing to do these checks ourselves.

provisioning/transport/amqp/src/Utils.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
+{
+    static class Utils
```
CIPop: Used only by the ClientWebSocket - we should switch to ArraySegment and remove this class.

provisioning/transport/amqp/src/AmqpLinkFactory.cs
jasmineymlo: interface member

---
In reply to: [146718122](https://github.com/Azure/azure-iot-sdk-csharp/pull/204#discussion_r146718122) [](ancestors = 146718122)

provisioning/transport/amqp/src/AmqpLinkFactory.cs
jasmineymlo: interface member

---
In reply to: [146718068](https://github.com/Azure/azure-iot-sdk-csharp/pull/204#discussion_r146718068) [](ancestors = 146718068)

provisioning/transport/amqp/src/AmqpTransportClient.cs
jasmineymlo: Service side is not supporting multiplexing and there is no use case for now.  Multiplexing would not be possible for x509.

provisioning/transport/amqp/src/AmqpTransportHandler.cs
jasmineymlo: Refer to AmqpClientLink.SendMessageAsyn().  We use AmqpConstants.NullBinary as the txnId.

---
In reply to: [146720881](https://github.com/Azure/azure-iot-sdk-csharp/pull/204#discussion_r146720881) [](ancestors = 146720881)

provisioning/transport/amqp/src/ClientWebSocketTransport.cs
jasmineymlo: used Task.Run()

---
In reply to: [146723519](https://github.com/Azure/azure-iot-sdk-csharp/pull/204#discussion_r146723519) [](ancestors = 146723519)

PR #214, opened 2017-11-02
update sasl tpm handshake

jasmineymlo:
update  the AMQP SASL TPM flow which includes:
- initial SASL Init packet to contain the IdScope and the Registration Id of the device
- challenge key sent by the service is no longer base-64 encoded

### DIFF
--- a/provisioning/transport/amqp/src/SaslTpmHandler.cs
+++ b/provisioning/transport/amqp/src/SaslTpmHandler.cs
@@ -14,26 +14,26 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Amqp
     {
         static readonly byte[] EmptyByte = { 0 };
         private const string MechanismName = "TPM";
-        private const int ChallengeKeySegmentSize = 400;
-        private readonly StringBuilder _encodedNonceStringBuilder = new StringBuilder(ChallengeKeySegmentSize * 3);
 
         private readonly byte[] _endorsementKey;
 
-        private readonly string _hostName;
+        private readonly string _idScope;
         private readonly ProvisioningSecurityClientSasToken _security;
         private readonly byte[] _storageRootKey;
+        private byte[] _nonceBuffer = Array.Empty<byte>();
         private byte _nextSequenceNumber;
 
-        public SaslTpmHandler(byte[] endorsementKey, byte[] storageRootKey, string hostName,
-            ProvisioningSecurityClientSasToken security)
+        private string _hostname => $"{this._idScope}/registrations/{_security.RegistrationID}";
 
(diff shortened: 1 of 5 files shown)


provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -14,36 +14,35 @@ public class SaslTpmHandler : SaslHandler
...
 
             Mechanism = MechanismName;
             _endorsementKey = endorsementKey;
             _storageRootKey = storageRootKey;
-            _hostName = hostName;
+            _idScope = idScope;
             _security = security;
         }
 
+        private string HostName => $"{this._idScope}/registrations/{_security.RegistrationID}";
```
CIPop: I think this doesn't look like a hostname anymore. It should be renamed (`_saslScope` maybe?)
Also: please follow the naming conventions and the placing of member declarations (at the top, next to the other private members).

provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -14,36 +14,35 @@ public class SaslTpmHandler : SaslHandler
...
+
         protected bool Equals(SaslTpmHandler other)
         {
             return
                 Equals(_endorsementKey, other._endorsementKey) &&
                 Equals(_storageRootKey, other._storageRootKey) &&
-                string.Equals(_hostName, other._hostName, StringComparison.InvariantCulture) &&
-                Equals(_encodedNonceStringBuilder, other._encodedNonceStringBuilder) &&
+                string.Equals(_idScope, other._idScope, StringComparison.InvariantCulture) &&
```
CIPop: I think you want to use `string.CompareOrdinal` instead of string.Equals.

provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -61,8 +60,7 @@ public override int GetHashCode()
             {
                 var hashCode = _endorsementKey != null ? _endorsementKey.GetHashCode() : 0;
```
CIPop: Please add the type if not present on the same line.
i.e. var myVar = new MyType() is OK because the type is written in the same statement.

provisioning/transport/amqp/src/SaslTpmHandler.cs
jasmineymlo: it is the value assigned to init.hostname, I will keep the same name (_hostname).

---
In reply to: [148645292](https://github.com/Azure/azure-iot-sdk-csharp/pull/214#discussion_r148645292) [](ancestors = 148645292)

jasmineymlo:
Refer to #228.  Closing this one

PR #221, opened 2017-11-08
Updating LTS information.

CIPop:
The current LTS state is inaccurate. Updating based on the last git hash.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -151,7 +151,7 @@ Below is a table showing the mapping of the LTS branches to the packages release
 
 | Nuget Package | Github Branch | LTS Status | LTS Start Date | Maintenance End Date | Removed Date |
 | :-----------: | :-----------: | :--------: | :------------: | :------------------: | :----------: |
-| 1.x.x         | lts_07_2017   | Active     | 2017-07-01     | 2017-12-31           | 2018-06-30   |
+| Microsoft.Azure.Devices.Client v1.5.1, Microsoft.Azure.Devices v1.4.0, Microsoft.Azure.Devices.Shared v1.1.1 | lts_07_2017   | Active     | 2017-07-01     | 2017-12-31           | 2018-06-30   |
 
 * <sup>1</sup> All scheduled dates are subject to change by the Azure IoT SDK team.
 


readme.md
```diff
@@ -151,7 +151,7 @@ Below is a table showing the mapping of the LTS branches to the packages release
 
 | Nuget Package | Github Branch | LTS Status | LTS Start Date | Maintenance End Date | Removed Date |
 | :-----------: | :-----------: | :--------: | :------------: | :------------------: | :----------: |
-| 1.x.x         | lts_07_2017   | Active     | 2017-07-01     | 2017-12-31           | 2018-06-30   |
+| [2017-10-06](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2017-10-6)    | lts_07_2017   | Active     | 2017-07-01     | 2017-12-31           | 2018-06-30   |
```
yzhong94: I think the intention is to put down the version number?  Microsoft.Azure.Devices.Client v1.5.1, Microsoft.Azure.Devices v1.4.0, Microsoft.Azure.Devices.Shared v1.1.1.

CIPop: I thought that won't fit nicely. Today we have 3 packages but soon we'll grow that to 9 or 10.
Should we design a new table that lists individual packages that are part of the LTS support instead and have the above fields in a list?

yzhong94: I think we should list them unless @olivierbloch thinks otherwise.  We can modify the table to make it look better.

olivierbloch: Are you saying that there could be 9 to 10 packages for 1 LTS? How is that? We update LTS ONLY if there is a security patch pushed into it, which shouldn't happen that often... If we don't push changes, we don't regenerate a new NuGet package. and for each LTS there should be only 1 NuGet package listed (the latest).
Did I miss something?

ryan-z-johnson: @olivierbloch  we have the Device Client nuget, Service Client nuget, and Shared nuget, as we add more IoT functionally (like DPS) we will grow that list.

yzhong94: I think the idea is that we don't have 1 package for C#, we have 3 now, each of them would have an LTS version.

CIPop: > Are you saying that there could be 9 to 10 packages for 1 LTS? How is that? 

@olivierbloch Yes, this LTS will reach end of maintenance in 2017/12/31 as stated above when we'll create a new one that includes the new DPS packages.

CIPop: No other comments on this: I'll change to what @yzhong94 proposed and submit.

PR #225, opened 2017-11-11
Write Configurations entry from the Twin

damonbarry:


damonbarry:
On behalf of @ajithg-msft

shared/Microsoft.Azure.Devices.Shared/TwinJsonConverter.cs
```diff
@@ -119,7 +119,14 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
                 writer.WritePropertyName(X509ThumbprintTag);
                 serializer.Serialize(writer, twin.X509Thumbprint);
             }
-
+ 
+#if ENABLE_MODULES_SDK
```
CIPop: Nit: I think it's OK to remove the conditionals now.

shared/Microsoft.Azure.Devices.Shared/TwinJsonConverter.cs
```diff
@@ -119,7 +119,14 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+            if (twin.Configurations != null)
+            {
+                writer.WritePropertyName(ConfigurationsJsonTag);
```
CIPop: Missing unit test.

shared/Microsoft.Azure.Devices.Shared/TwinJsonConverter.cs
damonbarry: I tried, but if we do that then we have to go into those regions and add the other #ifdefs (WINDOWS_UWP, PCL, NETMF) that we avoid currently because we simply don't define ENABLE_MODULES_SDK in those projects.

CIPop: Doesn't have to be in this check-in but is required before we merge into master.

shared/Microsoft.Azure.Devices.Shared/TwinJsonConverter.cs
CIPop: Offline discussion: tests will be added in a future PR.

shared/Microsoft.Azure.Devices.Shared/TwinJsonConverter.cs
damonbarry: Agreed

PR #228, opened 2017-11-14
update Amqp Sasl Tpm Handshake

jasmineymlo:
- update Amqp Sasl Tpm handshake
- extract sas token builder code from http to common for all transport path

### DIFF
--- a/common/src/device/provisioning/transport/ProvisioingSasBuilder.cs
+++ b/common/src/device/provisioning/transport/ProvisioingSasBuilder.cs
@@ -1,0 +1,71 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Net;
+using System.Text;
+using Microsoft.Azure.Devices.Shared;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
+{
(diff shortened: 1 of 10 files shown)


common/src/device/provisioning/transport/ProvisioingSasBuilder.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Text;
+using Microsoft.Azure.Devices.Shared;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
+{
+    //TODO: merge with other Sas Token Builder in IoTHubClient
+    internal static class ProvisioningSasBuilder
+    {
+        private const string KeyName = "registration";
+        private static readonly TimeSpan TimeToLive = TimeSpan.FromDays(1);
```
CIPop: @tameraw @simonporter should the provisioning token be shorter than 1 day?
/cc @prmathur-microsoft @jebrando @pierreca - what are the other SDKs using for default here?

jasmineymlo: This was extracted from existing code.  I will make note of this and modify accordingly when there is a conclusion on the ttl to use.  Thanks.

CIPop:
Integration failed with `Failed                Microsoft.Azure.Devices.E2ETests.FileUploadE2ETests.FileUploadSuccess_QuotaExceed_Amqp`

PR #233, opened 2017-11-17
Provisioning Service Sample documentation.

CIPop:
Provisioning dev guide and sample how-to.

### DIFF
--- a/provisioning/device/samples/ProvisioningDeviceClientTpm/readme.md
+++ b/provisioning/device/samples/ProvisioningDeviceClientTpm/readme.md
@@ -1,0 +1,18 @@
+# Provisioning Device Client Sample - TPM Attestation
+
+## Overview
+
+This is a quick tutorial with the steps to register a device in the Microsoft Azure IoT Hub Device Provisioning Service using the Trusted Platform Module (TPM) attestation.
+
+## How to run the sample
+
+Ensure that all prerequisite steps presented in [samples](..\) have been performed.
+To run the sample, in a developer command prompt enter:
+
+`dotnet run <IDScope>`
+
+replacing `IDScope` with the value found within the Device Provisioning Service Overview tab. E.g. `dotnet run 0ne1234ABCD`
+
+Continue by following the instructions presented by the sample.
+
+The sample is currently using `SecurityClientTpmSimulator` which is not supported on Linux. To run against a real TPM2.0 device, replace this with `SecurityCleintTpm`.
--- a/provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
+++ b/provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
@@ -1,0 +1,53 @@
+# Provisioning Device Client Sample - X.509 Attestation
+
+## Overview
+
+This is a quick tutorial with the steps to register a device in the Microsoft Azure IoT Hub Device Provisioning Service using X.509 certificates locally generated.
+
+## How to run the sample
+
+Ensure that all prerequisite steps presented in [samples](..\) have been performed.
+
+The sample code is set up to use X.509 certificates stored within a password-protected PKCS12 formatted file (certificate.pfx). To generate a self-signed certificate run the following command:
+
+`powershell .\GenerateTestCertificate.ps1`
+
+The script will prompt for a PFX password. The same password must be used when running the sample.
+
+In your Device Provisioning Service go to "Manage enrollments" and select "Individual Enrollments".
+Select "Add" then fill in the following:
+Mechanism: X.509
+Certificate: Select the public key 'certificate.cer' file.
+DeviceID: iothubx509device1
+
+To run the sample, in a developer command prompt enter:
+`dotnet run <IDScope>`
+
+replacing `IDScope` with the value found within the Device Provisioning Service Overview tab. E.g. `dotnet run 0ne1234ABCD`
+
+Continue by following the instructions presented by the sample.
+
+## Using your certificates
+
+The SDK requires an [X509Certificate2](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2(v=vs.110).aspx) object with private key ([HasPrivateKey](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2.hasprivatekey(v=vs.110).aspx)==true) and, optionally, the certificate chain within an [X509Certificate2Colection](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2collection(v=vs.110).aspx) object.
+
+This can be achieved by changing the following line:
+
+```C# 
+    using (var security = new SecurityClientX509(certificate))
+```
+
+to 
+
+```C# 
+    var myCertificate = new X509Certificate2("myCertificate.pfx", "mypassword");
+    var myChain = new X509Certificate2Collection();
+    myChain.Import("myChain.p7b");
+    using (var security = new SecurityClientX509(myCertificate, myChain)) 
+```
+
+A tool for creating _test_ certificates is available at https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/CACertificateOverview.md
+
+If a Windows compatible Hardware Security Module is used, the certificate must be obtained by opening it from the Certificate Store using [X509Store](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509store(v=vs.110).aspx).
+
+On Linux, .Net Core is using OpenSSL. Using [PInvoke](https://msdn.microsoft.com/en-us/library/55d3thsc.aspx) it is possible to configure the OpenSSL Engine to use a Hardware Security Module and create an X509Certificate2 object from the context.
--- a/provisioning/device/samples/readme.md
+++ b/provisioning/device/samples/readme.md
@@ -1,0 +1,101 @@
+# Provisioning Device Client Sample - Microsoft Azure IoT SDK for .NET
+
+## Overview
+
+_This documentation is preliminary and subject to change._
+
+This folder contains samples demonstrating the steps required to dynamically associate devices with IoT hubs using the Microsoft Azure IoT Device Provisioning Service.
+
+To ensure that only authorized devices can be provisioned, two device attestation mechanisms are supported by the service: one based on X.509 certificates and another based on Trusted Platform Module (TPM) devices. In both cases, public/private key authentication will be performed during provisioning.
+
+_Preview only:_ The SDK currently supports two communication protocols: HTTP and AMQP over TLS.
+
+## Device provisioning in a nutshell
+
+Provisioning is achieved by using a single call to the `ProvisioningDeviceClient.RegisterAsync()` API specifying the IDScope (unique for each Provisioning Service deployment), a `SecurityClient` and a `ProvisioningTransportHandler`:
+
+```C#
+    ProvisioningDeviceClient provClient = ProvisioningDeviceClient.Create(s_idScope, security, transport);
+    DeviceRegistrationResult result = await provClient.RegisterAsync();
+    if (result.Status != ProvisioningRegistrationStatusType.Assigned) 
+    {
+       Console.WriteLine($"ProvisioningClient AssignedHub: {result.AssignedHub}; DeviceID: {result.DeviceId}");
+    }
+```
+
+### Provisioning devices using X.509 certificate-based attestation
+
+Devices must have access to a single certificate with a private key. The private key can be hidden from the application using PKCS #11 Hardware Security Modules.
+
+When an X.509 certificate is used, both the _RegistrationID_ as well as the _DeviceID_ will be equal to the Common Name portion of the certificate Subject. (e.g. If the subject is `CN=mydevice O=Contoso C=US`, the RegistrationID and DeviceID will be `mydevice`.) The name must respect the [DeviceID naming constraints](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-identity-registry).
+
+X.509 attestation comes in two flavors: 
+
+1. Group Enrollment
+In this case, a single (Intermediate) Certificate Authority certificate is uploaded to the Provisioning Service. Devices have access to certificates issued by this CA. All devices will be using the same policies during IoT Hub provisioning.
+
+The service requires proof of possession of the uploaded CA certificate private key. This can be achieved using the [Proof of possession for X.509 Group Enrollment](..\..\service\samples\GroupCertificateVerificationSample) tool.
+
+Because Intermediate Authorities may have been issued by the uploaded CA, the application must present the full chain of certificates from the one used during authentication to the one uploaded to the service. E.g.: If `My CA Certificate` was uploaded to the service and `MyDevice1` is the device certificate, the entire chain must be available on the device: 
+`<"My CA Certificate", "My Intermediate 1", ..., "My Intermediate N", "MyDevice1>`.
+
+2. Individual Enrollment
+In this case, each device certificate gets its own certificate uploaded to the service.
+_Note:_ the device certificate can also be a signing certificate of the actual certificate used to authenticate with the service during provisioning.
+
+An example of specifying the authentication X509Certificate using a PKCS12 PFX password-protected file:
+
+```C#
+    using (var certificate = new X509Certificate2(s_certificateFileName, certificatePassword))
+    using (var security = new SecurityClientX509(certificate))
+    {
+        // ... (see sample for details)
+    }
+```
+
+The SDK provides an extension model [SecurityClientHsmX509](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/shared/Microsoft.Azure.Devices.Shared/SecurityClientHsmX509.cs) that allows hardware vendors to implement custom Hardware Security Modules that store the device certificates. On Windows, PKCS11 HSM devices are supported through the [Certificate Store](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/certificate-stores).
+
+An example of implementation for this extension module is the [SecurityClientX509](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/shared/Microsoft.Azure.Devices.Shared/SecurityClientX509.cs) class.
+
+### Provisioning devices using TPM based attestation
+
+In the case of TPM attestation, a RegistrationID must be supplied by the application.
+
+The TPM attestation supports only Individual Enrollments. The [Endorsement Key](https://technet.microsoft.com/en-us/library/cc770443(v=ws.11).aspx) (EK) must be supplied to the service. During provisioning, the [Storage Root Key](https://technet.microsoft.com/en-us/library/cc753560(v=ws.11).aspx) (SRK) will also be used to ensure that TPM ownership changes result in devices provisioned to the correct owner.
+
+
+```C#
+    using (var security = new SecurityClientTpmSimulator(RegistrationId))
+    {
+        // ... (see sample for details)
+    }
+```
+
+The SDK provides an extension model [SecurityClientHsmTpm](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/shared/Microsoft.Azure.Devices.Shared/SecurityClientHsmTpm.cs) that allows hardware vendors to implement custom TPM v2.0 Hardware Security Modules.
+
+The samples use a TPMv2.0 simulator that uses a loopback TCP connection for communication. This is provided for demonstration purposes only and does not provide any security.
+
+## List of samples
+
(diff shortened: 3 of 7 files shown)


provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -0,0 +1,53 @@
+# Provisioning Device Client Sample - TPM Attestation
```
yzhong94: Do you mean X.509 instead of TPM?

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -0,0 +1,53 @@
...
+Continue by following the instructions presented by the sample.
+
+## Using your certificates
+
+The SDK requires an [X509Certificate2](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2(v=vs.110).aspx) object with private key ([HasPrivateKey](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2.hasprivatekey(v=vs.110).aspx)==true) and, optionally, the certificate chain within an [X509Certificate2Colection](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2collection(v=vs.110).aspx) object.
+
+This can be achieved by changing the following line:
+
+```C# 
+    using (var security = new SecurityClientX509(certificate))
```
yzhong94: Where is this line located?  What file?

yzhong94:
Not sure the if the two sample docs necessarily add more information? It seems like the main readme also covers most of the content.  I'm okay with it the way it is for bug bash

provisioning/device/samples/ProvisioningDeviceClientTpm/readme.md
```diff
@@ -0,0 +1,18 @@
+# Provisioning Device Client Sample - TPM Attestation
+
+## Overview
+
+This is a quick tutorial with the steps to register a device in the Microsoft Azure IoT Hub Device Provisioning Service using the Trusted Platform Module (TPM) attestation.
+
+## How to run the sample
+
+Ensure that all prerequisite steps presented in [samples](..\) have been performed.
+To run the sample, in command prompt enter:
```
nberdy: Just to clarify, this is a normal cmd command prompt right? Not a developer command prompt for VS 2017?

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -0,0 +1,53 @@
+# Provisioning Device Client Sample - TPM Attestation
+
+## Overview
+
+This is a quick tutorial with the steps to register a device in the Microsoft Azure IoT Hub Device Provisioning Service using X509 certificates locally generated.
```
nberdy: Nit: should be X.509 (missing the dot)

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -0,0 +1,53 @@
...
+
+## How to run the sample
+
+Ensure that all prerequisite steps presented in [samples](..\) have been performed.
+
+The sample code is set up to use X509 certificates stored within a password-protected PKCS12 formatted file (certificate.pfx). To generate a self-signed certificate run the following command:
```
nberdy: Nit again, missing the dot in X.509

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -0,0 +1,53 @@
...
+
+`powershell .\GenerateTestCertificate.ps1`
+
+The script will prompt for a PFX password. The same password must be used when running the sample.
+
+In your Device Provisioning Service go to "Manage enrollments" and select "Individual Enrollments".
+Select "Add" then fill in the following:
+Mechanism: X.509
+Certificate: Select the public key 'certificate.cer' file.
+DeviceID: iothubx509device1 (must match the CN part of the certificate Subject)
```
nberdy: This is incorrect. The device ID does NOT have to match the CN of the cert - that's the registration ID that has to match, and we don't ask for it in the portal because we pull it from the cert automagically. The device ID is an optional field.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -0,0 +1,53 @@
...
+
+To run the sample, in command prompt enter:
```
nberdy: Why is the ID scope passed in as a command line parameter?

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -0,0 +1,53 @@
...
+```C# 
+    var myCertificate = new X509Certificate2("myCertificate.pfx", "mypassword");
+    var myChain = new X509Certificate2Collection();
+    myChain.Import("myChain.p7b");
+    using (var security = new SecurityClientX509(myCertificate, myChain)) 
+```
+
+A tool for creating _test_ certificates is available at https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/CACertificateOverview.md
+
+If a Windows compatible Hardware Security Module is used, the certificate must be obtained by opening it from the Certificate Store using [X509Store](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509store(v=vs.110).aspx).
```
nberdy: I just want to say thank you for linking to other MSDN articles.

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
+# Provisioning Device Client Sample - Microsoft Azure IoT SDK for .NET
+
+## Overview
+
+_This documentation is preliminary and subject to change._
+
+This folder contains samples demonstrating the steps required to dynamically associate devices with IoT Hubs using the Microsoft Azure IoT Device Provisioning Service.
```
nberdy: Nit: "IoT Hubs" should be "IoT hubs" (lowercase h). Yes I would like to have Words with the person who made the capitalization rules.

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
...
+## List of samples
+
+- [Provisioning Devices with X509 certificate attestation](ProvisioningDeviceClientX509)
+- [Provisioning Devices with TPM attestation](ProvisioningDeviceClientTpm)
+
+### How to run the samples
+
+_Preview only:_ Running the samples requires building from sources. This will not be necessary once NuGet packages are released.
+
+1. Prepare your development environment. Follow the instructions at https://github.com/Azure/azure-iot-sdk-csharp/blob/master/device/doc/devbox_setup.md
```
yzhong94: What does a bug basher need in order to get started quickly?  Can we provide a list of recommended steps (build locally v. get package, if get package, then how do we proceed etc.)?

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
...
+
+2. Build the environment. In the root of your clone, type:
+
+```build -clean -wip_provisioning```
```
yzhong94: azureiot-sdk-csharp or the root of the sample folder?

yzhong94: When I run this command from azure-iot-sdk-csharp, I get 

.\build.ps1 : File C:\Users\yizhon\azure-iot-sdk-csharp\build.ps1 cannot be loaded because running scripts is disabled
on this system. For more information, see about_Execution_Policies at http://go.microsoft.com/fwlink/?LinkID=135170.
At line:1 char:1
+ .\build.ps1 -clean -wip_provisioning ; exit $LASTEXITCODE
+ ~~~~~~~~~~~
    + CategoryInfo          : SecurityError: (:) [], PSSecurityException
    + FullyQualifiedErrorId : UnauthorizedAccess

When I run it from azure-iot-sdk-csharp\provisioning\device\samples\ProvisioningDeviceClientTpm>build -clean -wip_provisioning
'build' is not recognized as an internal or external command,
operable program or batch file.

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
...
+    DeviceRegistrationResult result = await provClient.RegisterAsync();
+    if (result.Status != ProvisioningRegistrationStatusType.Assigned) 
+    {
+    Console.WriteLine($"ProvisioningClient AssignedHub: {result.AssignedHub}; DeviceID: {result.DeviceId}");
+    }
+```
+
+### Provisioning devices using X509 certificate-based attestation
+
+Devices must have access to a single certificate with a private key. The private key can be hidden from the application using PCKS11 Hardware Security Modules.
```
nberdy: Might want to verify the PKCS11 support with Simon.

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
...
+
+When an X509 certificate is used, both the _RegistrationID_ as well as the _DeviceID_ will be equal to the Common Name portion of the certificate Subject. (e.g. If the subject is `CN=mydevice O=Contoso C=US`, the RegistrationID and DeviceID will be `mydevice`.) The name must respect the [DeviceID naming constraints](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-identity-registry).
```
nberdy: The device ID does NOT have to be the same as the CN, only the registration ID has to match the CN.

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
...
+
+X509 attestation comes in two flavors: 
+
+1. Group Enrollment
+In this case, a single Certificate Authority certificate is uploaded to the Provisioning Service. Devices have access to certificates issued by this CA. All devices will be using the same policies during IoT Hub provisioning.
```
nberdy: Might want to specify that intermediate certs can be used to - I'd like to avoid confusion with the later paragraph about why the entire cert chain is required to be presented.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
CIPop: There is only one source file in this folder so I assumed that's clear. I will add the file name and a comment in the source.

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
...
+
+```C#
+    using (var certificate = new X509Certificate2(s_certificateFileName, certificatePassword))
+    using (var security = new SecurityClientX509(certificate))
+    {
+        //...
+    }
+```
+
+The SDK provides an extension model [SecurityClientHsmX509](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/shared/Microsoft.Azure.Devices.Shared/SecurityClientHsmX509.cs) that allows hardware vendors to implement custom Hardware Security Modules that store the device certificates. On Windows, PKCS11 HSM devices are supported through the [Certificate Store](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/certificate-stores).
```
nberdy: Yay thank you for this call-out!

CIPop:
> Not sure the if the two sample docs necessarily add more information? It seems like the main readme also covers most of the content. I'm okay with it the way it is for bug bash

There is substantial more information as prerequisites are different for X509 for example. The main readme doesn't cover the specific TPM vs X509 sample implementation.

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
...
+
+In the case of TPM attestation, a RegistrationID must be supplied by the application.
+
+The TPM attestation supports only Individual Enrollments. The [Endorsement Key](https://technet.microsoft.com/en-us/library/cc770443(v=ws.11).aspx) (EK) must be supplied to the service. During provisioning, the [Storage Root Key](https://technet.microsoft.com/en-us/library/cc753560(v=ws.11).aspx) (SRK) will also be used to ensure that TPM ownership changes result in devices provisioned to the correct owner.
+
+
+```C#
+    using (var security = new SecurityClientTpmSimulator(RegistrationId))
+    {
+        //...
```
nberdy: Is this complete?

provisioning/device/samples/ProvisioningDeviceClientTpm/readme.md
CIPop: I wanted to avoid that: on Linux there is no dev command prompt...
The `dotnet` tool must be available.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
CIPop: Will fix here and everywhere I can find X509 in docs.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
CIPop: @nberdy I never tested that scenario - I assume that's only true for Individual Enrollments (where you can specify it).
In Enrollment Groups I've tested that above is always true and there is no way to change the device ID.

/cc @simonporter

CIPop: Will change the comment here since this is only about Individual Enrollments but, if you can confirm, I'll also add the group explanation.

provisioning/device/samples/readme.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Provisioning devices using X509 certificate-based attestation
+
+Devices must have access to a single certificate with a private key. The private key can be hidden from the application using PCKS11 Hardware Security Modules.
+
+When an X509 certificate is used, both the _RegistrationID_ as well as the _DeviceID_ will be equal to the Common Name portion of the certificate Subject. (e.g. If the subject is `CN=mydevice O=Contoso C=US`, the RegistrationID and DeviceID will be `mydevice`.) The name must respect the [DeviceID naming constraints](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-identity-registry).
+
+X509 attestation comes in two flavors: 
+
+1. Group Enrollment
```
nberdy: Please add details about how a device can use the derived certificate to provision via DPS as part of an enrollment group. This may need to be a sample using the entire cert chain as mentioned below somewhere.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
CIPop: Configuration should never be added to source control - promotes bad practices and enables a high risk scenario where PII or secrets get checked in by mistake.

Because in the last month we had 3 incidents where people submitted PRs containing secrets, I plan to change all samples in the future to take these via command-line parameters.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
CIPop: 👍 Please let me know if we should link to other docs. (E.g. if you have a permanent link to your blog-posts.)

provisioning/device/samples/readme.md
CIPop: /cc @simonporter 

Windows handles PKCS11 the same way as a normal certificate except that the private key never leaves the device.

provisioning/device/samples/readme.md
CIPop: See above discussion.

provisioning/device/samples/readme.md
CIPop: I could say In this case, a single _(Intermediate)_ Certificate Authority certificate is uploaded...
I'm open to suggestions.

provisioning/device/samples/readme.md
CIPop: I only wanted to point out the single line describing which security type to use. The `//...` is the part that's not TPM specific.

provisioning/device/samples/readme.md
CIPop: At the moment we don't have packaging and so, unfortunately, this is the shortest path - as mentioned in line 85.

At the very minimum, the system should have the following on Windows or Linux:
1. http://dot.net
2. Powershell (https://github.com/PowerShell/PowerShell)

.Net Framework / Xamarin will not be testable with only that.

provisioning/device/samples/readme.md
CIPop: The root of the clone. 

Your system is blocking Powershell script execution. This is a common case. Developers usually lower the security requirement by running the following in an Admin Powershell script:

`Set-ExecutionPolicy RemoteSigned`

Since this comes with a security risk and the error correctly informs the user and gives out the link on what that means, I prefer not documenting how to disable the security feature.

provisioning/device/samples/readme.md
yzhong94: Do we need to build the SDK in this step?

provisioning/device/samples/readme.md
yzhong94: Maybe we can say, Powershell script execution must be enabled on this device

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
nberdy: I confirm - I have an individual enrollment with a device ID set to something that is not the CN.

In enrollment groups, you cannot specify the device ID because it's a group of devices and we don't yet have the ability to specify a naming rule for the group.

provisioning/device/samples/readme.md
CIPop: From the client's standpoint there is no difference: we only need the client certificate and its chain.

The X509 sample has this automated via the cert generation script for Individual Enrollment.
Developers already having certificates and their chains should follow instructions here:
https://github.com/Azure/azure-iot-sdk-csharp/pull/233/files#diff-9a7ac6d32ecad3deadba51e4b1fba932R43

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
nberdy: What about just a placeholder in the source code and instruction to the user to fill it in with their own ID scope? This is a best practice for folks who will straight up copy+paste the sample code to bootstrap their own development efforts. I don't want prod devices taking ID scope as a startup parameter because it goes against some of the design principles of the DPS flow.

provisioning/device/samples/readme.md
nberdy: Nit: this line has the K and the C backwards.

provisioning/device/samples/readme.md
nberdy: Seriously, the device ID does not have to be the same as the CN, I have a DPS enrollment entry as proof :)

provisioning/device/samples/readme.md
nberdy: @dsk-2015 ideas?

provisioning/device/samples/readme.md
nberdy: Ah. Maybe add a "your code here" to the comment?

provisioning/device/samples/readme.md
CIPop: Yes, everything needs to be built at the moment. We won't need to do this after we have packages.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
CIPop: > I don't want prod devices taking ID scope as a startup parameter because it goes against some of the design principles of the DPS flow.

What are the design principles that this is going against?

I see the IDScope similar to a SQL connection string. These should never be hard-coded within the source-code and are instead stored in a configuration file. 

How would one check-in the code to GitHub if it has an IDScope in it?
Also, why would a customer purchasing a software solution be required to have access to the source code and recompile it in order to use it on their devices?

provisioning/device/samples/readme.md
CIPop: Sounds good. I can also add a link to the docs on how to do that.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
nberdy: The DPS flow was designed such that the ID scope will never have to be changed once a device is first enrolled in DPS - the device may change owners or may change IoT hubs, but the ID scope does not change. The ID scope will be the same for all devices in a batch; in fact, the only differences between devices connecting to DPS are the registration ID and the TPM/cert information that is unique per device. The ultimate goal for devices registering with DPS is that they can use the exact same firmware image across a batch of devices and differ only by info stored in the on-board HSM.

CIPop: This is only a sample that can't be "burned-in" as firmware to a device. .Net (except NetMF which isn't supported for DPS) cannot run on bare-metal devices. The same exe/dll should be reusable on all devices without recompilation.
Even more, this is a sample which is meant to be run by everybody that wants a quick and safe way to try out our SDK. Requiring sample editing is neither given that the code is under git source control.

Issue #85, opened 2017-04-12
Massive memory leak in RetryDelegatingHandler


JochenKalmbach:
We have a massive memory leak in the RetryDelegatingHandler if there is no connection to the IoTHub (we use Http1).
The leaking objects are: `System.Threading.Tasks.UnwrapPromise~1[[Microsoft.Azure.Devices.Client.Message, Microsoft.Azure.Devices.Client]]`

If you take a deeper look into it, it shows the GC root as:
```
0:000> !GCRoot -all 25076704
Thread 1b14:
    09a1df64 69fcf89e System.Threading.Tasks.ContinuationResultTaskFromResultTask-2[[System.__Canon, mscorlib],[System.__Canon, mscorlib]].InnerInvoke()
        esi: 
            ->  250d28ac System.Threading.Tasks.ContinuationResultTaskFromResultTask-2[[Microsoft.Azure.Devices.Client.Message, Microsoft.Azure.Devices.Client],[System.Threading.Tasks.Task-1[[Microsoft.Azure.Devices.Client.Message, Microsoft.Azure.Devices.Client]], mscorlib]]
            ->  250d2974 System.Threading.Tasks.UnwrapPromise-1[[Microsoft.Azure.Devices.Client.Message, Microsoft.Azure.Devices.Client]]
            ->  250ccf78 System.Threading.Tasks.UnwrapPromise~1[[Microsoft.Azure.Devices.Client.Message, Microsoft.Azure.Devices.Client]]
            ->  250c93e0 System.Threading.Tasks.UnwrapPromise-1[[Microsoft.Azure.Devices.Client.Message, Microsoft.Azure.Devices.Client]]
    <...cut...>
```

If you then take a look at the root object:
```
0:000> !do 250d28ac
Name:        System.Threading.Tasks.ContinuationResultTaskFromResultTask`2[[Microsoft.Azure.Devices.Client.Message, Microsoft.Azure.Devices.Client],[System.Threading.Tasks.Task`1[[Microsoft.Azure.Devices.Client.Message, Microsoft.Azure.Devices.Client]], mscorlib]]
```

You will see this Action:
```
0:000> !do 250d288c 
[...]

PR #236, opened 2017-11-17
Fix Memory Leak

ryan-z-johnson:
Fixing issues in #85 

Needed to update the ASyncExectuion class to run a method async to allow for use of the await keyword on a task execution.

### DIFF
--- a/device/Microsoft.Azure.Devices.Client/TransientFaultHandling/AsyncExecution.cs
+++ b/device/Microsoft.Azure.Devices.Client/TransientFaultHandling/AsyncExecution.cs
@@ -30,11 +30,11 @@ namespace Microsoft.Azure.Devices.Client.TransientFaultHandling
     /// Provides a wrapper for a non-generic <see cref="T:System.Threading.Tasks.Task" /> and calls into the pipeline
     /// to retry only the generic version of the <see cref="T:System.Threading.Tasks.Task" />.
     /// </summary>
-    internal class AsyncExecution : AsyncExecution<bool>
+    internal class AsyncExecution : AsyncExecution<Task<bool>>
     {
         private static Task<bool> cachedBoolTask;
 
-        public AsyncExecution(Func<Task> taskAction, ShouldRetry shouldRetry, Func<Exception, bool> isTransient, Action<int, Exception, TimeSpan> onRetrying, bool fastFirstRetry, CancellationToken cancellationToken) : base(() => AsyncExecution.StartAsGenericTask(taskAction), shouldRetry, isTransient, onRetrying, fastFirstRetry, cancellationToken)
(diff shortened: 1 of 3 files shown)


device/Microsoft.Azure.Devices.Client/TransientFaultHandling/AsyncExecution.cs
```diff
@@ -30,11 +30,11 @@ namespace Microsoft.Azure.Devices.Client.TransientFaultHandling
     /// Provides a wrapper for a non-generic <see cref="T:System.Threading.Tasks.Task" /> and calls into the pipeline
     /// to retry only the generic version of the <see cref="T:System.Threading.Tasks.Task" />.
     /// </summary>
-    internal class AsyncExecution : AsyncExecution<bool>
+    internal class AsyncExecution : AsyncExecution<Task<bool>>
```
CIPop: I don't understand why the change is within the Microsoft Patterns & Practices code. 
If we have a leak caused by this, we are most likely using it incorrectly, not that the definition is incorrect.

device/Microsoft.Azure.Devices.Client/TransientFaultHandling/AsyncExecution.cs
```diff
@@ -43,8 +43,16 @@ public AsyncExecution(Func<Task> taskAction, ShouldRetry shouldRetry, Func<Excep
...
         /// <returns>A <see cref="T:System.Threading.Tasks.Task" /> that wraps the non-generic <see cref="T:System.Threading.Tasks.Task" />.</returns>
-        private static Task<bool> StartAsGenericTask(Func<Task> taskAction)
+        private async static Task<Task<bool>> StartAsGenericTask(Func<Task> taskAction, CancellationToken cancellationToken)
         {
+            if (cancellationToken.IsCancellationRequested)
+            {
+
+                TaskCompletionSource<bool> taskCompletionSource = new TaskCompletionSource<bool>();
+                taskCompletionSource.TrySetCanceled();
+                return taskCompletionSource.Task;
```
CIPop: Looks like [Task.FromCancelled](https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.fromcanceled(v=vs.110).aspx)?

device/Microsoft.Azure.Devices.Client/TransientFaultHandling/AsyncExecution.cs
```diff
@@ -65,7 +73,7 @@ private static Task<bool> StartAsGenericTask(Func<Task> taskAction)
                 }), "taskAction");
             }
             TaskCompletionSource<bool> tcs = new TaskCompletionSource<bool>();
-            task.ContinueWith(delegate (Task t)
+            await task.ContinueWith(delegate (Task t)
```
CIPop: This is very odd - the code is configuring a task continuation then you are awaiting it?
It's basically the same thing as writing 
```C# 
await task; 
await continuation;
```

I don't think this was the intent here. Continuations are the right thing to use: the problem is when we retry, we should _not_ append a continuation, rather we should complete the original task and start another one.

CIPop:
Tentatively closing this as @ryan-z-johnson  is on vacation.

PR #237, opened 2017-11-18
Adding .NetStandard Multiplex with AMQP Sample

sergaz-msft:
Adding .NetStandard Sample with Multiplex over Amqp_WebSocket_Only or Amqp_Tcp_Only

- Send and Receive (similar with DeviceClientAmqpSample)

### DIFF
--- a/device/samples/NetStandardMultiplexSample/Program.cs
+++ b/device/samples/NetStandardMultiplexSample/Program.cs
@@ -1,0 +1,138 @@
+﻿using System;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Client;
+using System.Threading;
+
+namespace multiplexingDotNetCore
+{
+    class Program
+    {
+        // String containing IotHub Hostname:
+        //"hostname.azure-devices.net"
+        private const string IoTHubHostName = "<iothubhostname>.azure-devices.net";
+
+        //Number of devices for multiplexing
+        private static int MUX_DEVICES = 4;
+        private static DeviceClient[] deviceClients = new DeviceClient[MUX_DEVICES];
+
+        //Array of device Ids (should have the same entries as defined on MUX_DEVICES):
+        //{"device1", "device2", "device3", "device4"}
+        private static String[] deviceIds = new String[] { "<device1>", "<device2>", "<device3>", "<device4>" };
+
+        //Array of device PrimaryKeys (following the same order as above):
+        //{"PrimaryKey device1", "PrimaryKey device2", "PrimaryKey device3", "PrimaryKey device4"}
+        private static String[] devicePrimaryKeys = new String[] { "<PrimaryKey device1>",
+                                                                    "<PrimaryKey device2>",
+                                                                    "<PrimaryKey device3>",
+                                                                    "<PrimaryKey device4>"    };
+
+        private static int MESSAGE_COUNT = 5;
+        private const int TEMPERATURE_THRESHOLD = 30;
+
+        private static float temperature;
+        private static float humidity;
+        private static Random rnd = new Random();
+
+        static void Main(string[] args)
+        {
+
+            try
+            {
+                MultiplexConnection().Wait();
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine("Error in sample: {0}", ex.Message);
+            }
+
+            Console.WriteLine("Exited!\n");
+        }
+
+        static async Task MultiplexConnection()
+        {
+            for (int i = 0; i < 4; i++)
+            {
+                var auth = new DeviceAuthenticationWithRegistrySymmetricKey(deviceIds[i], devicePrimaryKeys[i]);
+                deviceClients[i] = DeviceClient.Create(
+                    IoTHubHostName,
+                    auth,
+                    new ITransportSettings[]
+                    {
+                        //Must specify Amqp_WebSocket_Only or Amqp_Tcp_Only
+                        new AmqpTransportSettings(TransportType.Amqp_WebSocket_Only)
+                        {
+                            //By default, connection pooling is turned off. It must be turned on explicitly using the below settings.
+                            AmqpConnectionPoolSettings = new AmqpConnectionPoolSettings()
+                            {
+                                Pooling = true,
+                                //If you would like to have only one TCP Connection set MaxPoolSize=1
+                                MaxPoolSize = 2
+                            }
+                        }
+                    });
+                await deviceClients[i].OpenAsync();
+            }
+
+            SendEvent().Wait();
+            ReceiveCommands().Wait();
+        }
+
+        static async Task SendEvent()
+        {
+            string dataBuffer;
+
+            Console.WriteLine("Devices sending {0} messages to IoTHub...\n", MESSAGE_COUNT);
+
+            for (int count = 0; count < MESSAGE_COUNT; count++)
+            {
+                for (int i = 0; i < MUX_DEVICES; i++)
+                {
+                    temperature = rnd.Next(20, 35);
+                    humidity = rnd.Next(60, 80);
+                    dataBuffer = string.Format("{{\"deviceId\":\"{0}\",\"messageId\":{1},\"temperature\":{2},\"humidity\":{3}}}", deviceIds[i], count, temperature, humidity);
+                    Message eventMessage = new Message(Encoding.UTF8.GetBytes(dataBuffer));
+                    eventMessage.Properties.Add("temperatureAlert", (temperature > TEMPERATURE_THRESHOLD) ? "true" : "false");
+                    Console.WriteLine("\t{0}> Sending message: {1}, Data: [{2}]", DateTime.Now.ToLocalTime(), count, dataBuffer);
+                    try
+                    {
+                        await deviceClients[i].SendEventAsync(eventMessage);
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.WriteLine("Error in sending message: {0}", ex.Message);
+                    }
+                    Thread.Sleep(1000);
+                }
+            }
+        }
+
+        static async Task ReceiveCommands()
+        {
+            Console.WriteLine("\nDevices waiting for commands from IoTHub...\n");
+            Message receivedMessage;
+            string messageData;
+
+            while (true)
+            {
+                for (int i = 0; i < MUX_DEVICES; i++)
+                {
+                    receivedMessage = await deviceClients[i].ReceiveAsync(TimeSpan.FromSeconds(1));
+                    if (receivedMessage != null)
+                    {
+                        messageData = Encoding.ASCII.GetString(receivedMessage.GetBytes());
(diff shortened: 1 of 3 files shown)


device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+{
+    class Program
+    {
+        // String containing IotHub Hostname:
+        //"hostname.azure-devices.net"
+        private const string IoTHubHostName = "<iothubhostname>.azure-devices.net";
+
+        //Number of devices for multiplexing
+        private static int MUX_DEVICES = 4;
+        private static DeviceClient[] deviceClients = new DeviceClient[MUX_DEVICES];
```
CIPop: Why are all of these static? To promote better coding practices we should have our samples properly use classes representing a single temperature/humidity sensor and simply instantiate multiple of those in the `Main` function.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        //Array of device Ids (should have the same entries as defined on MUX_DEVICES):
+        //{"device1", "device2", "device3", "device4"}
```
CIPop: Nit: I don't think this particular line has any value - it's almost the same as line 21 below.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+    {
+        // String containing IotHub Hostname:
+        //"hostname.azure-devices.net"
+        private const string IoTHubHostName = "<iothubhostname>.azure-devices.net";
+
+        //Number of devices for multiplexing
+        private static int MUX_DEVICES = 4;
+        private static DeviceClient[] deviceClients = new DeviceClient[MUX_DEVICES];
+
+        //Array of device Ids (should have the same entries as defined on MUX_DEVICES):
```
CIPop: Nit: Space after // and the first word of the comment.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+        //{"device1", "device2", "device3", "device4"}
+        private static String[] deviceIds = new String[] { "<device1>", "<device2>", "<device3>", "<device4>" };
+
+        //Array of device PrimaryKeys (following the same order as above):
+        //{"PrimaryKey device1", "PrimaryKey device2", "PrimaryKey device3", "PrimaryKey device4"}
```
CIPop: Same as above.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+        //"hostname.azure-devices.net"
+        private const string IoTHubHostName = "<iothubhostname>.azure-devices.net";
+
+        //Number of devices for multiplexing
+        private static int MUX_DEVICES = 4;
+        private static DeviceClient[] deviceClients = new DeviceClient[MUX_DEVICES];
+
+        //Array of device Ids (should have the same entries as defined on MUX_DEVICES):
+        //{"device1", "device2", "device3", "device4"}
+        private static String[] deviceIds = new String[] { "<device1>", "<device2>", "<device3>", "<device4>" };
```
CIPop: `string` instead of `String`, also see my OOP comment above. Otherwise this is not properly named (should be `s_deviceIds`).

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+        private static int MESSAGE_COUNT = 5;
+        private const int TEMPERATURE_THRESHOLD = 30;
+
+        private static float temperature;
+        private static float humidity;
+        private static Random rnd = new Random();
+
+        static void Main(string[] args)
+        {
+
```
CIPop: Nit: extra space.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+            try
+            {
+                MultiplexConnection().Wait();
```
CIPop: For useful exceptions, instead of `.Wait()` use `.GetAwaiter().GetResult();` in Main.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+            catch (Exception ex)
```
CIPop: This is bad practice. Applications should not capture exceptions. Allowing them to crash enables automatic exception handlers within the OS to aggregate them and present useful information (e.g. if it's a Store app, the developer can inspect the crashes across multiple apps).

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+        private static int MUX_DEVICES = 4;
+        private static DeviceClient[] deviceClients = new DeviceClient[MUX_DEVICES];
+
+        //Array of device Ids (should have the same entries as defined on MUX_DEVICES):
+        //{"device1", "device2", "device3", "device4"}
+        private static String[] deviceIds = new String[] { "<device1>", "<device2>", "<device3>", "<device4>" };
+
+        //Array of device PrimaryKeys (following the same order as above):
+        //{"PrimaryKey device1", "PrimaryKey device2", "PrimaryKey device3", "PrimaryKey device4"}
+        private static String[] devicePrimaryKeys = new String[] { "<PrimaryKey device1>",
```
CIPop: This is really bad practice - at absolutely no point should the user hard-code any Azure keys within C# code.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                Console.WriteLine("Error in sample: {0}", ex.Message);
+            }
+
+            Console.WriteLine("Exited!\n");
+        }
+
+        static async Task MultiplexConnection()
+        {
+            for (int i = 0; i < 4; i++)
```
CIPop: Since you have defined `MUX_DEVICES`, I would expect the code uses it here.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+                            //By default, connection pooling is turned off. It must be turned on explicitly using the below settings.
+                            AmqpConnectionPoolSettings = new AmqpConnectionPoolSettings()
+                            {
+                                Pooling = true,
+                                //If you would like to have only one TCP Connection set MaxPoolSize=1
+                                MaxPoolSize = 2
+                            }
+                        }
+                    });
+                await deviceClients[i].OpenAsync();
```
CIPop: This serializes the OpenAsync() operations. Why not create a Task array and allow all Open operations to happen asynchronously using Task.WhenAny?

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            SendEvent().Wait();
```
CIPop: Definitely never combine synchronous (Wait/GetAwaiter) with async code.
`await SendEvent()` instead.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+            ReceiveCommands().Wait();
```
CIPop: Why is Send and Receive factored in other methods while open is not.

I would rather have something on the lines of 
```C#

await OpenAllAsync();
await SendAllAsync();
await ReceiveAllAsync();
```

if this is the semantic you are looking for (i.e. all devices must complete the previous operations before they start the next).

I would not recommend doing this in production though: each operation should execute asynchronously. The only ordering required is that a device Opens, Sends then Receives, regardless of other devices do at that time. (Ties in with the OOP comment I had above.)

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+                    humidity = rnd.Next(60, 80);
+                    dataBuffer = string.Format("{{\"deviceId\":\"{0}\",\"messageId\":{1},\"temperature\":{2},\"humidity\":{3}}}", deviceIds[i], count, temperature, humidity);
+                    Message eventMessage = new Message(Encoding.UTF8.GetBytes(dataBuffer));
+                    eventMessage.Properties.Add("temperatureAlert", (temperature > TEMPERATURE_THRESHOLD) ? "true" : "false");
+                    Console.WriteLine("\t{0}> Sending message: {1}, Data: [{2}]", DateTime.Now.ToLocalTime(), count, dataBuffer);
+                    try
+                    {
+                        await deviceClients[i].SendEventAsync(eventMessage);
+                    }
+                    catch (Exception ex)
```
CIPop: We should not catch `Exception`. Only capture DeviceClient specific exceptions that we are OK to ignore.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+                    {
+                        Console.WriteLine("Error in sending message: {0}", ex.Message);
+                    }
+                    Thread.Sleep(1000);
```
CIPop: Same as above: never mix sync and async operations. 
`await Task.Delay(1000);` instead.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+                            Console.WriteLine("\t\tProperty[{0}> Key={1} : Value={2}", propCount++, prop.Key, prop.Value);
+                        }
+
+                        await deviceClients[i].CompleteAsync(receivedMessage);
+                    }
+                }
+            }
+        }
+    }
+}
```
CIPop: Nit: missing NL at EOF.

device/samples/NetStandardMultiplexSample/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+                        Console.WriteLine("Error in sending message: {0}", ex.Message);
+                    }
+                    Thread.Sleep(1000);
+                }
+            }
+        }
+
+        static async Task ReceiveCommands()
+        {
+            Console.WriteLine("\nDevices waiting for commands from IoTHub...\n");
```
CIPop: I'd add something "Press Ctrl+C" to stop and describe a way for someone to use either DeviceExplorer or the Azure Portal to send some commands to the devices. Otherwise someone running this sample could think that the app is hanging.

CIPop:
@sergaz-msft any updates on this PR? If you expect it'll take longer to update, please close then reopen later.

sergaz-msft:
Hi @CIPop . Planning to tackle it this week. Thanks.

CIPop:
Closing per offline discussion with @sergaz-msft. Please reopen when convenient.

PR #241, opened 2017-11-21
Add filtering capability to the Device Explorer

zposten:
# Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-c/blob/master/.github/CONTRIBUTING.md).
- [ ] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down). **<-- Are there tests for the Device Explorer tool?**

# Description of the problem
When an IoT hub has more than 1000 devices, the device explorer can no longer be used to find every device.  Searching for a particular device in such a large IoT hub (a not uncommon use case) is also difficult.

# Description of the solution
  - Up the max number of devices that the Device Explorer will show, to a value that is unlikely to be exceeded (admittedly, this is akin to delaying the problem, but it keeps closer with the current implementation).   I'm not sure why this arbitrary limit was put in here in the first place, but I'm assuming it was for performance.  However, a user with more than 1000 devices in his IoT hub has a choice between the tool maybe being slow, and not being able to use the tool at all to configure some of their devices.  The choice seems like a simple one for me.

  - Add a filter text box to the management tab to allow users to search common fields

### DIFF
--- a/tools/DeviceExplorer/DeviceExplorer/MainForm.cs
+++ b/tools/DeviceExplorer/DeviceExplorer/MainForm.cs
@@ -10,6 +10,7 @@ using Microsoft.Azure.Devices.Common;
 using Microsoft.Azure.Devices.Common.Security;
 using Microsoft.ServiceBus.Messaging;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
@@ -52,6 +53,8 @@ namespace DeviceExplorer
         private static DataGridViewRow messageSysPropCorrelationId = new DataGridViewRow();
         private static DataGridViewRow messageSysPropContentType = new DataGridViewRow();
         private static DataGridViewRow messageSysPropContentEncoding = new DataGridViewRow();
+
+        private SortableBindingList<DeviceEntity> allDevices;
 
         private string deviceIDSearchPattern = String.Empty;
         private DateTime deviceIDSearchPatternLastUpdateTime = DateTime.Now;
@@ -341,7 +344,8 @@ namespace DeviceExplorer
             var devicesProcessor = new DevicesProcessor(activeIoTHubConnectionString, MAX_COUNT_OF_DEVICES, protocolGatewayHost.Text);
             var devicesList = await devicesProcessor.GetDevices();
             devicesList.Sort();
-            var sortableDevicesBindingList = new SortableBindingList<DeviceEntity>(devicesList);
+            allDevices = new SortableBindingList<DeviceEntity>(devicesList);
+
             string deviceCurrentlySelected = null;
 
             // Save the device ID currently selected on the grid.
@@ -350,18 +354,12 @@ namespace DeviceExplorer
                 deviceCurrentlySelected = (string)devicesGridView.SelectedRows[0].Cells[0].Value;
             }
 
-            devicesGridView.DataSource = sortableDevicesBindingList;
             devicesGridView.ReadOnly = true;
             devicesGridView.SelectionMode = DataGridViewSelectionMode.FullRowSelect;
 
-            if (devicesList.Count() > MAX_COUNT_OF_DEVICES)
-            {
-                deviceCountLabel.Text = MAX_COUNT_OF_DEVICES + "+";
-            }
-            else
-            {
-                deviceCountLabel.Text = devicesList.Count().ToString();
-            }
+            deviceCountLabel.Text = devicesList.Count > MAX_COUNT_OF_DEVICES
+                ? MAX_COUNT_OF_DEVICES + "+"
+                : devicesList.Count.ToString();
 
             // Re-select the device ID previously selected before the update.
             // This avoids the super-annoying need to scroll down every time the management grid gets updated.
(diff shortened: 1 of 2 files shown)


zposten:
@CIPop Thank you for looking at this pull request.  I cannot duplicate this error, even when I build for release.  I could add another null check, but I don't understand how null could possibly get in there with the current implementation.

The only (handled) exception that I would expect is if you entered something like `[`, which alone is an invalid regex.

I don't know that I'm using the JIT though, I'm wondering if that has something to do with it?  Is there a way I can enable that?

CIPop:
@zposten Try replicating my environment: I have only 16 devices in my hub and ran the exe without attaching Visual Studio. This could be a race condition which means you may not always hit it.

If you add any instrumentation to your branch that would help you debug (using Debug.WriteLine for example) I could try with that.

Let me know when you have added a fix or the instrumentation in your PR branch.

zposten:
@CIPop I created a new hub with only 6 devices, ran the exe directly from the file system in both Debug and Release modes, and still was not able to replicate the issue.  I also built an installer and had someone else run it on their system; they also did not experience any issues.

As requested, I added some log statements around the filtering.  Everything is printing on my side exactly as I would expect it to.  I'm hoping these log statements shed some light on what's happening on your side?

Your stack trace above still confuses me.  `StringMatchesFilterText` is only called in one place, `DeviceMatchesFilterText`.  That also is only called in one place, `filterDevices`, and at the very beginning of that method, I check for `null` filter text inside of the `IsValidRegex` method.  

When you're typing, the `filterText` method should only be called from one place; `filterDevicesTextBox_TextChanged`.  This makes it hard for me to see where race conditions would come into play.  Especially when I have messed with the filter a great deal and not seen anything.

If the log statements prove unhelpful, would it be possible for one of the other reviewers to give it a quick run to see if they encounter the same issue?  I apologize that this is being troublesome, and I appreciate your time.  My team uses the device explorer frequently and gets a lot of value out of it.

CIPop:
Thanks @zposten ! I'll give this a try when I get a chance (more likely next week).

tools/DeviceExplorer/DeviceExplorer/MainForm.cs
```diff
@@ -1102,5 +1103,71 @@ private void devicesGridViewContextMenu_Opened(object sender, EventArgs e)
...
             }
         }
+
+        #region Device Filtering
+        private void filterDevicesTextBox_TextChanged(object sender, EventArgs e)
+        {
+            filterDevices();
+        }
+
+        private void filterDevices() { filterDevices(filterDevicesTextBox.Text); }
```
CIPop: Nit: please start function names with caps. `FilterDevices()`. (See coding standards.)

I would not change the above function though since that is prefixed by the name of the textbox.

tools/DeviceExplorer/DeviceExplorer/MainForm.cs
```diff
@@ -1102,5 +1103,71 @@ private void devicesGridViewContextMenu_Opened(object sender, EventArgs e)
...
+                device.SecondaryKey,
+                device.ConnectionState,
+                device.State,
+
+            }.Any(s => StringMatchesFilterText(s, filterText));
+        }
+
+        private static bool StringMatchesFilterText(string s, string filterText)
+        {
+            Debug.WriteLine($"\t\tChecking if string {s} matches filter {filterText}");
```
CIPop: Some of the properties can and will be `null`. E.g. Keys can be null if the device was created by the new Azure IoT Provisioning Service.

Adding the following fixes the issue:
`if (string.IsNullOrWhiteSpace(s)) return false;`

tools/DeviceExplorer/DeviceExplorer/MainForm.cs
```diff
@@ -25,7 +27,7 @@ public partial class MainForm : Form
     {
         #region fields
         private const int MAX_TTL_VALUE = 365;
-        private const int MAX_COUNT_OF_DEVICES = 1000;
+        private const int MAX_COUNT_OF_DEVICES = 100000;
```
tameraw: >100000 [](start = 49, length = 6)

This will not take effect as the GetDevicesAsync() call caps at 1000 devices max. 
Please see note in this article:
[https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-bulk-identity-mgmt](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-bulk-identity-mgmt)

zposten: I have more changes coming that will allow for searching for devices that are not in the first thousand, but I thank you for that article, and I have put it back for the time being 😄

tools/DeviceExplorer/DeviceExplorer/MainForm.cs
zposten: Thank you for finding this!  This has been implemented!

zposten:
@CIPop @tameraw This PR has been updated with your suggestions, I appreciate your help with it!

tools/DeviceExplorer/DeviceExplorer/MainForm.cs
zposten: Done!

PR #248, opened 2017-11-25
Adding MQTT provisioning implementation.

CIPop:
Adding implementation for MQTT over TCP.
Testing MQTT over WS is blocked on Windows by .Net Core.

`ClientWebSocketChannel.cs` and `ClientWebSocketChannelConfiguration.cs` are copies from IoT Hub DeviceClient. I'm not sure why but in the original code they are public API which I don't want for Provisioning.

### DIFF
--- a/common/src/device/provisioning/transport/ClientApiVersionHelper.cs
+++ b/common/src/device/provisioning/transport/ClientApiVersionHelper.cs
@@ -5,9 +5,8 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
 {
     internal class ClientApiVersionHelper
     {
+        // TODO: Split ApiVersionName as it is only used by HTTP and AMQP.
         public const string ApiVersionName = "api-version";
-
-        // TODO: Change this to GA version when the service is ready.
         public const string ApiVersion = "2017-11-15";
     }
 }
--- a/common/src/device/transport/mqtt/ClientWebSocketChannel.cs
+++ b/common/src/device/transport/mqtt/ClientWebSocketChannel.cs
@@ -1,0 +1,343 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
+{
+    using System;
+    using System.Diagnostics.Contracts;
+    using System.Net;
+    using System.Net.WebSockets;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using DotNetty.Buffers;
+    using DotNetty.Transport.Channels;
+    using System.Diagnostics;
+
+#pragma warning disable CA1001 // Types that own disposable fields should be disposable - WS is owned by the caller.
+    internal class ClientWebSocketChannel : AbstractChannel
+#pragma warning restore CA1001 // Types that own disposable fields should be disposable
+    {
+        private readonly ClientWebSocket _webSocket;
+        private readonly CancellationTokenSource _writeCancellationTokenSource;
+        private bool _active;
+
+        internal bool ReadPending { get; set; }
+
+        internal bool WriteInProgress { get; set; }
+
+        public ClientWebSocketChannel(IChannel parent, ClientWebSocket webSocket)
+            : base(parent)
+        {
+            _webSocket = webSocket;
+            _active = true;
+            Metadata = new ChannelMetadata(false, 16);
+            Configuration = new DefaultChannelConfiguration(parent);
+            _writeCancellationTokenSource = new CancellationTokenSource();
+        }
+
+        public override IChannelConfiguration Configuration { get; }
+
+        public override bool Open => _webSocket.State == WebSocketState.Open;
+
+        public override bool Active => _active;
+
+        public override ChannelMetadata Metadata { get; }
+
+        public ClientWebSocketChannel Option<T>(ChannelOption<T> option, T value)
+        {
+            Contract.Requires(option != null);
+
+            Configuration.SetOption(option, value);
+            return this;
+        }
+
+        protected override EndPoint LocalAddressInternal { get; }
+
+        protected override EndPoint RemoteAddressInternal { get; }
+
+        protected override IChannelUnsafe NewUnsafe() => new WebSocketChannelUnsafe(this);
+
+        protected class WebSocketChannelUnsafe : AbstractUnsafe
+        {
+            public WebSocketChannelUnsafe(AbstractChannel channel)
+                : base(channel)
+            {
+            }
+
+            public override Task ConnectAsync(EndPoint remoteAddress, EndPoint localAddress)
+            {
+                throw new NotSupportedException("ClientWebSocketChannel does not support ConnectAsync()");
+            }
+
+            protected override void Flush0()
+            {
+                // Flush immediately only when there's no pending flush.
+                // If there's a pending flush operation, event loop will call FinishWrite() later,
+                // and thus there's no need to call it now.
+                if (((ClientWebSocketChannel)channel).WriteInProgress)
+                {
+                    return;
+                }
+
+                base.Flush0();
+            }
+        }
+
+        protected override bool IsCompatible(IEventLoop eventLoop) => true;
+
+        protected override void DoBind(EndPoint localAddress)
+        {
+            throw new NotSupportedException("ClientWebSocketChannel does not support DoBind()");
+        }
+
+        protected override void DoDisconnect()
+        {
+            throw new NotSupportedException("ClientWebSocketChannel does not support DoDisconnect()");
+        }
+
+        protected override async void DoClose()
+        {
+            try
+            {
+                WebSocketState webSocketState = _webSocket.State;
+                if (webSocketState != WebSocketState.Closed && webSocketState != WebSocketState.Aborted)
+                {
+                    // Cancel any pending write
+                    CancelPendingWrite();
+                    _active = false;
+
+                    using (var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(30)))
+                    {
+                        await _webSocket.CloseAsync(
+                            WebSocketCloseStatus.NormalClosure, 
+                            string.Empty, 
+                            cancellationTokenSource.Token).ConfigureAwait(false);
+                    }
+                }
+            }
+            catch (Exception)
+            {
+                Abort();
+            }
+        }
+
+        protected override async void DoBeginRead()
+        {
+            IByteBuffer byteBuffer = null;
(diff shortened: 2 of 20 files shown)


provisioning/transport/mqtt/transport_mqtt.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26730.12
+VisualStudioVersion = 15.0.27004.2008
```
jasmineymlo: when is it necessary to update this version?

provisioning/transport/mqtt/src/Microsoft.Azure.Devices.Provisioning.Transport.Mqtt.csproj
```diff
@@ -27,6 +51,9 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="DotNetty.Codecs.Mqtt" Version="0.4.7" />
```
jasmineymlo: >7 [](start = 66, length = 1)

device client is using 0.4.6...we should align on all clients

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -1,14 +1,37 @@
...
 {
     public class ProvisioningTransportHandlerMqtt : ProvisioningTransportHandler
     {
+        private static MultithreadEventLoopGroup s_eventLoopGroup = new MultithreadEventLoopGroup();
+        private const int MaxMessageSize = 256 * 1024;
+        private const int MqttTcpPort = 8883;
+
+        private const string WsMqttSubprotocol = "mqtt";
+        private const string WsScheme = "wss";
+        private const int WsPort = 443;
```
jasmineymlo: we have these (similar) define in WebSocketConstants.  Check if we can consolidate.

common/src/device/provisioning/transport/ClientApiVersionHelper.cs
```diff
@@ -5,9 +5,8 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
 {
     internal class ClientApiVersionHelper
     {
+        // TODO: remove - this is protocol specific.
```
jasmineymlo: >specific [](start = 43, length = 8)

both http and amqp are using it.  mqtt should use it as well??

CIPop: You're right - I missed the one in HTTP.

provisioning/transport/mqtt/src/Microsoft.Azure.Devices.Provisioning.Transport.Mqtt.csproj
CIPop: Sure, I'll try to increase the DeviceClient version as well and see if that passes the tests.

provisioning/transport/mqtt/transport_mqtt.sln
CIPop: VS got an update and for some reason it likes to update the SLNs that it touched. 
I'll just go with it unless you see a problem opening SLNs on systems that are not up to date.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -0,0 +1,343 @@
...
+                while (allocHandle.ContinueReading());
+
+                allocHandle.ReadComplete();
+                ReadPending = false;
+                Pipeline.FireChannelReadComplete();
+            }
+            catch (Exception e)
+            {
+                // Since this method returns void, all exceptions must be handled here.
+                byteBuffer?.Release();
```
nayato: I'd recommend SafeRelease here instead as release may throw

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -1,14 +1,37 @@
...
+using System.Net;
+using System.Net.WebSockets;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Azure.Devices.Shared;
 
 namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
 {
     public class ProvisioningTransportHandlerMqtt : ProvisioningTransportHandler
```
jasmineymlo: name this to suggest x509 handling

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -0,0 +1,343 @@
...
+                WebSocketState webSocketState = _webSocket.State;
+                if (webSocketState != WebSocketState.Closed && webSocketState != WebSocketState.Aborted)
+                {
+                    // Cancel any pending write
+                    CancelPendingWrite();
+                    _active = false;
+
+                    using (var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(30)))
+                    {
+                        await _webSocket.CloseAsync(
```
nayato: if it acts the same as server version, it may deadlock. you can check our server websocket channel impl for details on how we work around this.

common/src/device/transport/mqtt/ClientWebSocketChannelConfig.cs
```diff
@@ -0,0 +1,114 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
+{
+    using System;
+    using System.Diagnostics.Contracts;
+    using DotNetty.Buffers;
+    using DotNetty.Transport.Channels;
+
+    internal class ClientWebSocketChannelConfig : IChannelConfiguration
```
nayato: can't you inherit from https://github.com/Azure/DotNetty/blob/dev/src/DotNetty.Transport/Channels/DefaultChannelConfiguration.cs?

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -0,0 +1,423 @@
...
+using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
+{
+    internal class ProvisioningChannelHandlerAdapter : ChannelHandlerAdapter
```
nayato: no liveness support for now?

CIPop: Good point but the comment is a bit ambiguous: please add more information. 
I do see one problem with the cooperative cancellation not being checked if no packets are received. Anything else?

(PS: I do see this is a very good candidate for TLA modeling.)

common/src/device/transport/mqtt/ClientWebSocketChannelConfig.cs
CIPop: Will do.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
CIPop: You mean https://github.com/Microsoft/dotnet/issues/560?

nayato: Yes. Good to see the fix is coming. Meanwhile it's likely a good idea not to deadlock executor's thread

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
nayato: No, I mean MQTT level keep alive handling with pings

CIPop: Provisioning should take no longer than a few seconds. I don't think that would be required in this case.
@simonporter @rajeevmv - Please confirm.

rajeevmv: It normally takes a few seconds. However, it can be delayed for a variety of reasons - like IotHub not responding for example.

---
In reply to: [153912021](https://github.com/Azure/azure-iot-sdk-csharp/pull/248#discussion_r153912021) [](ancestors = 153912021)

provisioning/transport/mqtt/src/Microsoft.Azure.Devices.Provisioning.Transport.Mqtt.csproj
CIPop: This is more complicated than I originally thought. Tracking with #250.

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
CIPop: The file containing these defines, for some reason, mixes AMQP and MQTT. Leaving this as a TODO.

common/src/device/provisioning/transport/ClientApiVersionHelper.cs
CIPop: After checking - this is protocol specific to HTTP and AMQP - it should't exist for MQTT.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
CIPop: @nayato there's no such API on `IByteBuffer`.

common/src/device/transport/mqtt/ClientWebSocketChannelConfig.cs
CIPop: I removed this class completely and used DefaultChannelConfiguration instead.

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
CIPop: This is part of the public API and will not be renamed (in case we do allow TPM with MQTT). I've renamed the internals instead.

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
simonporter: Yes it can take longer for a variety of reasons especially in the future when we have out of band attestation, external calls etc.

CIPop: Thanks @rajeevmv @simonporter - I've opened #274 to track this.

PR #256, opened 2017-12-01
Fixing EventSource duplication and adding log collection script.

### LINKED ISSUE #252: Intermittent test failures detected in the CI
Disabling the following tests blocking CI:

```
Twin_DeviceReportedPropertiesGracefulShutdownRecovery_Amqp 
Twin_DeviceReportedPropertiesGracefulShutdownRecovery_AmqpWs

```

CIPop:
`EventSource`s cannot be defined with the same name in multiple assemblies. This is creating a different `EventSource` for each binary and consolidating common code.
I have some small fixes for the logging code, and e2e tests. Adding a script for log collection for Windows. Linux collection is also possible using [LTTNG and perfcollect](http://blogs.microsoft.co.il/sasha/2017/03/30/tracing-runtime-events-in-net-core-on-linux/).

I'm re-enabling temporarily disabled tests:
Fixes #252


tools/CaptureLogs/iot_providers.txt
```diff
@@ -0,0 +1,5 @@
+"{e927240b-7198-5cc8-72f1-7ddcf31bb8cb}"
```
jasmineymlo: Add readme to capture guid to event source mapping and delete the guid from source code comments

CIPop: 👍 We need to point to https://github.com/dotnet/corefx/blob/master/Documentation/debugging/windows-instructions.md#traces and also document our own mapping.
Adding a quick readme.md but we'll need to enhance it later.

CIPop: Done: added readme.md in the tools folder.

PR #271, opened 2017-12-06
Add support for ModuleAuthenticationWithTokenRefresh 

varunpuranik:
# Description of the problem
No support exists for ModuleAuthenticationWithTokenRefresh. Adding that to enable modules as well.

# Description of the solution
Added ModuleAuthenticationWithTokenRefresh and its implementation ModuleAuthenticationWithSakRefresh to enable support for modules as well.


device/Microsoft.Azure.Devices.Client/ModuleAuthenticationWithSakRefresh.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            var builder = new SharedAccessSignatureBuilder()
+            {
+                Key = _connectionString.SharedAccessKey,
+                TimeToLive = TimeSpan.FromSeconds(suggestedTimeToLive),
+            };
+
+            if (_connectionString.SharedAccessKeyName == null)
+            {
+                builder.Target = "{0}/devices/{1}/modules/{2}".FormatInvariant(
```
CIPop: Except the target, what is different from the other DeviceAuthenticationWithSakRefresh?

device/Microsoft.Azure.Devices.Client/ModuleAuthenticationWithTokenRefresh.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Microsoft.Azure.Devices.Client
+{
+    using System;
+    using Microsoft.Azure.Devices.Client.Extensions;
+
+    /// <summary>
+    /// Authentication method that uses a shared access signature token and allows for token refresh. 
+    /// </summary>
+    public abstract class ModuleAuthenticationWithTokenRefresh : DeviceAuthenticationWithTokenRefresh
```
CIPop: Would simply adding "ModuleId" to the current type achieve the same functionality?

device/Microsoft.Azure.Devices.Client.Shared/ITokenRefresher.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Microsoft.Azure.Devices.Client
+{
+    using System;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Interface to get a new token for the device or module
+    /// </summary>
+    interface ITokenRefresher
```
CIPop: Adding an interface is usually limiting the expansion of this type, especially if it's public.
If any kinds of additions are possible to this (and I believe there are), we should instead use `abstract class` which allows for default implementations.

varunpuranik: I am not sure I understand how adding the interface limits this. Especially when you think that the interface is internal only. 
We can use the Device* abstract class, but that is not what we have done for other AuthenticationMethods.

device/Microsoft.Azure.Devices.Client/ModuleAuthenticationWithSakRefresh.cs
varunpuranik: Not much.. just that it has the moduleId

device/Microsoft.Azure.Devices.Client/ModuleAuthenticationWithTokenRefresh.cs
varunpuranik: It would... it is just now what we have done with other AuthenticationMethod types...

device/Microsoft.Azure.Devices.Client.Shared/ITokenRefresher.cs
CIPop: From offline discussion: we can extract an `internal AuthenticateWithTokenRefresh` and move all common code within it exposing only the differences.
That way `ModuleAuthenticationWithTokenRefresh` won't have to publicly inherit from `DeviceAuthenticationWithTokenRefresh` (in case they diverge in public API in the future.

device/Microsoft.Azure.Devices.Client/ModuleAuthenticationWithTokenRefresh.cs
CIPop: We should try to limit code duplication as much as possible: it could be that a different abstract can be extracted between this and the `Device` equivalent.

device/tests/Microsoft.Azure.Devices.Client.Test/ModuleAuthenticationWithTokenRefreshTests.cs
```diff
@@ -0,0 +1,260 @@
...
+        {
+            var rnd = new Random();
+            var rndBytes = new byte[32];
+            rnd.NextBytes(rndBytes);
+
+            TestSharedAccessKey = Convert.ToBase64String(rndBytes);
+        }
+
+        [TestMethod]
+        public void ModuleAuthenticationWithTokenRefresh_Ctor_WrongArguments_Fail()
```
CIPop: If we extract a common abstract, we should unify testing against that (a single copy of the tests) and only test differences between `Device` and `Method`.

device/Microsoft.Azure.Devices.Client/AuthenticationWithTokenRefresh.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// Gets a snapshot of the security token associated with the device. This call is thread-safe.
+        /// </summary>
+        public async Task<string> GetTokenAsync(string iotHub)
+        {
+            if (!IsExpiring)
+            {
+                return _token;
+            }
+
+            await _lock.WaitAsync();
```
CIPop: Please add `.ConfigureAwait(false)`

device/Microsoft.Azure.Devices.Client/AuthenticationWithTokenRefresh.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            try
+            {
+                if (!IsExpiring)
+                {
+                    return _token;
+                }
+
+                _token = await SafeCreateNewToken(iotHub, _suggestedTimeToLiveSeconds);
```
CIPop: Please add .ConfigureAwait(false)

device/Microsoft.Azure.Devices.Client/ModuleAuthenticationWithSakRefresh.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public ModuleAuthenticationWithSakRefresh(
+            string deviceId,
+            string moduleId,
+            IotHubConnectionString connectionString) : base(deviceId, moduleId)
+        {
+            _connectionString = connectionString;
+        }
+
+        protected override Task<string> SafeCreateNewToken(string iotHub, int suggestedTimeToLive)
```
CIPop: Nit: to avoid code duplication you could extract the common implementation into the abstract (similar to how you did with Populate).

device/tests/Microsoft.Azure.Devices.Client.Test/ModuleAuthenticationWithTokenRefreshTests.cs
```diff
@@ -0,0 +1,161 @@
...
+            TestAssert.Throws<ArgumentNullException>(() => new TestImplementation(TestDeviceId, null));
+            TestAssert.Throws<ArgumentNullException>(() => new TestImplementation("   ", TestModuleId));
+            TestAssert.Throws<ArgumentNullException>(() => new TestImplementation(TestDeviceId, "  "));
+            TestAssert.Throws<ArgumentOutOfRangeException>(() => new TestImplementation(TestDeviceId, TestModuleId, -1, 10));
+            TestAssert.Throws<ArgumentOutOfRangeException>(() => new TestImplementation(TestDeviceId, TestModuleId, 60, -1));
+            TestAssert.Throws<ArgumentOutOfRangeException>(() => new TestImplementation(TestDeviceId, TestModuleId, 60, 101));
+        }
+
+        [TestMethod]
+        public void ModuleAuthenticationWithTokenRefresh_Ctor_DefaultsGetProperties_Ok()
```
CIPop: This is already tested in the base class UnitTest. If you want to, you can test the particular behavior for Module (i.e. token is created with the module name, etc)
Otherwise I think you can delete these duplicates as you've deleted the Device ones.

varunpuranik: I thought in this one case, the some duplicate test code is okay. Want to have a test that tests it for the derived type.

device/Microsoft.Azure.Devices.Client/ModuleAuthenticationWithSakRefresh.cs
varunpuranik: You know I thought of that, but in this case it is not very clear how to do that, while keeping the code logically correct. Refactoring the code to the base will leave the code in patches (you won't be able to look at the code an understand what is going on). Maybe you have a better idea, but I will leave it like this for now.

device/Microsoft.Azure.Devices.Client/AuthenticationWithTokenRefresh.cs
varunpuranik: Done.

device/Microsoft.Azure.Devices.Client/AuthenticationWithTokenRefresh.cs
varunpuranik: Done

PR #279, opened 2017-12-08
Fix DPS matrix

yzhong94:
Update TPM over MQTT


CIPop:
@yzhong94 Please also update to no support for Provisioning over WebSockets (both AMQP and MQTT).

PR #285, opened 2017-12-13
WIP (do not merge): Fixing IoT Hub Device Client memory leak

### LINKED ISSUE #85: Massive memory leak in RetryDelegatingHandler

### LINKED ISSUE #263: AMQP Token doesn't refresh - Intermittently failing : Microsoft.Azure.Devices.E2ETests.DeviceTokenRefreshE2ETests.DeviceClient_TokenIsRefreshed_Ok_Amqp
```
Assert.IsTrue failed. [12/3/2017 9:16:47 AM] Token should have been refreshed after TTL expired.

RemoveDeviceAsync Enumerating devices.
RemoveDeviceAsync Removing E2E_Message_TokenRefresh_0b096a2d-970b-4226-98ef-54f36975f095.
Removing device E2E_Message_TokenRefresh_0b096a2d-970b-4226-98ef-54f36975f095
Device successfully removed
Creating device E2E_Message_TokenRefresh_7dd8c1ef-f4b6-4ac4-9845-5df56aee6c55
[12/3/2017 9:16:36 AM] OpenAsync
[12/3/2017 9:16:37 AM] Refresher: Creating new token 0
Token: SharedAccessSignature sr=[...] se=1512292603
[12/3/2017 9:16:37 AM] SendEventAsync (1)
[12/3/2017 9:16:41 AM] Waiting 6 seconds.
[12/3/2017 9:16:47 AM] SendEventAsync (2)

Microsoft.Azure.Devices.E2ETests.DeviceTokenRefreshE2ETests.<DeviceClient_TokenIsRefreshed_Internal>d__4.MoveNext() in D:\jenkins\workspace\csharp-windows-vs2017\e2e\Microsoft.Azure.Devices.E2ETests\DeviceTokenRefreshE2ETests.cs: line 79
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
Microsoft.Azure.Devices.E2ETests.DeviceTokenRefreshE2ETests.<DeviceClient_TokenIsRefreshed_Internal>d__4.MoveNext() in D:\jenkins\workspace\csharp-windows-vs2017\e2e\Microsoft.Azure.Devices.E2ETests\DeviceTokenRefreshE2ETests.cs: line 98
System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
Microsoft.Azure.Devices.E2ETests.DeviceTokenRefreshE2ETests.<DeviceClient_TokenIsRefreshed_Ok_Amqp>d__2.MoveNext() in D:\jenkins\workspace\csharp-windows-vs2017\e2e\Microsoft.Azure.Devices.E2ETests\DeviceTokenRefreshE2ETests.cs: line 29
System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
System.Runtime.CompilerServices.TaskAwaiter.GetResult()

```

CIPop:
I found several reasons for the memory leak as described in #85. This is also fixing the TaskHelper cancellation mechanism and is adding propagation and cooperative cancellation throughout the I/O pipeline.
The ConfigureAwait should in theory fix #263. We can reopen if we start seeing the test fail again in the CI.

After this fix, running the repro test (added in this PR) I'm seeing this memory usage evolution:
image

image

Fixes #85, #263


device/Microsoft.Azure.Devices.Client.NetStandard/Microsoft.Azure.Devices.Client.NetStandard.csproj
```diff
@@ -50,6 +50,10 @@
     <DocumentationFile>bin\Release_Delay_Sign\Microsoft.Azure.Devices.Client.XML</DocumentationFile>
   </PropertyGroup>
 
+  <PropertyGroup>
```
jasmineymlo: where is this used?

device/Microsoft.Azure.Devices.Client/Common/TaskHelpers.cs
```diff
@@ -384,20 +390,25 @@ public static Task WithTimeout(this Task task, TimeSpan timeout, Func<string> er
...
+                    delayCts.Cancel();
+                    await task.ConfigureAwait(false);
                     return;
                 }
             }
 
+            using (var operationCts = CancellationTokenSource.CreateLinkedTokenSource(token))
+            {
+                operationCts.Cancel();
+            }
```
jasmineymlo: can you explain why do we need this? what i understand is this will create a cancellation token source and cancel immediately, but it doesn't affect any on original token.

device/Microsoft.Azure.Devices.Client/Common/TaskHelpers.cs
```diff
@@ -384,20 +390,25 @@ public static Task WithTimeout(this Task task, TimeSpan timeout, Func<string> er
 
             if (task.IsCompleted || (timeout == Timeout.InfiniteTimeSpan && token == CancellationToken.None))
             {
-                await task;
+                await task.ConfigureAwait(false);
                 return;
             }
 
-            using (var cts = CancellationTokenSource.CreateLinkedTokenSource(token))
+            using (var delayCts = new CancellationTokenSource())
```
jasmineymlo: >new CancellationTokenSource() [](start = 34, length = 29)

why do you remove CreateLinkedTokenSource method call from line 397? if input cancellation token in line 384 is canceled from caller, line 399 still need to wait till timeout trigger.

device/Microsoft.Azure.Devices.Client.NetStandard/Microsoft.Azure.Devices.Client.NetStandard.csproj
CIPop: It's not currently used but I'd rather not remove it since I'll need it for the next round of build enhancements.

device/Microsoft.Azure.Devices.Client/Common/TaskHelpers.cs
CIPop: That was the bug: there are 2 separate tokens here:
- delayCts is used to cancel the Task.Delay
- operationCts is used to attach to the token given by the user

[Here](https://github.com/Azure/azure-iot-sdk-csharp/pull/285/files#diff-5c60a411b2230a22a19ebff0240453beR401) the user operation completed before the Task.Delay and so we are cancelling the delay and returning the task result (by awaiting `task`).

[Here](https://github.com/Azure/azure-iot-sdk-csharp/pull/285/files#diff-5c60a411b2230a22a19ebff0240453beR407) the delay completed before the user operation and so we are cancelling the user operation (based on the `token` passed as argument [here](https://github.com/Azure/azure-iot-sdk-csharp/pull/285/files#diff-5c60a411b2230a22a19ebff0240453beL378)) then throwing the TimeoutException.

The original code was incorrectly combining the two processes into a single call which meant that if the operation completed successfully, the code would then try to cancel it [here](https://github.com/Azure/azure-iot-sdk-csharp/pull/285/files#diff-5c60a411b2230a22a19ebff0240453beL395).

device/Microsoft.Azure.Devices.Client/Transport/RetryDelegatingHandler.cs
```diff
@@ -12,14 +12,15 @@ namespace Microsoft.Azure.Devices.Client.Transport
...
+    using System.Diagnostics;
 
     internal class RetryDelegatingHandler : DefaultDelegatingHandler
     {
         const int UndeterminedPosition = -1;
         const string StopRetrying = "StopRetrying";
 
         // RetryCount is used for testing purpose and is equal to MaxValue in prod.
-        internal static int RetryCount = Int32.MaxValue;
+        internal static int RetryCount = 10;
```
jasmineymlo: >10 [](start = 41, length = 2)

This is breaking change for current customer who is doing infinite retries.  The existing behavior is that retry is limited by operation timeout ONLY...but with this change, retry is also limited by the retry count which means that retry could stop before the operation timeout...  This is in violation of spec.

device/Microsoft.Azure.Devices.Client/Common/TaskHelpers.cs
jasmineymlo: So what happen if task and Task.Delay is not finished yet, but pass-in token is canceled? we want to stop asasp, right? 

---
In reply to: [156762250](https://github.com/Azure/azure-iot-sdk-csharp/pull/285#discussion_r156762250) [](ancestors = 156762250)

device/Microsoft.Azure.Devices.Client/Transport/RetryDelegatingHandler.cs
CIPop: @yzhong94 @tameraw @olivierbloch In that case I think the spec is wrong. For more details and my recommendation for IoT developers in case of permanent network failures, please see https://github.com/Azure/azure-iot-sdk-csharp/issues/85#issuecomment-350385365

device/Microsoft.Azure.Devices.Client/Transport/RetryDelegatingHandler.cs
```diff
@@ -98,12 +101,9 @@ public RetryDelegatingHandler(IPipelineContext context)
                 var sendState = new SendMessageState();
                 await this.internalRetryPolicy.ExecuteAsync(() =>
                 {
-                    if (cancellationToken.IsCancellationRequested)
-                    {
-                        return Common.TaskConstants.Completed;
-                    }
+                    cancellationToken.ThrowIfCancellationRequested();
```
jasmineymlo: >ThrowIfCancellationRequested [](start = 38, length = 28)

In case of cancellation, why should we throw?  This looks more like a normal operation than exception?  If cancellation is due to operation timeout.... the exception is throw from TaskHelper.WithTimout().

CIPop: This is the standard TPL pattern: if the operation timed out or was cancelled for other reasons through the cancellation token, the Task.Status should be set to "Cancelled" and not "Done".

Setting it to "Done" would make the application believe that the operation completed successfully (e.g. telemetry was sent successfully) which is not the case. 

An application can be written to avoid await/GetResult() and instead query its [Status](https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.status(v=vs.110).aspx) which in today's implementation is incorrectly set to "Done".

device/Microsoft.Azure.Devices.Client/Transport/RetryDelegatingHandler.cs
yzhong94: Just so I'm clear, MaxValue is set by the customer right?  If it's not set, then it will do infinite retries?  In this case, why should we hard code it to 10?

CIPop: A retry policy can be defined by the customer (see the recommended workaround in https://github.com/Azure/azure-iot-sdk-csharp/issues/85#issuecomment-350385365).

This is for the default retry count which should not be infinite (or MaxValue).

jasmineymlo: Default retry count should be the number of retries fitting into the default 4 minutes window (or whatever specified by the user) using a specific retry strategy.  It shouldn't be hardcoded to 10 either...

CIPop: > Default retry count should be the number of retries fitting into the default 4 minutes window

Why should we mix number of retries with retry times? Right now, if the request fails immediately we execute for about 1 minute with most likely no chance of recovery (e.g. if the cable is pulled):

image

Instead of doing this, we should allow the application to understand what the failure is and stop retrying due to the number of retries failed not the 4 minute timeout. (In this case it should get the SocketException with media failure instead of getting a TimeoutException.)

jasmineymlo: I am only speaking from the view point of what the spec has defined which was previously agreed upon within the team.  I will leave it to @yzhong94 and  @olivierbloch to comment.  We will need to sync this with other SDK owners if we decided to make any modification ...

yzhong94: Agree, if we are changing the behavior of something already released, we need to have a discussion.

device/Microsoft.Azure.Devices.Client/Common/TaskHelpers.cs
CIPop: The passed in `token` in these helpers represents the same token passed to the original operation and so when it's cancelled, the operation is cancelled. `task` will complete in this case which means `delayCts` will cancel the Delay task.

The whole `TaskHelpers` is a mess and must be removed in the future completely. The case you're describing is the way a task should be cancelled: only through the cancellationToken and not by starting another delay task. CancellationTokenSource supports the timeout mechanism.

device/Microsoft.Azure.Devices.Client/Transport/RetryDelegatingHandler.cs
CIPop: @jasmineymlo @yzhong94 I've discussed this offline with @olivierbloch and @jebrando and agreed to add a minimum backoffTime + executionTime in case the operation completes very fast.
The new default retry curve for the case where the operation completes immediately is:

image

I've added back the infinite retry and opened #290 to track proper error reporting after retry.

device/Microsoft.Azure.Devices.Client/Common/TaskHelpers.cs
CIPop: @jasmineymlo I misunderstood CreateLinkedTokenSource(token). I originally thought calling the linked cts will cancel `token` but instead the linked cts will be set to cancelled.
The leak is still present with infinite retries because of this (the operation is not actually cancelled...). I need to figure out a way to cancel the original token when the operation times out.

shared/Microsoft.Azure.Devices.Shared.NetStandard/Microsoft.Azure.Devices.Shared.NetStandard.csproj
```diff
@@ -12,7 +12,9 @@
     <AssemblyName>Microsoft.Azure.Devices.Shared</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <FileAlignment>512</FileAlignment>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors> 
+    <!-- TODO #176: Enable warnings as errors. -->
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
tameraw: >false [](start = 27, length = 5)

Why are we turning these off now?

device/Microsoft.Azure.Devices.Client/Transport/RetryDelegatingHandler.cs
yzhong94: Okay, thanks!  I like this solution

shared/Microsoft.Azure.Devices.Shared.NetStandard/Microsoft.Azure.Devices.Shared.NetStandard.csproj
CIPop: This project was missing the FXCop include [below](https://github.com/Azure/azure-iot-sdk-csharp/pull/285/commits/7b962c2b8932f2471f280e667c190e235b6c1dc1#diff-5ec035dad8b7639dca1223004f53893aR84). Adding it triggered a huge amount of warnings. We need to spend time fixing all of those before properly enabling the flag.

CIPop:
Marking this as WIP: I'm now investigating high memory usage in a different portion.

CIPop:
Closing - needs more work.

PR #289, opened 2017-12-13
fix debug build

jasmineymlo:


provisioning/service/src/Config/X509Certificates.cs
```diff
@@ -7,6 +7,7 @@
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 
+#if (RELEASE)
```
CIPop: This is the wrong file to add these statements. Instead we can have a separate .cs file for Debug. 
Release builds should not run UnitTests. I'm working on addressing these in master.

PR #297, opened 2017-12-17
Add query to Provisioning Service Client

mamokarz:
Add query to Provisioning Service Client


provisioning/service/samples/ProvisioningServiceBulkOperation/Program.cs
```diff
@@ -85,18 +85,17 @@ public static void Main(string[] args)
...
-
-                //while (query.hasNext())
-                //{
-                //    Console.WriteLine("\nQuerying the next enrollments...");
-                //    QueryResult queryResult = query.next();
-                //    Console.WriteLine(queryResult);
-                //}
+                Console.WriteLine("\nCreating a query for enrollments...");
+                QuerySpecification querySpecification = new QuerySpecification("SELECT * FROM enrollments");
+                using (IQuery query = provisioningServiceClient.CreateIndividualEnrollmentQuery(querySpecification, 2))
```
zolvarga: >2 [](start = 116, length = 1)

Just out of curiosity: what is this magic number?

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -35,24 +35,23 @@ internal sealed class ContractApiHttp : IContractApiHttp
         private HttpClient _httpClientObj;
         private HttpClient _httpClientObjWithPerRequestTimeout;
         private bool _isDisposed;
-        private readonly TimeSpan _defaultOperationTimeout;
+
+        static readonly TimeSpan DefaultOperationTimeout = TimeSpan.FromSeconds(100);
```
zolvarga: >100 [](start = 80, length = 3)

Use constants

provisioning/service/samples/ProvisioningServiceBulkOperation/Program.cs
mamokarz: This is just a random number, to be different than 0, it is a page size, I'll define a constant with a nice name to better clarify it.

mamokarz: Done

provisioning/service/src/Contract/ContractApiHttp.cs
mamokarz: This is the constant definition.

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -384,6 +383,126 @@ private static bool IsMappedToException(HttpResponseMessage message)
...
+                msg.Headers.Add(HttpRequestHeader.UserAgent.ToString(), Utils.GetClientVersion());
+
+                Task modifyRequestMessageAsync(HttpRequestMessage requestMsg, CancellationToken token)
+                {
+                    InsertIfMatch(requestMsg, ifMatch);
+                    AddCustomHeaders(requestMsg, customHeaders);
+                    return Task.FromResult(default(VoidTaskResult));
+                }
+
+                await modifyRequestMessageAsync(msg, cancellationToken);
```
CIPop: ? I don't understand the logic here. 
Why are we creating a task then awaiting it in the same context?

Also, await is missing ConfigureAwait(false).

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -384,6 +383,126 @@ private static bool IsMappedToException(HttpResponseMessage message)
...
+
+                try
+                {
+                    using (HttpResponseMessage httpResponse = await _httpClientObj.SendAsync(msg, cancellationToken))
```
CIPop: Whenever awaiting use .ConfigureAwait(false) in library code.

provisioning/service/src/Contract/ContractApiResponse.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Net;
+
+namespace Microsoft.Azure.Devices.Provisioning.Service
+{
+    public class ContractApiResponse
```
CIPop: Public?

provisioning/service/src/Contract/ContractApiResponse.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        public HttpStatusCode StatusCode { get; internal set; }
```
CIPop: Let's not use internal set with no validation.
Any reason not to have a ctor and make this immutable?

provisioning/service/src/Contract/IContractApiHttp.cs
```diff
@@ -13,6 +13,14 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
 {
     internal interface IContractApiHttp : IDisposable
     {
+        Task<ContractApiResponse> RequestAsync(
```
CIPop: What's the reason for this interface?

provisioning/service/src/IQuery.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Provisioning.Service
+{
+    public interface IQuery : IDisposable
```
CIPop: We should avoid public interfaces where the API can change in the future.
If needed, this should instead be an abstract class that allows addition of new APIs.

Please add the reasoning on why we need this interface in the first place.

mamokarz:
Thanks @zolvarga, @CIPop!

provisioning/service/src/Manager/IndividualEnrollmentManager.cs
```diff
@@ -184,26 +186,37 @@ internal static class IndividualEnrollmentManager
...
-            /* SRS_INDIVIDUAL_ENROLLMENT_MANAGER_21_015: [The CreateQuery shall return a new Query for IndividualEnrollments.] */
+            if(querySpecification == null)
+            {
+                throw new ArgumentNullException("querySpecification cannot be null");
+            }
 
-            throw new NotSupportedException("Query is not supported yet");
+            if(pageSize < 0)
+            {
+                throw new ArgumentException("pageSize cannot be negative");
```
CIPop: Please use `nameof()` everywhere the name of a parameter is required. Allows keeping the code and text in-sync. End sentences in ".".
`$"{nameof(pageSize)} cannot be negative."`

provisioning/service/src/Manager/IndividualEnrollmentManager.cs
```diff
@@ -184,26 +186,37 @@ internal static class IndividualEnrollmentManager
...
+            CancellationToken cancellationToken, 
+            int pageSize = 0)
         {
-            //TODO: Implement.
-
             /* SRS_INDIVIDUAL_ENROLLMENT_MANAGER_21_014: [The CreateQuery shall throw ArgumentException if the provided querySpecification is null.] */
-            /* SRS_INDIVIDUAL_ENROLLMENT_MANAGER_21_015: [The CreateQuery shall return a new Query for IndividualEnrollments.] */
+            if(querySpecification == null)
+            {
+                throw new ArgumentNullException("querySpecification cannot be null");
```
CIPop: Avoid adding unnecessary text where the exception will properly construct it:
`ArgumentNullException(nameof(querySpecification));`

provisioning/service/src/Config/QueryResult.cs
```diff
@@ -0,0 +1,173 @@
...
+        public string ContinuationToken { get; private set; }
+
+        /// <summary>
+        /// CONSTRUCTOR
+        /// </summary>
+        /// <param name="typeString">the <code>string</code> with type of the content in the body. It cannot be <code>null</code>.</param>
+        /// <param name="bodyString">the <code>string</code> with the body in a JSON list format. It cannot be <code>null</code>, or empty, if the type is different than `unknown`.</param>
+        /// <param name="continuationToken">the <code>string</code> with the continuation token. It can be <code>null</code>.</param>
+        /// <exception cref="ArgumentException">if one of the provided parameters is invalid.</exception>
+        internal QueryResult(string typeString, string bodyString, string continuationToken)
```
CIPop: This should be public: developers need a way to construct types like these for their own unit testing.

provisioning/service/samples/ProvisioningServiceEnrollment/Program.cs
```diff
@@ -84,18 +84,17 @@ public static void Main(string[] args)
...
-                //                .createSqlQuery();
-                //Query query = provisioningServiceClient.createIndividualEnrollmentQuery(querySpecification);
-
-                //while (query.hasNext())
-                //{
-                //    Console.WriteLine("\nQuerying the next enrollments...");
-                //    QueryResult queryResult = query.next();
-                //    Console.WriteLine(queryResult);
-                //}
+                Console.WriteLine("\nCreating a query for enrollments...");
```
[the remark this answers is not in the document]
mamokarz: the `...` is to indicate the it is a processing. The `\n` is to better organize the result.

provisioning/service/src/Config/QueryResult.cs
mamokarz: As a general ruler, we do not expose unnecessary methods, this constructor was made to fill this object with the server result. Make it public will create a unnecessary public API.

provisioning/service/src/Config/QueryResult.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+            else
+            {
+                ContinuationToken = continuationToken;
+            }
+
+            /* SRS_QUERY_RESULT_21_002: [The constructor shall throw ArgumentException if the provided body is null or empty and the type is not `unknown`.] */
+            if ((Type != QueryResultType.Unknown) && string.IsNullOrWhiteSpace(bodyString))
+            {
+                throw new ArgumentException("body cannot be null or empty");
```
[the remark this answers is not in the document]
mamokarz: Done

provisioning/service/src/Contract/ContractApiHttp.cs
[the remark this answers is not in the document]
mamokarz: No, it cannot be const. Done

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -384,6 +383,126 @@ private static bool IsMappedToException(HttpResponseMessage message)
...
+            string ifMatch,
+            CancellationToken cancellationToken)
+        {
+            ContractApiResponse response;
+
+            using (HttpRequestMessage msg = new HttpRequestMessage(httpMethod, requestUri))
+            {
+                msg.Content = new StringContent(body, Encoding.UTF8, CommonConstants.MediaTypeForDeviceManagementApis);
+
+                msg.Headers.Add(HttpRequestHeader.Authorization.ToString(), _authenticationHeaderProvider.GetAuthorizationHeader());
```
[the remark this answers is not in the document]
mamokarz: I dont' understand it.

provisioning/service/src/Contract/ContractApiHttp.cs
mamokarz: function removed

mamokarz: done

provisioning/service/src/Contract/ContractApiHttp.cs
mamokarz: Done

provisioning/service/src/Contract/ContractApiResponse.cs
mamokarz: Internal.

mamokarz: Done

provisioning/service/src/Contract/ContractApiResponse.cs
mamokarz: Ctor added.

mamokarz: Done

provisioning/service/src/Contract/IContractApiHttp.cs
mamokarz: This is a generic version that can be used instead of the other ones. The idea is replace all the others by this one.

provisioning/service/src/Config/QueryResult.cs
CIPop: I think it's OK not to expose this: users can still wrap this in their own interface and use mock that instead.

provisioning/service/src/Contract/ContractApiHttp.cs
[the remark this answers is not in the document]
CIPop: See for example these handlers:
https://github.com/Azure/azure-iot-sdk-csharp/blob/master/provisioning/transport/http/src/TPM/TpmDelegatingHandler.cs#L17
https://github.com/Azure/azure-iot-sdk-csharp/blob/master/provisioning/transport/http/src/ApiVersionDelegatingHandler.cs#L13

We don't need to do this now but we could improve later.

provisioning/service/src/Contract/ContractApiHttp.cs
CIPop: FYI: my latest PR fixing the IoT Hub DeviceClient fixed ConfigureAwait everywhere else in the code.

provisioning/service/src/IQuery.cs
mamokarz: Convert to an abstract class.

mamokarz: done

provisioning/service/src/Contract/IContractApiHttp.cs
mamokarz: done

provisioning/service/src/Manager/IndividualEnrollmentManager.cs
mamokarz: done

provisioning/service/src/Manager/IndividualEnrollmentManager.cs
mamokarz: done

PR #298, opened 2017-12-18
Fix TwinState on the Provisioning Service Client

mamokarz:
Fix TwinState on the Provisioning Service Client


shared/Microsoft.Azure.Devices.Shared.NetStandard/Microsoft.Azure.Devices.Shared.NetStandard.csproj
```diff
@@ -62,9 +62,6 @@
     <Compile Include="..\Microsoft.Azure.Devices.Shared\TwinCollectionValue.cs" />
     <Compile Include="..\Microsoft.Azure.Devices.Shared\TwinJsonConverter.cs" />
     <Compile Include="..\Microsoft.Azure.Devices.Shared\TwinProperties.cs" />
-    <Compile Include="..\Microsoft.Azure.Devices.Shared\TwinState.cs">
```
CIPop: This is breaking API: we can't simply remove this type from Shared since we've already released this as public API.

shared/Microsoft.Azure.Devices.Shared/TwinState.cs
```diff
@@ -1,35 +0,0 @@
...
-
-    /// <summary>
-    /// Twin State
-    /// </summary>
-    public class TwinState
-    {
-        /// <summary>
-        /// Creates an instance of <see cref="TwinState"/>
-        /// </summary>
-        public TwinState()
```
CIPop: We can't remove this ctor now.

provisioning/service/src/Config/TwinState.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// Getter and setter the <see cref="TwinState"/> properties.
+        /// </summary>
+        [JsonIgnore]
+        public TwinCollection DesiredProperties
+        {
+            get
+            {
+                /* SRS_TWIN_STATE_21_002: [If the _properties is null, the get.DesiredProperties shall return null.] */
+                if (_properties == null)
+                    return null;
```
CIPop: Please always use braces or write the if statement on a single line.

CIPop: This is also a behavior breaking change: it would never return null before.

provisioning/service/src/Config/TwinState.cs
```diff
@@ -0,0 +1,164 @@
...
+                /* SRS_TWIN_STATE_21_003: [The get.DesiredProperties shall return the content of _properties.Desired.] */
+                return _properties.Desired;
+            }
+
+            set
+            {
+                /* SRS_TWIN_STATE_21_004: [If the value is null, the set.DesiredProperties shall set _properties as null.] */
+                if (value == null)
+                {
+                    _properties = null;
```
CIPop: Behavior breaking change.

provisioning/service/src/Config/TwinState.cs
mamokarz: Preview behavior was wrong and under `preview` label. So, we can fix it now.

mamokarz: Braces done

provisioning/service/src/Config/TwinState.cs
mamokarz: Same, it was under `preview`.

shared/Microsoft.Azure.Devices.Shared.NetStandard/Microsoft.Azure.Devices.Shared.NetStandard.csproj
mamokarz: This is not the right place for this file, and its implementation is not correct. It was added by mistake when we created the `preview` version of the provisioning service client. I'll remove it, and apologize if any customer is using it in someway. All implementation must use the new correct version in the .../provisioning/service/src/Config

shared/Microsoft.Azure.Devices.Shared/TwinState.cs
mamokarz: Same comment as above.

mamokarz:
Thanks @CIPop

PR #304, opened 2017-12-19
Unify HTTP requests and exceptions

mamokarz:
Unify HTTP requests and exceptions


provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -414,7 +89,10 @@ private static bool IsMappedToException(HttpResponseMessage message)
 
             using (HttpRequestMessage msg = new HttpRequestMessage(httpMethod, requestUri))
             {
-                msg.Content = new StringContent(body, Encoding.UTF8, CommonConstants.MediaTypeForDeviceManagementApis);
+                if (body != null)
```
CIPop: Nit: `!string.IsNullOrEmpty(body)`

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -487,143 +165,44 @@ private static void ValidateHttpResponse(ContractApiResponse response)
         {
             if (response.StatusCode >= HttpStatusCode.InternalServerError)
             {
-                throw new ProvisioningServiceClientTransientException(response);
+                throw new ProvisioningServiceClientHttpException(response, true);
```
CIPop: Nit: whenever we can't set param names, for code clarity it's nice to have something like 
```C#
 throw new ProvisioningServiceClientHttpException(response, isTransient: true);
```

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -487,143 +165,44 @@ private static void ValidateHttpResponse(ContractApiResponse response)
...
-                }
+                return;
+            }
 
-                if (isMappedToException(responseMsg))
-                {
-                    Exception mappedEx = await MapToExceptionAsync(responseMsg, mergedErrorMapping).ConfigureAwait(false);
-                    throw mappedEx;
-                }
+            if (!ifMatch.StartsWith("\"", StringComparison.OrdinalIgnoreCase))
```
CIPop: Perf: a StringBuilder would do these string operations with less allocations.

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -487,143 +165,44 @@ private static void ValidateHttpResponse(ContractApiResponse response)
...
-            var mapToExceptionFunc = errorMapping[response.StatusCode];
-            var exception = mapToExceptionFunc(response);
-            return await exception.ConfigureAwait(false);
+            requestMessage.Headers.IfMatch.Add(new EntityTagHeaderValue(ifMatch));
         }
 
         /// <summary>
         /// Release all HTTP resources.
         /// </summary>
         public void Dispose()
```
CIPop: This will fail FXCop. 
See https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern for the correct pattern with the `protected void Dispose(bool)` function.

provisioning/service/src/Exceptions/ProvisioningServiceClientException.cs
```diff
@@ -12,104 +11,41 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
...
-        }
-
-        public ProvisioningServiceClientException(string message, string trackingId)
-            : this(message, false, trackingId)
-        {
-        }
-
-        public ProvisioningServiceClientException(string message, bool isTransient, string trackingId)
-            : this(message, null, isTransient, trackingId)
+            : this(message, null, false)
```
CIPop: Nit: Same as above. Where the params are not easily understood it's nice to specify their names.

provisioning/service/src/Manager/EnrollmentGroupManager.cs
```diff
@@ -24,28 +24,36 @@ internal static class EnrollmentGroupManager
...
-            /* SRS_ENROLLMENT_GROUP_MANAGER_28_003: [The CreateOrUpdateAsync shall return an enrollmentGroup object created from the body of the HTTP response.] */
-            return contractApiHttp.PutAsync(
+            ContractApiResponse contractApiResponse = await contractApiHttp.RequestAsync(
+                HttpMethod.Put,
                 GetEnrollmentUri(enrollmentGroup.EnrollmentGroupId),
-                enrollmentGroup,
                 null,
+                JsonConvert.SerializeObject(enrollmentGroup),
                 null,
                 cancellationToken);
```
CIPop: .ConfigureAwait(false);

provisioning/service/src/Manager/EnrollmentGroupManager.cs
```diff
@@ -67,12 +75,21 @@ internal static class EnrollmentGroupManager
...
             /* SRS_ENROLLMENT_GROUP_MANAGER_28_008: [The GetAsync shall sent the Get HTTP request to get the enrollmentGroup information.] */
-            /* SRS_ENROLLMENT_GROUP_MANAGER_28_009: [The GetAsync shall return an EnrollmentGroup object created from the body of the HTTP response.] */
-            return contractApiHttp.GetAsync<EnrollmentGroup>(
+            ContractApiResponse contractApiResponse = await contractApiHttp.RequestAsync(
+                HttpMethod.Get,
                 GetEnrollmentUri(enrollmentGroupId),
                 null,
                 null,
+                null,
                 cancellationToken);
```
CIPop: .ConfigureAwait(false);

provisioning/service/src/Manager/EnrollmentGroupManager.cs
```diff
@@ -81,24 +98,25 @@ internal static class EnrollmentGroupManager
...
             /* SRS_ENROLLMENT_GROUP_MANAGER_28_011: [The DeleteAsync shall sent the Delete HTTP request to remove the enrollmentGroup.] */
-            return contractApiHttp.DeleteAsync(
+            await contractApiHttp.RequestAsync(
+                HttpMethod.Delete,
                 GetEnrollmentUri(enrollmentGroup.EnrollmentGroupId),
-                enrollmentGroup.ETag,
+                null,
                 null,
                 null,
                 cancellationToken);
```
CIPop: Here and on every await : we need to always specify .ConfigureAwait. For libraries that's usually set to false, to allow the task continuation to execute on any available thread.

CIPop: I won't add more comments about ConfigureAwait but this applies to the entire code base.

provisioning/service/src/Contract/ContractApiHttp.cs
mamokarz: done

provisioning/service/src/Contract/ContractApiHttp.cs
mamokarz: done

provisioning/service/src/Contract/ContractApiHttp.cs
mamokarz: done

provisioning/service/src/Exceptions/ProvisioningServiceClientException.cs
mamokarz: done

provisioning/service/src/Manager/EnrollmentGroupManager.cs
mamokarz: done for all await

provisioning/service/src/Manager/EnrollmentGroupManager.cs
mamokarz: done

provisioning/service/src/Manager/EnrollmentGroupManager.cs
mamokarz: done for all.

provisioning/service/src/Contract/ContractApiHttp.cs
mamokarz: done

mamokarz:
Thanks @CIPop

PR #311, opened 2017-12-27
Add Update sample for the provisioning service, add if match etag for…

mamokarz:
Add Update sample for the provisioning service, add if match etag for createOrUpdate and Delete, and fix issues on JSON parser.


mamokarz:
Thanks @jasmineymlo

PR #326, opened 2018-01-09
Added a try-catch around the OnMessageReceived method in the MQTT tra…

oldblindman:
…nsport handler to prevent a thread crash deep in the kernel after the service has run for 24 to 48 hours.

 

# Checklist

Service crashes in the MQTT Message handler with an unknown threading exception after the service runs for 24 to 48 hours.  There are no signs of memory leaks that could cause this issue.  The try-catch prevents the crash, and the service then runs for days.
# Description of the problem

# Description of the solution
Put a try-catch around the whole OnMessageRecieved method.  It is brute force and ugly, but it works.


oldblindman:
A fix for a crash in the MQTT Device handler OnMessageReceived method.

PR #327, opened 2018-01-09
Fix to prevent thread exception in OnMessageReceived in MQTT message handler.

oldblindman:
# Checklist

# Description of the problem
MQTT Transport Handler OnMessageReceived method has an unknown exception when a Windows Service runs for more than 24 hours.

# Description of the solution
Put a try-catch around the OnMessageReceived method body.

### DIFF
--- a/device/Microsoft.Azure.Devices.Client/Transport/Mqtt/MqttTransportHandler.cs
+++ b/device/Microsoft.Azure.Devices.Client/Transport/Mqtt/MqttTransportHandler.cs
@@ -454,25 +454,34 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
         public async void OnMessageReceived(Message message)
         {
-            if ((this.State & TransportState.Open) == TransportState.Open)
+            // Added Try-Catch to avoid unknown thread exception
+            // after running for more than 24 hours
+            try
             {
-                if (message.MqttTopicName.StartsWith(twinResponseTopicPrefix))
+                if ((this.State & TransportState.Open) == TransportState.Open)
                 {
-                    twinResponseEvent(message);
+                    if (message.MqttTopicName.StartsWith(twinResponseTopicPrefix))
+                    {
+                        twinResponseEvent(message);
+                    }
+                    else if (message.MqttTopicName.StartsWith(twinPatchTopicPrefix))
+                    {
+                        await HandleIncomingTwinPatch(message).ConfigureAwait(false);
+                    }
+                    else if (message.MqttTopicName.StartsWith(methodPostTopicPrefix))
+                    {
+                        await HandleIncomingMethodPost(message).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        this.messageQueue.Enqueue(message);
+                        this.receivingSemaphore.Release();
+                    }
                 }
-                else if (message.MqttTopicName.StartsWith(twinPatchTopicPrefix))
-                {
-                    await HandleIncomingTwinPatch(message).ConfigureAwait(false);
-                }
-                else if (message.MqttTopicName.StartsWith(methodPostTopicPrefix))
-                {
-                    await HandleIncomingMethodPost(message).ConfigureAwait(false);
-                }
-                else
-                {
-                    this.messageQueue.Enqueue(message);
-                    this.receivingSemaphore.Release();
-                }
+            }
+            catch(Exception ex)
+            {
+                this.OnError(ex);
             }
         }
 


oldblindman:
Try-catch to prevent thread exception in MQTT Transport Handler method OnMessageReceived.

oldblindman:
One more time

device/Microsoft.Azure.Devices.Client/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -454,26 +454,34 @@ public void OnConnected()
...
+                        await HandleIncomingMethodPost(message).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        this.messageQueue.Enqueue(message);
+                        this.receivingSemaphore.Release();
+                    }
                 }
             }
+            catch(Exception)
```
CIPop: Catching the exception is the right fix but after this is caught, we need to flow the exception to the application, not hide it.
This would be similar to https://github.com/Azure/azure-iot-sdk-csharp/blob/master/provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs#L305
 
I believe that in this context the code needs to call `this.OnError(ex)` to move the state machine into a faulted state and cleanup.

CIPop: Also we should catch non-fatal exceptions:
`            catch (Exception ex) when (!ex.IsFatal())`

device/Microsoft.Azure.Devices.Client/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -454,26 +454,34 @@ public void OnConnected()
 
         public async void OnMessageReceived(Message message)
         {
-            if ((this.State & TransportState.Open) == TransportState.Open)
+            // Added Try-Catch to avoid unknown thread exception
```
CIPop: Shouldn't we add this to all event handlers within this file?

oldblindman:
Cristian,

You lost me on what you want done on the exception handling.  Do you want something like:

catch(Exception ex)
{
var wrapperEx = new ProvisioningTransportException($"{ExceptionPrefix} Error while processing OnMessageReceived.", ex, false, message);
                await FailWithExceptionAsync(context, wrapperEx).ConfigureAwait(false);
}

Or do you want:

catch(Exception ex)
{
var wrapperEx = new ProvisioningTransportException($"{ExceptionPrefix} Error while processing OnMessageReceived.", ex, false, message);
              this.OnError(wrapperEx);
}


Or, do you want something else?

In any case, I do not have a clue what the last parameter inProvisioningTransportException call should be.  Since it was a string in the other example, I just inserted the “message” string.

Thanks,
Gene

oldblindman:
Cristian,

After looking at the code a little closer, I see that the wrapper method is not defined in the MqttTransportHandler context.  Therefore, I presume the call should be:


            catch(Exception ex)
            {
                this.OnError(ex);
            }


However, this routine can throw a fatal exception.  I already know that the exception that I am seeing is fatal, so this would not solve my issue.  Therefore, I have implemented it as follows:

     catch(Exception ex)
            {
                try
                {
                    this.OnError(ex);
                }
                catch(Exception)
                {
                }
            }

I know that this is ugly and probably not acceptable to Microsoft, but I need for the application to continue to run no matter what.

Gene

oldblindman:
Cristian,

I pushed the change to handle the error.  It will catch even fatal errors, since the error we are seeing is fatal.  However, since the OnError method rethrows fatal errors in some cases it should not matter.

I did not add the error processing on other methods since we have not seen any issues with them, and there is some overhead with try catches.

I have pushed the changes for your review.

Gene

device/Microsoft.Azure.Devices.Client/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -454,26 +454,38 @@ public void OnConnected()
...
+                    else
+                    {
+                        this.messageQueue.Enqueue(message);
+                        this.receivingSemaphore.Release();
+                    }
                 }
             }
+            catch(Exception ex)
+            {
+                // This can still throw a fatal exception
```
CIPop: I'm not sure what you mean by this comment: "which defeats the reason ..." 
Applications should never catch exceptions that are fatal such as OutOfMemory or AccessViolation. In that case the application should immediately terminate, which is what is happening today. 

That said, the recommendation is that all expected exceptions (network, timeout) are caught in the DotNetty callbacks and properly handled (in this case by the OnError call which will, in turn, correctly switch the state machine and report back to the initial application call.

Please catch only exceptions that are not fatal so that the stack is preserved correctly. See https://github.com/Azure/azure-iot-sdk-csharp/pull/327/files#diff-95d848600b3ef1d9923119856d544fe5R538 on how to do

oldblindman:
Cristian,

Done!

Gene

CIPop:
Thanks @oldblindman. I think you didn't push but since you have allowed contributors I was able to rebase. I have scheduled this PR through CI.

PR #345, opened 2018-01-20
Adding support for PFX collections for the Provisioning X509 sample

CIPop:
PFX files (PKCS12) can contain a collection of certificates with or without private keys. This change allows usage of PFX files containing a single certificate with private key and multiple certificates without (i.e. the chain).
Apparently, the tools at https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/CACertificateOverview.md will produce a PFX containing both the root certificate as well as the leaf+private key.

### DIFF
--- a/provisioning/device/samples/ProvisioningDeviceClientX509/Program.cs
+++ b/provisioning/device/samples/ProvisioningDeviceClientX509/Program.cs
@@ -70,10 +70,36 @@ namespace ProvisioningDeviceClientX509
             s_idScope = args[0];
 
             string certificatePassword = ReadCertificatePassword();
-            using (var certificate = new X509Certificate2(s_certificateFileName, certificatePassword))
+
+            var certificateCollection = new X509Certificate2Collection();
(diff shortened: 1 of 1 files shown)


provisioning/device/samples/ProvisioningDeviceClientX509/Program.cs
```diff
@@ -70,10 +70,36 @@ public static void Main(string[] args)
...
+
+            X509Certificate2 certificate = null;
+
+            foreach (X509Certificate2 element in certificateCollection)
             {
-                RunSample(certificate).GetAwaiter().GetResult();
+                Console.WriteLine($"Found certificate: {element?.Thumbprint} {element?.Subject}; PrivateKey: {element?.HasPrivateKey}");
+                if (certificate == null && element.HasPrivateKey)
+                {
+                    certificate = element;
```
jspaith: Consider a "break;" here, since once you get one match there's no need continuing to enumerate.

CIPop: @jspaith The reason the code isn't breaking is to ensure all other X509 certs are disposed in the loop.

PR #348, opened 2018-01-23
Bump package versions for release 2018-1-23

jasmineymlo:


provisioning/device/src/Microsoft.Azure.Devices.Provisioning.Client.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionPrefix>1.0.1</VersionPrefix>
+    <VersionPrefix>1.0.2</VersionPrefix>
```
CIPop: Were there any ProvisioningClient versions? Do we need to release this?

security/tpm/src/Microsoft.Azure.Devices.Provisioning.Security.Tpm.csproj
```diff
@@ -26,7 +26,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionPrefix>1.0.1</VersionPrefix>
+    <VersionPrefix>1.0.2</VersionPrefix>
```
CIPop: Same questions as above - why do we need to release Transports and Security? Was there a fix?

provisioning/device/src/Microsoft.Azure.Devices.Provisioning.Client.csproj
jasmineymlo: Removing unnecessary CodeAnalysis.FxCopAnalyzers IOperation declaration changed the csproj.  I didn't look at every change in details.  There were some changes in "shared" in other PR.  If you are sure they do not need to be updated/release, we could revert.

security/tpm/src/Microsoft.Azure.Devices.Provisioning.Security.Tpm.csproj
jasmineymlo: same as above.

provisioning/device/src/Microsoft.Azure.Devices.Provisioning.Client.csproj
CIPop: Discussed offline: OK to leave as is. The FXCop change should not affect the behavior but there were changes to shared that might have been functional.

PR #357, opened 2018-02-01
Update TokenRefresh loop wait period

### LINKED ISSUE #319: AMQP TokenRefresh will randomly fail to re-initiate the CBS link
.Net Core
OS:  Ubuntu 16.04.1 or Windows 10.0.16299.192
SDK: 1.6.1

After ~220 minutes calling a registered method I get the error "Not Registered" on Ubuntu or Windows. I am calling this method every 15 minutes but after ~220 minutes I get this error.

Using the following code:

using System;
using System.Text;
using System.Threading;
using System.Threading.Tasks;
namespace Linux
{
    class Program
    {
        static DeviceClient mydevice;
        static String connectionString = "HostName=kevinsayIoT.azure-devices.net;DeviceId=Linux;SharedAccessKey=7R2clREMOVEDDTI=";
        static void Main(string[] args)
        {
            mydevice = DeviceClient.CreateFromConnectionString(connectionString, TransportType.Amqp);
            mydevice.SetMethodHandlerAsync("testMethod", testMethodCallback, null);
            mydevice.OpenAsync();
            int minute = 0;
            while (true)
            {
                // Sleep a minute
                Thread.Sleep(1000 * 60);
                minute++;
                Console.WriteLine("Slept for: " + minute.ToString() + " minutes");
            }
        }
        private static Task<MethodResponse> testMethodCallback(MethodRequest methodRequest, object userContext)
        {
            Console.WriteLine("Method called: " + methodRequest.DataAsJson);
            return Task.FromResult(new MethodResponse(Encoding.UTF8.GetBytes("{\"status\":\"Success\"}"), 200));
        }
    }
}

I get the following error when I call the method from Device Explorer after ~220 minutes:

One or more errors occurred. (Device {"Message":"{\"errorCode\":404103,\"trackingId\":\"d8248f029def4348b4f51915c766fdf4-G:11-TimeStamp:01/02/2018 01:01:17\",\"message\":\"Timed out waiting for device to connect.\",\"info\":{\"timeout\":\"00:00:10\"},\"timestampUtc\":\"2018-01-02T01:01:17.9205026Z\"}","ExceptionMessage":""} not registered)

varunpuranik:
# Checklist

# Reference/Link to the issue solved with this PR (if any) - https://github.com/Azure/azure-iot-sdk-csharp/issues/319

# Description of the problem
IoTHubConnectionString.GetTokenAsync sets the expiresOn value to RefreshesOn (instead of the actual expiresOn value for the token). This causes IoTHubTokenRefresher loop to terminate at times. When that happens, a new token is never sent on the CBS link, and the link dies.  

# Description of the solution
This change sets the expiresOn value to the actual expiry time for the token. This way the loop in IoTHubTokenRefresher does not terminate, and a new token keeps getting sent as expected. 

Fixes #319

### DIFF
--- a/device/Microsoft.Azure.Devices.Client/IotHubConnectionString.Core.cs
+++ b/device/Microsoft.Azure.Devices.Client/IotHubConnectionString.Core.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.Devices.Client
             else
             {
                 tokenValue = await this.TokenRefresher.GetTokenAsync(this.Audience).ConfigureAwait(false);
-                expiresOn = this.TokenRefresher.RefreshesOn;
+                expiresOn = this.TokenRefresher.ExpiresOn;
             }
 
             return new CbsToken(tokenValue, CbsConstants.IotHubSasTokenType, expiresOn);
--- a/device/Microsoft.Azure.Devices.Client/IotHubTokenRefresher.cs
+++ b/device/Microsoft.Azure.Devices.Client/IotHubTokenRefresher.cs
@@ -15,6 +15,8 @@ namespace Microsoft.Azure.Devices.Client
         private static readonly string[] AccessRightsStringArray = 
             AccessRightsHelper.AccessRightsToStringArray(AccessRights.DeviceConnect);
 
+        static readonly TimeSpan BufferPeriod = TimeSpan.FromSeconds(120);
+
         private readonly AmqpSession amqpSession;
         private readonly IotHubConnectionString connectionString;
         private readonly string audience;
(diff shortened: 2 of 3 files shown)


CIPop:
Thanks @varunpuranik ! As we've talked offline I'd like to properly fix this in the future so that the AMQP CBS token refresher uses the TokenRefresher object. The current fix will ignore the policy set by DeviceAuthenticationWithToken but fixes the bigger issue of the token not getting refreshed.

That would require some larger refactoring and so we can leave this code in place for now.

CIPop:
@varunpuranik please squash (`git rebase -i master`) to a single commit before merging.

PR #358, opened 2018-02-01
updated to compile against .NET Framework 4.5.2

yorek:
Updated to make sure it compiles with .NET Framework 4.5.2 instead of 4.5.1


CIPop:
Thanks @yorek for the proposal. We can't approve this because it is a breaking change - we can only remove support for 4.5.1 when we switch the major version.
Applications targetting .NET 4.5.2 and above should still be able to reference this library and all supported .NET Framework runtimes (4.5.2+) would be binary compatible with the 4.5.1 target.

PR #359, opened 2018-02-02
Update TokenRefresh loop wait period

varunpuranik:
# Checklist

# Reference/Link to the issue solved with this PR (if any) - https://github.com/Azure/azure-iot-sdk-csharp/issues/319

# Description of the problem
IoTHubConnectionString.GetTokenAsync sets the expiresOn value to RefreshesOn (instead of the actual expiresOn value for the token). This causes IoTHubTokenRefresher loop to terminate at times. When that happens, a new token is never sent on the CBS link, and the link dies.

# Description of the solution
This change sets the expiresOn value to the actual expiry time for the token. This way the loop in IoTHubTokenRefresher does not terminate, and a new token keeps getting sent as expected.

Fixes https://github.com/Azure/azure-iot-sdk-csharp/issues/319

### DIFF
--- a/device/Microsoft.Azure.Devices.Client/IotHubConnectionString.Core.cs
+++ b/device/Microsoft.Azure.Devices.Client/IotHubConnectionString.Core.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.Devices.Client
             else
             {
                 tokenValue = await this.TokenRefresher.GetTokenAsync(this.Audience);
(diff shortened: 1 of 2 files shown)


CIPop:
Will this follow #367?

varunpuranik:
No, I am actually going to abandon it. Initially we wanted to make the same change in modules-preview branch, but now we are FI'ing master to modules-preview instead.

PR #367, opened 2018-02-08
Merge branch 'master' into modules-preview

damonbarry:
Forward-integrate all the latest changes to the SDK from the master branch into modules-preview. Note that normally I'd merge changes from the last official release, but we specifically need a commit that went in after the release. I also confirmed that any other post-release commits in this merge shouldn't impact modules features.


varunpuranik:
Looks good.

PR #374, opened 2018-02-15
Update readme.md

nberdy:
Fixing the cert chain code

 

# Checklist

# Description of the problem
The existing recommended code for using a cert chain when simulating devices doesn't work.

# Description of the solution
I played around with the recommended code and landed on the changes in this PR as the fix.

### DIFF
--- a/provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
+++ b/provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
@@ -42,10 +42,16 @@ to
 ```C# 
     var myCertificate = new X509Certificate2("myCertificate.pfx", "mypassword");
     var myChain = new X509Certificate2Collection();
-    var security = new SecurityProviderX509Certificate(myCertificate, myChain);
+    
+    // Comment out the below line if you do not have a .p7b file (e.g. if you generated certificates using the tool below)
+    myChain.Import("myChain.p7b");
+    
+    using (var security = new SecurityProviderX509Certificate(myCertificate, myChain))
 ```
 
 A tool for creating _test_ certificates is available at https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/CACertificateOverview.md
+
+If you generate _test_ certificates as an administrator using the above tool, please note you must run the sample as administrator as well.
 
 If a Windows compatible Hardware Security Module is used, the certificate must be obtained by opening it from the Certificate Store using [X509Store](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509store(v=vs.110).aspx).
 
--- a/provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
+++ b/provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
@@ -42,8 +42,7 @@ to
 ```C# 
     var myCertificate = new X509Certificate2("myCertificate.pfx", "mypassword");
     var myChain = new X509Certificate2Collection();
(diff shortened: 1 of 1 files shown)


provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -42,8 +42,7 @@ to
 ```C# 
     var myCertificate = new X509Certificate2("myCertificate.pfx", "mypassword");
     var myChain = new X509Certificate2Collection();
-    myChain.Import("myChain.p7b");
```
CIPop: @nberdy any reason this didn't work for you? What was the error?

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
```diff
@@ -42,8 +42,7 @@ to
...
-    using (var security = new SecurityProviderX509Certificate(myCertificate, myChain)) 
+    var security = new SecurityProviderX509Certificate(myCertificate, myChain);
```
CIPop: 1. It's important to Dispose of this object (why the `using` statement was originally there).
2. `myChain` is empty after this change. This is fine as long as the certificates are correctly installed in your Intermediate Authorities store.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
nberdy: There's no .p7b file with that name, and I never create one in any of the steps (nor am I instructed to do so). As far as I can tell, there's no need for this line.

nberdy: (Nor do I know how to debug why I have no .p7b file or where to get one)

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
nberdy: 1. Noob question: Why is it important to Dispose the object?
2. How do I check that the certs are correctly installed?

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
CIPop: Good point regarding the temporary scripts. 
We should create a p7b or comment this line saying that if a collection of intermediate certificates is available we provide the functionality to import them in the correct store.

provisioning/device/samples/ProvisioningDeviceClientX509/readme.md
CIPop: > Why is it important to Dispose the object?

Disposing ensures that resources are de-allocated immediately. Otherwise GC will collect later or this will be performed at process exit.

> How do I check that the certs are correctly installed?

Verify the Intermediate Certificate Authorities store using certmgr.msc in Windows. In Linux, .NET Core provides a similar store construct that can only be accessed via .NET APIs.
image

CIPop:
@nberdy - any progress on this? 
To summarize my comments:
1. A p7b may be required if a customer did not previously add the intermediate authorities in the correct store. We provide a convenience method to add them but it's definitely not mandatory.
2. The `using` statement is always recommended when operating on `IDisposable` objects to ensure the native resources are released as soon as they aren't needed by the application. (Apps can also manually call `Dispose` but the `using` statement will automatically ensure that will happen even in the case of an exception getting thrown. - See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement for more details.)

PR #379, opened 2018-02-21
Remove the Uri encoding for query

### LINKED ISSUE #377: Provisioning Service - CreateEnrollmentGroupQuery always returns 0 results
- **OS and version used:** Windows 10

- **SDK version used:** Microsoft.Azure.Devices.Provisioning.Service 1.0.1

# Description of the issue:
RegistrationStatusManager.CreateEnrollmentGroupQuery() always returns 0 members. Tracked the issue down to the following -

1) RegistrationStatusManager class calls GetGetDeviceRegistrationStatus() to build a path that looks something like: registrations/groupName
2) Query class calls GetQueryUri() which calls WebUtility.UrlEncode(path); and changes the path to:

registrations%2FgroupName

This causes the lookup to fail.

mounicaArroju:
# Description of the problem
There was a bug in Query Uri with a '\' being escaped. This was returning a 404 for CreateEnrollmentGroupRegistrationStateQuery Api. This is a bug reported [here](https://github.com/Azure/azure-iot-sdk-csharp/issues/377)

# Description of the solution
Removed the Uri encoding in the query

Fixes #377

### DIFF
--- a/provisioning/service/src/Query.cs
+++ b/provisioning/service/src/Query.cs
@@ -257,7 +257,6 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
 
         private static Uri GetQueryUri(string path)
         {
-            path = WebUtility.UrlEncode(path);
             return new Uri(QueryUriFormat.FormatInvariant(path, SDKUtils.ApiVersionQueryString), UriKind.Relative);
         }
     }


CIPop:
@mounicaArroju could you please add a Unit Test as well?

PR #387, opened 2018-03-01
Fix bugs when enable e2e diagnostic with HTTP protocol

SLdragon:
# Checklist
- [ ] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-c/blob/master/.github/CONTRIBUTING.md).

# Description of the problem

# Description of the solution

### DIFF
--- a/device/Microsoft.Azure.Devices.Client/DeviceClient.cs
+++ b/device/Microsoft.Azure.Devices.Client/DeviceClient.cs
@@ -125,9 +125,14 @@ TODO: revisit DefaultDelegatingHandler - it seems redundant as long as we have t
                         "The range of diagnostic sampling percentage should between [0,100].");
                 }
 
-                _diagnosticSamplingPercentage = value;
+                if (IsE2EDiagnosticSupportedProtocol())
+                {
+                    _diagnosticSamplingPercentage = value;
+                }
             }
         }
+
+        ITransportSettings[] transportSettings;
 
 #if !WINDOWS_UWP && !PCL
         internal X509Certificate2 Certificate { get; set; }
@@ -265,6 +270,7 @@ TODO: revisit DefaultDelegatingHandler - it seems redundant as long as we have t
             IDelegatingHandler innerHandler = pipelineBuilder.Build(pipelineContext);
 
             this.InnerHandler = innerHandler;
+            this.transportSettings = transportSettings;
         }
 
         DeviceClient(IotHubConnectionString iotHubConnectionString)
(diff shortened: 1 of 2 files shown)


erich-wang:
@CIPop the issue is that we will fail to send telemetry message if setting **DeviceClient.DiagnosticSamplingPercentage = 100** with Http protocol.
the following code snippet describes the issue
``` csharp
            var client = DeviceClient.CreateFromConnectionString(deviceConnectionString, TransportType.Http1);
            client.DiagnosticSamplingPercentage = 100;
            var message = new Message(Encoding.ASCII.GetBytes("hello"));
            client.SendEventAsync(message).Wait(); //get exception here
```
The root cause is diagnostic related system properties are added into message but underlying http protocol handler could not recognize these properties.
The fix is to not allow set DiagnosticSamplingPercentage for Http protocol to align with e2e diagnostic design.

device/Microsoft.Azure.Devices.Client/DeviceClient.cs
```diff
@@ -1463,6 +1469,21 @@ internal void OnReportedStatePatchReceived(TwinCollection patch)
                 await ApplyTimeout(operationTimeoutCancellationToken => this.InnerHandler.DisableMethodsAsync(operationTimeoutCancellationToken)).ConfigureAwait(false);
             }
         }
+
+        private bool IsE2EDiagnosticSupportedProtocol()
+        {
+            var transportSetting = this.transportSettings.FirstOrDefault();
```
CIPop: This is looking at only the first item in the list - what about the fallback case where multiple protocol configurations are specified like {Amqp, Mqtt, Http}. If for some reason both Amqp and Mqtt fail, the code will fail when Diagnostics are attempted with Http.

device/tests/Microsoft.Azure.Devices.Client.Test/DeviceClientTests.cs
```diff
@@ -75,6 +75,22 @@ public void DeviceClient_SetDiagnosticSamplingPercentageOutOfRange_Fail()
             }
         }
 
+        [TestMethod]
+        [TestCategory("IoTHubClientDiagnostic")]
+        public void DeviceClient_StartDiagLocallyThatDoNotSupport_ThrowException()
+        {
+            DeviceClient deviceClient = DeviceClient.CreateFromConnectionString(fakeConnectionString, TransportType.Http1);
```
CIPop: Please add another test where multiple transport types are used: transport choices w/o Http1 should work while those containing Http1 anywhere in the list should fail.

device/Microsoft.Azure.Devices.Client/DeviceClient.cs
SLdragon: Thanks for your suggestions, we are missing that case, we will fix it.

device/tests/Microsoft.Azure.Devices.Client.Test/DeviceClientTests.cs
SLdragon: fixed

PR #399, opened 2018-03-13
Ability for a device to authenticate with registry using token

dsajanice:
IoTHub service supports devices connecting to registry manager with a device scoped key for module operations. This change allows devices to also connect to registry manager for the same set of module operations with a token. This is used when the key is obtained on the device by the device provisioning flow and a token is generated using the key.

### DIFF
--- a/common/src/service/AuthenticationMethodFactory.cs
+++ b/common/src/service/AuthenticationMethodFactory.cs
@@ -15,7 +15,14 @@ namespace Microsoft.Azure.Devices
         {
             if (!string.IsNullOrWhiteSpace(iotHubConnectionStringBuilder.DeviceId))
             {
-                return new ServiceAuthenticationWithDeviceCredentials(iotHubConnectionStringBuilder.DeviceId, iotHubConnectionStringBuilder.SharedAccessKey);
+                if (!string.IsNullOrWhiteSpace(iotHubConnectionStringBuilder.SharedAccessKey))
+                {
+                    return new ServiceAuthenticationWithDeviceSharedAccessPolicyKey(iotHubConnectionStringBuilder.DeviceId, iotHubConnectionStringBuilder.SharedAccessKey);
+                }
+                if (!string.IsNullOrWhiteSpace(iotHubConnectionStringBuilder.SharedAccessSignature))
+                {
+                    return new ServiceAuthenticationWithDeviceSharedAccessPolicyToken(iotHubConnectionStringBuilder.DeviceId, iotHubConnectionStringBuilder.SharedAccessSignature);
+                }
             }
             if (string.IsNullOrWhiteSpace(iotHubConnectionStringBuilder.SharedAccessKey))
             {
--- a/service/Microsoft.Azure.Devices/Microsoft.Azure.Devices.csproj
+++ b/service/Microsoft.Azure.Devices/Microsoft.Azure.Devices.csproj
@@ -289,7 +289,8 @@
     <Compile Include="Receiver.cs" />
     <Compile Include="RegistryManager.cs" />
     <Compile Include="RegistryStatistics.cs" />
-    <Compile Include="ServiceAuthenticationWithDeviceCredentials.cs" />
+    <Compile Include="ServiceAuthenticationWithDeviceSharedAccessPolicyToken.cs" />
+    <Compile Include="ServiceAuthenticationWithDeviceSharedAccessPolicyKey.cs" />
     <Compile Include="ServiceAuthenticationWithSharedAccessPolicyKey.cs" />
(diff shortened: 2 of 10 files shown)


common/src/service/AuthenticationMethodFactory.cs
```diff
@@ -15,7 +15,7 @@ static internal IAuthenticationMethod GetAuthenticationMethod(IotHubConnectionSt
         {
             if (!string.IsNullOrWhiteSpace(iotHubConnectionStringBuilder.DeviceId))
             {
-                return new ServiceAuthenticationWithDeviceCredentials(iotHubConnectionStringBuilder.DeviceId, iotHubConnectionStringBuilder.SharedAccessKey);
+                return new ServiceAuthenticationWithDeviceCredentials(iotHubConnectionStringBuilder.DeviceId, iotHubConnectionStringBuilder.SharedAccessKey, iotHubConnectionStringBuilder.SharedAccessSignature);
```
CIPop: This doesn't really make sense: you already have the Shared Access Key which is used to generate a signature (SAS). 
Why would we put both in this class ctor?

service/Microsoft.Azure.Devices/ServiceAuthenticationWithDeviceCredentials.cs
```diff
@@ -22,7 +23,7 @@ public IotHubConnectionStringBuilder Populate(IotHubConnectionStringBuilder iotH
 
             iotHubConnectionStringBuilder.SharedAccessKey = this.Key;
             iotHubConnectionStringBuilder.DeviceId = this.DeviceId;
-            iotHubConnectionStringBuilder.SharedAccessSignature = null;
+            iotHubConnectionStringBuilder.SharedAccessSignature = this.Token;
```
CIPop: If you have a key you can't have a token and vice-versa.

dsajanice:
@CIPop I have separated the token auth from key auth by creating a new class for it. Creating a different constructor would not have worked because the signature of the constructors would have been the same. Please take a look. Thanks.

service/Microsoft.Azure.Devices/ServiceAuthenticationWithDeviceSharedAccessPolicyToken.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public string DeviceId { get; set; }
+
+        public string Token { get; set; }
+    }
+}
```
CIPop: Nit: missing NL at EOF.

service/Microsoft.Azure.Devices/ServiceAuthenticationWithDeviceSharedAccessPolicyToken.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            return iotHubConnectionStringBuilder;
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public string DeviceId { get; set; }
+
+        public string Token { get; set; }
```
CIPop: Nit: this and above: missing documentation.

service/Microsoft.Azure.Devices/ServiceAuthenticationWithDeviceSharedAccessPolicyToken.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Azure.Devices
+{
+    public class ServiceAuthenticationWithDeviceSharedAccessPolicyToken : IAuthenticationMethod
```
CIPop: Nit: Public classes must have documentation - this would add yet a new warning in the build system.

service/Microsoft.Azure.Devices/Microsoft.Azure.Devices.csproj
```diff
@@ -289,7 +289,8 @@
     <Compile Include="Receiver.cs" />
     <Compile Include="RegistryManager.cs" />
     <Compile Include="RegistryStatistics.cs" />
-    <Compile Include="ServiceAuthenticationWithDeviceCredentials.cs" />
+    <Compile Include="ServiceAuthenticationWithDeviceSharedAccessPolicyToken.cs" />
```
CIPop: Note for -preview adopters: This is a public API breaking change between -preview packages. There is no expectation that API compat is maintained but existing applications may be impacted.

Workaround for apps using `ServiceAuthenticationWithDeviceCredentials` is to change to `ServiceAuthenticationWithDeviceSharedAccessPolicyKey`.

PR #401, opened 2018-03-13
Adding package testing and increment the versions after LTS

CIPop:
- Signing is controlled via AZURE_IOT_DONOTSIGN (@alextolp FYI)
- Binary and package signing integration
- Package testing is controlled via the AZURE_IOT_LOCALPACKAGES (nuget packages locally built will be automatically tested in the E2E and Samples builds)


CIPop:
Thanks @zolvarga !

PR #405, opened 2018-03-16
Add support for receiving events for modules over AMQP

varunpuranik:
# Checklist
- [x] This pull-request is submitted against the `modules-preview` branch, since this change is modules specific

# Description of the changes
- Added support for receiving events for Modules over AMQP. Currently this support exists only for MQTT.


device/Microsoft.Azure.Devices.Client/IotHubConnection.cs
```diff
@@ -50,7 +50,8 @@ public enum SendingLinkType {
         public enum ReceivingLinkType {
             C2DMessages,
             Methods,
-            Twin
+            Twin,
+            Messages
```
damonbarry: `Events` instead of `Messages`?

PR #417, opened 2018-03-20
Bump version of shared.NetStandard

dsajanice:
Looks like there is a bug in the bump_version script against modules-preview branch. It missed updating the assembly version of shared.NetStandard. But this might not be a problem after we FI with master.


dsajanice:
@varun please take a look

CIPop:
@dsajanice FYI - I've updated the version bump script last week and tested it on LTS. 
master has a new version bump system that is part of the normal check-in process.

CIPop:
Already addressed in #429.

PR #426, opened 2018-03-26
Merge branch 'master' into modules-preview

damonbarry:


damonbarry:
@CIPop I created another PR on my fork that is the same changes, but against master instead of modules-preview so that it's easier for my team to review (the diff shows all the changes that make modules-preview different from the latest master, which is only 58 files). 👍

PR #442, opened 2018-04-02
Bump package versions for release 2018-04-02

abhipsaMisra:
# Description of the changes

Bumping the package versions for release 2018-04-02

### DIFF
--- a/e2e/test/TwinE2ETests.cs
+++ b/e2e/test/TwinE2ETests.cs
@@ -391,6 +391,8 @@ namespace Microsoft.Azure.Devices.E2ETests
             await Twin_DeviceSetsReportedPropertyAndServiceReceivesIt(Client.TransportType.Amqp_WebSocket_Only);
         }
 
+        // TODO: #243 : Once the service team deploys the fix for Issue #243, the test can be executed
+        [Ignore]
         [TestMethod]
         [TestCategory("Twin-E2E")]
         public async Task Twin_ServiceDoesNotCreateNullPropertyInCollection_Mqtt()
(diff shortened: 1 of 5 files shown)


e2e/test/TwinE2ETests.cs
```diff
@@ -750,10 +750,7 @@ static public void ClassCleanup()
                 Assert.IsTrue(serviceTwin.Properties.Reported.Contains(propName1));
                 String value1 = serviceTwin.Properties.Reported[propName1].ToString();
 
-                // TODO: #243 
-                // If service team fixed the Issue #243 the following Assert.AreNotEqual should fail 
-                // and need to be changed to Assert.AreEqual
-                Assert.AreNotEqual(value1, propEmptyValue);
+                Assert.AreEqual(value1, propEmptyValue);
```
CIPop: @abhipsaMisra @zolvarga The tests are going to be in an odd state as our service gets upgraded.
My suggestion is to disable the tests against a GitHub issue (maybe the original one) until things get stable. At that point we can simply enable the test validating the new service behavior and close the bug.

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -22,7 +22,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.7.0</Version>
+    <Version>1.7.1</Version>
```
CIPop: What changes do we have to IoTHub DeviceClient?

versions.csv
```diff
@@ -1,10 +1,10 @@
 AssemblyPath, Version
-iothub\device\src\Microsoft.Azure.Devices.Client.csproj, 1.7.0
+iothub\device\src\Microsoft.Azure.Devices.Client.csproj, 1.7.1
```
CIPop: See above question.

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
abhipsaMisra: The following change was made to the ProductInfo class: Update ProductInfo to use semver-compliant AssemblyInformationalVersion

versions.csv
abhipsaMisra: same as above

PR #458, opened 2018-04-20
Added gatewayHostname to Create device client with authentication method

ancaantochi:
# Description of the changes
Added overloads for DeviceClient.Create to accept gatewayHostname


varunpuranik:
Changes look good. However, GatewayHostName has hitherto been a somewhat hidden feature of the Device SDK, parsed via the connection string, without any real API visibility. This changes that...
While I am okay with it, we should once check with the Device SDK team as well. 
Please pull me in on that.

PR #461, opened 2018-04-20
Bump version of device and service

ancaantochi:
Version update for Client and Service

### DIFF
--- a/iothub/service/src/Microsoft.Azure.Devices.csproj
+++ b/iothub/service/src/Microsoft.Azure.Devices.csproj
@@ -22,7 +22,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -22,7 +22,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.16.0-preview-003</Version>
+    <Version>1.16.0-preview-004</Version>
```
varunpuranik: Why do we need to update the Service SDK Version?

ancaantochi: There were changes to the service done in master that were merged to modules-preview

PR #462, opened 2018-04-27
Add HSM HTTP authentication and DeviceClientFactory

ancaantochi:
# Description of the changes
- Added HSM Http Authentication for modules
- Added DeviceClientFactory to create DeviceClient based on IoT Edge environment variables


iothub/device/src/Edge/DeviceClientFactory.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// Initializes a new instance of the <see cref="DeviceClientFactory"/> class.
+        /// </summary>
+        public DeviceClientFactory()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DeviceClientFactory"/> class with transport type.
+        /// </summary>
+        /// <param name="transporType">Specifies whether Amqp, MQTT or Http transport is used.</param>
```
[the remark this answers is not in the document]
ancaantochi: thanks

iothub/device/src/Edge/DeviceClientFactory.cs
```diff
@@ -0,0 +1,138 @@
...
+                string moduleId = this.GetValueFromEnvironment(envVariables, ModuleIdVariableName) ?? throw new InvalidOperationException($"Environement variable {ModuleIdVariableName} is required.");
+                string hostname = this.GetValueFromEnvironment(envVariables, IotHubHostnameVariableName) ?? throw new InvalidOperationException($"Environement variable {IotHubHostnameVariableName} is required.");
+                string authScheme = this.GetValueFromEnvironment(envVariables, AuthSchemeVariableName) ?? throw new InvalidOperationException($"Environement variable {AuthSchemeVariableName} is required.");
+                string gateway = this.GetValueFromEnvironment(envVariables, GatewayHostnameVariableName);
+                string apiVersion = this.GetValueFromEnvironment(envVariables, IotEdgedApiVersionVariableName);
+
+                if (!string.Equals(authScheme, SasTokenAuthScheme, StringComparison.OrdinalIgnoreCase))
+                {
+                    throw new InvalidOperationException($"Unsupported authentication scheme. Supported scheme is {SasTokenAuthScheme}.");
```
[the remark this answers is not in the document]
ancaantochi: SAS token is the only supported auth scheme for now when using edged. If connection string is set in the env variable it will use that to connect to Edge Hub/IoT Hub

iothub/device/src/HsmAuthentication/ModuleAuthenticationWithHsm.cs
```diff
@@ -0,0 +1,45 @@
...
+            this.signatureProvider = signatureProvider ?? throw new ArgumentNullException(nameof(signatureProvider)); 
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="iotHub"></param>
+        /// <param name="suggestedTimeToLive"></param>
+        /// <returns></returns>
+        protected override async Task<string> SafeCreateNewToken(string iotHub, int suggestedTimeToLive)
```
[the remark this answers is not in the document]
ancaantochi: it is iothub hostname, I added the description

iothub/device/src/Edge/DeviceClientFactory.cs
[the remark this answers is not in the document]
varunpuranik: Yes, SAS token is the only auth scheme supported by EdgeLet now.. and that is what this Factory is really giving us.

iothub/device/src/Edge/DeviceClientFactory.cs
```diff
@@ -0,0 +1,138 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+using Microsoft.Azure.Devices.Client.HsmAuthentication;
+
+namespace Microsoft.Azure.Devices.Client.Edge
```
varunpuranik: The code looks good. My concern with this would be with the API - now we have 2 "Factory" patterns - one on the DeviceClient itself (it has Create, CreateFromConnectionString), and then we also have this DeviceClientFactory that has a create method for a different purpose. 
The 2 should be unified. We should move the API methods in this file to DeviceClient which can then call this internal type to construct the DeviceClient.

ancaantochi: Because this factory is specific to Edge, I think would be better to keep it separate from DeviceClient Create methods. When we have a separate moduleSDK, we can move it there.

iothub/device/tests/HsmAuthentication/ModuleAuthenticationWithHsmTest.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Azure.Devices.Client.HsmAuthentication;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace Microsoft.Azure.Devices.Client.Test.HsmAuthentication
+{
+    [TestClass]
+    public class ModuleAuthenticationWithHsmTest
+    {
+        string signature = "signature";
```
zolvarga: >signature [](start = 15, length = 9)

From the coding standards:
"We use _camelCase for internal and private fields and use readonly where possible. Prefix internal and private instance fields with _, static fields with s_ and thread static fields with t_."

Please use underscore here

iothub/device/src/HsmAuthentication/HttpHsmSignatureProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+using System;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Client.TransientFaultHandling;
+
+namespace Microsoft.Azure.Devices.Client.HsmAuthentication
+{
+    class HttpHsmSignatureProvider : ISignatureProvider
+    {
+        const string DefaultApiVersion = "2018-06-28";
```
zolvarga: >DefaultApiVersion [](start = 21, length = 17)

Please take a look at the coding standards here:
[https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)

Specify visibility even if it is default.

iothub/device/src/HsmAuthentication/HttpHsmSignatureProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+        const SignRequestAlgo DefaultSignRequestAlgo = SignRequestAlgo.HMACSHA256;
+        readonly string apiVersion;
```
zolvarga: >apiVersion [](start = 24, length = 10)

Coding standards: Prefix is missing

iothub/device/src/Edge/DeviceClientFactory.cs
zolvarga: I agree with Varun.  Please move it to the DeviceCLient

---
In reply to: [184792231](https://github.com/Azure/azure-iot-sdk-csharp/pull/462#discussion_r184792231) [](ancestors = 184792231)

PR #465, opened 2018-05-02
DeviceClient / ModuleClient split

varunpuranik:
# Checklist

# Description of the changes
Split DeviceClient API into DeviceClient (for devices) and ModuleClient (for modules). Refactored shared code into a common base class. 
Also fixed the tests as necessary.

### DIFF
--- a/iothub/device/src/BaseClient.cs
+++ b/iothub/device/src/BaseClient.cs
@@ -7,7 +7,6 @@ namespace Microsoft.Azure.Devices.Client
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
-    using System.Linq;
 #if NETSTANDARD1_3
     using System.Net.Http;
 #endif
@@ -17,7 +16,6 @@ namespace Microsoft.Azure.Devices.Client
     using Microsoft.Azure.Devices.Client.Extensions;
     using Microsoft.Azure.Devices.Client.Transport;
     using Microsoft.Azure.Devices.Shared;
-    using Newtonsoft.Json.Linq;
     using Microsoft.Azure.Devices.Client.Transport.Mqtt;
     using System.Security.Cryptography.X509Certificates;
 
@@ -504,7 +502,11 @@ TODO: revisit DefaultDelegatingHandler - it seems redundant as long as we have t
             });
         }
 
-        internal Task CompleteInternalAsync(string lockToken)
+        /// <summary>
+        /// Deletes a received message from the device queue
+        /// </summary>
+        /// <returns>The lock identifier for the previously received message</returns>
+        public Task CompleteAsync(string lockToken)
         {
             if (string.IsNullOrEmpty(lockToken))
             {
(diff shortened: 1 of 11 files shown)


iothub/device/src/BaseClient.cs
```diff
@@ -0,0 +1,1083 @@
...
+                                                                                       |  everything |
+                                                                                       |             |
+                                                                                       +-------------+
+TODO: revisit DefaultDelegatingHandler - it seems redundant as long as we have to many overloads in most of the classes.
+*/
+
+    /// <summary>
+    /// Contains methods that a device can use to send messages to and receive from the service.
+    /// </summary>
+    public class BaseClient : IDisposable
```
BertKleewein: >public [](start = 4, length = 6)

Should this be internal?  (assuming the language allows a public object to inherit from an internal base).  I'm mostly concerned that making this public means that we need to worry about breaking changes at this level.

varunpuranik: Yeah, unfortunately a public class cannot derive from an internal class.

iothub/device/src/ModuleClient.cs
```diff
@@ -0,0 +1,511 @@
...
+            this.InnerHandler = innerHandler;
+        }
+
+        /// <summary>
+        /// Create an Amqp ModuleClient from individual parameters
+        /// </summary>
+        /// <param name="hostname">The fully-qualified DNS hostname of IoT Hub</param>
+        /// <param name="authenticationMethod">The authentication method that is used</param>
+        /// <returns>ModuleClient</returns>
+        public static ModuleClient Create(string hostname, IAuthenticationMethod authenticationMethod)
```
tameraw: Can we pull in all the common code for the Create...() functions into the Base Class as well?

iothub/device/src/ModuleClient.cs
```diff
@@ -0,0 +1,511 @@
...
+            return ApplyTimeout(operationTimeoutCancellationToken => this.InnerHandler.SendEventAsync(message, operationTimeoutCancellationToken));
+        }
+
+        /// <summary>
+        /// Sends a batch of events to device hub
+        /// <param name="outputName">The output target for sending the given message</param>
+        /// <param name="messages">A list of one or more messages to send</param>
+        /// </summary>
+        /// <returns>The task containing the event</returns>
+        public Task SendEventBatchAsync(string outputName, IEnumerable<Message> messages)
```
tameraw: >string outputName, IEnumerable<Message> messages [](start = 40, length = 48)

Consider swapping out the parameters to have "Message" first then Output. Here and elsewhere.

iothub/device/src/ModuleClient.cs
varunpuranik: I couldn't because it finally calls ModuleClient.Create or DeviceClient.Create, and pulling it into Base would require exposing some methods. 
But I am planning a further refactor of this code... will have that iteration out by tonight.

iothub/device/src/BaseClient.cs
pierreca: here's why: [https://ericlippert.com/2012/11/13/why-is-deriving-a-public-class-from-an-internal-class-illegal/](https://ericlippert.com/2012/11/13/why-is-deriving-a-public-class-from-an-internal-class-illegal/)

have you looked at using composition instead? is that too much overhead?

---
In reply to: [185608658](https://github.com/Azure/azure-iot-sdk-csharp/pull/465#discussion_r185608658) [](ancestors = 185608658)

varunpuranik: Yes, that is the plan.. I mentioned to Tamer and Bert as well. I will send an update with composition by tomorrow.

iothub/device/src/ModuleClient.cs
varunpuranik: Refactored these into a separate ClientFactory class.

iothub/device/src/BaseClient.cs
varunpuranik: Changed the code to make it internal.

iothub/device/src/ModuleClient.cs
varunpuranik: Emm, this was how it was designed originally (by Olivier, etc.), and I am afraid this will have far reaching impact on current module developers. I am inclined to keep it as it is.

iothub/device/src/DeviceClient.cs
```diff
@@ -3,336 +3,33 @@
...
-                throw new NotSupportedException();
-            }
-
-            retryDelegatingHandler.SetRetryPolicy(retryPolicy);
-        }
-
+        private readonly InternalClient internalClient;
 
-        private T GetDelegateHandler<T>() where T: DefaultDelegatingHandler
+        DeviceClient(InternalClient internalClient)
```
tameraw: >DeviceClient [](start = 8, length = 12)

nitpick: Explicit private please

iothub/device/src/ModuleClient.cs
```diff
@@ -0,0 +1,363 @@
...
+    using Microsoft.Azure.Devices.Shared;
+
+    /// <summary>
+    /// Contains methods that a device can use to send messages to and receive from the service.
+    /// </summary>
+    public sealed class ModuleClient : IDisposable
+    {
+        private readonly InternalClient internalClient;
+
+        internal ModuleClient(InternalClient internalClient)
```
tameraw: >internal [](start = 8, length = 8)

private to stay consistent with DeviceClient

varunpuranik: Emm, this is internal by design. It is used by the EdgeModuleClientFactory to instantiate the module client.

PR #468, opened 2018-05-03
Adding '?' prefix to MQTT username property bag and updating the clie…

rafilho:
…nt API version.

# Checklist

This pull request is against the module-preview branch, since the fix is only required when modules support is available.

# Description of the changes
- Adding '?' prefix to MQTT username property bag, indicating the beginning of the property bag and making easier to identify if the MQTT username is referring to a device or module, by the # of segments before the property bag;
- Updating client API version to the edge preview version, since the service will only accept property bags under this new format starting on the edge preview version ("2017-11-08-preview");

### DIFF
--- a/iothub/device/src/Common/Api/ClientApiVersionHelper.cs
+++ b/iothub/device/src/Common/Api/ClientApiVersionHelper.cs
@@ -8,7 +8,15 @@ namespace Microsoft.Azure.Devices.Client
         const string ApiVersionQueryPrefix = "api-version=";
         const string ApiVersionNov2016 = "2016-11-14";
         const string ApiVersionJune2017 = "2017-06-30";
-        public const string ApiVersionString = ApiVersionJune2017;
+
+#if ENABLE_MODULES_SDK
+        const string ApiVersionEdgePublicPreview = "2017-11-08-preview";
+        const string ApiVersionLatest = ApiVersionEdgePublicPreview;
+#else
+        const string ApiVersionLatest = ApiVersionJune2017;
+#endif
+
+        public const string ApiVersionString = ApiVersionLatest;
 
         public const string ApiVersionQueryString = ApiVersionQueryPrefix + ApiVersionString;
     }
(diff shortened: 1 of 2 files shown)


iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -242,7 +242,7 @@ public override void UserEventTriggered(IChannelHandlerContext context, object @
                 {
                     ClientId = id,
                     HasUsername = true,
-                    Username = $"{this.iotHubHostName}/{id}/api-version={ClientApiVersionHelper.ApiVersionString}&DeviceClientType={Uri.EscapeDataString(this.productInfo.ToString())}",
+                    Username = $"{this.iotHubHostName}/{id}/?api-version={ClientApiVersionHelper.ApiVersionString}&DeviceClientType={Uri.EscapeDataString(this.productInfo.ToString())}",
```
varunpuranik: Should we remove the '/' here?

rafilho: We shouldn't. The '?' should always be in the first character of the last segment. Otherwise, we shouldn't be able to identify if '?' is part of the id or the beginning of the property bag.

varunpuranik: Fair enough.

iothub/device/src/Common/Api/ClientApiVersionHelper.cs
```diff
@@ -8,7 +8,15 @@ class ClientApiVersionHelper
         const string ApiVersionQueryPrefix = "api-version=";
         const string ApiVersionNov2016 = "2016-11-14";
         const string ApiVersionJune2017 = "2017-06-30";
-        public const string ApiVersionString = ApiVersionJune2017;
+
+#if ENABLE_MODULES_SDK
+        const string ApiVersionEdgePublicPreview = "2017-11-08-preview";
```
tameraw: >ApiVersionEdgePublicPreview [](start = 21, length = 27)

Why can't this be the latest regardless of the #if flag? This is the modules_repo anyhow..

PR #470, opened 2018-05-04
Bump Device client version

ancaantochi:
# Description of the changes
Bump Device client to 1.17.0-preview-007

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -22,7 +22,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.17.0-preview-006</Version>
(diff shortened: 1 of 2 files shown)


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -22,7 +22,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.17.0-preview-006</Version>
+    <Version>1.17.0-preview-007</Version>
```
varunpuranik: Are these the only places we need to change? Also, can you please confirm that nothing has changed in Shared/Service SDKs, say maybe because of the FIs?

ancaantochi: only device sdk changes

PR #471, opened 2018-05-05
Replaced EscapeDataString with UrlEncode

ancaantochi:
# Description of the changes
HttpHsmClient generated code manually updated to replace System.Uri.EscapeDataString with System.Net.WebUtility.UrlEncode


iothub/device/src/HsmAuthentication/HttpHsmClient.cs
```diff
@@ -3,6 +3,7 @@
 //     Generated using the NSwag toolchain v11.17.0.0 (NJsonSchema v9.10.42.0 (Newtonsoft.Json v9.0.0.0)) (http://NSwag.org)
 // </auto-generated>
 //----------------------
+// Note: Code manually changed to replace System.Uri.EscapeDataString with System.Net.WebUtility.UrlEncode
```
damonbarry: I would append to this note: "...to be consistent with the rest of the SDK."

ancaantochi: thanks

PR #490, opened 2018-05-20
Don't subscribe to C2D messages for modules over AMQP

varunpuranik:
# Checklist

# Description of the changes
C2D is not supported for modules. So, don't subscribe to C2D messages for modules, over AMQP.

### DIFF
--- a/iothub/device/src/Transport/AmqpTransportHandler.cs
+++ b/iothub/device/src/Transport/AmqpTransportHandler.cs
@@ -126,7 +126,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
                          this.faultTolerantEventSendingLink,
                          new ConnectionEventArgs { ConnectionType = ConnectionType.AmqpTelemetry, ConnectionStatus = ConnectionStatus.Connected, ConnectionStatusChangeReason = ConnectionStatusChangeReason.Connection_Ok });
 
-                     if (!string.IsNullOrWhiteSpace(this.moduleId))
+                     if (string.IsNullOrWhiteSpace(this.moduleId))
                      {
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/AmqpTransportHandler.cs
```diff
@@ -987,7 +990,7 @@ private Twin TwinFromResponse(AmqpMessage message)
         {
             string path = this.BuildPath(CommonConstants.DeviceEventPathTemplate, CommonConstants.ModuleEventPathTemplate);
 
-            ReceivingAmqpLink messageReceivingLink = await this.IotHubConnection.CreateReceivingLinkAsync(path, this.iotHubConnectionString, this.deviceId, IotHubConnection.ReceivingLinkType.Events, this.prefetchCount, timeout, this.productInfo, cancellationToken);
+            ReceivingAmqpLink messageReceivingLink = await this.IotHubConnection.CreateReceivingLinkAsync(path, this.iotHubConnectionString, $"{this.deviceId}/{this.moduleId}", IotHubConnection.ReceivingLinkType.Events, this.prefetchCount, timeout, this.productInfo, cancellationToken);
```
ancaantochi: Is this ok if deviceId has '/' when moduleId id null or empty?

damonbarry: Can `this.ModuleId` be null (or empty string) here? Do you need to use something like the local method `BuildPath` for this?

varunpuranik: No, this is in module receive events code path.. so moduleId should never be null or empty.

varunpuranik: No, this is in module receive events code path.. so moduleId should never be null or empty.

PR #501, opened 2018-06-07
Added InvokeMethod to ModuleClient

ancaantochi:
# Checklist

# Description of the changes
- Added InvokeMethod to ModuleClient
- Removed custom header for ModuleId from AmqpServiceClient
 InvokeMethod because module call should use ModuleClient
- Refactored TrustBundle certificate validation to be able to use the callback for InvokeMethod which is over HTTP


iothub/device/src/DirectMethodRequest.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Creates an instance of DirectMethodRequest type with Zero timeout
+        /// </summary>
+        /// <param name="methodName">Method name</param>
+        /// <exception cref="ArgumentException">If <b>methodName</b> is null or whitespace</exception>
+        public DirectMethodRequest(string methodName)
+            : this(methodName, TimeSpan.Zero, TimeSpan.Zero)
```
varunpuranik: Is TimeSpan.Zero the default? Won't it timeout immediately?

iothub/device/src/Edge/CustomCertificateValidator.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            this.certs = certs;
+            this.transportSettings = transportSettings;
+        }
+
+        public void SetupCertificateValidation()
+        {
+            Debug.WriteLine("CustomCertificateValidator.SetupCertificateValidation()");
+
+            var certValidationFunc = this.GetCustomCertificateValidation();
```
varunpuranik: This API looks a bit odd. SetupCertificateValidation is calling GetCustomCertificateValidation...?

iothub/device/src/Edge/TrustBundleProvider.cs
```diff
@@ -1,119 +1,24 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Net.Http;
-using System.Net.Security;
```
varunpuranik: As per the C# SDK conventions (as in other files..), the usings are outside the namespaces. Let's keep the same conventions..

iothub/device/src/ModuleClient.cs
```diff
@@ -359,6 +373,88 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+        public Task<DirectMethodResult> InvokeMethodAsync(string deviceId, string moduleId, DirectMethodRequest directMethodRequest, CancellationToken cancellationToken)
+        {
+            return InvokeMethodAsync(GetModuleMethodUri(deviceId, moduleId), directMethodRequest, cancellationToken);
+        }
+
+        private Task<DirectMethodResult> InvokeMethodAsync(Uri uri, DirectMethodRequest directMethodRequest, CancellationToken cancellationToken)
+        {
+            var httpClientHandler = new HttpClientHandler();
+
+#if NETSTANDARD1_3 || NETSTANDARD2_0
```
varunpuranik: How does cert validation work in the other cases?

iothub/device/src/ModuleClient.cs
```diff
@@ -359,6 +373,88 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+
+        static Uri GetDeviceMethodUri(string deviceId)
+        {
+            deviceId = WebUtility.UrlEncode(deviceId);
+            return new Uri(string.Format(CultureInfo.InvariantCulture, DeviceMethodUriFormat, deviceId), UriKind.Relative);
+        }
+
+        static Uri GetModuleMethodUri(string deviceId, string moduleId)
+        {
+            deviceId = WebUtility.UrlEncode(deviceId);
```
varunpuranik: url encode moduleId

iothub/device/src/ModuleClient.cs
```diff
@@ -359,6 +373,88 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+        {
+            return InvokeMethodAsync(GetModuleMethodUri(deviceId, moduleId), directMethodRequest, cancellationToken);
+        }
+
+        private Task<DirectMethodResult> InvokeMethodAsync(Uri uri, DirectMethodRequest directMethodRequest, CancellationToken cancellationToken)
+        {
+            var httpClientHandler = new HttpClientHandler();
+
+#if NETSTANDARD1_3 || NETSTANDARD2_0
+            httpClientHandler.ServerCertificateCustomValidationCallback = this.certValidator.GetCustomCertificateValidation();
```
varunpuranik: In the InstalledCertificate case as well, this callback will be set and the install cert validation won't happen?

iothub/device/src/DirectMethodRequest.cs
ancaantochi: When the method is invoked it checks if the timeout is <= 0 it will set a default timeout of 100 seconds. Logic is in HttpTransportHandler and is the same as on ServiceClient.

---
In reply to: [193909920](https://github.com/Azure/azure-iot-sdk-csharp/pull/501#discussion_r193909920) [](ancestors = 193909920)

iothub/device/src/ModuleClient.cs
ancaantochi: I added WebRequestHandler for dotnet framework

---
In reply to: [193912087](https://github.com/Azure/azure-iot-sdk-csharp/pull/501#discussion_r193912087) [](ancestors = 193912087)

iothub/device/src/ModuleClient.cs
ancaantochi: In that case the callback is null so it uses default validation

---
In reply to: [193912949](https://github.com/Azure/azure-iot-sdk-csharp/pull/501#discussion_r193912949) [](ancestors = 193912949)

iothub/device/src/ModuleClient.cs
ancaantochi: done, also added in ServiceClient

---
In reply to: [193912221](https://github.com/Azure/azure-iot-sdk-csharp/pull/501#discussion_r193912221) [](ancestors = 193912221)

iothub/device/src/Edge/CustomCertificateValidator.cs
ancaantochi: Added Create method that does the setup

---
In reply to: [193910624](https://github.com/Azure/azure-iot-sdk-csharp/pull/501#discussion_r193910624) [](ancestors = 193910624)

iothub/device/src/Edge/TrustBundleProvider.cs
ancaantochi: done

---
In reply to: [193910901](https://github.com/Azure/azure-iot-sdk-csharp/pull/501#discussion_r193910901) [](ancestors = 193910901)

iothub/device/src/Edge/ICertificateValidator.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Azure.Devices.Client.Edge
+{
+    internal interface ICertificateValidator
```
varunpuranik: Nit: suggest rename to ICertificateValidatorProvider

iothub/device/src/ModuleClient.cs
```diff
@@ -359,6 +371,100 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+            if (customCertificateValidation != null)
+            {
+#if NETSTANDARD1_3 || NETSTANDARD2_0
+                httpClientHandler = new HttpClientHandler();
+                httpClientHandler.ServerCertificateCustomValidationCallback = customCertificateValidation;
+#else
+            httpClientHandler = new WebRequestHandler();
+            ((WebRequestHandler)httpClientHandler).ServerCertificateValidationCallback = (sender, certificate, chain, errors) =>
+            {
+                return customCertificateValidation(sender, certificate, chain, errors);
```
varunpuranik: Cool.

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -395,6 +396,33 @@ public override Task RejectAsync(string lockToken, CancellationToken cancellatio
                     cancellationToken), cancellationToken);
         }
 
+        internal Task<DirectMethodResult> InvokeMethodAsync(DirectMethodRequest directMethodRequest, Uri uri, CancellationToken cancellationToken)
+        {
+            TimeSpan timeout = GetInvokeDeviceMethodOperationTimeout(directMethodRequest);
+            var customHeaders = new Dictionary<string, string>
+            {
+                { CustomHeaderConstants.ModuleId, $"{this.deviceId}/{this.moduleId}" }
```
varunpuranik: Should never happen, but wonder if we should check that moduleId exists here.. (for future proofing).

PR #507, opened 2018-06-12
Fix trustbundle when gateway hostname is not set

ancaantochi:
# Checklist

# Description of the changes
CreateFromEnvironmentAsync should not make a call to get the trustbundle when gateway hostname is not set


mrohera:
LGTM.

PR #520, opened 2018-06-15
Recover AMQP Event Receiving connection

varunpuranik:
## Checklist
- [X] This pull-request is submitted against the `modules-preview` branch.

## Description of the changes
For AMQP, recover the link for receiving events if there is an error.

### DIFF
--- a/iothub/device/src/Transport/AmqpTransportHandler.cs
+++ b/iothub/device/src/Transport/AmqpTransportHandler.cs
@@ -238,6 +238,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
         {
             bool needMethodRecovery = false;
             bool needTwinRecovery = false;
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/AmqpTransportHandler.cs
```diff
@@ -238,6 +238,7 @@ sealed class AmqpTransportHandler : TransportHandler
         {
             bool needMethodRecovery = false;
             bool needTwinRecovery = false;
+            bool needEventReceivingLinkRecoverty = false;
```
damonbarry: typo "Recoverty"

varunpuranik: thanks

varunpuranik:
Unfortunately I couldn't find an easy way to add tests for this change, nor an existing pattern to do so.

PR #522, opened 2018-06-18
Add sample for ADM APIs

zolvarga:
## Checklist

### DIFF
--- a/iothub/service/samples/ADMSample/ADMSample.csproj
+++ b/iothub/service/samples/ADMSample/ADMSample.csproj
@@ -1,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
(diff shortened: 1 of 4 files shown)


iothub/service/samples/ADMSample/ADMSample.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
```
CIPop: We should always reference NuGets for samples instead of projects.
Is a pre-release NuGet available? If not, please add a TODO and associated Issue to track the change.

zolvarga: Yes, but the latest modules-preview Nuget package doesn't have the latest ADM APIs

---
In reply to: [196508474](https://github.com/Azure/azure-iot-sdk-csharp/pull/522#discussion_r196508474) [](ancestors = 196508474)

PR #539, opened 2018-06-28
Added Edge device capability in Provisioning Service.

abhipsaMisra:
## Checklist

## Description of the changes

- Added Edge Device capability

### DIFF
--- a/provisioning/service/src/Config/IndividualEnrollment.cs
+++ b/provisioning/service/src/Config/IndividualEnrollment.cs
@@ -141,6 +141,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
         /// <param name="createdDateTimeUtc">the <code>DateTime</code> with the date and time that the enrollment was created. This is optional and can be <code>null</code>.</param>
         /// <param name="lastUpdatedDateTimeUtc">the <code>DateTime</code> with the date and time that the enrollment was updated. This is optional and can be <code>null</code>.</param>
         /// <param name="eTag">the <code>string</code> with the eTag that identify the correct instance of the enrollment in the service. It cannot be <code>null</code> or empty.</param>
+        /// <param name="capabilities">the <see cref="DeviceCapabilities"/> that identifies the device capabilities. This is optional and can be <code>null</code>.</param>
         /// <exception cref="ProvisioningServiceClientException">if the received JSON is invalid.</exception>
         [JsonConstructor]
         internal IndividualEnrollment(
@@ -152,7 +153,8 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
             ProvisioningStatus? provisioningStatus,
             DateTime createdDateTimeUtc,
             DateTime lastUpdatedDateTimeUtc,
-            string eTag)
+            string eTag,
+            DeviceCapabilities capabilities)
         {
             /* SRS_INDIVIDUAL_ENROLLMENT_21_003: [The constructor shall throws ProvisioningServiceClientException if one of the 
                                                     provided parameters in JSON is not valid.] */
@@ -173,6 +175,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
                 CreatedDateTimeUtc = createdDateTimeUtc;
                 LastUpdatedDateTimeUtc = lastUpdatedDateTimeUtc;
                 ETag = eTag;
+                Capabilities = capabilities;
             }
             catch (ArgumentException e)
             {
@@ -313,5 +316,11 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
         /// </summary>
         [JsonProperty(PropertyName = "etag", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string ETag { get; set; }
+
+        /// <summary>
+        /// Capabilities of the device
+        /// </summary>
+        [JsonProperty(PropertyName = "capabilities", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public DeviceCapabilities Capabilities { get; set; }
     }
 }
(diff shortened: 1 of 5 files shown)


provisioning/service/src/Config/IndividualEnrollment.cs
```diff
@@ -141,6 +141,7 @@ public IndividualEnrollment(string registrationId, Attestation attestation)
         /// <param name="createdDateTimeUtc">the <code>DateTime</code> with the date and time that the enrollment was created. This is optional and can be <code>null</code>.</param>
         /// <param name="lastUpdatedDateTimeUtc">the <code>DateTime</code> with the date and time that the enrollment was updated. This is optional and can be <code>null</code>.</param>
         /// <param name="eTag">the <code>string</code> with the eTag that identify the correct instance of the enrollment in the service. It cannot be <code>null</code> or empty.</param>
+        /// <param name="capabilities">the <see cref="DeviceCapabilities"/> that identifies if the device is an Edge-enabled device. This is optional and can be <code>null</code>.</param>
```
CIPop: I think the comment shouldn't mention Edge. That is only one of the possibilities but we may be adding more in the future.

CIPop: Ah: this is internal - doesn't matter that much. Still incorrect.

provisioning/service/src/Config/IndividualEnrollment.cs
```diff
@@ -313,5 +316,12 @@ public Attestation Attestation
         /// </summary>
         [JsonProperty(PropertyName = "etag", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string ETag { get; set; }
+
+        /// <summary>
+        /// Capabilities of the device
+        /// </summary>
+ 
+       [JsonProperty(PropertyName = "capabilities", DefaultValueHandling = DefaultValueHandling.Ignore)]
```
CIPop: Nit: not properly aligned.
Nit2: Extra space between doc and attribute.

provisioning/service/src/Config/IndividualEnrollment.cs
```diff
@@ -313,5 +316,12 @@ public Attestation Attestation
         /// </summary>
         [JsonProperty(PropertyName = "etag", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string ETag { get; set; }
+
+        /// <summary>
+        /// Capabilities of the device
+        /// </summary>
+ 
+       [JsonProperty(PropertyName = "capabilities", DefaultValueHandling = DefaultValueHandling.Ignore)] 
+        public DeviceCapabilities Capabilities { get; set; }
```
CIPop: Public API change - please increase the minor version of this NuGet.

provisioning/service/src/Contract/SDKUtils.cs
```diff
@@ -5,7 +5,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
 {
     internal class SDKUtils
     {
-        private const string ApiVersionProvisioningPreview = "2017-11-15";
+        private const string ApiVersionProvisioningPreview = "2018-04-01";
```
CIPop: Please rename - we probably forgot to update the variable name: `ApiVersion` is what we use in `ClientApiVersionHelper.cs` in the IoT Hub SDK.

provisioning/service/tests/Config/IndividualEnrollmentTests.cs
```diff
@@ -74,7 +76,10 @@ public class IndividualEnrollmentTests
             "   \"provisioningStatus\":\"" + SampleProvisioningStatus + "\",\n" +
             "   \"createdDateTimeUtc\": \"" + SampleCreateDateTimeUTCString + "\",\n" +
             "   \"lastUpdatedDateTimeUtc\": \"" + SampleLastUpdatedDateTimeUTCString + "\",\n" +
-            "   \"etag\": \"" + SampleEtag + "\"\n" +
+            "   \"etag\": \"" + SampleEtag + "\",\n" +
```
CIPop: I would rather leave this as is (to check that if the Capabilities is missing, we will not send it over the wire) and create a different UT.
String concatenation should work for the expected JSon.

provisioning/service/tests/Config/IndividualEnrollmentTests.cs
```diff
@@ -259,6 +264,7 @@ public void IndividualEnrollmentConstructorJSONSucceed()
             Assert.AreEqual(SampleCreateDateTimeUTC, individualEnrollment.CreatedDateTimeUtc);
             Assert.AreEqual(SampleLastUpdatedDateTimeUTC, individualEnrollment.LastUpdatedDateTimeUtc);
             Assert.AreEqual(SampleEtag, individualEnrollment.ETag);
+            Assert.AreEqual(SampleEdgeCapability.IotEdge, individualEnrollment.Capabilities.IotEdge);
```
CIPop: We should enhance the existing test (3rd test) to assert something if the Capability was not set (should be either `false` or null).

provisioning/service/tests/Config/IndividualEnrollmentTests.cs
CIPop: We could have 2 UTs for this - when set to true and when set to false.

provisioning/service/src/Config/IndividualEnrollment.cs
```diff
@@ -313,5 +316,12 @@ public Attestation Attestation
         /// </summary>
         [JsonProperty(PropertyName = "etag", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string ETag { get; set; }
+
+        /// <summary>
+        /// Capabilities of the device
+        /// </summary>
+
```
CIPop: Nit: please remove extra space between XMLDoc and member.

PR #540, opened 2018-06-29
added RemoteCertificateChainErrors flag

paloukari:
## Checklist
- [x ] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/.github/CONTRIBUTING.md).
- [ x] This pull-request is submitted against the `master` branch.

## Description of the changes

When the **DisableServerCertificateValidation** configuration flag is set to true and the server certificate has invalid **certificate chain**, then the server certificate validation is not disabled. 

Added **RemoteCertificateChainErrors** next to **RemoteCertificateNameMismatch** as two of the disabled validation server certificate errors.

### DIFF
--- a/iothub/device/src/IotHubConnection.cs
+++ b/iothub/device/src/IotHubConnection.cs
@@ -471,7 +471,7 @@ namespace Microsoft.Azure.Devices.Client
                 return true;
             }
 
-            if (DisableServerCertificateValidation.Value && sslPolicyErrors == SslPolicyErrors.RemoteCertificateNameMismatch)
+            if (DisableServerCertificateValidation.Value && (sslPolicyErrors.HasFlag(SslPolicyErrors.RemoteCertificateNameMismatch) || sslPolicyErrors.HasFlag(SslPolicyErrors.RemoteCertificateChainErrors)))
             {
                 return true;
             }


iothub/device/src/IotHubConnection.cs
```diff
@@ -471,7 +471,7 @@ public static bool OnRemoteCertificateValidation(object sender, X509Certificate
                 return true;
             }
 
-            if (DisableServerCertificateValidation.Value && sslPolicyErrors == SslPolicyErrors.RemoteCertificateNameMismatch)
+            if (DisableServerCertificateValidation.Value && (sslPolicyErrors.HasFlag(SslPolicyErrors.RemoteCertificateNameMismatch) || sslPolicyErrors.HasFlag(SslPolicyErrors.RemoteCertificateChainErrors)))
```
CIPop: We can't take this change because it breaks app-compat by increasing the security risk for apps using the property (although there is already virtually no security for apps using this property - should only be used during debugging).

We've added public API to allow apps to have access to the CertificateValidationCallback itself:
https://github.com/Azure/azure-iot-sdk-csharp/blob/2c9bce0bc060b1869ec7a5bcfe9e63e736ed450f/iothub/device/src/AmqpTransportSettings.cs#L81
https://github.com/Azure/azure-iot-sdk-csharp/blob/cd93e75fd8914b6247f23eb6a5cfd7aea459676f/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs#L95

@paloukari While I'll reject this due to app-compat and security implications, I'm open to adding the public certificate validation for the HTTP protocol (if that's what you are missing).
__Important security consideration__ : If your application disables _any_ certificate validation (without replacing it with something else) it exposes itself to MITM attacks as the TLS channel can be easily intercepted.

/cc @varunpuranik

CIPop:
@paloukari While I'll reject this due to app-compat and security implications, I'm open to adding the public certificate validation for the HTTP protocol (if that's what you are missing).
Feel free to submit another PR (public API should be similar to the ones on AMQP/MQTT).

iothub/device/src/IotHubConnection.cs
paloukari: @CIPop, in my case the sslPolicyErrors value is 
`(RemoteCertificateNameMismatch | RemoteCertificateChainErrors)`
and the check 
`sslPolicyErrors == SslPolicyErrors.RemoteCertificateNameMismatch`
evaluates to false, resulting an invalid server certificate.

My assumption was that when the DisableServerCertificateValidation is used, **all server certificate validation errors should be ignored**. That is also my intention and my understanding, that this flag should be used only for debugging & testing purposes. 

My scenario is primarily IoT Edge unit testing and debugging. The problem I have is that when multiple `ModuleClient` instances are created in the same `AppDomain` with different `ITransportSettings[]` each, only the first connection `ITransportSettings[]` settings are honored due to this code:

https://github.com/Azure/azure-iot-sdk-csharp/blob/c93b167058708aead0d2849972c21dabf815b945/iothub/device/src/IotHubConnectionCache.cs#L26-L34

So if the first `ModuleClient` doesn't have a callback defined, no callback will be triggered for all subsequent clients that define a `CertificateValidationCallback`.

I understand your app-compat considerations and I guess you don't want to introduce a second flag for this. 

Maybe there is a better way to work this problem around. Should I create a new issue for this?

varunpuranik:
@CIPop I believe this is for AMQP. As @paloukari said, even though AMQP TransportSettings takes a RemoteCertificateValidationCallback, because of the way AMQP code works, only the first AmqpTransportSettings are used. We also ran into this problem. 
@paloukari - Since your scenario is only testing/debugging, a workaround would be to open a dummy AMQP connection first up, so that your AmqpTransportSettings are registered, and the callback gets into effect. 
If your describe your scenario a little more, maybe we can help you better.

CIPop:
> As @paloukari said, even though AMQP TransportSettings takes a RemoteCertificateValidationCallback, because of the way AMQP code works, only the first AmqpTransportSettings are used. We also ran into this problem.

@paloukari @varunpuranik please open a GitHub issue to track the AMQP issue: that sounds like a bug that should be fixed.

paloukari:
@varunpuranik I tried your recommended workaround and it works, thanks for that! 
I'm hosting the Edge Hub with all edge modules in the same app. I'm firing up the hub first to avoid the modules' `ModuleClient` to try to connect to something that is not already there, that seems to bring the clients to an unrecoverable state sometimes, this is why I cannot override the validation callback so easy.

@CIPop I'll create an issue for this shortly.

PR #548, opened 2018-07-07
Fix InvalidCast Exception (DateTimeOffset to DateTime) deserializing

JohnLTaylor:
when JsonConvert.DefaultSettings returns a JsonSerializerSettings with DateParseHandling set to DateParseHandling.DateTimeOffset

Changed the JsonReader in TwinJsonConvert calls to avoid casting and use the ReadAs... where possible.


CIPop:
Started a test run to see the current state.

PR #551, opened 2018-07-13
Fixing DNS iteration when multiple IPs are returned for the same host.

### LINKED ISSUE #549: Multipe connections detected for device in MQTT
- **OS, version, SKU and CPU architecture used:** <VERSION> 
Windows 10, Docker 18.03.1-ce-win65 linux container mode
- **SDK version used:** <VERSION> 
1.17.0

# Description of the issue:

I am trying to run an edge module in local machine (without docker). The steps are as follows
1. Start edgeHub container with all credential, network and volume. (Port is exposed correctly)
2. Open the csharp module in VSCode, and inject the connection, certificate value as environment variable
3. Run the csharp module,  error from edgeHub

 Closing AMQP device proxy for mylaptop2/target because no handler was registered.
Microsoft.Azure.Devices.Edge.Hub.Core.MultipleConnectionsException: Multiple connections detected for device mylaptop2/target

Tried to investigate, it seems the issue is caused by 
https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs#L942

It creates multiple connections in one session and cause the failure 

related to the fix of issue #495

CIPop:
## Checklist

## Description of the changes

Fixing MQTT DNS iteration to stop once a viable connection was made.

Fixes #549

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -945,6 +945,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                     {
                         Debug.WriteLine("Connecting to {0}.", address.ToString());
                         channel = await bootstrap.ConnectAsync(address, port).ConfigureAwait(false);
+                        break;
                     }
                     catch (AggregateException ae)
                     {
@@ -955,9 +956,9 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                                 Debug.WriteLine("ConnectException trying to connect to {0}: {1}", address.ToString(), ex.ToString());
                                 return true;
                             }
+
                             return false; // Let anything else stop the application.
                         });
-
                     }
                 }
 
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -22,7 +22,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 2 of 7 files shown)


CIPop:
Pending tests as described in https://github.com/Azure/azure-iot-sdk-csharp/issues/549#issuecomment-404671209 and https://github.com/Azure/azure-iot-sdk-csharp/issues/549#issuecomment-404778355

CIPop:
Thank you all for your CR! Will resolve conflicts and send this to the integration job.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -955,9 +956,9 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
                                 Debug.WriteLine("ConnectException trying to connect to {0}: {1}", address.ToString(), ex.ToString());
                                 return true;
                             }
+
```
adashen: the return in catch logic causes another bug. If there is error throws in the first call to ConnectAsync, there is no channel returned. We found this when trying the new SDK 1.17.1.  So it seems we need keep trying connection when exception happens

CIPop: @abhipsaMisra PTAL.

@adashen Please add more details: if the lambda passed to Handle returns `true` it means that the exception was handled and the `foreach` statement should continue (retry). If the return is false then we throw that exception.
https://msdn.microsoft.com/en-us/library/system.aggregateexception.handle(v=vs.110).aspx#Anchor_2

PR #570, opened 2018-07-27
Version update of all NuGet packages to pick up the Shared library

CIPop:
Increasing all versions of all NuGet packages to pick up the new Shared library version.


CIPop:
Blocked due to internal CI issues.

PR #579, opened 2018-08-14
Serialize empty method invoke body

dsajanice:
This change allows us to invoke a method from a module with an empty body. That behavior is currently broken in Edge and the invoking module gets back an error indicating that the body must not be null.

### DIFF
--- a/iothub/device/src/MethodInvokeRequest.cs
+++ b/iothub/device/src/MethodInvokeRequest.cs
@@ -23,7 +23,11 @@ namespace Microsoft.Azure.Devices.Client
         /// <exception cref="ArgumentException">If <b>methodName</b> is null or whitespace</exception>
         public MethodInvokeRequest(string methodName, string payload, TimeSpan? responseTimeout, TimeSpan? connectionTimeout)
         {
-            ValidatePayloadIsJson(payload);
+            if (!string.IsNullOrEmpty(payload))
+            {
+                ValidatePayloadIsJson(payload);
+                this.Payload = new JRaw(payload);
+            }
             if (string.IsNullOrWhiteSpace(methodName))
             {
                 throw new ArgumentException("Canot be empty", nameof(methodName));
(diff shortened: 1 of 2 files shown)


iothub/device/src/MethodInvokeRequest.cs
```diff
@@ -23,16 +23,19 @@ internal class MethodInvokeRequest
         /// <exception cref="ArgumentException">If <b>methodName</b> is null or whitespace</exception>
         public MethodInvokeRequest(string methodName, string payload, TimeSpan? responseTimeout, TimeSpan? connectionTimeout)
         {
-            ValidatePayloadIsJson(payload);
+            if (!string.IsNullOrEmpty(payload))
+            {
+                ValidatePayloadIsJson(payload);
+                this.Payload = new JRaw(payload);
+            }
             if (string.IsNullOrWhiteSpace(methodName))
```
CIPop: Two observations:
1. Unless there's any reason against it, please move the `methodName` parameter check above. We should validate params from left to right.
2. The message in line 33 is wrong - please replace with ArgumentNullException(nameof(methodName)); (Nit: why perform such validation in the first place? The method is on an internal class - our application should use internal APIs correctly.)

iothub/device/src/MethodInvokeRequest.cs
```diff
@@ -23,16 +23,19 @@ internal class MethodInvokeRequest
         /// <exception cref="ArgumentException">If <b>methodName</b> is null or whitespace</exception>
         public MethodInvokeRequest(string methodName, string payload, TimeSpan? responseTimeout, TimeSpan? connectionTimeout)
         {
-            ValidatePayloadIsJson(payload);
+            if (!string.IsNullOrEmpty(payload))
+            {
+                ValidatePayloadIsJson(payload);
+                this.Payload = new JRaw(payload);
```
CIPop: Is Leaving this.Payload == null ok?

dsajanice: Yes there are checks for Payload being null in other parts of the code.

iothub/device/src/MethodInvokeRequest.cs
dsajanice: 1. Done.
2. Replaced. From the perspective of building secure layers, I think checking in the internal class makes sense.

dsajanice:
I added a simple test that checks the code path that was originally blowing up with a null payload.

PR #597, opened 2018-08-22
Build the SLN instead of individual csproj files.

CIPop:
Speedup of the build process: using `dotnet build` to build the SLN file instead of individual csproj files. Testing and packaging remain the same.


CIPop:
Thanks @abhipsaMisra - I had to also modify the SLN for Release to avoid building UTs that require access to internals.

PR #614, opened 2018-09-05
Marking schema version as optional.

CIPop:
## Checklist

## Description of the changes
* ADM: marking the `schemaVersion` component as optional.
* Adding unit tests
* Enabling previously disabled serialization tests (test class was not marked as public).

Internal: 

> It appears that C# SDK has an requirement of “schemaVersion” to be non-null in Configuration object which is not a requirement from the service. Can we relax this in the next SDK release?


CIPop:
Thanks @varunpuranik and @sudeepster for reviewing this!

PR #637, opened 2018-09-27
Edge E2E wrapper objects

BertKleewein:
These are the wrapper objects that I'm using to test Edge E2E.  They're not particularly pretty, but they work.  Most of the code is auto-generated by swagger.io tools.  The only exceptions are the *Glue.cs files which are hand written by me and have most of the interesting logic.


CIPop:
We'll keep this portion internal until we get the common runtime public as well.

PR #638, opened 2018-09-28
Fix bug where DeviceClient amqp event links sent device id as correlation id

### LINKED ISSUE #618: AMQP should use unique correlation IDs
Thanks @nayato for reporting this offline:

We're using deviceId as correlation token: https://github.com/Azure/azure-iot-sdk-csharp/blob/2c9bce0bc060b1869ec7a5bcfe9e63e736ed450f/iothub/device/src/Transport/AmqpTransportHandler.cs#L814 
This means that when connection is re-established, client opens links with the same correlation id that we had been using before. There are some negative side effects to this:
1. On server side we can't really distinguish attempts to create req-rep channels and can link req side of one attempt to rep side of another (if there happens to be more than one attempt at a time)
2. When we see an attempt to create two links of the same type (sender or receiver) with the same correlation id, we will error out with 409 Conflict error

The recommendation is to switch to something more unique here. GUID would surely do, so would simple counter I think.

- [ ] IoT Hub DeviceClient AMQP
- [ ] IoT Hub ServiceClient AMQP
- [ ] ProvisioningTransportHandlerAmqp

timtay-microsoft:
## Checklist

## Description of the changes
In the AmqpTransportHandler class used by the Device Client, the event links wrongly provide a device id as a correlation id. The service requires a unique string here instead so that it can better correlate request and receiving links during reconnection.

This commit also factors out some common logic when assigning properties to an amqp link

Fixes #618

### DIFF
--- a/iothub/device/src/IotHubConnection.cs
+++ b/iothub/device/src/IotHubConnection.cs
@@ -53,6 +53,38 @@ namespace Microsoft.Azure.Devices.Client
             Events
         };
 
+        private String SendingLinkTypeToString(SendingLinkType linkType)
+        {
+            switch (linkType)
+            {
+                case SendingLinkType.TelemetryEvents:
+                    return "telemetry";
+                case SendingLinkType.Methods:
+                    return "methods";
+                case SendingLinkType.Twin:
+                    return "twin";
+                default:
+                    return "unknown";
+            }
+        }
+
+        private String ReceivingLinkTypeToString(ReceivingLinkType linkType)
+        {
+            switch (linkType)
+            {
+                case ReceivingLinkType.C2DMessages:
+                    return "c2d";
+                case ReceivingLinkType.Methods:
+                    return "methods";
+                case ReceivingLinkType.Twin:
+                    return "twin";
+                case ReceivingLinkType.Events:
+                    return "events";
+                default:
+                    return "unknown";
+            }
+        }
+
         protected IotHubConnection(string hostName, int port, AmqpTransportSettings amqpTransportSettings)
         {
             this.hostName = hostName;
@@ -110,14 +142,10 @@ namespace Microsoft.Azure.Devices.Client
                         break;
                 }
 
-                SetLinkSettingsCommonProperties(linkSettings, timeoutHelper.RemainingTime(), productInfo);
-                if (linkType == SendingLinkType.Methods)
+                SetLinkSettingsCommonProperties(linkSettings, timeoutHelper.RemainingTime(), productInfo, corrId, SendingLinkTypeToString(linkType));
+                if (linkType == SendingLinkType.Methods || linkType == SendingLinkType.Twin)
                 {
-                    SetLinkSettingsCommonPropertiesForMethod(linkSettings, corrId);
-                }
-                else if (linkType == SendingLinkType.Twin)
-                {
-                    SetLinkSettingsCommonPropertiesForTwin(linkSettings, corrId);
+                    linkSettings.AddProperty(IotHubAmqpProperty.ApiVersion, ClientApiVersionHelper.ApiVersionString);
                 }
 
                 var link = new SendingAmqpLink(linkSettings);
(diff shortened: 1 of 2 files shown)


iothub/device/src/IotHubConnection.cs
```diff
@@ -473,29 +465,15 @@ static AmqpSettings CreateAmqpSettings()
...
-            linkSettings.AddProperty(IotHubAmqpProperty.ApiVersion, ClientApiVersionHelper.ApiVersionString);
-            linkSettings.AddProperty(IotHubAmqpProperty.ChannelCorrelationId, "methods:" + corrId);
-            return linkSettings;
-        }
-
-        AmqpLinkSettings SetLinkSettingsCommonPropertiesForTwin(AmqpLinkSettings linkSettings, string corrId)
-        {
-            linkSettings.AddProperty(IotHubAmqpProperty.ApiVersion, ClientApiVersionHelper.ApiVersionString);
-            linkSettings.AddProperty(IotHubAmqpProperty.ChannelCorrelationId, "twin:" + corrId);
-            return linkSettings;
```
prmathur-microsoft: is it ok to ommit "twin:" or "methods:" from correlation id ?

prmathur-microsoft: Why was it needed in the first place ?

iothub/device/src/Transport/AmqpTransportHandler.cs
```diff
@@ -398,6 +399,7 @@ public override async Task EnableEventReceiveAsync(CancellationToken cancellatio
                 {
                     if (this.eventReceivedListener != null)
                     {
+                        eventConnectionCorrelationId = Guid.NewGuid().ToString("N");
```
prmathur-microsoft: this. ?

CIPop:
Failed CI: https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=1712&view=results
Please rebase over master to pick up the VSTS changes.

iothub/device/src/IotHubConnection.cs
timtay-microsoft: Those values are still being provided in the new common correlation id setter:
```cSharp
 protected static AmqpLinkSettings SetLinkSettingsCommonProperties(AmqpLinkSettings linkSettings, TimeSpan timeSpan, ProductInfo productInfo, string corrId, string linkType)
{
...
linkSettings.AddProperty(IotHubAmqpProperty.ChannelCorrelationId, linkType + ":" + corrId);
...
}
```

where linkType is either twin, methods, events

iothub/device/src/Transport/AmqpTransportHandler.cs
timtay-microsoft: Sure, I'll add the "this."

timtay-microsoft:
I went ahead and rebased from master, so now I should be even

CIPop:
@prmathur-microsoft  @timtay-microsoft @zolvarga - I would like to verify this using the logging that @zolvarga is currently working on. 
After his changes are in we can run with and w/o this code and diff logs to ensure we got the correlation IDs right.

iothub/device/src/IotHubConnection.cs
```diff
@@ -110,14 +110,10 @@ protected IotHubConnection(string hostName, int port, AmqpTransportSettings amqp
                         break;
                 }
 
-                SetLinkSettingsCommonProperties(linkSettings, timeoutHelper.RemainingTime(), productInfo);
-                if (linkType == SendingLinkType.Methods)
+                SetLinkSettingsCommonProperties(linkSettings, timeoutHelper.RemainingTime(), productInfo, corrId, linkType.ToString());
```
CIPop: Please avoid using linkType.ToString() (for both perf issues and separation of concerns - we should be able to change the internal enum name without affecting bytes on the wire).

iothub/device/src/IotHubConnection.cs
```diff
@@ -475,29 +467,15 @@ static AmqpSettings CreateAmqpSettings()
             return amqpSettings;
         }
 
-        protected static AmqpLinkSettings SetLinkSettingsCommonProperties(AmqpLinkSettings linkSettings, TimeSpan timeSpan, ProductInfo productInfo)
+        protected static AmqpLinkSettings SetLinkSettingsCommonProperties(AmqpLinkSettings linkSettings, TimeSpan timeSpan, ProductInfo productInfo, string corrId, string linkType)
         {
             linkSettings.AddProperty(IotHubAmqpProperty.TimeoutName, timeSpan.TotalMilliseconds);
-
             linkSettings.AddProperty(IotHubAmqpProperty.ClientVersion, productInfo.ToString());
+            linkSettings.AddProperty(IotHubAmqpProperty.ChannelCorrelationId, linkType + ":" + corrId);
```
CIPop: Instead of passing these two values, why not just take the correlationId (already built)?

timtay-microsoft: Previously, the linktype was being sent, too. See the old code in IotHubConnection.cs:
```cSharp
linkSettings.AddProperty(IotHubAmqpProperty.ChannelCorrelationId, "twin:" + corrId);
```

I'm not sure if the link type is necessary, but for now I'd rather not remove it altogether.

iothub/device/src/IotHubConnection.cs
timtay-microsoft: Sure thing, I'll change this

timtay-microsoft:
@CIPop I've gone ahead and pushed a new iteration addressing your comments.

CIPop:
@timtay-microsoft we're seeing a new out of memory exception with this change. Could you PTAL?

CIPop:
@zolvarga please include this with your larger architectural changes of the AMQP transport.

@timtay-microsoft @prmathur-microsoft closing for now - the change appears to cause some new memory exceptions within our E2E tests.

PR #653, opened 2018-10-03
Adding TRX file publishing and enabling CD.

CIPop:


CIPop:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=1874&view=logs passed

PR #654, opened 2018-10-04
Limit unsafe code to NET451

prmathur-microsoft:
## Checklist

### DIFF
--- a/iothub/device/src/Common/IOThreadScheduler.cs
+++ b/iothub/device/src/Common/IOThreadScheduler.cs
@@ -595,9 +595,15 @@ namespace Microsoft.Azure.Devices.Client
         // by the GC anyway.
         [Fx.Tag.SecurityNote(Critical = "manages NativeOverlapped instance, can be called outside user context")]
         [SecurityCritical]
+
+#if NET451
         unsafe class ScheduledOverlapped
+
+#else
+        class ScheduledOverlapped
+#endif
         {
-#if !NETSTANDARD1_3
+#if NET451
             readonly NativeOverlapped* nativeOverlapped;
 #endif
             IOThreadScheduler scheduler;
@@ -606,10 +612,10 @@ namespace Microsoft.Azure.Devices.Client
             {
 #if NETSTANDARD1_3
                 throw new NotImplementedException();
-#else
+#elif NET451
                 this.nativeOverlapped = (new Overlapped()).UnsafePack(
                     Fx.ThunkCallback(new IOCompletionCallback(IOCallback)), null);
-#endif
+
             }
 
             [Fx.Tag.SecurityNote(Miscellaneous = "note that in some hosts this runs without any user context on the stack")]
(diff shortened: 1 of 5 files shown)


iothub/device/src/Common/IOThreadScheduler.cs
```diff
@@ -595,9 +595,15 @@ public void DebugVerifyEmpty()
         // by the GC anyway.
         [Fx.Tag.SecurityNote(Critical = "manages NativeOverlapped instance, can be called outside user context")]
         [SecurityCritical]
+
+#if NET451
         unsafe class ScheduledOverlapped
+
+#else
+        class ScheduledOverlapped
```
CIPop: Do we need this class in anything but NET451?
Overlapped operations are always `unsafe`. Can't we just remove the file from ns13/ns20?

iothub/device/src/net451/Common/IOThreadTimer.cs
```diff
@@ -651,17 +645,13 @@ public long DueTime
...
-#else
                 this.dueTime = TimerHelper.Set(this.SafeWaitHandle, newDueTime);
-#endif
             }
             [Fx.Tag.SecurityNote(Critical = "Provides a set of unsafe methods used to work with the WaitableTimer")]
             [SecurityCritical]
             static class TimerHelper
             {
-                public static unsafe SafeWaitHandle CreateWaitableTimer()
+                public static SafeWaitHandle CreateWaitableTimer()
```
CIPop: This function is using UnsafeNativeMethods so it must be marked as `unsafe`.

iothub/device/src/net451/Common/IOThreadTimer.cs
```diff
@@ -672,7 +662,7 @@ public static unsafe SafeWaitHandle CreateWaitableTimer()
                     }
                     return handle;
                 }
-                public static unsafe long Set(SafeWaitHandle timer, long dueTime)
+                public static long Set(SafeWaitHandle timer, long dueTime)
```
CIPop: Same as above.

iothub/service/src/Common/IOThreadScheduler.cs
```diff
@@ -595,9 +595,13 @@ public void DebugVerifyEmpty()
         // by the GC anyway.
         [Fx.Tag.SecurityNote(Critical = "manages NativeOverlapped instance, can be called outside user context")]
         [SecurityCritical]
+#if NET451
         unsafe class ScheduledOverlapped
+#else
+        class ScheduledOverlapped
```
CIPop: Same as above - class contains unsafe code. It should be marked as `unsafe`. It would be ideal if we can remove it from ns builds.

prmathur-microsoft: I have limited the unsafe code to net451 and hence changing the class to unsafe only for net451.

CIPop: The nature of the class (performing IOCP / Overlapped I/O) is unsafe. If all unsafe references have been removed then the class should not be doing anything.

CIPop:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=2160&view=logs

CIPop:
One unrelated file upload test failed.

PR #656, opened 2018-10-04
E2etests part1

CIPop:
## Checklist

## Description of the changes
Production code:
* Unifying logging across `net451` and `netstandard*`
* Adding more MQTT logs.

Testing / debugging:
* Console logger was enhanced to allow VSCode colorization rules and time delta calculation.
* E2E test unification between NetFX and CoreFX for the EventHub listener.
* Template method for FaultInjection.
* Increasing VSTS test execution (newly enabled Fault Injection tests take longer)
* Disabled the device cleanup to allow Linux and Windows builds to run in parallel.


e2e/test/FaultInjection.cs
```diff
@@ -62,5 +63,121 @@ public static Client.Message ComposeErrorInjectionProperties(string faultType, s
...
+
+                s_log.WriteLine($">>> {nameof(FaultInjection)} Testing baseline");
+                await testOperation(deviceClient, testDevice).ConfigureAwait(false);
+                
+                await ActivateFaultInjection(transport, faultType, reason, delayInSec, durationInSec, deviceClient).ConfigureAwait(false);
+
+                s_log.WriteLine($">>> {nameof(FaultInjection)} Testing fault handling");
+                watch.Start();
+                s_log.WriteLine($"{nameof(FaultInjection)}: Waiting for fault injection to be active: {FaultInjection.WaitForDisconnectMilliseconds}ms");
+                await Task.Delay(FaultInjection.WaitForDisconnectMilliseconds).ConfigureAwait(false);
```
CIPop: TODO: why not delayInSec?

CIPop:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=2107&view=logs

e2e/test/FaultInjection.cs
```diff
@@ -62,5 +63,121 @@ public static Client.Message ComposeErrorInjectionProperties(string faultType, s
...
+
+            try
+            {
+                // For MQTT FaultInjection will terminate the connection prior to a PUBACK 
+                // which leads to an infinite loop trying to resend the FaultInjection message.
+                if (transport == Client.TransportType.Mqtt ||
+                    transport == Client.TransportType.Mqtt_Tcp_Only ||
+                    transport == Client.TransportType.Mqtt_WebSocket_Only)
+                {
+                    deviceClient.OperationTimeoutInMilliseconds = 1000;
```
timtay-microsoft: Could this operation timeout be shorter? I imagine you will be sending and re-sending this error injection message a few times which may be hard to account for later.

e2e/test/FaultInjection.cs
```diff
@@ -62,5 +63,121 @@ public static Client.Message ComposeErrorInjectionProperties(string faultType, s
...
+        public static async Task TestErrorInjectionTemplate(
+            string devicePrefix,
+            TestDeviceType type,
+            Client.TransportType transport,
+            string faultType,
+            string reason,
+            int delayInSec,
+            int durationInSec,
+            Func<DeviceClient, TestDevice, Task> initOperation,
+            Func<DeviceClient, TestDevice, Task> testOperation,
```
timtay-microsoft: rename "testOperation" to "baselineOperation" for clarity. We aren't testing that basic twin/methods/events scenarios work, we're testing connection recovery.

e2e/test/FaultInjection.cs
CIPop: Good point: I think we'll need to revising timings in the future. Right now each test takes 10s...

e2e/test/FaultInjection.cs
CIPop: testOperation is used for both baseline and the recovery test, not just baseline.

e2e/test/MessageReceiveTests.cs
```diff
@@ -0,0 +1,456 @@
...
+        private readonly ConsoleEventListener _listener;
+
+        public MessageReceiveTests()
+        {
+            _listener = new ConsoleEventListener("Microsoft-Azure-");
+        }
+
+        [TestMethod]
+        public async Task Message_DeviceReceiveSingleMessage_Amqp()
+        {
```
timtay-microsoft: Is there a better way to parameterize these tests? In Java, we just have a test for sending and receiving a single message, and the test suite passes in each protocol to use. Here you are still hardcoding all of the protocols to use, so you risk accidentally omitting a test case.

timtay-microsoft: It would also cut down on the space taken up by all the defined Message_DeviceReceiveSingleMessage_XXXXX methods here.

e2e/test/net451/EventHubTestListener.NetFramework.cs
```diff
@@ -0,0 +1,62 @@
...
+            EventHubReceiver receiver = null;
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
+
+            EventHubClient eventHubClient = EventHubClient.CreateFromConnectionString(Configuration.IoTHub.ConnectionString, "messages/events");
+            var eventHubPartitionsCount = eventHubClient.GetRuntimeInformation().PartitionCount;
+            string partition = EventHubPartitionKeyResolver.ResolveToPartition(deviceName, eventHubPartitionsCount);
+            string consumerGroupName = Configuration.IoTHub.EventHubConsumerGroup;
+
+            while (receiver == null && sw.Elapsed.Minutes < 1)
```
timtay-microsoft: Magic number, replace with a static constant declared at the top of the file

e2e/test/netstandard20/EventHubTestListener.NetCore.cs
```diff
@@ -0,0 +1,61 @@
...
+                EntityPath = Configuration.IoTHub.EventHubCompatibleName
+            };
+
+            EventHubClient eventHubClient = EventHubClient.CreateFromConnectionString(builder.ToString());
+            var eventRuntimeInformation = await eventHubClient.GetRuntimeInformationAsync().ConfigureAwait(false);
+            var eventHubPartitionsCount = eventRuntimeInformation.PartitionCount;
+            string partition = EventHubPartitionKeyResolver.ResolveToPartition(deviceName, eventHubPartitionsCount);
+            string consumerGroupName = Configuration.IoTHub.EventHubConsumerGroup;
+
+            while (receiver == null && sw.Elapsed.Minutes < 1)
```
timtay-microsoft: Magic number

e2e/test/MessageReceiveTests.cs
CIPop: We can easily do that when we switch to XUnit which supports data driven tests. 
I used to do this for VSTest but logging is very bad in that case so I had to roll-back.

CIPop:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=2137&view=results 
All tests passed except Provisioning. The service team is looking into the unrelated failures (no provisioning test changed in this PR).

common/test/TestModule.cs
```diff
@@ -28,9 +28,7 @@ private TestModule(Module module)
         public static async Task<TestModule> GetTestModuleAsync(string deviceNamePrefix, string moduleNamePrefix)
         {
             var log = TestLogging.GetInstance();
-            string prefix = deviceNamePrefix + "Module" + "_";
-
-            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(prefix).ConfigureAwait(false);
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(deviceNamePrefix).ConfigureAwait(false);
```
prmathur-microsoft: Why didn't you want to prefix with module? It is easy to identify with prefix..isn't it?

e2e/test/DeviceTokenRefreshE2ETests.cs
```diff
@@ -169,16 +172,23 @@ public TestTokenRefresher(string deviceId, string key) : base(deviceId)
...
+                _transport = transport;
             }
 
             protected override Task<string> SafeCreateNewToken(string iotHub, int suggestedTimeToLive)
             {
                 Console.WriteLine($"[{DateTime.UtcNow}] Refresher: Creating new token {_callCount}");
 
+                if (_transport == Client.TransportType.Mqtt)
+                {
+                    suggestedTimeToLive = -4 * 60 - 59; // server side time allowance.
```
prmathur-microsoft: Can you describe these magic numbers for future reference?

common/test/TestModule.cs
CIPop: TestDevice creates the device not the module. The module is then attached to this particular device.

e2e/test/DeviceTokenRefreshE2ETests.cs
CIPop: I've added a comment on what the 4:59 means. Do you mean we should document how the server applies the 10 minute time allowance in two separate chunks?

PR #661, opened 2018-10-08
Fix for CreateChannelFactory not correctly handling exceptions

### LINKED ISSUE #648: CreateChannelFactory not correctly handling exceptions
The CreateChannelFactory method in MqttTransportHandler.cs (line 907) contains a "foreach (IPAddress address in addresses)" with the aim to try to find the proper IP address to use.

If the address is a IPv6 one, in scenarios like IoT Edge Transparent Gateway, the ConnectAsync will fail and the exception will not be correctly catched, letting the foreach fail

Apart from asking to please catch the correct exception: do we really need to try to connect to IPv6 addresses right now? Neither IoT Hub or IoT Edge support IPv6 by default as of today
 
Could it be possible to check whether the AddressFamily of the addresses is InterNetwork or InterNetworkV6 and then only try to connect to IPv4 addresses?

danigian:
## Checklist

## Description of the changes
- Added a check on AddressFamily to accept only IPv4 addresses, as IPv6 is not supported yet for IoT Hub / Edge Hub
- Correct handling of DotNetty.Transport.Channels.ConnectException in CreateChannelFactory
Fixes #648

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -943,9 +943,14 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                 {
                     try
                     {
-                        Debug.WriteLine("Connecting to {0}.", address.ToString());
-                        channel = await bootstrap.ConnectAsync(address, port).ConfigureAwait(false);
-                        break;
+                        //Check if IPv4 as IPv6 not supported yet for IoT Hub / Edge Hub
+                        if (address.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork)
+                        {
+                            Debug.WriteLine("Connecting to {0}.", address.ToString());
+                            channel = await bootstrap.ConnectAsync(address, port).ConfigureAwait(false);
+                            break;
+                        }
+                        continue;
                     }
                     catch (AggregateException ae)
                     {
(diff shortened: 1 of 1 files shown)


CIPop:
Thanks @danigian ! I do understand the reason you want to filter DNS A records but I'm not so sure this is the right thing to solve the issue.

@tameraw @jspaith @myagley @varunpuranik I don't think disabling support for IPv6 is the right thing to do. To me, if the host doesn't support IPv6 then it shouldn't advertise it. IoT Hub doesn't and so won't run into this issue. The only problem is when the client connects to an incorrectly configured docker container (edgeHub) that has an IPv6 address but doesn't bind to it.

Any way to just enable IPv6 support for Edge? If not, can we disable IPv6 name resolution in Edge instructions?

CIPop:
1. @danigian please add the e2e scenario causing the exception you are adding. In all our tests we weren't able to generate it.
2. Please remove the IPv6 filtering. IPv6 to v4 forwarding is something that we would like to keep as a feature.

Once these changes are made, please re-open the PR.

danigian:
@CIPop thank you for the feedback about IPv6.
Maybe, you are not able to generate the exception because the MqttTransportHandlerTests is mocking the channel factory and not using any created with the CreateChannelFactory method.

To be able to reproduce the error you have to:
1) Have a machine with IPv6 enabled on network interfaces
2) Configure IoT Edge as Transparent Gateway (actually it should work even with just loading the edgeHub, it does not matter the certificates checks right now)
3) Start this [sample console application](https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/master/iot-hub/Samples/device/MessageSample) using the connection string with the GatewayHostName set to your machine hostname

If you directly reference this project you will notice the behavior of issue #648 (CreateChannelFactory looping over the foreach infinitely because of the correct exception not catched)

Please, let me know if you need more details

PR #671, opened 2018-10-12
Improve Device Explorer for accessibility users

jenoola:
Fix accessibility issues for Data tab and Create Device form


CIPop:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=2620&view=logs

PR #676, opened 2018-10-17
Provisioning Service Client v2

abhipsaMisra:
## Checklist
- [] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).

## Description of the changes

- Provisioning Service Client v2


e2e/test/netstandard20/ProvisioningCertificateValidationTest.cs
```diff
@@ -29,30 +31,29 @@ public ProvisioningCertificateValidationTest()
...
+                ProvisioningServiceClientFactory.CreateFromConnectionString(
                     Configuration.Provisioning.ConnectionStringInvalidServiceCertificate))
             {
-                Query q = provisioningServiceClient.CreateEnrollmentGroupQuery(
-                    new QuerySpecification("SELECT * FROM enrollmentGroups"));
+                var querySpecification = new QuerySpecification("SELECT * FROM enrollmentGroups");
 
-                var exception = await Assert.ThrowsExceptionAsync<ProvisioningServiceClientTransportException>(
-                    () => q.NextAsync()).ConfigureAwait(false);
+                var exception = await Assert.ThrowsExceptionAsync<HttpRequestException>(
```
abhipsaMisra: Throws HttpRequestException

common/src/ProductInfo.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Azure.Devices.Common
+{
+    internal class ProductInfo
```
CIPop: Missing test. Given the current implementation for DeviceClient, I'm not sure this actually works as expected.
We have one within DeviceClient which has a unit test. If you decide to move that into common (although Name would always be different), please ensure that everything else that's using ProductInfo uses the same implementation (DeviceClient, ProvisioningClient).

CIPop: Missing test. Given the current implementation for DeviceClient, I'm not sure this actually works as expected.

common/src/ProductInfo.cs
```diff
@@ -0,0 +1,35 @@
...
+            Assembly executingAssembly = System.Reflection.Assembly.GetExecutingAssembly();
+
+            string Name = executingAssembly.GetName().Name;
+            String version = executingAssembly.GetName().Version.ToString(3);
+
+            string runtime = RuntimeInformation.FrameworkDescription.Trim();
+            string operatingSystem = RuntimeInformation.OSDescription.Trim();
+            string processorArchitecture = RuntimeInformation.ProcessArchitecture.ToString().Trim();
+
+            string userAgent = $"{Name}/{version} ({runtime}; {operatingSystem}; {processorArchitecture})";
```
CIPop: Since we're redesigning this, I would like the initial name to be smaller. We can correctly identify the `Name` by using `IOTDC`, `IOTSC`, `DPSDC`, `DPSSC`.
/cc @yzhong94

common/src/provisioning/Authentication/AuthenticationMethodFactory.cs
```diff
@@ -1,14 +1,15 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.Azure.Devices.Common;
```
CIPop: This is moved into common - what else is referencing it?

common/src/provisioning/Authentication/AuthenticationMethodFactory.cs
```diff
@@ -1,14 +1,15 @@
...
 using System;
 
-namespace Microsoft.Azure.Devices.Common.Service.Auth
+namespace Microsoft.Azure.Devices.Provisioning.Service
 {
     /// <summary>
     /// Creates an instance of an implementation of <see cref="IAuthenticationMethod"/> based on known authentication parameters.
     /// </summary>
-    internal sealed class AuthenticationMethodFactory
+    internal static class AuthenticationMethodFactory
```
CIPop: Why is this static?

common/src/provisioning/Authentication/IAuthenticationMethod.cs
```diff
@@ -1,13 +1,18 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.Azure.Devices.Common.Service.Auth
+namespace Microsoft.Azure.Devices.Provisioning.Service
 {
     /// <summary>
     /// Authentication interface to use for IoTHub communications.
     /// </summary>
-    internal interface IAuthenticationMethod
+    public interface IAuthenticationMethod
```
CIPop: We shouldn't use IAuthenticationMethod - it didn't work well in DeviceClient.

common/src/provisioning/Authentication/IAuthorizationHeaderProvider.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.Azure.Devices.Common.Service.Auth
+namespace Microsoft.Azure.Devices.Provisioning.Service
 {
     internal interface IAuthorizationHeaderProvider
     {
```
CIPop: Same as above - this got moved, namespace changed but nothing else but service client is using it. What's the reason it resides in common?

common/src/provisioning/Authentication/ServiceConnectionString.cs
```diff
@@ -19,10 +19,9 @@ namespace Microsoft.Azure.Devices.Common.Service.Auth
     /// 
     /// This object parse and store the connection string. It is responsible to provide the authorization token too. 
     /// </remarks>
-    internal sealed class ServiceConnectionString : IAuthorizationHeaderProvider
+    public sealed class ServiceConnectionString : IAuthorizationHeaderProvider
```
CIPop: I don't think it makes sense for this class to derive from the IAuthHeaderProvider.

common/src/provisioning/Authentication/SharedAccessSignatureCredentials.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.Azure.Devices.Provisioning.Service
+{
+    /// <summary>
+    /// Shared Access Key Signature class.
+    /// </summary>
+    public class SharedAccessSignatureCredentials : ServiceClientCredentials
+    {
+        private string _sasKey;
+        private ProductInfo _productInfo = new ProductInfo();
```
CIPop: This is duplicated code - please use a base class instead.

e2e/test/netstandard20/ProvisioningCertificateValidationTest.cs
```diff
@@ -29,30 +31,29 @@ public ProvisioningCertificateValidationTest()
...
+                    () => provisioningServiceClient.QueryEnrollmentGroupsAsync(querySpecification)).ConfigureAwait(false);
 
 #if NET47 || NET451
-                Assert.IsInstanceOfType(exception.InnerException.InnerException.InnerException, typeof(AuthenticationException));
+                Assert.IsInstanceOfType(exception.InnerException.InnerException, typeof(AuthenticationException));
 #elif NETCOREAPP2_0
```
CIPop: We don't support netcoreapp 2.0 anymore - the framework is deprecated. Please remove all #if defs.

common/src/provisioning/Authentication/ServiceConnectionString.cs
```diff
@@ -113,11 +126,20 @@ public string GetPassword()
             return password;
         }
 
+        /// <summary>
+        /// Returns the Authorization header value
+        /// </summary>
+        /// <returns></returns>
         public string GetAuthorizationHeader()
```
CIPop: We should not mix things up - this is a connection string parser that should not get a particular HTTP style header.

common/src/provisioning/Authentication/ServiceConnectionString.cs
```diff
@@ -113,11 +126,20 @@ public string GetPassword()
...
         {
             return GetPassword();
         }
 
+        /// <summary>
+        /// Parser for the Provisioning Service Connection String
+        /// </summary>
+        /// <param name="connectionString"> The DPS Connection String </param>
+        /// <returns></returns>
         public static ServiceConnectionString Parse(string connectionString)
```
CIPop: Code below here doesn't seem to make sense for a connection string parser. (e.g. Token creation).

common/src/provisioning/Authentication/AuthenticationMethodFactory.cs
abhipsaMisra: This class was not required for the sas token implementation; I have removed it

common/src/provisioning/Authentication/AuthenticationMethodFactory.cs
abhipsaMisra: Same, not being used; has been removed

common/src/ProductInfo.cs
abhipsaMisra: GetExecutingAssembly() is not available in netstandard1.3. So for now, the ProductInfo in common can only be used by the ProvisioningDevoceClient and ProvisioningServiceClient. I will added unit tests for this separately.

abhipsaMisra:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=3741&view=logs

abhipsaMisra:
@CIPop Yes, since there were changes to the provisioning e2e tests in master, I am waiting until everything else is merged and rebased until I make the changes to tests here for provisioning service client v2 (to avoid conflicts).

common/src/ProductInfo.cs
abhipsaMisra: I left it in common as it is now used by both provisioning device client and provisioning service client. I also had to rearrange the "common" folder structure so that it was aligned with our project folder structure.

common/src/provisioning/Authentication/IAuthorizationHeaderProvider.cs
abhipsaMisra: Same, not being used; has been removed

common/src/provisioning/Authentication/ServiceConnectionString.cs
abhipsaMisra: The token generation logic has been removed from the parser

common/src/provisioning/Authentication/ServiceConnectionString.cs
abhipsaMisra: The token generation logic has been removed from the parser

CIPop:
@abhipsaMisra were other builds scheduled? Please add links here for reference.

abhipsaMisra:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=4138&view=logs

common/src/ProductInfo.cs
CIPop: Note this is changing in #698.

Issue #686, opened 2018-10-30
ProvisioningServiceClient_v2: Refactor the connection string parsing logic


abhipsaMisra:
Refactor the provisioning service connection string parser - remove token generation logic from the connection string class

CIPop:
@abhipsaMisra please create a new GitHub Project for these tracking issues.

abhipsaMisra:
Closing this issue; the updated autorest generator expects the credentials to be supplied in a different format; so this will need to be reworked entirely in a separate project.

az-iot-builder-01:
@CIPop, @abhipsaMisra, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

Issue #685, opened 2018-10-30
ProvisioningServiceClient_v2: Add token refreshing logic


abhipsaMisra:
Add token refreshing logic to the provisioning service client authentication

az-iot-builder-01:
@abhipsaMisra, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

Issue #684, opened 2018-10-30
ProvisioningServiceClient_v2: Add unit tests for ProductInfo in common


abhipsaMisra:
Need to add unit tests for the ProductInfo class in commons

abhipsaMisra:
The updated autorest generator defines an API surface different from the one implemented here above; once we get back to this item the entire thing will have to be re-worked. Closing this for now.

az-iot-builder-01:
@abhipsaMisra, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #697, opened 2018-11-05
Implement provisioning service token refresh logic

abhipsaMisra:
Implement token refresh logic for provisioning service client v2.

Addresses the following GitHub issues: 
[Token refreshing logic](https://github.com/Azure/azure-iot-sdk-csharp/issues/685)
[Refactor connection string parser](https://github.com/Azure/azure-iot-sdk-csharp/issues/686)


abhipsaMisra:
Open Issue:
[Unit tests for ProductInfo](https://github.com/Azure/azure-iot-sdk-csharp/issues/684)

iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -42,64 +42,64 @@
     <Compile Include="$(Common)\Logging.Common.cs">
       <Link>Common\Logging.Common.cs</Link>
     </Compile>
-    <Compile Include="$(Common)\service\AssertionFailedException.cs">
+    <Compile Include="$(Common)\iothub\service\AssertionFailedException.cs">
```
CIPop: These files are common between which projects? 
If you needed to specify "iothub" it means they are particular to service and should be moved together with the IoT Service SDK?

provisioning/service/src/Authentication/CredentialsWithTokenRefresh.cs
```diff
@@ -0,0 +1,140 @@
...
+                    return _token;
+                }
+
+                _token = await SafeCreateNewToken(_suggestedTimeToLiveSeconds).ConfigureAwait(false);
+
+                SharedAccessSignature sas = SharedAccessSignature.Parse(".", _token);
+                _expiryTime = sas.ExpiresOn;
+                UpdateTimeBufferSeconds((int)(_expiryTime - DateTime.UtcNow).TotalSeconds);
+
+                /* TODO: Add logging CaptureLogs - DPS Service
```
CIPop: TODO: missing GH Issue#.

provisioning/service/src/Authentication/CredentialsWithTokenRefresh.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Provisioning.Service
+{
+    /// <summary>
+    /// Authentication class implementing SAS token refresh
+    /// </summary>
+    public abstract class CredentialsWithTokenRefresh: ProvisioningServiceClientCredentials, IDisposable
```
CIPop: This code is copy / paste from AuthenticationWithToken refresh. Please unify or leave a TODO if you can't get to it within this PR. 
We already have a lot of code duplication - let's not add more.

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
+        private const char ValuePairSeparator = '=';
+        private const string HostNameSeparator = ".";
+
+        private const string HostNamePropertyName = nameof(HostName);
+        private const string SharedAccessKeyNamePropertyName = nameof(SharedAccessKeyName);
+        private const string SharedAccessKeyPropertyName = nameof(SharedAccessKey);
+        private const string SharedAccessSignaturePropertyName = nameof(SharedAccessSignature);
+
+        private static readonly TimeSpan regexTimeoutMilliseconds = TimeSpan.FromMilliseconds(500);
+        private static readonly Regex HostNameRegex = new Regex(@"[a-zA-Z0-9_\-\.]+$", RegexOptions.Compiled | RegexOptions.IgnoreCase, regexTimeoutMilliseconds);
```
CIPop: Please look at the memory profile - we're wasting a lot of memory compiling 4 Regex objects. 
Is there any time when we would only need to run one but not the others? If no, let's write a single RegEx (or, even better, our own string parser) that captures all elements.

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
-        /// CONSTRUCOR
+        /// Factory for new Connection String object.
         /// </summary>
-        /// <param name="parser">the <see cref="ServiceConnectionStringParser"/> with the connection string content.</param>
-        /// <exception cref="ArgumentNullException">if the provided parser is null.</exception>
-        public ServiceConnectionString(ServiceConnectionStringParser parser)
+        /// <remarks>
+        /// The connection string contains a set of information that uniquely identify an IoT Service.
+        /// 
+        /// A valid connection string shall be in the following format:
```
CIPop: Please remove:
1. Avoid the "shall" wording (not really commonly used and shouldn't be part of our public documentation). Regardless, the formatting may change and our implementation should be resilient to addition so "shall" is too strong either way.
2. We should not specify the connection string. Instead, either point to the online spec if it exists or don't add it at all. The only place where one would get a connection string is from the Azure Portal. Devices should actually avoid them since it takes considerably more memory to parse (because of the RegEx situation above).

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
             }
 
-            HostName = parser.HostName;
-            SharedAccessKeyName = parser.SharedAccessKeyName;
-            SharedAccessKey = parser.SharedAccessKey;
-            SharedAccessSignature = parser.SharedAccessSignature;
-            ServiceName = parser.ServiceName;
-            HttpsEndpoint = new UriBuilder("https", parser.HostName).Uri;
-        }
+            var ServiceConnectionStringParser = new ServiceConnectionString();
```
CIPop: Seems like parsing should be part of the ctor instead of another function call.

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
-        public Uri HttpsEndpoint
-        {
-            get;
-            private set;
-        }
+        public string SharedAccessKeyName { get; internal set; }
 
         /// <summary>
-        /// The Provisioning Service Audience
+        /// The Provisioning Service Shared Access Key for the specified
```
CIPop: Why is this on 2 lines. Please use 120 width.

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
         {
-            get;
-            private set;
+            get { return _hostName; }
+            set { SetHostName(value); }
         }
 
         /// <summary>
-        /// The Provisioning Service Client Https Endpoint
+        /// The Provisioning Service Access Policy Name
```
CIPop: Why is it policy name (like in portal) but the API is "Key Name"? This is very confusing.

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
-        }
+        public string SharedAccessSignature { get; internal set; }
 
         /// <summary>
-        /// The Provisioning Service Shared Access Key for the specified
-        /// access policy
+        /// The Provisioning Service Name
         /// </summary>
-        public string SharedAccessKey
+        public string ServiceName
```
CIPop: What's the service name? Hostname? The property name and docs are confusing.

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
         {
-            get;
-            private set;
+            get { return _serviceName; }
         }
 
         /// <summary>
-        /// The Provisioning Service Shared Access Signature
+        /// Returns the Provisioning Service Connection string
```
CIPop: Please correct capitalization. Also, end all sentences in "." (here and everywhere).
In general, for documentation please follow the MSDN guidelines or examples from the .NET Framework. The correct structure for properties is "Gets or sets ...".

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
         /// </summary>
-        public string SharedAccessSignature
+        /// <returns></returns>
```
CIPop: Don't leave any empty doc tags please.

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
-            string password;
-            if (string.IsNullOrWhiteSpace(SharedAccessSignature))
+            IDictionary<string, string> map = serviceConnectionString.ToDictionary(ValuePairDelimiter, ValuePairSeparator);
+
+            HostName = GetConnectionStringValue(map, HostNamePropertyName);
+            SharedAccessKeyName = GetConnectionStringOptionalValue(map, SharedAccessKeyNamePropertyName);
+            SharedAccessKey = GetConnectionStringOptionalValue(map, SharedAccessKeyPropertyName);
+            SharedAccessSignature = GetConnectionStringOptionalValue(map, SharedAccessSignaturePropertyName);
+
+            Validate();
```
CIPop: This is very oddly placed here: it can throw ArgumentException.

provisioning/service/src/Authentication/ServiceConnectionString.cs
```diff
@@ -1,156 +1,235 @@
...
+        }
+
+        private void Validate()
+        {
+            if (string.IsNullOrWhiteSpace(SharedAccessKeyName))
```
CIPop: Please move parameter validation within the public APIs.
1. In general the stack should be one level deep when verifying parameters.
2. This function is called several times in private functions. If something is empty but it's always required we should signal it immediately.
3. The type should be immutable - once initialized we shouldn't allow any setters.

provisioning/service/src/Microsoft.Azure.Devices.Provisioning.Service.csproj
```diff
@@ -34,8 +34,15 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="$(Common)\CommonExtensions.cs" />
-    <Compile Include="$(Common)\ProductInfo.cs" />
+    <Compile Include="$(Common)\CommonExtensions.cs">
+      <Link>Common\CommonExtensions.cs</Link>
+    </Compile>
+    <Compile Include="$(Common)\Provisioning\ProductInfo.cs">
```
CIPop: Are we sharing this between Provisioning service and client only?

provisioning/service/src/ProvisioningServiceClientFactory.cs
```diff
@@ -36,19 +34,21 @@ public static ProvisioningServiceClient CreateFromConnectionString(String connec
         /// <returns></returns>
         public static ProvisioningServiceClient CreateFromConnectionString(String connectionString, HttpClientHandler rootHandler, params DelegatingHandler[] handlers)
```
CIPop: Per our last meeting, please change this into an extension method (add `this ProvisioningServiceClient` as the first parameter).

CIPop:
@abhipsaMisra @prmathur-microsoft - this has been opened for almost a month. I'll close for now. Feel free to reopen when changes to comments above are made.

prmathur-microsoft:
This is not the appropriate place to track work items. Please use Kanban and track it against your PBI. @abhipsaMisra

PR #704, opened 2018-11-14
Remove rethrowing of exception to preserve callstack.

varunpuranik:
## Checklist

## Description of the changes
ApplyTimeout methods in InternalClient rethrow exceptions coming from TransportHandlers. Because of this, we lose the callstack of where the exceptions are coming from, making the issues hard to debug. For example, we have seen NREs being thrown by the DeviceClient in some cases, but we have no idea which area of the code they are coming from, since the callstack is lost. 
It looks like the rethrow was unnecessary, so removing it to preserve the original callstack of the exception.


prmathur-microsoft:
As a part of bigger design change @CIPop  is working on, this could potentially be handled or become a by product of it.

Issue #110, opened 2017-05-17
Make `DeviceClient` mockable


avranju:
Could we please have `DeviceClient` implement an interface called `IDeviceClient` so it's mockable for unit testing purposes in downstream projects? Thanks!

yfakariya:
Personally, I use own wrapper of `DeviceClient` for mocking in my project. It is great if `DeviceClient` become mockable!

CIPop:
Regardless of the language, the described problem is a [code smell](https://en.wikipedia.org/wiki/Code_smell) in your application's design. See chapter 8 of [Growing Object-Oriented Software, Guided by Tests](https://www.amazon.com/Growing-Object-Oriented-Software-Guided-Tests/dp/0321503627) for reasons to "Only Mock Types That You Own".

Larger SDKs (e.g. .NET Framework) are not providing public interfaces unless those interfaces (a) will never change and (b) must be implemented by application code to achieve certain behaviors (e.g. IDisposable).
Adding such an interface will work only in the version that it is added. Once we add another member to DeviceClient, to respect our binary compatibility promise the IDeviceClient will remain the same and you won't be able to mock new functionality. (We won't consider adding a new interface for each new public API.)

@yfakariya's answer is the correct one: create your own wrapper (facade or adapter). I would also recommend a factory method/class together with the wrapper. That way: 
- You know exactly what functionality your application uses from the type
- You can adapt in case of undesired changes in behavior
- You can easily mock

Here are a few online resources:
http://tech.findmypast.com/dont-mock-what-you-dont-own/
https://8thlight.com/blog/eric-smith/2011/10/27/thats-not-yours.html

StefanSchoof:
@CIPop it is controversial if "Only Mock Types That You Own" is a Code smell, see [Stack Overflow: Should you only mock types you own?](https://stackoverflow.com/questions/1906344/should-you-only-mock-types-you-own/31938571.)

I have have a azure function which is currently written in Node, but since Node Function have no Claims Support (https://github.com/Azure/azure-functions-nodejs-worker/pull/183) I considered, to move this code to c#. But with adding extra wrapper would blow this little function (39 loc and two method calls to the iothub library) disproportionately.

So please reconsider and let the user of the Library decide if they want to mock types they not own.

cdjc:
It seems to me that creating a wrapper for the sole purpose of unit testing is more of a code smell than mocking an object you don't own.

PR #718, opened 2018-11-30
Added IModuleClient Interface for public sealed ModuleClient class.

sandervandevelde:
With the current sealed ModuleClient class it is hard to make custom IoT Edge modules unit testable.

By adding this IModuleClient, programmers are now able to inject their own ModuleClient implementation in their own custom module code for injection and unit test purposes. This makes custom IoT Edge modules unit testable for almost 100 percent.

I only added the interface description and implementated in in the existing ModuleClient class. 

Adding an interface does not affect the code coverage.

### DIFF
--- a/iothub/device/src/IModuleClient.cs
+++ b/iothub/device/src/IModuleClient.cs
@@ -1,0 +1,61 @@
+﻿namespace Microsoft.Azure.Devices.Client
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Devices.Shared;
+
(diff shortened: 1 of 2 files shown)


CIPop:
Thanks @sandervandevelde - we're following .NET Framework's guidance for public API which considers interfaces to have the meaning "_object_ can do _interface_" (e.g. Socket can do disposing (IDisposable) or a List can do enumeration (IEnumerable)) and not "is a" which should be accomplished by an abstract class.

Please also reference the response for DeviceClient: https://github.com/Azure/azure-iot-sdk-csharp/issues/110

PR #722, opened 2018-12-03
Add module support for UpdateTwins2Async

### LINKED ISSUE #554: RegistryManager.UpdateTwins2Async method doesn't work when trying to update a module twin
- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64 
- **Application's .NET Target Framework :** net462 
- **Device:** Laptop 
- **SDK version used:** Microsoft.Azure.Devices.1.16.0 

## Description of the issue:
`RegistryManager.UpdateTwins2Async` method doesn't work when trying to update a module twin, instead it updates the device twin itself.
I checked with Fiddler, and the request body indeed lacks the moduleId property. I added it manually and sent the request again and the module twin updated as expected.

```
var twin = new Twin();
twin.ModuleId = "mymod";
twin.DeviceId = "device1";
twin.Tags = new TwinCollection();
twin.Tags["mytag"] = "tag";
var rm = RegistryManager.CreateFromConnectionString("<connstring>");
rm.UpdateTwins2Async(new List<Twin> { twin }, true).Wait();
```

jenoola:
## Checklist

## Description of the changes
- Add support for udpating module twins in bulk
- Add end-to-end tests to cover new functionality

Fixes #554

### DIFF
--- a/e2e/test/BulkOperationsE2ETests.cs
+++ b/e2e/test/BulkOperationsE2ETests.cs
@@ -1,0 +1,162 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.Devices.Shared;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    [TestClass]
+    [TestCategory("IoTHub-E2E")]
+    public class BulkOperationsE2ETests : IDisposable
+    {
+        private readonly string DevicePrefix = $"E2E_{nameof(BulkOperationsE2ETests)}_";
+        private readonly string ModulePrefix = $"E2E_{nameof(BulkOperationsE2ETests)}_";
+
+        [TestMethod]
+        public async Task BulkOperations_UpdateTwins2Device_Ok()
+        {
+            var tagName = Guid.NewGuid().ToString();
+            var tagValue = Guid.NewGuid().ToString();
+
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
+
+            using (RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString))
+            {
+                Twin twin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+
+                twin.Tags = new TwinCollection();
+                twin.Tags[tagName] = tagValue;
+
+                var result = await registryManager.UpdateTwins2Async(new List<Twin> { twin }, true).ConfigureAwait(false);
+                Assert.IsTrue(result.IsSuccessful, $"UpdateTwins2Async error:\n{ResultErrorsToString(result)}");
+
+                Twin twinUpd = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+
+                Assert.AreEqual(twin.DeviceId, twinUpd.DeviceId, "Device ID changed");
+                Assert.IsNotNull(twinUpd.Tags, "Twin.Tags is null");
+                Assert.IsTrue(twinUpd.Tags.Contains(tagName), "Twin doesn't contain the tag");
+                Assert.AreEqual((string)twin.Tags[tagName], (string)twinUpd.Tags[tagName], "Tag value changed");
+
+                await registryManager.CloseAsync().ConfigureAwait(false);
+            }
+        }
+
+        [TestMethod]
+        public async Task BulkOperations_UpdateTwins2DevicePatch_Ok()
+        {
+            var tagName = Guid.NewGuid().ToString();
+            var tagValue = Guid.NewGuid().ToString();
+
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
+
+            using (RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString))
+            {
+                Twin twin = new Twin();
+                twin.DeviceId = testDevice.Id;
+                twin.Tags = new TwinCollection();
+                twin.Tags[tagName] = tagValue;
+
+                var result = await registryManager.UpdateTwins2Async(new List<Twin> { twin }, true).ConfigureAwait(false);
+                Assert.IsTrue(result.IsSuccessful, $"UpdateTwins2Async error:\n{ResultErrorsToString(result)}");
+
+                Twin twinUpd = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+
+                Assert.AreEqual(twin.DeviceId, twinUpd.DeviceId, "Device ID changed");
+                Assert.IsNotNull(twinUpd.Tags, "Twin.Tags is null");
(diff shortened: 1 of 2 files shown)


e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,96 @@
...
+        private readonly string DevicePrefix = $"E2E_{nameof(BulkOperationsE2ETests)}_";
+        private readonly string ModulePrefix = $"E2E_{nameof(BulkOperationsE2ETests)}_";
+
+        private static TestLogging _log;
+        private readonly ConsoleEventListener _listener;
+
+        public BulkOperationsE2ETests()
+        {
+            _listener = TestConfig.StartEventListener();
+            _log = TestLogging.GetInstance();
```
CIPop: Static initializers should never be in a non-static ctor. See other tests for reference.

e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        [TestMethod]
+        public async Task BulkOperations_UpdateTwins2_Device()
```
CIPop: Naming: please name tests as "TestComponent_Action_ExpectedResult." See http://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html for more details.

e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            await BulkOperations_UpdateTwins2(false).ConfigureAwait(false);
```
CIPop: Always name parameters if it's not clear what they are. `setModule: false` in this case.

e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        [TestMethod]
+        public async Task BulkOperations_UpdateTwins2_Module()
+        {
+            await BulkOperations_UpdateTwins2(true).ConfigureAwait(false);
```
CIPop: Same as above.

e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        private async Task BulkOperations_UpdateTwins2(bool setModule)
+        {
+            var tagName = Guid.NewGuid().ToString();
+            var tagValue = Guid.NewGuid().ToString();
+
+            TestDevice testDevice = setModule ? null : await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
```
CIPop: This code is very hard to read. Instead of switching between testDevice and testModule, it would have been clearer to write two separate tests.

e2e/test/BulkOperationsE2ETests.cs
CIPop: This isn't used - please remove everything that's not used.

e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    [TestClass]
+    [TestCategory("IoTHub-E2E")]
+    public class BulkOperationsE2ETests : IDisposable
+    {
+        private readonly string DevicePrefix = $"E2E_{nameof(BulkOperationsE2ETests)}_";
+        private readonly string ModulePrefix = $"E2E_{nameof(BulkOperationsE2ETests)}_";
+
+        public BulkOperationsE2ETests() {}
```
CIPop: Why is this needed?

e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,105 @@
...
+                var result = await registryManager.UpdateTwins2Async(new List<Twin> { twin }, true).ConfigureAwait(false);
+                Assert.IsTrue(result.IsSuccessful, $"UpdateTwins2Async error:\n{ResultErrorsToString(result)}");
+
+                Twin twinUpd = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+
+                Assert.AreEqual(twin.DeviceId, twinUpd.DeviceId, "Device ID changed");
+                Assert.AreEqual(twin.ModuleId, twinUpd.ModuleId, "Module ID changed");
+                Assert.IsNotNull(twinUpd.Tags, "Twin.Tags is null");
+                Assert.IsTrue(twinUpd.Tags.Contains(tagName), "Twin doesn't contain the tag");
+                Assert.AreEqual(twin.Tags[tagName] as string, twinUpd.Tags[tagName] as string, "Tag value changed");
```
CIPop: It's not a good idea to do as string - if it's not string, it'll be `null` which means that AreEqual will pass which is wrong.
Use (string)twin.Tags[tagName] instead to ensure that if the type is not string the code will throw.

e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,105 @@
...
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
+
+            using (RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString))
+            {
+                Twin twin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+
+                twin.Tags = new TwinCollection();
+                twin.Tags[tagName] = tagValue;
+
+                var result = await registryManager.UpdateTwins2Async(new List<Twin> { twin }, true).ConfigureAwait(false);
```
CIPop: Sorry, I missed this in the initial review: this doesn't seem to test the bug referenced. 
Please add another test which is creating a twin patch for a particular device and one more for a twin patch for a particular device+module (without using `GetTwinAsync` initially).

e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -0,0 +1,105 @@
...
+        public async Task BulkOperations_UpdateTwins2Device_Ok()
+        {
+            var tagName = Guid.NewGuid().ToString();
+            var tagValue = Guid.NewGuid().ToString();
+
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
+
+            using (RegistryManager registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString))
+            {
+                Twin twin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
```
CIPop: Isn't the code fix below also populating now ModuleId in the twin obtained here? We should then have some asserts that the twin deviceId (and below moduleId) match the ones you expect (TestDevice.Id / TestModule.Id).

jenoola: no, it's gonna be a different device for TestModule, because TestModule generates it's own device

iothub/service/src/HttpRegistryManager.cs
```diff
@@ -828,6 +828,7 @@ static IEnumerable<ExportImportDevice> GenerateExportImportDeviceListForTwinBulk
 
                 var exportImportDevice = new ExportImportDevice();
                 exportImportDevice.Id = twin.DeviceId;
+                exportImportDevice.ModuleId = twin.ModuleId;
```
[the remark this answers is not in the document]
jenoola: no, I still need to figure out what to do with other fields. I want to fix customer bug sooner, so this PR is only about supporting module twins

CIPop:
@jenoola - did the tests pass?

CIPop:
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=6146

PR #746, opened 2019-01-04
Test enhancements

CIPop:
A few test enhancements to lower the number of false positives and provide more information when things fail.


e2e/test/TwinFaultInjectionTests.cs
```diff
@@ -199,7 +199,10 @@ private async Task Twin_DeviceReportedPropertiesRecovery(Client.TransportType tr
 
                 await deviceClient.UpdateReportedPropertiesAsync(props).ConfigureAwait(false);
 
-                var deviceTwin = await deviceClient.GetTwinAsync().ConfigureAwait(false);
+                Twin deviceTwin = await deviceClient.GetTwinAsync().ConfigureAwait(false);
```
zolvarga: >Twin [](start = 16, length = 4)

Just curious: The type needed for the Assert?

CIPop: Some testing found a null ref within the final assert line. This is to get more information of what exactly was null.

PR #747, opened 2019-01-04
DeviceExplorer version update and enhancements

CIPop:
1. Increasing DeviceExplorer dependency versions.
2. Increasing DeviceExplorer version.
3. Small optimization to allow operating with a high number of devices.

### DIFF
--- a/tools/DeviceExplorer/DeviceExplorer/MainForm.cs
+++ b/tools/DeviceExplorer/DeviceExplorer/MainForm.cs
@@ -407,7 +407,6 @@ namespace DeviceExplorer
                 await updateDevicesGridView();
                 devicesListed = true;
                 listDevicesButton.Text = "Refresh";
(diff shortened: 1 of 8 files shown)


tools/DeviceExplorer/DeviceExplorer/MainForm.cs
```diff
@@ -403,9 +403,11 @@ private async void UpdateListOfDevices()
         {
             try
             {
+                listDevicesButton.Enabled = false;
                 await updateDevicesGridView();
                 devicesListed = true;
                 listDevicesButton.Text = "Refresh";
+                listDevicesButton.Enabled = true;
```
zolvarga: >listDevicesButton.Enabled = true; [](start = 16, length = 33)

Move out from the try-catch into the finally otherwise in a case of exception the button will never be enabled

PR #771, opened 2019-01-25
ModuleAuthenticationWithHSM should encode deviceId and moduleId

varunpuranik:
## Checklist

## Description of the changes
The deviceId and moduleId should be encoded before being used. This is to allow all the characters that IoTHub supports for the deviceId and moduleId.


iothub/device/src/HsmAuthentication/ModuleAuthenticationWithHsm.cs
```diff
@@ -36,7 +37,7 @@ public ModuleAuthenticationWithHsm(ISignatureProvider signatureProvider, string
         protected override async Task<string> SafeCreateNewToken(string iotHub, int suggestedTimeToLive)
         {
             DateTime startTime = DateTime.UtcNow;
-            string audience = SasTokenBuilder.BuildAudience(iotHub, this.DeviceId, this.ModuleId);
+            string audience = SasTokenBuilder.BuildAudience(iotHub, WebUtility.UrlEncode(this.DeviceId), WebUtility.UrlEncode(this.ModuleId));
```
arsing: Shouldn't this one be done by `BuildAudience` itself?

arsing: Also, `BuildAudience` first constructs the `"{0}/devices/{1}/modules/{2}"` string and then `UrlEncode`s the whole thing. Should it be changed to encode the parts instead to be consistent with the other cases you've added?

varunpuranik: yeah, moved it to BuildAudience

prmathur-microsoft:
You may have to rebase to latest master for this to get in.

varunpuranik:
> You may have to rebase to latest master for this to get in.

Thanks for the PR. I rebased with master.

iothub/device/tests/HsmAuthentication/ModuleAuthenticationWithHsmTest.cs
```diff
@@ -33,16 +33,44 @@ public async Task TestSafeCreateNewToken_ShouldReturnSasToken()
...
+            Assert.AreEqual(string.Empty, token.KeyName);
+        }
+
+        [TestMethod]
+        public async Task TestSafeCreateNewToken_ShouldReturnSasToken_DeviceIdWithChars()
+        {
+            // Arrange
+            string deviceId = "n@m.et#st";
+            string moduleId = "$edgeAgent";
+            var httpClient = new Mock<ISignatureProvider>();
```
prmathur-microsoft: Is it possible to add E2E for this scenario ?

varunpuranik: Any Edge E2E scenario with Device IDs that have special characters should cover it. 
We will probably add E2E tests for this as well, once the SDK gets fixed.

CIPop: Can we get an E2E test where we create such a device, use it then delete it? I'm not very familiar with the HSM Module code or how to test this out with Edge or IotHub.

iothub/device/tests/HsmAuthentication/ModuleAuthenticationWithHsmTest.cs
```diff
@@ -33,16 +33,44 @@ public async Task TestSafeCreateNewToken_ShouldReturnSasToken()
             string sasToken = await moduleAuthenticationWithHsm.GetTokenAsync(this.iotHub);
             SharedAccessSignature token = SharedAccessSignature.Parse(iotHub, sasToken);
 
-            string audience = WebUtility.UrlEncode(string.Format(CultureInfo.InvariantCulture, "{0}/devices/{1}/modules/{2}",
```
CIPop: This is different from the prod code yet it works. Can you please add an international domain name (IDN) for the hub. Do we even support that? (If not, maybe we should not UrlEncode the whole string.)

varunpuranik: It is not different, it is the same. Just that encoding is done in 2 different places with the IoTHub/Prod code (which is probably why it went unnoticed). I just moved it to be in the same place, so that there is no confusion.

iothub/device/src/HsmAuthentication/SasTokenBuilder.cs
```diff
@@ -36,10 +36,11 @@ public static string BuildExpiresOn(DateTime startTime, TimeSpan timeToLive)
 
         public static string BuildAudience(string iotHub, string deviceId, string moduleId)
         {
+            // DeviceId and ModuleId need to be double encoded.
             string audience = WebUtility.UrlEncode("{0}/devices/{1}/modules/{2}".FormatInvariant(
                 iotHub,
-                deviceId,
-                moduleId));
+                WebUtility.UrlEncode(deviceId),
```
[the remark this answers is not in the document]
CIPop: Talked offline with @varunpuranik - test can be made only with Edge and must have the following:
            string deviceId = "n@m.et#st";
            string moduleId = "$edgeAgent";

Tracking with #829 as our Edge test integration is still work in progress.

CIPop:
Working with @leantk to unblock testing for this PR.

PR #779, opened 2019-01-30
Updated fake keys used in unit tests and added corresponding suppression file

abhipsaMisra:
## Checklist

## Description of the changes

Suppress false positives of fake connection strings flagged by CredScan

### DIFF
--- a/vsts/CredScanSuppressions.json
+++ b/vsts/CredScanSuppressions.json
@@ -1,0 +1,37 @@
+{
+    "tool": "Credential Scanner",
+    "suppressions": [
+        {
+            "placeholder": "dGVzdFN0cmluZzE=",
+            "_justification": "Dummy secret used in unit tests, it is fake"
+        },
+        {
+            "placeholder": "dGVzdFN0cmluZzI=",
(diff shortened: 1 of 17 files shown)


vsts/CredScanSuppressions.json
```diff
@@ -0,0 +1,37 @@
...
+            "_justification": "Dummy secret used in unit tests, it is fake"
+        },
+        {
+            "placeholder": "notsafe",
+            "_justification": "Dummy secret used in unit tests, it is fake"
+        },
+        {
+            "file": ["\\device\\external_modules\\dotnetty\\shared\\dotnetty.com.pfx", "\\device\\external_modules\\dotnetty\\shared\\contoso.com.pfx "],
+            "_justification": "The PFX certificate is required for building the UWP Sample"
+        }
```
prmathur-microsoft: Is this needed during runtime? Why can't you leave it at "Your PFX cert here" as a string placeholder?

abhipsaMisra: These are required by dotnetty tests and examples, which are included in our app as an external module. Since the certs and their references are both a part of the dotnetty repo, we have explicitly suppress those occurrences.

PR #790, opened 2019-02-08
Reprovisioning support for dps

timtay-microsoft:
This PR includes all feature implementation needed to enable reprovisioning.

This includes all e2e tests.

### DIFF
--- a/e2e/test/netstandard20/ProvisioningE2ETests.cs
+++ b/e2e/test/netstandard20/ProvisioningE2ETests.cs
@@ -47,515 +47,508 @@ namespace Microsoft.Azure.Devices.E2ETests
             Group
         }
 
-        public enum TransportProtocol
-        {
-            Http,
-            Mqtt,
-            Amqp
-        }
-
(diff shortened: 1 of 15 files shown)


timtay-microsoft:
Temporarily disabling non-provisioning tests. DO NOT MERGE YET

timtay-microsoft:
//AzurePipelines run all

timtay-microsoft:
//AzurePipelines run

e2e/test/netstandard20/ProvisioningE2ETests.cs
```diff
@@ -50,274 +47,664 @@ public enum EnrollmentType
             Group
         }
 
+        public enum TransportProtocol
```
jenoola: Why are you introducing this enum instead of using existing one?

timtay-microsoft: I'll see if I can rewrite these tests to reuse that enum, sure

PR #750, opened 2019-01-07
Dps symmetric key support

timtay-microsoft:
## Description of the changes
Adding support for the new Symmetric Key attestation type for DPS. These changes impact both the provisioning service client, and the provisioning device client.

PR #791, opened 2019-02-11
Adding retry to DPS device registration operations

timtay-microsoft:
This PR adds a passive retry mechanism to all device registration operations. If a provisioning device client receives a transient exception (429 and above) during registration status polling, the sdk will log that it happened, and then continue to poll for updates until a final status is reached.


provisioning/transport/http/src/Generated/RuntimeRegistration.cs
```diff
@@ -159,13 +159,22 @@ public RuntimeRegistration(DeviceProvisioningServiceRuntimeClient client)
             HttpStatusCode _statusCode = _httpResponse.StatusCode;
             cancellationToken.ThrowIfCancellationRequested();
             string _responseContent = null;
+
+            if ((int)_statusCode == 429 || (int)_statusCode >= 500)
+            {
+                var ex = new ProvisioningTransportException("Received status code " + _statusCode, null, true);
```
rajeevmv: >ProvisioningTransportException [](start = 29, length = 30)

Will this cause a retry? i.e., is this a transient exception?

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -309,7 +322,25 @@ private async Task VerifyPublishPacketTopicAsync(IChannelHandlerContext context,
...
+                    if (statusCode == HttpStatusCode.BadRequest)
+                    {
+                        await FailWithExceptionAsync(
+                         context,
+                         new ProvisioningTransportException(
+                             jsonData,
+                             null,
+                             false)).ConfigureAwait(false);
+                    }
+                    else if ((int)statusCode == 429)
```
rajeevmv: >(int)statusCode == 429) [](start = 29, length = 23)

looks like this condition is covered in line 347 below.

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -309,7 +322,25 @@ private async Task VerifyPublishPacketTopicAsync(IChannelHandlerContext context,
             {
                 if (Enum.TryParse(match.Groups[1].Value, out HttpStatusCode statusCode))
                 {
-                    if (statusCode >= HttpStatusCode.BadRequest)
+                    if (statusCode == HttpStatusCode.BadRequest)
```
rajeevmv: >BadRequest [](start = 53, length = 10)

even this condition appears to be covered by line 347

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: As per your note [here](https://github.com/Azure/azure-iot-sdk-csharp/pull/750#discussion_r249187971) aren't 429 and 400 special cases where we cannot parse into a ProvisioningErrorDetails object? So don't we need these special cases where we do not parse into a json object?

rajeevmv: Sorry, I was answering the question of whether the response would contain the status of 'Assigning'.  Anush is correct and you should receive an object of type 'ProvisioningServiceErrorDetails' which you are deserializing on line 345.

---
In reply to: [255747413](https://github.com/Azure/azure-iot-sdk-csharp/pull/791#discussion_r255747413) [](ancestors = 255747413)

timtay-microsoft: Ah okay, I'll leave the error parsing as it was, then.

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: I'll revert this to maintain the current error parsing then

timtay-microsoft:
@rajeevmv I have updated the PR to no longer treat 400/429 as special cases. Can you take one last look today?

timtay-microsoft:
//AzurePipelines run all

timtay-microsoft:
Closing this since the feature was abandoned for now

Issue #751, opened 2019-01-08
ModuleClient not able to send/receive messages over MQTT after EdgeHub restart with v1.19.0


varunpuranik:
## Description of the issue:
I have a setup on IoT Edge with 2 modules, both connecting to EdgeHub over MQTT. The routes have been set up so that one module generates messages and the other module receives messages.
The communication goes on successfully, but if EdgeHub restarts (or is restarted), then both modules stop communicating (sending or receiving messages). 

Points to note - 
- This issue does not repro if the modules are connecting to the EdgeHub over AMQP
- This does not repro with the previous version of the Client SDK - Microsoft.Azure.Devices.Client v1.18.1. So it seems like a regression from the latest changes.

Here are the console logs from the client with the repro. Looks like the MqttTransportHandler is not properly closed before the retry operation.

_k-csharp\iothub\device\src\Transport\Mqtt\MqttTransportHandler.cs:line 1001
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttTransportHandler.<OpenAsync>d__48.MoveNext() in E:\workspace\azure-iot-sdk-csharp\iothub\device\src\Transport\Mqtt\MqttTransportHandler.cs:line 167
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Azure.Devices.Client.Transport.ProtocolRoutingDelegatingHandler.<OpenAsync>d__5.MoveNext() in E:\workspace\azure-iot-sdk-csharp\iothub\device\src\Transport\ProtocolRoutingDelegatingHandler.cs:line 69
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.<>c__DisplayClass22_0.<<ExecuteWithErrorHandlingAsync>b__0>d.MoveNext() in E:\workspace\azure-iot-sdk-csharp\iothub\device\src\Transport\ErrorDelegatingHandler.cs:line 152
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #795, opened 2019-02-13
Add IsReady flag to IDelegatingHandler to check MqttTransportHandler state

varunpuranik:
## Checklist

## Description of the changes

- With the changes to the Retry logic that were made as part of commit https://github.com/Azure/azure-iot-sdk-csharp/commit/35ab2ac37a220b8ff558e2f3a85881a980ff3efb, the retry logic no longer closes the underlying handlers. However, this causes problems when using MqttTransportHandler.  

- MqttTransportHandler has state. So when an error occurs, it goes into an Error state, and is not usable anymore. EnsureValidState which is called before every operation, will throw if the current state is Error. 

- This is however different from AmqpTransportHandler, where each link has its own state, and the entire transport should not be closed/abandoned if one link runs into errors. 

- To handle this, I added an IsReady flag to IDelegatingHandler. It returns true for all cases, but for MqttTransportHandler, where it returns true only if the state is not Error or Closed. 

- This way, the ProtocolRoutingDelegatingHandler, which is responsible for creating the actual "transport" handlers can create a fresh transport handler if it is not ready. This should happen only when the MqttTransportHandler is in an Error state.

https://github.com/Azure/azure-iot-sdk-csharp/issues/751


iothub/device/src/IDelegatingHandler.cs
```diff
@@ -15,6 +15,7 @@ interface IDelegatingHandler : IContinuationProvider<IDelegatingHandler>, IDispo
         Task OpenAsync(CancellationToken cancellationToken);
         Task CloseAsync(CancellationToken cancellationToken);
         Task WaitForTransportClosedAsync();
+        bool IsReady { get; }
 
         // Telemetry uplink.
```
prmathur-microsoft: Quick question: why isn't it possible to close before handing over the control to these layers from mqtt layer?

varunpuranik: Yes, we can close, but the upper layers won't be aware of it..

prmathur-microsoft:
As we discussed, post testing we will merge it.

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
```diff
@@ -66,6 +66,7 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
 
                 try
                 {
+                    CreateNewTransportIfNotReady();
```
CIPop: I don't really like the addition of this function as it shows TransportHandler.OpenAsync is not callable again after failure. This should be fixed to all transports in the future.

varunpuranik: Note that the failure need not be in OpenAsync. This simply says whether the Transport is in a bad state or not. If it is not, then you can call OpenAsync on it. 
We can extend this to do what you are saying by having MqttTransportHandler not get into Error state when OpenAsync fails. This will still work in that case.
So, in my mind, this is a most generic check.

CIPop:
@leantk Can you please take a look at why the azure bot doesn't work anymore? Looks like 2 tests are pending yet the bot is reporting none are associated...

leantk:
@CIPop taking a look...

PR #805, opened 2019-02-19
Test Varunpuranik mqtt retry fix1

jenoola:
## Checklist


iothub/device/src/Transport/ErrorDelegatingHandler.cs
```diff
@@ -30,6 +31,7 @@ public ErrorDelegatingHandler(IPipelineContext context, IDelegatingHandler inner
         {
             typeof(IOException),
             typeof(SocketException),
+            typeof(ClosedChannelException),
             typeof(TimeoutException),
```
prmathur-microsoft: Is this dotnetty exception or a type of system exception ?

prmathur-microsoft: The idea over here is to bundle up network exceptions so I don't follow this.

jenoola:
The testing this PR was created for went successful, the fix will be introduced in the next release along with other changes

PR #846, opened 2019-03-15
Release 2019 3 11

CIPop:
Changes made for the preview release on 3/11.


CIPop:
@leantk could you PTAL why ebelouso-csharp-canary is triggered automatically from VSTS. Looks like some bug in VSTS triggers (we have everything disabled in triggers).

PR #854, opened 2019-03-18
Increase test workers to 32.

abhipsaMisra:
Increase the test runners to 32


CIPop:
@abhipsaMisra I think it would be valuable if we move Windows to parallel and we keep Linux as-is until we can push a TSS.MSR fix to avoid the deadlock.

CIPop:
Closing for now - we can change this only on Windows for now.

PR #866, opened 2019-03-21
Add DeviceCapabilities to EnrollmentGroup

timtay-microsoft:
Bringing the DeviceCapabilities field to enrollmentGroup now, too. Adding e2e tests that provision a device through enrollmentGroups and individual enrollments with the flag and verify that the device has the correct capabilities.

This pr does not include a full e2e scenario wherein the provisioned device is used to run edgehub.

feature is currently still in preview on service side


e2e/test/netstandard20/ProvisioningE2ETests.cs
```diff
@@ -263,6 +263,84 @@ public async Task ProvisioningDeviceClient_ValidRegistrationId_MqttWs_SymmetricK
             await ProvisioningDeviceClient_ValidRegistrationId_Register_Ok(Client.TransportType.Mqtt, AttestationType.SymmetricKey, EnrollmentType.Individual, false).ConfigureAwait(false);
         }
 
+        [TestMethod]
```
CIPop: Please write another test - this is already getting too crowded.
Also, please vary only relevant items - I don't see how using TPM vs X509 would be relevant for how we fill in `DeviceCapabilities()`. I think having 3 tests for the 3 protocols should be sufficient (WSS won't matter here either).

e2e/test/netstandard20/ProvisioningE2ETests.cs
```diff
@@ -1012,7 +1105,22 @@ private async Task ConfirmRegisteredDeviceWorks(DeviceRegistrationResult result,
             }
         }
 
-        private async Task<SecurityProvider> CreateSecurityProviderFromName(AttestationType attestationType, EnrollmentType? enrollmentType, string groupId, ReprovisionPolicy reprovisionPolicy, AllocationPolicy allocationPolicy, CustomAllocationDefinition customAllocationDefinition, ICollection<string> iothubs)
+        private async Task ConfirmExpectedDeviceCapabilities(DeviceRegistrationResult result, DeviceCapabilities capabilities)
```
CIPop: I think this isn't testing the right thing. In my mind, the right scenario would be:
1. DPS Service sets the capabilities (via service client)
2. Device performs DPS provisioning.
3. Device connects to IoT Hub and queries the Twin to see the capabilities.

The test below is querying RegistryManager for some reason - that isn't a device-side API.

e2e/test/netstandard20/ProvisioningE2ETests.cs
timtay-microsoft: Protocol shouldn't matter, nor should attestation type, you are right. Individual enrollment vs group enrollment does matter, so I'll just have one test case for each of those instead of what I have right now.

timtay-microsoft: As for the test suite getting too large, I'll take some time to move all of the reprovisioning tests into a separate suite for this PR. The tests I'm writing for device capabilities do belong in this suite, though.

e2e/test/netstandard20/ProvisioningE2ETests.cs
timtay-microsoft: I'll change this to use the device side twin to verify the flag is set as expected instead of the registry manager, sure

timtay-microsoft: Actually, I don't see a way to query this information from a device side query. We have to use the registry manager here to see if the device is an edge device or not.

e2e/test/netstandard20/ProvisioningE2ETests.cs
```diff
@@ -1241,10 +989,8 @@ public static string ComputeDerivedSymmetricKey(byte[] masterKey, string registr
             }
         }
 
-        private bool ImplementsWebProxy(Client.TransportType transportProtocol)
+        public static bool ImplementsWebProxy(Client.TransportType transportProtocol)
```
davilu: Why public?

provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -179,7 +179,7 @@ private static async Task CreateLinksAsync(AmqpClientConnection connection, stri
             AmqpClientConnection client,
             string correlationId)
         {
-            var amqpMessage = AmqpMessage.Create(new AmqpValue() { Value = DeviceOperations.Register });
+            var amqpMessage = AmqpMessage.Create(new MemoryStream(), true);
```
davilu: DPS server API change?

timtay-microsoft: yes

e2e/test/netstandard20/ProvisioningE2ETests.cs
timtay-microsoft: It is used by the reprovisioning test suite now, too

PR #873, opened 2019-03-23
Amqp redesign preview

davilu:
## Checklist


e2e/test/DeviceTokenRefreshE2ETests.cs
```diff
@@ -157,10 +156,7 @@ public int SafeCreateNewTokenCallCount
             {
                 get
                 {
-                    lock (_lock)
-                    {
-                        return _callCount;
-                    }
+                    return _callCount;
```
CIPop: Why not interlocked? These can / will be on different threads.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -2,54 +2,35 @@
...
     internal class AmqpAuthenticationRefresher : IAmqpAuthenticationRefresher
     {
+        private readonly TimeSpan MinWaitTime = TimeSpan.FromSeconds(1);
         private readonly AmqpCbsLink AmqpCbsLink;
         private readonly IotHubConnectionString ConnectionString;
         private readonly string Audience;
         private CancellationTokenSource CancellationTokenSource;
         private TimeSpan OperationTimeout;
-        private Task RefreshLoop;
+        private ConfiguredTaskAwaitable RefreshLoop;
```
CIPop: This should be a Task.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -85,7 +66,11 @@ private async Task RefreshLoopAsync(DateTime expiry, CancellationToken cancellat
             {
                 if (Logging.IsEnabled) Logging.Info(this, expiry, $"Before {nameof(RefreshLoopAsync)}");
 
-                if (waitTime.Milliseconds > 0)
+                if (waitTime < MinWaitTime)
```
CIPop: There shouldn't be a min wait time. Having one may unnecessarily delay a refresh which could take a very long time within the process of getting/sending the token.

iothub/device/src/Transport/DeviceIdentity.cs
```diff
@@ -38,6 +42,25 @@ internal DeviceIdentity(IotHubConnectionString iotHubConnectionString, AmqpTrans
                 AuthenticationModel = AuthenticationModel.X509;
             }
         }
+        private static string CreateAudience(IotHubConnectionString connectionString)
+        {
+            if (connectionString.SharedAccessKeyName.IsNullOrWhiteSpace())
+            {
+                if (connectionString.ModuleId.IsNullOrWhiteSpace())
+                {
+                    return $"{connectionString.HostName}/devices/{WebUtility.UrlEncode(connectionString.DeviceId)}";
```
CIPop: Please pick-up from `master`: this appears to be missing recent fixes.

iothub/device/src/ModuleClient.cs
```diff
@@ -43,6 +43,7 @@ internal ModuleClient(InternalClient internalClient, ICertificateValidator certV
             {
                 throw new ArgumentException("A valid module ID should be specified to create a ModuleClient");
             }
+            if (Logging.IsEnabled) Logging.Associate(this, this, internalClient, nameof(this.internalClient));
```
CIPop: Nit: missing NL after closing brace. (Please follow our coding style guidance.)

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
davilu: I added one second min interval delay to avoid infinite loop.

iothub/device/src/Transport/DeviceIdentity.cs
davilu: done

iothub/device/src/ModuleClient.cs
davilu: done

e2e/test/DeviceTokenRefreshE2ETests.cs
davilu: done

iothub/device/src/Transport/DeviceIdentity.cs
davilu: done

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
davilu: done

e2e/test/DeviceTokenRefreshE2ETests.cs
```diff
@@ -205,7 +205,7 @@ protected override Task<string> SafeCreateNewToken(string iotHub, int suggestedT
                 {
                     _callCount++;
                 }
-
+
```
CIPop: Nit: please remove empty spaces

iothub/device/src/HsmAuthentication/SasTokenBuilder.cs
```diff
@@ -36,12 +36,7 @@ public static string BuildExpiresOn(DateTime startTime, TimeSpan timeToLive)
 
         public static string BuildAudience(string iotHub, string deviceId, string moduleId)
         {
-            string audience = WebUtility.UrlEncode("{0}/devices/{1}/modules/{2}".FormatInvariant(
-                iotHub,
-                deviceId,
-                moduleId));
-
-            return audience;
+            return WebUtility.UrlEncode($"{iotHub}/devices/{WebUtility.UrlEncode(deviceId)}/modules/{WebUtility.UrlEncode(moduleId)}");
```
CIPop: @varunpuranik I don't think this is correct. I believe the "/" characters should not be escaped.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
CIPop: @davilu it's the same here...

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -58,21 +41,21 @@ public async Task InitLoopAsync(TimeSpan timeout)
                     ConnectionString,
                     ConnectionString.AmqpEndpoint,
                     Audience,
-                    Audience,
-                    AccessRightsHelper.AccessRightsToStringArray(AccessRights.DeviceConnect),
-                    timeoutHelper.RemainingTime()
+                    WebUtility.UrlEncode(Audience),
```
CIPop: @varunpuranik @davilu I doubt this is correct - we need to Uri (System.Uri) encode which is different what WebUtility does (some form of percent encoding).

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -58,21 +41,21 @@ public async Task InitLoopAsync(TimeSpan timeout)
...
             }
-            if (Logging.IsEnabled) Logging.Exit(this, timeoutHelper.RemainingTime(), $"{nameof(InitLoopAsync)}");
+            if (Logging.IsEnabled) Logging.Exit(this, timeout, $"{nameof(InitLoopAsync)}");
         }
 
         private void StartLoop(DateTime expiry, CancellationToken cancellationToken)
         {
             if (Logging.IsEnabled) Logging.Enter(this, expiry, $"{nameof(StartLoop)}");
-            RefreshLoop = RefreshLoopAsync(expiry, cancellationToken);
+            RefreshLoop = RefreshLoopAsync(expiry, cancellationToken).ConfigureAwait(false);
```
CIPop: ConfigureAwait is wrong here. Task needs to be assigned only.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
CIPop: In that case, we should include the refresh time (start a Stopwatch and check to see how long that took). This code will _add_ a second even if the refresh took minutes.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -98,8 +85,8 @@ private async Task RefreshLoopAsync(DateTime expiry, CancellationToken cancellat
                             ConnectionString,
                             ConnectionString.AmqpEndpoint,
                             Audience,
-                            Audience,
-                            AccessRightsHelper.AccessRightsToStringArray(AccessRights.DeviceConnect),
+                            WebUtility.UrlEncode(Audience),
```
CIPop: @varunpuranik PTAL - I don't think this is right either.

iothub/device/src/Transport/Amqp/AmqpLinkHelper.cs
```diff
@@ -39,8 +39,7 @@ TimeSpan timeout
                 LinkName = CommonResources.GetNewStringGuid(linkSuffix),
                 Role = false,
                 InitialDeliveryCount = 0,
-                Target = new Target() { Address = BuildLinkAddress(deviceIdentity, deviceTemplate, moduleTemplate) },
-                Source = new Source() { Address = deviceIdentity.IotHubConnectionString.DeviceId }
```
CIPop: Can we get confirmation if this is needed?

iothub/device/src/HsmAuthentication/SasTokenBuilder.cs
varunpuranik: This looks correct - IoThub will UrlDecode the audience (and then UrlDecode the deviceId and moduleId)

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
varunpuranik: BuildAudience is already doing a UrlEncode of the Audience - this would do this twice, right? 
We need to make sure all code paths that get the Audience are consistent (including the non-connection string ones like the Edge scenario)

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
varunpuranik: Yeah, this looks unnecessary, but I haven't looked at other code paths.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
davilu: This parameter is not used in the code. It is used in `CbsLink` to call `ICbsTokenProvider::GetTokenAsync(Uri namespaceAddress, string appliesTo, string[] requiredClaims)` as parameter `appliesTo`. It is ignore in SDK code. Please refer to `IotHubConnectionString` and `IotHubConnectionString.core`.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
davilu: If not, it will fail token refresh e2etest.

iothub/device/src/Transport/Amqp/AmqpLinkHelper.cs
davilu: just aligning with what master's doing.

e2e/test/DeviceTokenRefreshE2ETests.cs
davilu: done.

davilu:
Too many redundant work here, moving to 876
.

Issue #834, opened 2019-03-12
TaskScheduler.UnobservedTaskException being thrown from Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler


SwatantraYadav:
- **OS, version, SKU and CPU architecture used:**Windows 10 Pro  6.3.14393 x64 

- **Application's .NET Target Framework :** net471, net472, 4.7.02053 
- **Device:** Laptop 
- **SDK version used:** Microsoft.Azure.Devices.Client -Version 1.19.0 

## Description of the issue:

We are using DeviceClient to send and receive messages. Entire code to initialize device client, send messages is enclosed within a Try-Catch block and we are disposing the client in finally. However because the library is probably not catching an exception somewhere there is an UnobservedTaskException somewhere. 

This is how I am able to catch this exception:
TaskScheduler.UnobservedTaskException += (sender, eventArgs) =>
            {
                Log.Critical("An exceptiored in an unobserved Task.");
eventArgs.SetObserved();
            };
## Code sample exhibiting the issue:

Please see here: https://gist.github.com/SwatantraYadav/1fea89c6e763ec656832e67b8667c3d2

## Console log of the issue:

An exception occured in an unobserved Task.
Type: System.AggregateException
Message: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread.
HResult: -2146233088

---- Inner Exception #1-1 ----

Type: System.Net.Sockets.SocketException
ErrorCode: 10054
Message: An existing connection was forcibly closed by the remote host
NativeErrorCode: 10054
StackTrace:    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.<SubscribeAsync>d__49.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #885, opened 2019-03-29
Fix for Task unobserved exceptions

SwatantraYadav:
Always return Task from async method instead of void. This ensure that caller can await the Task and will not get an (unexpected) exception thrown by a method that returns void in the caller's Synchronization context.

This change is inline with the Asynchronous programming best practices given here: https://msdn.microsoft.com/en-us/magazine/jj991977.aspx
https://devblogs.microsoft.com/pfxteam/task-exception-handling-in-net-4-5/

In this PR there is no change to signature of any existing methods. Any method with following signature
async void MethodName()
will now internally call a new method
async Task MethodNameAsync

All such calls have ConfigureAwait(false) so that all tests pass.

## Checklist

https://github.com/Azure/azure-iot-sdk-csharp/issues/834


common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -96,6 +96,11 @@ protected override void DoDisconnect()
         }
 
         protected override async void DoClose()
+        {
+            await DoCloseAsync().ConfigureAwait(false);
```
CIPop: This won't change much: the async void isn't awaited and if DoClose throws it will still be unobserved.

iothub/device/src/IotHubConnection.cs
```diff
@@ -1,452 +1,326 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.Azure.Devices.Client
+namespace Microsoft.Azure.Devices
```
CIPop: I don't think this is correct: looks like a copy of the ServiceClient connection string parser?

iothub/device/src/Transport/Mqtt/ClientWebSocketChannel.cs
```diff
@@ -94,6 +94,11 @@ protected override void DoDisconnect()
         }
 
         protected override async void DoClose()
+        {
+            await DoCloseAsync().ConfigureAwait(false);
```
CIPop: Same as above for all of these: async void won't be observed - the async actions need to be awaited and the exceptions dispatched to one of the API stacks.

iothub/device/src/Transport/Mqtt/ClientWebSocketChannel.cs
```diff
@@ -175,6 +186,11 @@ protected override async void DoBeginRead()
         }
 
         protected override async void DoWrite(ChannelOutboundBuffer channelOutboundBuffer)
+        {
+            await DoWriteAsync(channelOutboundBuffer).ConfigureAwait(false);
```
CIPop: No difference here: just wrapped with another Task which adds overhead.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
SwatantraYadav: @CIPop I agree, and I think the issue cannot be fixed in this repository alone as the underlying library called DoNetty (https://github.com/Azure/DotNetty) needs to be fixed first. I took a crack at it but soon realized it is bigger task than i anticipated or can afford to spend time to fix. Basically the IChannel and all its hierarchy needs to be converted to return Task instead of void. That change should be done all the way up and ideally with name changes to methods such as Open to OpenAsync. I wish there was a short cut but it does look like it.

CIPop: @SwatantraYadav I agree - this is definitely non-trivial and requires a lot of investigations to ensure all tasks get properly awaited or at least results get observed. 
We work closely with the DotNetty team and we will raise this up for a future release.

CIPop:
Closing this for now: as @SwatantraYadav mention this is a complicated fix that would require rework of the Mqtt transport. 
At the moment we are actively trying to fix the AMQP transport to remove any such unobserved tasks. 1.20.0 still has at least one task not being properly awaited (the CBS token refresher) but we plan to design a garbage collector like structure where we can ensure that all tasks completed and all exceptions are observed and flowed to either the application or the log.

PR #878, opened 2019-03-28
Basic mux tests (with check that connection does not disconnect for basic scenarios)

abhipsaMisra:
Basic Scenarios:
- [x] Message Send E2E muxed
- [x] Message Receive E2E muxed
- [x] Methods E2E muxed
- [x] Twin E2E muxed 
- [x] Device client - all operations over muxed connection

There is a connection status check enabled to verify the muxing does not cause previously added devices to disconnect.

Open items:
- [ ] Message, method and twin common code is still in util classes.
- [ ] Modules operations - basic and combined operations.

PR #888, opened 2019-04-02
Fixing graceful MQTT disconnect.

CIPop:


CIPop:
Changes checked-in from #878.

PR #900, opened 2019-04-08
AMQP should allow null AMQP app properties just like MQTT and HTTP. Properties with null value should not be dropped.

aklintsevich:
AMQP

### DIFF
--- a/iothub/device/src/MessageConverter.cs
+++ b/iothub/device/src/MessageConverter.cs
@@ -222,10 +222,6 @@ namespace Microsoft.Azure.Devices.Client
         public static bool TryGetAmqpObjectFromNetObject(object netObject, MappingType mappingType, out object amqpObject)
         {
             amqpObject = null;
-            if (netObject == null)
-            {
-                return false;
-            }
 
             switch (SerializationUtilities.GetTypeId(netObject))
             {
@@ -288,6 +284,9 @@ namespace Microsoft.Azure.Devices.Client
                         amqpObject = new AmqpMap((IDictionary)netObject);
                     }
                     break;
+                case PropertyValueType.Null:
+                    amqpObject = string.Empty;
+                    break;
                 default:
                     break;
             }
@@ -298,10 +297,6 @@ namespace Microsoft.Azure.Devices.Client
         public static bool TryGetNetObjectFromAmqpObject(object amqpObject, MappingType mappingType, out object netObject)
         {
             netObject = null;
-            if (amqpObject == null)
-            {
-                return false;
-            }
 
             switch (SerializationUtilities.GetTypeId(amqpObject))
             {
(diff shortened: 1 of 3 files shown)


aklintsevich:
@vishal-gandhi @kapil-ms

aklintsevich:
@davilu @timtay-microsoft @CIPop @abhipsaMisra @jenoola Can someone review this pull request?

iothub/device/src/MessageConverter.cs
```diff
@@ -222,10 +222,6 @@ public static void UpdateAmqpMessageHeadersAndProperties(AmqpMessage amqpMessage
         public static bool TryGetAmqpObjectFromNetObject(object netObject, MappingType mappingType, out object amqpObject)
         {
             amqpObject = null;
-            if (netObject == null)
```
davilu: Shouldn't remove this check.

iothub/device/tests/MessageConverterTests.cs
```diff
@@ -44,6 +44,7 @@ public void UpdateMessageHeaderAndPropertiesTest()
             
                 amqpMessage.ApplicationProperties.Map["Prop1"] = "Value1";
                 amqpMessage.ApplicationProperties.Map["Prop2"] = "Value2";
+                amqpMessage.ApplicationProperties.Map["Prop3"] = null;
```
davilu: Shouldn't use `null` but `string.Empty`

aklintsevich: This is testing that a null value does not get dropped.

aklintsevich:
@davilu if MQTT and HTTP should be consistent with AMQP then this library should accept null values. We shouldn't be dropping app properties with null value. See image above.

iothub/device/tests/MessageConverterTests.cs
davilu: Shouldn't allow `null` value. Should use `string.Empty`.

davilu:
`null` value has no standard serialization. Once this type of data received, it won't be able to tell it was set to`null` or empty string.

aklintsevich:
If you do not want to allow null values, then you should not allow null values for MQTT and HTTP as well. If you would like to allow null values then you should allow them for all protocols rather than a subset.,

aklintsevich:
MQTT and HTTP null values get serialized to empty string

davilu:
Another question is it's not sure how `Amqp` library handle `null` value.

aklintsevich:
I don't know how Amqp library will handle null values. I only tested the serialization of null values with empty string, which is consistent behavior with MQTT and HTTP protocols. When serializing with empty string there were no issues found when running IotHub BVTs.

davilu:
We should try to fix HTTP/MQTT instead of AMQP. Pair with `null` value should be dropped. There are a lot of properties are ignore such as `Message.To`, `Message.ExpiryTimeUtc` etc. We shouldn't bring in new inconsistency.

CIPop:
> I don't know how Amqp library will handle null values.
> When serializing with empty string there were no issues found when running IotHub BVTs.

@aklintsevich please create an issue describing the problem you are trying to solve or add an E2E test for us to understand the issue.

aklintsevich:
https://github.com/Azure/azure-iot-sdk-csharp/issues/919

PR #924, opened 2019-04-25
Add deviceScope property to Device object (from preview to master)

abhipsaMisra:
## Checklist

## Description of the changes

Add deviceScopes property to Device object.
This field stores an identifier to other edge devices, with whom the device can connect and communicate with even in the absence of Azure IoT Hub connectivity.
https://azure.microsoft.com/en-us/blog/extended-offline-operation-with-azure-iot-edge/


varunpuranik:
@abhipsaMisra Looks good, but if you cherry pick the commit, we won't lose history of the change.

abhipsaMisra:
I cherry-picked the commits from Azure:preview; so we should be able to preserve the history.

varunpuranik:
Oh okay.. cool

e2e/test/RegistryManagerTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            Device receivedDevice = await registryManager.AddDeviceAsync(leafDevice).ConfigureAwait(false);
+
+            Assert.IsNotNull(receivedDevice);
+            Assert.AreEqual(leafDevice.Id, receivedDevice.Id);
+            Assert.AreEqual(leafDevice.Scope, receivedDevice.Scope);
+            await registryManager.RemoveDeviceAsync(leafDevice.Id).ConfigureAwait(false);
+            await registryManager.RemoveDeviceAsync(edgeDevice.Id).ConfigureAwait(false);
+        }
+    }
+}
```
CIPop: Nit: NL at EOF.

timtay-microsoft:
When I run this new e2e test, the security scope is always null. Is this the expected behavior for the test @varunpuranik 

I was expecting the scope field to have a non-null value at some point in this test

abhipsaMisra:
Good catch @timtay-microsoft . Confirmed with @varunpuranik that we need to update the api version to "2019-03-30"

CIPop:
@abhipsaMisra @varunpuranik Marking as blocked per @anthonyvercolano who got customer reports saying this doesn't work on a particular scale unit.
/cc @prmathur-microsoft @tameraw

PR #925, opened 2019-04-25
Add support for Retry-After header to provisioning device client

timtay-microsoft:
Http handled the header already, but added support to amqp and mqtt.

DPS has added a new header to DeviceRegistrationOperation responses that should dictate how long to wait before retrying. When this value is not provided, the SDK defaults to 2 seconds between operations.

### DIFF
--- a/provisioning/transport/amqp/src/ProvisioningErrorDetailsAmqp.cs
+++ b/provisioning/transport/amqp/src/ProvisioningErrorDetailsAmqp.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.Amqp;
+using Microsoft.Azure.Amqp.Framing;
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
+{
+    [SuppressMessage("Microsoft.Performance", "CA1812", Justification = "Is instantiated by json convertor")]
+    internal class ProvisioningErrorDetailsAmqp : ProvisioningErrorDetails
+    {
+        /// <summary>
+        /// The time to wait before trying again if this error is transient
+        /// </summary>
+        internal TimeSpan? RetryAfter { get; set; }
+
+        public const string RetryAfterKey = "Retry-After";
+
+        public static TimeSpan? GetRetryAfterFromApplicationProperties(AmqpMessage amqpResponse, TimeSpan defaultInterval)
+        {
+            object retryAfter;
+            if (amqpResponse.ApplicationProperties != null && amqpResponse.ApplicationProperties.Map.TryGetValue(RetryAfterKey, out retryAfter))
+            {
+                int secondsToWait;
+                if (int.TryParse(retryAfter.ToString(), out secondsToWait))
+                {
+                    TimeSpan serviceRecommendedDelay = TimeSpan.FromSeconds(secondsToWait);
+
+                    if (serviceRecommendedDelay.TotalSeconds < defaultInterval.TotalSeconds)
+                    {
+                        return defaultInterval;
+                    }
+                    else
+                    {
+                        return serviceRecommendedDelay;
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        public static TimeSpan? GetRetryAfterFromRejection(Rejected rejected, TimeSpan defaultInterval)
+        {
+            if (rejected.Error != null && rejected.Error.Info != null)
+            {
+                object retryAfter;
+                if (rejected.Error.Info.TryGetValue(RetryAfterKey, out retryAfter))
+                {
+                    int secondsToWait = 0;
+                    if (int.TryParse(retryAfter.ToString(), out secondsToWait))
+                    {
+                        if (secondsToWait < defaultInterval.Seconds)
+                        {
+                            return defaultInterval;
+                        }
+                        else
+                        {
+                            return TimeSpan.FromSeconds(secondsToWait);
+                        }
+                    }
+                }
+                
+            }
+
+            return null;
+        }
+    }
+}
--- a/provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
+++ b/provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
     /// </summary>
     public class ProvisioningTransportHandlerAmqp : ProvisioningTransportHandler
     {
-        private static readonly TimeSpan DefaultOperationPoolingIntervalMilliseconds = TimeSpan.FromSeconds(2);
+        private static readonly TimeSpan DefaultOperationPoolingInterval = TimeSpan.FromSeconds(2);
         private static readonly TimeSpan TimeoutConstant = TimeSpan.FromMinutes(1);
 
         /// <summary>
@@ -123,15 +123,22 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
 
                     await Task.Delay(
                         operation.RetryAfter ??
-                        DefaultOperationPoolingIntervalMilliseconds).ConfigureAwait(false);
+                        RetryJitter.GenerateDelayWithJitterForRetry(DefaultOperationPoolingInterval)).ConfigureAwait(false);
 
                     cancellationToken.ThrowIfCancellationRequested();
 
-                    operation = await OperationStatusLookupAsync(
+                    try
+                    {
+                        operation = await OperationStatusLookupAsync(
                         connection,
                         operationId,
                         correlationId).ConfigureAwait(false);
-
+                    }
+                    catch (ProvisioningTransportException e) when (e.ErrorDetails is ProvisioningErrorDetailsAmqp &&  e.IsTransient)
+                    {
+                        operation.RetryAfter = ((ProvisioningErrorDetailsAmqp) e.ErrorDetails).RetryAfter;
+                    }
+                    
                     attempts++;
                 }
 
@@ -207,7 +214,9 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
             client.AmqpSession.ReceivingLink.AcceptMessage(amqpResponse);
             string jsonResponse = await new StreamReader(amqpResponse.BodyStream).ReadToEndAsync()
                 .ConfigureAwait(false);
-            return JsonConvert.DeserializeObject<RegistrationOperationStatus>(jsonResponse);
+            RegistrationOperationStatus status = JsonConvert.DeserializeObject<RegistrationOperationStatus>(jsonResponse);
+            status.RetryAfter = ProvisioningErrorDetailsAmqp.GetRetryAfterFromApplicationProperties(amqpResponse, DefaultOperationPoolingInterval);
+            return status;
         }
 
         private async Task<RegistrationOperationStatus> OperationStatusLookupAsync(
@@ -229,7 +238,9 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
             client.AmqpSession.ReceivingLink.AcceptMessage(amqpResponse);
             string jsonResponse = await new StreamReader(amqpResponse.BodyStream).ReadToEndAsync()
                 .ConfigureAwait(false);
-            return JsonConvert.DeserializeObject<RegistrationOperationStatus>(jsonResponse);
+            RegistrationOperationStatus status = JsonConvert.DeserializeObject<RegistrationOperationStatus>(jsonResponse);
+            status.RetryAfter = ProvisioningErrorDetailsAmqp.GetRetryAfterFromApplicationProperties(amqpResponse, DefaultOperationPoolingInterval);
+            return status;
         }
 
         private static DeviceRegistrationResult ConvertToProvisioningRegistrationResult(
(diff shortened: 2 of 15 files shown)


provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -326,6 +333,29 @@ private async Task VerifyPublishPacketTopicAsync(IChannelHandlerContext context,
             }
         }
 
+        private static TimeSpan? getRetryAfterFromTopic(string topic)
```
CIPop: Naming - please check our coding standards. (`GetRetry..`)

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -326,6 +333,29 @@ private async Task VerifyPublishPacketTopicAsync(IChannelHandlerContext context,
...
+        {
+            string[] topicAndQueryString = topic.Split('?');
```
CIPop: This is complex parsing - please add a UT.

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -326,6 +333,29 @@ private async Task VerifyPublishPacketTopicAsync(IChannelHandlerContext context,
...
+                        int secondsToWait;
+                        if (int.TryParse(queryKeyAndValue[1], out secondsToWait))
+                        {
+                            return TimeSpan.FromSeconds(secondsToWait);
+                        }
+                    }
+                }
+            }
+
+            return null;
```
CIPop: I think we should throw if invalid data is sent from the service and crash the application.
(If the topic _does_ have the data then it _must_ be correct. Otherwise we can ignore.) /cc @alextolp

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -344,11 +374,12 @@ private async Task ProcessRegistrationStatusAsync(IChannelHandlerContext context
                 //"{\"operationId\":\"0.indcertdevice1.e50c0fa7-8b9b-4b3d-8374-02d71377886f\",\"status\":\"assigning\"}"
                 var operation = JsonConvert.DeserializeObject<RegistrationOperationStatus>(jsonData);
                 operationId = operation.OperationId;
+                operation.RetryAfter = getRetryAfterFromTopic(packet.TopicName);
 
                 if (string.CompareOrdinal(operation.Status, RegistrationOperationStatus.OperationStatusAssigning) == 0 ||
                     string.CompareOrdinal(operation.Status, RegistrationOperationStatus.OperationStatusUnassigned) == 0)
                 {
-                    await Task.Delay(s_defaultOperationPoolingIntervalMilliseconds).ConfigureAwait(true);
+                    await Task.Delay(operation.RetryAfter ?? s_defaultOperationPoolingIntervalMilliseconds).ConfigureAwait(true);
```
CIPop: This is dangerous:
1. Please add a check to ensure a minimum retry time. That will avoid high-CPU spinning if the RetryAfter is too short due to bugs.
2. Check if RetryAfter is negative.

For violations of above rules we need to log a Warn message.

CIPop: Also, we should add some jitter if RetryAfter is not present. (We do see this as a bug in the minimum time implementation in DeviceClient.)

provisioning/device/src/ProvisioningErrorDetails.cs
```diff
@@ -66,5 +67,10 @@ public string CreateMessage(string message)
 
             return sb.ToString();
         }
+
+        /// <summary>
+        /// The time to wait before trying again if this error is transient
+        /// </summary>
+        public TimeSpan? RetryAfter { get; set; }
```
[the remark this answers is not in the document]
timtay-microsoft: https://github.com/Azure/iot-sdks-internals/pull/82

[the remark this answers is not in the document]
timtay-microsoft: I created a new PR in the internals repo to review this new API. Please take a look

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: Why would we need to validate a value provided by the service here? If the service does not provide a retryafter value, we assume a 2 second delay. The service should never provide a value to causes high-CPU spinning

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: Sure thing

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: Sure thing

provisioning/device/src/ProvisioningErrorDetails.cs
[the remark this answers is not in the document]
prmathur-microsoft: Why do you need additional public api

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: This method is buried deep enough into internal classes and methods that unit testing it without mocks doesn't seem feasible. Would you be okay with using reflection to isolate this method for unit testing?

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: Sure we can validate the value provided by service, and add some jitter to the default retry

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: I don't see a reason to drop the whole connection simply because the retry-after field is off. If we just fall back to the default retry operation, the SDK will handle this scenario just fine.

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: I factored out this logic to the ProvisioningErrorDetails implementations for each protocol in order to make this logic unit testable

provisioning/device/src/ProvisioningErrorDetails.cs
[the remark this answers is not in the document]
timtay-microsoft: It isn't needed, I'll remove it

provisioning/transport/amqp/src/ProvisioningErrorDetailsAmqp.cs
```diff
@@ -0,0 +1,70 @@
+﻿using Microsoft.Azure.Amqp;
```
CIPop: Missing header.

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
```diff
@@ -0,0 +1,149 @@
+﻿using Microsoft.Azure.Amqp;
```
CIPop: Missing header.

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        private static TimeSpan defaultInterval = TimeSpan.FromSeconds(2);
+
+        [TestMethod]
+        public void getRetryFromRejectedSuccess()
+        {
+            int expectedSeconds = 32;
+            Rejected rejected = new Rejected();
+            rejected.Error = new Error();
+            rejected.Error.Info = new Fields();
+            rejected.Error.Info.Add(new Azure.Amqp.Encoding.AmqpSymbol("Retry-After"), expectedSeconds);
```
CIPop: missing `using` statement instead of the fully qualified name here.

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
```diff
@@ -0,0 +1,149 @@
...
+namespace Microsoft.Azure.Devices.Provisioning.Transport.Amqp.UnitTests
+{
+    [TestClass]
+    [TestCategory("Unit")]
+    public class ProvisioningErrorDetailsAmqpTests
+    {
+        private static TimeSpan defaultInterval = TimeSpan.FromSeconds(2);
+
+        [TestMethod]
+        public void getRetryFromRejectedSuccess()
```
CIPop: Please follow our naming convention for tests <Unit>_<Scenario>_<ExpectedResult>

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
CIPop: Also, names begin with caps.

provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -286,5 +302,15 @@ private void ValidateOutcome(Outcome outcome)
                 }
             }
         }
+
+        private static int jitterMax = 5;
+        private static int jitterMin = 0;
+        private static TimeSpan GenerateDelayWithJitterForRetry(TimeSpan defaultDelay)
```
CIPop: Please add a UT to test this.

provisioning/transport/http/src/ProvisioningErrorDetailsHttp.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
```
CIPop: Here and other places: please ensure that you have added the proper file headers.

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -219,5 +223,15 @@ private static DeviceRegistrationResult ConvertToProvisioningRegistrationResult(
                 result.Etag,
                 result?.Payload?.ToString(CultureInfo.InvariantCulture));
         }
+
+        private static int jitterMax = 5;
+        private static int jitterMin = 0;
+        private static TimeSpan GenerateDelayWithJitterForRetry(TimeSpan defaultDelay)
```
CIPop: Duplicated code. Please remove/move to common.

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -533,5 +541,15 @@ private ushort GetCurrentPacketId()
                 return (ushort)Volatile.Read(ref _packetId);
             }
         }
+
+        private static int jitterMax = 5;
+        private static int jitterMin = 0;
+        private static TimeSpan GenerateDelayWithJitterForRetry(TimeSpan defaultDelay)
```
CIPop: Same as above: please keep all common code in `common\src`

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
timtay-microsoft: sure thing

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
timtay-microsoft: sure thing

provisioning/transport/amqp/src/ProvisioningErrorDetailsAmqp.cs
timtay-microsoft: good catch

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
timtay-microsoft: I'll make them start with capital letters, sure

provisioning/transport/http/src/ProvisioningErrorDetailsHttp.cs
timtay-microsoft: Sure thing

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
timtay-microsoft: Sure thing

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: Sure thing

provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
timtay-microsoft: I can add a test to check that a single call for this random jitter falls within the expected bounds, but seeing as how the jitter relies on random number generation, I'm not sure how useful it will be.

CIPop: It's useful to check if it falls in a range and it's random :).

provisioning/transport/amqp/tests/RetryJitterTests.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+
+namespace Microsoft.Azure.Devices.Provisioning.Transport.Amqp.UnitTests
+{
+    [TestClass]
+    [TestCategory("Unit")]
+    public class RetryJitterTests
+    {
+        [TestMethod]
+        public void RetryJitterGeneratedDelayLargerOrEqualToDefaultDelay()
```
CIPop: The naming convention is actually <Unit>_<Scenario>_<Expectation>. This way you can read it as a story: `RetryJitter_DefaultDelay_WithinRange`.

PR #930, opened 2019-05-01
added missing AMQP token refresh cleanup on device client close or di…

davilu:
…spose

## Checklist


davilu:
It's internal logic and it's really difficult to write unit test since it relies on AMQP connection, CBS link. One thing we can sure is in the existing code that token refresher is not disposed when AMQP transport is stopped or even disposed.

PR #932, opened 2019-05-02
Adding single-device stress/perf test

CIPop:
- Updating stress test sln.
- Adding single device stress test.

### DIFF
--- a/e2e/stress/DeviceTelemetryTest/DeviceTelemetryMetrics.cs
+++ b/e2e/stress/DeviceTelemetryTest/DeviceTelemetryMetrics.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    public struct DeviceTelemetryMetrics
+    {
+        public double? WallTime;
+        public string Protocol;
+        public int? MessageSize;
+        public double? SendMs;
+        public double? AckMs;
+        public string ErrorMessage;
+
+        public static string GetHeader()
+        {
+            return
+                "WallTime," +
+                "Protocol, " +
+                "MessageSize," +
+                "SendMs, " +
+                "AckMs, " +
+                "ErrorMessage, ";
+        }
+
+        public override string ToString()
+        {
+            var sb = new StringBuilder(); 
+            Add(sb, WallTime);
+            Add(sb, Protocol);
+            Add(sb, MessageSize);
+            Add(sb, SendMs);
+            Add(sb, AckMs);
+            Add(sb, ErrorMessage);
+
+            return sb.ToString();
+        }
+
+        private void Add(StringBuilder sb, object data)
+        {
+            if (data != null)
+            {
+                sb.Append(data.ToString());
+            }
+
+            sb.Append(',');
+        }
+    }
+}
--- a/e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
+++ b/e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
@@ -1,0 +1,131 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.Devices.Client;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    public class DeviceTelemetryTest
+    {
+        int _messageSize;
+        int _messageCount;
+        int _parallelCount;
+        TransportType _protocol;
+        ResultWriter _writer;
+        DeviceClient _client;
+        Stopwatch _wall = new Stopwatch();
+
+        public DeviceTelemetryTest(
+            TransportType protocol,
+            int messageCount,
+            int messageSize,
+            int parallelCount,
+            ResultWriter resultWriter)
+        {
+            _messageCount = messageCount;
+            _messageSize = messageSize;
+            _protocol = protocol;
+            _parallelCount = parallelCount;
+            _writer = resultWriter;
+
+            if (_messageCount < _parallelCount) throw new ArgumentException($"{nameof(messageCount)} needs to be greater than {nameof(parallelCount)}");
+        }
+
+        public void CreateDevice()
+        {
+            string connectionString =
+                Environment.ExpandEnvironmentVariables(Environment.GetEnvironmentVariable("IOTHUB_DEVICE_CONN_STRING"));
+
+            _client = DeviceClient.CreateFromConnectionString(connectionString, _protocol);
+
+            _wall.Start();
+        }
+
+        public async Task SendMessageAsync()
+        {
+            var metrics = new DeviceTelemetryMetrics();
+            metrics.MessageSize = _messageSize;
+            metrics.Protocol = _protocol.ToString();
+            metrics.WallTime = _wall.Elapsed.TotalMilliseconds;
+
+            try
+            {
+                Stopwatch watch = new Stopwatch();
+                watch.Restart();
+                Message message = new Message(new byte[_messageSize]);
+                Task sendTask = _client.SendEventAsync(message);
+                watch.Stop();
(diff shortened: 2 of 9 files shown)


CIPop:
@prmathur-microsoft @abhipsaMisra  Note - this doesn't impact build at this time. 
1. We need to keep the build -stresstests up-to-date but that requires minor modifications of the tests to return non-zero on failure (which could be caused by high CPU, failing KPIs, etc).
2. We need to define a DevOps pipeline to run stress tests (these take considerably more and are highly dependent on network and VM SKU).

@jebrando - please let me know if we have a public common output spec for perf. We would like to adapt this as well as a future multi-device test to output the same CSV (or JSON).

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
```diff
@@ -0,0 +1,131 @@
...
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    public class DeviceTelemetryTest
+    {
+        int _messageSize;
+        int _messageCount;
+        int _parallelCount;
+        TransportType _protocol;
+        ResultWriter _writer;
+        DeviceClient _client;
```
davilu: `_client` could be local variable in `RunTestAsync`

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            _messageCount = messageCount;
+            _messageSize = messageSize;
+            _protocol = protocol;
+            _parallelCount = parallelCount;
+            _writer = resultWriter;
+
+            if (_messageCount < _parallelCount) throw new ArgumentException($"{nameof(messageCount)} needs to be greater than {nameof(parallelCount)}");
+        }
+
+        public void CreateDevice()
```
davilu: `private` and return `DeviceClient`

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            metrics.WallTime = _wall.Elapsed.TotalMilliseconds;
+
+            try
+            {
+                Stopwatch watch = new Stopwatch();
+                watch.Restart();
+                Message message = new Message(new byte[_messageSize]);
+                Task sendTask = _client.SendEventAsync(message);
+                watch.Stop();
+                metrics.SendMs = watch.Elapsed.TotalMilliseconds;
```
davilu: It's not `sendMs` but `task` was created. We're not able to get `sendMs`  in existing implementation.

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
```diff
@@ -0,0 +1,131 @@
...
+
+                watch.Restart();
+                await sendTask.ConfigureAwait(false);
+                watch.Stop();
+                metrics.AckMs = watch.Elapsed.TotalMilliseconds;
```
davilu: This is time cost of sending and receiving ack.

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
```diff
@@ -0,0 +1,131 @@
...
+
+
+            var sw = new Stopwatch();
+            sw.Start();
+            double startTimeSeconds = _wall.Elapsed.TotalSeconds;
+            int interimStatCompleted = 0;
+
+            for (int i = 0; i < _parallelCount; i++)
+            {
+                tasks.Add(SendMessageAsync());
```
davilu: In the existing stress test metrics, it took about 150 seconds to initialize 50K tasks. Start `_parallelCount` tasks here is not test `DeviceClient` performance but more like dotnet task scheduler.

e2e/stress/DeviceTelemetryTest/DeviceTelemetryMetrics.cs
```diff
@@ -0,0 +1,51 @@
...
+            return
+                "WallTime," +
+                "Protocol, " +
+                "MessageSize," +
+                "SendMs, " +
+                "AckMs, " +
+                "ErrorMessage, ";
+        }
+
+        public override string ToString()
```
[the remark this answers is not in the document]
CIPop: The only concern would be the `null` check (not sure if that translates into "(null)" or something else).
In this case (no formatting) it appears that Join might be faster.

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    public class DeviceTelemetryTest
+    {
+        int _messageSize;
```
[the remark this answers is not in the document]
CIPop: 👍

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
CIPop: In C# is preferred to init in declaration. In dotnet/corefx we were following something similar to this: https://stackoverflow.com/questions/24551/initialize-class-fields-in-constructor-or-at-declaration (can't find the doc on dotnet/corefx).

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
```diff
@@ -0,0 +1,131 @@
...
+        int _messageCount;
+        int _parallelCount;
+        TransportType _protocol;
+        ResultWriter _writer;
+        DeviceClient _client;
+        Stopwatch _wall = new Stopwatch();
```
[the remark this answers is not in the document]
CIPop: In C# is preferred to init in declaration. In dotnet/corefx we were following something similar to this: https://stackoverflow.com/questions/24551/initialize-class-fields-in-constructor-or-at-declaration (can't find the doc on dotnet/corefx).

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
CIPop: 👍

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
CIPop: SendMs measures the time (overheads) it takes to execute the `SendEventAsync` method.
Not sure what you mean by not being able to get sendMs (you mean until we write to the wire?) End-to-end send is actually measured as SendMs + AckMs.

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
CIPop: I don't understand. Please clarify.
`_parallelCount` is one of the meta-parameters that need tuning based on the machine specification. In general it's small. I don't think we actually need 1000 here (I think ~100 will get the same result).

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
davilu:
```
public async Task RunTestAsync()
{
    DeviceClient deviceClient = CreateDevice();
    ...
    await deviceClient.OpenAsync(timeout).ConfigAwait(false);
    ...
    for (int i = 0; i < _parallelCount; i++)
    {
        tasks.Add(SendMessageAsync(deviceClient));
    }
    ....
```

e2e/stress/DeviceTelemetryTest/DeviceTelemetryMetrics.cs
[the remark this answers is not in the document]
CIPop: I take that back: these are double and int, etc - join only works on a string array.

e2e/stress/DeviceTelemetryTest/DeviceTelemetryTest.cs
CIPop: I don't understand why I would need to pass in the deviceClient as a parameter to member functions when things are encapsulated in a single class.

CIPop:
Closing for now: I have moved this into a more generic framework.

PR #937, opened 2019-05-07
fix of NPE

davilu:
## Checklist


iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
```diff
@@ -83,22 +89,20 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
 
         private void CreateNewTransportIfNotReady()
         {
-            if (InnerHandler == null || !InnerHandler.IsUsable)
+            if (!(InnerHandler?.IsUsable ?? false))
```
CIPop: Nit: please change back.

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
```diff
@@ -68,6 +69,11 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
                 {
                     CreateNewTransportIfNotReady();
                     await base.OpenAsync(cancellationToken).ConfigureAwait(false);
+                    if (_disposed)
```
CIPop: This isn't interlocked so it won't work: Dispose will change the flag on its thread but it won't change value here.
We could use an interlocked / volatile or a lock although I would like to avoid locks in Dispose.

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
```diff
@@ -83,22 +89,20 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
-            {
-                InnerHandler.Dispose();
-                InnerHandler = null;
-            }
+            IDelegatingHandler innerHandler = InnerHandler;
 
             // Ask the ContinuationFactory to attach the proper handler given the Context's ITransportSettings.
             InnerHandler = ContinuationFactory(Context, null);
+
+            innerHandler?.Dispose();
```
CIPop: Please add a comment explaining why CreateNewTransportHandler() during dispose is OK in `WaitForTransportClosedAsync`.

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
CIPop: Please add a comment on why we're doing this (to allow safe Dispose).

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -992,7 +992,7 @@ bool TryStateTransition(TransportState fromState, TransportState toState)
             return (TransportState)Interlocked.CompareExchange(ref this.state, (int)toState, (int)fromState) == fromState;
         }
 
-        void EnsureValidState()
+        private void EnsureValidState(bool ensureOpen)
```
CIPop: Nit: EnsureValidState(bool throwIfNotOpened=true)

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -166,7 +166,7 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
 
                 cancellationToken.ThrowIfCancellationRequested();
 
-                this.EnsureValidState();
+                this.EnsureValidState(false);
```
CIPop: Nit `throwIfNotOpen=false`.

CIPop:
For the DE failure we need to update vsts.yaml:376 from `timeoutInMinutes: 5` to maybe 15min?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
davilu: done

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
davilu: done

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
davilu: done

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
davilu: done

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
davilu: done

PR #938, opened 2019-05-09
Implement connection status check with statistical results from multiple runs.

abhipsaMisra:
## Checklist

## Description of the changes

- Implement connection status check with results from multiple runs for the test (to account for network instability issues).
- Await on baseline/ test operation for all devices at once, instead of individually.

### DIFF
--- a/e2e/test/CombinedClientOperationsMultiplexingOverAmqpTests.cs
+++ b/e2e/test/CombinedClientOperationsMultiplexingOverAmqpTests.cs
@@ -115,7 +115,6 @@ namespace Microsoft.Azure.Devices.E2ETests
             ServiceClient serviceClient = ServiceClient.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
 
             // Message payload for C2D operation
-            string payload, messageId, p1Value;
             Dictionary<string, List<string>> messagesSent = new Dictionary<string, List<string>>();
 
             // Twin properties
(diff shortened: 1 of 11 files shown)


e2e/test/CombinedClientOperationsMultiplexingOverAmqpTests.cs
```diff
@@ -170,8 +160,8 @@ ConnectionStringAuthScope authScope
                 // C2D Operation
                 _log.WriteLine($"{nameof(CombinedClientOperationsMultiplexingOverAmqpTests)}: Operation 2: Receive C2D for device={testDevice.Id}");
                 List<string> msgSent = messagesSent[testDevice.Id];
-                payload = msgSent[0];
-                p1Value = msgSent[1];
+                var payload = msgSent[0];
```
CIPop: Nit: `string` instead of `var`.

e2e/test/MultiplexingOverAmqp.cs
```diff
@@ -15,6 +15,8 @@ public static class MultiplexingOverAmqp
         public const int MuxWithoutPoolingPoolSize = 1; // For enabling multiplexing without pooling, the pool size needs to be set to 1
         public const int MuxWithPoolingDevicesCount = 4;
         public const int MuxWithPoolingPoolSize = 2;
+        public const int maxTestRunCount = 5;
+        public const float testSuccessRate = 0.8f; // 4 out of 5 test runs should pass (even after accounting for network instability issues).
```
CIPop: Nit: 
`We expect network issues during test execution [ ... ] `.

Please explain the reason we need to do this kind of strict assertion that is sensitive to noise (network failing during test execution).

PR #957, opened 2019-06-04
Fix to close device and service WS clients (SGW clients) in parallel in Device Streaming e2e tests

ewertons:
…in Device Streaming e2e tests

The Streaming Gateway was handling closure of websocket clients
incorrectly, waiting individually for the service-side client to close
first, then waiting for the device client-side to close, in this order.
Dmitry provided a fix that now makes it wait in parallel for both sides to close,
which is correct since there is no way to predict who will close the connection first.

The same fix had to be applied to the Device Streaming E2E tests as it
had similar behavior.

## Checklist

### DIFF
--- a/e2e/test/DeviceStreamingE2ETests.cs
+++ b/e2e/test/DeviceStreamingE2ETests.cs
@@ -415,8 +415,10 @@ namespace Microsoft.Azure.Devices.E2ETests
                 }, TaskScheduler.Current)
             ).ConfigureAwait(false);
 
-            await deviceWSClient.CloseAsync(WebSocketCloseStatus.NormalClosure, "End of test", cts.Token).ConfigureAwait(false);
-            await serviceWSClient.CloseAsync(WebSocketCloseStatus.NormalClosure, "End of test", cts.Token).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


e2e/test/DeviceStreamingE2ETests.cs
```diff
@@ -415,8 +415,10 @@ private async Task TestEchoThroughStreamingGatewayAsync(ClientDeviceStreamingReq
                 }, TaskScheduler.Current)
             ).ConfigureAwait(false);
 
-            await deviceWSClient.CloseAsync(WebSocketCloseStatus.NormalClosure, "End of test", cts.Token).ConfigureAwait(false);
-            await serviceWSClient.CloseAsync(WebSocketCloseStatus.NormalClosure, "End of test", cts.Token).ConfigureAwait(false);
+            Task.WaitAll(
```
CIPop: Mixing Wait and Await is extremely dangerous and can cause deadlocks.
Use `await Task.WhenAll(...).ConfigureAwait(false);` instead.

ewertons: Thanks Cris. Shall be changed.

Issue #764, opened 2019-01-18
Intermittent test failure for KillTcp and MqttWs fault injection tests


abhipsaMisra:
Intermittent test failure for fault injection tests having a combination of fault type `KillTcp` and transport type `MqttWs`. Disabled them for unblocking CI.

```
Message_TcpConnectionLossReceiveRecovery_MqttWs
Message_TcpConnectionLossSendRecovery_MqttWs
Method_DeviceReceivesMethodAndResponseRecovery_MqttWs
Twin_DeviceReportedPropertiesTcpConnRecovery_MqttWs
Twin_DeviceDesiredPropertyUpdateTcpConnRecovery_MqttWs
```

Logs: 
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=8752&view=logs
https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=8762&view=logs

abhipsaMisra:
```
2019-01-18T22:00:16.6045705Z  2019-01-18T22:00:16.5397360 [Microsoft-Azure-Devices-TestLogging-TestMessage] (ActivateFaultInjection: Fault injection requested.).
2019-01-18T22:00:16.6046398Z  2019-01-18T22:00:16.5398145 [Microsoft-Azure-Devices-TestLogging-TestMessage] (FaultInjection: Waiting for fault injection to be active: 0ms).
2019-01-18T22:00:16.6047081Z  2019-01-18T22:00:16.5399394 [Microsoft-Azure-Devices-Device-Client-Enter] (RetryDelegatingHandler#55649057, SendTwinPatchAsync, ({
2019-01-18T22:00:16.6047646Z    "37dd86e1-fb53-4224-a992-22e85f75b35c": "196f5455-98f9-4705-afdf-c93ef387afc5"
2019-01-18T22:00:16.6047873Z  }, CancellationToken#1324048)).
2019-01-18T22:00:16.6048577Z  2019-01-18T22:00:16.5399801 [Microsoft-Azure-Devices-Device-Client-Enter] (ErrorDelegatingHandler#36744480, ExecuteWithErrorHandlingAsync, ErrorDelegatingHandler.ExecuteWithErrorHandlingAsync).
[...]

PR #962, opened 2019-06-06
Add null check for deviceId and moduleId

### LINKED ISSUE #883: Calling InvokeDeviceMethodAsync with moduleId set to null fails in obscure way
Microsoft.Azure.Devices v1.17.2
Calling an overload of `InvokeDeviceMethodAsync` that has moduleId and passing null to it results in SDK letting it through to service with incorrect URL and as a result user sees:
```
The requested resource does not support http method 'POST'
```
It may be worth checking moduleId for != null and failing fast with ArgumentNullException.

stefanzueger:
## Checklist

Fixes #883


iothub/service/src/AmqpServiceClient.cs
```diff
@@ -235,6 +235,16 @@ public override Task<CloudToDeviceMethodResult> InvokeDeviceMethodAsync(string d
 
         public override Task<CloudToDeviceMethodResult> InvokeDeviceMethodAsync(string deviceId, string moduleId, CloudToDeviceMethod cloudToDeviceMethod, CancellationToken cancellationToken)
         {
+            if (string.IsNullOrWhiteSpace(deviceId))
+            {
+                throw new ArgumentException("Value should be non null and non empty", nameof(deviceId));
```
CIPop: Please switch to `ArgumentNullException` and omit the text message keeping just the argument. That allows the dotnet stack to choose the proper text string in any language and keeps the resulting binary smaller.

stefanzueger: I thought, I'll do it consistent to the other methods within the same class, which also throw `ArgumentException`. But I agree it is better to throw `ArgumentNullExcption`, so I changed it also in the other methods in the same class.
While scanning through the class I've found two places which do not specify ConfigureAwait, which I also fixed now.

stefanzueger:
@CIPop one build is failing because of an E2E test that reports that the token refresh took more than the allowed ttl of 6s. I assume this is a flaky test and has nothing to do with my changes.

CIPop:
Failed with #764. Rerunning.

CIPop:
Known server-side errors in Canary.

CIPop:
Thank you for the contribution @stefanzueger !

PR #964, opened 2019-06-10
Adding stress / perf test 

CIPop:
Adding stress/perf test and unifying previous versions of these tests.
Removing the jenkins folder which is no longer in use.


e2e/stress/IoTClientPerf/readme.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Prerequisites
+
+##### Create an IoT Hub for performance testing.
+##### Create one IoT Hub device with SAS token and one with X509 certificate authentication.
+
+Fill in the following environment variables (see the e2e\test prerequisites for details):
+```
+IOTHUB_CONN_STRING_CSHARP=
+IOTHUB_DEVICE_CONN_STRING=
```
abhipsaMisra: I had removed these (IOTHUB_DEVICE_CONN_STRING, IOTHUB_DEVICE_CONN_STRING2) from our pre-requisites, since the E2E tests were no longer referencing them. We will need to add them back in if they are needed by the stress test.

e2e/stress/IoTClientPerf/readme.md
```diff
@@ -0,0 +1,58 @@
+# Azure IoT Hub Stress Test
+
+Use `dotnet run` to display the help page for this tool.
```
abhipsaMisra: FYI: I got the following error on running dotnet run from cmd:
`Reporting\ResultWriterFile.cs(24,89): error CS1738: Named argument specifications must appear after all fixed arguments have been specified. Please use language version 7.2 or greater to allow non-trailing named arguments.`
It ran after I modified the Language version in VS 2017 to use latest minor version.

e2e/stress/IoTClientPerf/readme.md
CIPop: These are needed for samples as well. We need them back.

e2e/stress/IoTClientPerf/readme.md
CIPop: Thanks! I think I'm on latest and using VS2019 which means it'll work for me w/o the warning.
I'll remove this to allow older .NET versions.

CIPop:
@abhipsaMisra one of the new tests failed:

`Microsoft.Azure.Devices.E2ETests.MessageSendE2EPoolAmqpTests.Message_DeviceSak_DeviceSendSingleMessage_MultipleConnections_Amqp`
Test method Microsoft.Azure.Devices.E2ETests.MessageSendE2EPoolAmqpTests.Message_DeviceSak_DeviceSendSingleMessage_MultipleConnections_Amqp threw exception: 
System.Threading.Tasks.TaskCanceledException: A task was canceled.

PTAL

e2e/stress/IoTClientPerf/readme.md
```diff
@@ -0,0 +1,58 @@
...
+IOTHUB_DEVICE_CONN_STRING=
+IOTHUB_DEVICE_CONN_STRING2=
+IOTHUB_X509_PFX_CERTIFICATE=
+```
+
+##### Provision sufficient devices for the hub:
+
+`dotnet run -- -a sas -n 100 -o out.csv -f generate_iothub_config`
+
+Follow the instructions in the output for how to import the `devices.txt` file into your hub.
```
abhipsaMisra: Note: You will need the [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest) and the [IoT extension for Azure CLI](https://github.com/Azure/azure-iot-cli-extension#installation).

e2e/stress/IoTClientPerf/Scenarios/GenerateIotHubConfigTest.cs
```diff
@@ -0,0 +1,87 @@
...
+        // - X509 key devices generated will have the same certificate as specified in IOTHUB_X509_PFX_CERTIFICATE
+        private static StreamWriter s_outputFile = new StreamWriter("devices.txt");
+        private static SemaphoreSlim s_semaphore = new SemaphoreSlim(1);
+
+        static GenerateIotHubConfigTest()
+        {
+            Console.WriteLine("----------------------");
+            Console.WriteLine("To import the generated file:");
+            Console.WriteLine("1. Create an Azure Storage Container");
+            Console.WriteLine("2. Get a container SAS key:");
```
abhipsaMisra: Reverse the order of step 2 and 3 - so that we generate the container SAS key, and then use it immediately after.

common/test/Configuration.IoTHub.cs
```diff
@@ -27,6 +27,8 @@ public static X509Certificate2 GetCertificateWithPrivateKey()
 
             public static string DeviceConnectionString => GetValue("IOTHUB_DEVICE_CONN_STRING");
 
+            public static string DeviceConnectionString2 => GetValue("IOTHUB_DEVICE_CONN_STRING2");
```
abhipsaMisra: We should set the default value as Empty (since these are used only by samples and perf tests).

CIPop: I need this to be set in order to run the tests. If we set the default value, it could create issues when the perf test runs.
No exceptions are going to be thrown until it's used. E2E tests don't use it yet.

CIPop:
Unrelated file notification tests failed.

PR #965, opened 2019-06-12
Fix to close device and service WS clients (SGW clients) in parallel in Device Streaming e2e tests

ewertons:
## Checklist

This is a change in preview branch only.

## Description of the changes

(This is being ported from DeviceStreaming branch, commit# 2d36a276213e15f1909bbd5bed31a12c62762dd4)

The Streaming Gateway was handling closure of websocket clients
incorrectly, waiting individually for the service-side client to close
first, then waiting for the device client-side to close, in this order.
Dmitry provided a fix that now makes it wait in parallel for both sides to close,
which is correct since there is no way to predict who will close the connection first.

The same fix had to be applied to the Device Streaming E2E tests as it
had similar behavior.


CIPop:
Obsoleted by #971.

Issue #969, opened 2019-06-14
High CPU load if Disconnected_Retrying status


SergiiChurakov:
- **OS, version, SKU and CPU architecture used:** <VERSION>  Windows 10 Pro x64
- **Application's .NET Target Framework :** <NETVERSION> netcoreapp2.1, net472
- **Device:** <DEVICETYPE> Laptop
- **SDK version used:** <VERSION> 1.20.2

## Description of the issue:
Connect to the device with your credentials and register any direct method. Disconnect from the network. After 2-4 minutes, the application begins to heavily load the CPU.

This is because:
- Default retry strategy in RetryDelegatingHandler class created ExponentialBackoff with retryCount parameter = int.MaxValue.
- When dicsonnect happen ExponentialBackoff.GetShouldRetry() method after 20-25 attempts retrun negative value in retryInterval parameter.
- In AsyncExecution<TResult>.ExecuteAsyncContinueWith(Task<TResult> runningTask) method, if retryInterval parameter is negative, set TimeSpan.Zero. Thus, there is no interval between attempts.

## Code sample exhibiting the issue:
       public class IoTHubClient
       {
        private DeviceClient deviceClient;

        public async Task Connect(string deviceId, string connectionString, string hostname)
        {
            string key;

            using (RegistryManager registryManager = RegistryManager.CreateFromConnectionString(connectionString))
            {
                Device device = await registryManager.GetDeviceAsync(deviceId).ConfigureAwait(false);
                if (device == null)
                {
                    device = await registryManager.AddDeviceAsync(new Device(deviceId)).ConfigureAwait(false);
                }
                key = device.Authentication.SymmetricKey.PrimaryKey;
            }

            if (deviceClient == null)
            {
                deviceClient = DeviceClient.Create(hostname, new DeviceAuthenticationWithRegistrySymmetricKey(deviceId, key), TransportType.Mqtt);
                deviceClient.SetConnectionStatusChangesHandler(SomeConnectionStatusChangesHandler);
            }
        }

        public async Task RegisterDirectMethodAsync(string Name, MethodCallback MethodHandler)
        {
            await deviceClient.SetMethodHandlerAsync(Name, MethodHandler, new object()).ConfigureAwait(false);
        }

        private void SomeConnectionStatusChangesHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
        {
            Console.WriteLine($"Status changed to: {status}. Reason : {reason}");
        }
    }

[...]

PR #970, opened 2019-06-14
Change type to prevent overflow and negative value.

SergiiChurakov:
## Description of the changes
Change type to prevent overflow and negative value.

#969

### DIFF
--- a/iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
+++ b/iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
@@ -98,8 +98,8 @@ namespace Microsoft.Azure.Devices.Client.TransientFaultHandling
                 if (currentRetryCount < this.retryCount)
                 {
                     Random random = new Random();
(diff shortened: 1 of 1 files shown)


iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
```diff
@@ -98,8 +98,8 @@ public override ShouldRetry GetShouldRetry()
                 if (currentRetryCount < this.retryCount)
                 {
                     Random random = new Random();
-                    int num = (int)((Math.Pow(2.0, (double)currentRetryCount) - 1.0) * (double)random.Next((int)(this.deltaBackoff.TotalMilliseconds * 0.8), (int)(this.deltaBackoff.TotalMilliseconds * 1.2)));
```
CIPop: I don't think this fixes it. Instead of changing the type, I think you need to add `checked { ... }` around the operation.
Please also add a unit test to ensure this will throw on overflow.

CIPop:
@SergiiChurakov no activity for 11 days. Closing for now.
Feel free to resubmit taking into account the observations above.

PR #1002, opened 2019-07-12
Supporting timeouts when calling receive async

debugerr:
## Checklist

## Description of the changes
When calling ReceiveAsync on the DeviceClient the timeout nor the cancellation token are effective in aborting the request when using AmqpTransport. There are 2 issues we found: 
1. The AmqpTransportHandler does have an endless while(true) and won't exit until a message was received, or the operation timed out. 
2. The actual timeout being supplied to the method is not reaching the AmqpTransportHandler because it is wrapped in the CancellationToken that has no effect on the AmqpTransportHandler level.

@fbeltrao 

Fixes:  #1003

### DIFF
--- a/e2e/test/DeviceTokenRefreshE2ETests.cs
+++ b/e2e/test/DeviceTokenRefreshE2ETests.cs
@@ -116,7 +116,14 @@ namespace Microsoft.Azure.Devices.E2ETests
 
                 var message = new Client.Message(Encoding.UTF8.GetBytes("Hello"));
 
-                using (var cts = new CancellationTokenSource(TimeSpan.FromSeconds(ttl * 10)))
+                // calculating test run timeout: 
+                // we use the ttl + max percentage buffer time * number of calls we make to WaitForTokenRefreshAsync
+                // and add a buffer for opening and sending on the deviceclient.
+                const int operationCompletionBufferTimeInSeconds = 10;
+                const int numberOfWaitForTokenCalls = 2;
+                var testRunTimeout = (numberOfWaitForTokenCalls * (ttl * (float)(1 + buffer / 100))) + operationCompletionBufferTimeInSeconds;
+
+                using (var cts = new CancellationTokenSource(TimeSpan.FromSeconds(testRunTimeout)))
                 {
                     try
                     {
@@ -137,7 +144,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                     // Wait for the Token to expire.
                     if (transport == Client.TransportType.Http1)
                     {
-                        float waitTime = (float)ttl * ((float)buffer / 100) + 1;
+                        float waitTime = ttl * (1 + (float)buffer / 100);
                         Console.WriteLine($"[{DateTime.UtcNow}] Waiting {waitTime} seconds.");
                         await Task.Delay(TimeSpan.FromSeconds(waitTime)).ConfigureAwait(false);
                     }
@@ -162,7 +169,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                     // Ensure that the token was refreshed.
                     Console.WriteLine($"[{DateTime.UtcNow}] Token was refreshed after {refresher.DetectedRefreshInterval} (ttl = {ttl} seconds).");
                     Assert.IsTrue(
-                        refresher.DetectedRefreshInterval.TotalSeconds < (float)ttl * (1 + (float)buffer/100), // Wait for more than what we expect.
+                        refresher.DetectedRefreshInterval.TotalSeconds < ((float)ttl * (1 + (float)buffer/100)) + 1, // Wait for more than what we expect.
                         $"Token was refreshed after {refresher.DetectedRefreshInterval} although ttl={ttl} seconds.");
 
                     Console.WriteLine($"[{DateTime.UtcNow}] CloseAsync");
--- a/e2e/test/DeviceTokenRefreshE2ETests.cs
+++ b/e2e/test/DeviceTokenRefreshE2ETests.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Azure.Devices.E2ETests
         {
             TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
 
-            int buffer = 50;
+            const int bufferInPercentage = 50;
             Device device = testDevice.Identity;
             SemaphoreSlim deviceDisconnected = new SemaphoreSlim(0);
 
@@ -98,7 +98,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                 device.Id, 
                 device.Authentication.SymmetricKey.PrimaryKey, 
                 ttl, 
-                buffer,
+                bufferInPercentage,
                 transport);
 
             using (DeviceClient deviceClient = DeviceClient.Create(testDevice.IoTHubHostName, refresher, transport))
@@ -121,7 +121,8 @@ namespace Microsoft.Azure.Devices.E2ETests
                 // and add a buffer for opening and sending on the deviceclient.
                 const int operationCompletionBufferTimeInSeconds = 10;
                 const int numberOfWaitForTokenCalls = 2;
-                float testRunTimeout = (numberOfWaitForTokenCalls * (ttl * (float)(1 + buffer / 100))) + operationCompletionBufferTimeInSeconds;
+                int bufferInSeconds = (int)(ttl * ((float)bufferInPercentage / 100));
+                float testRunTimeout = (numberOfWaitForTokenCalls * (ttl + bufferInSeconds)) + operationCompletionBufferTimeInSeconds;
 
                 using (var cts = new CancellationTokenSource(TimeSpan.FromSeconds(testRunTimeout)))
                 {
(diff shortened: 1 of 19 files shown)


CIPop:
Thanks @debugerr !

> The actual timeout being supplied to the method is not reaching the AmqpTransportHandler because it is wrapped in the CancellationToken that has no effect on the AmqpTransportHandler level

The cancellation token worked before but with the granularity of `OperationTimeout` so it had some effect but not cancel immediately.

debugerr:
Hi @CIPop, thanks for driving this through so quickly! I think we may still have potential for an optimization in the submitted PR:

```
while (true)
{
      message = await _amqpUnit.ReceiveMessageAsync(timeout).ConfigureAwait(false);
      if (message != null || DateTime.UtcNow >= absoluteTimeout)
      {
            break;
      }
}
```

The while true seems to be unnecessary now. If the timeout handling in the _amqpUnit is correct (I quickly had a look a that and it seems to be the case) then the result of ReceivMessageAsync will always be either:

1. a message is returned within < timeout
2. no message is returned and the timeout fires >= timeout
3. an exception is thrown for some reason < timeout

I think it should not happen that we get a null result < timeout, only that case would warrant the while(true) but then we would need to adjust the timeout that is passed in to the additional call to ensure we keep the promise we made upstream with the original timeout. 

I would be voting for getting rid of the while(true) loop completely. What do you think?

CIPop:
> I would be voting for getting rid of the while(true) loop completely. What do you think?

@debugerr If the code isn't shared by the cancellation token path then it should be fine to remove the loop. The cancellation token path can't take advantage of knowing when cancellation needs to occur so for that we still need a variant of ReceiveAsync with the loop for AMQP.

e2e/test/DeviceTokenRefreshE2ETests.cs
```diff
@@ -116,7 +116,14 @@ private async Task DeviceClient_TokenIsRefreshed_Internal(Client.TransportType t
 
                 var message = new Client.Message(Encoding.UTF8.GetBytes("Hello"));
 
-                using (var cts = new CancellationTokenSource(TimeSpan.FromSeconds(ttl * 10)))
+                // calculating test run timeout: 
+                // we use the ttl + max percentage buffer time * number of calls we make to WaitForTokenRefreshAsync
+                // and add a buffer for opening and sending on the deviceclient.
+                const int operationCompletionBufferTimeInSeconds = 10;
+                const int numberOfWaitForTokenCalls = 2;
+                var testRunTimeout = (numberOfWaitForTokenCalls * (ttl * (float)(1 + buffer / 100))) + operationCompletionBufferTimeInSeconds;
```
CIPop: Nit: `int` instead of `var` (see coding guidelines on using var only when the type is written on the same line.

debugerr: Ok. Fixed.

mfalkvidd:
Would it be possible to have this fix included in a release soon?

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -199,9 +204,22 @@ public async Task<Message> ReceiveMessageAsync(TimeSpan timeout)
...
+
+                var openLinkStopWatch = Stopwatch.StartNew();
+
                 await EnsureReceivingLinkIsOpenedAsync(timeout).ConfigureAwait(false);
                 Debug.Assert(_messageSendingLink != null);
 
+                openLinkStopWatch.Stop();
+                timeout = timeout - openLinkStopWatch.Elapsed;
+
+                if (timeout.TotalMilliseconds <= 0)
```
davilu: `if (openLinkStopWatch.Elapsed>=timeout)`

iothub/device/src/Transport/ErrorDelegatingHandler.cs
```diff
@@ -33,6 +33,7 @@ public ErrorDelegatingHandler(IPipelineContext context, IDelegatingHandler inner
             typeof(ClosedChannelException),
             typeof(TimeoutException),
             typeof(OperationCanceledException),
+            typeof(TaskCanceledException),
```
davilu: `TaskCanncelledException` shouldn't map to network category.

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -202,26 +202,17 @@ public override Task<Twin> SendTwinGetAsync(CancellationToken cancellationToken)
...
-        }
-
-        public override async Task<Message> ReceiveAsync(TimeSpan timeout, CancellationToken cancellationToken)
-        {
-            // Long-polling is not supported
-            if (!TimeSpan.Zero.Equals(timeout))
-            {
-                throw new ArgumentOutOfRangeException(nameof(timeout), "Http Protocol does not support a non-zero receive timeout");
-            }
+            HttpResponseMessage responseMessage = null;
```
davilu: This local variable with `null` value is not required.

iothub/device/src/Transport/TransportHandler.cs
```diff
@@ -37,7 +32,29 @@ protected override void Dispose(bool disposing)
             {
                 OnTransportClosedGracefully();
             }
+        }
+
+        public override async Task OpenAsync(TimeSpan timeout)
+        {
+            using (var cts = new CancellationTokenSource((int)timeout.TotalMilliseconds))
```
davilu: `CancellationTokenSource` accepts `TimeSpan`

iothub/device/src/Transport/TransportHandler.cs
```diff
@@ -37,7 +32,29 @@ protected override void Dispose(bool disposing)
             {
                 OnTransportClosedGracefully();
             }
+        }
+
+        public override async Task OpenAsync(TimeSpan timeout)
```
davilu: Shouldn't map `TimeSpan` to `CancellationToken`

iothub/device/src/Transport/TransportHandler.cs
```diff
@@ -37,7 +32,29 @@ protected override void Dispose(bool disposing)
...
+        {
+            using (var cts = new CancellationTokenSource((int)timeout.TotalMilliseconds))
+            {
+                await this.OpenAsync(cts.Token).ConfigureAwait(false);
+            }
+        }
 
+        public override async Task<Message> ReceiveAsync(TimeSpan timeout)
```
davilu: Shouldn't map `TimeSpan` to `CancellationToken`

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -458,7 +474,7 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
         /// <summary>
         /// Implicit open handler.
         /// </summary>
-        private async Task EnsureOpenedAsync(CancellationToken cancellationToken)
+        private async Task EnsureOpenedAsync(CancellationToken cancellationToken, TimeSpan? timeout = null)
```
davilu: Shouldn't have both `CancellationToken` and `TimeSpan`.  If required `CancellationToken` should be created from `TimeSpan`.

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -482,16 +500,24 @@ private async Task EnsureOpenedAsync(CancellationToken cancellationToken)
             }
         }
 
-        private Task OpenAsyncInternal(CancellationToken cancellationToken)
+        private Task OpenAsyncInternal(CancellationToken cancellationToken, TimeSpan? timeout = null)
```
davilu: The same as above.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -159,20 +159,39 @@ public override async Task SendEventAsync(IEnumerable<Message> messages, Cancell
...
             Message message = null;
+
             while (true)
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                message = await _amqpUnit.ReceiveMessageAsync(timeout).ConfigureAwait(false);
+                // The AMQP protocol implementation does not take a cancellation token. 
+                // We use a shorter timeout to be able to check on the cancellation state of the passed
+                // in token
+                message = await _amqpUnit.ReceiveMessageAsync(defaultReceiveTimeout).ConfigureAwait(false);
```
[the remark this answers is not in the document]
mfalkvidd: Shorter times is the whole point of this PR. I need sub second response time when there are no messages in the queue.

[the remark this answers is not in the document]
davilu: The cost with shorter time is to very expensive and it's going to impact some major products. As a trade off, if there is tight restrict on time, please use timeout instead of CancellationToken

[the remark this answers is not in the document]
mfalkvidd: https://github.com/Azure/azure-iot-sdk-csharp/pull/1027 says this:
> Deprecating ReceiveAsync(Timeout): we recommend using the CancellationToken APIs instead.
Maybe I'm misinterpreting the meaning of that statement, but doesn't it mean that we shouldn't use timeout?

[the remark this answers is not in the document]
davilu: I disagree to deprecate timeout API since AMQP library doesn't support `CancellationToken`.

prmathur-microsoft:
I'm closing this PR as changes requested in the form of this PR is now addressed in master from [this ](https://github.com/Azure/azure-iot-sdk-csharp/commit/305112308d836c6813455c86749dfe47025e5a2b)commit.

PR #1012, opened 2019-07-25
Fix a would-be breaking change by making bodystream public again

timtay-microsoft:
Needed before the release


timtay-microsoft:
Closing this since a different pr contains these changes

PR #1013, opened 2019-07-25
Revert message api and cbs link break changes

timtay-microsoft:
Also fix would-be breaking change in bodystream


prmathur-microsoft:
As SDL tasks are currently unavailable (devops issue) merging it with admin rights.

PR #1018, opened 2019-07-30
Fix MqttTransportHandler not handling ConnectExceptions correctly

timtay-microsoft:
issue #648

While iterating over IP addresses for the openAsync call within MqttTransportHandler, if a ConnectException is thrown within an AggregateException, then the exception is not thrown, and the method continues iterating over IP addresses.

This PR extends the same behavior to a ConnectException that is not nested within an aggregate exception.


timtay-microsoft:
> Followup : Please produce a scenario where ipv6 can be used
> Everything else LGTM

In the [issue thread](https://github.com/Azure/azure-iot-sdk-csharp/issues/648) Cristian came up with a scenario where IPv6 is possible. His scenario:

>I believe the answer is yes: we need to support IPv6. If a customer chooses to IPv6 / IPv4 translation (to maintain only an IPv6 network) then the SDK will fail.
>

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -556,6 +556,11 @@ async Task OpenAsyncInternal(CancellationToken cancellationToken)
                 });
             }
 
+            if (this.channel == null)
```
CIPop: When would this happen?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -880,6 +885,11 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
                             return false; // Let anything else stop the application.
                         });
                     }
+                    catch (DotNetty.Transport.Channels.ConnectException ex)
```
CIPop: (coding std): Please add `using` instead of expanding to FQN.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -880,6 +885,11 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
...
+                    {
+                        //same as above, we will handle DotNetty.Transport.Channels.ConnectException
+                        Debug.WriteLine("ConnectException trying to connect to {0}: {1}", address.ToString(), ex.ToString(), nameof(CreateChannelFactory));
```
CIPop: Please follow logging guidelines (see the architecture.md file).

CIPop: (also this and other files)

iothub/device/tests/MqttTransportHandlerTests.cs
```diff
@@ -138,6 +138,26 @@ MqttTransportHandler CreateTransportHandlerWithMockChannel(out IChannel channel,
             return transport;
         }
 
+        MqttTransportHandler CreateTransportHandlerWithRealChannel(out IChannel channel, string connectionString = DummyConnectionString)
```
CIPop: Please add an E2E test. While the UT is good to have, we need to have the scenario handled.
My recommendation would be to use TcpListener and attempt a channel connect.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Sure, I'll fix this to use the proper logger like the above catch does

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Sure thing

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: This happens if CreateChannelFactory iterates through all the provided IP addresses passed in and it receives ConnectExceptions for all of them. The method would just return a null channel object that is saved into this.channel

CIPop: Argh... this is the 3rd or 4th time we have issues around this area (@abhipsaMisra @varunpuranik / cc @prmathur-microsoft ).

We really need the e2e test and dotNetty integration test to ensure we're getting it right on both Windows and Linux. Please also include @nayato (Max) to review that portion of the code since this issue was caused by a wrong assumption about dotNetty.

timtay-microsoft:
This PR is old, and cluttered, so I'm closing it for a new one

PR #1029, opened 2019-08-12
Add proxy handling to HttpClientHelper.httpClientObjWithPerRequestTim…

jenoola:
…eout

## Checklist

## Description of the changes

The change fixes proxy support for HTTP requests sent by service client or registry manager if custom operation timeout is specified and it is not equal to default operation timeout.

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -66,7 +66,18 @@ namespace Microsoft.Azure.Devices
             this.httpClientObj.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(CommonConstants.MediaTypeForDeviceManagementApis));
             this.httpClientObj.DefaultRequestHeaders.ExpectContinue = false;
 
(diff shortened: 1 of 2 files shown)


jenoola:
@abhipsaMisra this object is used when timeout is specified and it is not the same as the default timeout. I haven't added it, I just corrected proxy handling.

abhipsaMisra:
> @abhipsaMisra this object is used when timeout is specified and it is not the same as the default timeout. I haven't added it, I just corrected proxy handling.

Yes, I agree; and the changes for proxy handling look good; what I am trying to understand is - 
- Why we needed a separate `HttpClient()` object only for specifying timeout?
- The tests are all device streaming related, so is this object used in device streaming cases?

jenoola:
Upon further investigation it was determined that the mentioned code duplication in HttpClientHelper created some dead code (httpClientObjWithPerRequestTimeout is not used in SDK code and is not exposed to customer as a part of API), so this PR should be closed and dead code should be removed.

timtay-microsoft:
Closing this PR in lieu of #1178

PR #1031, opened 2019-08-16
Throw exception when TopicName in MQTT packet is too long

### LINKED ISSUE #853: IotHubException "Work must be complete in the same order as it was started. Expected work id: '2', actual work id: '9522' "
- **OS, version, SKU and CPU architecture used:** Windows 10 Enterprise 1809 x64
- **Application's .NET Target Framework :** Targeting UWP 17763
- **Device:** Laptop
- **SDK version used:** Microsoft.Azure.Devices.Client 1.19.0 (also repros with 1.29.0-preview-002)

## Description of the issue:

When using ModuleClient in a UWP device client application I am getting the following exception when sending a message with several properties. The app is intended to run on Windows 10 IoT Core device, and the bug also repros when run on PC (i.e. no need to set up device to repro). The same message can be sent successfully using DeviceClient. Also it works from a plain console app with both Module and Device client.
```
An exception of type 'Microsoft.Azure.Devices.Client.Exceptions.IotHubException' occurred in System.Private.CoreLib.dll but was not handled in user code
Work must be complete in the same order as it was started. Expected work id: '2', actual work id: '9522'

   at Microsoft.Azure.Devices.Client.Transport.Mqtt.OrderedTwoPhaseWorkQueue`2.CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.ProcessMessage(IChannelHandlerContext context, Packet packet)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.SendMessageAsync(IChannelHandlerContext context, Message message)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.WriteAsync(IChannelHandlerContext context, Object data)
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.<>c__DisplayClass27_0.<<ExecuteWithErrorHandlingAsync>b__0>d.MoveNext()
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.ExecuteWithErrorHandlingAsync[T](Func`1 asyncOperation)
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.<>c__DisplayClass18_0.<<SendEventAsync>b__0>d.MoveNext()
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.SendEventAsync(Message message, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.InternalClient.SendEventAsync(Message message)
   at ModuleClientTest.MainPage.<>c__DisplayClass3_0.<<MainPage_Loaded>b__0>d.MoveNext()
```

```C#
private async void MainPage_Loaded(object sender, RoutedEventArgs e)
{
    ModuleClient client1 = ModuleClient.CreateFromConnectionString("<insert_module_connection_string>", TransportType.Mqtt);
    client1.SetConnectionStatusChangesHandler(ConnectionStatusChanged);
    await client1.OpenAsync().ConfigureAwait(true);
    Message msg = CreatReproMessage();
    await client1.SendEventAsync(msg);
}

Message CreatReproMessage()
{
    Random rnd = new Random();
    int len = 1024 * 128;
    byte[] bytes = new byte[len];
    rnd.NextBytes(bytes);
    Message msg = new Message(bytes);
    int numProp = 82;
    for (int i = 0; i < numProp; i++)
    {
        msg.Properties.Add(Guid.NewGuid().ToString(), new string('*', 256));
    }
    return msg;
}

private void ConnectionStatusChanged(ConnectionStatus status, ConnectionStatusChangeReason reason)
{
    Debug.WriteLine($"Status: {status} - Reason: {reason}");
}
```
## Console log of the issue:

Status: Connected - Reason: Connection_Ok
Status: Disconnected - Reason: Communication_Error
Exception thrown: 'Microsoft.Azure.Devices.Client.Exceptions.IotHubException' in System.Private.CoreLib.dll
An exception of type 'Microsoft.Azure.Devices.Client.Exceptions.IotHubException' occurred in System.Private.CoreLib.dll but was not handled in user code
Work must be complete in the same order as it was started. Expected work id: '2', actual work id: '9522'

jenoola:
## Checklist

Fixes #853

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -22,6 +22,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
     using System.Diagnostics.Contracts;
     using System.Globalization;
     using System.IO;
+    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
 
@@ -50,6 +51,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         private const string IotHubTrueString = "true";
         private const string SegmentSeparator = "/";
         private const int MaxPayloadSize = 0x3ffff;
+        private const int MaxTopicNameLength = 0xffff;
 
         private static readonly Action<object> PingServerCallback = PingServer;
         private static readonly Action<object> CheckConnAckTimeoutCallback = ShutdownIfNotReady;
@@ -1173,6 +1175,11 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             else
             {
                 msg = topicName;
+            }
+
+            if (Encoding.UTF8.GetByteCount(msg) > MaxTopicNameLength)
+            {
+                throw new MessageTooLargeException($"TopicName for MQTT packet cannot be larger than {MaxTopicNameLength} bytes, current length is {Encoding.UTF8.GetByteCount(msg)}. The probable cause is the list of message.Properties and/or message.systemProperties is too long. Please use AMQP or HTTP.");
             }
 
             return msg;
--- a/iothub/device/tests/Mqtt/MqttIotHubAdapterTest.cs
+++ b/iothub/device/tests/Mqtt/MqttIotHubAdapterTest.cs
@@ -5,10 +5,12 @@ namespace Microsoft.Azure.Devices.Client.Test.Mqtt
 {
     using DotNetty.Codecs.Mqtt.Packets;
     using DotNetty.Transport.Channels;
+    using Microsoft.Azure.Devices.Client.Exceptions;
     using Microsoft.Azure.Devices.Client.Transport.Mqtt;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using Moq;
     using System;
+    using System.Collections.Generic;
     using System.Threading.Tasks;
 
     [TestClass]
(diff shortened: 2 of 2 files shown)


iothub/device/tests/Mqtt/MqttIotHubAdapterTest.cs
```diff
@@ -37,6 +39,32 @@ public void TestPopulateMessagePropertiesFromPacket_NormalMessage()
             Assert.AreEqual("MessageId1", message.SystemProperties["message-id"]);
         }
 
+        [TestMethod]
+        [ExpectedException(typeof(MessageTooLargeException), AllowDerivedTypes = false)]
+        public void TestPopulateMessagePropertiesFromMessage_TopicNameTooLong()
+        {
+            var message = new Message();
+
+            for (int i = 0; i < 65; i++)
```
bikamani: Talked offline about doing this more efficiently

iothub/device/tests/Mqtt/MqttIotHubAdapterTest.cs
```diff
@@ -37,6 +39,32 @@ public void TestPopulateMessagePropertiesFromPacket_NormalMessage()
...
+            {
+                message.Properties.Add(Guid.NewGuid().ToString(), new string('1', 1024));
+            }
+
+            var publishPacket = new PublishPacket(QualityOfService.AtMostOnce, false, false)
```
bikamani: Is it better to wrap this common code in a method? I see this used in unit test above this as well.

iothub/device/tests/Mqtt/MqttIotHubAdapterTest.cs
jenoola: turns out, I can't set just one super long property (UriHelper doesn't allow it). seems that the loop should stay

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -1175,6 +1177,11 @@ internal string PopulateMessagePropertiesFromMessage(string topicName, Message m
                 msg = topicName;
             }
 
+            if (Encoding.ASCII.GetByteCount(msg) > MaxTopicNameLength)
```
CIPop: Why do you need the encoding? The `msg` is passed as-is to message. I doubt ASCII is the right char-set - likely a small subset anyway. 
I think you can assume correct chars are used and simply count characters instead of total bytes from the string.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -1175,6 +1177,11 @@ internal string PopulateMessagePropertiesFromMessage(string topicName, Message m
...
+            {
+                throw new MessageTooLargeException($"TopicName for MQTT packet cannot be larger than {MaxTopicNameLength} bytes, current size is {Encoding.ASCII.GetByteCount(msg)}. Please send a smaller message or use AMQP. TopicName is \n\"{topicName}\"");
```
CIPop: The message isn't clear: devs don't set topic names. Please explain what are the likely causes for the topic name to be large due to the application using our API.
1. Smaller message is wrong here: sending a smaller payload won't affect topic name.
2. Also, definitely do _not_ leak the topic name itself: it could contain PII/HBI information and the error is caused by input. Doing so could allow someone with a device to display information such as device name, etc.
3. We should display the current len and expected len. Also, using AMQP or HTTP (not only AMQP).

/cc @nayato

iothub/device/tests/Mqtt/MqttIotHubAdapterTest.cs
CIPop: What is the max-len for UriHelper?

iothub/device/tests/Mqtt/MqttIotHubAdapterTest.cs
CIPop: Are there other things contributing to the length of the topic? (E.g. I see SystemProperties affecting this as well: what are all our the features using SystemProperties?)
/cc @prmathur-microsoft @jebrando @pierreca - please check your SDKs for max topic len as new features add to SystemProperties and app can set a lot of properties as well.

timtay-microsoft:
Closing this PR for a fresh PR in #1197

PR #1033, opened 2019-08-19
Reference to latest version of Microsoft.IdentityModel.Clients.ActiveDirectory as per security alert

bikamani:
Reference to latest version of Microsoft.IdentityModel.Clients.ActiveDirectory as per security alert


bikamani:
> Did we figure out why the SDL tasks did not capture this?

Opened thread with the team handling Component Governance.

PR #1034, opened 2019-08-19
Reference to latest version of Microsoft.IdentityModel.Clients.ActiveDirectory

bikamani:
Reference to latest version of Microsoft.IdentityModel.Clients.ActiveDirectory as per security alert

### DIFF
--- a/e2e/test/E2ETests.csproj
+++ b/e2e/test/E2ETests.csproj
@@ -139,4 +139,4 @@
     </PackageReference>
   </ItemGroup>
 
-gi</Project>
+</Project>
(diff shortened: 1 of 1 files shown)


e2e/test/E2ETests.csproj
```diff
@@ -139,4 +139,4 @@
     </PackageReference>
   </ItemGroup>
 
-</Project>
\ No newline at end of file
+gi</Project>
```
abhipsaMisra: "gi"?

bikamani: Not sure how did that come here. Thanks for noticing. updated.

bikamani:
> Thanks @bikamani ! This is required in `master` too I believe.

There is a PR out there for master as well.

PR #1035, opened 2019-08-20
Add DeviceNotOnlineException for ServiceClient

jenoola:
## Checklist

### DIFF
--- a/common/src/service/ExceptionHandlingHelper.cs
+++ b/common/src/service/ExceptionHandlingHelper.cs
@@ -8,8 +8,9 @@ namespace Microsoft.Azure.Devices
     using System.Net;
     using System.Net.Http;
     using System.Threading.Tasks;
-
+    using Microsoft.Azure.Devices.Common;
     using Microsoft.Azure.Devices.Common.Exceptions;
+    using Microsoft.Azure.Devices.Common.Extensions;
 
     class ExceptionHandlingHelper
     {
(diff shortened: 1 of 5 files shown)


common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -37,5 +38,21 @@ public static Task<string> GetExceptionMessageAsync(HttpResponseMessage response
...
+            if (Enum.TryParse(httpErrorCode, out errorCode))
+            {
+                switch (errorCode)
+                {
+                    case ErrorCode.DeviceNotOnline:
+                        return new DeviceNotOnlineException(exMsg);
+                }
+            }
+            return new DeviceNotFoundException(exMsg);
+        }
```
prmathur-microsoft: Should it return the generic exception DeviceNotFoundException in all other cases?

prmathur-microsoft: Can you add a test for this scenario as well?

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -18,7 +19,7 @@ class ExceptionHandlingHelper
             var mappings = new Dictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>();
 
             mappings.Add(HttpStatusCode.NoContent, async (response) => new DeviceNotFoundException(await GetExceptionMessageAsync(response).ConfigureAwait(false)));
-            mappings.Add(HttpStatusCode.NotFound, async (response) => new DeviceNotFoundException(await GetExceptionMessageAsync(response).ConfigureAwait(false)));
+            mappings.Add(HttpStatusCode.NotFound, HttpNotFoundMapping);
```
bikamani: Trying to understand why is this done differently now?

iothub/service/src/Common/Exceptions/DeviceNotOnlineException.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Common.Exceptions
+{
+    [Serializable]
```
bikamani: Why is this Serializable?

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -40,6 +40,7 @@ public enum ErrorCode
         JobNotFound = 404002,
         PartitionNotFound = 404003,
         ModuleNotFound = 404010,
+        DeviceNotOnline = 404103,
```
bikamani: How is this value defined? Is it documented anywhere?

common/src/service/ExceptionHandlingHelper.cs
jenoola: It's been returning generic DeviceNotFoundException before, it's just according to exception in http headers. But DeviceNotFound and DeviceNotOnline both have 404 error code, that's why the issue appeared

common/src/service/ExceptionHandlingHelper.cs
jenoola: 'DeviceNotFound' and 'DeviceNotOnline' both have 404 error code in http header, so there is only one mapping for both

iothub/service/src/Common/Exceptions/ErrorCode.cs
jenoola: this code is returned by IoT Hub when device is not online

iothub/service/src/Common/Exceptions/DeviceNotOnlineException.cs
jenoola: this class is implemented as DeviceNotFoundException which is serializable

prmathur-microsoft:
Closing this PR.

PR #1038, opened 2019-08-23
Deprecating support for .NET MicroFramework.

CIPop:
Deprecating support for .NET MicroFramework in future SDK versions.
Because NETMF is no longer supported or actively in-development, we will remove all references and code related to it in a future release.

### DIFF
--- a/doc/devguide.md
+++ b/doc/devguide.md
@@ -83,6 +83,9 @@ If you would like to develop an application using the Azure IoT SDK for C# (usin
 Contains the IoT Hub Device SDK source, unit-tests and samples. 
 This produces the `Microsoft.Azure.Devices.Client` NuGet package.
 
+```diff
+- .NET MicroFramework will not be supported in future versions of the SDK.
+```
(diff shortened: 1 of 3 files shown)


doc/devguide.md
```diff
@@ -83,6 +83,9 @@ If you would like to develop an application using the Azure IoT SDK for C# (usin
 Contains the IoT Hub Device SDK source, unit-tests and samples. 
 This produces the `Microsoft.Azure.Devices.Client` NuGet package.
 
+```diff
+- .NET MicroFramework will not be supported in future versions of the SDK.
+```
 The src.NetMF folder contains the .NET Microframework port.
```
prmathur-microsoft: Is this true for provisioning as well ?

CIPop: No: DPS was a new feature and so it is supported only for ns20. We will need to redo our top-level readme.md as we don't make support very well understood (table with actual support would be great) - not per feature but per DLL.

PR #1065, opened 2019-09-17
E2E test fixes

CIPop:
## Checklist

## Description of the changes

- Removing 1:1 file upload notification checks. 
- Removing test device creation delay.
- Increase token refresh interval.

### DIFF
--- a/common/test/TestDevice.cs
+++ b/common/test/TestDevice.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.Devices.E2ETests
 
     public class TestDevice
     {
-        private const int DelayAfterDeviceCreationSeconds = 5;
+        private const int DelayAfterDeviceCreationSeconds = 0;
         private static Dictionary<string, TestDevice> s_deviceCache = new Dictionary<string, TestDevice>();
         private static TestLogging s_log = TestLogging.GetInstance();
         private static SemaphoreSlim s_semaphore = new SemaphoreSlim(1, 1);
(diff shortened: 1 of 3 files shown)


common/test/TestDevice.cs
```diff
@@ -25,7 +25,7 @@ public enum ConnectionStringAuthScope
 
     public class TestDevice
     {
-        private const int DelayAfterDeviceCreationSeconds = 5;
+        private const int DelayAfterDeviceCreationSeconds = 0;
```
davilu: Service used to have an issue, new registered device could be connected with not found error. Please confirm it's addressed.

e2e/test/FileUploadE2ETests.cs
```diff
@@ -73,18 +73,16 @@ private async Task UploadFile(Client.TransportType transport, string filename, b
...
-
-                
                 using (FileStream fileStreamSource = new FileStream(filename, FileMode.Open, FileAccess.Read))
                 {
                     await deviceClient.UploadToBlobAsync(filename, fileStreamSource).ConfigureAwait(false);
                 }
 
                 FileNotification fileNotification = await VerifyFileNotification(notificationReceiver, testDevice.Id).ConfigureAwait(false);
 
+                // The following checks allow running these tests multiple times in parallel.
```
davilu: The problem with this change is if there is one file notification doesn't happen, we won't know which test failed.

common/test/TestDevice.cs
CIPop: According to @simonporter we should not need to wait anymore right after device creation.
I've tested on my hub and Canary and WestUS proved that to be true, at least in the last runs.

I don't want to remove the code though: if this comes back, we can quickly increase the time back to something that works around the issue.

e2e/test/FileUploadE2ETests.cs
CIPop: Good point - one would need to go through the large ETL file to figure that out.
1. Instead of just removing the assert, I'll replace it with a logging line.
2. I'll log the device name.

The debugging then would be a matter of looking at all parallel runs and creating a list of device names+filenames. Matching the ones that were supposed to be uploaded with the notifications will get you the device name and file name.

CIPop:
@abhipsaMisra @davilu the following test failed: Method_IoTHubSak_DeviceMethodReqLinkDropRecovery_MultipleConnections_Amqp on Linux in both Canary and WestUS.

PR #1066, opened 2019-09-17
Perf/Stress test fixes

CIPop:
- Switching to CPU Load instead of percentage (which can't be obtained easily in Linux).
- Removing no-retry scenarios except for the `device_all_noretry`.
- Fixing and enhancing logging.
- Fixing device_all_noretry dispose issues.


CIPop:
Included in https://github.com/Azure/azure-iot-sdk-csharp/pull/1077

PR #1078, opened 2019-10-01
fix of mqtt droping message while disconnected

davilu:
## Checklist

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -143,7 +143,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             {
                 if (this.IsInState(StateFlags.Closed))
                 {
-                    return;
+                    throw new IotHubCommunicationException("MQTT is disconnected.");
                 }
 
                 var message = data as Message;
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -534,11 +534,6 @@ async void ProcessMessage(IChannelHandlerContext context, Packet packet)
         {
             if (Logging.IsEnabled) Logging.Enter(this, context.Name, packet.PacketType, nameof(ProcessMessage));
 
-            if (this.IsInState(StateFlags.Closed))
-            {
-                return;
```
abhipsaMisra: Why do we not throw IoTHub exception here as well?

davilu: `ProcessMessage` is while receiving. It means there is pending packet after MQTT connection is closed. The packet should still be processed by application instead of ignore to avoid duplicated operations. An example is a PUBACK should be processed which will tell apllication the message was delivered, no retry is required.

abhipsaMisra: Makes sense, since the packet was received and then MQTT connection was closed, we should process the packet without needing a retry.

PR #1082, opened 2019-10-09
Receive Async timeout changes

bikamani:
ReceiveAsync timeout changes to fix open issues regarding timeout not being respected on AMQP.


iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -304,6 +298,12 @@ public override async Task<Message> ReceiveAsync(TimeSpan timeout, CancellationT
             return message;
         }
 
+
+        public override async Task<Message> ReceiveAsync(TimeoutHelper timeoutHelper)
+        {
+            throw new NotImplementedException("Http Protocol does not support receive timeout");
```
davilu:
```
TimeSpan timeout = timeoutHelper.remaining();
if (timeout > TimeSpan.Zero) {
   throw new ArgumentOutOfRangeException(nameof(timeout), "Http Protocol does not support a non-zero receive timeout");
}
else
{
  await ReceiveAsync(new CancellationTokenSource(timeout).Token).ConfigAwait(false);
}
```

e2e/test/MessageReceiveE2ETests.cs
```diff
@@ -153,7 +160,7 @@ private async Task ReceiveSingleMessage(TestDeviceType type, Client.TransportTyp
                     transport == Client.TransportType.Mqtt_WebSocket_Only)
                 {
                     // Dummy ReceiveAsync to ensure mqtt subscription registration before SendAsync() is called on service client.
-                    await deviceClient.ReceiveAsync(TimeSpan.FromSeconds(1)).ConfigureAwait(false);
+                    await deviceClient.ReceiveAsync(TimeSpan.FromSeconds(10)).ConfigureAwait(false);
```
CIPop: Why does this need to change? (Or exist in the first place?).

iothub/device/src/InternalClient.cs
```diff
@@ -354,10 +351,10 @@ public async Task<Message> ReceiveAsync(TimeSpan timeout)
         /// </summary>
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <returns>The receive message or null if there was no message until the default timeout</returns>
-        public Task<Message> ReceiveAsync(CancellationToken cancellationToken)
+        public async Task<Message> ReceiveAsync(CancellationToken cancellationToken)
         {
             // Codes_SRS_DEVICECLIENT_28_011: [The async operation shall retry until time specified in OperationTimeoutInMilliseconds property expire or unrecoverable(authentication, quota exceed) error occurs.]
-            return InnerHandler.ReceiveAsync(cancellationToken);
+            return await InnerHandler.ReceiveAsync(cancellationToken);
```
CIPop: This is adding an extra task and making the binary larger by producing an async class. Why is it needed?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -75,6 +75,29 @@ private void OnDisconnected()
...
+                {
+                    return;
+                }
+
+                _closed = false;
+            }
+
+            try
+            {
+                await _amqpUnit.OpenAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
```
CIPop: Note on this design: having the times here very small is risky and can cause some very odd behaviors.

iothub/device/src/Transport/DefaultDelegatingHandler.cs
```diff
@@ -73,16 +78,16 @@ public virtual Task WaitForTransportClosedAsync()
             return InnerHandler.WaitForTransportClosedAsync();
         }
 
-        public virtual Task<Message> ReceiveAsync(CancellationToken cancellationToken)
+        public virtual async Task<Message> ReceiveAsync(CancellationToken cancellationToken)
         {
             ThrowIfDisposed();
-            return InnerHandler?.ReceiveAsync(cancellationToken) ?? s_dummyResultObject;
+            return await InnerHandler.ReceiveAsync(cancellationToken).ConfigureAwait(false);
```
CIPop: Same for all of these - you can simply return the same Task object.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -203,11 +207,44 @@ public override async Task SendEventAsync(IEnumerable<Message> messages, Cancell
...
 
-        public override async Task<Message> ReceiveAsync(TimeSpan timeout, CancellationToken cancellationToken)
+        public override async Task<Message> ReceiveAsync(CancellationToken cancellationToken)
         {
+            cancellationToken.ThrowIfCancellationRequested();
+
             Message message = null;
 
-            cancellationToken.ThrowIfCancellationRequested();
+            this.EnsureValidState();
```
CIPop: Is this common code? If yes, can you avoid code duplication here by extracting a method?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -233,18 +271,55 @@ public override async Task<Message> ReceiveAsync(TimeSpan timeout, CancellationT
...
+            bool hasMessage = false;
+            cancellationToken.ThrowIfCancellationRequested();
+            var disconnectToken = this.disconnectAwaitersCancellationSource.Token;
+            this.EnsureValidState();
+
+            using (CancellationTokenSource linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, disconnectToken))
+            {
+                try
+                {
+                    hasMessage = await this.receivingSemaphore.WaitAsync(timeout, linkedCts.Token).ConfigureAwait(true);
```
CIPop: Can we do the same receivingSemaphore on AMQP instead of creating a whole new branch using TimeoutHelper?
That way the CancellationToken can be used without having the issues of maintaining two separate timeout modes while solving the issue of immediate cancellation.

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
```diff
@@ -31,6 +31,63 @@ internal class ProtocolRoutingDelegatingHandler : DefaultDelegatingHandler
         {
         }
 
+        public override async Task OpenAsync(TimeoutHelper timeoutHelper)
+        {
+            try
+            {
+                if (Logging.IsEnabled) Logging.Enter(this, timeoutHelper, $"{nameof(ProtocolRoutingDelegatingHandler)}.{nameof(OpenAsync)}");
+                
+                bool gain = await _handlerLock.WaitAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
```
CIPop: Same as above: isn't this the same code as the OpenAsync with cancellation token? Can you extract a method and avoid code duplication?

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -483,6 +483,38 @@ private async Task EnsureOpenedAsync(CancellationToken cancellationToken)
             }
         }
 
+        private async Task EnsureOpenedAsync(TimeoutHelper timeoutHelper)
```
CIPop: Same as others: avoid code duplication by extracting a method.

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -508,6 +540,31 @@ private Task OpenAsyncInternal(CancellationToken cancellationToken)
             cancellationToken);
         }
 
+        private Task OpenAsyncInternal(TimeoutHelper timeoutHelper)
```
CIPop: Dup code here as well.

e2e/test/MessageReceiveE2ETests.cs
bikamani: 1 s may be too short for receiveAsync - subscription, receive and all these together. As we are checking remaining time later with new changes, it seemed right to change this to a bigger number so it does not fail in e2e tests.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: What behaviors in specific? What would be the ideal approach if we were to change this?

e2e/test/MessageReceiveE2ETests.cs
```diff
@@ -114,7 +114,14 @@ public static async Task VerifyReceivedC2DMessageAsync(Client.TransportType tran
                 Client.Message receivedMessage = null;
 
                 _log.WriteLine($"Receiving messages for device {deviceId}.");
-                receivedMessage = await dc.ReceiveAsync(TimeSpan.FromSeconds(1)).ConfigureAwait(false);
+                if (transport == Client.TransportType.Http1)
```
abhipsaMisra: It might be worth adding a note here as to why we don't pass the TimeSpan to the Http call

iothub/device/src/DeviceClient.cs
```diff
@@ -280,11 +280,11 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
         public Task<Message> ReceiveAsync() => this.internalClient.ReceiveAsync();
 
         /// <summary>
-        /// Receive a message from the device queue using the default timeout.
+        /// Receive a message from the device queue using the cancellation token.
         /// </summary>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <exception cref="OperationCanceledException">Thrown when the operation has been canceled.</exception>  
-        /// <returns>The receive message or null if there was no message until the default timeout</returns>
+        /// <returns>The receive message</returns>
```
abhipsaMisra: Did the behavior change for the scenario when the cancellation token has expired but no message was received yet (do we not return a null message now)?

e2e/test/MessageReceiveE2ETests.cs
bikamani: On HTTP timeout is not supported, we throw an exception.

iothub/device/src/DeviceClient.cs
abhipsaMisra: We still throw an `OperationCanceledException` on cancellation token expiry; but I would no longer be getting a null message. Is this correct?

bikamani: no, still getting NULL when cancellation token is expired. I will add that in comment to description

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
bikamani: Can you explain more? What do you mean by two separate timeout modes here?

e2e/test/MessageReceiveE2ETests.cs
bikamani: Added a comment

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
bikamani: Done

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
bikamani: only disposing part is common, leaving it as it is. Extracting just disposing part does not make much of sense

iothub/device/src/Transport/RetryDelegatingHandler.cs
bikamani: same as above, does not make much sense here for extracting disposal code.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
CIPop: OpenAsync with very small timeout may cause rapid retries, etc. 

The right approach would be to properly implement CancellationToken cancellation which would be covering _all_ aspects, including the retry handler, etc.

CIPop: Also, I'm not sure this pattern applied to ReceiveAsync won't cause messages to be lost. Hub will only allow 10 pending messages before it stops sending which means apps may stop receiving messages.

PR #1095, opened 2019-10-24
Make ConfigurationMetrics on Configuration settable

timtay-microsoft:
For editing configurations using the registry manager, users need to be able to set the ConfigurationMetrics.

Node SDK already allows customers to do this, and so does the Java SDK.

[The original java issue](https://github.com/Azure/azure-iot-sdk-java/issues/600)

### DIFF
--- a/iothub/service/src/Configurations/Configuration.cs
+++ b/iothub/service/src/Configurations/Configuration.cs
@@ -101,7 +101,7 @@ namespace Microsoft.Azure.Devices
         /// Custom Configuration Metrics
         /// </summary>
         [JsonProperty(PropertyName = "metrics", NullValueHandling = NullValueHandling.Ignore)]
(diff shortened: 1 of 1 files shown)


iothub/service/src/Configurations/Configuration.cs
```diff
@@ -101,7 +101,7 @@ internal Configuration()
         /// Custom Configuration Metrics
         /// </summary>
         [JsonProperty(PropertyName = "metrics", NullValueHandling = NullValueHandling.Ignore)]
-        public ConfigurationMetrics Metrics { get; internal set; }
+        public ConfigurationMetrics Metrics { get; set; }
 
         /// <summary>
```
prmathur-microsoft: Can you add a E2E test for this change since this is public API I assume setting this will be reflected on the hub.

PR #1099, opened 2019-10-29
Fix issue where integer underflow caused exponential backoff time to retry to underflow

timtay-microsoft:
Issue #969

During the exponential backoff algorithm, some double values were cast to integer. When these values were large, the cast would result in an underflow which would cause the recommended next delay to be negative unexpectedly.

~~This pr alters the algorithm to only use doubles (which go to infinity, not to underflow) to fix this issue~~

Switched back to integers, and using checked statements to lookout for integer overflow

### DIFF
--- a/iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
+++ b/iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
@@ -98,24 +98,13 @@ namespace Microsoft.Azure.Devices.Client.TransientFaultHandling
                 if (currentRetryCount < this.retryCount)
                 {
                     Random random = new Random();
-                    try
-                    {
-                        checked
-                        {
-                            int num = (int) (Math.Pow(2.0, currentRetryCount) - 1.0) * random.Next((int)(this.deltaBackoff.TotalMilliseconds * 0.8), (int)(this.deltaBackoff.TotalMilliseconds * 1.2));
-                            int num2 = (int) Math.Min(this.minBackoff.TotalMilliseconds + num, this.maxBackoff.TotalMilliseconds);
-                            retryInterval = TimeSpan.FromMilliseconds(num2);
-                            return true;
-                        }
-                    }
-                    catch (OverflowException e)
-                    {
-                        //Integer overflow only occurs due to the exponential backoff calculations at high retry count
-                        // Therefore, if integer overflow occurs, it is reasonable to default to the default maxBackoff
-                        retryInterval = TimeSpan.FromMilliseconds(this.maxBackoff.TotalMilliseconds);
-                        return true;
-                    }
+                    double exponentialInterval = (Math.Pow(2.0, currentRetryCount) - 1.0) * random.Next((int) this.deltaBackoff.TotalMilliseconds * 8 / 10, (int) this.deltaBackoff.TotalMilliseconds * 12 / 10) + this.minBackoff.TotalMilliseconds;
+                    double maxInterval = this.maxBackoff.TotalMilliseconds;
+                    double num2 = Math.Min(exponentialInterval, maxInterval);
+                    retryInterval = TimeSpan.FromMilliseconds(num2);
+                    return true;
                 }
+
                 retryInterval = TimeSpan.Zero;
                 return false;
             };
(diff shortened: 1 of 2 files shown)


iothub/device/tests/ExponentialBackoffTests.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Microsoft.Azure.Devices.Client.TransientFaultHandling;
```
CIPop: Missing required header.

iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
```diff
@@ -98,9 +98,9 @@ public override ShouldRetry GetShouldRetry()
                 if (currentRetryCount < this.retryCount)
                 {
                     Random random = new Random();
-                    int num = (int)((Math.Pow(2.0, (double)currentRetryCount) - 1.0) * (double)random.Next((int)(this.deltaBackoff.TotalMilliseconds * 0.8), (int)(this.deltaBackoff.TotalMilliseconds * 1.2)));
-                    int num2 = (int)Math.Min(this.minBackoff.TotalMilliseconds + (double)num, this.maxBackoff.TotalMilliseconds);
-                    retryInterval = TimeSpan.FromMilliseconds((double)num2);
+                    double num = ((Math.Pow(2.0, currentRetryCount) - 1.0) * random.Next((int)(this.deltaBackoff.TotalMilliseconds * 0.8), (int)(this.deltaBackoff.TotalMilliseconds * 1.2)));
```
CIPop: 1. Why do you need to use `double` which is significantly slower and harder to debug. The end result is a TimeSpan which is based on `milliseconds`. I doubt anyone would care of timings less than a millisecond. Instead you can use something like ulong which should be large enough for these purposes. You may also want to use a `checked` statement to ensure no overflows.

2. What does it mean here to have a timespan formed from Double.MinValue, Double.MaxValue or Infinity? That basically makes the device either not wait (spin-wait causing high CPU) or wait forever (hang).

iothub/device/tests/ExponentialBackoffTests.cs
timtay-microsoft: Sure, I'll add it

iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
timtay-microsoft: 1) I'll make use of a checked block here and keep the numbers as integers then. If overflow occurs, the retry policy will just default to the max backoff.

2) This code does not create TimeSpans with Double.MinValue or Double.MaxValue. It creates TimeSpans within the min backoff and the max backoff

iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
```diff
@@ -98,10 +98,23 @@ public override ShouldRetry GetShouldRetry()
                 if (currentRetryCount < this.retryCount)
                 {
                     Random random = new Random();
-                    int num = (int)((Math.Pow(2.0, (double)currentRetryCount) - 1.0) * (double)random.Next((int)(this.deltaBackoff.TotalMilliseconds * 0.8), (int)(this.deltaBackoff.TotalMilliseconds * 1.2)));
-                    int num2 = (int)Math.Min(this.minBackoff.TotalMilliseconds + (double)num, this.maxBackoff.TotalMilliseconds);
-                    retryInterval = TimeSpan.FromMilliseconds((double)num2);
-                    return true;
+                    try
```
davilu: The code is too complicated to read. How about `if(num2<0) return maxBackoff;`

timtay-microsoft: Overflow may loop back around to look like a normal positive value again, so we do need the checked statement to correctly implement the exponential backoff.

davilu: Then don't convert it to int. 
```
double interval = Math.Pow(2.0, currentRetryCount) - 1.0) * random.Next((int)(this.deltaBackoff.TotalMilliseconds * 0.8), (int)(this.deltaBackoff.TotalMilliseconds * 1.2) + this.minBackoff.TotalMilliseconds;
if (interval > this.maxBackoff.TotalMilliseconds) {
   retryInterval = maxBackoff;
} else [
  retryInterval = TimeSpan.FromMilliseconds(interval);
}
```

timtay-microsoft: As Cristian mentioned above, double's are unnecessarily heavy here. We can keep everything as an int as long as we have the checked statement

Issue #1081, opened 2019-10-07
Default RetryPolicy is broken


dmytro-gokun:
Microsoft.Azure.Devices.Client, Version=1.21.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35

In my app, i've implemented LoggingRetryPolicy, so that i can monitor how the default ExponentialBackoff policy performs. Here's the code:

```
        private class LoggingRetryPolicy : IRetryPolicy
        {
            private readonly IRetryPolicy _policy;

            public LoggingRetryPolicy(IRetryPolicy policy)
            {
                _policy = policy;
            }

            public bool ShouldRetry(int currentRetryCount, Exception lastException, out TimeSpan retryInterval)
            {
                var result = _policy.ShouldRetry(currentRetryCount, lastException, out retryInterval);
                if (result)
                    Log.Warn(lastException, "Retrying. CurrentRetryCount: {0}, RetryInterval: {1}", currentRetryCount, retryInterval);
                return result;
            }
        }
```

So, nothing special here, just logging the current retry count & delay.
And here's how i use it:

```
var deviceClient = DeviceClient.Create(...);
// This is the default retry policy according to the docs.
var defaultRetryPolicy = new ExponentialBackoff(int.MaxValue, TimeSpan.FromMilliseconds(100), TimeSpan.FromSeconds(10), TimeSpan.FromMilliseconds(100));
deviceClient.SetRetryPolicy(new LoggingRetryPolicy(defaultRetryPolicy));
```

And then i do bunch of theses whenever needed:

```
await deviceClient.SendEventAsync(message, cancellationToken);
```

So far, so good. Mostly this works just fine. In fact it worked perfectly (for many months) until the last night. And here is the log from the last night:

2019-10-07T02:02:36.8839Z	Retrying. CurrentRetryCount: 0, RetryInterval: 00:00:00.1000000
2019-10-07T02:02:37.8372Z	Retrying. CurrentRetryCount: 1, RetryInterval: 00:00:00.2130000
...
2019-10-07T02:02:47.0596Z	Retrying. CurrentRetryCount: 6, RetryInterval: 00:00:06.0850000
2019-10-07T02:02:53.1893Z	Retrying. CurrentRetryCount: 7, RetryInterval: 00:00:10
...
2019-10-07T02:06:00.9973Z	Retrying. CurrentRetryCount: 24, RetryInterval: 00:00:10
... and BOOM:
[...]

PR #1104, opened 2019-11-05
Add clarification to device operation timeout documentation

timtay-microsoft:
Device Operation Timeout, by design, is only used for operations where user does not provide their own cancellation token. Adding clarification to the header for this timeout so users know not to expect it to matter for any operations with explicitly provided cancellation tokens

Solves part of issue #1081

### DIFF
--- a/iothub/device/src/DeviceClient.cs
+++ b/iothub/device/src/DeviceClient.cs
@@ -206,7 +206,10 @@ namespace Microsoft.Azure.Devices.Client
         }
 
         /// <summary>
-        /// Stores the timeout used in the operation retries.
+        /// Stores the timeout used in the operation retries. Note that this value is ignored for operations
+        /// where a cancellation token is provided. For example, SendEventAsync(Message) will use this timeout, but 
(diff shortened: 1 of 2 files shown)


iothub/device/src/DeviceClient.cs
```diff
@@ -206,7 +206,10 @@ public int DiagnosticSamplingPercentage
         }
 
         /// <summary>
-        /// Stores the timeout used in the operation retries.
+        /// Stores the timeout used in the operation retries. Note that this value is ignored for operations
+        /// where a cancellation token is provided. For example, SendEventAsync(Message) will use this timeout, but 
+        /// SendEventAsync(Message, CancellationToken) will not. The latter operation will only be canceled by the 
+        /// provided cancellation token.
         /// </summary>
```
prmathur-microsoft: Can you confirm if this is true for receiveAsync(), receiveAsync(timeout) ?

prmathur-microsoft:
[Nit] Please try to follow commit guidelines in commit message.

PR #1106, opened 2019-11-06
Made Message class partially cloneable

vipeller:
Message class has some system properties open to set by user. In certain scenarios Message class needs to be replicated which involves copying over message properties and system properties. The internal system properties dictionary is not exposed, and the specific getter methods on the top of 'system properties' don't distinguish between the case that "property is not defined" and "property has a null (default) value".

If we use the getter method, we cannot decide that the property is undefined or was set to null.

Currently we are fixing the following bug:

https://github.com/Azure/iotedge/issues/1923

Without this PR we use the assumption that null value means 'property is not defined', however long term we want to use the newly introduced CloneWithBody() method


iothub/device/tests/MessageTests.cs
```diff
@@ -147,5 +147,29 @@ public void SettingMessageAsSecurityMessageTest()
...
+
+            var clone = original.CloneWithBody(Encoding.UTF8.GetBytes("Cloned version"));
+
+            Assert.AreEqual("test_v_1", clone.Properties["test1"]);
+            Assert.AreEqual("test_v_2", clone.Properties["test2"]);
+
+            Assert.AreEqual("gzip", clone.ContentEncoding);
+            Assert.AreEqual("JohnDoe", clone.UserId);
+
+            var clonedContent = new StreamReader(clone.BodyStream, Encoding.UTF8).ReadToEnd();
```
drwill-ms: StreamReader is IDisposable, so put inside a using block

iothub/device/tests/MessageTests.cs
```diff
@@ -147,5 +147,29 @@ public void SettingMessageAsSecurityMessageTest()
...
+
+        [TestMethod]
+        public void CloneWithBodyTest()
+        {
+            var original = new Message(Encoding.UTF8.GetBytes("Original copy"));
+
+            original.Properties["test1"] = "test_v_1";
+            original.Properties["test2"] = "test_v_2";
+
+            original.ContentEncoding = "gzip";
```
drwill-ms: add ContentType prop

iothub/device/src/Message.cs
```diff
@@ -107,6 +107,32 @@ internal Message(Stream stream, bool ownStream)
...
+        /// <summary>
+        /// Clones an existing <see cref="Message"/> instance and sets content body defined by <paramref name="byteArray"/> on it.
+        /// </summary>
+        /// <param name="byteArray">Message content to be set after clone.</param>
+        /// <returns>A new instance of <see cref="Message"/> with body content defined by <paramref name="byteArray"/>,
+        /// and user/system properties of the cloned <see cref="Message"/> instance.
+        /// </returns>
+        /// <remarks>user should treat the input byte array as immutable when
+        /// sending the message.</remarks>
+        public Message CloneWithBody(byte[] byteArray)
```
drwill-ms: Will VS let you use the `in` keyword for this parameter?

iothub/device/tests/MessageTests.cs
```diff
@@ -165,5 +165,29 @@ public void SettingMessageAsSecurityMessageTest()
             Assert.AreEqual(CommonConstants.SecurityMessageInterfaceId, msg.SystemProperties[MessageSystemPropertyNames.InterfaceId]);
             Assert.IsTrue(msg.IsSecurityMessage);
         }
+
+        [TestMethod]
+        public void CloneWithBodyTest()
```
abhipsaMisra: Can we add a unit test around the scenario where Property/SystemProperty does have null value

abhipsaMisra:
Just to confirm, you are able to see null properties flow through, but only face issue when you try to clone the Message?
I ask this because we had recently updated our SDK to not drop null property (Message.Property) over AMQP anymore (MQTT and HTTP work as expected): PR #1243 .

vipeller:
@abhipsaMisra maybe the description of the fix is misleading: the origin of the problem is that we wanted to clone a message, including the system properties. Because the underlying dictionary is not exposed, we couldn't take a look that which system properties are set (by the user, e.g. content type)
What we could do is to call the related getter methods. It either returned a normal value (e.g. a string as "UTF8"), or null. But when it returned null, in an edge case this could have happened because the user set this property to null specifically. We don't have a good way to decide that the value is missing (not set by anyone), or it us set to null. Because of that, we have no way to clone the internal system properties map 100% accurately. 
Right now, we treat 'null' as missing value, and ignore the edge case when the user sets it to null (as it doesn't make much sense to set ContentType to null)

I lost the full context (it was 4 months ago), but I assume I saw this important because defined null properties get encoded later (and your comment tells me that it might be a correct assumption)

For the same reason, I cannot write a test easily to check if the clone replicated the properties explicitly set to null or they are just not set. I can get the dictionary via reflection and to a comparison, however some developers don't like testing private members via reflection, so before I invest any work in it, I need to ask if this is something you would accept, or maybe you have a different approach for the same test. Let me know.

iothub/device/src/Message.cs
vipeller: yes, it works (after I updated VS)

iothub/device/tests/MessageTests.cs
vipeller: I wonder if it is necessary here - everything is memory bound and gets dereferenced in a moment - dispose just unchains some dependencies in this case which would not make cleanup faster. On the other hand, it is easier to read without the using. Message instances are also disposable, yet the majority of the tests don't call dispose on them, even those requesting a body stream.

iothub/device/tests/MessageTests.cs
vipeller: the check needs reflection to access the internal maps and see if the cloned instance has entries with explicit null-s where it was defined in the original, and missing entries where it was not defined in the original Messages.
Let me know if you are ok these sort of tests (not everybody is ok with testing private parts) or you have a suggestion for different approach

iothub/device/tests/MessageTests.cs
drwill-ms: It won't be disposed when it goes out of scope, right? That's the point of IDisposable objects, they have unmanaged backing resources that need to be explicitly disposed.

I know it is a test, but for a long test run, I'd prefer not to gain resource leaks. Also if we have fxcop validation in place, it prevents a code warning.

iothub/device/src/Message.cs
```diff
@@ -79,6 +79,32 @@ internal Message(Stream stream, bool ownStream)
...
```
drwill-ms: Method needs to go below the properties

iothub/device/tests/MessageTests.cs
drwill-ms: You are right. The `Properies` property is private set, and initialized in the ctor with a dictionary. So it cannot be null.

iothub/device/tests/MessageTests.cs
vipeller: if it has unmanaged resources, then it has a finalizer that will be called sooner or later and cleans up, so no resource leak would happen. It just cleans up later - but in this particular example, because it is all in memory, no difference I can see.
Anyway, I can easily put it into a using (or call Dispose)

vipeller: fixed and now dispose is called on disposable objects

iothub/device/src/Message.cs
vipeller: moved the method down

iothub/device/tests/MessageTests.cs
vipeller: added

abhipsaMisra:
> @abhipsaMisra maybe the description of the fix is misleading: the origin of the problem is that we wanted to clone a message, including the system properties. Because the underlying dictionary is not exposed, we couldn't take a look that which system properties are set (by the user, e.g. content type)
> What we could do is to call the related getter methods. It either returned a normal value (e.g. a string as "UTF8"), or null. But when it returned null, in an edge case this could have happened because the user set this property to null specifically. We don't have a good way to decide that the value is missing (not set by anyone), or it us set to null. Because of that, we have no way to clone the internal system properties map 100% accurately.
> Right now, we treat 'null' as missing value, and ignore the edge case when the user sets it to null (as it doesn't make much sense to set ContentType to null)
> 
> I lost the full context (it was 4 months ago), but I assume I saw this important because defined null properties get encoded later (and your comment tells me that it might be a correct assumption)
> 
> For the same reason, I cannot write a test easily to check if the clone replicated the properties explicitly set to null or they are just not set. I can get the dictionary via reflection and to a comparison, however some developers don't like testing private members via reflection, so before I invest any work in it, I need to ask if this is something you would accept, or maybe you have a different approach for the same test. Let me know.

@vipeller Our Device Client library is set to have it's internal members visible to the unit test project in debug mode. So you should be able to call on Message.SystemProperty.
Let me know if you still face issues.

iothub/device/src/Message.cs
```diff
@@ -326,6 +326,32 @@ public byte[] GetBytes()
...
+        /// <returns>A new instance of <see cref="Message"/> with body content defined by <paramref name="byteArray"/>,
+        /// and user/system properties of the cloned <see cref="Message"/> instance.
+        /// </returns>
+        /// <remarks>user should treat the input byte array as immutable when
+        /// sending the message.</remarks>
+        public Message CloneWithBody(in byte[] byteArray)
+        {
+            var result = new Message(byteArray);
+
+            foreach (string key in this.Properties.Keys)
```
drwill-ms: we've chosen to not use "this." where not necessary. The rest of this file looks to be following that mode, so you can remove them here too.

vipeller: removed 'this.'

vipeller:
@abhipsaMisra oh cool, then I added a test, pls check if this is something you had on your mind

iothub/device/tests/MessageTests.cs
```diff
@@ -165,5 +165,62 @@ public void SettingMessageAsSecurityMessageTest()
...
+                original.ContentType = null;
+
+                using (var clone = original.CloneWithBody(Encoding.UTF8.GetBytes("Cloned version")))
+                {                    
+                    Assert.AreEqual("test_v_1", clone.Properties["test1"]);
+                    Assert.AreEqual(null, clone.Properties["test2"]);
+                    Assert.AreEqual(2, clone.Properties.Count);
+
+                    Assert.AreEqual("gzip", clone.ContentEncoding);
+                    Assert.AreEqual(null, clone.ContentType);
```
drwill-ms: Assert.IsNull. same on line 216

iothub/device/tests/MessageTests.cs
```diff
@@ -165,5 +165,62 @@ public void SettingMessageAsSecurityMessageTest()
...
+                    Assert.IsFalse(clone.SystemProperties.Keys.Contains(MessageSystemPropertyNames.UserId));
+                }
```
abhipsaMisra: original only has Properties set, not SystemProperties. What do we want to check with this assertion?

iothub/device/tests/MessageTests.cs
```diff
@@ -165,5 +165,62 @@ public void SettingMessageAsSecurityMessageTest()
...
+                        clonedContent = reader.ReadToEnd();
+                    }
+                
+                    Assert.AreEqual("Cloned version", clonedContent);
+                }
+            }
+        }
+
+        [TestMethod]
+        public void CloneWithBodyWithNullTest()
```
abhipsaMisra: Add SystemProperties as well, so that we cover all cases being impacted.

abhipsaMisra:
> @abhipsaMisra oh cool, then I added a test, pls check if this is something you had on your mind

@vipeller Looks good! Could you also make sure the test checks for SystemProperty allowing null values to be cloned?

iothub/device/tests/MessageTests.cs
vipeller: the ' original.ContentType = null;' part goes to SystemProperties I believe

vipeller: ok, I improved the check reading the SystemProperties directly, not through the getter method, which would return "null" even if the value is not in the dictionary

iothub/device/tests/MessageTests.cs
vipeller: There are some system properties opened by getter/setters, e.g. ContentType. In the background those settings are stored in SystemProperties

Issue #1100, opened 2019-10-31
An error "Malformed token" occur if DeviceID has Semicolon (e.g., X:1234;X:12;X:123456789)


wachi-yoshitaka:
An error "Malformed token" occured when access to IoTEdge via DPS (Individual Enrollments, X.509).
I specify DeviceID which has Semicolon (e.g., X:1234;X:12;X:123456789) to DPS as 'IoT Hub Device ID'.

The following is the code which is occured the error.

https://github.com/Azure/azure-iot-sdk-csharp/blob/41ae9b00c8d41f6829873714a07a4b0abdb49ea1/iothub/device/src/Common/Extensions/CommonExtensions.cs#L67

timtay-microsoft:
Right now, the SDK uses semicolons as delimiters when parsing connection strings, so when semicolons are in the device id itself, the SDK throws this exception.

That being said, IoT Hub does allow for device ids to contain semicolons, so this behavior from the SDK is incorrect. We'll have to rethink how we parse these connection strings.

timtay-microsoft:
The PR that you provided has now been merged into our master branch. Thanks for the contribution! I'll update this thread again once we have released a version of the SDK that contains this fix.

wachi-yoshitaka:
Thanks.

PR #1110, opened 2019-11-09
Fix for github issue #1100

wachi-yoshitaka:
Fixed an issue which is occurring 'Malformed Token' when specified device id has a semicolon.

This is not the best way.
Because, if device id is specified similar to 'connection string' (has both a semicolon and an equal sign), the code still has problems.
However, in the case that device id has one of a semicolon and an equal sign, the fix is helpful.

## Checklist

### DIFF
--- a/iothub/device/src/Common/Extensions/CommonExtensions.cs
+++ b/iothub/device/src/Common/Extensions/CommonExtensions.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Azure.Devices.Client.Extensions
                         (m.NextMatch().Success ? m.NextMatch().Index : valuePairString.Length) - (m.Index + m.Value.Length))
                 });
 
-            if (parts.Any(p => p.Length != 2))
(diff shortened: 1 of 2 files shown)


timtay-microsoft:
Can you provide a few unit tests around this scenario? The DeviceClientConnectionStringTests class should be a good place to test connection strings where the deviceId contains semicolons

wachi-yoshitaka:
I have modified the unit tests.
I added a semicolon to the device id of the unit test.

I checked the following.
- The source code before fix gets NG as result of unit test.
- The source code after fix gets OK as result of unit test.

timtay-microsoft:
These unit tests look good, now. I'll work to get this merged in soon. Thanks for the PR!

timtay-microsoft:
I will also need you to sign the Contributor License Agreement before I can merge this in, if you haven't already

wachi-yoshitaka:
Thanks.
I have signed the CLA.

Issue #1122, opened 2019-11-15
RegistryManager.GetTwinAsync throws exceptions


hanabi1224:
Message: 
    Microsoft.Azure.Devices.Common.Exceptions.IotHubException : Error converting value 1/1/0001 12:00:00 AM +00:00 to type 'System.DateTime'. Path 'statusUpdateTime', line 1, position 142.
      ----> Newtonsoft.Json.JsonSerializationException : Error converting value 1/1/0001 12:00:00 AM +00:00 to type 'System.DateTime'. Path 'statusUpdateTime', line 1, position 142.
      ----> System.InvalidCastException : Object must implement IConvertible.
  Stack Trace: 
    HttpClientHelper.ExecuteAsync(HttpClient httpClient, HttpMethod httpMethod, Uri requestUri, Func`3 modifyRequestMessageAsync, Func`2 isMappedToException, Func`3 processResponseMessageAsync, IDictionary`2 errorMappingOverrides, CancellationToken cancellationToken)
    HttpClientHelper.GetAsync[T](Uri requestUri, TimeSpan operationTimeout, IDictionary`2 errorMappingOverrides, IDictionary`2 customHeaders, Boolean throwIfNotFound, CancellationToken cancellationToken)
    AzureIoTUtils.GetTwinAsync(String deviceId) line 41
    AzureIoTUtilsTests.TestGetTwinAsync(String deviceId) line 33
    GenericAdapter`1.BlockUntilCompleted()
    NoMessagePumpStrategy.WaitForCompletion(AwaitAdapter awaitable)
    AsyncToSyncAdapter.Await(Func`1 invoke)
    TestMethodCommand.RunTestMethod(TestExecutionContext context)
    TestMethodCommand.Execute(TestExecutionContext context)
    SimpleWorkItem.PerformWork()
    --JsonSerializationException
    JsonSerializerInternalReader.EnsureType(JsonReader reader, Object value, CultureInfo culture, JsonContract contract, Type targetType)
    JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
    JsonSerializerProxy.DeserializeInternal(JsonReader reader, Type objectType)
    JsonSerializer.Deserialize(JsonReader reader, Type objectType)
    JsonSerializer.Deserialize[T](JsonReader reader)
    TwinJsonConverter.ReadJson(JsonReader reader, Type objectType, Object existingValue, JsonSerializer serializer)
[...]

PR #1123, opened 2019-11-15
Fix issue 1122

hanabi1224:
Fix  https://github.com/Azure/azure-iot-sdk-csharp/issues/1122
It fails when default JsonSerializerSettings.DateParseHandling is set to DateParseHandling.DateTimeOffset

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.Devices
     using Microsoft.Azure.Devices.Common.Exceptions;
     using Microsoft.Azure.Devices.Common.Extensions;
     using Microsoft.Azure.Devices.Shared;
+    using Newtonsoft.Json;
     using System;
     using System.Collections.Generic;
     using System.Collections.ObjectModel;
@@ -24,6 +25,7 @@ namespace Microsoft.Azure.Devices
 
     sealed class HttpClientHelper : IHttpClientHelper
     {
+        static readonly JsonSerializerSettings s_jsonSerializerSettings = new JsonSerializerSettings { DateParseHandling = DateParseHandling.DateTime, };
 #if !NETSTANDARD1_3 && !NETSTANDARD2_0
         static readonly JsonMediaTypeFormatter JsonFormatter = new JsonMediaTypeFormatter();
 #endif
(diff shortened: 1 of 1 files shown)


hanabi1224:
I believe a better fix would be changing all DateTime usage to DateTimeOffset given that timezone info is part of timezone string in actual twin

abhipsaMisra:
Thank you for submitting this PR @hanabi1224, I'll take a look at it. 
In the meanwhile, could you rebase your branch off of `master`, so that you bring in the latest changes!

hanabi1224:
@abhipsaMisra Thanks! It’s now up to date

hanabi1224:
@abhipsaMisra Kindly ping, any feedback?

abhipsaMisra:
@hanabi1224 Thanks for your patience with this, reviewing it now.

common/src/service/HttpClientHelper.cs
```diff
@@ -24,6 +25,7 @@ namespace Microsoft.Azure.Devices
 
     sealed class HttpClientHelper : IHttpClientHelper
     {
+        static readonly JsonSerializerSettings s_jsonSerializerSettings = new JsonSerializerSettings { DateParseHandling = DateParseHandling.DateTime, };
```
abhipsaMisra: The DateParseHandling property has a default value of DateTime. Could you please explain the advantage of setting it explicitly as above?

hanabi1224: @abhipsaMisra The problem is that, user can override default settings,  as described in https://github.com/Azure/azure-iot-sdk-csharp/issues/1122

abhipsaMisra: Are you setting the json serialization settings in your application? Can you share a code snippet that reproduces this issue.

hanabi1224:
```csharp
JsonConvert.DefaultSettings = () =>new JsonSerializerSettings
            {
                DateParseHandling = DateParseHandling.DateTimeOffset,
            };
```

abhipsaMisra:
This issue requires some further investigation on if we should support DateTimeOffset directly in the twin structure returned, and throughout (based on the service response). As you've pointed out above, that would be a better fix rather than forcing all applications to use DateTime mandatorily.
I'll create a separate PR addressing this bigger issue, and close this PR for now.

abhipsaMisra:
Closing this PR. This issue will be addressed in #1175

abhipsaMisra:
Thank you for

abhipsaMisra:
Thank you for highlighting this issue and submitting the PR @hanabi1224 .

This issue requires some further investigation on if we should support DateTimeOffset directly in the twin structure returned, and throughout (based on the service response). As you've pointed out above, that would be a better fix rather than forcing all applications to use DateTime mandatorily.

I've created a separate PR #1175 to fix this; you can use this to track this issue.

PR #1126, opened 2019-11-16
fixed exception mapping

davilu:
## Checklist

## Description of the changes

Modified to throw TimeoutException if operation timeout to fix break change since device client 1.19.0
Modified file upload notification E2E test to use singleton to avoid parallel file upload tests impact each other which causes test failure
Modified receive C2D message test to ignore message leftover to avoid false failure

### DIFF
--- a/e2e/test/FaultInjection.cs
+++ b/e2e/test/FaultInjection.cs
@@ -171,6 +171,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                 s_log.WriteLine($">>> {nameof(FaultInjection)} Testing baseline");
                 await testOperation(deviceClient, testDevice).ConfigureAwait(false);
 
+                int countBeforeFaultInjection = connectionStatusChangeCount;
                 watch.Start();
                 s_log.WriteLine($">>> {nameof(FaultInjection)} Testing fault handling");
                 await ActivateFaultInjection(transport, faultType, reason, delayInSec, durationInSec, deviceClient).ConfigureAwait(false);
@@ -185,7 +186,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                     bool isFaulted = false;
                     for (int i = 0; i < LatencyTimeBufferInSec; i++)
                     {
-                        if (connectionStatusChangeCount >= 2)
+                        if (connectionStatusChangeCount > countBeforeFaultInjection)
                         {
                             isFaulted = true;
                             break;
--- a/e2e/test/MessageReceiveE2ETests.cs
+++ b/e2e/test/MessageReceiveE2ETests.cs
@@ -182,14 +182,14 @@ namespace Microsoft.Azure.Devices.E2ETests
 
         public static async Task VerifyReceivedC2DMessageAsync(Client.TransportType transport, DeviceClient dc, string deviceId, string payload, string p1Value)
         {
-            var wait = true;
-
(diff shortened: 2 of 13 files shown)


e2e/test/FaultInjection.cs
```diff
@@ -185,7 +186,7 @@ public static async Task ActivateFaultInjection(Client.TransportType transport,
                     bool isFaulted = false;
                     for (int i = 0; i < LatencyTimeBufferInSec; i++)
                     {
-                        if (connectionStatusChangeCount >= 2)
+                        if (connectionStatusChangeCount > countBeforeFaultInjection)
```
bikamani: Can you elaborate why is this changed to countBeforeFaultInjection which is set to connectionStatusChangeCount ?

e2e/test/MessageReceiveE2ETests.cs
```diff
@@ -282,16 +246,31 @@ private async Task ReceiveMessageInOperationTimeout(TestDeviceType type, Client.
...
 
-                await ReceiveMessageTimeoutCheck(deviceClient).ConfigureAwait(false);
-
-                await deviceClient.CloseAsync().ConfigureAwait(false);
+                try
+                {
+                    deviceClient.OperationTimeoutInMilliseconds = Convert.ToUInt32(TIMESPAN_SIXTYFIVE_SECONDS.TotalMilliseconds);
+                    if (transport == Client.TransportType.Amqp || transport == Client.TransportType.Amqp_Tcp_Only || transport == Client.TransportType.Amqp_WebSocket_Only)
+                    {
+                        await ReceiveMessageWithoutTimeoutCheck(deviceClient, TIMESPAN_SIXTYFIVE_SECONDS).ConfigureAwait(false);
```
bikamani: Add comment for why is this different for AMQP vs MQTT for future references.

e2e/test/MessageReceiveE2ETests.cs
```diff
@@ -184,91 +180,59 @@ public static (Message message, string messageId, string payload, string p1Value
...
                     var prop = receivedMessage.Properties.Single();
                     Assert.AreEqual("property1", prop.Key, $"The key \"property1\" did not match for device {deviceId}");
                     Assert.AreEqual(p1Value, prop.Value, $"The value of \"property1\" did not match for device {deviceId}");
-
-                    await dc.CompleteAsync(receivedMessage).ConfigureAwait(false);
                     break;
                 }
 
-                if (sw.Elapsed.TotalMilliseconds > FaultInjection.RecoveryTimeMilliseconds)
+                try
```
bikamani: Code will never reach here if the message matched above.

e2e/test/MessageReceiveE2ETests.cs
```diff
@@ -184,91 +180,59 @@ public static (Message message, string messageId, string payload, string p1Value
...
+                {
+                    // always complete message
+                    await dc.CompleteAsync(receivedMessage).ConfigureAwait(false);
+                }
+                catch (Exception)
                 {
-                    throw new TimeoutException("Test is running longer than expected.");
+                    // ignore exception from CompleteAsync
```
bikamani: Aren't we supposed to check a specific i.e. locktoken exception?

Also ignore but just log it so will help with debugging.

bikamani:
@davilu  can you please add description for this PR, include summary of the changes added?

PR #1129, opened 2019-11-20
Update IoT Hub API version for PnP features.

abhipsaMisra:
Update IoT Hub API version for PnP features.

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -22,7 +22,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.29.0-preview-003</Version>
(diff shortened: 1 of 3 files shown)


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -22,7 +22,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.29.0-preview-003</Version>
+    <Version>1.29.0-preview-004</Version>
     <Title>Microsoft Azure IoT Device Client SDK</Title>
```
prmathur-microsoft: Don't we do this during release?

abhipsaMisra: Yes, but since this is the only change going into this release, I included it within the same PR

Issue #1130, opened 2019-11-20
Missing evaluation of this.receivingSemaphore.WaitAsync(linkedCts.Token) in MqttTransportHandler.cs


mma-mm:
Hi,

I'm currently investigating an issue where i receive null Messages in ReceiveAsync even when timeout has not occured yet.
While trying to find out how DeviceClient works, i stumbled about this:

In iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs:306 in Method `async Task<bool> ReceiveMessageArrivalAsync(CancellationToken cancellationToken)` there is no evaluation of the WaitAsync:

    
    using (CancellationTokenSource linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, this.disconnectAwaitersCancellationSource.Token))
            {
                try
                {
      HERE ---->     await this.receivingSemaphore.WaitAsync(linkedCts.Token).ConfigureAwait(true);
                }
                catch (OperationCanceledException)
                {
                    if (!cancellationToken.IsCancellationRequested)
                    {

In the correspondig overload in iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs:282 the WaitAsync is evaluated:

    
     try
                {
    HERE--->         hasMessage = await this.receivingSemaphore.WaitAsync(timeout, linkedCts.Token).ConfigureAwait(true);
                }
                catch (OperationCanceledException)
                {
                    if (disconnectToken.IsCancellationRequested)
                    {

It might be able that messages are lost this way.

davilu:
@mma-mm  Whenever there is message arrived `receivingSemaphore.Release()` is called on line 490 which will increase the count. While `ReceiveAsync` is called, it always acquire from `receivingSemaphore`. It could fail either device is disconnected or timeout. Otherwise  it will decrease the count and return a message. Could you please give an example of message lost?

mma-mm:
What i'm observing:

In my Application im Receiveing Messages on the same connection object at 2 different places, once using ReceiveAsync with a Timeout set using a TimeSpan, once by handing over a CancellationToken. The one using a TimeSpan works flawlessly, receives all messages. The one using hte CancellationToken receives only null messages. I'm quite sure these are not timeouts, as is can control the sending of C2D-Messages. The null messages occur after the default timeout, or if i press the corresponding button in the cloud. 

The both calls to ReceiveAsync don't take place at the same time. 

[...]

PR #1136, opened 2019-11-22
fixed mqtt receive message with cancellation token and added e2e test

davilu:
## Checklist

## Description of the changes

fixed mqtt receive message with cancellation token and added e2e test
fixed amqp dispose exception when open failed

#1137 and #1130

### DIFF
--- a/common/test/TestDevice.cs
+++ b/common/test/TestDevice.cs
@@ -201,7 +201,7 @@ namespace Microsoft.Azure.Devices.E2ETests
             return deviceClient;
         }
 
-        private static string GetHostName(string iotHubConnectionString)
+        public static string GetHostName(string iotHubConnectionString)
         {
             Regex regex = new Regex("HostName=([^;]+)", RegexOptions.None);
             return regex.Match(iotHubConnectionString).Groups[1].Value;
--- a/e2e/test/DeviceClientX509AuthenticationE2ETests.cs
+++ b/e2e/test/DeviceClientX509AuthenticationE2ETests.cs
@@ -28,11 +28,81 @@ namespace Microsoft.Azure.Devices.E2ETests
         }
 
         
-
         [TestMethod]
         public async Task X509_InvalidDeviceId_Amqp()
         {
-            var deviceClient = CreateDeviceClientWithInvalidId(Client.TransportType.Amqp_Tcp_Only);
+            await X509InvalidDeviceIdOpenAsyncTest(Client.TransportType.Amqp).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Amqp_Tcp()
+        {
+            await X509InvalidDeviceIdOpenAsyncTest(Client.TransportType.Amqp_Tcp_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Amqp_WebSocket()
+        {
+            await X509InvalidDeviceIdOpenAsyncTest(Client.TransportType.Amqp_WebSocket_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Mqtt()
+        {
+            await X509InvalidDeviceIdOpenAsyncTest(Client.TransportType.Mqtt).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Mqtt_Tcp()
+        {
+            await X509InvalidDeviceIdOpenAsyncTest(Client.TransportType.Mqtt_Tcp_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Mqtt_WebSocket()
+        {
+            await X509InvalidDeviceIdOpenAsyncTest(Client.TransportType.Mqtt_WebSocket_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Twice_Amqp()
+        {
+            await X509InvalidDeviceIdOpenAsyncTwiceTest(Client.TransportType.Amqp).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Twice_Amqp_TCP()
+        {
+            await X509InvalidDeviceIdOpenAsyncTwiceTest(Client.TransportType.Amqp_Tcp_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Twice_Amqp_WebSocket()
+        {
+            await X509InvalidDeviceIdOpenAsyncTwiceTest(Client.TransportType.Amqp_WebSocket_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Twice_Mqtt()
+        {
+            await X509InvalidDeviceIdOpenAsyncTwiceTest(Client.TransportType.Mqtt).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Twice_Mqtt_Tcp()
+        {
+            await X509InvalidDeviceIdOpenAsyncTwiceTest(Client.TransportType.Mqtt_Tcp_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Twice_Mqtt_WebSocket()
+        {
+            await X509InvalidDeviceIdOpenAsyncTwiceTest(Client.TransportType.Mqtt_WebSocket_Only).ConfigureAwait(false);
+        }
+
+        private async Task X509InvalidDeviceIdOpenAsyncTest(Client.TransportType transportType)
+        {
+            var deviceClient = CreateDeviceClientWithInvalidId(transportType);
             using (deviceClient)
             {
                 try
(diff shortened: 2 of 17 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -621,7 +697,14 @@ public Task SendEventBatchAsync(IEnumerable<Message> messages, CancellationToken
             }
 
             // Codes_SRS_DEVICECLIENT_28_019: [The async operation shall retry until time specified in OperationTimeoutInMilliseconds property expire or unrecoverable error(authentication or quota exceed) occurs.]
-            return InnerHandler.SendEventAsync(messages, cancellationToken);
+            try
+            {
```
bikamani: Can you not do this before calling sendeventasync?

e2e/test/MessageReceiveE2ETests.cs
```diff
@@ -234,6 +295,53 @@ public static async Task VerifyReceivedC2DMessageAsync(Client.TransportType tran
...
+                {
+                    // always complete message
+                    await dc.CompleteAsync(receivedMessage).ConfigureAwait(false);
+                }
+                catch (Exception)
+                {
+                    // ignore exception from CompleteAsync
+                }
+
+                string messageData = Encoding.ASCII.GetString(receivedMessage.GetBytes());
```
bikamani: duplicate code, can be shared with other verify tasks

e2e/test/MessageReceiveE2ETests.cs
```diff
@@ -234,6 +295,53 @@ public static async Task VerifyReceivedC2DMessageAsync(Client.TransportType tran
...
+            sw.Start();
+
+            Client.Message receivedMessage;
+
+            while (!received && sw.ElapsedMilliseconds < FaultInjection.RecoveryTimeMilliseconds)
+            {
+                _log.WriteLine($"Receiving messages for device {deviceId}.");
+
+                
+                receivedMessage = await dc.ReceiveAsync(new CancellationTokenSource(TIMESPAN_ONE_MINUTE).Token).ConfigureAwait(false);
```
bikamani: Why does this need to be 1 minute?
Can it be smaller value? As this is adding more time for tests. 

Adding any comments for the same will be helpful.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -330,29 +281,7 @@ public override async Task CompleteAsync(string lockToken, CancellationToken can
...
-                    lockToken.Length != actualLockToken.Length + GenerationPrefixLength)
-                {
-                    throw new IotHubException(
-                        $"Client MUST send PUBACK packets in the order in which the corresponding PUBLISH packets were received (QoS 1 messages) per [MQTT-4.6.0-2]. Expected lock token: '{actualLockToken}'; actual lock token: '{lockToken}'.",
-                        isTransient:false);
-                }
-
-                this.completionQueue.Dequeue();
-                completeOperationCompletion = this.channel.WriteAndFlushAsync(actualLockToken);
+                completeOperationCompletion = this.channel.WriteAndFlushAsync(lockToken);
```
bikamani: This change requires more test to be added. Add test or hold on this change.

e2e/test/MessageReceiveE2ETests.cs
davilu: No it won't add more time. If C2D function is good, it should be fast, otherwise it will get a `null` and fail the test in a minute. There is no SLA for C2D latency so one minute is the least interval we should wait.

e2e/test/MessageReceiveE2ETests.cs
davilu: Sure. It could be extracted. But it is coupling with `while` loop. It could be extracted as `bool VerifyMessage(Client.Message message, string payload, string payloadKey, string payloadValue)`. It will return `false` if `payload` doesn't match.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
davilu: Reverted. There is more code coupling together. Will leave as it is.

iothub/device/src/InternalClient.cs
davilu: No you can't. But it's a good idea add  `cancellationToken.ThrowIfCancellationRequested()` once cause is operation cancelled otherwise will just throw exception as it is.


`

e2e/test/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -0,0 +1,160 @@
...
+        private static readonly TimeSpan TIMESPAN_FIVE_SECONDS = TimeSpan.FromSeconds(5);
+        private static readonly TimeSpan TIMESPAN_TWENDY_SECONDS = TimeSpan.FromSeconds(20);
+
+        private readonly ConsoleEventListener _listener;
+        private readonly string _hostName;
+
+        public DeviceClientX509AuthenticationE2ETests()
+        {
+            _listener = TestConfig.StartEventListener();
+            _hostName = TestDevice.GetHostName(Configuration.IoTHub.ConnectionString);
```
bikamani: Why not use testDevice.IoTHubHostName instead at line 152?

e2e/test/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Amqp_Tcp()
+        {
+            await X509InvalidDeviceIdOpenAsyncTest(Client.TransportType.Amqp_Tcp_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Amqp_WebSocket()
```
bikamani: Should the test names be more specific to the operation it's performing underneath? That will be more readable for all of these tests.

e2e/test/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -0,0 +1,160 @@
...
+                    try
+                    {
+                        await deviceClient.OpenAsync().ConfigureAwait(false);
+                        Assert.Fail("Should throw UnauthorizedException but didn't.");
+                    }
+                    catch (UnauthorizedException)
+                    {
+                        // It should always throw UnauthorizedException
+                    }
+                }
```
bikamani: redundant code in task above, can be extracted. Also can you explain the reason of doing it twice? probably hard code value 2 if it makes more sense readability wise.

e2e/test/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Mqtt_WebSocket()
+        {
+            await X509InvalidDeviceIdOpenAsyncTest(Client.TransportType.Mqtt_WebSocket_Only).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task X509_InvalidDeviceId_Twice_Amqp()
```
bikamani: What is this test for? can be named better based on the scenario tested.

e2e/test/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -0,0 +1,160 @@
...
+                    Assert.Fail("Should throw UnauthorizedException but didn't.");
+                }
+                catch (UnauthorizedException)
+                {
+                    // It should always throw UnauthorizedException
+                }
+
+                // Check TCP connection to verify there is no connection leak
+                // netstat -na | find "[Your Hub IP]" | find "ESTABLISHED" 
+                await Task.Delay(TimeSpan.FromSeconds(10)).ConfigureAwait(false);
```
bikamani: just wondering what is purpose of adding delay here?

common/test/TestDevice.cs
```diff
@@ -201,7 +201,7 @@ public DeviceClient CreateDeviceClient(ITransportSettings[] transportSettings, C
             return deviceClient;
         }
 
-        private static string GetHostName(string iotHubConnectionString)
+        public static string GetHostName(string iotHubConnectionString)
```
bikamani: This change may not be required. Follow up comments in DeviceClientX509AuthenticationE2ETests.cs

e2e/test/MessageFeedbackE2ETests.cs
```diff
@@ -0,0 +1,77 @@
...
+                    if (message == null)
+                    {
+                        Assert.Fail("No message received.");
+                    }
+                    messages.Add(message);
+                }
+
+                for (int i = 0; i < MESSAGE_COUNT; i++)
+                {
+                    await deviceClient.CompleteAsync(messages[MESSAGE_COUNT - 1 - i]).ConfigureAwait(false);
```
bikamani: Are we sure at this point if the complete for messages is done within 1 minute of interval? Because this may result in locklost Exception otherwise.

e2e/test/MessageFeedbackE2ETests.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Diagnostics.Tracing;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    [TestClass]
+    [TestCategory("IoTHub-E2E")]
+    [Ignore]
+    // TODO MQTT OrderedTwoPhaseWorkQueue disallow message feedback to be called mix order, enable this test once it's fixed
```
bikamani: Correct.

But are we adding tests for AMQP though?

e2e/test/MessageFeedbackE2ETests.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        [TestMethod]
+        public async Task Message_CompleteMixOrder_Mqtt()
+        {
+            await CompleteMessageMixOrder(TestDeviceType.Sasl, Client.TransportType.Mqtt_Tcp_Only).ConfigureAwait(false);
+        }
+
+        private static async Task CompleteMessageMixOrder(TestDeviceType type, Client.TransportType transport)
+        {
```
bikamani: Do we need to start a stop watch to keep some time check for the test for worst case scenarios?

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -181,8 +180,9 @@ private void Cleanup()
             _amqpAuthenticationRefresher?.StopLoop();
             if (!_deviceIdentity.IsPooling())
             {
-                _amqpConnectionHolder?.Dispose();
+                _amqpConnectionHolder?.Shutdown();
```
bikamani: Can it instead call RemoveAmqpUnit?

e2e/test/MessageFeedbackE2ETests.cs
davilu: Add timeout with operation timeout later.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
davilu: To avoid to call upper layer(cycle dependency).

PR #1142, opened 2019-11-26
clarified sendeventbatchasync docs

gregman-msft:
docs:clarified SendEventBatchAsync requires AMQP over WebSockets.

### DIFF
--- a/iothub/device/src/DeviceClient.cs
+++ b/iothub/device/src/DeviceClient.cs
@@ -394,14 +394,14 @@ namespace Microsoft.Azure.Devices.Client
 
 
         /// <summary>
-        /// Sends an event to device hub. Requires AMQP over WebSockets.
+        /// Sends an event to device hub
         /// </summary>
         /// <param name="message">The message.</param>
         /// <returns>The message containing the event</returns>
         public Task SendEventAsync(Message message) => this.internalClient.SendEventAsync(message);
 
         /// <summary>
-        /// Sends an event to device hub. Requires AMQP over WebSockets.
+        /// Sends an event to device hub
         /// </summary>
         /// <param name="message">The message.</param>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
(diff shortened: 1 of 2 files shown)


iothub/device/src/DeviceClient.cs
```diff
@@ -394,14 +394,14 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
 
 
         /// <summary>
-        /// Sends an event to device hub
+        /// Sends an event to device hub. Requires AMQP over WebSockets.
         /// </summary>
```
prmathur-microsoft: This is generic one which works on all protocols. Batching is the one that works only on AMQP. Please revert this.

iothub/device/src/DeviceClient.cs
```diff
@@ -394,14 +394,14 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
         /// <param name="message">The message.</param>
         /// <returns>The message containing the event</returns>
         public Task SendEventAsync(Message message) => this.internalClient.SendEventAsync(message);
 
         /// <summary>
-        /// Sends an event to device hub
+        /// Sends an event to device hub. Requires AMQP over WebSockets.
         /// </summary>
```
prmathur-microsoft: Same comment as above.

iothub/device/src/ModuleClient.cs
```diff
@@ -349,14 +349,14 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
         public Task SendEventAsync(Message message, CancellationToken cancellationToken) => this.internalClient.SendEventAsync(message, cancellationToken);
 
         /// <summary>
-        /// Sends a batch of events to device hub
+        /// Sends a batch of events to device hub. Requires AMQP over WebSockets.
         /// </summary>
```
prmathur-microsoft: AMQP or AMQP over websocket.

@abhipsaMisra - Do we have E2E for this? If not can you add?.

abhipsaMisra: Currently we don't have any E2E for SendEventBatchAsync. Tracked in backlog.

PR #1143, opened 2019-11-26
Decreased MQTT packet id range to 15 bit. 

vipeller:
The reason is that Publish packets clear the highest (15th) bit of the counter in some scenarios, and so packetId = 0x8000 becomes 0x0000 which in turn is an invalid value for a packet id.

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -900,7 +900,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             Debug.Assert(lockTaken);
             unchecked
             {
-                ret = ++_packetId == 0 ? ++_packetId : _packetId;
+                _packetId = (ushort)(++_packetId % 0x8000);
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -900,7 +900,8 @@ private ushort GetNextPacketId()
             Debug.Assert(lockTaken);
             unchecked
             {
-                ret = ++_packetId == 0 ? ++_packetId : _packetId;
+                _packetId = (ushort)(++_packetId % 0x8000);
```
davilu: Should better use `++packetId&0x7FFF` instead. `%` is much worse than `&` to permanence concern.

vipeller: for me, "Modulo" expresses more the intention that "the result should not be greater than" - as a clock arithmetic,  it is usual to use it to restrict numbers into a certain range and I think it is more readable because of this (I even considered using an "if" to make the code less cryptic :) ) But probably it is more like a matter of preference. As this is not a central peace of a loop calculating something, I would not worry about the performance hit.

PR #1152, opened 2019-12-09
Merge Digital Twin into preview branch

timtay-microsoft:
Strictly merges with no extra code added. The only caveat is that the digital twin samples will not be brought along with the rest of the code. Those samples will live in a samples repo.

See [this commit](https://github.com/Azure/azure-iot-sdk-csharp/commit/39b9ed9d350d3b5ae8132cbc4adeb612ef94d86c) in this PR for the details surrounding removing the samples


abhipsaMisra:
Why are all of the commits repeated (with different hashes)?

PR #1156, opened 2019-12-13
Fix ASC tests - mark test methods as async

abhipsaMisra:
The ASC tests were not marked as async, so the dotnet tool was not picking them up correctly.


abhipsaMisra:
Closing this since this will be tracked in #1157

PR #1157, opened 2019-12-13
Fix ASC tests - mark test methods as async

abhipsaMisra:
The ASC tests were not marked as async, so the dotnet tool was not picking them up correctly.


davilu:
The changes looks fine but the test seems to be invalid.

abhipsaMisra:
> The changes looks fine but the test seems to be invalid.

Invalid, how so?

abhipsaMisra:
Closing this PR, since it is allowed to return the Task to the TestMethod. The method being returned is awaiting the actual async operation.

PR #1165, opened 2019-12-27
Enable some tests in ConnectionStatusChangeHandlerTests

shantanu1singh:
## Checklist

## Description of the changes
Enable some tests in ConnectionStatusChangeHandlerTests.

Removing the code blocks that were attempting to assert whether the ConnectionStatusChangeHandler is called multiple times when a new API on the client is invoked (after the deletion/disablement of a device). This is because the InternalClient.cs class has logic to prevent the triggering of the ConnectionStatusChangeHandler multiple times when the last connection status is the same as the new status.


shantanu1singh:
@davilu can you please help review this change?

prmathur-microsoft:
Can you update this PR to closure? It has been sitting for a long time without progress

PR #1024, opened 2019-08-08
Updated the ToJson method to ensure backward compatibility with Newtonsoft.Json 9.0.1

wmmihaa:
## Checklist
- [X] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/.github/CONTRIBUTING.md).
- [ ] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
- [X] This pull-request is submitted against the `master` branch.

## Description of the changes
Using Microsoft.Devices.* from an Azure Function (.Net Framwork) is not working due to the built in dependency to Newtonsoft.Json v9.0.1 in Azure Functions (v1). 

In the context of the features used by Microsoft.Devices and Microsoft.Devices.Shared, -Newtonsoft.Json v10.0.3 is backward compatible to 9.0.1 apart from the *Newtonsoft.Json.Formating* used in the **ToJson** methods in Twin.cs and TwinCollection.cs in the Microsoft.Devices.Shared project.

To resolve the issue, I've changed the *ToJson* method from:
```csharp
/// <summary>
/// Gets the Twin as a JSON string
/// </summary>
/// <param name="formatting">Optional. Formatting for the output JSON string.</param>
/// <returns>JSON string</returns>
public string ToJson(Formatting formatting = Formatting.None)
{
    return JsonConvert.SerializeObject(this, formatting);
}
```

to...

```csharp
/// <summary>
/// Gets the Twin as a JSON string
/// </summary>
/// <param name="formatting">Optional. Formatting for the output JSON string.</param>
/// <returns>JSON string</returns>
public string ToJson(Formatting formatting)
{
    return JsonConvert.SerializeObject(this, formatting);
}
/// <summary>
/// Gets the Twin as a JSON string
/// </summary>
/// <returns>JSON string</returns>
public string ToJson()
{
    return JsonConvert.SerializeObject(this);
}
```
With this change you can downgrade the Newtonsoft package to 9.0.1 for net451 (**Not part of this PR**) and ensure the Microsoft.Azure.Devices works for Azure Functions v1.

Cheers
Mikael

Issue #940, opened 2019-05-10
upgrade 'newtonsoft.json' to latest version


seanmcd-msft:
I’m intermittently hitting a bug with a stack that looks like the below.  there can't be a dupe since there's only 1 item in the collection.  i also hit a variant where the same code complains about a null ref when i can't be passing a null since the parm i'm passing is local and i print it right before the call. this appears to be a newtonsoft bug that was fixed back in November.
`
2019/05/06 18:32:01.463020 :                                    ConnectionBase UpdateObject lambda exception System.ArgumentException: Can not add property Orientation to Newtonso
ft.Json.Linq.JObject. Property with the same name already exists on object.
   at Newtonsoft.Json.Linq.JObject.ValidateToken(JToken o, JToken existing)
   at Newtonsoft.Json.Linq.JContainer.InsertItem(Int32 index, JToken item, Boolean skipParentCheck)
   at Microsoft.Azure.Devices.Shared.TwinCollection.TrySetMemberInternal(String propertyName, Object value)
   at CallSite.Target(Closure , CallSite , TwinCollection , String , Object )
   at System.Dynamic.UpdateDelegates.UpdateAndExecuteVoid3[T0,T1,T2](CallSite site, T0 arg0, T1 arg1, T2 arg2)
   at Microsoft.Azure.Devices.Shared.TwinCollection.set_Item(String propertyName, Object value)
   at EdgeModuleSamples.Common.Azure.AzureModuleBase.UpdateReportedPropertiesAsync(KeyValuePair``2 u) in D:\src\Windows-iotcore-samples\Samples\EdgeModules\Common\CS\AzureConnectio
n.cs:line 118
   at EdgeModuleSamples.Common.Azure.AzureConnectionBase.<UpdateObject>b__17_0() in D:\src\Windows-iotcore-samples\Samples\EdgeModules\Common\CS\AzureConnection.cs:line 398
`

my code looks like:

            TwinCollection delta = new TwinCollection();
            delta[u.Key] = u.Value;
            Log.WriteLine("updating twin reported properties with key = {0}, vt = {1}:{2}", u.Key, u.Value.GetType(), u.Value.ToString());
            await _moduleClient.UpdateReportedPropertiesAsync(delta).ConfigureAwait(false);

Please upgrade the SDK  from v10.x to latest(v12).

timtay-microsoft:
[...]

PR #1177, opened 2020-01-09
Upgrade newtonsoft dependency

timtay-microsoft:
Issue #940 fix combined with PR #1024 (code courtesy of @wmmihaa)


shared/src/Twin.cs
```diff
@@ -147,9 +147,17 @@ public Twin(TwinProperties twinProperties)
         /// </summary>
         /// <param name="formatting">Optional. Formatting for the output JSON string.</param>
         /// <returns>JSON string</returns>
-        public string ToJson(Formatting formatting = Formatting.None)
+        public string ToJson(Formatting formatting)
         {
             return JsonConvert.SerializeObject(this, formatting);
         }
+        /// <summary>
```
abhipsaMisra: Looking at the issue details from #1024 , why are we modifying the SDK to be compatible with 9.0.1, instead of having the referencing application use the updated version of NewtonSoft json. That way, the responsibility of backward-compatibility rests on NewtonSoft json directly.

timtay-microsoft: Okay, I'll re-open the other PR and just let this PR focus on upgrading the existing dependency

PR #1174, opened 2020-01-09
Restrict 4.5.1 to TLS 1.2, and centralize version choice

drwill-ms:
## Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/.github/CONTRIBUTING.md).
- [x] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
- [x] This pull-request is submitted against the `master` branch.

## Description of the changes
Retarget .NET Framework 4.5.1 code to only specify TLS 1.2 (removing 1.0 and 1.1). It does not have a SystemDefault setting, which would be preferred, so we must be explicit. Note this is one of many reasons why 4.5.1 support should be deprecated.

## Reference/Link to the issue solved with this PR (if any)
6024224

PR #1178, opened 2020-01-09
fix(iot-service): Fix proxy handling for requests with per-request timeouts

timtay-microsoft:
The http client helper class defined two different http client instances where the only difference was the client level timeout. Because of this, one of those http clients did not respect the proxy settings, if any were provided.

This PR consolidates the logic such that only one http client is maintained, and that one client does respect the proxy settings if any are provided

Now an http client is created for each request. This is done because the httpClient.Timeout property will be set before each request, but the http client library will throw if that property is changed after the client has been used to send a request. The alternative is to just have two httpClients, but that was what we had already and it was very confusing and prone to being inconsistently maintained

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -30,10 +30,10 @@ namespace Microsoft.Azure.Devices
         readonly Uri baseAddress;
         readonly IAuthorizationHeaderProvider authenticationHeaderProvider;
         readonly IReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping;
-        HttpClient httpClientObj;
-        HttpClient httpClientObjWithPerRequestTimeout;
-        bool isDisposed;
         readonly TimeSpan defaultOperationTimeout;
+        readonly IWebProxy customHttpProxy;
+        readonly Action<HttpClient> preRequestActionForAllRequests;
+
 
         public HttpClientHelper(
             Uri baseAddress,
@@ -49,34 +49,7 @@ namespace Microsoft.Azure.Devices
                 new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
             this.defaultOperationTimeout = timeout;
 
-            if (customHttpProxy != DefaultWebProxySettings.Instance)
-            {
-                HttpClientHandler httpClientHandler = new HttpClientHandler();
-                httpClientHandler.UseProxy = (customHttpProxy != null);
-                httpClientHandler.Proxy = customHttpProxy;
-                this.httpClientObj = new HttpClient(httpClientHandler);
-            }
-            else
-            {
-                this.httpClientObj = new HttpClient();
-            }
 
-            this.httpClientObj.BaseAddress = this.baseAddress;
-            this.httpClientObj.Timeout = timeout;
-            this.httpClientObj.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(CommonConstants.MediaTypeForDeviceManagementApis));
-            this.httpClientObj.DefaultRequestHeaders.ExpectContinue = false;
-
-            this.httpClientObjWithPerRequestTimeout = new HttpClient();
-            this.httpClientObjWithPerRequestTimeout.BaseAddress = this.baseAddress;
-            this.httpClientObjWithPerRequestTimeout.Timeout = Timeout.InfiniteTimeSpan;
-            this.httpClientObjWithPerRequestTimeout.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(CommonConstants.MediaTypeForDeviceManagementApis));
-            this.httpClientObjWithPerRequestTimeout.DefaultRequestHeaders.ExpectContinue = false;
-
-            if (preRequestActionForAllRequests != null)
-            {
-                preRequestActionForAllRequests(this.httpClientObj);
-                preRequestActionForAllRequests(this.httpClientObjWithPerRequestTimeout);
-            }
         }
 
         public Task<T> GetAsync<T>(
@@ -112,7 +85,7 @@ namespace Microsoft.Azure.Devices
             {
                 if (throwIfNotFound)
                 {
-                    await this.ExecuteWithOperationTimeoutAsync(
+                    await this.ExecuteWithCustomOperationTimeoutAsync(
                         HttpMethod.Get,
                         new Uri(this.baseAddress, requestUri),
                         operationTimeout,
@@ -124,7 +97,7 @@ namespace Microsoft.Azure.Devices
                 }
                 else
                 {
-                    await this.ExecuteWithOperationTimeoutAsync(
+                    await this.ExecuteWithCustomOperationTimeoutAsync(
                        HttpMethod.Get,
                        new Uri(this.baseAddress, requestUri),
                         operationTimeout,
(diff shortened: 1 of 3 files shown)


common/src/service/HttpClientHelper.cs
```diff
@@ -763,6 +751,28 @@ public static bool IsMappedToException(HttpResponseMessage message)
...
+        async Task ExecuteAsync(
+            HttpClient httpClient,
+            HttpMethod httpMethod,
+            Uri requestUri,
+            Func<HttpRequestMessage, CancellationToken, Task> modifyRequestMessageAsync,
+            Func<HttpResponseMessage, bool> isMappedToException,
+            Func<HttpResponseMessage, CancellationToken, Task> processResponseMessageAsync,
+            IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> errorMappingOverrides,
+            bool infinteClientTimeout,
+            CancellationToken cancellationToken)
```
abhipsaMisra: What is the difference in behavior between the operation timeout vs cancellation token? In the `ExecuteAsync` method above, the operation timeout passed is used to create a cancellation token source, which is then passed to the HTTP request: https://github.com/Azure/azure-iot-sdk-csharp/pull/1178/files#diff-4e06d203c3b4754233cdbe29cc0ccfc8L724

e2e/test/ServiceClientE2ETests.cs
```diff
@@ -37,6 +41,58 @@ public async Task Message_NoTimeoutPassed()
...
+                    HttpProxy = new WebProxy(Configuration.IoTHub.ProxyServerAddress)
+                }).ConfigureAwait(false);
+        }
+
+        private async Task ServiceClient_InvokeMethodWithCustomTimeout(TransportType transportType)
+        {
+            await ServiceClient_InvokeMethodWithCustomTimeout(TransportType.Amqp_WebSocket_Only, new ServiceClientTransportSettings()).ConfigureAwait(false);
+        }
+
+        private async Task ServiceClient_InvokeMethodWithCustomTimeout(TransportType transportType, ServiceClientTransportSettings transportSettings)
```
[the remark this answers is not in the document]
timtay-microsoft: That's a bit out of scope for now. This PR is solely focused on bringing proxy support to that code path

e2e/test/ServiceClientE2ETests.cs
```diff
@@ -14,6 +15,9 @@ namespace Microsoft.Azure.Devices.E2ETests
     [TestCategory("IoTHub-E2E")]
     public class ServiceClientE2ETests : IDisposable
     {
+        private readonly string DirectMethodName = $"{nameof(ServiceClientE2ETests)}Method";
+        private TimeSpan CustomTimeout = TimeSpan.FromMinutes(5);
```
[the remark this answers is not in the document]
timtay-microsoft: I redid these e2e tests, but we do still need a long timeout in one of these tests because that is the only way to exercise the code path that this PR is bringing proxy support to. Any method invocation with a read timeout + connect timeout that is less than the default operation timeout will use the default operation timeout code path.

e2e/test/ServiceClientE2ETests.cs
```diff
@@ -37,6 +41,58 @@ public async Task Message_NoTimeoutPassed()
...
+
+        [TestMethod]
+        public async Task ServiceClient_InvokeMethodWithCustomTimeout_Amqp_WithProxy()
+        {
+            await ServiceClient_InvokeMethodWithCustomTimeout(
+                TransportType.Amqp,
+                new ServiceClientTransportSettings()
+                {
+                    AmqpProxy = new WebProxy(Configuration.IoTHub.ProxyServerAddress),
+                    HttpProxy = new WebProxy(Configuration.IoTHub.ProxyServerAddress)
```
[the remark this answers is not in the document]
timtay-microsoft: I redid the e2e tests for clarity. You are right, the code being fixed in this PR is only HTTP code, so the AMQP proxy didn't make sense.

e2e/test/ServiceClientE2ETests.cs
```diff
@@ -37,6 +41,58 @@ public async Task Message_NoTimeoutPassed()
...
+
+        [TestMethod]
+        public async Task ServiceClient_InvokeMethodWithCustomTimeout_AmqpWs_WithProxy()
+        {
+            await ServiceClient_InvokeMethodWithCustomTimeout(
+                TransportType.Amqp_WebSocket_Only,
+                new ServiceClientTransportSettings()
+                {
+                    AmqpProxy = new WebProxy(Configuration.IoTHub.ProxyServerAddress),
+                    HttpProxy = new WebProxy(Configuration.IoTHub.ProxyServerAddress)
```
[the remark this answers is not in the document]
timtay-microsoft: I redid the e2e tests for clarity. You are right, the code being fixed in this PR is only HTTP code, so the AMQP proxy didn't make sense.

timtay-microsoft:
Hold off on merging this pr until #1174 is checked in. That PR also makes changes to the httpclienthelper class, and we want to avoid merge conflicts

abhipsaMisra:
Device Client HTTP Client for reference: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/HttpClientHelper.cs#L111

timtay-microsoft:
> @timtay-microsoft FYI, Device Client HTTP Client for reference: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/HttpClientHelper.cs#L111

So creating a single http client works for device client because it never needs to assign different values to that ```httpClientObj.Timeout``` property. In this service client implementation, we do need to assign different values here if we only want to maintain one http client. The best solution is to create and dispose one http client per request

common/src/service/HttpClientHelper.cs
```diff
@@ -149,15 +122,18 @@ sealed class HttpClientHelper : IHttpClientHelper
...
-                       new Uri(this.baseAddress, requestUri),
-                       (requestMsg, token) => AddCustomHeaders(requestMsg, customHeaders),
-                       message => !(message.IsSuccessStatusCode || message.StatusCode == HttpStatusCode.NotFound),
-                       async (message, token) => result = message.StatusCode == HttpStatusCode.NotFound ? (default(T)) : await ReadResponseMessageAsync<T>(message, token).ConfigureAwait(false),
-                       errorMappingOverrides,
-                       cancellationToken).ConfigureAwait(false);
+                    using (var httpClient = buildHttpClient(defaultOperationTimeout))
+                    {
+                        await this.ExecuteAsync(
+                            httpClient,
```
drwill-ms: there is an extra space before `httpClient,`

common/src/service/HttpClientHelper.cs
```diff
@@ -863,18 +845,33 @@ public static bool IsMappedToException(HttpResponseMessage message)
             return await exception.ConfigureAwait(false);
         }
 
-        public void Dispose()
+        private HttpClient buildHttpClient(TimeSpan timeout)
```
drwill-ms: Not Java, TitleCasedMethods. :)

common/src/service/HttpClientHelper.cs
timtay-microsoft: I'll remove it, good catch

common/src/service/HttpClientHelper.cs
timtay-microsoft: BuildHttpClient instead? I can do that

e2e/test/MethodE2ETests.cs
```diff
@@ -102,6 +105,28 @@ public async Task Method_DeviceReceivesMethodAndResponseWithDefaultMethodHandler
...
+                HttpProxy = new WebProxy(Configuration.IoTHub.ProxyServerAddress)
+            };
+
+            await SendMethodAndRespond(Client.TransportType.Mqtt_Tcp_Only, SetDeviceReceiveMethod, serviceClientTransportSettings).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task Method_ServiceSendsMethodThroughProxyWithCustomTimeout()
+        {
+            ServiceClientTransportSettings serviceClientTransportSettings = new ServiceClientTransportSettings()
```
drwill-ms: When you use an initializer list and don't send ctor params, you may leave out the parenthesis.
Also, when the type is obvious, we prefer `var`.

```
var serviceClientTransportSettings = new ServiceClientTransportSettings
{
    HttpProxy = ...,
};
```

e2e/test/MethodE2ETests.cs
```diff
@@ -23,6 +24,8 @@ public class MethodE2ETests : IDisposable
         private const string MethodName = "MethodE2ETest";
         private static TestLogging _log = TestLogging.GetInstance();
 
+        private const double defaultMethodTimeoutMinutes = 1;
```
drwill-ms: Instead of a double, consider making it a static readonly TimeSpan, with value of `TimeSpan.FromMinutes(1)`

timtay-microsoft: Good idea

e2e/test/MethodE2ETests.cs
timtay-microsoft: Sure thing

PR #1196, opened 2020-01-17
Add additional logging info for E2E tests

abhipsaMisra:
This PR contains changes for the 2 tests that have been failing quite consistently in the past:
1. `ReceiveMessageInOperationTimeout` test - this was failing since the `ReceiveAsync` sometimes takes a couple of ms over a min to return. I've added an additional second of wait time, and also added logs that prints the time the operation actually took to return.
2. `Message_TimeOutReachedResponse` test - this test sometimes fails to timeout in 1 tick. I am unable to repro this in local runs, so I've increased the timeout to 5 ticks (can be reverted), and also added logs to indicate when the operation actually completes/ throws and exception.

These changes should give us further information if the tests fail in future runs.

### DIFF
--- a/e2e/test/MessageReceiveE2ETests.cs
+++ b/e2e/test/MessageReceiveE2ETests.cs
@@ -473,7 +473,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                     TimeSpan maxLatency = TimeSpan.FromMilliseconds(dc.OperationTimeoutInMilliseconds) + bufferTime;
                     if (sw.Elapsed > maxLatency)
                     {
-                        Assert.Fail($"ReceiveAsync did not return in {maxLatency}.");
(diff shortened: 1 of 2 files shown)


abhipsaMisra:
@davilu : Could you take a look at this PR. It contains changes for the 2 tests that have been failing quite consistently in the past:
1. `ReceiveMessageInOperationTimeout` test - this was failing since the `ReceiveAsync` sometimes takes a couple of ms over a min to return. I've added an additional second of wait time, and also added logs that prints the time the operation actually took to return.
2. `Message_TimeOutReachedResponse` test - this test sometimes fails to timeout in 1 tick. I am unable to repro this in local runs, so I've increased the timeout to 5 ticks (can be reverted), and also added logs to indicate when the operation actually completes/ throws and exception.

These changes should give us further information if the tests fail in future runs.

PR #1199, opened 2020-01-23
Forward Integration Merge from master to preview

timtay-microsoft:
This PR brings all the latest code from master into preview

commit aa3094d12f2a978a89a85b89e15bf0179dda0828 doesn't show in the list of commits, but its changes are captured in the merge commit

Also including a commit that bumps the newtonsoft and csharp dependency versions of digital twin device client and its unit tests so that it is consistent with all other projects that have that dependency

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,41 @@
+<!-- BEGIN MICROSOFT SECURITY.MD V0.0.3 BLOCK -->
+
+## Security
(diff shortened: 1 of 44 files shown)


timtay-microsoft:
Closing this PR because the FI merge got messed up, and a new PR will be opened with it and the bump versions needed to release

PR #1163, opened 2019-12-20
Update E2E tests to optionally run against signed packages

abhipsaMisra:

PR #1200, opened 2020-01-24
Forward Integration Merge from master to preview, Bump versions for 1/24/2020 preview release

timtay-microsoft:
This PR brings all the latest code from master into preview

Had to add a few commits to make the eventhub listener changes work for the preview branch only test classes that were using it

Also including a commit that bumps the newtonsoft and csharp dependency versions of digital twin device client and its unit tests so that it is consistent with all other projects that have that dependency

I'm also including the bump versions commit since we will release preview from this point

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,41 @@
+<!-- BEGIN MICROSOFT SECURITY.MD V0.0.3 BLOCK -->
+
+## Security
+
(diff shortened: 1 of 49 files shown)


abhipsaMisra:
@timtay-microsoft FYI - Once we have the updates from `master`, #1163 needs to be merged into preview (before starting the release).

timtay-microsoft:
> @timtay-microsoft FYI - Once we have the updates from `master`, #1163 needs to be merged into preview (before starting the release).

Gotcha, I'll leave out the bump versions from this PR then. We can bump the versions after we get your PR merged in, too

timtay-microsoft:
#1205 instead

PR #1211, opened 2020-01-31
PR build refactor and TlsVersion instance change

drwill-ms:
This is a start on the Windows script.

- Refactored the build.ps1 script so all parameters are opt-in
- Updated the cmd file that calls it to pass the right parameters
- When running tests, instead of passing in a path to run, the function finds all *tests.csproj files, and starts runs with the specified filter to all projects.
  - This required renaming of a few test projects, and updates to the InternalsVisibleTo attributes in product assemblies.
- Split up test categories, such as IotHubE2E to "IotHub" and "E2E". That way all E2E tests can be run with a single tag.
- Added more test categories for tests to exclude during PR validation, such as LongRunning and Flaky

In order to get TlsVersion tests to run, I had to:
- Change their test trait to match MSTest
- Make the class one that can be instantiated, so it can be tested in parallel
- Update calling classes to use the static Instance property
- Update documentation to indicate the Instance property


vsts/vsts.yaml
```diff
@@ -1,467 +1,457 @@
...
+        displayName: 'GIT checkout'
+        env:
+          COMMIT_FROM: $(COMMIT_FROM)
+
+      - script: |
+          rem Run dotnet first experience.
+          dotnet new
+
+          rem Start build
+          build.cmd -clean -build -configuration Debug
```
drwill-ms: View this file with whitespace diff off, otherwise it is impossible.

This is the only intentional meaningful line change, matching the parameter changes in build.ps1.

build.ps1
```diff
@@ -84,124 +86,121 @@ Function CheckPublishTools()
...
+    {
+        & dotnet clean $projectPath --verbosity $verbosity --configuration $configuration
+        if ($LASTEXITCODE -ne 0)
+        {
             throw "Clean failed: $label"
         }
     }
 
-    & dotnet build --verbosity $verbosity --configuration $configuration
+    & dotnet build $projectPath --verbosity $verbosity --configuration $configuration
```
drwill-ms: By specifying the project path, we don't need to change to the directory (line 103 removed)

build.ps1
```diff
@@ -84,124 +86,121 @@ Function CheckPublishTools()
...
     Write-Host -ForegroundColor Cyan $label
-    cd (Join-Path $rootDir $path)
 
-    if ($framework -eq "*")
+    $testProjects = Get-ChildItem *tests.csproj -Recurse
+
+    $runTestCmd = "dotnet test --verbosity $verbosity --configuration $configuration --logger trx"
+    if ($filterTestCategory -ne "*")
     {
-        if ($filterTestCategory -eq "*")
```
drwill-ms: We can get rid of all these DRY violations by building a string of the command and running it

e2e/test/MessageReceiveE2ETests.cs
```diff
@@ -1,22 +1,22 @@
...
+using Microsoft.Azure.Devices.Client;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Azure.Devices.E2ETests
 {
     [TestClass]
-    [TestCategory("IoTHub-E2E")]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    [TestCategory("LongRunning")]
```
abhipsaMisra: Looks like an oversight; we don't need the TestCategory on both the class and tests in it

e2e/test/MessageSendE2ETests.cs
```diff
@@ -80,7 +81,8 @@ public async Task Message_DeviceSendSingleMessage_AmqpWs_WithHeartbeats()
         }
 
         [TestMethod]
-        [TestCategory("ProxyE2ETests")]
+        [TestCategory("Proxy")]
+        [TestCategory("LongRunning")]
```
abhipsaMisra: This takes long?

drwill-ms: I've been trimming away anything that takes more than 10 seconds

vsts/linux_csharp_netcore.sh
```diff
@@ -2,12 +2,7 @@
...
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-./build.sh -clean -configuration Debug
-if [ "$?" -ne 0 ]; then
-	exit 1
-fi
-
-./build.sh -clean -configuration Release -e2etests
+./build.sh -clean -build -configuration Debug -prtests
 if [ "$?" -ne 0 ]; then
```
abhipsaMisra: why not test it as Release?

vsts/windows_csharp.cmd
```diff
@@ -4,24 +4,15 @@
...
-rem Build Debug + run UTs
-call build.cmd -clean -configuration Debug
-if errorlevel 1 goto :eof
-
+@rem Build Debug + run UTs and select E2E
 taskkill /F /IM Simulator.exe
 start /D .\vsts\TpmSimulator Simulator.exe
-
-rem Build Release + run E2E Tests
-rem If the build is Delay Signed, this requires strong name validation disabled for our key:
```
abhipsaMisra: we should retain these comments; for reference

vsts/windows_csharp.cmd
```diff
@@ -4,24 +4,15 @@
...
-rem (As Administrator:)  sn -Vr *,31bf3856ad364e35
-rem Re-enable strong name validation for this key using:
-rem (As Administrator:)  sn -Vx
-call build.cmd -clean -configuration Release -e2etests
+call build.cmd -clean -build -configuration Debug -prtests
```
abhipsaMisra: we should discuss the impact of modifying this on the release pipeline

abhipsaMisra: shouldn't this have -unittests as well?

build.ps1
```diff
@@ -84,124 +86,121 @@ Function CheckPublishTools()
...
-    $projectName = (dir (Join-Path $projectPath *.csproj))[0].BaseName
+    $projectName = (Get-ChildItem (Join-Path $projectPath *.csproj))[0].BaseName
 
     if ($sign)
     {
-        Write-Host -ForegroundColor Magenta "`tSigning binaries: $projectName"
-        $filesToSign = dir -Recurse .\bin\Release\$projectName.dll
+        Set-Location $projectPath
+        Write-Host -ForegroundColor Magenta "`tSigning binaries for: $projectName"
+        $filesToSign = Get-ChildItem -Path "$projectPath\bin\$configuration\*\$projectName.dll" -Recurse
```
abhipsaMisra: we would not want to sign debug packages; maybe we could add some other validation to check that sign is always called on release only?

build.ps1
```diff
@@ -84,124 +86,121 @@ Function CheckPublishTools()
...
+Function RunTests($message, $framework = "*", $filterTestCategory = "*")
+{
+    $label = "TEST: --- $message $configuration $filterTestCategory ---"
 
     Write-Host
     Write-Host -ForegroundColor Cyan $label
-    cd (Join-Path $rootDir $path)
 
-    if ($framework -eq "*")
+    $testProjects = Get-ChildItem *tests.csproj -Recurse
```
abhipsaMisra: how to differentiate between unit test and E2E test projects?

abhipsaMisra: By TestCategory; I see!

vsts/windows_csharp.cmd
abhipsaMisra: > shouldn't this have -unittests as well?

default is unit test+e2e

abhipsaMisra: The release script calls into build.cmd directly; so modifying this file should have no impact.

vsts/linux_csharp_netcore.sh
drwill-ms: PR tests include unit tests, so they must be debug. Previously I was running unit tests in debug and then building for release and running the subset of E2E. By only doing debug in a PR we save another 90 seconds from building 2 flavors, which seemed like another nice gain.

vsts/windows_csharp.cmd
drwill-ms: added back

build.ps1
drwill-ms: I put in some extra error handling in the initial $sign check and throw if not Release.

PR #1217, opened 2020-02-05
Smart Gate + Gate Refactor + E2E Refactor

timtay-microsoft:
E2E Test changes:
-Test devices are no longer shared between test methods. New devices are created for each test as needed
-Simplified amqp pooling tests to not check for connection status since their is not reason to check them there
-Parallelize e2e tests at class level (except for tests using tpm simulator) for windows and linux (used to only be for Windows)
-Unify Windows and Linux config since they now have the same settings
-Add better exception message to provisioning failed cases so that we can investigate why the provisioning failed when it does

Gate changes:
-Add smart gate feature so only relevant tests are run
-Unifies build scripts into powershell scripts that can be used for both windows and linux
-Move tpm simulator into separate phase for windows rather than keep it in build scripts
-Remove manual checkout from all jobs. VSTS already does that
-Remove Net452 targetting pack install. Did not seem top be used, so no reason to download it
-Re-label test output so that it distinguishes between failed tests from framework to framework
-Parallelize windows build so that each of its 3 frameworks run their tests simultaneously


vsts/determine_tests_to_run.ps1
```diff
@@ -0,0 +1,92 @@
...
+#  variables to signal later vsts scripts to run or not to run certain e2e tests
+
+# For all other builds (nightly builds, for instance), this script will set environment variables to run all e2e tests.
+
+Write-Host "Determining tests to run..."
+$Env:RUN_IOTHUB_TESTS = "false"
+$Env:RUN_PROVISIONING_TESTS = "false"
+
+$targetBranch = ($env:TARGET_BRANCH)
+if (($env:TARGET_BRANCH).toLower().Contains("system.pullrequest.targetbranch"))
```
abhipsaMisra: Question: what is stored in `$env:TARGET_BRANCH`?

vsts/determine_tests_to_run.ps1
```diff
@@ -0,0 +1,92 @@
...
+			$Env:RUN_PROVISIONING_TESTS = "true"
+		}
+
+        # Any changes to e2e test folder should run all tests
+		if ($line.toLower().Contains("e2e"))
+        {
+            $Env:RUN_IOTHUB_TESTS = "true"
+            $Env:RUN_PROVISIONING_TESTS = "true"
+        }
+	}
```
abhipsaMisra: need another block here for "common" folder

build.ps1
```diff
@@ -21,6 +21,10 @@ Parameters:
     -stresstests: Runs Stress tests.
     -publish: (Internal use, requires nuget toolset) Publishes the nuget packages.
     -verbosity: Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+	-framework: Select which framework to run tests on. Allowed values examples include, but are not limited to, "netcoreapp2.1", "net47", "net451"
```
drwill-ms: fix spacing to be the same as the others above

build.ps1
```diff
@@ -272,7 +274,18 @@ try
...
+        if ($skipIotHubTests)
+        {
+            $testCategory += "&TestCategory!=IoTHub"
+        }
+		
+        if ($skipDPSTests)
+        {
+            $testCategory += "&TestCategory!=DPS"
+        }
+
```
drwill-ms: combine double empty lines to 1

e2e/test/PoolingOverAmqp.cs
```diff
@@ -6,10 +6,12 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using static Microsoft.Azure.Devices.E2ETests.FaultInjectionPoolingOverAmqp;
 
 namespace Microsoft.Azure.Devices.E2ETests
 {
-    public static class PoolingOverAmqp
+    [TestClass]
+    public class PoolingOverAmqp
```
drwill-ms: Should these have a TestCategory?

e2e/test/PoolingOverAmqp.cs
```diff
@@ -65,90 +92,123 @@ public static class PoolingOverAmqp
...
-                        {
-                            deviceConnectionStatusAsExpected = false;
-                        }
+                        Assert.AreEqual(2, amqpConnectionStatuses[i].ConnectionStatusChangeCount);
 
                         // The connection status should be "Disabled", with connection status change reason "Client_close"
                         Assert.AreEqual(ConnectionStatus.Disabled, amqpConnectionStatuses[i].LastConnectionStatus, $"The actual connection status is = {amqpConnectionStatuses[i].LastConnectionStatus}");
                         Assert.AreEqual(ConnectionStatusChangeReason.Client_Close, amqpConnectionStatuses[i].LastConnectionStatusChangeReason, $"The actual connection status change reason is = {amqpConnectionStatuses[i].LastConnectionStatusChangeReason}");
+
+                        reRunTest = false;
```
abhipsaMisra: reRunTest is never set as true; it should be set to true in the catch block (when Exception is AssertionFailedException)

e2e/test/config/TestConfig.cs
```diff
@@ -1,14 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System.Diagnostics.Tracing;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-[assembly: Parallelize(Workers = 1, Scope = ExecutionScope.ClassLevel)]
+[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.ClassLevel)]
```
drwill-ms: Can you comment the effect of this attribute? Is it to run tests in a given test class serially?

e2e/test/PoolingOverAmqp.cs
```diff
@@ -65,90 +92,123 @@ public static class PoolingOverAmqp
...
-                    var amqpConnectionStatusChange = new AmqpConnectionStatusChange();
+                    var amqpConnectionStatusChange = new AmqpConnectionStatusChange(devicePrefix);
                     deviceClient.SetConnectionStatusChangesHandler(amqpConnectionStatusChange.ConnectionStatusChangesHandler);
 
                     testDevices.Add(testDevice);
                     deviceClients.Add(deviceClient);
                     amqpConnectionStatuses.Add(amqpConnectionStatusChange);
 
-                    operations.Add(initOperation(deviceClient, testDevice));
+                    operations.Add(deviceClient.OpenAsync());
```
abhipsaMisra: Only opening the client will not cause all links to be opened. It might be a better idea to add this check inside the `CombinedClientOperationsPoolAmqpTests`; or alleast have them perform all operations

e2e/test/config/TestConfig.cs
abhipsaMisra: Why 0 workers?

vsts/determine_tests_to_run.ps1
```diff
@@ -0,0 +1,92 @@
...
+ForEach ($line in $($GitDiff -split "`r`n"))
+{
+	if ($line.EndsWith(".md", "CurrentCultureIgnoreCase") -or $line.EndsWith(".png", "CurrentCultureIgnoreCase"))
+	{
+        # These file types are ignored when determining if source code changes require running e2e tests
+	}
+	elseif ($line.toLower().Contains("sample"))
+	{
+	    # Sample changes don't necessitate running e2e tests
+    }
```
drwill-ms: tabbing is off here

drwill-ms: MIght consider installing a powershell extension for VSCode (ms-vscode.powershell) and it'll do all your formatting for you.

vsts/vsts.yaml
```diff
@@ -1,460 +1,405 @@
...
-        enabled: false
+    displayName: build
 
-      - task: securedevelopmentteam.vss-secure-development-tools.build-task-codemetrics.CodeMetrics@1
-        displayName: 'Run CodeMetrics'
-        inputs:
-          Files: '$(Build.SourcesDirectory)\**\DeviceExplorer.exe;$(Build.SourcesDirectory)\**\Microsoft.Azure.Devices.*.dll'
+- job: DEVEXP
+  displayName: Device Explorer
+  steps:
```
abhipsaMisra: We should retain the timeout property; unless there's a reason not to

vsts/determine_tests_to_run.ps1
```diff
@@ -0,0 +1,92 @@
...
+	else
+	{
+	    # If code changes were made to provisioning package, only need to run provisioning tests
+		if ($line.toLower().Contains("provisioning") -or $line.toLower().Contains("security"))
+		{
+			$Env:RUN_PROVISIONING_TESTS = "true"
```
drwill-ms: Why use environment variables, instead of powershell variables?

vsts/determine_tests_to_run.ps1
timtay-microsoft: It is equal to the azure devops built-in variable $(System.PullRequest.TargetBranch)

Its value is the target branch of the PR build ("master", usually) or it is a default value for builds that are not PR builds. I'll add a note in the script about this

vsts/determine_tests_to_run.ps1
timtay-microsoft: Good catch, I'll add it

build.ps1
timtay-microsoft: sure thing

build.ps1
timtay-microsoft: sure thing

e2e/test/PoolingOverAmqp.cs
timtay-microsoft: Definitely, I'll add iothub and e2e tags here

e2e/test/PoolingOverAmqp.cs
timtay-microsoft: You are absolutely right, I'll add it

e2e/test/config/TestConfig.cs
timtay-microsoft: Sure, I'll add a comment explaining what that means. Workers = 0 makes the test engine use one worker per available core on the computer running it.

timtay-microsoft: That means that the test engine will use as many workers as there are available cores on the computer running it. I'll add an explanation in the comments

vsts/vsts.yaml
timtay-microsoft: I'll add it back in, sure

vsts/determine_tests_to_run.ps1
timtay-microsoft: sure thing

vsts/determine_tests_to_run.ps1
timtay-microsoft: I can just make these global powershell variables, sure

e2e/test/PoolingOverAmqp.cs
timtay-microsoft: Sure, I can refactor this test so that CombinedClientOperationsPoolAmqpTests behaves like it used to (cares about connection status), while the other amqp pool tests don't

PR #1219, opened 2020-02-08
Test code cleanup

drwill-ms:
Some test code cleanup


e2e/test/AzureSecurityCenterForIoTSecurityMessageE2ETests.cs
```diff
@@ -19,10 +19,9 @@ public class AzureSecurityCenterForIoTSecurityMessageE2ETests : IDisposable
     {
         private readonly string _devicePrefix = $"E2E_{nameof(AzureSecurityCenterForIoTSecurityMessageE2ETests)}_";
         private readonly string _modulePrefix = $"E2E_{nameof(AzureSecurityCenterForIoTSecurityMessageE2ETests)}_";
-        private static TestLogging _log = TestLogging.GetInstance();
 
-        private readonly ConsoleEventListener _listener;
```
abhipsaMisra: why remove the readonly? It is initialized only in the constructor.

drwill-ms: Assigned to null during Dispose

PR #1221, opened 2020-02-11
added API to enable/disable certificate revocation check for MQTT

davilu:
## Checklist

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -910,7 +910,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
                 var clientTlsSettings = new ClientTlsSettings(
                      protocols,
-                     certs.Any(),
+                     settings.CertificateRevocationCheck,
                      certs,
                      iotHubConnectionString.HostName);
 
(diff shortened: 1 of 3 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -910,7 +910,7 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
 
                 var clientTlsSettings = new ClientTlsSettings(
                      protocols,
-                     certs.Any(),
+                     settings.CertificateRevocationCheck,
```
drwill-ms: Should be certs.Any() && settings.CertificateRevocationCheck, because you don't want to check if there aren't certs.

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -26,6 +26,10 @@ public class MqttTransportSettings : ITransportSettings
         private const QualityOfService DefaultReceivingQoS = QualityOfService.AtLeastOnce;
         private static readonly TimeSpan DefaultConnectArrivalTimeout = TimeSpan.FromSeconds(10);
         private static readonly TimeSpan DefaultDeviceReceiveAckTimeout = TimeSpan.FromSeconds(300);
+        /// <summary>
+        /// To enable certificate revocation check. Default to be false.
+        /// </summary>
+        public bool CertificateRevocationCheck = false;
```
davilu: Flipped default value to false based on following facts: 
- avoid break change to customer
- based on this [article](https://medium.com/@alexeysamoshkin/how-ssl-certificate-revocation-is-broken-in-practice-af3b63b9cb3)
> There is no ready-to-go 100% working and reliable solution to make browsers detect revoked certificates in a timely manner and refuse connecting

PR #1223, opened 2020-02-11
added API to enable/disable certificate revocation check for AMQP

davilu:
## Checklist

### DIFF
--- a/iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
+++ b/iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
@@ -210,15 +210,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
 
         private bool CausedByRevocationCheckError(X509Chain chain)
         {
-            foreach (X509ChainStatus status in chain.ChainStatus)
-            {
-                if(status.Status != X509ChainStatusFlags.RevocationStatusUnknown)
-                {
-                    return false;
-                }
-            }
-
-            return true;
+            return chain.ChainStatus.All(status => status.Status == X509ChainStatusFlags.RevocationStatusUnknown);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
```diff
@@ -200,7 +200,25 @@ private bool OnRemoteCertificateValidation(object sender, X509Certificate certif
...
 
+            if (!_amqpTransportSettings.CertificateRevocationCheck && sslPolicyErrors == SslPolicyErrors.RemoteCertificateChainErrors && CausedByRevocationCheckError(chain))
+            {
+                return true;
+            }
+
             return false;
         }
+
+        private bool CausedByRevocationCheckError(X509Chain chain)
```
philipktlin: I think we want to ensure all status are not X509ChainStatusFlags.RevocationStatusUnknown in order to return false.  
```suggestion
        private bool CausedByRevocationCheckError(X509Chain chain) => chain.ChainStatus.Any(s => s.Status == X509ChainStatusFlags.RevocationStatusUnknown);
```

philipktlin: I tested this method with my suggested change.

iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
```diff
@@ -200,7 +200,25 @@ private bool OnRemoteCertificateValidation(object sender, X509Certificate certif
                 return true;
             }
 
+            if (!_amqpTransportSettings.CertificateRevocationCheck && sslPolicyErrors == SslPolicyErrors.RemoteCertificateChainErrors && CausedByRevocationCheckError(chain))
```
philipktlin: David, I don't test this condition "_amqpTransportSettings.CertificateRevocationCheck" as I used old sdk to test.

davilu: I think it's OK. If `_amqpTransportSettings.CertificateRevocationCheck` is  `true` it will bypass this block and return `false` which will throw exception as expected.

iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
davilu: We want to make sure there is no other  `X509ChainStatusFlags` but `X509ChainStatusFlags.RevocationStatusUnknown`.

PR #1225, opened 2020-02-11
Make TlsVersions constructor internal

drwill-ms:


shared/src/TlsVersions.cs
```diff
@@ -18,6 +18,10 @@ public class TlsVersions
         /// </summary>
         public static readonly TlsVersions Instance = new TlsVersions();
 
+        internal TlsVersions()
```
abhipsaMisra: nit: Can add a comment here that this is for internal testing

Issue #919, opened 2019-04-23
AMQP drops null app properties while MQTT and HTTP do not. All protocols should be consistent in their implementation.


aklintsevich:
![image](https://user-images.githubusercontent.com/4860052/56604614-3be82180-65b7-11e9-8e27-2fa5044d56c3.png)

aklintsevich:
CustomNullProperty was added to all protocols, but as we can see in the picture it appears in MQTT and HTTP, but not AMQP.

prmathur-microsoft:
Can you give details on 
OS, version, SKU and CPU architecture used: 

Application's .NET Target Framework : 

Device: 

SDK version used:

aklintsevich:
@prmathur-microsoft 

"OS, version, SKU and CPU architecture used:

Application's .NET Target Framework :

Device:"

This is not necessary to know. This occurs on the latest SDK version and can be easily verified on any
OS, version, SKU, CPU architecture.

jenoola:
@aklintsevich please provide the information requested, as well as the actual number for SDK version you are using and a code sample for repro

az-iot-builder-01:
@aklintsevich, @prmathur-microsoft, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1243, opened 2020-02-18
Add null Application Property support

bikamani:
Adding changes to fix github [issue](https://github.com/Azure/azure-iot-sdk-csharp/issues/919)

AMQP drops null app properties while MQTT and HTTP do not. All protocols should be consistent in their implementation.
Fixes: #919


e2e/test/EventHubTestListener.cs
```diff
@@ -133,7 +134,7 @@ private static bool VerifyKeyValue(string checkForKey, string checkForValue, IDi
             {
                 if (checkForKey == key)
                 {
-                    if ((string)properties[checkForKey] == checkForValue)
```
abhipsaMisra: You don't want to return true for all `null` values. You should instead pass `null` into `checkForValue` for application properties

iothub/device/src/Transport/AmqpIoT/AmqpIoTMessageConverter.cs
```diff
@@ -339,7 +339,7 @@ private static bool TryGetNetObjectFromAmqpObject(object amqpObject, MappingType
             netObject = null;
             if (amqpObject == null)
             {
-                return false;
```
abhipsaMisra: Is this `amqpObject` specific to application properties for a message only? Is it being passed from any other method? If yes, then we are introducing new behavior where any `amqpObject` can now be `null`.

bikamani: Nope. I see only one reference for messages

PR #1254, opened 2020-02-27
Add support for JobProperties.StorageAuthenticationType property

drwill-ms:


iothub/service/src/ImportExportJobParameters.cs
```diff
@@ -0,0 +1,63 @@
+﻿using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Input parameters for jobs
+    /// </summary>
+    public class ImportExportJobParameters
```
bikamani: Should it instead be called ImportExportJobInputParameters or ImportExportJobRequestParameters

drwill-ms: I find the extra words "input" and "request" a bit redundant, don't you? Parameters are things one passes as input or to a request.

drwill-ms: I should note that I originally called this JobPropertiesParameters, but after looking at more of our source code we have several different ambiguously named Job* classes for jobs other than import and export. I figured I'd disambiguate here. I don't take naming lightly, and do welcome feedback on the name.

At the same time, we should generally strive for concisely named types that clearly convey their identity. Obviously those two things are a trade-off.

iothub/service/src/ClientApiVersionHelper.cs
```diff
@@ -3,12 +3,20 @@
...
+    /// Holds the API version numbers required in data-plane calls to the service
+    /// </summary>
+    internal class ClientApiVersionHelper
     {
-        const string ApiVersionQueryPrefix = "api-version=";
-        const string ApiVersionGA = "2016-02-03";
-        const string ApiVersionLatest = "2019-03-30";
+        private const string ApiVersionQueryPrefix = "api-version=";
+        private const string ApiVersionGA = "2016-02-03";
+        private const string ApiVersionLatest = "2020-03-13";
```
drwill-ms: Updated to latest api version

iothub/service/src/ExportDevicesRequest.cs
```diff
@@ -1,15 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-namespace Microsoft.Azure.Devices
-{
-    /// <summary>
-    /// Used to provide parameters when requesting an export through the Resource Provider.
-    /// </summary>
-    public sealed class ExportDevicesRequest
```
drwill-ms: Unreferenced class that looks to be related to the export devices job.

Yes, it is a public class, so technically part of the published API. Should I just tag it as obsolete, instead?

As it is not referenced anywhere, likely, no one was using it.

iothub/service/src/HttpRegistryManager.cs
```diff
@@ -959,58 +959,86 @@ public override Task CancelJobAsync(string jobId)
...
+            return ExportDevicesAsync(
+                JobProperties.CreateForExportJob(
+                    exportBlobContainerUri,
+                    excludeKeys,
+                    outputBlobName),
+                ct);
+        }
 
-            return CreateJobAsync(jobProperties, ct);
+        public override Task<JobProperties> ExportDevicesAsync(JobProperties jobParameters, CancellationToken cancellationToken = default)
```
drwill-ms: New API impl for export.

All previous methods now call this one. It enforces the job type.

iothub/service/src/HttpRegistryManager.cs
```diff
@@ -959,58 +959,86 @@ public override Task CancelJobAsync(string jobId)
...
+            return ImportDevicesAsync(
+                JobProperties.CreateForImportJob(
+                    importBlobContainerUri,
+                    outputBlobContainerUri,
+                    inputBlobName),
+                ct);
+        }
 
-            return CreateJobAsync(jobProperties, ct);
+        public override Task<JobProperties> ImportDevicesAsync(JobProperties jobParameters, CancellationToken cancellationToken = default)
```
drwill-ms: New API impl for import.

All previous methods now call this one. It enforces the job type.

iothub/service/src/ExportImportRequest.cs
```diff
@@ -8,11 +8,7 @@ namespace Microsoft.Azure.Devices
     /// </summary>
     public sealed class ExportImportRequest
     {
-        private string storageConnectionString;
-
-        private string containerName;
-
-        public string StorageConnectionString { get => storageConnectionString; set => storageConnectionString = value; }
-        public string ContainerName { get => containerName; set => containerName = value; }
+        public string StorageConnectionString { get; set; }
```
drwill-ms: Unrelated to this change - just some code cleanup I did while looking at files that might be related.

iothub/service/src/ImportDevicesRequest.cs
```diff
@@ -1,15 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-namespace Microsoft.Azure.Devices
-{
-    /// <summary>
-    /// Used to provide parameters when requesting an import through the Resource Provider.
-    /// </summary>
-    public sealed class ImportDevicesRequest
```
drwill-ms: Unreferenced class that looks to be related to the import devices job.

Yes, it is a public class, so technically part of the published API. Should I just tag it as obsolete, instead?

As it is not referenced anywhere, likely, no one was using it.

iothub/service/src/JobProperties.cs
```diff
@@ -52,49 +53,121 @@ public JobProperties()
...
+        /// <param name="inputBlobContainerUri">URI containing SAS token to a blob container that contains registry data to sync</param>
+        /// <param name="outputBlobContainerUri">URI containing SAS token to a blob container.  This is used to output the status of the job and the results.</param>
+        /// <param name="inputBlobName">The blob name to be used when importing from the provided input blob container</param>
+        /// <param name="storageAuthenticationType">Specifies authentication type being used for connecting to storage account</param>
+        /// <returns>An instance of JobProperties</returns>
+        public static JobProperties CreateForImportJob(
+            string inputBlobContainerUri,
+            string outputBlobContainerUri,
+            string inputBlobName = null,
+            StorageAuthenticationType storageAuthenticationType = StorageAuthenticationType.KeyBased)
```
drwill-ms: Factory method to help users know which parameters are required and which are optional for import job. Method below for export job.

iothub/service/src/RegistryManager.cs
```diff
@@ -620,10 +623,18 @@ public void Dispose()
...
         /// <returns>JobProperties of the newly created job.</returns>
         public abstract Task<JobProperties> ExportDevicesAsync(string exportBlobContainerUri, string outputBlobName, bool excludeKeys, CancellationToken cancellationToken);
 
+        /// <summary>
+        /// Creates a new bulk job to export device registrations to the container specified by the provided URI.
+        /// </summary>
+        /// <param name="jobParameters">Parameters for the job</param>
+        /// <param name="cancellationToken">Task cancellation token</param>
+        /// <returns>JobProperties of the newly created job.</returns>
+        public abstract Task<JobProperties> ExportDevicesAsync(JobProperties jobParameters, CancellationToken cancellationToken = default);
```
drwill-ms: New API for customers for export job.

iothub/service/src/RegistryManager.cs
```diff
@@ -656,10 +667,20 @@ public void Dispose()
...
 
+#pragma warning restore CA1054 // Uri parameters should not be strings
+
+        /// <summary>
+        /// Creates a new bulk job to import device registrations into the IoT Hub.
+        /// </summary>
+        /// <param name="jobParameters">Parameters for the job</param>
+        /// <param name="cancellationToken">Task cancellation token</param>
+        /// <returns>JobProperties of the newly created job.</returns>
+        public abstract Task<JobProperties> ImportDevicesAsync(JobProperties jobParameters, CancellationToken cancellationToken = default);
```
drwill-ms: New API for customers for import job.

shared/src/StorageAuthenticationType.cs
```diff
@@ -0,0 +1,26 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Specifies authentication type being used for connecting to storage account.
+    /// </summary>
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum StorageAuthenticationType
```
drwill-ms: New enum def

iothub/service/src/JobProperties.cs
```diff
@@ -52,49 +53,121 @@ public JobProperties()
...
 
         /// <summary>
-        /// System genereated.  Ignored at creation.
-        /// If status == failure, this represents a string containing the reason.
+        /// Specifies authentication type being used for connecting to storage account
         /// </summary>
-        [JsonProperty(PropertyName = "failureReason", NullValueHandling = NullValueHandling.Ignore)]
-        public string FailureReason { get; set; }
+        [JsonProperty(PropertyName = "storageAuthenticationType", NullValueHandling = NullValueHandling.Ignore)]
+        public StorageAuthenticationType StorageAuthenticationType { get; set; }
```
drwill-ms: New enum inclusion in job properties

iothub/service/src/HttpRegistryManager.cs
bikamani: On out side, do we not need to validate number of parameters required for ImportDevices - InputBlobContainerUri and OutputBlobContainerUri at least?

iothub/service/src/RegistryManager.cs
bikamani: Should there be tests for these APIs?

iothub/service/src/RegistryManager.cs
```diff
@@ -656,10 +667,20 @@ public void Dispose()
         /// <param name="importBlobContainerUri"></param>
```
tonygleck: Shouldn't this be InputBlobContainer?

iothub/service/src/JobProperties.cs
```diff
@@ -52,49 +53,121 @@ public JobProperties()
...
+
+#pragma warning disable CA1054 // Uri parameters should not be strings
+
+        /// <summary>
+        /// Creates an instance of JobProperties with paramters ready to start an Import job
```
bikamani: typo parameters

iothub/service/src/RegistryManager.cs
drwill-ms: Yes, but I can't run/test them until this new version is available in production.

iothub/service/src/RegistryManager.cs
drwill-ms: It probably should have been, but I cannot rename it now in case some customer used named parameters.

iothub/service/src/JobProperties.cs
drwill-ms: fixed

iothub/service/src/HttpRegistryManager.cs
drwill-ms: We can let the service validate and refuse.

iothub/service/src/ClientApiVersionHelper.cs
```diff
@@ -1,14 +1,34 @@
...
+        private const string ApiVersionDefault = "2019-03-30";
+        private const string ApiVersionLimitedAvailability = "2020-03-13";
+
+        // For import/export devices jobs, a new parameter is available in a
+        // new api-version, which is only available in a few initial regions.
+        // Control access via an environment variable. If a user wishes to try it out,
+        // they can set "EnabledStorageIdentity" to "1". Otherwise, the SDK will still
+        // default to the latest, broadly-supported api-version used in this SDK.
+        internal static bool IsStorageIdentityEnabled => StringComparer.OrdinalIgnoreCase.Equals("1", Environment.GetEnvironmentVariable("EnableStorageIdentity"));
+
```
prmathur-microsoft: 👍

iothub/service/src/ClientApiVersionHelper.cs
```diff
@@ -1,14 +1,34 @@
...
+    /// </summary>
+    internal class ClientApiVersionHelper
     {
-        const string ApiVersionQueryPrefix = "api-version=";
-        const string ApiVersionGA = "2016-02-03";
-        const string ApiVersionLatest = "2019-03-30";
-        public const string ApiVersionQueryString = ApiVersionQueryPrefix + ApiVersionLatest;
+        private const string ApiVersionQueryPrefix = "api-version=";
+        private const string ApiVersionGA = "2016-02-03";
+        private const string ApiVersionDefault = "2019-03-30";
```
prmathur-microsoft: Do we still need two?

iothub/service/src/ExportDevicesRequest.cs
prmathur-microsoft: If it isn't referenced anywhere I think it is safe to delete.

iothub/service/src/ExportImportRequest.cs
```diff
@@ -8,11 +8,7 @@ namespace Microsoft.Azure.Devices
...
+        public string ContainerName { get; set; }
```
prmathur-microsoft: Does this need to be public? I thought we are using RBAC for storage? is this for non identity path?

iothub/service/src/HttpRegistryManager.cs
```diff
@@ -1357,7 +1399,7 @@ public override Task<BulkRegistryOperationResult> UpdateTwins2Async(IEnumerable<
         {
             return BulkDeviceOperationsAsync<BulkRegistryOperationResult>(
                 GenerateExportImportDeviceListForTwinBulkOperations(twins, forceUpdate ? ImportMode.UpdateTwin : ImportMode.UpdateTwinIfMatchETag),
-                ClientApiVersionHelper.ApiVersionQueryString,
+                ClientApiVersionHelper.ApiVersionQueryStringDefault,
                 cancellationToken);
```
prmathur-microsoft: Do we always return default here?

prmathur-microsoft: even if the env is set to 1?

drwill-ms: For this method, for sure. We don't want to use the new api-version unless it is the import/export devices job

iothub/service/src/ClientApiVersionHelper.cs
drwill-ms: Need two what?

I chose not to do the work to make sure we could get rid of the GA version for now, but I put it on a personal task list to look at later. It seems we just never did the work to move 3 calls from this older api-version.

iothub/service/src/ExportImportRequest.cs
drwill-ms: I didn't change the visibility. We shouldn't change it now that it has been public.

This class is unrelated to the import/export job, but I looked at it because of the name. Then I saw an opportunity to use auto-properties, and cleaned it up.

iothub/service/src/ClientApiVersionHelper.cs
prmathur-microsoft: okay

iothub/service/src/ClientApiVersionHelper.cs
```diff
@@ -1,14 +1,34 @@
...
+        // they can set "EnabledStorageIdentity" to "1". Otherwise, the SDK will still
+        // default to the latest, broadly-supported api-version used in this SDK.
+        internal static bool IsStorageIdentityEnabled => StringComparer.OrdinalIgnoreCase.Equals("1", Environment.GetEnvironmentVariable("EnableStorageIdentity"));
+
+        /// <summary>
+        /// The default API version to use for all data-plane service calls
+        /// </summary>
+        public const string ApiVersionQueryStringDefault = ApiVersionQueryPrefix + ApiVersionDefault;
+
+        public const string ApiVersionQueryStringLimitedAvailability = ApiVersionQueryPrefix + ApiVersionLimitedAvailability;
```
prmathur-microsoft: I wish these were privates. However as they were already public wouldn't effect us much.

drwill-ms: they are in an internal class, so they are effectively internal. yay!

PR #1260, opened 2020-03-06
Remove NETMF ifdefs

drwill-ms:


CodeMaid.config
```diff
@@ -32,6 +32,9 @@
       <setting name="Cleaning_PerformPartialCleanupOnExternal" serializeAs="String">
         <value>1</value>
       </setting>
+      <setting name="Progressing_ShowBuildProgressOnBuildStart" serializeAs="String">
```
drwill-ms: I finally found where this annoying progress tool window was coming from. Buh bye.

PR #1266, opened 2020-03-11
Fix more warnings, and possible resource leaks

drwill-ms:
Updated to latest FxCop nuget. That fixed some false warnings, but introduces new standards.

Two such standards that we consistently break is not having hard-coded strings as error messages, and catching System.Exception. Rather than fix all those (ha) or suppress each one individually, I chose to add globalsuppression files to each project.

Most warning fixes are:

- using nameof for parameter validation exception params
- parameter null checking
- calling dispose on IDisposable objects just before they would be out of scope (oops! - pay special attention to these changes please)
- adding some missed parameter xml tags
- methods marked as async that didn't make await anything
- some exceptions had a public ctor with serialization info param, which apparently shouldn't be public. As we can't change the public API, these have been suppressed.
- compare strings using StringComparer.Ordinal/Invariant.Equals
- using string.IsNullOrWhiteSpace or equiv
- do not use Count() method if Length or Count properties are available

Also included are style changes including:

- usings outside of namespace
- using auto properties where possible
- auto-property getter and setter all on one line
- simply property getter or setter use lambda notation for fewer lines
- class member names, visibility, and readonly when appropriate
- removing "this." now that ambiguity is gone with field names
- formatting method calls and parameters for easier reading, when they get long

### DIFF
--- a/e2e/test/FaultInjectionPoolAmqpTests.MessageSendFaultInjectionPoolAmqpTests.cs
+++ b/e2e/test/FaultInjectionPoolAmqpTests.MessageSendFaultInjectionPoolAmqpTests.cs
@@ -821,27 +821,31 @@ namespace Microsoft.Azure.Devices.E2ETests
                 Assert.IsTrue(isReceived, $"Message is not received for device {testDevice.Id}.");
             };
 
-            Func<IList<DeviceClient>, Task> cleanupOperation = async (deviceClients) =>
+            Func<IList<DeviceClient>, Task> cleanupOperation = (deviceClients) =>
             {
                 foreach (DeviceClient deviceClient in deviceClients)
                 {
                     deviceClient.Dispose();
                 }
+
+                return Task.FromResult(0);
             };
 
-            await FaultInjectionPoolingOverAmqp.TestFaultInjectionPoolAmqpAsync(
-                MessageSend_DevicePrefix,
-                transport,
-                poolSize,
-                devicesCount,
-                faultType,
-                reason,
-                delayInSec,
-                durationInSec,
-                (d, t) => { return Task.FromResult<bool>(false); },
-                testOperation,
-                cleanupOperation,
-                authScope).ConfigureAwait(false);
+            await FaultInjectionPoolingOverAmqp
+                .TestFaultInjectionPoolAmqpAsync(
+                    MessageSend_DevicePrefix,
+                    transport,
+                    poolSize,
+                    devicesCount,
+                    faultType,
+                    reason,
+                    delayInSec,
+                    durationInSec,
+                    (d, t) => { return Task.FromResult(false); },
+                    testOperation,
+                    cleanupOperation,
+                    authScope)
+                .ConfigureAwait(false);
         }
     }
 }
--- a/iothub/device/src/Common/Exceptions/IotHubException.cs
+++ b/iothub/device/src/Common/Exceptions/IotHubException.cs
@@ -18,41 +18,15 @@ namespace Microsoft.Azure.Devices.Client.Exceptions
         [NonSerialized]
         private const string TrackingIdValueSerializationStoreName = "IotHubException-TrackingId";
 
-        [NonSerialized]
-        private bool _isTransient;
-
-        [NonSerialized]
-        private string _trackingId;
-
         /// <summary>
         /// Gets a value indicating if the error is transient.
         /// </summary>
-        public bool IsTransient
-        {
-            get
-            {
-                return _isTransient;
-            }
-            private set
-            {
-                _isTransient = value;
-            }
-        }
+        public bool IsTransient { get; private set; }
 
         /// <summary>
         /// Gets or sets the Azure IoT service-side Tracking ID in Support Requests.
         /// </summary>
-        public string TrackingId
-        {
-            get
-            {
-                return _trackingId;
-            }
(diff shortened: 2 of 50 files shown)


provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -191,66 +196,88 @@ private static async Task CreateLinksAsync(AmqpClientConnection connection, stri
             string correlationId,
             DeviceRegistration deviceRegistration)
         {
-            AmqpMessage amqpMessage;
-            if (deviceRegistration == null)
+            AmqpMessage amqpMessage = null;
+
+            try
             {
-                amqpMessage = AmqpMessage.Create(new MemoryStream(), true);
```
drwill-ms: lots of resource leaks in this file :(

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -293,12 +299,13 @@ private async Task PublishRegisterAsync(IChannelHandlerContext context)
             if (_message.Payload != null && _message.Payload.Length > 0)
             {
                 var deviceRegistration = new DeviceRegistration { Payload = new JRaw(_message.Payload) };
-                var customContentStream =
```
drwill-ms: Another resource leak

e2e/test/FaultInjectionPoolAmqpTests.MessageSendFaultInjectionPoolAmqpTests.cs
```diff
@@ -821,27 +821,31 @@ public async Task Message_IoTHubSak_AuthenticationNoRecovery_MultipleConnections
...
-                devicesCount,
-                faultType,
-                reason,
-                delayInSec,
-                durationInSec,
-                (d, t) => { return Task.FromResult<bool>(false); },
-                testOperation,
-                cleanupOperation,
-                authScope).ConfigureAwait(false);
+            await FaultInjectionPoolingOverAmqp
```
vinagesh: minor: I don't know if this is public facing or not but async function name does not seem to be ending with async

iothub/device/src/Common/Exceptions/IotHubException.cs
```diff
@@ -18,41 +18,15 @@ public class IotHubException : Exception
...
         /// <summary>
         /// Gets a value indicating if the error is transient.
         /// </summary>
-        public bool IsTransient
-        {
-            get
-            {
-                return _isTransient;
-            }
-            private set
```
barustum: changing from private set to set. Intentional?

iothub/device/src/Common/Security/SharedAccessSignature.cs
```diff
@@ -140,29 +103,29 @@ public static bool IsSharedAccessSignature(string rawSignature)
...
+            return ExpiresOn + SharedAccessSignatureConstants.MaxClockSkew < DateTime.UtcNow;
         }
 
         public void Authenticate(SharedAccessSignatureAuthorizationRule sasAuthorizationRule)
         {
-            if (this.IsExpired())
+            if (IsExpired())
             {
-                throw new UnauthorizedAccessException($"The specified SAS token is expired on {this.ExpiresOn}.");
+                throw new UnauthorizedAccessException($"The specified SAS token is expired on { ExpiresOn}.");
```
barustum: space after ExpiresOn

iothub/device/src/Common/Exceptions/IotHubException.cs
drwill-ms: no, good catch

iothub/device/src/Common/Security/SharedAccessSignature.cs
```diff
@@ -12,105 +12,69 @@ namespace Microsoft.Azure.Devices.Client
...
         }
 
-        public string IotHubName
-        {
-            get
-            {
-                return this.iotHubName;
-            }
-        }
+        public string IotHubName { get; private set; }
```
drwill-ms: readonly

provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -191,66 +196,88 @@ private static async Task CreateLinksAsync(AmqpClientConnection connection, stri
...
 
         private static DeviceRegistrationResult ConvertToProvisioningRegistrationResult(
             Models.DeviceRegistrationResult result)
         {
-            var status = ProvisioningRegistrationStatusType.Failed;
-            Enum.TryParse(result.Status, true, out status);
-
-            var substatus = ProvisioningRegistrationSubstatusType.InitialAssignment;
-            Enum.TryParse(result.Substatus, true, out substatus);
+            Enum.TryParse(result.Status, true, out ProvisioningRegistrationStatusType status);
```
vinagesh: Default is Unassigned so will this change the behavior now?

drwill-ms: The .net contract for `out` parameters is that they must be assigned to. So our previous assignment of Failed didn't matter.

Does that answer your question?

PR #1269, opened 2020-03-12
Add support for storage identity auth

drwill-ms:


e2e/test/FileUploadE2ETests.cs
```diff
@@ -93,7 +93,7 @@ private static async Task<string> GetTestFileNameAsync(int fileSize)
 
             string filePath = Path.Combine(Path.GetTempPath(), Path.GetTempFileName());
 
-#if NET451 || NET47
+#if NET451 || NET472
```
abhipsaMisra: Do we need the check for net472? net472 should have ns2.0 API support, so I would expect these APIs to be available

drwill-ms: It had a compile error otherwise, so your expectation was invalidated.

PR #1271, opened 2020-03-13
Change dotnet core version for test projects to 3.1

barustum:
Updated test projects to use .netcoreapp 3.1. 

Reason for the upgrade is that .NET Core version 2.1 has a vulnerability where an unauthenticated attacker can cause a denial of service remotely, by issuing specially crafted requests to the .NET core app.


e2e/test/E2ETests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net472;net451</TargetFrameworks>
```
abhipsaMisra: Question - .net core 2.1 is on [LTS support](https://dotnet.microsoft.com/download/dotnet-core), did it not receive a fix for this issue? Also, it looks like .net core 2.2 has already reached end-of-life. Would .net core 3.1 be a better option here?

barustum: Per the work item filed against us, we can upgrade to 2.2.1, 2.1.7 or higher. Do you recommend doing 2.1.7 instead?

abhipsaMisra: I think .net core v2.1.16 seems like a better option, just because it is on LTS, and so should receive any security updates. .net core 2.2 is no longer supported.

abhipsaMisra: Also, does it work with .net core 3.1? It should still be able to target ns2.0 dll.

barustum: Yes I just checked the compatibility chart. 3.1 should work. I'll make the change to target 3.1.

PR #1274, opened 2020-03-17
Fix (Im|Ex)port tests with older storage nuget

drwill-ms:
Due to a bug between System.Diagnostics.DiagnosticSource and Azure.Storage.Blob failing async container creation, these tests required console logging to be off, but when run with other tests it would still be on in the process. I've fallen back from Azure.Storage.Blob to Microsoft.Azure.Storage.Blob nuget, where the issue does not repro. It looks like it is an issue in the track2 pipeline.

### DIFF
--- a/e2e/test/Helpers/StorageContainer.cs
+++ b/e2e/test/Helpers/StorageContainer.cs
@@ -4,9 +4,8 @@ using System;
 using System.Globalization;
 using System.Text;
 using System.Threading.Tasks;
-using Azure.Storage;
-using Azure.Storage.Blobs;
-using Azure.Storage.Sas;
+using Microsoft.Azure.Storage;
+using Microsoft.Azure.Storage.Blob;
 
 namespace Microsoft.Azure.Devices.E2ETests.Helpers
 {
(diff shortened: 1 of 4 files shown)


e2e/test/Helpers/StorageContainer.cs
```diff
@@ -129,21 +125,23 @@ protected virtual void Dispose(bool disposing)
...
             }
 
             _disposed = true;
         }
 
         private async Task InitializeAsync()
         {
-            BlobContainerClient = new BlobContainerClient(Configuration.Storage.ConnectionString, ContainerName);
-            await BlobContainerClient.CreateIfNotExistsAsync().ConfigureAwait(false);
+            _ = CloudStorageAccount.TryParse(Configuration.Storage.ConnectionString, out CloudStorageAccount storageAccount);
```
barustum: not familiar with the '_' syntax. What does this do?

drwill-ms: It means discard the output from this method. I don't care about it. It signals to any code reader that the return parameter was not ignored by accident.

drwill-ms: https://docs.microsoft.com/en-us/dotnet/csharp/discards

drwill-ms: I checked and there is a Parse method instead, so I'll change to use that.

PR #1279, opened 2020-03-19
Allow to enable/disable CheckCertificateRevocationList for Http 

bikamani:
By default CheckCertificateRevocationList value is set to true.

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -34,6 +34,7 @@ namespace Microsoft.Azure.Devices
         private readonly TimeSpan _defaultOperationTimeout;
         private readonly IWebProxy _customHttpProxy;
         private readonly Action<HttpClient> _preRequestActionForAllRequests;
+        private readonly bool _certificateRevocationCheck;
 
         public HttpClientHelper(
             Uri baseAddress,
@@ -41,7 +42,8 @@ namespace Microsoft.Azure.Devices
             IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping,
             TimeSpan timeout,
             Action<HttpClient> preRequestActionForAllRequests,
-            IWebProxy customHttpProxy)
+            IWebProxy customHttpProxy,
+            bool certificateRevocationCheck)
         {
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
@@ -49,6 +51,7 @@ namespace Microsoft.Azure.Devices
             _defaultOperationTimeout = timeout;
             _preRequestActionForAllRequests = preRequestActionForAllRequests;
             _customHttpProxy = customHttpProxy;
+            _certificateRevocationCheck = certificateRevocationCheck;
             TlsVersions.Instance.SetLegacyAcceptableVersions();
         }
 
--- a/iothub/service/src/HttpTransportSettings.cs
+++ b/iothub/service/src/HttpTransportSettings.cs
@@ -17,5 +17,10 @@ namespace Microsoft.Azure.Devices
         }
 
         public IWebProxy Proxy { get; set; }
+
+        /// <summary>
+        /// To enable certificate revocation check. Default to be false.
+        /// </summary>
+        public bool CertificateRevocationCheck { get; set; } = true;
     }
 }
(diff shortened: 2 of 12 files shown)


iothub/service/src/HttpTransportSettings.cs
```diff
@@ -17,5 +17,10 @@ public HttpTransportSettings()
         }
 
         public IWebProxy Proxy { get; set; }
+
+        /// <summary>
+        /// To enable certificate revocation check. Default to be false.
```
drwill-ms: default to true

iothub/service/src/ServiceClientTransportSettings.cs
```diff
@@ -20,5 +20,7 @@ public ServiceClientTransportSettings()
         public IWebProxy AmqpProxy { get; set; }
 
         public IWebProxy HttpProxy { get; set; }
+
+        public bool CertificateRevocationCheck { get; set; } = true;
```
drwill-ms: add meaningful xml comments

provisioning/service/src/HttpTransportSettings.cs
```diff
@@ -23,5 +23,11 @@ public HttpTransportSettings()
         /// Gets or sets proxy information for the request.
         /// </summary>
         public IWebProxy Proxy { get; set; }
+
+        /// <summary>
+        /// To enable certificate revocation check. Default to be false.
```
drwill-ms: default to true

drwill-ms:
Why not put it centrally on TlsVersions.Instance? That way it is available for all SDKs and all transports across the board.

provisioning/service/src/HttpTransportSettings.cs
```diff
@@ -23,5 +23,11 @@ public HttpTransportSettings()
...
+        /// </summary>
+        public bool CertificateRevocationCheck { get; set; } = true;
```
vinagesh: Wondering if this should be something you can send to the constructor and can be optional (set to true by default). When someone creates an instance, they will easily be able to tell what options are possible rather than going and check what can be set in the class. As a user, I would do new HttpTransportSettings() and it would be much easier for me if the intellisense shows that i can send this parameter.

drwill-ms: The question is, which of the many, many properties should be in the ctor? It gets too long and then the ctor is hard to use.

That is why, with option classes like this, the guidance is to rely on public properties that someone sets, ideally via initializer list.

vinagesh: There were only two here but the explanation makes sense.

common/src/service/HttpClientHelper.cs
```diff
@@ -34,21 +34,24 @@ internal sealed class HttpClientHelper : IHttpClientHelper
...
+            IWebProxy customHttpProxy,
+            bool certificateRevocationCheck)
         {
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
             _defaultOperationTimeout = timeout;
             _preRequestActionForAllRequests = preRequestActionForAllRequests;
             _customHttpProxy = customHttpProxy;
+            _certificateRevocationCheck = certificateRevocationCheck;
```
barustum: I might be missing something here, but where is _certificateRevocationCheck  every used in this file?

bikamani:
> Why not put it centrally on TlsVersions.Instance? That way it is available for all SDKs and all transports across the board.

That's a good idea actually. I do not have any issues going that way. Does it seem to be still the right name for this class with adding this change?

common/src/service/HttpClientHelper.cs
bikamani: we are basically setting property this for httpClientHelper constructor which is referenced in other places such as HttpRegistryManager.cs

drwill-ms:
> > Why not put it centrally on TlsVersions.Instance? That way it is available for all SDKs and all transports across the board.
> 
> That's a good idea actually. I do not have any issues going that way. Does it seem to be still the right name for this class with adding this change?

Yeah, the name isn't great, but the pattern is what we want to adopt. I think it would be fine to put it in that class.

bikamani:
Closing as opened a PR as per new design.

PR #1280, opened 2020-03-19
Test service client with net472 and net451

drwill-ms:


iothub/service/tests/ClientWebSocketTransportTests.cs
```diff
@@ -1,4 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
barustum: duplicates

iothub/service/tests/ClientWebSocketTransportTests.cs
```diff
@@ -79,6 +79,7 @@ public async Task ClientWebSocketTransportReadWithoutConnectTest()
 
         // The following tests can only be run in Administrator mode
         [TestMethod]
+        [Timeout(10000)] // if it is going to fail, fail fast. Otherwise it can go on for 4+ minutes. :(
```
vinagesh: Just curious - why does this fail?

drwill-ms: The comment on 80 says it has to be run in admin mode.

In practice, I find it fails if run with other tests and not in admin mode. If you run it by itself, even not in admin mode, it'll succeed.

abhipsaMisra: nit: save this timeout value as a constant.

drwill-ms: Why?

abhipsaMisra: So that we don't end up having to maintain different values everywhere

drwill-ms: General rule of thumb is to make something a constant if it is referenced from more than one location. In this case, we aren't. The value only matters here. If I found a second method that needed the same value, then I'd consider it.

drwill-ms: I'm going to resolve this as by design, for now. Hit me up on teams if you want to talk about it. :)

PR #1281, opened 2020-03-19
Removing incorrect comments and minor improvements

vinagesh:
## Description of the changes
- Removed unnecessary serialization of class properties  Did not change the class that has already been made public to keep the contract intact.
- Changed the class access modifiers explicitly to internal as that is sufficient
- Renamed the private variables in the class according to coding standards

This change was made to mainly address the TODO in one of the files that said - Encrypt the Json property. This class is internal and does not seem like it needs any seiralization. I've removed the Json attribute and removed the unnecessary TODO from the files. The code is building and E2E tests passed. If I'm missing something, please let me know and i'll make changes accordingly.

https://msazure.visualstudio.com/One/_workitems/edit/5867874

### DIFF
--- a/iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
+++ b/iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
@@ -1,22 +1,22 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using System;
+using Newtonsoft.Json;
 
 namespace Microsoft.Azure.Devices.Common.Data
 {
-    internal sealed class SharedAccessSignatureAuthorizationRule : IEquatable<SharedAccessSignatureAuthorizationRule>
+    public sealed class SharedAccessSignatureAuthorizationRule : IEquatable<SharedAccessSignatureAuthorizationRule>
     {
         private string _primaryKey;
         private string _secondaryKey;
 
+        [JsonProperty(PropertyName = "keyName")]
         public string KeyName { get; set; }
 
+        [JsonProperty(PropertyName = "primaryKey")]
         public string PrimaryKey
         {
-            get
-            {
-                return _primaryKey;
-            }
+            get => _primaryKey;
 
             set
             {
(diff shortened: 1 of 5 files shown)


vinagesh:
azp/run

iothub/service/src/Common/Security/ISharedAccessSignatureCredential.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.Azure.Devices.Common.Security
 
     using Microsoft.Azure.Devices.Common.Data;
       
-    public interface ISharedAccessSignatureCredential
+    internal interface ISharedAccessSignatureCredential
```
abhipsaMisra: Is this change intentional?

vinagesh: yes. If we want to make SharedAccessSignatureAuthorizationRule internal, we have to make this internal. And it seems like that visibility is sufficient.

drwill-ms: Well, we shouldn't be making any types that were public not public any more, unless we have reason to believe the types were of no use.

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -55,12 +48,12 @@ public bool Equals(SharedAccessSignatureAuthorizationRule other)
...
-                string.Equals(this.SecondaryKey, other.SecondaryKey, StringComparison.Ordinal) &&
-                Equals(this.Rights, other.Rights);
+            bool equals = string.Equals(KeyName, other.KeyName, StringComparison.OrdinalIgnoreCase) &&
+                string.Equals(PrimaryKey, other.PrimaryKey, StringComparison.Ordinal) &&
+                string.Equals(SecondaryKey, other.SecondaryKey, StringComparison.Ordinal) &&
+                Equals(Rights, other.Rights);
 
             return equals;
-        }
+        }
```
drwill-ms: somehow, you introduced extra space after this brace

PR #1284, opened 2020-03-24
SDK comments clearer about when user should dispose

drwill-ms:


iothub/device/src/DeviceClient.cs
```diff
@@ -35,143 +33,143 @@ private DeviceClient(InternalClient internalClient)
         }
 
         /// <summary>
-        /// Create an Amqp DeviceClient from individual parameters
+        /// Creates a disposable, Amqp DeviceClient from the specified parameters
```
abhipsaMisra: Does this have to AMQP? This is legacy documentation perhaps

abhipsaMisra: Actually no, disregard this. AMQP is the default protocol selected I think

drwill-ms: It defaults to Amqp, yes

PR #1285, opened 2020-03-24
Allow to enable/disable CheckCertificateRevocationList for Http, Amqp, Mqtt

bikamani:
TlsVersions has a new get set property which is allowing to enable/disable CheckCertificateRevocationList for all transports.

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -34,6 +34,7 @@ namespace Microsoft.Azure.Devices
         private readonly TimeSpan _defaultOperationTimeout;
         private readonly IWebProxy _customHttpProxy;
         private readonly Action<HttpClient> _preRequestActionForAllRequests;
+        private readonly bool _certificateRevocationCheck;
 
         public HttpClientHelper(
             Uri baseAddress,
@@ -50,6 +51,7 @@ namespace Microsoft.Azure.Devices
             _preRequestActionForAllRequests = preRequestActionForAllRequests;
             _customHttpProxy = customHttpProxy;
             TlsVersions.Instance.SetLegacyAcceptableVersions();
+            _certificateRevocationCheck = TlsVersions.Instance.CertificateRevocationCheck;
         }
 
         public Task<T> GetAsync<T>(
@@ -851,8 +853,9 @@ namespace Microsoft.Azure.Devices
             {
 #if !NET451
                 SslProtocols = TlsVersions.Instance.Preferred,
+                CheckCertificateRevocationList = _certificateRevocationCheck,
 #endif
-            };
+        };
 
             if (_customHttpProxy != DefaultWebProxySettings.Instance)
             {
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -32,23 +32,60 @@ namespace Microsoft.Azure.Devices
         private readonly IAuthorizationHeaderProvider _authenticationHeaderProvider;
         private readonly IReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> _defaultErrorMapping;
         private readonly TimeSpan _defaultOperationTimeout;
-        private readonly IWebProxy _customHttpProxy;
-        private readonly Action<HttpClient> _preRequestActionForAllRequests;
+
+        // IDisposables
+
+        private readonly HttpClient _httpClientWithDefaultTimeout;
+
+        private readonly HttpClient _httpClientWithNoTimeout;
+        private readonly HttpClientHandler _httpClientHandler;
 
         public HttpClientHelper(
             Uri baseAddress,
             IAuthorizationHeaderProvider authenticationHeaderProvider,
             IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping,
             TimeSpan timeout,
-            Action<HttpClient> preRequestActionForAllRequests,
             IWebProxy customHttpProxy)
         {
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
             _defaultOperationTimeout = timeout;
-            _preRequestActionForAllRequests = preRequestActionForAllRequests;
-            _customHttpProxy = customHttpProxy;
+
+            // HttpClientHandler is IDisposable, so save onto it for disposing.
+            _httpClientHandler = new HttpClientHandler
+            {
+#if !NET451
+                SslProtocols = TlsVersions.Instance.Preferred,
+                CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck,
+#endif
+            };
+
+            if (customHttpProxy != DefaultWebProxySettings.Instance)
+            {
+                _httpClientHandler.UseProxy = customHttpProxy != null;
+                _httpClientHandler.Proxy = customHttpProxy;
+            }
+
+            // We need two types of HttpClients, one with our default operation timeout, and one without. The one without will rely on
+            // a cancellation token.
+
+            _httpClientWithDefaultTimeout = new HttpClient(_httpClientHandler)
+            {
+                BaseAddress = _baseAddress,
+                Timeout = _defaultOperationTimeout,
+            };
+            _httpClientWithDefaultTimeout.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(CommonConstants.MediaTypeForDeviceManagementApis));
+            _httpClientWithDefaultTimeout.DefaultRequestHeaders.ExpectContinue = false;
+
+            _httpClientWithNoTimeout = new HttpClient(_httpClientHandler)
+            {
+                BaseAddress = _baseAddress,
+                Timeout = Timeout.InfiniteTimeSpan,
+            };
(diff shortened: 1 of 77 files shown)


e2e/test/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -140,20 +149,22 @@ private async Task SendMessageTest(ITransportSettings transportSetting)
             }
         }
 
+        private ITransportSettings CreateHttpTransportSettingWithCertificateRevocationCheck()
+        {
+            TlsVersions.Instance.CertificateRevocationCheck = true;
```
drwill-ms: This can be a problem for tests that run in parallel, because changing this static instance will impact all the others. You should make sure tests that change this value do not run in parallel.

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -17,10 +17,6 @@ public sealed class AmqpTransportSettings : ITransportSettings
         private readonly TransportType _transportType;
         private TimeSpan _operationTimeout;
         private TimeSpan _openTimeout;
-        /// <summary>
-        /// To enable certificate revocation check. Default to be false.
-        /// </summary>
-        public bool CertificateRevocationCheck = false;
```
drwill-ms: If we shipped this bool, we can't remove it.

We could, however, proxy the value to TslVersions.Instance.CertificateRevocationCheck... treat it like a backing field.

shared/src/TlsVersions.cs
```diff
@@ -100,5 +100,10 @@ public void SetLegacyAcceptableVersions()
             System.Net.ServicePointManager.SecurityProtocol = _net451Protocol;
 #endif
         }
+
+        /// <summary>
+        /// To enable certificate revocation check. Default to be true.
```
drwill-ms: properties above methods

shared/src/TlsVersions.cs
```diff
@@ -100,5 +100,10 @@ public void SetLegacyAcceptableVersions()
...
+        /// </summary>
+        public bool CertificateRevocationCheck { get; set; } = true;
```
drwill-ms: Would you also hook this value up to all the other locations we instantiate an HttpClient (or Mqtt or Amqp connection)?

e2e/test/DeviceClientX509AuthenticationE2ETests.cs
bikamani: Even if all of these tests set value to "true", is a problem running it in parallel?

iothub/device/src/AmqpTransportSettings.cs
bikamani: added changes to set it from tlsversions

shared/src/TlsVersions.cs
bikamani: done

e2e/test/DeviceClientX509AuthenticationE2ETests.cs
drwill-ms: Nope, as soon as one of them wants to be false, it will be though.

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -20,7 +20,7 @@ public sealed class AmqpTransportSettings : ITransportSettings
         /// <summary>
         /// To enable certificate revocation check. Default to be false.
         /// </summary>
-        public bool CertificateRevocationCheck = false;
+        public bool CertificateRevocationCheck = TlsVersions.Instance.CertificateRevocationCheck;
```
drwill-ms: Change this to:

```csharp
public bool CertificcateRevocationCheck
{
    get => TlsVersions.Intance.CertificateRevocationCheck;
    set => TlsVersions.Instance.CertificateRevocationCheck = value;
}
```

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -35,7 +35,7 @@ public class MqttTransportSettings : ITransportSettings
         /// </summary>
 #pragma warning disable CA1051 // Do not declare visible instance fields
 #pragma warning disable IDE1006 // Naming Styles
-        public bool CertificateRevocationCheck = false;
+        public bool CertificateRevocationCheck = TlsVersions.Instance.CertificateRevocationCheck;
```
drwill-ms: Same as the Amqp class

shared/src/TlsVersions.cs
```diff
@@ -98,7 +103,8 @@ public void SetLegacyAcceptableVersions()
         {
```
drwill-ms: Update the xml comments

shared/src/TlsVersions.cs
```diff
@@ -47,6 +47,11 @@ internal TlsVersions()
         private const SslProtocols AllowedProtocols = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
         private const SslProtocols PreferredProtocol = SslProtocols.Tls12;
 
+        /// <summary>
+        /// To enable certificate revocation check. Default to be true.
+        /// </summary>
+        public bool CertificateRevocationCheck { get; set; } = true;
```
drwill-ms: You could add a unit test to the TlsVersionsTests file to confirm this defaults to true.

e2e/test/DeviceClientX509AuthenticationE2ETests.cs
bikamani: Is it good idea to user ordering? Or is there a better way to specify "not to run tests in parallel"?

common/src/service/HttpClientHelper.cs
```diff
@@ -50,6 +51,7 @@ internal sealed class HttpClientHelper : IHttpClientHelper
             _preRequestActionForAllRequests = preRequestActionForAllRequests;
             _customHttpProxy = customHttpProxy;
             TlsVersions.Instance.SetLegacyAcceptableVersions();
+            _certificateRevocationCheck = TlsVersions.Instance.CertificateRevocationCheck;
```
drwill-ms: You can cut out this extra variable, and just assign TlsVersions.Instance.CertificateRevocationCheck directly on line 868.

This way if the value is changed after the fact, they don't get the cached value.

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -31,11 +31,15 @@ public class MqttTransportSettings : ITransportSettings
         private static readonly TimeSpan DefaultDeviceReceiveAckTimeout = TimeSpan.FromSeconds(300);
 
         /// <summary>
-        /// To enable certificate revocation check. Default to be false.
+        /// To enable certificate revocation check. Default to be true.
         /// </summary>
 #pragma warning disable CA1051 // Do not declare visible instance fields
```
drwill-ms: Can remove this warning now

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -53,6 +53,8 @@ internal class ContractApiHttp : IContractApiHttp
                 //SslProtocols = TlsVersions.Preferred,
             };
 
+            _httpClientHandler.CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck;
```
drwill-ms: Can this be put up in the initializer list?

shared/tests/Shared/TlsVersionsTests.cs
```diff
@@ -25,6 +25,13 @@ public void Preferred_DefaultsToNone()
             new TlsVersions().Preferred.Should().Be(SslProtocols.None);
         }
 
+        [Fact]
+        public void CheckCertificationList_DefaultsToNone()
```
drwill-ms: DefaultsToTrue

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -18,9 +18,13 @@ public sealed class AmqpTransportSettings : ITransportSettings
         private TimeSpan _operationTimeout;
         private TimeSpan _openTimeout;
         /// <summary>
-        /// To enable certificate revocation check. Default to be false.
+        /// To enable certificate revocation check. Default to be true.
```
drwill-ms: default has changed to false

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -31,13 +31,15 @@ public class MqttTransportSettings : ITransportSettings
         private static readonly TimeSpan DefaultDeviceReceiveAckTimeout = TimeSpan.FromSeconds(300);
 
         /// <summary>
-        /// To enable certificate revocation check. Default to be false.
+        /// To enable certificate revocation check. Default to be true.
```
drwill-ms: comment update here too

shared/tests/Shared/TlsVersionsTests.cs
```diff
@@ -25,6 +25,13 @@ public void Preferred_DefaultsToNone()
             new TlsVersions().Preferred.Should().Be(SslProtocols.None);
         }
 
+        [Fact]
+        public void CheckCertificationList_DefaultsToTrue()
```
drwill-ms: no longer true

shared/tests/Shared/TlsVersionsTests.cs
```diff
@@ -25,6 +25,13 @@ public void Preferred_DefaultsToNone()
...
+        {
+            // assert
+            new TlsVersions().CertificateRevocationCheck.Should().Be(true);
```
drwill-ms: Btw, use `Should().BeFalse()` and `Shoudl().BeTrue()` instead of `Be(bool)`

shared/tests/Shared/TlsVersionsTests.cs
bikamani: right, had missed this, modified.

PR #1286, opened 2020-03-25
fix(iot-service): reuse HttpClient instances

drwill-ms:
HttpClient is IDisposable, but will leak ports used for a period of time, which can lead to port exhaustion if the client sends a lot of requests in a short period of time.

The solution is to hold onto and reuse HttpClient instances for the duration of the client. In our case, we need 2 flavors, so we keep 2 of them.

Also includes:
- code cleanup in the files impacted.
- removing an unused ctor param from HttpClientHelper

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -32,23 +32,59 @@ namespace Microsoft.Azure.Devices
         private readonly IAuthorizationHeaderProvider _authenticationHeaderProvider;
         private readonly IReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> _defaultErrorMapping;
         private readonly TimeSpan _defaultOperationTimeout;
-        private readonly IWebProxy _customHttpProxy;
-        private readonly Action<HttpClient> _preRequestActionForAllRequests;
+
+        // IDisposables
+
+        private readonly HttpClient _httpClientWithDefaultTimeout;
+
+        private readonly HttpClient _httpClientWithNoTimeout;
+        private readonly HttpClientHandler _httpClientHandler;
 
         public HttpClientHelper(
             Uri baseAddress,
             IAuthorizationHeaderProvider authenticationHeaderProvider,
             IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping,
             TimeSpan timeout,
-            Action<HttpClient> preRequestActionForAllRequests,
             IWebProxy customHttpProxy)
         {
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
             _defaultOperationTimeout = timeout;
-            _preRequestActionForAllRequests = preRequestActionForAllRequests;
-            _customHttpProxy = customHttpProxy;
+
+            // HttpClientHandler is IDisposable, so save onto it for disposing.
+            _httpClientHandler = new HttpClientHandler
+            {
+#if !NET451
+                SslProtocols = TlsVersions.Instance.Preferred,
+#endif
+            };
+
+            if (customHttpProxy != DefaultWebProxySettings.Instance)
+            {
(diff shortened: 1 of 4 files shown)


common/src/service/HttpClientHelper.cs
```diff
@@ -826,10 +855,8 @@ public static bool IsMappedToException(HttpResponseMessage message)
 
                 throw new IotHubCommunicationException(string.Format(CultureInfo.InvariantCulture, "The {0} operation timed out.", httpMethod), ex);
             }
-            catch (Exception ex)
+            catch (Exception ex) when (!Fx.IsFatal(ex))
```
drwill-ms: Just don't catch

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -1,121 +1,97 @@
...
+            _faultTolerantSendingLink = new FaultTolerantAmqpObject<SendingAmqpLink>(CreateSendingLinkAsync, Connection.CloseLink);
+            _feedbackReceiver = new AmqpFeedbackReceiver(Connection);
+            _fileNotificationReceiver = new AmqpFileNotificationReceiver(Connection);
+            _iotHubName = iotHubConnectionString.IotHubName;
+            _httpClientHelper = new HttpClientHelper(
                 iotHubConnectionString.HttpsEndpoint,
                 iotHubConnectionString,
                 ExceptionHandlingHelper.GetDefaultErrorMapping(),
-                DefaultOperationTimeout,
-                client => { },
```
drwill-ms: 1 of 3 cases calling the ctor. No impl on this callback.

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -139,25 +115,24 @@ public async override Task SendAsync(string deviceId, Message message, TimeSpan?
...
                     }
                 }
-                catch (TimeoutException exception)
+                catch (TimeoutException)
                 {
-                    throw exception;
+                    throw;
                 }
-                catch (Exception exception)
+                catch (Exception ex) when (!ex.IsFatal())
```
drwill-ms: Just don't catch

iothub/service/src/HttpRegistryManager.cs
```diff
@@ -1273,10 +1274,10 @@ public override Task<Twin> ReplaceTwinAsync(string deviceId, Twin newTwin, strin
 
         public override Task<Twin> ReplaceTwinAsync(string deviceId, Twin newTwin, string etag, CancellationToken cancellationToken)
         {
-            return UpdateTwinInternalAsync(deviceId, newTwin, etag, cancellationToken, true);
+            return UpdateTwinInternalAsync(deviceId, newTwin, etag, true, cancellationToken);
         }
 
-        private Task<Twin> UpdateTwinInternalAsync(string deviceId, Twin twin, string etag, CancellationToken cancellationToken, bool isReplace)
+        private Task<Twin> UpdateTwinInternalAsync(string deviceId, Twin twin, string etag, bool isReplace, CancellationToken cancellationToken)
```
drwill-ms: fixed param order on this private method (and next)

iothub/service/src/JobClient/HttpJobClient.cs
```diff
@@ -1,42 +1,39 @@
...
         internal HttpJobClient(IotHubConnectionString connectionString, HttpTransportSettings transportSettings)
         {
-            this.iotHubName = connectionString.IotHubName;
-            this.httpClientHelper = new HttpClientHelper(
+            _httpClientHelper = new HttpClientHelper(
                 connectionString.HttpsEndpoint,
                 connectionString,
                 ExceptionHandlingHelper.GetDefaultErrorMapping(),
-                DefaultOperationTimeout,
-                client => { },
```
drwill-ms: 2 of 3 cases calling the ctor. No impl on this callback.

iothub/service/src/HttpRegistryManager.cs
```diff
@@ -64,7 +64,6 @@ internal HttpRegistryManager(IotHubConnectionString connectionString, HttpTransp
                 connectionString,
                 ExceptionHandlingHelper.GetDefaultErrorMapping(),
                 s_defaultOperationTimeout,
-                client => { },
```
drwill-ms: 2 of 3 cases calling the ctor. No impl on this callback.

PR #1287, opened 2020-03-25
Add runsettings to MS test project. remove and sort usings

abhipsaMisra:
- Added runsettings to MS test project.
    - This setting currently has the per-test timeout for MS test projects. 
      The current timeout is set to 6mins (Long running tests - DeviceClient_TokenIsRefreshed_Ok_Mqtt, DeviceClient_DeviceDisabled_Gives_ConnectionStatus_DeviceDisabled_AMQP take a little over 5mins to complete.)
- Cleaned up using statements in E2E Test project


test.runsettings
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+   <MSTest>
+      <!-- Specify timeout in milliseconds. A valid value should be greater than 0 -->
+      <TestTimeout>300000</TestTimeout>
```
abhipsaMisra: Check - do the long running tests require > 5mins to complete?

e2e/test/FaultInjectionPoolAmqpTests.MethodFaultInjectionPoolAmqpTests.cs
```diff
@@ -1,13 +1,14 @@
-﻿// Copyright(c) Microsoft.All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Copyright(c) Microsoft.All rights reserved.
```
timtay-microsoft: duplicate header?

test.runsettings
timtay-microsoft: This is per test, correct?

timtay-microsoft: It may be a bit aggressive. A timeout like this should only be used to stop stalling tests, so long running tests may hit this threshold if we keep it at 5 minutes

e2e/test/FaultInjectionPoolAmqpTests.MethodFaultInjectionPoolAmqpTests.cs
abhipsaMisra: Good catch

test.runsettings
abhipsaMisra: yes

abhipsaMisra: I agree, I am running the entire test suite to see how many tests get timeout out with this value

e2e/test/FaultInjectionPoolAmqpTests.MethodFaultInjectionPoolAmqpTests.cs
abhipsaMisra: So this got added because of spacing differences in the existing header vs what is defined in our codemaid settings! 
All the more reason to not add these manually.

test.runsettings
timtay-microsoft: Okay, I trust you will come up with an appropriate value here then, so I'll sign off

abhipsaMisra: Current timeout is set to 6mins (Long running tests - DeviceClient_TokenIsRefreshed_Ok_Mqtt, DeviceClient_DeviceDisabled_Gives_ConnectionStatus_DeviceDisabled_AMQP take a little over 5mins to complete.)

PR #1288, opened 2020-03-26
Expose a way for nested Edge to pass authentication chain to IoT Hub

richma-ms:
To enable nested Edge scenarios, the Edge runtime needs to pass up the chain of all parent Edge devices to IoT Hub when connecting on behalf of a child device.  This means the following for the two different supported protocols.

AMQP:
Add a new link property "AuthChain" of format "edge1;edge2;..." on both the Sending/Receiving links

MQTT:
Augment the username payload with a new "AuthChain" suffix of the format:
$"{iotHubHostName}/{id}/?api-version={version}&DeviceClientType={type}"&AuthChain=edge1;edge2;..."

Since this behavior is Edge specific, it doesn't make sense to expose a public API to support it.  This PR is a proposal for the smallest change we can think of to enable this.

AmqpTransportSettings and MqttTransportSettings now both have a **internal** property that Edge can set at client creation time using System.Reflection.  At connection/link establishing time, AmqpIoTSession and MqttIotHubAdapter would read this property from the corresponding transport setting, and if it exists, add it to the connection payload.

We then would make the corresponding change in IoT Hub service to consume this new payload as part of auth.

### DIFF
--- a/iothub/device/src/AmqpTransportSettings.cs
+++ b/iothub/device/src/AmqpTransportSettings.cs
@@ -17,6 +17,12 @@ namespace Microsoft.Azure.Devices.Client
         private readonly TransportType _transportType;
         private TimeSpan _operationTimeout;
         private TimeSpan _openTimeout;
+
+        /// <summary>
+        /// Used by Edge runtime to specify an authentication chain for Edge-to-Edge connections
+        /// </summary>
+        internal string AuthenticationChain { get; set; }
+        
         /// <summary>
         /// To enable certificate revocation check. Default to be false.
         /// </summary>
--- a/iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
+++ b/iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
@@ -208,6 +208,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
         )
         {
             if (Logging.IsEnabled) Logging.Enter(typeof(AmqpIoTSession), deviceIdentity, $"{nameof(OpenSendingAmqpLinkAsync)}");
+
             AmqpLinkSettings amqpLinkSettings = new AmqpLinkSettings
             {
                 LinkName = linkSuffix,
@@ -224,6 +225,11 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
             if (CorrelationId != null)
             {
                 amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ChannelCorrelationId, CorrelationId);
+            }
+
+            if (!deviceIdentity.AmqpTransportSettings.AuthenticationChain.IsNullOrWhiteSpace())
+            {
+                amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.AuthChain, deviceIdentity.AmqpTransportSettings.AuthenticationChain);
             }
 
             try
@@ -285,6 +291,12 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.TimeoutName, timeout.TotalMilliseconds);
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ClientVersion, deviceIdentity.ProductInfo.ToString());
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ApiVersion, ClientApiVersionHelper.ApiVersionString);
+
+            if (!deviceIdentity.AmqpTransportSettings.AuthenticationChain.IsNullOrWhiteSpace())
+            {
+                amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.AuthChain, deviceIdentity.AmqpTransportSettings.AuthenticationChain);
+            }
+
             if (correlationId != null)
             {
                 amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ChannelCorrelationId, correlationId);
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -43,6 +43,9 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         private const string DeviceCommandTopicFilterFormat = "devices/{0}/messages/devicebound/#";
         private const string DeviceTelemetryTopicFormat = "devices/{0}/messages/events/";
         private const string ModuleTelemetryTopicFormat = "devices/{0}/modules/{1}/messages/events/";
+        private const string ApiVersionParam = "api-version";
+        private const string DeviceClientTypeParam = "DeviceClientType";
+        private const string AuthChainParam = "auth-chain";
         private const char SegmentSeparatorChar = '/';
         private const char SingleSegmentWildcardChar = '+';
         private const char MultiSegmentWildcardChar = '#';
@@ -288,11 +291,18 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                     Debug.Assert(this.mqttTransportSettings.ClientCertificate != null);
                 }
 
+                string usernameString = $"{this.iotHubHostName}/{id}/?{ApiVersionParam}={ClientApiVersionHelper.ApiVersionString}&{DeviceClientTypeParam}={Uri.EscapeDataString(this.productInfo.ToString())}";
+
+                if (!this.mqttTransportSettings.AuthenticationChain.IsNullOrWhiteSpace())
+                {
+                    usernameString += $"&{AuthChainParam}={Uri.EscapeDataString(this.mqttTransportSettings.AuthenticationChain)}";
+                }
+
                 var connectPacket = new ConnectPacket
                 {
                     ClientId = id,
                     HasUsername = true,
-                    Username = $"{this.iotHubHostName}/{id}/?api-version={ClientApiVersionHelper.ApiVersionString}&DeviceClientType={Uri.EscapeDataString(this.productInfo.ToString())}",
+                    Username = usernameString,
                     HasPassword = !string.IsNullOrEmpty(password),
                     Password = password,
                     KeepAliveInSeconds = this.mqttTransportSettings.KeepAliveInSeconds,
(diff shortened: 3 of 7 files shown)


iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
```diff
@@ -285,6 +301,12 @@ TimeSpan timeout
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.TimeoutName, timeout.TotalMilliseconds);
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ClientVersion, deviceIdentity.ProductInfo.ToString());
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ApiVersion, ClientApiVersionHelper.ApiVersionString);
+            
+            if (authenticationChain != null)
```
ancaantochi: please use String.IsNullOrWhiteSpace

iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
```diff
@@ -226,6 +233,11 @@ TimeSpan timeout
                 amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ChannelCorrelationId, CorrelationId);
             }
 
+            if (authenticationChain != null)
```
ancaantochi: please use String.IsNullOrWhiteSpace

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -288,11 +289,22 @@ private async Task Connect(IChannelHandlerContext context)
                     Debug.Assert(this.mqttTransportSettings.ClientCertificate != null);
                 }
 
+                string usernameString = $"{this.iotHubHostName}/{id}/?api-version={ClientApiVersionHelper.ApiVersionString}&DeviceClientType={Uri.EscapeDataString(this.productInfo.ToString())}";
+                string authenticationChain =
+                    (string)typeof(MqttTransportSettings)
+                        .GetField("_authenticationChain", BindingFlags.NonPublic | BindingFlags.Instance)
+                        .GetValue(this.mqttTransportSettings);
+
+                if (authenticationChain != null)
```
ancaantochi: please use String.IsNullOrWhiteSpace

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -288,11 +289,22 @@ private async Task Connect(IChannelHandlerContext context)
...
+                {
+                    usernameString += $"&AuthChain={authenticationChain}";
```
ancaantochi: is it neede to use Uri.EscapeDataString?

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
     public class MqttTransportSettings : ITransportSettings
     {
         private readonly TransportType _transportType;
+        internal string AuthenticationChain;
```
varunpuranik, line 20:
```suggestion
        internal string AuthenticationChain { get; set; }
```

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -17,6 +17,7 @@ public sealed class AmqpTransportSettings : ITransportSettings
         private readonly TransportType _transportType;
         private TimeSpan _operationTimeout;
         private TimeSpan _openTimeout;
+        internal string AuthenticationChain;
```
varunpuranik, line 20:
```suggestion
        internal string AuthenticationChain { get; set; }
```

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -17,6 +17,7 @@ public sealed class AmqpTransportSettings : ITransportSettings
         private readonly TransportType _transportType;
         private TimeSpan _operationTimeout;
         private TimeSpan _openTimeout;
+        internal string AuthenticationChain { get; set; }
```
abhipsaMisra, line 20: Add a comment here explaining the usage, and also indicating that this is specific to IoT Edge implementation.

iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
```diff
@@ -226,6 +227,11 @@ TimeSpan timeout
                 amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ChannelCorrelationId, CorrelationId);
             }
 
+            if (!deviceIdentity.AmqpTransportSettings.AuthenticationChain.IsNullOrWhiteSpace())
+            {
+                amqpLinkSettings.AddProperty("AuthChain", deviceIdentity.AmqpTransportSettings.AuthenticationChain);
```
abhipsaMisra, line 232: Move "AuthChain" to a constant. There is a constants file maintained here: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/AmqpIoT/AmqpIoTConstants.cs

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
     public class MqttTransportSettings : ITransportSettings
     {
         private readonly TransportType _transportType;
+        internal string AuthenticationChain { get; set; }
```
abhipsaMisra, line 20: add a comment here as well

iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
ancaantochi: in Iothub amqp properties names have the format: AmqpConstants.Vendor + ":api-version", where AmqpConstants.Vendor = "com.microsoft". I suggest to use the same format  mqpConstants.Vendor + ":auth-chain"

iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
```diff
@@ -285,6 +291,12 @@ TimeSpan timeout
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.TimeoutName, timeout.TotalMilliseconds);
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ClientVersion, deviceIdentity.ProductInfo.ToString());
             amqpLinkSettings.AddProperty(AmqpIoTErrorAdapter.ApiVersion, ClientApiVersionHelper.ApiVersionString);
+
+            if (!deviceIdentity.AmqpTransportSettings.AuthenticationChain.IsNullOrWhiteSpace())
+            {
+                amqpLinkSettings.AddProperty("AuthChain", deviceIdentity.AmqpTransportSettings.AuthenticationChain);
```
ancaantochi, line 297: auth-chain?

iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
abhipsaMisra: I agree, this is better @ancaantochi !

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -288,11 +288,18 @@ private async Task Connect(IChannelHandlerContext context)
                     Debug.Assert(this.mqttTransportSettings.ClientCertificate != null);
                 }
 
+                string usernameString = $"{this.iotHubHostName}/{id}/?api-version={ClientApiVersionHelper.ApiVersionString}&DeviceClientType={Uri.EscapeDataString(this.productInfo.ToString())}";
+
+                if (!this.mqttTransportSettings.AuthenticationChain.IsNullOrWhiteSpace())
+                {
+                    usernameString += $"&AuthChain={Uri.EscapeDataString(this.mqttTransportSettings.AuthenticationChain)}";
```
ancaantochi, line 295: should this property be auth-chain? It seems both formats are used api-version and AuthChain

richma-ms: sure thing

iothub/device/tests/Microsoft.Azure.Devices.Client.Tests.csproj
```diff
@@ -30,10 +30,10 @@
     <PackageReference Include="MSTest.TestFramework" Version="2.0.0" />
     <PackageReference Include="NSubstitute" Version="3.1.0" />
     <PackageReference Include="Moq" Version="4.10.1" />
+    <Reference Include="System.Web" />
```
abhipsaMisra, line 33: I would like to know the error thrown if you don't specify global import for System.Web

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -288,11 +288,18 @@ private async Task Connect(IChannelHandlerContext context)
                     Debug.Assert(this.mqttTransportSettings.ClientCertificate != null);
                 }
 
+                string usernameString = $"{this.iotHubHostName}/{id}/?api-version={ClientApiVersionHelper.ApiVersionString}&DeviceClientType={Uri.EscapeDataString(this.productInfo.ToString())}";
+
+                if (!this.mqttTransportSettings.AuthenticationChain.IsNullOrWhiteSpace())
+                {
+                    usernameString += $"&auth-chain={Uri.EscapeDataString(this.mqttTransportSettings.AuthenticationChain)}";
```
abhipsaMisra, line 295: Why not use `AmqpIoTErrorAdapter.AuthChain` instead of `auth-chain` directly?

iothub/device/tests/Microsoft.Azure.Devices.Client.Tests.csproj
richma-ms: Error	CS0234	The type or namespace name 'HttpUtility' does not exist in the namespace 'System.Web' (are you missing an assembly reference?)	Microsoft.Azure.Devices.Client.Tests (netcoreapp3.1)	C:\src\azure-iot-sdk-csharp\iothub\device\tests\Mqtt\MqttIotHubAdapterTest.cs	132	N/A

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
richma-ms: AmqpIoTErrorAdapater.AuthChain has the : delimiter, which isn't used here for MQTT

abhipsaMisra: This is mqtt, my bad!

PR #1292, opened 2020-03-26
Specify to output the new snupkg format

drwill-ms:
Perhaps one step closer to https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg

I'd have preferred to add symbols directly to the nupkg file, but apparently that is not the way we do things these days.

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -27,6 +27,9 @@
   <PropertyGroup>
     <Version>1.24.0</Version>
     <Title>Microsoft Azure IoT Device Client SDK</Title>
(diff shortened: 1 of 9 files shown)


drwill-ms:
Not ready to push these to nuget yet. We'll attach them to our release as usual and I'll figure out if we want to publish there, somewhere else, or not.

abhipsaMisra:
We can remove the include-source and include-symbols from our CLI command as well: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/build.ps1#L143

drwill-ms:
> We can remove the include-source and include-symbols from our CLI command as well: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/build.ps1#L143

In the meantime, it doesn't hurt to include them on the command.

PR #1300, opened 2020-03-30
feat(e2e): Change e2e env file name and location

drwill-ms:


e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -37,29 +37,45 @@ $WarningActionPreference = "Continue"
...
         az account set --subscription $SubscriptionId
     }
 
     return $azureContext
 }
 
+Function CleanUp-Certs()
+{
+    Write-Host("`nCleaning up old certs and files that may cause conflicts.")
+    Get-ChildItem "Cert:\LocalMachine\My" | Where-Object { $_.Issuer.Contains("CN=$subjectPrefix") } | Remove-Item
```
drwill-ms: Changes these 3 lines to use the variables instead of hard coding them again.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -360,15 +399,19 @@ if (-not $spExists) {
...
         az role assignment create --role Reader --assignee $appId --resource-group $ResourceGroup --output none
         $retry = $false
     }
-    catch {
+    catch
+    {
         $retryCount++
-        if ($retryCount -eq 10) {
+        if ($retryCount -ge 10)
+        {
```
drwill-ms: What kind of failure causes this to require retry? Should we sleep between retries? If max retries is reached, should the script be allowed to continue?

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -360,15 +399,19 @@ if (-not $spExists) {
...
             Write-Host "Max retries reached"
```
drwill-ms: Write-Error?

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -360,15 +399,19 @@ if (-not $spExists) {
...
         Write-Host "`nGranting $appId Reader role assignment to the $Resourcegroup resource group - Trial $retryCount"
         az role assignment create --role Reader --assignee $appId --resource-group $ResourceGroup --output none
         $retry = $false
     }
-    catch {
+    catch
+    {
         $retryCount++
-        if ($retryCount -eq 10) {
+        if ($retryCount -ge 10)
```
drwill-ms: Using greater than or equal to in order to avoid any possible case where we didn't catch it on exactly 10. Obviously, we don't expect this code to do that, but it is a general best practice when testing boundaries.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -433,41 +479,38 @@ az keyvault secret set --vault-name $secretsKeyVaultName --name "LA-AAD-APP-CERT
...
-Get-ChildItem .\ | Where-Object { $_.Name.EndsWith(".p7b") } | Remove-Item
+CleanUp-Certs
 
-#Creating a file to run to load environment variables
-$currentDir = Get-Location
-Write-Host "`nOutput in '$currentDir\iotPowershellConfig.ps1'"
-$file = New-Item  -Path $currentDir -Name "iotPowershellConfig.ps1" -ItemType "file" -Force
-Add-Content -Path $file.PSPath -Value "Invoke-Expression `".\LoadEnvironmentVariablesFromKeyVault.ps1 -SubscriptionId $SubscriptionId -KeyVaultName $secretsKeyVaultName`""
+# Creating a file to run to load environment variables
+$loadScriptDir = Join-Path $PSScriptRoot "..\..\..\..\.." -Resolve
```
drwill-ms: I chose the simplest method, which is knowing how deep this script is. If someone moves it to another folder, then this would likely break.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -433,41 +479,38 @@ az keyvault secret set --vault-name $secretsKeyVaultName --name "LA-AAD-APP-CERT
...
+$loadScriptName = "Load-$secretsKeyVaultName.ps1"
```
drwill-ms: Script name is based on the key vault name generated.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
vinagesh: The script would fail on error. We don't continue on Error (Set at the beginning of the script). Sleep might not be a bad idea to add but I don't know how much to add. I wanted everyone to try and see and then add it accordingly. I occasionally hit the issue that we created the SP but it said not found when we tried to assign RBAC (I'm guessing we hit a different endpoint where changes were not propagated).

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -37,29 +37,52 @@ $WarningActionPreference = "Continue"
...
+    Write-Host("`nCleaning up old certs and files that may cause conflicts.")
+    Get-ChildItem "Cert:\LocalMachine\My" | Where-Object { $_.Issuer.Contains("CN=$subjectPrefix") } | Remove-Item
+    Get-ChildItem "Cert:\LocalMachine\My" | Where-Object { $_.Issuer.Contains("CN=$groupCertCommonName") } | Remove-Item
+    Get-ChildItem "Cert:\LocalMachine\My" | Where-Object { $_.Issuer.Contains("CN=$deviceCertCommonName") } | Remove-Item
+    Get-ChildItem .\ | Where-Object { $_.Name.EndsWith(".pfx") } | Remove-Item
+    Get-ChildItem .\ | Where-Object { $_.Name.EndsWith(".cer") } | Remove-Item
+    Get-ChildItem .\ | Where-Object { $_.Name.EndsWith(".p7b") } | Remove-Item
+}
+
+$isAdmin = ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole( [Security.Principal.WindowsBuiltInRole] "Administrator")
```
drwill-ms: Admin check

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
vinagesh: Sure makes sense

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
vinagesh: Cool!

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
drwill-ms: I don't see what causes the error to occur. We have a catch. If max retries reached, we exit the loop. What is throwing the error though?

vinagesh: I see what you are saying. We need to error it out to hint the user to run the script again. If we continue without failing, user can assume everything is done but some E2E tests might fail as the permissions were not assigned.

Issue #1276, opened 2020-03-18
RegistryManager causes TCP socket exhaustion under load


smolyakoff:
- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64, Azure App Service (P1V2 plan on Windows)
- **Application's .NET Target Framework :** netcoreapp2.2
- **SDK version used:** Microsoft.Azure.Devices 1.20.0

## Description of the issue:
We were running performance tests for our system that relies on IoT Hub. Under concurrent load from about 50 threads performing HTTP requests to our API endpoints, we started hitting `System.Net.Sockets.SocketException` with the following exception message:

```
Only one usage of each socket address (protocol/network address/port) is normally permitted
```

We use `RegistryManager`'s twin methods (retrieving and updating twins) quite heavily. After further investigation, we found out that too many TCP sockets stay in "TIME_WAIT" state blocking further TCP connections to be established. The limit is around 2K active sockets on Azure App Service P1V2 Plan we were using for testing.

![TCP Socket Usage](https://user-images.githubusercontent.com/1292603/76952436-25109300-691e-11ea-9b52-19ce5c990dd9.png)

I've checked the SDK's source code and it appears that a new `HttpClient` and `HttpClientHandler` instance is created for each independent request. See the [following line](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/common/src/service/HttpClientHelper.cs#L848) in `HttpClientHelper` class used for HTTP communication.

It would be great to have this fixed as it's the only limit that we are hitting during the test and other resources usage (CPU/memory) is pretty low.

## Code sample exhibiting the issue:

The following code sample reproduces the issue. There should be some devices registered in IoT Hub instance to be able to run the code sample. The number of active sockets grows quickly when the `RegistryClient` is used as seen in the console output.

```csharp
using System;
using System.Collections.Generic;
using System.Linq;
using System.Net.Http;
using System.Net.NetworkInformation;
using System.Threading;
using System.Threading.Tasks;
using Microsoft.Azure.Devices;
using Newtonsoft.Json.Linq;

namespace IoTHubSocketExhaustion
{
    internal class Program
    {
        internal static async Task Main(string[] args)
        {
            var cts = new CancellationTokenSource();
            Console.CancelKeyPress += (s, e) =>
            {
                e.Cancel = true;
                cts.Cancel();
            };
            var registryManager = RegistryManager.CreateFromConnectionString(
[...]

PR #1303, opened 2020-03-31
fix(iot-service): reuse HttpClient instances (#1276)

drwill-ms:
Compare to the fix made to master: https://github.com/Azure/azure-iot-sdk-csharp/commit/5e09aae4a14aa7a1b47994960644e5c05f44a36b

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -1,56 +1,91 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.Azure.Devices
-{
-    using Microsoft.Azure.Devices.Common;
-    using Microsoft.Azure.Devices.Common.Exceptions;
-    using Microsoft.Azure.Devices.Common.Extensions;
-    using Microsoft.Azure.Devices.Shared;
-    using System;
-    using System.Collections.Generic;
-    using System.Collections.ObjectModel;
-    using System.Globalization;
-    using System.IO;
-    using System.Linq;
-    using System.Net;
-    using System.Net.Http;
-    using System.Net.Http.Headers;
-    using System.Threading;
-    using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Common;
+using Microsoft.Azure.Devices.Common.Exceptions;
+using Microsoft.Azure.Devices.Common.Extensions;
+using Microsoft.Azure.Devices.Shared;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+
 #if NET451
     using System.Net.Http.Formatting;
 #endif
 
-    sealed class HttpClientHelper : IHttpClientHelper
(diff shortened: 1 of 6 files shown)


common/src/service/HttpClientHelper.cs
```diff
@@ -1,56 +1,84 @@
...
-        readonly IWebProxy customHttpProxy;
-        readonly Action<HttpClient> preRequestActionForAllRequests;
+        private readonly Uri _baseAddress;
+        private readonly IAuthorizationHeaderProvider _authenticationHeaderProvider;
+        private readonly IReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> _defaultErrorMapping;
+        private readonly TimeSpan _defaultOperationTimeout;
 
+        // IDisposables
+
+        private readonly HttpClient _httpClientWithDefaultTimeout;
```
drwill-ms: They key part of the fix is caching the only 2 instances of HttpClient we need, rather than building a new one for each request.

common/src/service/HttpClientHelper.cs
```diff
@@ -737,7 +774,14 @@ public static bool IsMappedToException(HttpResponseMessage message)
             return isMappedToException;
         }
 
-        async Task ExecuteAsync(
+        public void Dispose()
+        {
+            _httpClientHandler.Dispose();
```
drwill-ms: Cached instances need disposal along with this class now.

common/src/service/HttpClientHelper.cs
```diff
@@ -701,19 +736,21 @@ static void InsertEtag(HttpRequestMessage requestMessage, string etag)
         {
             cancellationToken.ThrowIfCancellationRequested();
 
-            using (var httpClient = BuildHttpClient(Timeout.InfiniteTimeSpan))
```
drwill-ms: don't build now, instead use the cached one without timeout, because this one relies on a custom cancellation token

common/src/service/HttpClientHelper.cs
```diff
@@ -652,44 +689,42 @@ static void InsertEtag(HttpRequestMessage requestMessage, string etag)
...
-        async Task ExecuteAsync(
+        private async Task ExecuteAsync(
             HttpMethod httpMethod,
             Uri requestUri,
             Func<HttpRequestMessage, CancellationToken, Task> modifyRequestMessageAsync,
             Func<HttpResponseMessage, CancellationToken, Task> processResponseMessageAsync,
             IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> errorMappingOverrides,
             CancellationToken cancellationToken)
         {
-            using (var httpClient = BuildHttpClient(this.defaultOperationTimeout))
```
drwill-ms: Don't build now, instead use the cached one with default timeout

common/src/service/HttpClientHelper.cs
```diff
@@ -80,61 +115,64 @@ sealed class HttpClientHelper : IHttpClientHelper
...
+                            new Uri(_baseAddress, requestUri),
+                            (requestMsg, token) => AddCustomHeaders(requestMsg, customHeaders),
+                            async (message, token) => result = await ReadResponseMessageAsync<T>(message, token).ConfigureAwait(false),
+                            errorMappingOverrides,
+                            cancellationToken)
+                        .ConfigureAwait(false);
                 }
                 else
                 {
-                    using (var httpClient = BuildHttpClient(defaultOperationTimeout))
```
drwill-ms: Don't build the client, use the cached version with default timeout.

PR #1311, opened 2020-04-03
refactor(e2e): Cleanup and add default parameters

vinagesh:
## Description of the changes
Cleaning up the ARM template and defining default values for inputs

https://msazure.visualstudio.com/One/_workitems/edit/6007037

### DIFF
--- a/e2e/test/prerequisites/E2ETestsSetup/e2eTestsArmTemplate.json
+++ b/e2e/test/prerequisites/E2ETestsSetup/e2eTestsArmTemplate.json
@@ -2,79 +2,141 @@
     "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
     "contentVersion": "1.0.0.0",
     "parameters": {
-        "Region": {
-            "type": "string",
-            "metadata": {
-                "description": "The main region for resources created by this template."
-            }
-        },
-        "FarRegion": {
-            "type": "string",
-            "metadata": {
-                "description": "The region for the second IoT hub in a DPS that is far away from the test devices."
-            }
-        },
-        "StorageAccountName": {
-            "type": "string",
-            "minLength": 3,
-            "maxLength": 24,
-            "metadata": {
-                "description": "The name of the storage account used by the IoT hub."
-            }
-        },
-        "ResourceGroup": {
-            "type": "string",
-            "metadata": {
-                "description": "The name of the resource group to create all resources used by tests."
-            }
-        },
         "UserObjectId": {
             "type": "string",
             "metadata": {
                 "description": "Signed in users's objectId"
             }
         },
+        "DpsCustomAllocatorRunCsxContent": {
+            "type": "string",
+            "metadata": {
+                "description": "Contents of run.csx file for the azure function."
+            }
+        },
+        "DpsCustomAllocatorProjContent": {
+            "type": "string",
+            "metadata": {
+                "description": "Contents of proj file for the azure function."
+            }
+        },
+        "KeyVaultName": {
+            "type": "string",
+            "defaultValue": "[uniqueString(resourceGroup().id)]",
+            "metadata": {
+                "description": "The name of the KeyVault that holds the secrets."
+            }
+        },
+        "BlobServiceName": {
+            "type": "string",
+            "defaultValue": "default",
+            "metadata": {
+                "description": "The name of BlobService inside the StorageAccount."
+            }
+        },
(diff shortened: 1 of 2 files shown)


e2e/test/prerequisites/E2ETestsSetup/e2eTestsArmTemplate.json
```diff
@@ -2,32 +2,85 @@
     "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
     "contentVersion": "1.0.0.0",
     "parameters": {
-        "Region": {
+        "KeyVaultName": {
```
drwill-ms, line 5: Please move all the parameters with a default after all the ones that don't have one.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsArmTemplate.json
```diff
@@ -36,45 +89,54 @@
...
+            "defaultValue": "[resourceGroup().name]",
             "metadata": {
                 "description": "The name of the DPS used by tests."
             }
-        },        
+        },
         "DpsCustomAllocatorFunctionName": {
             "type": "string",
-            "defaultValue": "DpsCustomAllocator"
+            "defaultValue": "DpsCustomAllocator",
```
drwill-ms, line 113: I think we should actually call this one `DpsCustomAllocatorByHubLength`, because that is the logic it uses. If we ever chose to add more custom allocators, we'd want to distinguish them.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsArmTemplate.json
```diff
@@ -282,8 +344,8 @@
         {
             "type": "Microsoft.OperationalInsights/workspaces",
             "apiVersion": "2017-03-15-preview",
-            "name": "[parameters('ResourceGroup')]",
-            "location": "[parameters('Region')]",
+            "name": "[parameters('WorkSpaceName')]",
```
drwill-ms, line 347: I think "Workspace" is generally considered to not have a capital S

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -84,13 +84,8 @@ if (-not $isAdmin)
 #################################################################################################
 
 $Region = $Region.Replace(' ', '')
-
-$storageAccountName = "$($ResourceGroup.ToLower())sa"
-$hubName = $ResourceGroup
+$ResourceGroup = $ResourceGroup.ToLower()
```
drwill-ms, line 87: Why lower case it?

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -464,7 +453,7 @@ az keyvault secret set --vault-name $keyVaultName --name "IOTHUB-EVENTHUB-CONN-S
 az keyvault secret set --vault-name $keyVaultName --name "IOTHUB-EVENTHUB-COMPATIBLE-NAME" --value $eventResourceGroup --output none
 az keyvault secret set --vault-name $keyVaultName --name "IOTHUB-EVENTHUB-CONSUMER-GROUP" --value $consumerGroups --output none
 az keyvault secret set --vault-name $keyVaultName --name "IOTHUB-PROXY-SERVER-ADDRESS" --value $proxyServerAddress --output none
-az keyvault secret set --vault-name $keyVaultName --name "FAR-AWAY-IOTHUB-HOSTNAME" --value "$farHubName.azure-devices.net" --output none
```
drwill-ms, line 467: Was this a bug?

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
vinagesh: Not needed to do this anymore.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
vinagesh: Not a bug. Previously the $farHubName was set in the .ps1 to be equal to ResouceGroupName. Now i'm no longer doing it. There is a default in the ARM template and the template will just output the hostname. There is no need ot the concatenation anymore.

PR #1313, opened 2020-04-03
Add a RegistryManager query test and minor dispose clean up

drwill-ms:


common/src/service/HttpClientHelper.cs
```diff
@@ -288,7 +293,7 @@ internal sealed class HttpClientHelper : IHttpClientHelper
                     requestMsg.Content = new ObjectContent<T>(entity, JsonFormatter);
 #else
                     var str = Newtonsoft.Json.JsonConvert.SerializeObject(entity);
```
barustum, line 295: 'var' to 'string'

common/src/service/HttpClientHelper.cs
```diff
@@ -25,6 +27,8 @@ namespace Microsoft.Azure.Devices
 {
     internal sealed class HttpClientHelper : IHttpClientHelper
     {
+        private const string ApplicationJson = "application/json";
```
drwill-ms, line 30: Centralize this string.

common/src/service/HttpClientHelper.cs
```diff
@@ -312,7 +317,7 @@ internal sealed class HttpClientHelper : IHttpClientHelper
                     requestMsg.Content = new ObjectContent<T>(entity, JsonFormatter);
 #else
                     var str = Newtonsoft.Json.JsonConvert.SerializeObject(entity);
```
barustum, line 319: 'var' to 'string'

common/src/service/HttpClientHelper.cs
```diff
@@ -70,15 +74,15 @@ internal sealed class HttpClientHelper : IHttpClientHelper
             // We need two types of HttpClients, one with our default operation timeout, and one without. The one without will rely on
             // a cancellation token.
 
-            _httpClientWithDefaultTimeout = new HttpClient(_httpClientHandler)
+            _httpClientWithDefaultTimeout = new HttpClient(_httpClientHandler, false)
```
drwill-ms, line 77: False means don't dispose the handler. We'll do it ourselves because we share it with the 2 HttpClient instances.

e2e/test/RegistryManagerE2ETests.cs
```diff
@@ -104,5 +107,49 @@ public async Task RegistryManager_AddDeviceWithProxy()
...
+            string deviceId = $"{_devicePrefix}{Guid.NewGuid()}";
+
+            try
+            {
+                Device device = await registryManager
+                    .AddDeviceAsync(new Device(deviceId))
+                    .ConfigureAwait(false);
+
+                // act
+
```
barustum, line 125: extra new line

drwill-ms: They are intentional.

The "// act" applies to the entire block under it, not just the next line.

So generally, when that is the case, we add an extra new line to communicate that.

PR #1316, opened 2020-04-06
ARM template allow a different region for website

drwill-ms:
Antares capacity varies by region, so allow website to a different region, and default it to one where capacity is currently available.


e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -506,6 +506,7 @@ az keyvault secret set --vault-name $keyVaultName --name "DPS-GLOBALDEVICEENDPOI
 # Below Environment variables are only used in Java
```
vinagesh, line 506: Pass the WebRegion as optional parameter so that it helps when we create this resource?

drwill-ms: YAGNI?

PR #1320, opened 2020-04-08
fix(iot-dev,tests): Code cleanup, tests, and disposing

drwill-ms:
While investigating an AMQP bug, I added some tests for a helper class, cleaned up code, and found several cases of IDisposables that weren't being disposed.


common/test/TestModule.cs
```diff
@@ -2,16 +2,15 @@
...
-using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 namespace Microsoft.Azure.Devices.E2ETests
 {
     public class TestModule
     {
-        private Module _module;
+        private static readonly Regex s_hostNameRegex = new Regex("(?<=HostName=).*?(?=;)", RegexOptions.Compiled);
```
drwill-ms, line 12: If you are interested in knowing more about this regex, it employs a negative look behind to require, but not match on, some text ahead of the match we want. Then it also uses a negative look ahead to know when to stop the match.

This way you get exactly what you want, and don't have to parse various matches and index on the one you want.

Also note, I used the non-greedy matcher because there are multiple semi-colons in the connection string, so you want to match to the smallest unit to just get the host name.

e2e/test/Helpers/HostNameHelper.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Text.RegularExpressions;
+
+namespace Microsoft.Azure.Devices.E2ETests.Helpers
+{
+    internal static class HostNameHelper
+    {
+        private static readonly Regex s_hostNameRegex = new Regex("(?<=HostName=).*?(?=;)", RegexOptions.Compiled);
```
drwill-ms, line 7: If you are interested in knowing more about this regex, it employs a negative look behind to require, but not match on, some text ahead of the match we want. Then it also uses a negative look ahead to know when to stop the match.

This way you get exactly what you want, and don't have to parse various matches and index on the one you want.

Also note, I used the non-greedy matcher because there are multiple semi-colons in the connection string, so you want to match to the smallest unit to just get the host name.

common/test/TestDevice.cs
```diff
@@ -25,26 +21,23 @@ public enum ConnectionStringAuthScope
         Device
     }
 
-    public class TestDevice
+    public class TestDevice : IDisposable
```
drwill-ms, line 24: Has a SemaphoreSlim which is IDisposable, so let's make this class disposable to stay cleaned up, even during a test run.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -155,19 +153,21 @@ internal sealed class MqttTransportHandler : TransportHandler, IMqttIotHubEventH
...
+        public TimeSpan TwinTimeout { get; set; } = TimeSpan.FromSeconds(60);
+        public override bool IsUsable => State != TransportState.Closed && State != TransportState.Error;
 
         #region Client operations
 
-        public override Task OpenAsync(TimeoutHelper timeoutHelper)
+        public override async Task OpenAsync(TimeoutHelper timeoutHelper)
         {
-            return OpenAsync(new CancellationTokenSource(timeoutHelper.RemainingTime()).Token);
+            using var cts = new CancellationTokenSource(timeoutHelper.GetRemainingTime());
```
drwill-ms, line 169: Was leaking this CTS

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -280,51 +281,49 @@ private Message ProcessMessage(Message message, bool hasMessage)
...
-            var disconnectToken = this.disconnectAwaitersCancellationSource.Token;
-            this.EnsureValidState();
+            var disconnectToken = _disconnectAwaitersCancellationSource.Token;
+            EnsureValidState();
 
-            using (CancellationTokenSource linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, disconnectToken))
-            {
-                return await this.receivingSemaphore.WaitAsync(timeout, linkedCts.Token).ConfigureAwait(true);
-            }
+            using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, disconnectToken);
```
drwill-ms, line 287: Another leak

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -638,136 +635,132 @@ private bool TryStop()
...
+            EnsureValidState();
 
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_005:  `SendMethodResponseAsync` shall allocate a `Message` object containing the method response.
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_006:  `SendMethodResponseAsync` shall set the message topic to '$iothub/methods/res/<STATUS>/?$rid=<REQUEST_ID>' where STATUS is the return status for the method and REQUEST_ID is the request ID received from the service in the original method call.
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_007:  `SendMethodResponseAsync` shall set the message body to the response payload of the `Method` object.
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_008:  `SendMethodResponseAsync` shall send the message to the service.
-            var message = new Message(methodResponse.BodyStream);
-
-            message.MqttTopicName = methodResponseTopic.FormatInvariant(methodResponse.Status, methodResponse.RequestId);
+            using var message = new Message(methodResponse.BodyStream)
```
drwill-ms, line 711: Another leak? Seems save to dispose this one here.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -638,136 +635,132 @@ private bool TryStop()
...
+            status = 500;
+            return false;
         }
 
         private async Task<Message> SendTwinRequestAsync(Message request, string rid, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
 
-            var responseReceived = new SemaphoreSlim(0);
+            using var responseReceived = new SemaphoreSlim(0);
```
drwill-ms, line 749: Looks limited to this method, so should be disposed when finished.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -805,19 +798,18 @@ private async Task<Message> SendTwinRequestAsync(Message request, string rid, Ca
...
         {
             cancellationToken.ThrowIfCancellationRequested();
 
-            Twin twin = null;
-            this.EnsureValidState();
+            EnsureValidState();
 
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_014:  `SendTwinGetAsync` shall allocate a `Message` object to hold the `GET` request
-            var request = new Message();
+            using var request = new Message();
```
drwill-ms, line 812: Also seems safe to dispose within this method

iothub/device/src/Transport/TransportHandlerBase.cs
```diff
@@ -1,298 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
```
drwill-ms, line 1: Unreferenced internal class.

iothub/device/src/net451/IotHubClientWebSocket.cs
```diff
@@ -1,73 +1,73 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
```
drwill-ms, line 1: Not sure how I first found this file, but now I can't find it in the solution. There is another with the same name, but this one from the net451 folder doesn't seem to be referenced in the solution.

iothub/device/src/net451/IotHubClientWebSocket.cs
```diff
@@ -834,106 +765,99 @@ string BuildUpgradeRequest()
...
-
-            Stream Stream { get; set; }
-
-            byte[] Buffer { get; set; }
-
-            int TotalBytesRead { get; set; }
-
-            int bytesRead;
+            // Commented out to avoid FxCop error for uncalled private code
+            //public string HttpVersion { get; private set; }
```
timtay-microsoft, line 828: Can you delete this commented out property?

iothub/device/src/Transport/TransportHandlerBase.cs
timtay-microsoft: Good find

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Good catch

e2e/test/Helpers/HostNameHelper.cs
vinagesh: Why don't we just store the IotHubHostName in an Env variable? We do that for Far away hub. That way we don't need to do the regex parsing.

drwill-ms: Actually, now that you mention it, I'd rather get rid of the far hub hostname and just parse the connection string.

e2e/test/TwinE2ETests.cs
```diff
@@ -202,52 +202,50 @@ public async Task Twin_ClientSetsReportedPropertyWithoutDesiredPropertyCallback(
...
+            await Task.Delay(TimeSpan.FromSeconds(10)).ConfigureAwait(false);
 
-                // clean up
-                await deviceClient.CloseAsync().ConfigureAwait(false);
-            }
+            // assert
+            Assert.AreEqual(0, connectionStatusChangeCount, "AMQP should not be disconnected.");
+
+            // clean up
+            await deviceClient.CloseAsync().ConfigureAwait(false);
```
timtay-microsoft, line 229: does the using statement already close the device client for us?

e2e/test/TwinE2ETests.cs
```diff
@@ -202,52 +202,50 @@ public async Task Twin_ClientSetsReportedPropertyWithoutDesiredPropertyCallback(
...
-            using (var deviceClient = DeviceClient.CreateFromConnectionString(testDevice.ConnectionString, transport))
-            {
-                await Twin_DeviceSetsReportedPropertyAndGetsItBack(deviceClient).ConfigureAwait(false);
-                await deviceClient.CloseAsync().ConfigureAwait(false);
-            }
+            using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(_devicePrefix).ConfigureAwait(false);
+            using var deviceClient = DeviceClient.CreateFromConnectionString(testDevice.ConnectionString, transport);
+
+            await Twin_DeviceSetsReportedPropertyAndGetsItBack(deviceClient).ConfigureAwait(false);
+            await deviceClient.CloseAsync().ConfigureAwait(false);
```
timtay-microsoft, line 238: same as above, does the using statement close the device client for us?

e2e/test/Helpers/HostNameHelper.cs
timtay-microsoft: Don't forget to update keyvault appropriately if you do. We only really need the hostname for our e2e tests, but I'm not opposed to this change

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -155,19 +153,21 @@ internal sealed class MqttTransportHandler : TransportHandler, IMqttIotHubEventH
...
-                this.channelFactory = channelFactory;
+                _channelFactory = channelFactory;
             }
 
-            this.closeRetryPolicy = new RetryPolicy(new TransientErrorIgnoreStrategy(), 5, TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(1));
+            _closeRetryPolicy = new RetryPolicy(new TransientErrorIgnoreStrategy(), 5, TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(1));
         }
 
-        public override bool IsUsable => this.State != TransportState.Closed && this.State != TransportState.Error;
+        public TimeSpan TwinTimeout { get; set; } = TimeSpan.FromSeconds(60);
```
abhipsaMisra, line 162: nit: we should save this value in a readonly var above; for easy reference.

iothub/device/src/net451/IotHubClientWebSocket.cs
abhipsaMisra: We should delete this then. It's an internal class, and if don't reference it anywhere, let's delete it.

e2e/test/TwinE2ETests.cs
abhipsaMisra: The end effect of disposing the device client instance would be that transport layer is marked as closed, retry policies/ token refreshers etc are all cancelled gracefully, and then the object is marked as disposed. So yes, disposing the client would essentially close the connection and mark the resources as disposed.

PR #1322, opened 2020-04-09
fix(iot-dev,iot-service,prov-dev,prov-service): dispose CancellationTokenSource

drwill-ms:


iothub/device/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -166,24 +162,24 @@ private async Task<int> ReadAsyncCore(TransportAsyncCallbackArgs args)
...
         }
 
         protected override bool CloseInternal()
         {
-            var webSocketState = this.webSocket.State;
+            WebSocketState webSocketState = _webSocket.State;
             if (webSocketState != WebSocketState.Closed && webSocketState != WebSocketState.Aborted)
             {
-                this.CloseInternalAsync(CloseTimeout);
+                CloseInternalAsync(s_closeTimeout).GetAwaiter().GetResult();
```
abhipsaMisra, line 182: We weren't awaiting the completion of the task here!

drwill-ms: Crazy, right?

PR #1323, opened 2020-04-09
fix(iot-dev): don't dispose a message being returned

drwill-ms:
A long-running test failed, making it clear one of the Message disposals shouldn't have been called.


iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -239,7 +239,7 @@ public override async Task<Message> ReceiveAsync(CancellationToken cancellationT
 
             byte[] byteContent = await responseMessage.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
 
-            using var message = byteContent != null
```
drwill-ms, line 242: This method returns that message, so it should be disposed by the caller. Oopsie!

Issue #1063, opened 2019-09-16
MQTT Websocket DeviceClient setup with HTTP Proxy relies on local DNS


kiranpradeep:
- **OS, version, SKU and CPU architecture used:** Ubuntu 18.04 
- **Application's .NET Target Framework :** netcoreapp 2.1
- **Device:** Laptop
- **SDK version used:** 1.21.1

## Description of the issue:
I assume, when connecting behind HTTP proxy, the local DNS server should have no impact on the DeviceClient (?). I tried running MessageSample [below link] inside a docker container as below. In step 2, please note the _--dns_ flag with invalid IP.

1. dotnet publish -c Release -o /tmp/out 
2. docker run --rm -it --dns=128.0.0.1 -v /tmp/out:/app microsoft/dotnet:2.1-runtime-stretch-slim /bin/bash 
3.   cd /app
4.   export https_proxy=PROXY_IP:PROXY_PORT
5.   dotnet MessageSample.dll

Below are observations.

1. The sample works perfectly using AMQP_WS under the same invalidated DNS server.  The issue is only for MQTT_WS protocol. 
2. The same program works perfectly with MQTT_WS if the DNS server is not wrong (not using wrong DNS with --dns=128.0.0.1 )

## Code sample exhibiting the issue:
https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/master/iot-hub/Samples/device/MessageSample

## Console log of the issue:
> Device sending 5 messages to IoTHub...
> Sending message: 0, Data: [{"messageId":0,"temperature":30,"humidity":79}]

> Unhandled Exception: Microsoft.Azure.Devices.Client.Exceptions.IotHubCommunicationException: Transient network error occurred, please retry. ---> System.Net.Internals.SocketExceptionFactory+ExtendedSocketException: Resource temporarily unavailable
>    at System.Net.Dns.InternalGetHostByName(String hostName)
>    at System.Net.Dns.ResolveCallback(Object context)
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Net.Dns.HostResolutionEndHelper(IAsyncResult asyncResult)
>    at System.Net.Dns.EndGetHostAddresses(IAsyncResult asyncResult)
>    at System.Net.Dns.<>c.<GetHostAddressesAsync>b__25_1(IAsyncResult asyncResult)
[...]

PR #1326, opened 2020-04-13
fix(iot-dev): Remove unnecessary DNS lookup in mqtt stack

timtay-microsoft:
When a proxy is present, there is no need to lookup the DNS of the host as the proxy will handle that. 

See #1063

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -568,7 +568,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
         private async Task OpenAsyncInternal(CancellationToken cancellationToken)
         {
-            if (_webProxy != DefaultWebProxySettings.Instance)
+            if (isProxyConfigured())
             {
                 //No need to do a DNS lookup since we have the proxy address already
                 _serverAddresses = new IPAddress[0];
@@ -976,7 +976,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
                 try
                 {
-                    if (_webProxy != DefaultWebProxySettings.Instance)
+                    if (isProxyConfigured())
                     {
                         // Configure proxy server
                         websocket.Options.Proxy = _webProxy;
@@ -1107,6 +1107,11 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
             if (Logging.IsEnabled) Logging.Info(null, "EventLoopGroup threads count was not set.");
             return new MultithreadEventLoopGroup();
+        }
+
+        private bool isProxyConfigured()
+        {
+            return _webProxy != null && _webProxy != DefaultWebProxySettings.Instance;
         }
     }
 }
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -568,7 +568,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
         private async Task OpenAsyncInternal(CancellationToken cancellationToken)
         {
-            if (isProxyConfigured())
+            if (IsProxyConfigured())
             {
                 //No need to do a DNS lookup since we have the proxy address already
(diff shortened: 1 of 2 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -962,15 +974,12 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
                 var websocket = new ClientWebSocket();
                 websocket.Options.AddSubProtocol(WebSocketConstants.SubProtocols.Mqtt);
 
-                // Check if we're configured to use a proxy server
-                IWebProxy webProxy = settings.Proxy;
-
                 try
                 {
-                    if (webProxy != DefaultWebProxySettings.Instance)
+                    if (_webProxy != DefaultWebProxySettings.Instance)
```
drwill-ms, line 979: could it be null or invalid either?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -565,11 +568,20 @@ private TransportState MoveToStateIfPossible(TransportState destination, Transpo
 
         private async Task OpenAsyncInternal(CancellationToken cancellationToken)
         {
+            if (_webProxy != DefaultWebProxySettings.Instance)
```
drwill-ms, line 571: Might put this logic in a helper method, because it is not necessarily an obvious conclusion by someone new to the code, and we repeat it several times.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -565,11 +568,20 @@ private TransportState MoveToStateIfPossible(TransportState destination, Transpo
...
+                _serverAddresses = new IPAddress[0];
+            }
+            else
+            {
 #if NET451
-            _serverAddresses = Dns.GetHostEntry(_hostName).AddressList;
+                _serverAddresses = Dns.GetHostEntry(_hostName).AddressList;
 #else
-            _serverAddresses = await Dns.GetHostAddressesAsync(_hostName).ConfigureAwait(true);
+                _serverAddresses = await Dns.GetHostAddressesAsync(_hostName).ConfigureAwait(true);
```
drwill-ms, line 581: why configureawait true?

timtay-microsoft: Not sure, I can set it to false though

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Sure thing

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: I can add a null check to the helper function that will replace this line, sure

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Actually, that seems to be a pattern throughout the entire class. I'll hold off on making that change since it isn't strictly relevant to this PR

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -962,15 +974,12 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
                 var websocket = new ClientWebSocket();
                 websocket.Options.AddSubProtocol(WebSocketConstants.SubProtocols.Mqtt);
 
-                // Check if we're configured to use a proxy server
-                IWebProxy webProxy = settings.Proxy;
-
                 try
                 {
-                    if (webProxy != DefaultWebProxySettings.Instance)
+                    if (isProxyConfigured())
```
drwill-ms, line 979: C# case would be IsProxyConfigured

timtay-microsoft: I guess I woke up on the Java side of my bed this morning, oops

PR #1327, opened 2020-04-13
Code cleanup, dispose client web socket instance in mqtt transport handler

abhipsaMisra:


common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -286,7 +285,7 @@ public override bool Equals(object obj)
                 return true;
             }
 
-            if (ReferenceEquals(obj, null))
+            if (obj is null)
```
vinagesh, line 288: is is usually used for type comparison. Why not obj == null?

abhipsaMisra: Good point; for `null` is resolves to the same result; so I used the `is` operator. We are also overriding the `==` operator below, so I didn't want to depend on it for this comparison.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -301,12 +300,7 @@ public override int GetHashCode()
 
         public static bool operator ==(ClientWebSocketChannel left, ClientWebSocketChannel right)
         {
-            if (ReferenceEquals(left, null))
-            {
-                return ReferenceEquals(right, null);
-            }
-
-            return left.Equals(right);
+            return left is null ? right is null : left.Equals(right);
```
vinagesh, line 303: Same here, why are we using the 'is' operator and not == ? Also I don't get the logic here. Are we just trying to check if these two are the same object (at least that is my understanding looking at the code) ? In that case, you can just return left.Equals(right).

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
```diff
@@ -328,7 +326,7 @@ private async Task VerifyPublishPacketTopicAsync(IChannelHandlerContext context,
                 {
                     if (statusCode >= HttpStatusCode.BadRequest)
                     {
-                        var errorDetails = JsonConvert.DeserializeObject<ProvisioningErrorDetailsMqtt>(jsonData);
+                        ProvisioningErrorDetailsMqtt errorDetails = JsonConvert.DeserializeObject<ProvisioningErrorDetailsMqtt>(jsonData);
```
vinagesh, line 329: Do you need to specify type here as you're specifying it during deserialization?

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -31,7 +31,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
     /// </summary>
     public class ProvisioningTransportHandlerMqtt : ProvisioningTransportHandler
     {
-        private static MultithreadEventLoopGroup s_eventLoopGroup = new MultithreadEventLoopGroup();
+        private static readonly MultithreadEventLoopGroup s_eventLoopGroup = new MultithreadEventLoopGroup();
```
vinagesh, line 34: Would be good to rename this variable. C# naming guidelines tell us not to use underscores.

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -56,14 +56,7 @@ public class ProvisioningTransportHandlerMqtt : ProvisioningTransportHandler
...
             FallbackType = transportFallbackType;
-            if (FallbackType == TransportFallbackType.WebSocketOnly)
-            {
-                Port = WsPort;
-            }
-            else
-            {
-                Port = MqttTcpPort;
-            }
+            Port = FallbackType == TransportFallbackType.WebSocketOnly ? WsPort : MqttTcpPort;
```
vinagesh, line 59: why is the above line needed? can we just not use transportFallbackType in the comparison?

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -316,34 +310,36 @@ public override int GetHashCode()
...
-            return ReferenceEquals(left, null) ? ReferenceEquals(right, null) : left.CompareTo(right) >= 0;
+            return left is null ? right is null : left.CompareTo(right) >= 0;
         }
 
 #pragma warning disable CA1822
+
         public int CompareTo(ClientWebSocketChannel other)
         {
-            if (ReferenceEquals(other, null))
+            if (other is null)
```
vinagesh, line 335: == ?

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
timtay-microsoft: +1 from me. Var seems appropriate here, still

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
drwill-ms: Our style guidelines state to use underscore as a prefix for private fields, and s_ for private static fields.

The mono repo guidelines are the same, actually.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
drwill-ms: https://www.gullberg.tk/blog/is-null-versus-null-in-c/

drwill-ms: But I agree, I'm used to seeing `== null` checks

vinagesh: Yeah I saw that the == is overloaded below. The logic was confusing in the beginning but I think I get it now.

vinagesh: For null, using == will surely resolve to the same even if you have an overload.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
vinagesh: I think you can leave it as 'is' as these are all overload methods

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -316,34 +310,36 @@ public override int GetHashCode()
 
         public static bool operator <(ClientWebSocketChannel left, ClientWebSocketChannel right)
         {
-            return ReferenceEquals(left, null) ? !ReferenceEquals(right, null) : left.CompareTo(right) < 0;
+            return left is null ? right is object : left.CompareTo(right) < 0;
```
vinagesh, line 313: The logic used for > looks easier to follow.
return left is object && left.CompareTo(right) < 0;

provisioning/transport/mqtt/src/ProvisioningChannelHandlerAdapter.cs
abhipsaMisra: sure, this was a VS recommended change, but I can that it isn't really relevant here as we are already specifying the type

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
abhipsaMisra: I don't understand your question here; are you asking why we set the port or why we have this comparison?

vinagesh: I think my question is not valid. Not able to see the full context of the code in the PR confused me. FallbackType is not another local variable. So nothing to change here.

PR #1344, opened 2020-04-30
 Update JobQuotaExceededException to include the associated http error reason

abhipsaMisra:
These tests have been failing consistently in our nightly, with the job quota having been exceeded. Marking them to run serially.

### DIFF
--- a/iothub/service/src/Common/Exceptions/JobQuotaExceededException.cs
+++ b/iothub/service/src/Common/Exceptions/JobQuotaExceededException.cs
@@ -1,17 +1,33 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
-    using System;
-
+    /// <summary>
+    /// Exception thrown when IoT Hub exceeds the available quota for active jobs
+    /// </summary>
     [Serializable]
     public sealed class JobQuotaExceededException : IotHubException
     {
-        public JobQuotaExceededException()
-            : base("Job quota has been exceeded")
-        {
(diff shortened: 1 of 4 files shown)


iothub/service/src/Common/Exceptions/JobQuotaExceededException.cs
```diff
@@ -1,17 +1,33 @@
...
 {
-    using System;
-
+    /// <summary>
+    /// Exception thrown when IoT Hub exceeds the available quota for active jobs
+    /// </summary>
     [Serializable]
     public sealed class JobQuotaExceededException : IotHubException
     {
+        private const string JOB_QUOTA_EXCEEDED_MESSAGE = "Job quota has been exceeded";
```
drwill-ms, line 14: lol why screaming snake case?

iothub/service/src/Common/Exceptions/JobQuotaExceededException.cs
```diff
@@ -1,17 +1,33 @@
...
         public JobQuotaExceededException()
-            : base("Job quota has been exceeded")
+            : base(JOB_QUOTA_EXCEEDED_MESSAGE)
         {
+        }
 
+        /// <summary>
+        /// Initializes an instance of JobQuotaExceededException with the message returned in the Http response
+        /// </summary>
+        /// <param name="message"></param>
```
drwill-ms, line 27: fill in

iothub/service/src/Common/Exceptions/JobQuotaExceededException.cs
abhipsaMisra: oops, have been looking at java code a bit too much; will change!

abhipsaMisra:
Keeping this change on hold for now; service is currently swallowing the exception message details (which contains the currently active job IDs). Once service passes that information to the client, we can update the SDK to forward the information to the user.

PR #1345, opened 2020-05-01
Twin array support

timtay-microsoft:
This PR brings twin array support to the iothub device and service client. Note that this requires an increase of the service API version.

As an unintended consequence of that API version increase, the SDK is now expected by the service to send a subscribe request to receive desired properties after opening the twin links. This PR modifies the SDK to do this on behalf of the users since they are used to not needing to do this. As a result, users shouldn't see any change in behavior.

This PR will not pass the gate until twin array support is broadly available. Right now, it is behind a feature flag

### DIFF
--- a/e2e/test/TwinE2ETests.cs
+++ b/e2e/test/TwinE2ETests.cs
@@ -8,6 +8,8 @@ using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client;
 using Microsoft.Azure.Devices.Shared;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.Devices.E2ETests
 {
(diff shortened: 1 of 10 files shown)


azabbasi:
I have noticed that the only changes are to the AMQP transport layer when it comes to sending those subscribe messages. Does that not apply to other protocols?

timtay-microsoft:
The MQTT layer already sends the subscribe request. The service folks considered it a bug on their side that AMQP didn't require it, so they made this change

iothub/service/src/ClientApiVersionHelper.cs
```diff
@@ -12,7 +12,7 @@ internal class ClientApiVersionHelper
     {
         private const string ApiVersionQueryPrefix = "api-version=";
         private const string ApiVersionGA = "2016-02-03";
-        private const string ApiVersionDefault = "2019-03-30";
```
azabbasi, line 15: This is a different API version we were targeting from what is specified in ClientApiVersionHelper on the device side ... how does that happen?

azabbasi: does that mean the device and the service clients don't necessarily target the same API versions? (even though they will after this PR)

iothub/device/src/Transport/AmqpIoT/AmqpIoTSendingLink.cs
```diff
@@ -194,6 +194,23 @@ internal async Task<AmqpIoTOutcome> SendTwinPatchMessageAsync(string correlation
...
 
+        internal async Task<AmqpIoTOutcome> SubscribeToDesiredPropertiesAsync(string correlationId, TimeSpan timeout)
+        {
+            if (Logging.IsEnabled) Logging.Enter(this, $"{nameof(SubscribeToDesiredPropertiesAsync)}");
+
+            AmqpMessage amqpMessage = AmqpMessage.Create();
+            amqpMessage.Properties.CorrelationId = correlationId;
+            amqpMessage.MessageAnnotations.Map["operation"] = "PUT";
+            amqpMessage.MessageAnnotations.Map["resource"] = "/notifications/twin/properties/desired";
+            amqpMessage.MessageAnnotations.Map["version"] = null;
```
prmathur-microsoft, line 205: Don't you need this for reported?

e2e/test/TwinE2ETests.cs
```diff
@@ -234,10 +284,21 @@ private async Task Twin_DeviceSetsReportedPropertyAndGetsItBackSingleDeviceAsync
...
+            await Twin_DeviceSetsReportedPropertyAndGetsItBackAsync(deviceClient, JArray.Parse("[1, 2, 3]")).ConfigureAwait(false);
+        }
+
+        public static async Task Twin_DeviceSetsReportedPropertyAndGetsItBackAsync(DeviceClient deviceClient, object propValue = null)
         {
             var propName = Guid.NewGuid().ToString();
-            var propValue = Guid.NewGuid().ToString();
+            if (propValue == null)
+            {
+                propValue = Guid.NewGuid().ToString();
```
prmathur-microsoft, line 300: Why is this necessary? Can't you ignore if there is no propvalue?

e2e/test/TwinE2ETests.cs
```diff
@@ -246,10 +307,11 @@ public static async Task Twin_DeviceSetsReportedPropertyAndGetsItBackAsync(Devic
...
 
             Twin deviceTwin = await deviceClient.GetTwinAsync().ConfigureAwait(false);
-            Assert.AreEqual<string>(deviceTwin.Properties.Reported[propName].ToString(), propValue);
+            var actual = deviceTwin.Properties.Reported[propName];
+            Assert.AreEqual(JsonConvert.SerializeObject(actual), JsonConvert.SerializeObject(propValue));
         }
 
-        public static async Task<Task> SetTwinPropertyUpdateCallbackHandlerAsync(DeviceClient deviceClient, string expectedPropName, string expectedPropValue)
+        public static async Task<Task> SetTwinPropertyUpdateCallbackHandlerAsync(DeviceClient deviceClient, string expectedPropName, object expectedPropValue)
         {
```
prmathur-microsoft, line 315: Does this mean if I used a string earlier to receive output "expectedPropValue", would it be broken now?

iothub/device/src/Common/Api/ClientApiVersionHelper.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.Azure.Devices.Client
     internal class ClientApiVersionHelper
     {
         internal const string ApiVersionQueryPrefix = "api-version=";
-        internal const string ApiVersionLatest = "2018-06-30";
+        internal const string ApiVersionLatest = "2019-10-01";
```
prmathur-microsoft, line 9: Would this only effect device client? or is this expected to impact both device and service SDK's

iothub/service/src/ClientApiVersionHelper.cs
timtay-microsoft: It means that, at some point, the service SDK needed to bump the API version to get a new feature, but the device side didn't care about that feature, or the feature didn't impact the device side

e2e/test/TwinE2ETests.cs
timtay-microsoft: The tests that call this method provide the value that should be tested (a twin array, for example), or they provide nothing if they are fine with the GUID

iothub/device/src/Transport/AmqpIoT/AmqpIoTSendingLink.cs
timtay-microsoft: This is the subscribe to desired properties request, what do you mean?

e2e/test/TwinE2ETests.cs
timtay-microsoft: No. The SDK always returned a map of <string, object>. It was just the tests that assumed the values would always be strings

iothub/device/src/Common/Api/ClientApiVersionHelper.cs
timtay-microsoft: The service client can send desired property updates that contain string arrays now, too

timtay-microsoft: I added some tests for sending arrays in desired property updates, for reference

iothub/device/src/Transport/AmqpIoT/AmqpIoTSendingLink.cs
prmathur-microsoft: I was wondering if you need to subscribe to reported prop as well ? Just like MQTT.

timtay-microsoft: The device side only sends reported properties, so I don't understand why we would

e2e/test/TwinE2ETests.cs
prmathur-microsoft: So would SDK work even without GUID and propvalue = null

timtay-microsoft: I'm not sure I follow. The SDK may or may not allow null values in the value field, but this PR has nothing to do with that

timtay-microsoft: This is just a default property value to use for a test of reported/desired properties, and this PR doesn't change anything about that. It just allows for the test to specify a different value to use

abhipsaMisra: Previously the tests were setting guid strings for both property name and property value, in this test block, and that was being used for updating twin properties. 
What you are trying to achieve now is, for new tests you are setting the `propValue` to be an array (which you are passing into the test method), while existing tests do not have `propValue` passed in, and hence will be using the above block to set a guid string.
The way this test block is structured is potentially causing the confusion. You should pass in `propValue` into the test method for string case as well, to avoid this confusion.

e2e/test/TwinE2ETests.cs
```diff
@@ -329,10 +391,13 @@ public static async Task RegistryManagerUpdateDesiredPropertyAsync(string device
             await registryManager.CloseAsync().ConfigureAwait(false);
         }
 
-        private async Task Twin_ServiceSetsDesiredPropertyAndDeviceReceivesEventAsync(Client.TransportType transport, Func<DeviceClient, string, string, Task<Task>> setTwinPropertyUpdateCallbackAsync)
+        private async Task Twin_ServiceSetsDesiredPropertyAndDeviceReceivesEventAsync(Client.TransportType transport, Func<DeviceClient, string, object, Task<Task>> setTwinPropertyUpdateCallbackAsync, object propValue = null)
         {
             var propName = Guid.NewGuid().ToString();
-            var propValue = Guid.NewGuid().ToString();
+            if (propValue == null)
```
abhipsaMisra, line 397: here as well

abhipsaMisra:
Before this service version bump, was the desired property update subscription implicit with OpenAsync(), or with DesiredPropertyUpdateCallback()?

abhipsaMisra:
> Before this service version bump, was the desired property update subscription implicit with OpenAsync(), or with DesiredPropertyUpdateCallback()? If it was with open async, then we would need to move the subscription to the open call, else it would qualify as a _breaking change_.

Along these lines, is this the corresponding subscription from mqtt: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs#L631?
If not, what does this topic filter do?

e2e/test/TwinE2ETests.cs
timtay-microsoft: I can remove the default value from this method parameter if that clears it up, sure

timtay-microsoft:
> Before this service version bump, was the desired property update subscription implicit with OpenAsync(), or with DesiredPropertyUpdateCallback()? If it was with open async, then we would need to move the subscription to the open call, else it would qualify as a _breaking change_.

Before this service version bump, the desired property update subscription was only implicit to opening the twin receiver link. OpenAsync did not open the twin receiver link, so this is safe.

timtay-microsoft:
> > Before this service version bump, was the desired property update subscription implicit with OpenAsync(), or with DesiredPropertyUpdateCallback()? If it was with open async, then we would need to move the subscription to the open call, else it would qualify as a _breaking change_.
> 
> Along these lines, is this the corresponding subscription from mqtt: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs#L631?
> If not, what does this topic filter do? `"$iothub/twin/res/#"`

It corresponds to this MQTT topic, actually:

```
private const string TwinPatchTopicFilter = "$iothub/twin/PATCH/properties/desired/#";
```

abhipsaMisra:
> It corresponds to this MQTT topic, actually:
> 
> ```
> private const string TwinPatchTopicFilter = "$iothub/twin/PATCH/properties/desired/#";
> ```

Yeah, I thought so too, but what does `$iothub/twin/res/#` do?

timtay-microsoft:
Good question, @abhipsaMisra , I actually don't see any reference to it being used in Java, and the C# references are pretty vague. I can see that it is implicitly subscribed to during open for mqtt. I can keep digging though

timtay-microsoft:
https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support

"A device must first subscribe to the $iothub/twin/res/# topic to receive the operation's responses from IoT Hub."

My understanding is that the C# SDK uses roundtrip messages, so this doesn't seem applicable to AMQP

abhipsaMisra:
> https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support
> 
> "A device must first subscribe to the $iothub/twin/res/# topic to receive the operation's responses from IoT Hub."
> 
> My understanding is that the C# SDK uses roundtrip messages in AMQP, so this doesn't seem applicable to AMQP. I.E. sending a subscribe request for desired properties, we do get the operation response already in AMQP

Ah, I see; it's for the GetTwin() call over MQTT, cool!

iothub/device/src/Transport/AmqpIoT/AmqpIoTSendingLink.cs
prmathur-microsoft: Offline : Since reported properties are published and get twin has a different sub, we don't need this for reported properties.

Issue #1185, opened 2020-01-13
[documentation] Request for connection and reliability documentation


kiranpradeep:
This is a documentation request and so does not adhere to the issue template. Is it possible to add  documentation equivalent to azure-iot-sdk-c-reliability document [[1](https://github.com/Azure/azure-iot-sdk-c/blob/master/doc/connection_and_messaging_reliability.md)]. This help in better understanding of the sdk. Example queries since other language sdk (c, java etc) provides callbacks on message delivery/failure, but this sdk relies on exceptions

1. I like to know if the retry option set is applicable only to connection retry or is it equally applicable to messgae retries? For e.g. for MQTT, I like to know when does `DeviceClient::SendEventAsync` throw an exception (#91) for a half closed TCP connection ? 

2. I like to when on what condition does a `DeviceClient::SentEventAsync` return successfully for each of HTTP, MQTT and AMQP protocols. 
> For e.g. for MQTT does a return from SentEventAsync mean 
> 1. message is confirmed delivered to IoTHub (mqtt puback)
> 2. message is written to an sdk internal queue 
> 3. message is written to wire (os flush) ?

abhipsaMisra:
Thanks for bringing this up @kiranpradeep ; a document like this will surely be helpful in understanding the SDK better. I am working on adding this in.

abhipsaMisra:
To provide an update here - we've been improving the API comments throughout the library iteratively, and have a device sample [here](https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/master/iot-hub/Samples/device/DeviceReconnectionSample) that demonstrates the recommended usage pattern for our client. We've also listed the expected client status scenarios that you might encounter and how to handle them.

While this isn't completely exhaustive at the moment, we are working on adding and updating this document as we encounter areas that require further explanation.
Also, our doc team is also working on adding public documents that specify the different behavior per client protocol used.

For the question regarding `DeviceClient::SentEventAsync`, the completion of the async API represents that the data has been successfully written to the wire.

vinagesh:
Documentation added - https://github.com/Azure/azure-iot-sdk-csharp/blob/master/DeviceConnectionAndReliabilityReadme.md

PR #1346, opened 2020-05-04
fix(iot-device): Fix out-of-date comments on DeviceClient

### LINKED ISSUE #515: Twin.DeviceId and Twin.ModuleId return null
- **OS, version, SKU and CPU architecture used:** <VERSION> Windows 10 Desktop x64
- **Application's .NET Target Framework :** <NETVERSION> netcoreapp2.1
- **Device:** <DEVICETYPE> Surface Pro (i5)
- **SDK version used:** <VERSION> Microsoft.Azure.Devices.Client 1.17.0-preview-007

## Description of the issue:
When calling `Twin.DeviceId` or `Twin.ModuleId`, null is returned.  
other properties (e.g. `Twin.Properties`) are properly populated

```C#
Twin myTwin = await ioTHubModuleClient.GetTwinAsync();
Debug.WriteLine($"Module ID:{myTwin.ModuleId}");
Debug.WriteLine($"Device ID: {myTwin.DeviceId}");
```

## Console log of the issue:
Module ID:
Device ID:

abhipsaMisra:
The only device client creation method where Http and Amqp are the only supported protocols is device client creation using iot-hub scope connection string. (FYI for #1185 )

Also, update comment for GetTwinAsync on Device Client.
Fixes #515

### DIFF
--- a/iothub/device/src/ClientFactory.cs
+++ b/iothub/device/src/ClientFactory.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Azure.Devices.Client
         /// </summary>
         /// <param name="hostname">The fully-qualified DNS hostname of IoT Hub</param>
         /// <param name="authenticationMethod">The authentication method that is used</param>
-        /// <param name="transportType">The transportType used (Http1 or Amqp)</param>
+        /// <param name="transportType">The transportType used (Http1, Amqp or Mqtt), <see cref="TransportType"/></param>
         /// <returns>InternalClient</returns>
         public static InternalClient Create(string hostname, IAuthenticationMethod authenticationMethod, TransportType transportType)
         {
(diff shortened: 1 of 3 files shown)


iothub/device/src/ClientFactory.cs
```diff
@@ -47,7 +47,7 @@ public static InternalClient Create(string hostname, string gatewayHostname, IAu
         /// </summary>
         /// <param name="hostname">The fully-qualified DNS hostname of IoT Hub</param>
         /// <param name="authenticationMethod">The authentication method that is used</param>
-        /// <param name="transportType">The transportType used (Http1 or Amqp)</param>
+        /// <param name="transportType">The transportType used (Http1, Amqp or Mqtt)</param>
```
prmathur-microsoft, line 50: Could we use a url to point to the definition of transport type? I believe there are more right? https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/TransportType.cs

abhipsaMisra: Amqp and Mqtt cover both the Tcp and WS variants. By defining the transport type as Amqp, the device client will create a prioritized transport list of Amqp_tcp, Amqp_ws (same for Mqtt).

abhipsaMisra: I can add a link to this file, sure

Issue #815, opened 2019-02-22
API Feedback: Remove empty DeviceStreamRequest type from device SDK


wickste:
- **OS, version, SKU and CPU architecture used:** <VERSION> 
- **Application's .NET Target Framework :** <NETVERSION> 
- **Device:** <DEVICETYPE> 
- **SDK version used:** <VERSION> 

## Description of the issue:

**API ReviewFeedback:** Remove empty DeviceStreamRequest type from device SDK

timtay-microsoft:
@wickste I see the DeviceStreamRequest type in the device SDK being used in several APIs. Can you expand a bit on what you wanted here?

timtay-microsoft:
Closing this issue as there is no empty DeviceStreamRequest type in this repo

az-iot-builder-01:
@wickste, @timtay-microsoft, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

Issue #814, opened 2019-02-22
API Feedback: DeviceStreamRequest.Url should be Uri


wickste:
- **OS, version, SKU and CPU architecture used:** <VERSION> 
- **Application's .NET Target Framework :** <NETVERSION> 
- **Device:** <DEVICETYPE> 
- **SDK version used:** <VERSION> 

## Description of the issue:

**API Review Feedback:** DeviceStreamRequest.Url should be renamed to Uri

timtay-microsoft:
Closing this issue as this has now been checked in

az-iot-builder-01:
@wickste, @timtay-microsoft, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

Issue #811, opened 2019-02-22
API Feedback: ServiceClient.CreateStreamAsync() should take stream name argument directly


wickste:
- **OS, version, SKU and CPU architecture used:** <VERSION> 
- **Application's .NET Target Framework :** <NETVERSION> 
- **Device:** <DEVICETYPE> 
- **SDK version used:** <VERSION> 

## Description of the issue:

**API Review Feedback:** ServiceClient.CreateStreamAsync() should take the stream name as argument directly. No need to ask the developer to first create a DeviceStreamRequest instance. If in the future more parameters are needed to create the request, we can add an overload that takes a structure to pass those in.

timtay-microsoft:
Closing this issue as we plan on re-working the device streaming API surface completely later

abhipsaMisra:
Re-opening and closing to work around dashboard issue.

abhipsaMisra:
Closing the issue.

PR #1347, opened 2020-05-05
Device Streaming API fixes

timtay-microsoft:
@wickste opened several github issues a while back about the device stream APIs, and how he would improve them. I took a couple of those suggestions for this PR:

See #814 and #812 

There were a couple of other issues he opened, one needs expanding on, and the other is a bit more controversial.

See #815 for the input needed issue
See #811 for the more controversial one that I have not implemented yet.

Both the changes I made are technically breaking changes, but considering these are preview features, it should be fine

### DIFF
--- a/e2e/test/DeviceStreamingE2ETests.cs
+++ b/e2e/test/DeviceStreamingE2ETests.cs
@@ -35,7 +35,8 @@ namespace Microsoft.Azure.Devices.E2ETests
             _listener = TestConfig.StartEventListener();
         }
 
-#region Device Client Tests
+        #region Device Client Tests
+
         [TestMethod]
         public async Task DeviceStreaming_RequestAccepted_Sas_Amqp()
         {
@@ -189,9 +190,11 @@ namespace Microsoft.Azure.Devices.E2ETests
                 }
             }
         }
+
         #endregion Device Client Tests
 
         #region Module Client Tests
+
         [TestMethod]
         public async Task ModuleStreaming_RequestAccepted_Sas_Amqp()
         {
@@ -267,9 +270,11 @@ namespace Microsoft.Azure.Devices.E2ETests
 
             await TestModuleStreamingAsync(TestDeviceType.Sasl, transportSettings, false).ConfigureAwait(false);
         }
-#endregion Module Client Tests
 
-#region Private Methods
+        #endregion Module Client Tests
+
+        #region Private Methods
+
         private async Task TestDeviceStreamingAsync(TestDeviceType type, ITransportSettings[] transportSettings, bool acceptRequest)
         {
             TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
@@ -289,7 +294,7 @@ namespace Microsoft.Azure.Devices.E2ETests
 
                 Assert.IsNotNull(clientRequest, "Received an unexpected null device streaming request");
 
-                _log.WriteLine("Device streaming request received (name=" + clientRequest.Name + "; url=" + clientRequest.Url + "; authToken=" + clientRequest.AuthorizationToken + ")");
+                _log.WriteLine("Device streaming request received (name=" + clientRequest.Name + "; uri=" + clientRequest.Uri + "; authToken=" + clientRequest.AuthorizationToken + ")");
 
                 if (acceptRequest)
                 {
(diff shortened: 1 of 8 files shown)


e2e/test/DeviceStreamingE2ETests.cs
```diff
@@ -289,7 +294,7 @@ private async Task TestDeviceStreamingAsync(TestDeviceType type, ITransportSetti
 
                 Assert.IsNotNull(clientRequest, "Received an unexpected null device streaming request");
 
-                _log.WriteLine("Device streaming request received (name=" + clientRequest.Name + "; url=" + clientRequest.Url + "; authToken=" + clientRequest.AuthorizationToken + ")");
+                _log.WriteLine("Device streaming request received (name=" + clientRequest.Name + "; uri=" + clientRequest.Uri + "; authToken=" + clientRequest.AuthorizationToken + ")");
```
drwill-ms, line 297: If you are up for it, I'd change these to string interpolation to make them easier to read.

iothub/device/src/DeviceStreamRequest.cs
```diff
@@ -19,13 +19,13 @@ public sealed class DeviceStreamRequest
         /// </summary>
         /// <param name="requestId">ID of this request.</param>
         /// <param name="name">Name of the stream</param>
-        /// <param name="url">Url to the Device Streaming gateway</param>
+        /// <param name="uri">Uri to the Device Streaming gateway</param>
```
drwill-ms, line 22: I wouldn't capitalize Device Streaming.

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -324,21 +313,9 @@ public override Task<DeviceStreamResponse> CreateStreamAsync(string deviceId, De
         /// <param name="deviceId">Device Id</param>
         /// <param name="moduleId">Module ID</param>
         /// <param name="deviceStreamRequest">Configuration needed for initiating a cloud-to-device stream.</param>
-        /// <returns>The result of the cloud-to-device stream request, or null of the request itself could not be completed.</returns>
-        public override Task<DeviceStreamResponse> CreateStreamAsync(string deviceId, string moduleId, DeviceStreamRequest deviceStreamRequest)
-        {
```
prmathur-microsoft, line 329: Why wouldn't module-id be needed anymore?

prmathur-microsoft: and the API without cancellation token?

iothub/service/src/DeviceStreamResponse.cs
```diff
@@ -42,7 +42,7 @@ public DeviceStreamResponse(string streamName, bool isAccepted, string authoriza
             StreamName = streamName;
             IsAccepted = isAccepted;
             AuthorizationToken = authorizationToken;
-            Url = streamingGatewayUri;
+            Uri = streamingGatewayUri;
         }
```
prmathur-microsoft, line 46: So are these just naming changes?

iothub/service/src/ServiceClient.cs
```diff
@@ -197,42 +197,24 @@ public static ServiceClient CreateFromConnectionString(string connectionString,
...
         /// <param name="cancellationToken">Token used for controlling the termination of the asynchronous call.</param>
         /// <returns>The result of the cloud-to-device stream request, or null of the request itself could not be completed.</returns>
-        public abstract Task<DeviceStreamResponse> CreateStreamAsync(string deviceId, string moduleId, DeviceStreamRequest deviceStreamRequest, CancellationToken cancellationToken);
+        public abstract Task<DeviceStreamResponse> CreateStreamAsync(string deviceId, DeviceStreamRequest deviceStreamRequest, CancellationToken cancellationToken = default);
 
         /// <summary>
         /// Initiates a new cloud-to-device stream.
         /// </summary>
         /// <param name="deviceId">Device Id</param>
+        /// <param name="moduleId"></param>
```
drwill-ms, line 214: fill in

iothub/service/src/AmqpServiceClient.cs
drwill-ms: A method with module param still exists.

One of the bugs was to eliminate overloads without cancellation tokens. The ones with it just have a default parameter for it, which is how we've been handling this elsewhere (like ADT).

iothub/service/src/DeviceStreamResponse.cs
timtay-microsoft: URL becomes URI, which is a breaking change. But it is more accurate

drwill-ms: Did you read the bugs?

iothub/service/src/AmqpServiceClient.cs
timtay-microsoft: It still is. This used to be 4 APIs, where 2 needed cancellation tokens, and the other 2 didn't. Git diff just makes it confusing to see that

iothub/service/src/DeviceStreamResponse.cs
timtay-microsoft: To expand on this, there is a getter and setter for this URI, so customer's could see this break their code. I stand by it being worthwile though as the field is a URI, not a URL

e2e/test/DeviceStreamingE2ETests.cs
timtay-microsoft: It's a bit out of scope for this PR

PR #1351, opened 2020-05-08
Fix C2D subscription over Mqtt to set QoS of 1

### LINKED ISSUE #206: System properties like sequence number is not populated in received C2D message using mqtt protocol
- **OS and version used:**  Windows 10
- **SDK version used:** 1.5.2 

# Description of the issue:
In received C2D message using mqtt protocol, system properties like sequence number is not populated

# Code sample exhibiting the issue:
Device:
```
var myDeviceClient = DeviceClient.CreateFromConnectionString(connectiongString, TransportType.Mqtt);
myDeviceClient.OpenAsync().Wait();
while (true)
{
         var receivedMessage = await MyDeviceClient.ReceiveAsync(TimeSpan.FromSeconds(1));
         if (receivedMessage == null) continue;

         Console.WriteLine($"{DateTime.Now} < Received message: {Encoding.UTF8.GetString(receivedMessage.GetBytes())} with SN {receivedMessage.SequenceNumber}");

          await MyDeviceClient.CompleteAsync(receivedMessage);
          await Task.Delay(100);
}

```
# Output:
image

If using other protocols, the sequence number shows correctly

abhipsaMisra:
- The Mqtt transport handler was setting C2D subscription packet with QoS of 0 (AtMostOnce), while the [documentation](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#receiving-cloud-to-device-messages) states that it should be QoS of 1 (AtLeastOnce). This was causing all mqtt packets received over devicebound topic to have a QoS of 0 and packet ID of 0.
- Since there is no API to return the message sequence no over Mqtt, this property is not settable over Mqtt

Fixes #206

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -665,7 +665,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         {
             if (TryStateTransition(TransportState.Open, TransportState.Subscribing))
             {
-                await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(_deviceboundMessageFilter, QualityOfService.AtMostOnce))).ConfigureAwait(true);
+                await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(_deviceboundMessageFilter, QualityOfService.AtLeastOnce))).ConfigureAwait(true);
 
                 if (TryStateTransition(TransportState.Subscribing, TransportState.Receiving))
                 {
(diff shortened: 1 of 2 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -665,7 +665,7 @@ private async Task SubscribeCloudToDeviceMessagesAsync()
         {
             if (TryStateTransition(TransportState.Open, TransportState.Subscribing))
             {
-                await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(_deviceboundMessageFilter, QualityOfService.AtMostOnce))).ConfigureAwait(true);
+                await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(_deviceboundMessageFilter, QualityOfService.AtLeastOnce))).ConfigureAwait(true);
```
prmathur-microsoft, line 669: What would be the impact on the QOS of messages to SDK users?

abhipsaMisra: We were initially subscribing with QoS of 0, so all C2D messages were sent from service with QoS of 0 -> at most once -> fire-and-forget. We now subscribe with QoS of at-least-once, so this will make c2d message delivery more reliable over Mqtt.

PR #1353, opened 2020-05-08
Add support for ns2.1

abhipsaMisra:
This PR brings in the following changes:

1. Update pipeline script to run unit tests for CI and nightly builds.
2. Add support for .netstandard2.1
3. Add an additional target framework for tests (.netcoreapp2.1.18).  This is the latest stable version of .netcore2.1, which is in [LTS until 2021-08-21](https://dotnet.microsoft.com/download/dotnet-core)
Now the .netcoreapp3.1 test project references the ns2.1 dll, while the .netstandard2.1.18 test project references the ns2.0 dll. Other test project references remain the same as before.
4. Addition of ns2.1 support required the `String.Replace(string, string)` calls to be updated to `Replace(String, String, StringComparison)`. However, the overload with `StringComparion` is not available in <ns2.1 implementations, so I had to add #if directives.

### DIFF
--- a/e2e/test/E2ETests.csproj
+++ b/e2e/test/E2ETests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net472;net451</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;netcoreapp2.1.18;net472;net451</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp3.1;netcoreapp2.1.18</TargetFrameworks>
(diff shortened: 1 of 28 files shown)


e2e/test/E2ETests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net472;net451</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;netcoreapp2.1.18;net472;net451</TargetFrameworks>
```
drwill-ms, line 3: Why is 3.1 insufficient to test?

abhipsaMisra: Due to nuget framework precedence rules, the netcoreapp3.1 test project will target the ns2.1 dll, while the netcoreapp2.1.18 test project will target the ns2.0 dll

abhipsaMisra:
> Anything to update in the readme?

Ah yes, good point! I'll update it.

PR #1333, opened 2020-04-15
Set remote certificate validation callback for WebSocket connections

jussi-raunio-abb:
## Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/.github/CONTRIBUTING.md).
- [ ] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
  - There were no unit tests for the methods that I changed in the first place.
- [x] This pull-request is submitted against the `master` branch.

## Description of the changes
- Remove private key checks from `ClientFactory`.
- Set remote certificate validation callback for WebSocket connections.
  - Since .NET core does not use global certificate validation callbacks, each instance has its own remote certificate validation callback. Unfortunately, the `WebSocketOptions` in .NET standard does not exposed a remote certificate callback property. When a library is used from .NET core, there is no way to set WebSocket's remote certificate callback. This PR sets the remote certificate callback through reflection. 
- Add remote certificate callback property to `SecurityProviderX509` to be used by the provisioning service.

## Reference/Link to the issue solved with this PR (if any)
We are using .NET core in Linux with custom OpenSSL client certificate engine. Since .NET does not expose OpenSSL interface, we have configured OpenSSL engine to be autoloaded. So, when a TLS connection is established, our custom OpenSSL engine completes the handshake by providing a private key for the certificate.

SDK does not need to know anything about OpenSSL or our custom engine. However, SDK needs to allow certificate validation to happen outside the application. Basically this means that:
- SDK should allow certificates without private key.
- SDK should allow setting a custom certificate validation callback for provision and device connections.

PR #1363, opened 2020-05-15
Add remote certificate validation callback for websocket connections

abhipsaMisra:
This PR brings in the following changes:
- Implement remote certificate validation callback over websocket connections for IoT Hub
- Implement remote certificate validation callback for tcp and websocket connections for DPS
- Remove x509 certificate private key check for device client - the service requires the public key to be sent for authentication, the private key is required for TLS negotiation, and can be a part of the `X509Certificate2` object sent to service, or installed on the system directly.

### DIFF
--- a/iothub/device/src/ClientFactory.cs
+++ b/iothub/device/src/ClientFactory.cs
@@ -83,11 +83,6 @@ namespace Microsoft.Azure.Devices.Client
                     throw new ArgumentException("certificate must be present in DeviceAuthenticationWithX509Certificate");
                 }
 
-                if (!connectionStringBuilder.Certificate.HasPrivateKey)
-                {
-                    throw new ArgumentException("certificate in DeviceAuthenticationWithX509Certificate must have a private key");
-                }
-
                 InternalClient dc = CreateFromConnectionString(connectionStringBuilder.ToString(), PopulateCertificateInTransportSettings(connectionStringBuilder, transportType));
                 dc.Certificate = connectionStringBuilder.Certificate;
                 return dc;
@@ -136,11 +131,6 @@ namespace Microsoft.Azure.Devices.Client
                 if (connectionStringBuilder.Certificate == null)
                 {
                     throw new ArgumentException("certificate must be present in DeviceAuthenticationWithX509Certificate");
-                }
-
-                if (!connectionStringBuilder.Certificate.HasPrivateKey)
-                {
-                    throw new ArgumentException("certificate in DeviceAuthenticationWithX509Certificate must have a private key");
                 }
 
                 InternalClient dc = CreateFromConnectionString(connectionStringBuilder.ToString(), PopulateCertificateInTransportSettings(connectionStringBuilder, transportSettings));
--- a/iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
+++ b/iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
@@ -176,6 +176,10 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
                 if (_amqpTransportSettings.RemoteCertificateValidationCallback != null)
                 {
                     websocket.Options.RemoteCertificateValidationCallback = _amqpTransportSettings.RemoteCertificateValidationCallback;
+                    if (Logging.IsEnabled)
+                    {
+                        Logging.Info(this, $"{nameof(CreateClientWebSocketAsync)} Setting RemoteCertificateValidationCallback");
+                    }
                 }
 #endif
 
--- a/iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
+++ b/iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
@@ -171,6 +171,14 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
                     websocket.Options.ClientCertificates.Add(_amqpTransportSettings.ClientCertificate);
                 }
 
+                // Support for RemoteCertificateValidationCallback for ClientWebSocket is introduced in .NET Standard 2.1
+#if NETSTANDARD2_1
+                if (_amqpTransportSettings.RemoteCertificateValidationCallback != null)
+                {
+                    websocket.Options.RemoteCertificateValidationCallback = _amqpTransportSettings.RemoteCertificateValidationCallback;
+                }
+#endif
+
                 using (var cancellationTokenSource = new CancellationTokenSource(timeout))
                 {
                     await websocket.ConnectAsync(websocketUri, cancellationTokenSource.Token).ConfigureAwait(false);
--- a/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
@@ -126,7 +126,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         public TimeSpan DefaultReceiveTimeout { get; set; }
 
         /// <summary>
-        /// A callback for remote certificate validation
+        /// A callback for remote certificate validation.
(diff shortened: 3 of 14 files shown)


iothub/device/src/ClientFactory.cs
```diff
@@ -83,11 +83,6 @@ public static InternalClient Create(string hostname, string gatewayHostname, IAu
                     throw new ArgumentException("certificate must be present in DeviceAuthenticationWithX509Certificate");
                 }
 
-                if (!connectionStringBuilder.Certificate.HasPrivateKey)
```
abhipsaMisra, line 86: The availability of the client certificate's private key will be validated when server presents the crypto challenge to the client.

timtay-microsoft: Should these if blocks be preserved when the user isn't on Netstandard2.1? From what I can tell, the remote callback only gets executed for Netstandard2.1

abhipsaMisra: This check is independent of the remote certificate validation callback. This is the client certificate that is presented to the remote server. The service doesn't actually require the private key, but instead relies on the client solving the challenge, in order to establish the mutual TLS connection.

iothub/device/src/ClientFactory.cs
```diff
@@ -138,11 +133,6 @@ public static InternalClient Create(string hostname, string gatewayHostname, IAu
...
-                {
```
prmathur-microsoft, line 142: Can we try to check why did we have this in the first place till now ? I agree private key shouldn't be presented. Also how did it work with this check until now?

iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
```diff
@@ -171,6 +171,18 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
                     websocket.Options.ClientCertificates.Add(_amqpTransportSettings.ClientCertificate);
                 }
 
+                // Support for RemoteCertificateValidationCallback for ClientWebSocket is introduced in .NET Standard 2.1
+#if NETSTANDARD2_1
+                if (_amqpTransportSettings.RemoteCertificateValidationCallback != null)
+                {
```
prmathur-microsoft, line 177: And what if _amqpTransportSettings.RemoteCertificateValidationCallback == null; is it safe to ignore the value of websocket.Options.RemoteCertificateValidationCallback ?

provisioning/transport/amqp/src/AmqpAuthStrategySymmetricKey.cs
```diff
@@ -41,14 +42,13 @@ public override AmqpSettings CreateAmqpSettings(string idScope)
             return settings;
         }
 
-        public override Task OpenConnectionAsync(AmqpClientConnection connection, TimeSpan timeout, bool useWebSocket, IWebProxy proxy)
+        public override Task OpenConnectionAsync(AmqpClientConnection connection, TimeSpan timeout, bool useWebSocket, IWebProxy proxy, RemoteCertificateValidationCallback remoteCertificateValidationCallback)
         {
```
prmathur-microsoft, line 46: Is this for trusted certs here? why would you need this in symmetric key code path ?

iothub/device/src/Transport/AmqpIoT/AmqpIoTTransport.cs
abhipsaMisra: If the application doesn't supply a validation callback then we don't set it

provisioning/transport/amqp/src/AmqpAuthStrategySymmetricKey.cs
abhipsaMisra: The remote certificate validation callback is for validating the remote certificate, which is the certificate presented by the iot hub, in our case. This is the certificate that the server would preset for TLS negotiation. It does not depend on the specific authentication mechanism used for validating the device identity.

iothub/device/src/ClientFactory.cs
abhipsaMisra: IoT hub doesn't care about the private key; all they require is the thumbprint for self-signed certificates, and the certificate chain for CA authenticated devices. If the device presents a certificate whose private key it doesn't own, then the authentication will fail when completing the TLS handshake.

abhipsaMisra: Also, here we weren't actually requiring the private key as a string/ some exportable format; we simply validated that the certificate being passed in has the private key available. This is usually true for most windows based systems; but other OS might be having different constraints.

abhipsaMisra:
From @jussi-raunio-abb #1333 :

We are using .NET core in Linux with custom OpenSSL client certificate engine. Since .NET does not expose OpenSSL interface, we have configured OpenSSL engine to be autoloaded. So, when a TLS connection is established, our custom OpenSSL engine completes the handshake by providing a private key for the certificate.

SDK does not need to know anything about OpenSSL or our custom engine. However, SDK needs to allow certificate validation to happen outside the application. Basically this means that:

- SDK should allow certificates without private key.
- SDK should allow setting a custom certificate validation callback for provision and device connections.

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -126,7 +126,8 @@ public MqttTransportSettings(TransportType transportType)
         public TimeSpan DefaultReceiveTimeout { get; set; }
 
         /// <summary>
-        /// A callback for remote certificate validation
+        /// A callback for remote certificate validation.
+        /// If incorrectly implemented, your device may fail to connect to IoTHub and/or be open to security vulrenabilities.
```
prmathur-microsoft, line 130: 👍

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -306,6 +310,18 @@ private static DeviceRegistrationResult ConvertToProvisioningRegistrationResult(
                 websocket.Options.ClientCertificates.Add(clientCertificate);
             }
 
+            // Support for RemoteCertificateValidationCallback for ClientWebSocket is introduced in .NET Standard 2.1
+#if NETSTANDARD2_1
+            if (RemoteCertificateValidationCallback != null)
```
bikamani, line 315: Just wondering, this is supported for netstandard only then?

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
drwill-ms, line 130: vulnerabilities

provisioning/device/src/ProvisioningTransportHandler.cs
```diff
@@ -23,6 +24,12 @@ public abstract class ProvisioningTransportHandler : IDisposable
         /// </summary>
         public IWebProxy Proxy { get; set; }
 
+        /// <summary>
+        /// A callback for remote certificate validation.
+        /// If incorrectly implemented, your device may fail to connect to DPS and/or be open to security vulrenabilities.
```
drwill-ms, line 29: misspelled here too

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
abhipsaMisra: Yes, the support for REmoteCertificateValidationCallback was added for websockets in .NET Standard 2.1: https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocketoptions.remotecertificatevalidationcallback?view=netcore-3.1

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
abhipsaMisra: my bad, fixing the typo.

PR #1375, opened 2020-05-29
refactor(iot-service): Bump service API version now that 2020-03-13 is available in all regions

timtay-microsoft:
This API version was only available in certain regions previously, but now that it is available in every region, we can make all service client operations use this

We can also get rid of the environment variable lookup that was previously used to toggle between a widely available service API version and the selectively available service api version

### DIFF
--- a/iothub/service/src/AmqpServiceClient.cs
+++ b/iothub/service/src/AmqpServiceClient.cs
@@ -17,10 +17,10 @@ namespace Microsoft.Azure.Devices
 {
     internal sealed class AmqpServiceClient : ServiceClient
     {
-        private const string StatisticsUriFormat = "/statistics/service?" + ClientApiVersionHelper.ApiVersionQueryStringDefault;
-        private const string PurgeMessageQueueFormat = "/devices/{0}/commands?" + ClientApiVersionHelper.ApiVersionQueryStringDefault;
-        private const string DeviceMethodUriFormat = "/twins/{0}/methods?" + ClientApiVersionHelper.ApiVersionQueryStringDefault;
-        private const string ModuleMethodUriFormat = "/twins/{0}/modules/{1}/methods?" + ClientApiVersionHelper.ApiVersionQueryStringDefault;
+        private const string StatisticsUriFormat = "/statistics/service?" + ClientApiVersionHelper.ApiVersionQueryString;
+        private const string PurgeMessageQueueFormat = "/devices/{0}/commands?" + ClientApiVersionHelper.ApiVersionQueryString;
+        private const string DeviceMethodUriFormat = "/twins/{0}/methods?" + ClientApiVersionHelper.ApiVersionQueryString;
+        private const string ModuleMethodUriFormat = "/twins/{0}/modules/{1}/methods?" + ClientApiVersionHelper.ApiVersionQueryString;
         private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromSeconds(100);
 
         private readonly FaultTolerantAmqpObject<SendingAmqpLink> _faultTolerantSendingLink;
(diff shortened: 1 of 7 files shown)


iothub/service/src/HttpRegistryManager.cs
```diff
@@ -1495,14 +1483,14 @@ private static Uri GetBulkRequestUri(string apiVersionQueryString)
 
         // apiVersion parameter should be removed after the "next" api-version is available in all regions
```
prmathur-microsoft, line 1484: remove this comment?

prmathur-microsoft: And address it?

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -17,10 +17,10 @@ namespace Microsoft.Azure.Devices
...
     {
-        private const string StatisticsUriFormat = "/statistics/service?" + ClientApiVersionHelper.ApiVersionQueryStringDefault;
-        private const string PurgeMessageQueueFormat = "/devices/{0}/commands?" + ClientApiVersionHelper.ApiVersionQueryStringDefault;
-        private const string DeviceMethodUriFormat = "/twins/{0}/methods?" + ClientApiVersionHelper.ApiVersionQueryStringDefault;
-        private const string ModuleMethodUriFormat = "/twins/{0}/modules/{1}/methods?" + ClientApiVersionHelper.ApiVersionQueryStringDefault;
+        private const string StatisticsUriFormat = "/statistics/service?" + ClientApiVersionHelper.ApiVersionQueryString;
+        private const string PurgeMessageQueueFormat = "/devices/{0}/commands?" + ClientApiVersionHelper.ApiVersionQueryString;
+        private const string DeviceMethodUriFormat = "/twins/{0}/methods?" + ClientApiVersionHelper.ApiVersionQueryString;
+        private const string ModuleMethodUriFormat = "/twins/{0}/modules/{1}/methods?" + ClientApiVersionHelper.ApiVersionQueryString;
         private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromSeconds(100);
```
prmathur-microsoft, line 24: I rem this had some effect on e2e as well? Isn't there anything to be cleaned up ?

timtay-microsoft: You remember correctly, there was a flag in the e2e tests around this. I've pushed changes to remove that flag now

prmathur-microsoft: 👍

iothub/service/src/ClientApiVersionHelper.cs
```diff
@@ -12,22 +12,12 @@ internal class ClientApiVersionHelper
...
-        private const string ApiVersionDefault = "2019-10-01";
-        private const string ApiVersionLimitedAvailability = "2020-03-13";
-
-        // For import/export devices jobs, a new parameter is available in a
-        // new api-version, which is only available in a few initial regions.
-        // Control access via an environment variable. If a user wishes to try it out,
-        // they can set "EnabledStorageIdentity" to "1". Otherwise, the SDK will still
-        // default to the latest, broadly-supported api-version used in this SDK.
-        internal static bool IsStorageIdentityEnabled => StringComparer.OrdinalIgnoreCase.Equals("1", Environment.GetEnvironmentVariable("EnableStorageIdentity"));
+        private const string ApiVersionDefault = "2020-03-13";
```
abhipsaMisra, line 15: Not related to this change, but why do we have a 2016 service API version supported here? Can we test if all operations can support the latest service API version?

timtay-microsoft: Good question. I'll remove it and see if the tests still pass as expected. I don't see why we should still have this reference

PR #1377, opened 2020-06-01
refactor(yaml): Make gate run in parallel fix

timtay-microsoft:
3 in parallel used to be correct, since we had 3 frameworks to test for. Now that we have 4 for windows, we may as well bump up this value to a max value so we don't have to increment it each time we add a supported framework

### DIFF
--- a/vsts/vsts.yaml
+++ b/vsts/vsts.yaml
@@ -22,7 +22,7 @@ jobs:
     displayName: Linux
     strategy:
       # Change maxParallel to 1 make builds run in serial rather than in parallel
-      maxParallel: 3
+      maxParallel: 100
       matrix:
         .Net Core 3.1:
           FRAMEWORK: netcoreapp3.1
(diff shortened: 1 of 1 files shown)


vsts/vsts.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
     displayName: Linux
     strategy:
       # Change maxParallel to 1 make builds run in serial rather than in parallel
-      maxParallel: 3
+      maxParallel: 100
```
drwill-ms, line 25: That's a big jump. Are we sure 100 is safe? I don't know much about the impact of changing this.

timtay-microsoft: This just means that if we add another framework to our parallel suite, we won't need to increment this max parallel value. It doesn't mean that we will run 100 agents in parallel right now. We'll just have one agent in parallel per supported framework + OS

PR #1379, opened 2020-06-03
feat(iot-device): Add support for PnP by passing in the DCM ID

abhipsaMisra:
## Checklist

### DIFF
--- a/iothub/device/src/ClientOptions.cs
+++ b/iothub/device/src/ClientOptions.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.Devices.Client
     public class ClientOptions
     {
         /// <summary>
-        /// The Digital Twin Model Id associated with the device client instance.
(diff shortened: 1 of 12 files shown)


iothub/device/src/ClientOptions.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Options that allow configuration of the device client instance during initialization.
+    /// </summary>
+    public class ClientOptions
+    {
+        /// <summary>
+        /// The Digital Twin Model Id associated with the device client instance.
```
drwill-ms, line 12: Digital Twins?

abhipsaMisra: sure, should it also be lower case?

drwill-ms: yes

PR #1380, opened 2020-06-03
refactor(iot-device): Fix documentation comments

abhipsaMisra:


iothub/device/src/ClientOptions.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.Devices.Client
     public class ClientOptions
     {
         /// <summary>
-        /// The Digital Twin Model Id associated with the device client instance.
+        /// The Digital Twins model Id associated with the device client instance.
```
drwill-ms, line 12: you can lower case digital twins too, per your other inquiry

vinagesh, line 12: do you need a plural here - twins?

abhipsaMisra: ok

abhipsaMisra: yes, that is the term that we are using when referring to this feature/ service.

PR #1368, opened 2020-05-15
Fix error mapping for error code iot-hub-not-found-error

abhipsaMisra:
AMQP - iot-hub-not-found error will now close the connection and reconnect.
The current behavior was to reconnect at session level itself, but for failover/failback scenarios we require the DNS resolution to happen again.

Fixes #1312

PR #1382, opened 2020-06-04
fix(iot-device): Revert iot-hub-not-found will close the connection and reconnect

abhipsaMisra:
This PR reverts the change made in #1368 .
This is because, based on the discussion with the service team today, they have confirmed that service will send the error code `:iot-hub-not-found-error` over amqp, for the scenario where the hub does not exist. In this case, any retry is useless and undesirable. 
Since the hub failover scenario is has a different behavior right now wrt the different protocols, any workaround that we add in the SDK has the potential to introduce more issues.

The service team is analyzing their code to come up with:
- a short term fix - the service team will work to have consistency in Error Codes/Disconnection Pattern/Link Closure across AMQP and MQTT for devices in hub which is failing over. They will provide us with instructions on how the application should attempt recovery for hub failover scenarios.
- long term fix - the service will provide different error codes to explicitly differentiate between “iot hub does not exist” vs “iot hub does not exist right now because it is failing over”.

### DIFF
--- a/iothub/device/src/Common/Exceptions/IotHubSuspendedException.cs
+++ b/iothub/device/src/Common/Exceptions/IotHubSuspendedException.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.Devices.Client.Exceptions
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="DeviceMaximumQueueDepthExceededException"/> class with the message string set to the message parameter and a reference to the inner exception that is the cause of this exception.
+        /// Initializes a new instance of the <see cref="IotHubSuspendedException"/> class with the message string set to the message parameter and a reference to the inner exception that is the cause of this exception.
         /// </summary>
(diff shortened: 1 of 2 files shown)


iothub/device/src/Common/Exceptions/IotHubSuspendedException.cs
```diff
@@ -40,7 +40,7 @@ public IotHubSuspendedException(string iotHubName, string trackingId)
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="DeviceMaximumQueueDepthExceededException"/> class with the message string set to the message parameter and a reference to the inner exception that is the cause of this exception.
+        /// Initializes a new instance of the <see cref="IotHubSuspendedException"/> class with the message string set to the message parameter and a reference to the inner exception that is the cause of this exception.
```
abhipsaMisra, line 43: This is an unrelated change, the documentation comment was incorrect.

PR #1385, opened 2020-06-05
feat(vsts): Add additional test frameworks to release pipeline

abhipsaMisra:


vsts/test-release-nuget.yaml
```diff
@@ -8,10 +8,12 @@ jobs:
     displayName: Linux
     strategy:
       # Change maxParallel to 1 make builds run in serial rather than in parallel
-      maxParallel: 3
+      maxParallel: 100
```
vinagesh, line 11: Do we need a limit at all? The schema says - If maxParallel is unspecified or set to 0, no limit is applied.

abhipsaMisra: good point, will make this change.

Issue #1037, opened 2019-08-23
Unable to upload file with HTTP proxy


motias:
- **OS, version, SKU and CPU architecture used:** <VERSION> Windows 10 Desktop x64
- **Application's .NET Target Framework :** <NETVERSION> netcoreapp2.2
- **Device:** <DEVICETYPE> Laptop (Simulated Device)
- **SDK version used:** <VERSION> Microsoft.Azure.Devices.Client 1.21.0

## Description of the issue:
It is not possible to upload file with proxy by use of `DeviceClient.UploadToBlobAsync`.
I found a similar question in Stack Overflow, but it has not beed resolved.
[Azure IoT Hub, File Upload through proxy?](https://stackoverflow.com/questions/54776386/azure-iot-hub-file-upload-through-proxy)

## Code sample exhibiting the issue:
```cs
var proxy = new WebProxy("http://proxy.example.com:3128");
var deviceClient = DeviceClient.CreateFromConnectionString("ConnectionStringForIoTHub", new ITransportSettings[]
{
    new AmqpTransportSettings(TransportType.Amqp_WebSocket_Only)
    {
        Proxy = proxy
    }
});

// DeviceTwin update works correctly
await deviceClient.GetTwinAsync().ConfigureAwait(false);
await deviceClient.UpdateReportedPropertiesAsync(new TwinCollection { ["status"] = "connected" }).ConfigureAwait(false); 

// File upload does not work correctly
var filePath = "test.log";
var blobName = "test.log";
using (var stream = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
{
    await deviceClient.UploadToBlobAsync(blobName, stream).ConfigureAwait(false); 
}
```

motias:
@bikamani Sorry, no errors and logs.
In my network environment, there are two network route. One is proxy-enabled, and the other is not enabled. I tried to upload file using the former route, but the latter is used (and file upload is succeeded).

It seems to me that the following implementation is incomplete. 
The code is very simple and `CloudBlockBlob` instance is not configured to use a proxy.

https://github.com/Azure/azure-iot-sdk-csharp/blob/a0824e20e1f87dae193e3c951f61772fd29796ce/iothub/device/src/Transport/HttpTransportHandler.cs#L162

I think it is necessary to be implemented as follows:

```cs
var client = new CloudBlobClient(new Uri(fileUploadResponse.HostName), 
[...]

PR #1388, opened 2020-06-05
Bump version for the release 06-05-2020

prmathur-microsoft:
Release Notes draft : 

## Microsoft.Azure.Devices.Shared v1.20.1

- Add Support for net standard 2.1

## Microsoft.Azure.Devices.Client v1.26.0

- Add Client Options to enable users to supply model id 
- Add Support for net standard 2.1
- Add a way for nested Edge to pass authentication chain to IoTHub
- Add twin array support and update to latest service API version
- Updated reference to Microsoft.Azure.Devices.Shared nuget.

#Bug fixes

- Enable RemoteCertificateValidationCallback for Websocket connections to allow users to validate remote certificates
- Fix Mqtt C2D subscription message to set QoS of 1
- Fix comments on twin/methods and create
- Fix issue to require explicit subscribe to desired properties in latest Service API version 
- Fix for each subscribe request declare its topic name on mqtt messages
- Remove DNS lookup in mqtt stack when proxy is present
- Don't dispose a message being returned
- Dispose CancellationTokenSource being used internally in the library

## Microsoft.Azure.Devices v1.21.0

- Use latest Service API version that is available in all regions
- Add Support for net standard 2.1
- Add twin array support
- Updated reference to Microsoft.Azure.Devices.Shared nuget.

#Bug fixes

- Update JobQuotaExceededException to include the associated http error reason
- Code clean up in IotHubConnectionStringBuilder based on coding standards

## Microsoft.Azure.Devices.Provisioning.Service v1.6.3
- Add Support for net standard 2.1

## Microsoft.Azure.Devices.Provisioning.Client v1.6.0

- Add Support for net standard 2.1
- Add RemoteCertificateValidationCallback to allow users to validate remote certs

#Bugs fixes
- Dispose client web socket instance in mqtt transport handler

## Microsoft.Azure.Devices.Provisioning.Transport.Amqp v1.3.0
- Add Support for net standard 2.1
- Add RemoteCertificateValidationCallback to allow users to validate remote certs

## Microsoft.Azure.Devices.Provisioning.Transport.Mqtt v1.3.0
- Add Support for net standard 2.1
- Add RemoteCertificateValidationCallback to allow users to validate remote certs

## Microsoft.Azure.Devices.Provisioning.Transport.Http v1.2.3
- Add Support for net standard 2.1

## Microsoft.Azure.Devices.Provisioning.Security.Tpm v1.2.3
- Add Support for net standard 2.1


timtay-microsoft:
Can you post the before and after look at the versions, so we can tell how much these versions are bumped (patch vs minor bump)?

bikamani:
> * code clean up in IotHubConnectionStringBuilder

Is this a bug fix, can we add more description then?

bikamani:
> * Don't dispose a message being returned
> * Dispose CancellationTokenSource

Some more description here would help too, other notes look fine.

abhipsaMisra:
@timtay-microsoft 
I have some questions about this change: Add Proxy support for File upload
We did not have proxy support for file upload since it wasn't supported for azure storage. I looked at the commit and we are doing 2 things that I don't understand:
- we are supplying the proxy info to establish the client -> iothub connection; but the same isn't available for client -> storage connection - to me this seems incorrect
- we have some sort of logic that is looking for proxy settings in amqp and mqtt transport settings, while file upload is only over http - this also is incorrect to me.

Can you explain the reasonings behind these changes.

provisioning/transport/http/src/Microsoft.Azure.Devices.Provisioning.Transport.Http.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.2.2</Version>
+    <Version>1.2.3</Version>
```
bikamani, line 23: why is this patch update? Should it not be minor?

timtay-microsoft:
> @timtay-microsoft
> I have some questions about this change: Add Proxy support for File upload
> We did not have proxy support for file upload since it wasn't supported for azure storage. I looked at the commit and we are doing 2 things that I don't understand:
> 
> * we are supplying the proxy info to establish the client -> iothub connection; but the same isn't available for client -> storage connection - to me this seems incorrect
> * we have some sort of logic that is looking for proxy settings in amqp and mqtt transport settings, while file upload is only over http - this also is incorrect to me.
> 
> Can you explain the reasonings behind these changes.

That's a good catch about file upload proxy support. You are right, that I only added proxy support for the step where we get the blob uri from iothub. We can skip claiming proxy support for file upload until we get that step figured out, too

As for why it looks at proxy settings in amqp or mqtt: If a client is using amqp to send telemetry through a proxy, and they also want to do file upload through a proxy, this was the only way to use the same proxy for both scenarios

abhipsaMisra:
> That's a good catch about file upload proxy support. You are right, that I only added proxy support for the step where we get the blob uri from iothub. We can skip claiming proxy support for file upload until we get that step figured out, too

Yes, we should not claim support for file upload, until we can actually honor those settings.

> As for why it looks at proxy settings in amqp or mqtt: If a client is using amqp to send telemetry through a proxy, and they also want to do file upload through a proxy, this was the only way to use the same proxy for both scenarios

I do not like this approach, it's a weird thing to say that you can give me your proxy details over amqp settings, but I will use them over http instead. I would prefer that we call out explicitly that file upload requires http connection, so if you would like to have any explicit configurations for your http client, then you would need to initialize the device client over http. Otherwise we would use an http client with default settings for file upload operation.

timtay-microsoft:
@abhipsaMisra leave your notes in this issues thread so that we can follow up on this discussion: https://github.com/Azure/azure-iot-sdk-csharp/issues/1037

abhipsaMisra:
> Release Notes draft :
> 
> ## Microsoft.Azure.Devices.Shared v1.21.0
> * Add Support for net standard 2.1
> 
> ## Microsoft.Azure.Devices.Client v1.26.0
> * Add Client Options to enable users to supply model id
> * **Add Proxy support for File upload**
***This needs to be removed
> * Add Support for net standard 2.1
> * **Add a way for nested Edge to pass authentication chain to IoTHub**
***Why do we want to publish this info? It is an internal property that has no meaning for end users.
> * Add twin array support and update to latest service API version
> * Updated reference to Microsoft.Azure.Devices.Shared nuget.
> 
> #Bug fixes
> 
> * **Enable RemoteCertificateValidationCallback to allow users to validate _remote_ certs**
*** I would add this under features and call out that we added support for this validation callback for websocket connections, and that it is available only in ns2.1.
> * Fix Mqtt C2D subscription message to set QoS of 1
> * Fix comments on twin/methods and create
> * Fix issue to require explicit subscribe to desired properties in latest Service API version
> * Fix for each subscribe request declare its topic name on mqtt messages
> * Remove DNS lookup in mqtt stack when proxy is present
> * Don't dispose a message being returned
> * **Dispose CancellationTokenSource**
*** Dispose CancellationTokenSource being used internally in the library.
> 
> ## Microsoft.Azure.Devices v1.21.0
> * Use latest Service API version that is available in all regions
> * Add Support for net standard 2.1
> * Add twin array support
> * Updated reference to Microsoft.Azure.Devices.Shared nuget.
> 
> #Bugs
> 
> * Update JobQuotaExceededException to include the associated http error reason
> * **code clean up in IotHubConnectionStringBuilder**
*** What kind of cleanup?
> 
> ## Microsoft.Azure.Devices.Provisioning.Service v1.7.0
> * Add Support for net standard 2.1
> 
> ## Microsoft.Azure.Devices.Provisioning.Client v1.6.0
*** * Add Support for net standard 2.1
> * Add RemoteCertificateValidationCallback to allow users to validate certs
> 
> #Bugs
> 
> * Dispose client web socket instance in mqtt transport handler
> 
> ## Microsoft.Azure.Devices.Provisioning.Transport.Amqp v1.3.0
> * Add Support for net standard 2.1
> * Add RemoteCertificateValidationCallback to allow users to validate certs
> 
> ## Microsoft.Azure.Devices.Provisioning.Transport.Mqtt v1.3.0
> * Add Support for net standard 2.1
> * Add RemoteCertificateValidationCallback to allow users to validate certs
> 
> ## Microsoft.Azure.Devices.Provisioning.Transport.Http **_v1.2.2_**
> * Add Support for net standard 2.1
*** Minor version bump for added framework support.
> 
> ## Microsoft.Azure.Devices.Provisioning.Security.Tpm v1.3.0
> * Add Support for net standard 2.1

provisioning/transport/http/src/Microsoft.Azure.Devices.Provisioning.Transport.Http.csproj
prmathur-microsoft: It is patch because the fixes for making it 2.1 compatible were minor and mostly bug fixes.

abhipsaMisra:
FYI - the semver guidance based on dependency changes: https://github.com/semver/semver/issues/148

For addition of a new target framework, we rely on the availability of a new API for websockets:
- For IoT hub device client packages - we didn't change our public API surface; so those should have patch version increased.
- For provisioning device client packages - we added a public API, so it will have minor version increased.
- For all service client packages - no public API change, patch version increase.

provisioning/transport/http/src/Microsoft.Azure.Devices.Provisioning.Transport.Http.csproj
bikamani: But I see other clients updating to minor version for making this compatibility change?

prmathur-microsoft: Based on changes to ns2.1 in the files, if it is small fix I updated it to reflect patch. If those were big changes then I updated it to reflect minor.

PR #1397, opened 2020-06-09
Expose a way for IoTEdge to pass DeviceCapabilityModelId to IoTHub

dylanbronson:
Currently Edge does not have any support for Plug and Play. This change adds the necessary Id, the Device Capability Model Id (referred to as "digital-twin-model-id"). Edge can set the property via reflection and the SDK will send the ID along to IoTHub so that the device is successfully registered as a PNP device with its DCMID. 

This PR is very similar to https://github.com/Azure/azure-iot-sdk-csharp/commit/ae85bab9d8e62a990c4238da3cf7d3105921a315

In that change, we added a way for IoTEdge to change the authentication chain via reflection to pass to IoTHub. 

This change does the exact same thing, but instead with the DeviceCapabilityModelId that is needed for Plug and Play devices. 

Note: The name of the DeviceCapabilityModelId might be moving towards just ModelId or CapabilityModelId. 

## Checklist

### DIFF
--- a/iothub/device/src/AmqpTransportSettings.cs
+++ b/iothub/device/src/AmqpTransportSettings.cs
@@ -22,6 +22,11 @@ namespace Microsoft.Azure.Devices.Client
         /// Used by Edge runtime to specify an authentication chain for Edge-to-Edge connections
         /// </summary>
         internal string AuthenticationChain { get; set; }
+
+        /// <summary>
+        /// Used by Edge runtime for Plug and Play devices
+        /// </summary>
+        internal string DeviceCapabilityModelId { get; set; }
         
         /// <summary>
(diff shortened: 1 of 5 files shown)


abhipsaMisra:
We have already exposed a way for setting the Model Id for Plug and play device. This feature is available over MQTT. 
Can IoT Edge use this: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/ClientOptions.cs#L15

Also, is there a use-case for making this available over AMQP? Afaik, this feature has been implemented by the service only over MQTT.

abhipsaMisra:
@dylanbronson This will be implemented via #1467 , closing this PR.

Issue #599, opened 2018-08-27
Wrong exception parameters for DeviceNotFoundException and DeviceAlreadyExistsException for IoT Hub service clients


CIPop:
Shared code in `\common\src\service\ExceptionHandlingHelper.cs` calls into several exception types which have `deviceId` as first parameter, not `message`:
- [ ] DeviceNotFoundException
- [ ] DeviceAlreadyExistsException

This results in confusing messages. For example the following is thrown when a Method is called for a disconnected device (spaces added for clarity):

```
Microsoft.Azure.Devices.Common.Exceptions.DeviceNotFoundException: 
Device 

{"Message":"{\"errorCode\":404103,\"trackingId\":\"cda96d6c6e65450d8ded4cea239f7401-G:4-TimeStamp:08/27/2018 07:55:43\",\
"message\":\"Timed out waiting for device to connect.\",\"info\":{\"timeout\":\"00:00:00\"},\"timestampUtc\":\"2018-08-27T07:55:43.0356528Z\"}","ExceptionMessage":""} 

not registered
```

RyanLilla:
@CIPop I am interested in taking on this issue if it is still available! As I am new to exception handling in this context,  is there somewhere/someone I could refer to for assistance?

I see that for the method definition where `message` is a parameter, there is a second parameter of type `Exception innerException` that would need to be used within ExceptionHandlingHelper.cs, correct? Or are we saying that `deviceId` shouldn't be the single parameter at all within these methods, but should be `message` instead?

CIPop:
@RyanLilla go for it! Really appreciate the contributions!

mprigge:
This issue bit me, so I'd be interested to fix it. Seems like the easiest way to fix it without breaking a bunch of things that are using the `public DeviceNotFoundException(string deviceId)` constructor would be to add a `(Exception)null` parameter to the calls in `\common\src\service\ExceptionHandlingHelper.cs` so that they hit the `public DeviceNotFoundException(string message, Exception innerException)` constructor instead. Other portions of the codebase already use that workaround and so that's easy enough to implement. I need to read the contribution guidelines, but I'll try to do that soon.

[...]

PR #1402, opened 2020-06-12
fix(iothub-client): Updated code to set the correct exception message

vinagesh:
## Checklist

## Description of the changes
Updated the code to set the correct exception message.

https://github.com/Azure/azure-iot-sdk-csharp/issues/599

### DIFF
--- a/common/src/service/ExceptionHandlingHelper.cs
+++ b/common/src/service/ExceptionHandlingHelper.cs
@@ -17,24 +17,23 @@ namespace Microsoft.Azure.Devices
         {
             var mappings = new Dictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>();
 
-            mappings.Add(HttpStatusCode.NoContent, async (response) => new DeviceNotFoundException(await GetExceptionMessageAsync(response).ConfigureAwait(false)));
-            mappings.Add(HttpStatusCode.NotFound, async (response) => new DeviceNotFoundException(await GetExceptionMessageAsync(response).ConfigureAwait(false)));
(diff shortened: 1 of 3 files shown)


common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -17,24 +17,23 @@ class ExceptionHandlingHelper
         {
             var mappings = new Dictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>();
```
vinagesh, line 19: Previously, this resulted in setting wrong error messages for DeviceNotFoundException as the message was being passed in as a deviceId. I have updated all methods to explicitly use the correct method and parameters so that if we add any overloads, they accidentally don't end up matching something incorrect.

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -51,7 +51,7 @@ public AmqpServiceClient(IotHubConnectionString iotHubConnectionString, bool use
                 transportSettings.HttpProxy);
         }
 
-        internal AmqpServiceClient(IotHubConnectionString iotHubConnectionString, bool useWebSocketOnly, IHttpClientHelper httpClientHelper) : base()
+        internal AmqpServiceClient(IHttpClientHelper httpClientHelper) : base()
```
vinagesh, line 54: This is internal and those parameters are not used so seems safe to remove.

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -11,30 +11,29 @@ namespace Microsoft.Azure.Devices
 
     using Microsoft.Azure.Devices.Common.Exceptions;
 
-    class ExceptionHandlingHelper
+    internal class ExceptionHandlingHelper
```
vinagesh, line 14: explicitly calling out as internal. Should have no impact as it was already internal.

Issue #1399, opened 2020-06-10
UploadToBlobAsync does not modify Stream.Position for progress tracking.


OsaPL:
It makes progress report basically imposible.

(Here I'm using Win10 x64 machine, would have to test it on a Linux also)
Lets do a file upload. 
```c#
using (var data = new FileStream(path, FileMode.Open))
{
     var watch = System.Diagnostics.Stopwatch.StartNew();
     ProgressReport(blobPath, data, watch, cancellationTokenSource.Token);
     await deviceClient.UploadToBlobAsync(blobPath, data, cancellationTokenSource.Token);
     watch.Stop();
     _logger.LogInformation("Time to upload file: {0}ms", watch.ElapsedMilliseconds);
}
```

Then in a seperate Task:
```c#
private async Task ProgressReport(string file,Stream data, System.Diagnostics.Stopwatch watch,CancellationToken token)
{
    var length = data.Length;
    while (!token.IsCancellationRequested)
    {
        await Task.Delay(5000);
        _logger.LogInformation("Progress: {0} {1}% {2}s", file, data.Position/data.Length, watch.Elapsed.TotalSeconds);
    }
}
```

You will get something like this:
```
[13:58:31 Information ]
 Progress: Test.mp3 0% 35.0627038s
[13:58:36 Information ]
 Progress: Test.mp3 0% 40.1025024s
[13:58:41 Information ]
 Progress: Test.mp3 0% 45.116194s
[13:58:46 Information ]
 Progress: Test.mp3 0% 50.129142099999996s
[13:58:51 Information ]
 Progress: Test.mp3 0% 55.1335366s
[13:58:56 Information ]
 Progress: Test.mp3 0% 60.1507109s
[13:59:01 Information ]
 Progress: Test.mp3 0% 65.1606468s
[13:59:06 Information ]
 Time to upload file: 69286ms
```

Would love to get a way to actually track progress, either by using ``Stream.Position`` or something exposed by the ``DeviceClient``.

timtay-microsoft:
Just a heads up, we are planning on deprecating that uplaod to blob API in favor of what I detailed [here](https://github.com/Azure/azure-iot-sdk-csharp/issues/1037#issuecomment-641521842).

It's to solve a different issue, but this would allow you to use the Azure Storage SDK directly, and they may have a better way for you to track each upload

OsaPL:
@timtay-microsoft 
[...]

PR #1403, opened 2020-06-12
feat(iot-dev): Deprecate existing file upload APIs, add new ones at granular level

### LINKED ISSUE #1037: Unable to upload file with HTTP proxy

timtay-microsoft:
Customers should be free to use whatever version of the Azure Storage SDK that they want, and to configure it however they want. Because of that, the existing file upload API has been split into the three service calls: Get SAS uri from Iot Hub, upload to Azure Storage blob using iothub provided credentials, and then notify Iot hub that a file upload has completed.

The JSON object types for these new APIs already existed, but their names were wrong and they were internal, so I renamed them and added documentation accordingly.

Once this is checked in, I can add a sample to the samples repo. After that, I can reference that sample in the documentation for these new APIs.

In doing this, I also made sure that we only create one http client for the lifetime of the client for all http operations. I also added proxy support for just the 

I couldn't design our APIs in such a way that the user provides the HTTP client, so I settled on taking an HttpTransportSettings object instead.

Fixes #1037 and probably #1399


vinagesh:
Do you want to add a link to the bug that will get fixed in the description?

iothub/device/src/DeviceClient.cs
```diff
@@ -447,9 +449,28 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
         public Task UploadToBlobAsync(string blobName, Stream source, CancellationToken cancellationToken) =>
             InternalClient.UploadToBlobAsync(blobName, source, cancellationToken);
 
+        /// <summary>
+        /// Get a file upload SAS URI which the Azure Storage SDK can use to upload a file to blob for this device. See <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-file-upload#initialize-a-file-upload">this documentation for more details</see>
+        /// </summary>
+        /// <param name="request">The request details for getting the SAS URI, including the destination blob name.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The file upload details to be used with the Azure Storage SDK in order to upload a file from this device.</returns>
+        public async Task<FileUploadSasUriResponse> GetFileUploadSasUri(FileUploadSasUriRequest request, CancellationToken cancellationToken = default) =>
```
drwill-ms, line 462: you can probably remove `async` from this, and the `await` and `.ConfigureAwait(false)` from the impl.

e2e/test/FileUploadE2ETests.cs
```diff
@@ -54,6 +59,97 @@ public async Task FileUpload_X509_SmallFile_Http()
...
+            {
+                Proxy = new WebProxy(Configuration.IoTHub.ProxyServerAddress)
+            };
+
+            await UploadFileManual(fileStreamSource, filename, fileUploadTransportSettings).ConfigureAwait(false);
+        }
+
+        private async Task UploadFileManual(Stream source, string filename, Http1TransportSettings fileUploadTransportSettings, bool x509auth = false)
+        {
+            await FileNotificationTestListener.InitAsync().ConfigureAwait(false);
```
vinagesh, line 100: This is just a suggestion - I think the tests will read better if you separate the auth from the upload. The auth part seems to belong more in the test as you are testing upload with different auth.

iothub/device/src/InternalClient.cs
```diff
@@ -149,6 +151,23 @@ public InternalClient(IotHubConnectionString iotHubConnectionString, ITransportS
             if (Logging.IsEnabled) Logging.Associate(this, transportSettings, nameof(InternalClient));
             _transportSettings = transportSettings;
 
+            if (options != null)
+            {
+                var fileUploadTransportSettings = options.FileUploadTransportSettings;
```
drwill-ms, line 156: Even though you provide a default impl, they could still set it to null, which would give you an NRE here

iothub/device/src/DeviceClient.cs
```diff
@@ -436,6 +437,7 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
         /// <param name="blobName">The name of the blob to upload.</param>
         /// <param name="source">A stream with blob contents. Should be disposed after upload completes.</param>
         /// <returns>AsncTask</returns>
+        [Obsolete("This API has been split into three APIs: GetFileUploadSasUri, uploading to blob directly using the Azure Storage SDK, and CompleteFileUploadAsync")]
```
vinagesh, line 440: woohoo

iothub/device/src/Transport/FileUploadNotificationResponse.cs
```diff
@@ -1,22 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information
-
-using Newtonsoft.Json;
-
-namespace Microsoft.Azure.Devices.Client.Transport
-{
-    internal class FileUploadNotificationResponse
```
drwill-ms, line 8: Were these unreferenced classes?

timtay-microsoft: These are the same Json objects that were used for this flow originally. They were just internal, and I changed their names to fit what they each did

timtay-microsoft: Even though github didn't interpret it this way, this class was just renamed to FileUploadSasUriResponse

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: Like only add tests for 
1) SAS based granular
2) SAS based granular with Prody

?

iothub/device/src/DeviceClient.cs
```diff
@@ -447,9 +449,28 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
         public Task UploadToBlobAsync(string blobName, Stream source, CancellationToken cancellationToken) =>
             InternalClient.UploadToBlobAsync(blobName, source, cancellationToken);
 
+        /// <summary>
+        /// Get a file upload SAS URI which the Azure Storage SDK can use to upload a file to blob for this device. See <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-file-upload#initialize-a-file-upload">this documentation for more details</see>
+        /// </summary>
+        /// <param name="request">The request details for getting the SAS URI, including the destination blob name.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The file upload details to be used with the Azure Storage SDK in order to upload a file from this device.</returns>
```
vinagesh, line 461: Returns file upload details or SAS Uri?

vinagesh: Should we also mention how long this will be valid?

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: I wanted to have an x509 test case here simply because the transportsettings of the file upload client are a bit different depending on if you use SAS or x509, so it seemed worth testing

iothub/device/src/InternalClient.cs
```diff
@@ -149,6 +151,23 @@ public InternalClient(IotHubConnectionString iotHubConnectionString, ITransportS
...
+                if (fileUploadTransportSettings.ClientCertificate == null)
+                {
+                    //Kind of strange to take file upload transport settings from the non-file upload transport settings,
```
vinagesh, line 159: nit: space before Kind

iothub/device/src/DeviceClient.cs
timtay-microsoft: The SAS URI is something that is constructed from the return type here. See the e2e test for details. It looks like this:

timtay-microsoft:
```
            string blobUriString = String.Format(
                CultureInfo.InvariantCulture,
                "https://{0}/{1}/{2}{3}",
                fileUploadSasUriResponse.HostName,
                fileUploadSasUriResponse.ContainerName,
                Uri.EscapeDataString(fileUploadSasUriResponse.BlobName), // Pass URL encoded device name and blob name to support special characters
                fileUploadSasUriResponse.SasToken);
```

timtay-microsoft: This is ideally where I would link the sample that shows this off, but I can't write the sample until this code is checked in since the samples live in a different repo that references this repo

timtay-microsoft: Actually, I may as well just add a helper function to do this formatting. Hang tight and I'll push changes for that

e2e/test/FileUploadE2ETests.cs
vinagesh: No I meant add the auth logic inside the test and the common method should just have the common logic. The Upload helper method seems to do a lot.

iothub/device/src/ClientOptions.cs
```diff
@@ -13,5 +13,13 @@ public class ClientOptions
...
         /// </summary>
         public string ModelId { get; set; }
+
+        /// <summary>
+        /// The transport settings to use for all file upload operations, regardless of what protocol the device
+        /// client is configured with. All file upload operations take place over https. If no client certificates
+        /// are configured to this object, then file upload operations will use the client certificates configured
+        /// in the transport settings set for the non-file upload operations.
+        /// </summary>
+        public Http1TransportSettings FileUploadTransportSettings { get; set; } = new Http1TransportSettings();
```
prmathur-microsoft, line 23: Can I use these settings if I need to use HTTP on device client?

iothub/device/src/DeviceClient.cs
prmathur-microsoft: While it is awesome to celebrate! We should let @sharmasejal take a call on letting customers know of this change and her inputs on this.

iothub/device/src/ClientOptions.cs
timtay-microsoft: No. They are strictly for file upload operations, hence the name

timtay-microsoft: You can provide the same instance of HTTP transport settings for the base device client operations and for this file upload transport settings option, though

iothub/device/src/DeviceClient.cs
prmathur-microsoft, line 440: Also please mention last SDK version where this API works in comments so that current users of this API know when/what to upgrade to.

iothub/device/src/ClientOptions.cs
prmathur-microsoft: That would confuse the customer. If I end up providing Http1TransportSettings  then I would ideally want it to work on messaging as well if I use HTTP.

iothub/device/src/ClientOptions.cs
```diff
@@ -13,5 +13,13 @@ public class ClientOptions
         /// This feature is currently supported only over MQTT.
         /// </summary>
         public string ModelId { get; set; }
+
+        /// <summary>
+        /// The transport settings to use for all file upload operations, regardless of what protocol the device
+        /// client is configured with. All file upload operations take place over https. If no client certificates
+        /// are configured to this object, then file upload operations will use the client certificates configured
```
abhipsaMisra, line 20: Why not make it mandatory to provide client certificate for file upload in `FileUploadTransportSettings`?
Also, what was the restriction in accepting an HttpClient directly?

iothub/device/src/DeviceClient.cs
```diff
@@ -447,9 +449,28 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+        public Task<FileUploadSasUriResponse> GetFileUploadSasUri(FileUploadSasUriRequest request, CancellationToken cancellationToken = default) =>
+            InternalClient.GetFileUploadSasUriAsync(request, cancellationToken);
+
+        /// <summary>
+        /// Notify IoT Hub that a device's file upload has finished. See <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-file-upload#notify-iot-hub-of-a-completed-file-upload">this documentation for more details</see>
+        /// </summary>
+        /// <param name="notification">The notification details, including if the file upload succeeded.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The task to await.</returns>
+        public Task CompleteFileUploadAsync(FileUploadCompletionNotification notification, CancellationToken cancellationToken = default) =>
```
prmathur-microsoft, line 471: Can a customer reject this? (to be inline with our other API's)

iothub/device/src/Transport/FileUploadCompletionNotification.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Whether the file upload was successful or not.
+        /// </summary>
+        [JsonProperty(PropertyName = "isSuccess")]
+        public bool IsSuccess { get; set; }
+
+        /// <summary>
+        /// The status code for the file upload.
+        /// </summary>
```
prmathur-microsoft, line 30: User defined Status Code?

iothub/device/src/InternalClient.cs
```diff
@@ -149,6 +151,24 @@ public InternalClient(IotHubConnectionString iotHubConnectionString, ITransportS
             if (Logging.IsEnabled) Logging.Associate(this, transportSettings, nameof(InternalClient));
             _transportSettings = transportSettings;
 
+            if (options != null && options.FileUploadTransportSettings != null)
+            {
+                var fileUploadTransportSettings = options.FileUploadTransportSettings;
+                if (fileUploadTransportSettings.ClientCertificate == null)
+                {
+                    // Kind of strange to take file upload transport settings from the non-file upload transport settings,
```
abhipsaMisra, line 159: Ah :pensive:

e2e/test/FileUploadE2ETests.cs
```diff
@@ -54,6 +59,90 @@ public async Task FileUpload_X509_SmallFile_Http()
...
+            using FileStream fileStreamSource = new FileStream(filename, FileMode.Open, FileAccess.Read);
+            Http1TransportSettings fileUploadTransportSettings = new Http1TransportSettings();
+
+            await UploadFileGranular(fileStreamSource, filename, fileUploadTransportSettings, true).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        [TestCategory("LongRunning")]
+        public async Task FileUpload_SmallFile_Http_GranularSteps_Proxy()
+        {
```
prmathur-microsoft, line 87: Even if this is a service feature, I would just check if the following behavior works and SDK did not impose any restrictions on it with new implemention.
1. 10 (get this number from what msdn says if it is incorrect) fileuploads in parallel

iothub/device/src/InternalClient.cs
timtay-microsoft: Yeah, I didn't see a way around this

iothub/device/src/Transport/FileUploadCompletionNotification.cs
timtay-microsoft: User defined, yes. I can add a comment to that effect. They will see this status code in the service client side when the service client receives the notification

e2e/test/FileUploadE2ETests.cs
```diff
@@ -54,6 +59,90 @@ public async Task FileUpload_X509_SmallFile_Http()
...
+            using FileStream fileStreamSource = new FileStream(filename, FileMode.Open, FileAccess.Read);
+            Http1TransportSettings fileUploadTransportSettings = new Http1TransportSettings();
+
+            await UploadFileGranular(fileStreamSource, filename, fileUploadTransportSettings).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        [TestCategory("LongRunning")]
+        public async Task FileUpload_SmallFile_Http_GranularSteps_x509()
+        {
```
prmathur-microsoft, line 76: I don't see a sample (update or add) to this new set of API's. Is it another PR?

iothub/device/src/DeviceClient.cs
timtay-microsoft: No, because it is not like those other APIs. The device client is the sender, not the receiver

iothub/device/src/DeviceClient.cs
timtay-microsoft: Users of older versions won't see this message, so I don't see the value

e2e/test/FileUploadE2ETests.cs
```diff
@@ -54,6 +59,90 @@ public async Task FileUpload_X509_SmallFile_Http()
...
+            var fileUploadSasUriRequest = new FileUploadSasUriRequest()
+            {
+                BlobName = filename
+            };
+
+            var fileUploadSasUriResponse = await deviceClient.GetFileUploadSasUri(fileUploadSasUriRequest).ConfigureAwait(false);
+
+            CloudBlockBlob blob = new CloudBlockBlob(fileUploadSasUriResponse.GetBlobUri());
+            var uploadTask = blob.UploadFromStreamAsync(source);
+            await uploadTask.ConfigureAwait(false);
```
prmathur-microsoft, line 133: Does it matter which storage SDK I need to use for this feature to work?

iothub/device/src/ClientOptions.cs
timtay-microsoft: Because it isn't mandatory to provide FileUploadTransportSettings at all, much less a field on them

iothub/device/src/InternalClient.cs
```diff
@@ -1070,6 +1102,7 @@ public void Dispose()
             InnerHandler?.Dispose();
             _methodsDictionarySemaphore?.Dispose();
             _receiveSemaphore?.Dispose();
+            _fileUploadHttpTransportHandler?.Dispose();
```
abhipsaMisra, line 1105: 👍

iothub/device/src/Transport/FileUploadCompletionNotification.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Copyright (c) Microsoft. All rights reserved.
```
abhipsaMisra, line 3: twice!

iothub/device/src/ClientOptions.cs
timtay-microsoft: I investigated taking the http client directly, but there is http client handler class that we could take that exists in all the frameworks that we support. Plus there are just layers and layers of HTTP code underneath all of this, and its a lot safer to just re-use that logic than it is to start dependency injecting clients into that mess

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: We shouldn't be documenting service side limitations in our SDK. They could change on a whim and without anyone telling us to update

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: The sample repo is separate for C#. I'll add a sample after this gets checked in

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: Nope! And that's the beauty of it all

iothub/device/src/Transport/FileUploadCompletionNotification.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.Client.Transport
+{
+    /// <summary>
+    /// The request payload to send to IoT Hub to notify it when a file upload is completed, whether successful or not.
+    /// </summary>
+    public class FileUploadCompletionNotification
```
abhipsaMisra, line 13: q - are these fields mandatory to be set?

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: For our e2e tests, I just used the storage SDK version that we were already using

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: This isn't exercising a new path on the service, so the restrictions are still the same

iothub/device/src/DeviceClient.cs
timtay-microsoft: The proper method name might be "NotifyHubThatFileUploadHasCompleted" but that is a little wordy

iothub/device/src/Transport/FileUploadCompletionNotification.cs
timtay-microsoft: good question, let me check

e2e/test/FileUploadE2ETests.cs
abhipsaMisra: let's use the non-deprecated sdk for our tests then :)

drwill-ms: The absolute latest one doesn't work with our tests. Remember the crash I ran into when our diagnostic logging is enabled?

I use 1 version previous in the jobs test. That one works. We can use that one.

iothub/device/src/Transport/FileUploadCompletionNotification.cs
timtay-microsoft: CorrelationId and Success are mandatory, the others are optional

iothub/device/src/ClientOptions.cs
abhipsaMisra: hmm, I don't mean to be a pain about this, but since this operation doesn't really have any of the features that we get from the delegating handlers pipeline, couldn't we forgo all of that and directly make the http call within the internal client (with the supplied http client)?

e2e/test/FileUploadE2ETests.cs
abhipsaMisra: ah yes, I remember :(

iothub/device/src/DeviceClient.cs
prmathur-microsoft: I think what I want is when I upgrade and if I happened to be using older API which will now show as deprecated, and can still compile. When I hover over it I want to know what should I do next and when was it deprecated so that I can decide to go back if I dont want to use new approach (for whatever reason) vs upgarde to latest for whatever benefits I am getting. If this is not the right place then where is it?

iothub/device/src/DeviceClient.cs
prmathur-microsoft: But as a user I can decide to not upload after receiving sas uri. How do I notify that?

e2e/test/FileUploadE2ETests.cs
prmathur-microsoft: I have not asked for documentation in the comment. My comment was to "I would just check if the following behavior works and SDK did not impose any restrictions on it with new implemention."

iothub/device/src/ClientOptions.cs
timtay-microsoft: That hasn't been the behavior in the past, so what you are suggesting would be a breaking change. It isn't an option

iothub/device/src/DeviceClient.cs
timtay-microsoft: When a method is deprecated isn't interesting to anyone, in my opinion. We don't break the deprecated API in question, so there is never a need for the customer to revert their SDK version

timtay-microsoft: As for what they should do instead, that is interesting to everyone, so I included it in the Obsolete comment

iothub/device/src/DeviceClient.cs
timtay-microsoft: You can decide to not upload, and there is nothing wrong with that. You can even notify iothub that an upload occurred even when it didn't, but that would just be user error.

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: I did check, and everything still works on the same quotas. No new restrictions are brought on from this

e2e/test/FileUploadE2ETests.cs
timtay-microsoft: Yeah, we can take that fix for another PR

PR #1404, opened 2020-06-16
fix(iot-service): Expose the exception Code to the user

### LINKED ISSUE #629: Code property on IotHubException is not populated
The property Code on IotHubException is never populated.

Like this:

```
try {
    await _serviceClient.InvokeDeviceMethodAsync(iotDeviceId, cloudToDeviceMethod);
} catch (DeviceNotFoundException ex) {
    var referenceCode = ex.GetReferenceCode(); //null
    var code = ex.Code; //ErrorCode.InvalidErrorCode
}
```
I have written the values I see when an error occur and it is always those values.

I would like to be able to get more detailed information on why the device was not found. 
If I could get access to the error so it is possible to distinguish a 404001 DeviceNotFound from a 404103 "Timed out waiting for device to connect" as an example.

Thanks
Mikkel

AB#7291056

barustum:
## Checklist

## Description of the changes
1. Expose the Code property in the IoTHubException class

Fixes #629


e2e/test/MethodE2ETests.cs
```diff
@@ -126,6 +127,31 @@ public async Task Method_ServiceSendsMethodThroughProxyWithCustomTimeout()
...
+        [TestMethod]
+        public async Task Method_ServiceInvokeDeviceMethodWithUnknownDeviceThrows()
+        {
+            // setup
+            using var serviceClient = ServiceClient.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+            var methodInvocation = new CloudToDeviceMethod("SetTelemetryInterval");
+            methodInvocation.SetPayloadJson("10");
+
+            // act
+            ErrorCode actualErrorCode = ErrorCode.InvalidErrorCode;
```
abhipsaMisra, line 139: Q - does this need to be initialized?

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
```diff
@@ -14,6 +14,11 @@ public DeviceMessageLockLostException(string message)
         {
         }
 
+        public DeviceMessageLockLostException(ErrorCode code, string message)
```
abhipsaMisra, line 17: this should have been an overload with `ErrorCode code, string message, Exception innerException = null`, correct?

iothub/service/src/Common/Exceptions/IotHubException.cs
```diff
@@ -47,11 +47,23 @@ public IotHubException(string message, Exception innerException)
         {
         }
 
+        public IotHubException(ErrorCode code, string message, Exception innerException = null)
+            : this(message, innerException, false, string.Empty)
+        {
```
abhipsaMisra, line 52: What do you think about having the ErrorCode overload on every constructor?

iothub/service/src/Common/Exceptions/QuotaExceededException.cs
```diff
@@ -14,6 +14,11 @@ public QuotaExceededException(string message)
         {
         }
 
+        public QuotaExceededException(ErrorCode code, string message)
+            : base(code, message, true)
```
abhipsaMisra, line 18: I would add named parameter for `isTransient: true`. The `true` sitting in there as is can get a bit confusing.

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -5,29 +5,53 @@ namespace Microsoft.Azure.Devices
...
-            mappings.Add(HttpStatusCode.PreconditionFailed, async (response) => new DeviceMessageLockLostException(message: await GetExceptionMessageAsync(response).ConfigureAwait(false)));
-            mappings.Add(HttpStatusCode.RequestEntityTooLarge, async (response) => new MessageTooLargeException(message: await GetExceptionMessageAsync(response).ConfigureAwait(false))); ;
-            mappings.Add(HttpStatusCode.InternalServerError, async (response) => new ServerErrorException(message: await GetExceptionMessageAsync(response).ConfigureAwait(false)));
-            mappings.Add(HttpStatusCode.ServiceUnavailable, async (response) => new ServerBusyException(message: await GetExceptionMessageAsync(response).ConfigureAwait(false)));
-            mappings.Add((HttpStatusCode)429, async (response) => new ThrottlingException(message: await GetExceptionMessageAsync(response).ConfigureAwait(false)));
+            mappings.Add(HttpStatusCode.NoContent, async (response) => new DeviceNotFoundException(await GetExceptionCodeAsync(response).ConfigureAwait(false),
```
abhipsaMisra, line 24: I would use named args here as well.

e2e/test/MethodE2ETests.cs
azabbasi: looks like an Enum, no initialization required

common/src/service/ExceptionHandlingHelper.cs
vinagesh: yupp I added those just one commit ago to fix a bug. It would be safer to keep the named args.
Also this is reverting the fix for #599 - There is no overload in DeviceNotFoundException for a code and message. It expects a deviceId and IotHubName and constructs a message, which will end up looking weird.

vinagesh: Nevermind you added an overload below.

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -36,5 +60,70 @@ public static Task<string> GetExceptionMessageAsync(HttpResponseMessage response
...
+                {
+                    if (Logging.IsEnabled)
+                    {
+                        Logging.Error(null, $"There is a mismatch between the error code retrieved from the response content and the response header." +
+                            $"Content error code: {errorCode}. Header error code description: {(int)headerErrorCode}.");
+                    }
+                }
+            }
+
+            return ErrorCode.InvalidErrorCode;
```
vinagesh, line 126: This is a lot of code to maintain in the SDK. This is great for now but we should probably let the service know to send back the right kind of response. This might be something they can fix as a non-breaking change.

iothub/service/src/Common/Exceptions/DeviceAlreadyExistsException.cs
```diff
@@ -24,6 +24,11 @@ public DeviceAlreadyExistsException(string message, Exception innerException)
         {
         }
 
+        public DeviceAlreadyExistsException(ErrorCode code, string message, Exception innerException = null)
```
vinagesh, line 27: This class is public, so wondering if you should just create a new constructor? I'm not sure if this will be considered a breaking change if someone has taken dependency on this class (though unlikely).

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
vinagesh: It would be good to keep all exception classes consistent though this class never had innerException in the parameters list.

iothub/service/src/Common/Exceptions/IotHubException.cs
vinagesh: Would that not be a breaking change as all these classes are public?

common/src/service/CommonConstants.cs
```diff
@@ -117,6 +117,9 @@ static class CommonConstants
         public const string IotHubErrorCode = "IotHubErrorCode";
         public const string HttpErrorCodeName = "iothub-errorcode";
 
+        // Custom HTTP response contents
+        public const string ErrorCode = "errorCode";
```
timtay-microsoft, line 121: Does this need to be public? How about just internal?

iothub/service/src/Common/Exceptions/IotHubException.cs
abhipsaMisra: We would be adding additional APIs, so no, it would be a public API surface change, but not breaking.

barustum: I thought about this as well. I can definitely add the ErrorCode on all the constructors. I added it for a single contructor based on who needs to consume it. For me it was how much time and effort do we want to put in for this bug fix, since track2 is going to replace this one. I am not opposed to the idea of adding ErrorCode to all constructors on all the exception types we implement.

vinagesh: @abhipsaMisra  - I misunderstood what you meant. Sure adding a new constructor overload will not be a breaking change. But I agree with @barustum that it is probably not worth spending the time on it. This is mostly consumed internally so this might be sufficient. It could be another improvement story on the board that can be prioritized accordingly.

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
barustum: I am seeing tons if inconsistencies between the exception classes. They're all over the place. I did not add innerException here because non of the methods expose it. So left it out. It should probably be there, but then I'll have to investigate all the other exception classes and make them consistent. Since the change is addressing a specific bug, and since we're working on track2, I thought I would make minimal changes to fix the issue. Let me know if you guys disagree.

common/src/service/ExceptionHandlingHelper.cs
barustum: Agreed. The integer ErrorCode and description should be coming from a single place, as opposed to two different places. Do you have any contact in IoT Hub who can help clarify this?

common/src/service/CommonConstants.cs
barustum: Good point. Changing to internal.

common/src/service/ExceptionHandlingHelper.cs
vinagesh: I don't. But we should ask Priyanka/Sejal and just send them an email.

abhipsaMisra: Agree with you guys here, we can put this workaround for now, but you should follow up with Srivani about this.

e2e/test/MethodE2ETests.cs
abhipsaMisra: not removed?

iothub/service/src/Common/Exceptions/IotHubException.cs
abhipsaMisra: that works for me

iothub/service/src/IoTHubExceptionResult.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// A class used as a model to deserialize error response object sent by IoT Hub
```
abhipsaMisra, line 9: nit: A class used as a model to deserialize error response object ~sent by IoT Hub~ received from IoT Hub

PR #1405, opened 2020-06-16
feat(iot-dev): Add client options to module client constructors

barustum:
This will allow user to set the module ID when setting up the module client

## Checklist

## Description of the changes
Change the constructors of the module client to allow passing in client options

PBI 7468678

### DIFF
--- a/iothub/device/src/ClientOptions.cs
+++ b/iothub/device/src/ClientOptions.cs
@@ -4,12 +4,12 @@
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// Options that allow configuration of the device client instance during initialization.
+    /// Options that allow configuration of the device or module client instance during initialization.
     /// </summary>
     public class ClientOptions
     {
         /// <summary>
-        /// The digital twins model Id associated with the device client instance.
+        /// The digital twins model Id associated with the device or module client instance.
         /// This feature is currently supported only over MQTT.
         /// </summary>
         public string ModelId { get; set; }
--- a/iothub/device/src/ModuleClient.cs
+++ b/iothub/device/src/ModuleClient.cs
@@ -47,10 +47,11 @@ namespace Microsoft.Azure.Devices.Client
         /// </summary>
         /// <param name="hostname">The fully-qualified DNS hostname of IoT Hub</param>
         /// <param name="authenticationMethod">The authentication method that is used</param>
+        /// <param name="options">The options that allow configuration of the module client instance during initialization.</param>
         /// <returns>ModuleClient</returns>
-        public static ModuleClient Create(string hostname, IAuthenticationMethod authenticationMethod)
+        public static ModuleClient Create(string hostname, IAuthenticationMethod authenticationMethod, ClientOptions options = default)
         {
-            return Create(() => ClientFactory.Create(hostname, authenticationMethod));
+            return Create(() => ClientFactory.Create(hostname, authenticationMethod, options));
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


iothub/device/tests/ModuleClientTests.cs
```diff
@@ -72,6 +72,22 @@ public void ModuleClient_CreateFromConnectionString_NoTransportSettings()
...
+            var clientOptions = new ClientOptions
+            {
+                ModelId = "tempModuleId"
+            };
+
+            // act
+            var moduleClient = ModuleClient.CreateFromConnectionString(FakeConnectionString, clientOptions);
+
+            // assert
+            Assert.IsNotNull(moduleClient);
```
vinagesh, line 88: Maybe also assert on the value of ModelId in ClientOption?

iothub/device/src/ClientOptions.cs
```diff
@@ -4,12 +4,12 @@
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// Options that allow configuration of the device client instance during initialization.
+    /// Options that allow configuration of the device or module client instance during initialization.
     /// </summary>
     public class ClientOptions
```
drwill-ms, line 9: We need to rename this to avoid the name conflcit with, and it needs to inherit from the Azure.Core.ClientOptions.

iothub/device/tests/ModuleClientTests.cs
```diff
@@ -72,6 +72,22 @@ public void ModuleClient_CreateFromConnectionString_NoTransportSettings()
...
+        public void ModuleClient_CreateFromConnectionString_WithClientOptions()
+        {
+            // setup
+            var clientOptions = new ClientOptions
+            {
+                ModelId = "tempModuleId"
+            };
+
+            // act
+            var moduleClient = ModuleClient.CreateFromConnectionString(FakeConnectionString, clientOptions);
```
drwill-ms, line 85: If this fails, wouldn't it throw?

timtay-microsoft:
We've got some communications going between us and the Edge team about whether the createFromEnvironment constructor should take the modelId as a parameter, as a new environment variable, or not at all. Please hold off on merging this in until we reach a conclusion on this

iothub/device/src/ModuleClient.cs
```diff
@@ -96,11 +101,12 @@ public static ModuleClient Create(string hostname, string gatewayHostname, IAuth
         /// <param name="hostname">The fully-qualified DNS hostname of IoT Hub</param>
         /// <param name="authenticationMethod">The authentication method that is used</param>
         /// <param name="transportSettings">Prioritized list of transportTypes and their settings</param>
+        /// <param name="options">The options that allow configuration of the module client instance during initialization.</param>
         /// <returns>ModuleClient</returns>
         public static ModuleClient Create(string hostname, IAuthenticationMethod authenticationMethod,
-            ITransportSettings[] transportSettings)
+            ITransportSettings[] transportSettings, ClientOptions options = default)
         {
```
prmathur-microsoft, line 108: Just want to make sure that this defaults to the same code path as for device client i.e. modelId sits on connect packet for MQTT ? Have you verified it with sample or other ways?

prmathur-microsoft: Also is it possible to e2e this?

barustum: Yes this is using the same ClientFactory that is used for DeviceClient. So it's following the same code path. 
@prmathur-microsoft @abhipsaMisra Does the backend already support us sending modelId on device modules?

abhipsaMisra: Azure IoT Explorer would be the easiest way to test this. If it doesn't work, I'd reach out to Miguel Nunez Avila (from the pnp service team).

iothub/device/tests/ModuleClientTests.cs
barustum: I traced through the code, and yes. It seems like we throw if something fails. I am assuming you're suggesting the Assert is pointless because the moduleClient cannot be null, correct?

iothub/device/src/ModuleClient.cs
barustum: Got it. I will write a small e2e test for module client and device client.

bikamani:
> Two Questions:
> 
> 1. Is the plan to do changes for CreateFromEnvironmentAsync in a separate PR?
> 2. Are these changes validated with a sample or just tests?

Synced up offline on this.

iothub/device/src/ModuleClient.cs
barustum: As Binal hinted offline during a sync we were having, I cannot at the moment do an e2e test because testing happens against a specific canary region. So what I did is wrote an e2e test locally, and I was able to verify from IoT Explorer the following:
"dt-dataschema": "dtmi:my:company:namespace;1"

So once this feature becomes available in all regions, we can add an e2e test. I will create a task to track this.

PR #1410, opened 2020-06-18
feat(samples): Add sample project for SimpleThermostatSample

abhipsaMisra:
These samples will live in the SDK code repo (for now), and will be moved to the sample repo when we go GA.
However, the sample does not take a project reference, but takes a package reference on the GA nuget package that has pnp support.

### DIFF
--- a/iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Models/SimpleThermostat.json
+++ b/iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Models/SimpleThermostat.json
@@ -1,0 +1,53 @@
+﻿{
+  "@context": "dtmi:dtdl:context;2",
+  "@id": "dtmi:com:example:simplethermostat;1",
+  "@type": "Interface",
+  "displayName": "Simple thermostat",
+  "contents": [
+    {
+      "@type": [
+        "Property",
+        "Temperature"
+      ],
+      "displayName": "Target Temperature",
+      "description": "Desired temperature to configure remotely.",
+      "name": "targetTemperature",
+      "schema": "double",
+      "unit": "degreeCelsius",
+      "writable": true
+    },
+    {
+      "@type": [
+        "Property",
+        "Temperature"
+      ],
+      "displayName": "Current Temperature",
+      "description": "Current temperature reported from the device as a property.",
+      "name": "currentTemperature",
+      "schema": "double",
+      "unit": "degreeCelsius",
+      "writable": false
+    },
+    {
+      "@type": [
+        "Telemetry",
+        "Temperature"
+      ],
+      "description": "Temperature reported from the device as telemetry.",
+      "displayName": "Temperature",
+      "name": "temperature",
+      "schema": "double",
+      "unit": "degreeCelsius"
+    },
+    {
+      "@type": "Command",
+      "description": "This command will reboot the device application.",
+      "displayName": "Reboot",
+      "name": "reboot",
+      "request": {
+        "name": "delay",
+        "schema": "integer"
+      }
+    }
+  ]
+}
--- a/iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Program.cs
+++ b/iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Program.cs
@@ -1,0 +1,153 @@
+﻿using System;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Shared;
+
+namespace SimpleThermostat
+{
+    public class Program
+    {
+        private const string DeviceConnectionString = "device_connection_string_here";
+        private const string ModelId = "dtmi:com:example:simplethermostat;1";
+
+        private static DeviceClient s_deviceClient;
+
+        public static async Task Main(string[] _)
+        {
+            await RunSampleAsync();
+        }
+
+        private static async Task RunSampleAsync()
+        {
+            // This sample follows the following workflow:
+            // -> Initialize device client instance.
+            // -> Set handler to receive "target temperature" updates.
+            // -> Set handler to receive "reboot" command.
+            // -> Retrieve current "target temperature".
+            // -> Send "current temperature" over both telemetry and property updates.
+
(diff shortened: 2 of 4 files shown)


iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Models/SimpleThermostat.json
```diff
@@ -0,0 +1,53 @@
+﻿{
+  "@context": "dtmi:dtdl:context;2",
```
abhipsaMisra, line 2: This interface is taken from here: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/SimpleThermostat.json

iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Program.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        private static async Task RunSampleAsync()
+        {
+            // This sample follows the following workflow:
+            // -> Initialize device client.
+            // -> Set handler to receive "target temperature" updates.
+            // -> Set handler to receive "reboot" command.
+            // -> Retrieve current "target temperature".
+            // -> Send "current temerature" update over both telemetry and property update.
```
barustum, line 27: temerature = temperature

iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Program.cs
```diff
@@ -0,0 +1,90 @@
...
+            var options = new ClientOptions
+            {
+                ModelId = ModelId,
+            };
+
+            // Initialize the device client instance using the device connection string, transport of Mqtt over TCP (with fallback to Websocket),
+            // and the device ModelId set in ClientOptions.
+            s_deviceClient = DeviceClient.CreateFromConnectionString(DeviceConnectionString, TransportType.Mqtt, options);
+
+            // Register a connection status change callback, that will get triggerred any time the device's connection status changes.
```
barustum, line 47: triggerred = triggered

iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        private static Task HandleRebootCommand()
+        {
+            return Task.CompletedTask;
+        }
+
+        private static T GetPropertyFromTwin<T>(TwinCollection collection, string propertyName)
+        {
+            return collection.Contains(propertyName) ? (T)collection[propertyName] : default;
```
abhipsaMisra, line 115: this logic is a placeholder for now, has not been tested.

bikamani:
> These samples will live in the SDK code repo (for now), and will be moved to the sample repo when we go GA.
> However, the sample does not take a project reference, but takes a package reference on the GA nuget package that has pnp support.

So with related to model id changes, I see ClientOptions being set but what else is in plan to add in for sample?

iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Program.cs
vinagesh: A cool feature is to install an extension on Visual Studio for spell check. I did it last week and it works amazingly well. There will be a pink squiggly line under anything that has a typo - Visual studio spell checker

iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Program.cs
```diff
@@ -0,0 +1,138 @@
...
+    public class Program
+    {
+        private const string DeviceConnectionString = "device_connection_string_here";
+        private const string ModelId = "dtmi:com:example:simplethermostat;1";
+
+        private static DeviceClient s_deviceClient;
+
+        public static async Task Main(string[] _)
+        {
+            await RunSampleAsync().ConfigureAwait(false);
```
vinagesh, line 18: This is a console app, so we don't need ConfigureAwait. Also I don't think we want to add this in samples based on what we heard in SDK review. @drwill-ms - Checking with you if we need to add it.

abhipsaMisra:
> > These samples will live in the SDK code repo (for now), and will be moved to the sample repo when we go GA.
> > However, the sample does not take a project reference, but takes a package reference on the GA nuget package that has pnp support.
> 
> So with related to model id changes, I see ClientOptions being set but what else is in plan to add in for sample?

I am not sure I understand the question @bikamani . 
Are you asking - what in these samples is pnp specific? If so, then yes, the only difference between this and a non-pnp sample is the fact that we are passing in the model Id; and then implementing callbacks that can handle commands and property changes as defined in the dtdl.

iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+    public class Program
+    {
+        private const string DeviceConnectionString = "device_connection_string_here";
+        private const string ModelId = "dtmi:com:example:simplethermostat;1";
+
+        private static DeviceClient s_deviceClient;
+
+        public static async Task Main(string[] _)
+        {
+            await RunSampleAsync();
```
abhipsaMisra, line 18: `.ConfigureAwait(false)` is used when we want to ensure that the callback isn't forced to be continued on the original context from where the `Task` was called. 
This is essential for library code, because if the `SynchronizationContext` has limitations on the no of threads that can be used to carry out operations, we would end up with a deadlock. 

For application code, however, this is not required, since the app environment, .NET Core in this case, will have it's own `SynchronizationContext`, which will do the right thing when a callback is posted to the original context from where the task was called. So we can let it use the default behavior.

vinagesh: I agree it is needed in our SDK(library) code for sure as it is safe for the consumer of the library. But this is not needed in samples (SDK board said that for track2 ADT samples). For application code, like you said it is not needed if someone is using .net core, otherwise it is needed.

PR #1412, opened 2020-06-19
fix(iot-device): Fix for retrieving message system properties over Http

abhipsaMisra:
An IoT Hub message is composed of -
- payload
- some pre-defined system properties
- user defined application properties

https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct

Some of the system properties are writeable by the SDK, others are populated only by the service.

This PR makes sure that system properties expected in c2d messages are parsed and returned properly by the SDK, over Http. Amqp and Mqtt currently do parse these properties as expected.

### DIFF
--- a/iothub/device/src/Transport/HttpTransportHandler.cs
+++ b/iothub/device/src/Transport/HttpTransportHandler.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
         private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromSeconds(60);
         private static readonly TimeSpan s_defaultMethodOperationTimeout = TimeSpan.FromSeconds(100);
 
-        private static readonly IDictionary<string, string> s_mapMessageProperties2HttpHeaders = new Dictionary<string, string>
+        private static readonly IDictionary<string, string> s_mapMessagePropertiesToHttpHeaders = new Dictionary<string, string>
         {
             { MessageSystemPropertyNames.Ack, CustomHeaderConstants.Ack },
             { MessageSystemPropertyNames.CorrelationId, CustomHeaderConstants.CorrelationId },
@@ -39,6 +39,20 @@ namespace Microsoft.Azure.Devices.Client.Transport
             { MessageSystemPropertyNames.ContentType, CustomHeaderConstants.ContentType },
             { MessageSystemPropertyNames.ContentEncoding, CustomHeaderConstants.ContentEncoding },
             { MessageSystemPropertyNames.InterfaceId, CustomHeaderConstants.InterfaceId }
+        };
+
+        private static readonly IDictionary<string, string> s_mapHttpHeadersToMessageProperties = new Dictionary<string, string>
+        {
+            { HttpResponseHeader.ETag.ToString(), MessageSystemPropertyNames.LockToken },
+            { CustomHeaderConstants.MessageId, MessageSystemPropertyNames.MessageId },
+            { CustomHeaderConstants.SequenceNumber, MessageSystemPropertyNames.SequenceNumber },
+            { CustomHeaderConstants.To, MessageSystemPropertyNames.To },
+            { CustomHeaderConstants.ExpiryTimeUtc, MessageSystemPropertyNames.ExpiryTimeUtc },
+            { CustomHeaderConstants.CorrelationId, MessageSystemPropertyNames.CorrelationId },
+            { CustomHeaderConstants.UserId, MessageSystemPropertyNames.UserId },
+            { CustomHeaderConstants.Ack, MessageSystemPropertyNames.Ack },
+            { CustomHeaderConstants.EnqueuedTime, MessageSystemPropertyNames.EnqueuedTime },
+            { CustomHeaderConstants.DeliveryCount, MessageSystemPropertyNames.DeliveryCount },
         };
 
         private readonly IHttpClientHelper _httpClientHelper;
@@ -87,7 +101,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
             foreach (var property in message.SystemProperties)
             {
                 string strValue = property.Value is DateTime ? ((DateTime)property.Value).ToString("o") : property.Value.ToString();
-                customHeaders.Add(s_mapMessageProperties2HttpHeaders[property.Key], strValue);
+                customHeaders.Add(s_mapMessagePropertiesToHttpHeaders[property.Key], strValue);
             }
 
             foreach (var property in message.Properties)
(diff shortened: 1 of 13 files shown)


timtay-microsoft:
Could you add an e2e test that verifies these system properties are retrieved as expected?

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -252,52 +265,33 @@ public override async Task<Message> ReceiveAsync(CancellationToken cancellationT
...
                 }
-            }
 
-            if (expiryTime != null)
-            {
-                if (DateTime.TryParse(expiryTime.First(), out DateTime absoluteExpiryTime))
+                if (s_mapHttpHeadersToMessageProperties.ContainsKey(headerKey))
                 {
-                    message.ExpiryTimeUtc = absoluteExpiryTime;
+                    string messagePropertyKey = s_mapHttpHeadersToMessageProperties[headerKey];
```
vinagesh, line 286: messagePropertyName?

abhipsaMisra:
> Could you add an e2e test that verifies these system properties are retrieved as expected?

Sure, will do.

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -427,6 +421,29 @@ private static TimeSpan GetInvokeDeviceMethodOperationTimeout(MethodInvokeReques
             return customHeaders;
         }
 
+        private static object ConvertToMessageSystemProperty(string messagePropertyName, IEnumerable<string> headerValue)
```
vinagesh, line 424: messagePropertyValues?

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -427,6 +421,29 @@ private static TimeSpan GetInvokeDeviceMethodOperationTimeout(MethodInvokeReques
...
+
+            if (messagePropertyName.Equals(MessageSystemPropertyNames.EnqueuedTime, StringComparison.Ordinal)
+                || messagePropertyName.Equals(MessageSystemPropertyNames.ExpiryTimeUtc, StringComparison.Ordinal))
+            {
+                return DateTime.TryParse(propertyValue, out DateTime dateTime) ? dateTime : (object)null;
+            }
+
+            if (messagePropertyName.Equals(MessageSystemPropertyNames.SequenceNumber, StringComparison.Ordinal))
+            {
+                return propertyValue == null ? 0 : Convert.ToUInt64(propertyValue, CultureInfo.InvariantCulture);
```
vinagesh, line 436: This convert could throw if the format is incorrect. Either we should have a try catch and throw appropriately or if it is a scenario to ignore, we should try catch and continue with the method.

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -427,6 +421,29 @@ private static TimeSpan GetInvokeDeviceMethodOperationTimeout(MethodInvokeReques
             return customHeaders;
         }
 
+        private static object ConvertToMessageSystemProperty(string messagePropertyName, IEnumerable<string> headerValue)
+        {
+            string propertyValue = headerValue?.First();
+
+            if (messagePropertyName.Equals(MessageSystemPropertyNames.EnqueuedTime, StringComparison.Ordinal)
```
vinagesh, line 428: Just double checking that case-sensitive comparison is intentional.

iothub/device/src/Transport/HttpTransportHandler.cs
abhipsaMisra: I used the key-value terminology, I can change it if it's unclear.

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -252,52 +265,33 @@ public override async Task<Message> ReceiveAsync(CancellationToken cancellationT
...
+                    object messagePropertyValue = ConvertToMessageSystemProperty(messagePropertyKey, header.Value);
+                    if (messagePropertyValue != null)
+                    {
+                        message.SystemProperties[messagePropertyKey] = messagePropertyValue;
+                    }
                 }
             }
 
-            message.CorrelationId = correlationId?.First();
-            message.SequenceNumber = sequenceNumber == null
```
vinagesh, line 297: What are these properties? After the fix, these would be null right? Should we keep the values in these fields and the SystemProperties for backward compatibility?

iothub/device/src/Transport/HttpTransportHandler.cs
abhipsaMisra: sure

iothub/device/src/Transport/HttpTransportHandler.cs
abhipsaMisra: yes, the case sensitive comparison is intentional

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -252,52 +265,33 @@ public override async Task<Message> ReceiveAsync(CancellationToken cancellationT
...
-                {
-                    message.EnqueuedTimeUtc = enqueuedTimeUtc;
-                }
-            }
+                string headerKey = header.Key;
 
-            if (deliveryCountAsStr != null)
-            {
-                if (byte.TryParse(deliveryCountAsStr.First(), out byte deliveryCount))
+                if (header.Key.Equals(HttpResponseHeader.ETag.ToString(), StringComparison.OrdinalIgnoreCase))
```
vinagesh, line 279: Can we move this logic inside ConvertToMessageSystemProperty? For all other cases, it is in that method. The logic for this key would be to trim instead of parse.

iothub/device/src/Transport/HttpTransportHandler.cs
vinagesh, line 428: Why not use a switch statement ?

iothub/device/src/Transport/HttpTransportHandler.cs
vinagesh: yeah I got that but felt like it would be easier to co-relate as I was trying to understand the code but leave it up to you.

iothub/device/src/Transport/HttpTransportHandler.cs
abhipsaMisra: I have it outside since the Etag mapping isn't available in `CustomHeaderConstants`, but I can add this as an exception.

iothub/device/src/Transport/HttpTransportHandler.cs
abhipsaMisra: `ConvertToMessageSystemProperty` will initialize it with the value, if available, else set it to null. This is the existing behavior.

iothub/device/src/Transport/HttpTransportHandler.cs
abhipsaMisra: Sequence no is not user-settable, so if service sets it to an invalid value, then the SDK will throw the exception. This is the current behavior.

iothub/device/src/Transport/HttpTransportHandler.cs
abhipsaMisra: no reason! I had started out with 2 properties, so `if` block was cleaner. I have 4 custom implementations in here now, so I can change it to switch case.

PR #1413, opened 2020-06-23
feat(iothub-service): Expose DTDL model Id property for pnp devices.

vinagesh:
## Description of the changes
Exposing the DTDL model Id on the device identity object.
This property will be returned on GET calls once we call the service with the new API version. Until then this should have no side effects.

https://msazure.visualstudio.com/One/_workitems/edit/7479580

### DIFF
--- a/iothub/service/src/Device.cs
+++ b/iothub/service/src/Device.cs
@@ -110,5 +110,13 @@ namespace Microsoft.Azure.Devices
         /// </summary>
         [JsonProperty(PropertyName = "deviceScope", NullValueHandling = NullValueHandling.Ignore)]
         public virtual string Scope { get; set; }
+
+        /// <summary>
+        /// The DTDL model id of the device.
+        /// The value will be null for a non-pnp device.
+        /// The value will be null for a pnp device until the device connects and registers with the model id.
+        /// </summary>
+        [JsonProperty(PropertyName = "digital-twin-model-id", NullValueHandling = NullValueHandling.Ignore)]
+        public virtual string ModelId { get; set; }
     }
 }


iothub/service/src/Device.cs
```diff
@@ -110,5 +110,13 @@ public Device(string id)
         /// </summary>
         [JsonProperty(PropertyName = "deviceScope", NullValueHandling = NullValueHandling.Ignore)]
         public virtual string Scope { get; set; }
+
+        /// <summary>
+        /// The DTDL model id provided by the device during the device connection.
+        /// The value will be null for a non-pnp device.
+        /// The value will be null for a pnp device until the device connects and registers with the model id.
```
vinagesh, line 117: This comment was stated for twins but i'm not sure if it applies to device identity as well. I will update it if someone knows better.

abhipsaMisra: Sounds good.

abhipsaMisra: This change needs to be a part of `Module` class as well.

timtay-microsoft:
What about the new modelId field for the device twin object?

iothub/service/src/Device.cs
abhipsaMisra: I would suggest including all three in this PR:
- model Id for device and module types
- model Id for twin type
- service API bump

iothub/service/src/Device.cs
```diff
@@ -110,5 +110,13 @@ public Device(string id)
...
         [JsonProperty(PropertyName = "deviceScope", NullValueHandling = NullValueHandling.Ignore)]
         public virtual string Scope { get; set; }
+
+        /// <summary>
+        /// The DTDL model id of the device.
+        /// The value will be null for a non-pnp device.
+        /// The value will be null for a pnp device until the device connects and registers with the model id.
+        /// </summary>
+        [JsonProperty(PropertyName = "digital-twin-model-id", NullValueHandling = NullValueHandling.Ignore)]
+        public virtual string ModelId { get; set; }
```
bikamani, line 120: +1 for the summary
Same for modules and Twin.

vinagesh:
@timtay-microsoft - I was going to add the property to twins in a separate PR as I wanted to ensure this does not break anything. I have made separate tasks on the board to represent both these tasks. If there is a downside to doing it one by one, let me know and I can do it all in one PR.

iothub/service/src/Device.cs
vinagesh: I was planning to do it in separate PR as you mentioned this morning Twin is in the Common nuget. Let me evaluate if making incremental changes will help or not and proceed accordingly.

PR #1415, opened 2020-06-24
feat(iothub-service): Expose DTDL model Id property for pnp devices.

vinagesh:
## Description of the changes
Adding changes to serialize and de-serialize DTDL modelId property returned by the Service on GET /twins/<deviceId> calls.


shared/src/TwinJsonConverter.cs
```diff
@@ -217,6 +224,10 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                         twin.DeviceId = reader.Value as string;
                         break;
 
+                    case ModelId:
```
vinagesh, line 227: This is to deserialize

shared/src/TwinJsonConverter.cs
```diff
@@ -72,10 +73,16 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+            {
+                writer.WritePropertyName(ModelId);
+                writer.WriteValue(twin.ModelId);
+            }
+
             writer.WritePropertyName(DeviceIdJsonTag);
             writer.WriteValue(twin.DeviceId);
 
-            if (!string.IsNullOrEmpty(twin.ModuleId))
+            if (!string.IsNullOrWhiteSpace(twin.ModuleId))
```
vinagesh, line 85: This is to serialize

shared/src/TwinJsonConverter.cs
```diff
@@ -39,6 +39,7 @@ public sealed class TwinJsonConverter : JsonConverter
         private const string CloudToDeviceMessageCountTag = "cloudToDeviceMessageCount";
         private const string AuthenticationTypeTag = "authenticationType";
         private const string X509ThumbprintTag = "x509Thumbprint";
+        private const string ModelId = "digital-twin-model-id";
```
prmathur-microsoft, line 43: Is this change working now? This is expected to change fyi.

iothub/service/src/ClientApiVersionHelper.cs
```diff
@@ -12,7 +12,7 @@ internal class ClientApiVersionHelper
     {
         private const string ApiVersionQueryPrefix = "api-version=";
         private const string ApiVersionGA = "2016-02-03";
-        private const string ApiVersionDefault = "2018-08-30-preview";
+        private const string ApiVersionDefault = "2020-05-31-preview";
         private const string ApiVersionLimitedAvailability = "2020-03-13";
```
prmathur-microsoft, line 16: How are you verifying this works e2e?

shared/src/Twin.cs
```diff
@@ -48,6 +48,13 @@ public Twin(TwinProperties twinProperties)
         /// </summary>
         public string DeviceId { get; set; }
 
+        /// <summary>
+        /// The DTDL model id of the device.
+        /// The value will be null for a non-pnp device.
+        /// The value will be null for a pnp device until the device connects and registers with the model id.
+        /// </summary>
+        public string ModelId { get; set; }
+
```
prmathur-microsoft, line 57: Is there an e2e possible for this scenario?

abhipsaMisra:
From Miguel's email today to PnPPartners alias, it looks like this is now available on East US 2 EUAP; we should be able to add a test around this

shared/src/TwinJsonConverter.cs
abhipsaMisra: Isn't the change specific to what we send on the mqtt connect packet? This shouldn't be affected in that case.
Or are we talking about a different change here?

shared/src/Twin.cs
timtay-microsoft: It'll depend on if the hub's that support this are available in our canary region or not.

shared/src/TwinJsonConverter.cs
vinagesh: Yes it is working. 1) There is backward compatibility as existing tests are passing 2) This has been manually tested using samples against the hub they provided

shared/src/Twin.cs
vinagesh: Not yet. There is a task on my card to do it when their service is deployed in Canary.

vinagesh: Based on Abhipsa's update below the service might be available today. So i'll add it in the next PR.

iothub/service/src/ClientApiVersionHelper.cs
vinagesh: The E2E is running in the pipeline so it works in terms of existing scenarios. For new scenarios, as I mentioned in all other places, a test will be added when the service is in Canary.

shared/src/TwinJsonConverter.cs
prmathur-microsoft: Great that it is working. Then lets take the followup change in next PR. The rename applies to both device and service SDK's.

bikamani:
Merging this change as per offline discussion with @vinagesh

Issue #1239, opened 2020-02-15
ModuleClient configured with Mqtt_Tcp_Only fails to reconnect after edgeHub restart then never recovers


and-rewsmith:
- **OS, version, SKU and CPU architecture used:** <VERSION> 
OS: Ubuntu 18.04 amd64

- **Application's .NET Target Framework :** 2.1 
- **Device:** Local VM Ubuntu 18.04 
- **SDK version used:** 
    PackageReference Include="Microsoft.Azure.Devices" Version="1.22.0" 

## Description of the issue:

This problem can be reproduced consistently. 

Start edgeHub with one [SimulatedTemperatureSensor](https://github.com/Azure/iotedge/tree/release/1.0.9/edge-modules/SimulatedTemperatureSensor) configured with Mqtt_Tcp_Only TransportType. The module will successfully make the initial connection and start sending messages. If edgeHub is restarted, the ModuleClient.SendEventAsync() will hang then eventually timeout. This then crashes the module, and a new ModuleClient will get created and start operating as expected. 

To further test, I added a try/catch around the ModuleClient.SendEventAsync() to detect if ModuleClient would recover after its initial timeout. Unfortunately, ModuleClient will **never recover** and timeout after **all** subsequent SendEventAsync() calls.

## Code sample exhibiting the issue:

Run iotedge with one [SimulatedTemperatureSensor module](https://github.com/Azure/iotedge/tree/release/1.0.9/edge-modules/SimulatedTemperatureSensor) configured with Mqtt_Tcp_Only. Restart edgeHub and this timeout will occur. 

## Console log of the issue:

SimulatedTemperatureSensor logs:
```
[2020-02-15 00:55:29 : Starting Module
SimulatedTemperatureSensor Main() started.
Initializing simulated temperature sensor to send unlimited messages, at an interval of 5 seconds.
To change this, set the environment variable MessageCount to the number of messages that should be sent (set it to -1 to send unlimited messages).
Information: Trying to initialize module client using transport type [Mqtt_Tcp_Only].
Information: Successfully initialized module client of transport type [Mqtt_Tcp_Only].
	02/15/2020 00:55:37> Sending message: 1, Body: [{"machine":{"temperature":21.842550837710757,"pressure":1.0959868042961622},"ambient":{"temperature":21.113807129028164,"humidity":26},"timeCreated":"2020-02-15T00:55:37.2835997Z"}]
[...]

PR #1416, opened 2020-06-25
feat(versions): Bump versions for .NET SDK preview 2020-6-25 release

abhipsaMisra:
**Versioning for preview nuget packages:**
Eg Microsoft.Azure.Devices.Client.1.29.0-preview-004
- Increment "1.29.0-preview-**_004_**" - when you are simply bringing in changes from master (doesn't matter if they are patch or minor updates), or when there are patch updates specifically on "preview" code.
- Increment "1.29.**_0_**-preview-004" - when there is an API surface change specifically for "preview" code.
- Increment "1.**_29_**.0-preview-004" - we don't do this.

**Release notes:**

## Microsoft.Azure.Devices.Client.1.29.1-preview-001
* Rename `DeviceStreamRequest.Url` to `DeviceStreamRequest.Uri`, since it represents a Uri.
   - **NOTE: This is a breaking change for `DeviceStreaming` APIs.**
* SDK fixes from `master` branch:
    * Fix issue where MQTT `ReceiveAsync` with cancellation token always returns null.
    * Fix issue where MQTT packet id becomes 0 on 32kth messages sent, which caused MQTT connection to drop.
    * Fix issue where AMQP `SendEventBatchAsync` always throws `InvalidOperationException`.
    * Fix issue where AMQP connection throws `ObjectDisposedException` when `OpenAsync` fails to open sender link.
    * Fix issue where AMQP `SendTwinPatchAsync` with cancellation token doesn’t cancel after cancellation token is cancelled.
    * Revert to throw `TimeoutException` if operation times out instead of `IotHubCommunicationException`.
    * Revert to throw `OperationCancelledException` if operation is cancelled instead of `IotHubCommunicationException`.
    * Fix connection string parser to allow special characters such as semicolon.
    * Fix a missing await for async task in MQTT which could cause race condition to the connection.
    * Add clarification to connection status callback documentation for HTTP.
    * Add clarification to device operation timeout documentation.
    * Upgraded `Newtonsoft.Json` dependency to 12.0.3
    * Fixed twin parsing to use `DateTimeOffset` instead of `DateTime` (#1122)
    * Fixed Mqtt adaptor not always cleaning up while closing (PR #1181)
    * Fixed `MqttTransportHandler` not handling `ConnectExceptions` correctly (#648)
    * Fixed the SDK not throwing an exception when the mqtt topic name was too large to publish (#1197)
    * Added API to specify client side AMQP heartbeat interval.
    * Centralize TLS version choice, prefer None where possible.
    * Added API to enable/disable remote SSL certificate revocation check.
    * Exposed the default transport settings values.
    * Fixed null reference exception when device sets reported property but does not specify a desired property callback with AMPQ.
    * Fix issue where `ModuleClient` configured with `Mqtt_Tcp_Only` fails to reconnect after edgeHub restart then never recovers (#1239)
    * When using AMQP, application properties with a null value will now be sent to the hub, matching MQTT/HTTP behavior (#919)
    * Improve AMQP reconnection resiliency
    * SDK will now retry on more error codes over AMQP. For more details, see [amqp transport exceptions](https://github.com/Azure/azure-iot-sdk-csharp/blob/preview/iothub/device/devdoc/amqpTransportExceptions.md).
        - `amqp:internal-error`
        - `amqp:link :transfer-limit-exceeded`
        - `amqp:resource-locked`
    * Dropped support for .NET Standard 1.3, and added support for .NET Framework 4.7.2.
    * Improved XML comments
    * `MqttTransportSettings.CertificateRevocationCheck` and `AmqpTransportSettings.CertificateRevocationCheck` are now linked with `TlsVersions.Instance.CertificateRevocationCheck` property.
    * Made `Message` class partially cloneable - to be able to copy system properties to a new instance (#1923). See `Message.CloneWithBody` method.
    * Updated `System.Net.Http` package from 4.3.3 to 4.3.4 to resolve a security vulnerability in the package.
    * Add `ClientOptions` to enable users to supply model id
    * Add support for .NET Standard 2.1
    * Add a way for nested Edge to pass authentication chain to IoTHub
    * Add twin array support and update to latest service API version
    * Enable `RemoteCertificateValidationCallback` for websocket connections to allow users to validate remote certificates
    * Fix Mqtt C2D subscription message to set QoS of 1
    * Fix comments on twin/methods and create
    * Fix issue to require explicit subscribe to desired properties in latest Service API version
    * Fix for each subscribe request declare its topic name on mqtt messages
    * Remove DNS lookup in mqtt stack when proxy is present
    * Don't dispose a message being returned
    * Dispose `CancellationTokenSource` being used internally in the library

## Microsoft.Azure.Devices.1.27.1-preview-001
* Rename `DeviceStreamResponse.Url` to `DeviceStreamResponse.Uri`, since it represents a Uri.
   - **NOTE: This is a breaking change for `DeviceStreaming` APIs.**
* SDK fixes from `master` branch:
    * Upgraded `Newtonsoft.Json` dependency to 12.0.3
    * Fix proxy handling on HTTP for requests with per-request timeout.
    * Centralize TLS version choice, prefer None where possible.
    * Fixed proxy handling for Http client.
    * `IotHubClientWebSocket.ConnectAsync` method now respects the `SslProtocol` value specified in `TlsVersions.Instance.Preferred`.
    * Dropped support for .NET Standard 1.3, and added support for .NET Framework 4.7.2.
    * Enables new security functionality to specify how hub should authenticate when working with a storage account when running import/export devices jobs. 
        - New enum `StorageAuthenticationType` to specify key-based or identity-based authentication.
        - New property `JobProperties.StorageAuthenticationType` for use in import/export devices jobs.
        - Added two overloaded methods: `RegistryManager.ImportDevicesAsync` and `RegistryManager.ExportDevicesAsync`. The overloads take a `JobProperties` instance so the user can pass all required and any optional parameters.
        - Two factory methods to make it easier to call `ImportDevicesAsync` and `ExportDevicesAsync`, making it clear which properties are required and which are optional. They are `JobProperties.CreateForInportJob` and `JobProperties.CreateForExportJob` respectively.
        - Removed two unreferenced classes: `ImportDevicesRequest` and `ExportDevicesRequest`.
    * Reuse `HttpClient` instances (#1276), avoiding a known issue with `HttpClient.Dispose()` that can cause TCP socket exhaustion under heavy load.
    * Use latest Service API version that is available in all regions
    * Add Support for .NET Standard 2.1
    * Add twin array support
    * Update `JobQuotaExceededException` to include the associated http error reason
    * Code clean up in `IotHubConnectionStringBuilder` based on coding standards


timtay-microsoft:
> Rename DeviceStreamRequest.Url to DeviceStreamRequest.Uri, since it represents an Uri.

It's worth calling out that this is actually a small breaking change

abhipsaMisra:
> > Rename DeviceStreamRequest.Url to DeviceStreamRequest.Uri, since it represents an Uri.
> 
> It's worth calling out that this is actually a small breaking change

Sure

timtay-microsoft:
> Reuse HttpClient instances (#1276)
> Reuse HttpClient instances (#1276), avoiding a known issue with HttpClient.Dispose() that can cause TCP socket exhaustion under heavy load.

Duplicate?

timtay-microsoft:
> Fix for each subscribe request declare its topic name on mqtt messages

This was just an internal code cleanup. It won't affect customers, so you can omit this note

abhipsaMisra:
> > Reuse HttpClient instances (#1276)
> > Reuse HttpClient instances (#1276), avoiding a known issue with HttpClient.Dispose() that can cause TCP socket exhaustion under heavy load.
> 
> Duplicate?

Well, no. It was fixed over multiple releases, but I can definitely consolidate it into a single note here.

abhipsaMisra:
> > Fix for each subscribe request declare its topic name on mqtt messages
> 
> This was just an internal code cleanup. It won't affect customers, so you can omit this note

I don't see any issue in mentioning this. We do mention changes that we make that doesn't affect customers directly.

versions.csv
```diff
@@ -1,12 +1,12 @@
...
-provisioning\transport\amqp\src\Microsoft.Azure.Devices.Provisioning.Transport.Amqp.csproj, 1.1.9
-provisioning\transport\http\src\Microsoft.Azure.Devices.Provisioning.Transport.Http.csproj, 1.1.6
-provisioning\transport\mqtt\src\Microsoft.Azure.Devices.Provisioning.Transport.Mqtt.csproj, 1.1.8
-security\tpm\src\Microsoft.Azure.Devices.Provisioning.Security.Tpm.csproj, 1.1.6
-provisioning\service\src\Microsoft.Azure.Devices.Provisioning.Service.csproj, 1.5.0
-digitaltwin\device\src\Azure.Iot.DigitalTwin.Device.csproj, 1.0.0-preview-001
-digitaltwin\service\src\Azure.Iot.DigitalTwin.Service.csproj, 1.0.0-preview-001
+iothub\device\src\Microsoft.Azure.Devices.Client.csproj, 1.29.1-preview-001
+iothub\service\src\Microsoft.Azure.Devices.csproj, 1.27.1-preview-001
+shared\src\Microsoft.Azure.Devices.Shared.csproj, 1.20.0
```
abhipsaMisra, line 4: non "-preview" version updates are simply updates from "master". They will not be released again.

PR #1417, opened 2020-06-26
fix(vsts): Update preview release pipeline to pick up correct artifacts

abhipsaMisra:


vsts/test-release-nuget.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
         inputs:
           buildType: 'specific'
           project: 'f9b79625-2860-4d92-a4ee-57b03fabfd10'
-          pipeline: '278'
+          pipeline: '355'
```
abhipsaMisra, line 27: This needs to be the Id of the pipeline that will generate the signed nugets for testing. 
For preview, it needs to be: https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build?definitionId=355&_a=summary

abhipsaMisra: This had been pointing to the GA release pipeline instead: https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build?definitionId=278

Issue #1042, opened 2019-08-29
Retry-after header should be used to delay and backup in case server returns statusCode >= 429


alextolp:
- **SDK version used:** 1.4 

## Description of the issue:

The issue we’re seeing is that retry-after header for HTTP is not observed when request are being throttled. In case of status code >= 429, there would be an exception thrown here:

https://github.com/Azure/azure-iot-sdk-csharp/blob/master/provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs#L105 

When that happens we need to extract retry-after header and delay before retrying the operation.

bikamani:
This is already been taken care of hence closing the issue.

PR #1419, opened 2020-06-26
fix(dps): Delay based on retryAfter value for statusCode >= 429

bikamani:
Fix for this [bug](https://github.com/Azure/azure-iot-sdk-csharp/issues/1042)

### DIFF
--- a/e2e/test/netstandard20/ProvisioningE2ETests.cs
+++ b/e2e/test/netstandard20/ProvisioningE2ETests.cs
@@ -584,7 +584,7 @@ namespace Microsoft.Azure.Devices.E2ETests
             }
             catch (ProvisioningTransportException ex)
             {
-                Assert.IsTrue(ex.Message.Contains("404201"));
+                Assert.IsTrue(ex.Message.Contains("HTTP transport exception"));
             }
         }
 
--- a/e2e/test/netstandard20/ProvisioningE2ETests.cs
+++ b/e2e/test/netstandard20/ProvisioningE2ETests.cs
@@ -584,7 +584,8 @@ namespace Microsoft.Azure.Devices.E2ETests
             }
             catch (ProvisioningTransportException ex)
             {
-                Assert.IsTrue(ex.Message.Contains("HTTP transport exception"));
+                // Exception message must contain the errorCode value as below
+                Assert.IsTrue(ex.Message.Contains("404201"));
             }
         }
 
@@ -599,6 +600,7 @@ namespace Microsoft.Azure.Devices.E2ETests
             }
             catch (ProvisioningTransportException ex)
             {
+                // Exception message must contain the errorCode value as below
                 Assert.IsTrue(ex.Message.Contains("404201"));
             }
         }
@@ -624,6 +626,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                 Assert.AreEqual(ProvisioningRegistrationStatusType.Failed, result.Status);
                 Assert.IsNull(result.AssignedHub);
                 Assert.IsNull(result.DeviceId);
+                // Exception message must contain the errorCode value as below
                 Assert.AreEqual(404201, result.ErrorCode);
             }
         }
--- a/provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
+++ b/provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
@@ -155,12 +155,21 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
                         operationId,
                         message.IdScope).ConfigureAwait(false);
                     }
-                    catch(HttpOperationException oe) when (oe.Response.StatusCode >= (HttpStatusCode)429)
+                    catch(HttpOperationException oe)
                     {
+                        bool isTransient = oe.Response.StatusCode >= HttpStatusCode.InternalServerError
+                            || (int)oe.Response.StatusCode == 429;
                         try
                         {
                             var errorDetails = JsonConvert.DeserializeObject<ProvisioningErrorDetailsHttp>(oe.Response.Content);
-                            serviceRecommendedDelay = errorDetails.RetryAfter;
+                            if (isTransient)
+                            {
+                                serviceRecommendedDelay = errorDetails.RetryAfter;
+                            }
+                            else
+                            {
+                                throw new ProvisioningTransportException(oe.Response.Content, oe, isTransient, errorDetails);
+                            }
                         }
                         catch (JsonException ex)
                         {
@@ -194,6 +203,35 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
                 }
 
                 return ConvertToProvisioningRegistrationResult(operation.RegistrationState);
+            }
+            catch (HttpOperationException oe)
+            {
+                if (Logging.IsEnabled) Logging.Error(
+                   this,
+                   $"{nameof(ProvisioningTransportHandlerHttp)} threw exception {oe}",
+                   nameof(RegisterAsync));
+
+                bool isTransient = oe.Response.StatusCode >= HttpStatusCode.InternalServerError
+                    || (int)oe.Response.StatusCode == 429;
+
+                try
+                {
+                    var errorDetails = JsonConvert.DeserializeObject<ProvisioningErrorDetailsHttp>(oe.Response.Content);
+                    throw new ProvisioningTransportException(oe.Response.Content, oe, isTransient, errorDetails);
+                }
+                catch (JsonException ex)
+                {
+                    if (Logging.IsEnabled) Logging.Error(
+                        this,
+                        $"{nameof(ProvisioningTransportHandlerHttp)} server returned malformed error response." +
+                        $"Parsing error: {ex}. Server response: {oe.Response.Content}",
+                        nameof(RegisterAsync));
+
+                    throw new ProvisioningTransportException(
+                        $"HTTP transport exception: malformed server error message: '{oe.Response.Content}'",
+                        ex,
+                        false);
+                }
             }
             catch (Exception ex) when (!(ex is ProvisioningTransportException))
             {
--- a/provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
+++ b/provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
@@ -158,7 +158,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
                     catch(HttpOperationException oe)
                     {
                         bool isTransient = oe.Response.StatusCode >= HttpStatusCode.InternalServerError
-                            || oe.Response.StatusCode == HttpStatusCode.TooManyRequests;
+                            || (int)oe.Response.StatusCode == 429;
                         try
                         {
                             var errorDetails = JsonConvert.DeserializeObject<ProvisioningErrorDetailsHttp>(oe.Response.Content);
@@ -216,7 +216,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
                    nameof(RegisterAsync));
 
                 bool isTransient = oe.Response.StatusCode >= HttpStatusCode.InternalServerError
-                    || oe.Response.StatusCode == HttpStatusCode.TooManyRequests;
+                    || (int)oe.Response.StatusCode == 429;
 
                 try
                 {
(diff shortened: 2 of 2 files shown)


timtay-microsoft:
I get that you can't add an e2e test for this, but did you verify this manually somehow?

e2e/test/netstandard20/ProvisioningE2ETests.cs
```diff
@@ -584,7 +584,7 @@ public async Task ProvisioningDeviceClient_InvalidRegistrationId_TpmRegister_Htt
             }
             catch (ProvisioningTransportException ex)
             {
-                Assert.IsTrue(ex.Message.Contains("404201"));
+                Assert.IsTrue(ex.Message.Contains("HTTP transport exception"));
```
azabbasi, line 587: do you have an example of this exception handy? this seems a bit odd that we have to rely on an exception message text ...

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -148,10 +148,34 @@ public ProvisioningTransportHandlerHttp()
...
 
-                    operation = await client.RuntimeRegistration.OperationStatusLookupAsync(
+                    try
+                    {
+                        operation = await client.RuntimeRegistration.OperationStatusLookupAsync(
                         registrationId,
                         operationId,
                         message.IdScope).ConfigureAwait(false);
+                    }
+                    catch(HttpOperationException oe) when (oe.Response.StatusCode >= (HttpStatusCode)429)
```
azabbasi, line 158: this also applies to all 500 error codes ... should we be catching 429 specifically ?

e2e/test/netstandard20/ProvisioningE2ETests.cs
vinagesh: +1 it would be good if there were a specific Exception code and http status combination to assert against.

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
vinagesh: Also if we decide to catch 500s, the header won't have the retry-after header as it will not be Too Many Request. In that case, we need to set it to default value.

e2e/test/netstandard20/ProvisioningE2ETests.cs
bikamani: I wish there is a better way to check this but there's not any code I can assert for.

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
bikamani: That's a good question. 

@vinagesh There is a logic above where retry-after is null it will set it to default value of 3s but the major question here is in case of internal error or something do we want to retry or not. I assume it to be safe to retry only for 429, in case of any other error just throw exception.

Tagging @prmathur-microsoft to confirm if it is safe to assume the above.

bikamani:
> I get that you can't add an e2e test for this, but did you verify this manually somehow?

verified the code path with a different failure.

e2e/test/netstandard20/ProvisioningE2ETests.cs
abhipsaMisra: what changed here, wrt behavior, that we cannot assert with 404201 anymore?

abhipsaMisra:
this is for the registration operation - wouldn't this be required for mqtt and amqp as well, or is this change already in place there?

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -148,10 +148,34 @@ public ProvisioningTransportHandlerHttp()
 
                     cancellationToken.ThrowIfCancellationRequested();
 
-                    operation = await client.RuntimeRegistration.OperationStatusLookupAsync(
+                    try
+                    {
+                        operation = await client.RuntimeRegistration.OperationStatusLookupAsync(
```
abhipsaMisra, line 153: Q - does `RegisterAsync()` throw 429, or only `OperationStatusLookupAsync()`? This wouldn't work if `RegisterAsync()` were to throw a throttling exception.

bikamani: OperationStatusLookupAsync returns RegistrationOperationStatus so it should be OperationStatusLookupAsync which throws 429

bikamani:
> this is for the registration operation - wouldn't this be required for mqtt and amqp as well, or is this change already in place there?

I see it implemented for AMQP. MQTT doesn't even have any retry mechanism or a way to check operation status as of now, if we were to implement it for mqtt (which is not the ask as per the github issue) I would add an item to backlog.

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
azabbasi: I don't think this logic as it is works in our favor, we have to be a lot more specific if we want to catch a range of exception codes. Best would be to only catch 429s and let the natural flow go through otherwise.

e2e/test/netstandard20/ProvisioningE2ETests.cs
azabbasi: apparently the message does not contain that string anymore (based on a screenshot i got from Binal) that's another reason why we shouldn't look for certain strings in an exception message since that is prone to change. but it doesn't seem like there is a status code on that exception type either. Who owns this exception? the Service? Can we ask for status code?

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
bikamani: I believe I do have an answer. This catch block logic would be for any transient errors. I have made necessary changes, re-review new commit.

e2e/test/netstandard20/ProvisioningE2ETests.cs
bikamani: So on this note, that behavior should not change as Abhipsa mentioned. I made sure to add a relevant catch block

Yes you are right @azabbasi , this is the exception we are converting too for the response. I made sure in new changes, it does capture this error code. I just added a note to better explain it.

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -148,10 +148,43 @@ public ProvisioningTransportHandlerHttp()
...
                         registrationId,
                         operationId,
                         message.IdScope).ConfigureAwait(false);
+                    }
+                    catch(HttpOperationException oe)
+                    {
+                        bool isTransient = oe.Response.StatusCode >= HttpStatusCode.InternalServerError
+                            || (int)oe.Response.StatusCode == 429;
```
azabbasi, line 161: use the HttpStatusCode enum here so you don't have to cast it to Int

bikamani: agreed, I'll take care of that in any follow up PR

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -148,10 +148,43 @@ public ProvisioningTransportHandlerHttp()
...
+                        try
+                        {
+                            var errorDetails = JsonConvert.DeserializeObject<ProvisioningErrorDetailsHttp>(oe.Response.Content);
+                            if (isTransient)
+                            {
+                                serviceRecommendedDelay = errorDetails.RetryAfter;
+                            }
+                            else
+                            {
```
abhipsaMisra, line 170: add error logging here

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
abhipsaMisra, line 161: use `HttpStatusCode.TooManyRequests` here instead

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -148,10 +148,43 @@ public ProvisioningTransportHandlerHttp()
...
+                    catch(HttpOperationException oe)
+                    {
+                        bool isTransient = oe.Response.StatusCode >= HttpStatusCode.InternalServerError
+                            || (int)oe.Response.StatusCode == 429;
+                        try
+                        {
+                            var errorDetails = JsonConvert.DeserializeObject<ProvisioningErrorDetailsHttp>(oe.Response.Content);
+                            if (isTransient)
+                            {
+                                serviceRecommendedDelay = errorDetails.RetryAfter;
```
abhipsaMisra, line 167: errorDetails will not have RetryAfter header for status code 5xx, those cases should set to the default retry after delay instead.

bikamani: It will be defaulted to s_defaultOperationPoolingIntervalMilliseconds at line 147

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -174,11 +207,12 @@ public ProvisioningTransportHandlerHttp()
             catch (HttpOperationException oe)
```
abhipsaMisra, line 207: any value in having this catch block: https://github.com/Azure/azure-iot-sdk-csharp/blob/bikamani/retryAfter/provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs#L207-L235?

bikamani: yes for RegisterDeviceAsync at line 117

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
bikamani: added

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
bikamani: done

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
abhipsaMisra: it does nothing more than throw a `ProvisioningTransportException`. You can merge both the catch blocks into a single block: if exception thrown is not PTE, then evaluate if it is transient, and throw the corresponding PTE.

bikamani: Can there not be other type of exception than HttpOperationException?

abhipsaMisra: the register operation will only throw an HttpOperationException, that's what the impl does afaik. But your catch block would handle other cases as well.

```
catch (Exception ex) when (!(ex is ProvisioningTransportException))
{
	if (Logging.IsEnabled) Logging.Error(
		this,
		$"{nameof(ProvisioningTransportHandlerHttp)} threw exception {ex}",
		nameof(RegisterAsync));

	bool isTransient = false;
	if (ex is HttpOperationException)
	{
		HttpOperationException oe = (HttpOperationException)ex;
		isTransient = oe.Response.StatusCode >= HttpStatusCode.InternalServerError || HttpStatusCode.TooManyRequests == oe.Response.StatusCode;

		try
		{
			var errorDetails = JsonConvert.DeserializeObject<ProvisioningErrorDetailsHttp>(oe.Response.Content);
			throw new ProvisioningTransportException(oe.Response.Content, oe, isTransient, errorDetails);
		}
		catch (JsonException jex)
		{
			if (Logging.IsEnabled) Logging.Error(
				this,
				$"{nameof(ProvisioningTransportHandlerHttp)} server returned malformed error response." +
				$"Parsing error: {jex}. Server response: {oe.Response.Content}",
				nameof(RegisterAsync));

			throw new ProvisioningTransportException(
				$"HTTP transport exception: malformed server error message: '{oe.Response.Content}'",
				jex,
				false);
		}
	}

	throw new ProvisioningTransportException($"HTTP transport exception", ex, isTransient);
}
```

abhipsaMisra: now that i've written this, I don't have a very strong preference to change it; I prefer it this way, but you can leave as is, if you like.

bikamani: yeah agree, that's more of a preference and how you like to see. I find less nesting is more readable

PR #1423, opened 2020-06-30
feat(samples): Add 2 component sample

abhipsaMisra:
DTDL model for TemperatureController sample: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json


iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Shared;
+using PnpHelpers;
+
+namespace TemperatureController
+{
+    public class Program
+    {
+        private const string DeviceConnectionString = "device_connection_string_here";
```
vinagesh, line 15: How about a command line input to the sample instead? Otherwise there is a chance one might accidentally check this value in.

abhipsaMisra: The cross-SDK consensus was to supply this to the sample via environmental variables; so I'll make the change for that.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        // Send the current temperature over telemetry and reported property.
+        private static async Task SendCurrentTemperatureAsync()
+        {
+            // Send current temperature over telemetry.
+            string telemetryName = "temperature";
+
+            // Generate a random value between 40F and 90F for the current temperature reading.
+            double currentTemperature = new Random().Next(40, 90);
```
vinagesh, line 73: It is a best practice to create a single static instance of Random in the class and use it wherever needed. In this case we have not used Random any other place but it will be useful when we need it.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            string telemetryPayload = $"{{ \"{telemetryName}\": {currentTemperature} }}";
```
prmathur-microsoft, line 75: Can this be built using helper?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+            };
+
+            message.Properties.Add(PnpHelper.TelemetryComponentPropertyName, Thermostat1);
+
+            await s_deviceClient.SendEventAsync(message);
+            PrintLog($"Sent current temperature {currentTemperature}F over telemetry.");
+
+            // Send current temperature over reported property.
+            string reportedPropertyPatch = PnpHelper.CreateReportedPropertiesPatch("currentTemperature", currentTemperature, Thermostat1);
+            var reportedProperty = new TwinCollection(reportedPropertyPatch);
```
prmathur-microsoft, line 89: 👍

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: good point, I'll make that change

pnp/PnpHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+                    $"{{" +
+                    $"  \"{propertyName}\": " +
+                    $"      {{ " +
+                    $"          \"value\": \"{propertyValue}\" " +
+                    $"      }} " +
+                    $"}}";
+            }
+
+            return $"" +
+                $"{{" +
```
prmathur-microsoft, line 27: Just a thought experiment, how would this work for nested component cases?

pnp/PnpHelper.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace PnpHelpers
+{
+    public static class PnpHelper
+    {
```
prmathur-microsoft, line 9: Will there be a followup PR for methods and desired properties?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: I didn't want to create the complete Message via the helper, since that'll bring device client dependency into it, and I wanted it to be as generic as possible. 
This seemed pretty simple to build, so I left it in the sample. If we feel this is something we can add into the helper, I can create a method to simply format the telemetry payload, which I think is what you're asking for here.

pnp/PnpHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+        public const string TelemetryComponentPropertyName = "$.sub";
+        private const string PropertyComponentIdentifier = "\"__t\": \"c\",";
+
+        public static string CreateReportedPropertiesPatch(string propertyName, object propertyValue, string componentName = null)
```
prmathur-microsoft, line 13: I would love to see createTelemetry message to be on same lines. Can we use Bytes API from device client so that helper does the needed pnp stuff on properties/headers?

pnp/PnpHelper.cs
abhipsaMisra: yes, this is just to get initial feedback on the approach for the helper class.

pnp/PnpHelper.cs
abhipsaMisra: good point, do we have a sample patch for that? All the samples in the spec doc were single component examples.
If I can get a nested component patch, I can see how to make this more flexible.

pnp/PnpHelper.cs
abhipsaMisra: I agree, I'll see what I can add without having to bring in an existing nuget dependency

iothub/device/samples/PnpDeviceSamples/Thermostat/Models/TemperatureController.json
```diff
@@ -0,0 +1,60 @@
...
+    {
+      "@type": [
+        "Telemetry",
+        "DataSize"
+      ],
+      "name": "workingSet",
+      "displayName": "Working Set",
+      "description": "Current working set of the device memory in KiB.",
+      "schema": "double",
+      "unit": "kibibyte"
```
bikamani, line 17: Wondering if this is a right value?

iothub/device/samples/PnpDeviceSamples/Thermostat/Models/Thermostat.json
```diff
@@ -0,0 +1,89 @@
...
+      "@type": [
+        "Property",
+        "Temperature"
+      ],
+      "name": "targetTemperature",
+      "schema": "double",
+      "displayName": "Target Temperature",
+      "description": "Allows to remotely specify the desired target temperature.",
+      "unit": "degreeCelsius",
+      "writable": true
```
bikamani, line 29: is the default false?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Shared;
+using PnpHelpers;
+
+namespace TemperatureController
+{
+    public class Program
+    {
+        private const string DeviceConnectionString = "device_connection_string_here";
+        private const string ModelId = "dtmi:com:example:TemperatureController;1";
```
bikamani, line 16: _camelCase for private properties here and everywhere?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+        public static async Task Main(string[] _)
+        {
+            await RunSampleAsync();
+        }
+
+        private static async Task RunSampleAsync()
+        {
+            // This sample follows the following workflow:
+            // -> Initialize device client instance.
+            // -> Send "current temperature" over both telemetry and property updates.
```
bikamani, line 31: Does this also need to add "Register callback and Handle command (reboot)"?

pnp/PnpHelper.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace PnpHelpers
+{
+    public static class PnpHelper
```
bikamani, line 8: Does this helper suppose to have PnP_Sample_Helper_Parse or did I miss something?

iothub/device/samples/PnpDeviceSamples/Thermostat/Models/TemperatureController.json
abhipsaMisra: yeah, it's 1024 bytes -> to differentiate that kilo is usually 1000 and not 1024: https://en.wikipedia.org/wiki/Kibibyte

iothub/device/samples/PnpDeviceSamples/Thermostat/Models/Thermostat.json
abhipsaMisra: yes, it is false: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/dtdlv2.md#property

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: yes, true for private properties, but this is a const, for which we follow CamelCase

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: yes, this is the initial impl, the commands and desired property callback implementations will follow

pnp/PnpHelper.cs
abhipsaMisra: I don't know what you mean by that, can you elaborate on it? What is `PnP_Sample_Helper_Parse`?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: @vinagesh @bikamani : FYI - the env var name to be used for service connection string: IOTHUB_CONNECTION_STRING

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
bikamani, line 19: why not readonly?

bikamani: you meant PascalCasing right? But isn't that for local variables/fields?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
bikamani: Got it.

pnp/PnpHelper.cs
bikamani: It is in the spec if you see.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: yes, you are right, I meant PascalCasing. 
By convention, C# uses PascalCase for all constants; you can find some documentation here: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

Also, this for reference: https://stackoverflow.com/questions/242534/c-sharp-naming-convention-for-constants

pnp/PnpHelper.cs
abhipsaMisra: Ah, you mean, this from the specs doc:
```C
ampleTwinCb(JSON update) { 
PnP_Sample_Helper_Parse(update, ...); 
//"Component:Foo", "Pro:bar", "val" 
 
char* jsonToSend = PnP_Helper_CreateReported("foo", "bar", 200, value); 
moduleClient.SendProprety(jsonToSend); 
}
```

This is just a sample code on how the helper is expected to work. We don't need to follow it word-to-word. 
What this does mean is, we need a helper that can create a pnp-compatible reported property patch; which is something we will have under `PnpHelper`.

Also, I would like to re-iterate, this is work-in-progress, so not all helpers are implemented yet. That being said, feel free to leave your comments on the ones that have been implemented, all feedback is much-appreciated! :)

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: you mean, define this as a `static readonly`, as opposed to a `const`? A `const` is always `static`, by default., and is preferred when you want to assign a constant value (available at compile time) to a variable. 

One place where I find with static readonly fields useful, is when you want to initialize it only when the class instance is accessed for the first time (if it is a resource heavy operation, etc).

You can reference this stackoverflow answer for more details: https://stackoverflow.com/questions/55984/what-is-the-difference-between-const-and-readonly-in-c

bikamani: Ok. The dotnet guide you mentioned doesn't say anything about private const though.

bikamani: Also our coding [guideline ](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/doc/coding-style.md)says to follow PascalCasing for constant local variables and fields

abhipsaMisra: > Ok. The dotnet guide you mentioned doesn't say anything about private const though.

The access modifier here doesn't matter, for constant fields, we use PascalCase.
Had it not been a const, then we would have declared it as internal/ private, and used camelCase (with the appropraite prefix for static, instance var etc).
I cannot see a use-case for having a public field, that is not a const. If there is a use-case for a public field, it is recommended to expose it as a property instead (with a private backing field).

>Also our coding guideline says to follow PascalCasing for constant local variables and fields

yes, you are correct, and this is a **constant field**. This is also echoed via the [VS editorconfig](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/.editorconfig#L204) that we use to enforce the C# coding style.

pnp/PnpHelper.cs
abhipsaMisra: No nested componenets - there is only one level of component supported.

timtay-microsoft:
I see a lot of "TODO"s in this PR. Did you mean to address them before merging?

abhipsaMisra:
> I see a lot of "TODO"s in this PR. Did you mean to address them before merging?

Before merging, yes. Those are small things that to do with how the sample (application) behaves, and isn't pnp related logic.
The PR is still open for review, though.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,193 @@
...
+                while (true)
+                {
+                    if (temperatureReset)
+                    {
+                        // Generate a random value between 5.0°C and 45.0°C for the current temperature reading.
+                        s_temperature = Math.Round(s_random.NextDouble() * 40.0 + 5.0, 1);
+                    }
+
+                    await SendTemperatureTelemetryAsync();
+                    await SendMaxTemperatureSinceLastRebootAsync();
```
abhipsaMisra, line 70: there is no "reboot" command in `Thermostat` model, so this property seems weirdly placed. There is an ongoing discussion with Eric on this.
@bikamani , @vinagesh  FYI.

pnp/PnpHelper.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                return collection.Contains(propertyName) ? (true, (T)collection[propertyName]) : (false, default);
+            }
+
+            // If the desired property update is for a nested component, verify that the property patch received contains the pnp component identifier ("__t": "c"),
+            // and also the desired property update.
+            if (collection.Contains(componentName))
+            {
+                JObject componentProperty = collection[componentName];
+                if (componentProperty.ContainsKey(PropertyComponentIdentifierKey) && componentProperty.Value<string>(PropertyComponentIdentifierKey).Equals(PropertyComponentIdentifierValue))
```
rido-min, line 107: dont check for the __t flag. Desired Property Updates might not include the __t flag. See #7574815 for more details.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    public class Program
+    {
+        private const string ModelId = "dtmi:com:example:TemperatureController;1";
+        private const string Thermostat1 = "thermostat1";
+        private const string Thermostat2 = "thermostat2";
+        private const string SerialNumber = "SR-123456";
+
+        private static readonly string s_deviceConnectionString = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_CONNECTION_STRING");
```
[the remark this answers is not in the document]
abhipsaMisra: I've used the C# coding guidelines [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/doc/coding-style.md):
We use `_camelCase` for internal and private fields and use `readonly` where possible. Prefix instance fields with `_`, static fields with `s_` and thread static fields with `t_`. When used on static fields, `readonly` should come after `static` (i.e. `static readonly` not `readonly static`).

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,193 @@
...
+            {
+                ModelId = ModelId,
+            };
+            s_deviceClient = DeviceClient.CreateFromConnectionString(s_deviceConnectionString, TransportType.Mqtt, options);
+            s_deviceClient.SetConnectionStatusChangesHandler((status, reason) =>
+            {
+                PrintLog($"Connection status change registered - status={status}, reason={reason}");
+            });
+
+            await s_deviceClient.OpenAsync();
```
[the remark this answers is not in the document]
abhipsaMisra: No, it is not required. I had added it for testing purpose, I'll remove it.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        // Initialize the device client instance over Mqtt protocol (TCP, with fallback over Websocket), setting the ModelId into ClientOptions, and open the connection.
+        // This method also sets a connection status change callback, that will get triggered any time the device's connection status changes.
+        private static async Task InitializeDeviceClientAsync()
+        {
+            var options = new ClientOptions
+            {
+                ModelId = ModelId,
+            };
+            s_deviceClient = DeviceClient.CreateFromConnectionString(s_deviceConnectionString, TransportType.Mqtt, options);
```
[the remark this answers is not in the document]
abhipsaMisra: @rido-min I don't understand what you mean here, can you explain this a bit.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -0,0 +1,193 @@
...
+            byte[] responsePayload = Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(report));
+            return await Task.FromResult(new MethodResponse(responsePayload, (int)StatusCode.Completed));
+        }
+
+        private static (bool, T) GetPropertyFromTwin<T>(TwinCollection collection, string propertyName)
+        {
+            return collection.Contains(propertyName) ? (true, (T)collection[propertyName]) : (false, default);
+        }
+
+        private static void PrintLog(string message)
```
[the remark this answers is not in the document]
abhipsaMisra: sure, will do!

iothub/device/samples/PnpDeviceSamples/PnpDeviceSamples/SimpleThermostat/Models/SimpleThermostat.json
```diff
@@ -1,53 +0,0 @@
-﻿{
-  "@context": "dtmi:dtdl:context;2",
-  "@id": "dtmi:com:example:simplethermostat;1",
```
prmathur-microsoft, line 3: Why is the name in consistent?
simplethermostat or thermostat

prmathur-microsoft: nm

abhipsaMisra: this looks like an outdated commit, could you try refreshing

abhipsaMisra: It should be Thermostat

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -0,0 +1,327 @@
...
+        private static DeviceClient s_deviceClient;
+        private static ILogger s_logger;
+
+        // A dictionary to hold all desired property change callbacks that this pnp device should be able to handle.
+        // The key for this dictionary is the componentName.
+        private static readonly IDictionary<string, DesiredPropertyUpdateCallback> s_desiredPropertyUpdateCallbacks = new Dictionary<string, DesiredPropertyUpdateCallback>();
+
+        // Dictionary to hold the temperature updates sent over each "Thermostat" component.
+        private static readonly Dictionary<string, Dictionary<DateTimeOffset, double>> s_temperatureReadings = new Dictionary<string, Dictionary<DateTimeOffset, double>>();
+
```
prmathur-microsoft, line 43: This is very strange to me. We want a device to hold all "temperature updates" info which might be restricted by memory and other resource constraints?

prmathur-microsoft: Why couldn't we just push it as a property and let service client hold it?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -0,0 +1,327 @@
...
+        private static async Task RunSampleAsync()
+        {
+            // This sample follows the following workflow:
+            // -> Initialize device client instance.
+            // -> Set handler to receive "reboot" command - root interface.
+            // -> Set handler to receive "getMaxMinReport" command - on "Thermostat" components.
+            // -> Set handler to receive "targetTemperature" property updates from service - on "Thermostat" components.
+            // -> Send initial device info - "workingSet" over telemetry, "serialNumber" over reported property update - root interface.
+            // -> Periodically send "temperature" over telemetry - on "Thermostat" components.
+            // -> Send "maxTempSinceLastReboot" over property update, when a new max temperature is set - on "Thermostat" components.
```
prmathur-microsoft, line 76: on line 75; why not immediately?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: We need a container to hold all temperature updates, in order to calculate the avg temp since last reboot -> this is required for "getMaxMinReport" command.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: It is sent immediately after device starts, and then every 5secs.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -0,0 +1,327 @@
...
+                    {
+                        // Generate a random value between 5.0°C and 45.0°C for the current temperature reading for each "Thermostat" component.
+                        s_temperature[Thermostat1] = Math.Round(s_random.NextDouble() * 40.0 + 5.0, 1);
+                        s_temperature[Thermostat2] = Math.Round(s_random.NextDouble() * 40.0 + 5.0, 1);
+                    }
+
+                    await SendTemperatureAsync(Thermostat1);
+                    await SendTemperatureAsync(Thermostat2);
+
+                    temperatureReset = s_temperature[Thermostat1] == 0 && s_temperature[Thermostat2] == 0;
```
prmathur-microsoft, line 114: I am not following the logic behind your reset. You want to reset when temp of each thermostat turns to 0 and from your logic on line 107, 108 this value will always be between 5 and 40 ?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -0,0 +1,327 @@
...
+                    await Task.Delay(5 * 1000);
+                }
+            });
+        }
+
+        // Initialize the device client instance over Mqtt protocol (TCP, with fallback over Websocket), setting the ModelId into ClientOptions.
+        // This method also sets a connection status change callback, that will get triggered any time the device's connection status changes.
```
prmathur-microsoft, line 121: 👍 for comments in sample.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: Yes, I should have added a comment here: this is an open item right now, the `Thermostat` interface does not have a "reboot" command, but has a property for holding max temp since last reset: https://github.com/Azure/azure-iot-sdk-csharp/pull/1423#discussion_r449310913

Based on his update, I'll remove the reset logic from `Thermostat` sample,. This logic makes sence for the `TemperatureController` sample, which does have a "reboot" command.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+        // Send working set of device memory over telemetry.
+        private static async Task SendDeviceMemoryAsync()
+        {
+            string telemetryName = "workingSet";
+            long workingSet = Environment.WorkingSet / 1024;
+
+            Message msg = PnpHelper.CreateIothubMessageUtf8(telemetryName, JsonConvert.SerializeObject(workingSet));
+
```
prmathur-microsoft, line 142: Have you aligned with Other SDK's on this PnPHelper design?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
```diff
@@ -0,0 +1,327 @@
...
+        // Send device serial number over property update.
+        private static async Task SendDeviceSerialNumberAsync()
+        {
+            string propertyName = "serialNumber";
+            string propertyPatch = PnpHelper.CreatePropertyPatch(propertyName, JsonConvert.SerializeObject(SerialNumber));
+            var reportedProperties = new TwinCollection(propertyPatch);
+
+            await s_deviceClient.UpdateReportedPropertiesAsync(reportedProperties);
+            s_logger.LogDebug($"Property: Update - {{ \"{propertyName}\": \"{SerialNumber}\" }} is {StatusCode.Completed}.");
+        }
```
prmathur-microsoft, line 156: Can we also get the updated format in your logger from PnP helper? This will be useful incase it changes over time. Sample maintenance will become easy.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: I can modify this to hold a running total instead, and calculate the avg from that; but do we need something like that for a sample?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: This is in line with what C has: https://github.com/Azure/azure-iot-sdk-c/blob/jspaith_pnp_sample_on_deviceclient/iothub_client/samples/pnp/common/pnp_protocol_helpers.c
Node and python have similar logic, with the actual method invocation within the helper, instead of in the sample directly. I have synced up with them on this, and they will be fixing that.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: Do you mean - log the format of the property patches being sent? Do you want this logged, or added into the comments?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: On second thought, even a running total for average wouldn't work, since the report needs to be generated from a point in time. So we need to store the historical data somewhere. 
We cannot send it to service since reported properties will get overwritten with each update, and sending it via telemetry will require integration with a storage related service.

prmathur-microsoft: Yeah that is my concern. That storing seems to be inbuilt into Model which I don't think is a good use case. Small devices wouldn't want to store rather won't be able to store. They need to able to send data as soon as they have it ready

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
prmathur-microsoft: Why not both :)

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: Yes, in those scenarios they would need to leverage the storage capabilities of a different service.  I will add that as a comment.

pnp/PnpHelper.cs
```diff
@@ -0,0 +1,192 @@
...
+            return string.IsNullOrWhiteSpace(componentName) ? commandName : $"{componentName}*{commandName}";
+        }
+
+        /// <summary>
+        /// Create a plug and play compatible command payload.
+        /// </summary>
+        /// <param name="serializedPayload">The serialized payload, in the format defined in the DTDL interface.</param>
+        /// <returns>A plug and play compatible command payload.</returns>
+        public static string CreatePnpCommandRequestPayload(string serializedPayload)
+        {
```
prmathur-microsoft, line 159: Command request? Does this mean this will be used on Service Sample as well?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
prmathur-microsoft: If the reboot doesn't happen often (which is usually true in case of IOT devices, sample demonstrates to store all the data generated which is not what a typical PnP user would want to do.

pnp/PnpHelper.cs
abhipsaMisra: Yes, since I was using these helpers to test my sample, I added them into `PnpHelper` directly. 
I've informed @bikamani that she can use/ modify these helpers based on her findings.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
abhipsaMisra: I feel like the property formatter in `PnpHelper` is pretty descriptive in itself (it is has descriptive parameter names, and is well indented). Do you feel additional comments there are required?

pnp/PnpHelper.cs
prmathur-microsoft: 👍

iothub/device/samples/PnpDeviceSamples/TemperatureController/Program.cs
prmathur-microsoft: I'm more concerned about the logger.

PR #1425, opened 2020-07-01
fix(e2e): Adding prompt during certificate cleanup in script

vinagesh:
Adding prompt before cleaning up certs so that we can ensure we are not removing OS certs or other unnecessary certs.

### DIFF
--- a/e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
+++ b/e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
@@ -65,9 +65,35 @@ Function Connect-AzureSubscription()
 Function CleanUp-Certs()
 {
     Write-Host("`nCleaning up old certs and files that may cause conflicts.")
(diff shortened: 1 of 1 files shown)


e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -65,9 +65,35 @@ Function Connect-AzureSubscription()
...
+    $certsToDelete = $certsToDelete1 + $certsToDelete2 + $certsToDelete3
+    
+    $title = "Clenaing up certs"
+    $certsToDeleteSubjectNames = $certsToDelete | foreach-object  {$_.Subject} 
+    $certsToDeleteSubjectNames = $certsToDeleteSubjectNames -join "`n"
+    $question = "Are you sure you want to delete the following certs?`n`n$certsToDeleteSubjectNames"
+    $choices  = '&Yes', '&No'
+    $decision = $Host.UI.PromptForChoice($title, $question, $choices, 1)
+
+    if($certsToDelete.Count -ne 0)
```
drwill-ms, line 81: spacea after `if`

PR #1431, opened 2020-07-02
refactor(iot-device): Model Id changes for MQTT Connect packet

bikamani:
Tested and verified with sample


abhipsaMisra:
changes for twin?

bikamani:
> changes for twin?

I initially thought to put in a same PR but that'll be in preview only

PR #1432, opened 2020-07-02
refactor(iot-service): Model Id changes for Twin

bikamani:
Property name is now changed to "modelId"
Tested and verified with Sample.

### DIFF
--- a/shared/src/TwinJsonConverter.cs
+++ b/shared/src/TwinJsonConverter.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Azure.Devices.Shared
         private const string CloudToDeviceMessageCountTag = "cloudToDeviceMessageCount";
         private const string AuthenticationTypeTag = "authenticationType";
         private const string X509ThumbprintTag = "x509Thumbprint";
-        private const string ModelId = "digital-twin-model-id";
+        private const string ModelId = "modelId";
 
         /// <summary>
         /// Converts <see cref="Twin"/> to its equivalent Json representation.


shared/src/TwinJsonConverter.cs
```diff
@@ -39,7 +39,7 @@ public sealed class TwinJsonConverter : JsonConverter
         private const string CloudToDeviceMessageCountTag = "cloudToDeviceMessageCount";
         private const string AuthenticationTypeTag = "authenticationType";
         private const string X509ThumbprintTag = "x509Thumbprint";
-        private const string ModelId = "digital-twin-model-id";
+        private const string ModelId = "modelId";
```
abhipsaMisra, line 42: so this is `modelId` and not `model-id`?

bikamani: yes

shared/src/TwinJsonConverter.cs
```diff
@@ -39,7 +39,7 @@ public sealed class TwinJsonConverter : JsonConverter
...
 
         /// <summary>
```
prmathur-microsoft, line 44: Will there be separate PR for Device Client?

bikamani: Separate [PR ](https://github.com/Azure/azure-iot-sdk-csharp/pull/1431)because it is for master branch, merged just a minute ago.

shared/src/TwinJsonConverter.cs
prmathur-microsoft: https://github.com/Azure/azure-iot-sdk-java/pull/835/files as per this it model-id. So is it different between device client and service client?

prmathur-microsoft: Okay this resolves it. https://github.com/Azure/azure-iot-sdk-java/pull/834/files

shared/src/TwinJsonConverter.cs
prmathur-microsoft: 👍

PR #1437, opened 2020-07-07
[DO NOT MERGE] Bring master to preview

bikamani:
Cherry-picking commits since [this](https://github.com/Azure/azure-iot-sdk-csharp/commit/f1da7a2142541ba4036be0987e14bdabc88aecf5)


vinagesh:
I know you resolved conflicts on the second commit but it might be good to keep the same commit name as in master so that if we check to find if a commit is here, we know it is here.

bikamani:
> ts on the second commit but it might be good to keep the same commit name as in master so that if we check to find if a commit is here, we know it

noted, agreed.

timtay-microsoft:
Wait, why are you cherry picking? Can't you just merge master into preview?

bikamani:
> Wait, why are you cherry picking? Can't you just merge master into preview?

Yes we can do merge too, there were only 3 commits so I did cherry picking. This approach was simpler as not much conflict.

prmathur-microsoft:
> These are not all the commits that separate master and preview. Please do a full merge from master into preview, not a cherry pick of just the PnP commits

I agree we should bring in all commits into preview before release. It is fine to unblock yourselves by cherrypicking and coding samples. But finally we need master and preview to be aligned before release.

bikamani:
Not required anymore.

bikamani:
> > These are not all the commits that separate master and preview. Please do a full merge from master into preview, not a cherry pick of just the PnP commits
> 
> I agree we should bring in all commits into preview before release. It is fine to unblock yourselves by cherrypicking and coding samples. But finally we need master and preview to be aligned before release.

Agreed, it is taken care of now.

PR #1446, opened 2020-07-09

abhipsaMisra:
Commit e57354c4b02bd2e355fb73c608a90a7949665971 fixes the current issues with service client samples. 
It was necessary to bring in this commit as a part of this PR, since the master -> preview changes remove some API calls from the `PnpHelper` class, which were being referenced by these samples. This was causing the build to fail.


vinagesh:
There are not files in the changes. Why is that? If the merge brought in changes that don't exist in preview, they should show up.

PR #1449, opened 2020-07-09
refactor(e2e): Disabling digital twin e2e test project in source rather than script

timtay-microsoft:
The "-skipPnPTests" flag wasn't working, and we should investigate that later. For now though, no one runs these tests, so they should be disabled in source

### DIFF
--- a/digitaltwin/e2e/tests/Microsoft.Azure.Devices.DigitalTwin.E2ETests.csproj
+++ b/digitaltwin/e2e/tests/Microsoft.Azure.Devices.DigitalTwin.E2ETests.csproj
@@ -4,6 +4,10 @@
     <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.2</TargetFrameworks>
     <IsPackable>false</IsPackable>
+
+    <!-- Disabling this test project because we don't need to run it -->
+    <IsTestProject>false</IsTestProject>
+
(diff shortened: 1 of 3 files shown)


vsts/gatedBuild.ps1
```diff
@@ -43,17 +43,6 @@ if (IsPullRequestBuild)
 	{
 		Write-Host "Will run Iot Hub tests"
 	}	
-	
-	#if (ShouldSkipPnPTests) 
-	#{
-		#Write-Host "Will skip PnP tests"
-		Write-Host "SKIPPING ALL PNP TESTS REGARDLESS"
-		$runTestCmd += " -skipPnPTests"
```
timtay-microsoft, line 51: This wasn't working. Something to investigate later though

digitaltwin/e2e/tests/Microsoft.Azure.Devices.DigitalTwin.E2ETests.csproj
```diff
@@ -4,6 +4,10 @@
     <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.2</TargetFrameworks>
     <IsPackable>false</IsPackable>
+
+    <!-- Disabling this test project because we don't need to run it -->
+    <IsTestProject>false</IsTestProject>
```
timtay-microsoft, line 9: Since we always want to skip the PnP tests, we may as well disable them in source instead

PR #1455, opened 2020-07-10
Fix release pipeline to use local nuget packages for testing

abhipsaMisra:


vsts/gatedBuild.ps1
```diff
@@ -18,7 +18,7 @@ docker ps -a
 #Load functions used to check what, if any, e2e tests should be run
 . .\vsts\determine_tests_to_run.ps1
 
-$runTestCmd = ".\build.ps1 -clean -build -configuration DEBUG -framework $env:FRAMEWORK"
+$runTestCmd = ".\build.ps1 -clean -build -configuration DEBUG -framework $env:FRAMEWORK -nobuildBeforeTesting"
```
timtay-microsoft, line 21: Does casing matter on these flags? "-nobuild..." vs "-noBuild..."?

abhipsaMisra: Ah, I would assume that it is. I'll fix this.

abhipsaMisra:
Will be fixed in #1459

PR #1456, opened 2020-07-10
vsts: Fix release pipeline to pick local nugets

abhipsaMisra:


build.ps1
```diff
@@ -69,7 +69,8 @@ Param(
     [string] $verbosity = "q",
     [string] $framework = "*",
     [switch] $skipIotHubTests,
-    [switch] $skipDPSTests
+    [switch] $skipDPSTests,
+    [switch] $noBuildBeforeTesting
```
prmathur-microsoft, line 73: Why do we have skipDPSTests ?

abhipsaMisra: this is for the check where, if nothing changed in dps folder, we don't need to run the dps tests.
This is valid for gated build only, not nightly or release.

PR #1461, opened 2020-07-13
feat(e2e): Adding tests for pnp model Id changes.

vinagesh:
## Checklist

## Description of the changes
Adding tests for pnp model Id changes.
This has been locally verified against a hub in canary.

### DIFF
--- a/common/test/TestDevice.cs
+++ b/common/test/TestDevice.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Azure.Devices.E2ETests
 
         private static async Task<TestDevice> CreateDeviceAsync(TestDeviceType type, string prefix)
         {
-            string deviceName = prefix + Guid.NewGuid();
+            string deviceName = "E2E_" + prefix + Guid.NewGuid();
 
             // Delete existing devices named this way and create a new one.
             using var rm = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
--- a/e2e/test/PnpServiceTests.cs
+++ b/e2e/test/PnpServiceTests.cs
@@ -1,0 +1,114 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.Azure.Devices.Shared;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    /// <summary>
+    /// Test class containing all tests to be run for plug and play.
+    /// </summary>
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    [TestCategory("PlugAndPlay")]
+    public class PnpServiceTests
+    {
+        private const string DevicePrefix = "plugAndPlayDevice";
+        private const string ModulePrefix = "plugAndPlayModule";
+        private const string TestModelId = "dtmi:com:example:testModel;1";
+
+        [TestMethod]
+        public async Task DeviceTwin_Contains_ModelId()
+        {
+            // Setup
+
+            // Create a device.
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
+            // Send model ID with MQTT connect packet to make the device plug and play.
+            var options = new ClientOptions
+            {
+                ModelId = TestModelId,
+            };
+            using var deviceClient = DeviceClient.CreateFromConnectionString(testDevice.ConnectionString, Client.TransportType.Mqtt_Tcp_Only, options);
+            await deviceClient.OpenAsync().ConfigureAwait(false);
+
+            // Act
+
+            // Get device twin.
+            using var registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+            Twin twin = await registryManager.GetTwinAsync(testDevice.Device.Id).ConfigureAwait(false);
+
+            // Assert
+            twin.ModelId.Should().Be(TestModelId, "because the device was created as plug and play");
+
+            // Cleanup
+            await registryManager.RemoveDeviceAsync(testDevice.Id).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task DeviceTwin_Contains_ModelId_X509()
+        {
+            // Setup
+
+            // Create a device.
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix, TestDeviceType.X509).ConfigureAwait(false);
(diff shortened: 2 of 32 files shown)


e2e/test/PnpServiceTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    [TestCategory("PlugAndPlay")]
+    public class PnpServiceTests
+    {
+        private const string DevicePrefix = "plugAndPlayDevice";
```
abhipsaMisra, line 20: We prefix our E2E test devices with `E2E_`. 
This makes it easier for the cleanup script to delete these devices periodically: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/service/CleanUpDevicesSample/csharp_devices_list.csv

e2e/test/PnpServiceTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        private const string ModulePrefix = "plugAndPlayModule";
+        private const string TestModelId = "dtmi:com:example:testModel;1";
+
+        [TestMethod]
+        public async Task DeviceTwin_Contains_ModelId()
+        {
+            // Setup
+
+            // Create a device.
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
```
abhipsaMisra, line 30: Since these tests runs under a .net core/ netfm console app, we shouldn't be requiring these `.ConfigureAwait(false)` statements here either, right?

e2e/test/PnpServiceTests.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    [TestCategory("PlugAndPlay")]
+    public class PnpServiceTests
+    {
+        private const string DevicePrefix = "plugAndPlayDevice";
+        private const string ModulePrefix = "plugAndPlayModule";
+        private const string TestModelId = "dtmi:com:example:testModel;1";
```
abhipsaMisra, line 22: this is a placeholder model Id (it's contents don't matter to us), maybe we should call it such?

e2e/test/PnpServiceTests.cs
vinagesh: On the other hand, I'm inclined to add delete device. Does remove device just remove the registration or delete it as well?

e2e/test/PnpServiceTests.cs
vinagesh: We don't need it but as this is not a sample that we will share, I am adding it for consistency. We've followed that in other E2E tests as well.

e2e/test/PnpServiceTests.cs
vinagesh: Hence I called it testModel to suggest it is a placeholder :) If it seems very unclear we can change it later.

e2e/test/PnpServiceTests.cs
abhipsaMisra: Ah, you want to handle the create - test - delete within the test directly? I don't see any issue with this, other than if this adds considerable test run duration because of the  addition of the "deletion" operation.

>Does remove device just remove the registration or delete it as well?

I am assuming you mean removing the device registration from IoT hub identity registry. If yes, then "removing the device registration from IoT hub identity registry" = "delete the device from IoT Hub".

vinagesh: However, it is possible the test fails so I will add the prefix to the helper class so that we always create a device with the right prefix.

common/test/TestDevice.cs
```diff
@@ -59,7 +59,7 @@ public static async Task<TestDevice> GetTestDeviceAsync(string namePrefix, TestD
 
         private static async Task<TestDevice> CreateDeviceAsync(TestDeviceType type, string prefix)
         {
-            string deviceName = prefix + Guid.NewGuid();
+            string deviceName = "E2E_" + prefix + Guid.NewGuid();
```
abhipsaMisra, line 62: you will need to modify other test classes that add this prefix explicitly, else, the test devices will now be called `E2E_E2E_MessageReceiveFaultInjectionTests_Sasl_dfcb23e9-f224-4fd1-8f70-7f55598d9941`.
This is in no way a blocker for tests, but instead just looks very very weird.

vinagesh: Yes I know. I thought it should be ok to have a double prefix rather than fixing all files. We will be future proof though.

vinagesh: I am looking at how many files though. I will be updating if they are just a few.

e2e/test/PnpServiceTests.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.Azure.Devices.Shared;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    [TestCategory("PlugAndPlay")]
+    public class PnpServiceTests
```
bikamani, line 18: nit: may be add a comment for this class.

e2e/test/PnpServiceTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            // Setup
+
+            // Create a module.
+            TestModule testModule = await TestModule.GetTestModuleAsync(DevicePrefix, ModulePrefix).ConfigureAwait(false);
+            // Send model ID with MQTT connect packet to make the module plug and play.
+            var options = new ClientOptions
+            {
+                ModelId = TestModelId,
+            };
+            using var moduleClient = ModuleClient.CreateFromConnectionString(testModule.ConnectionString, Client.TransportType.Mqtt_Tcp_Only, options);
```
bikamani, line 64: is there way to not hard code this?

vinagesh: I have not found a way but we can extend this test in the future if it is needed. For now it anyway only works with MQTT.

e2e/test/PnpServiceTests.cs
vinagesh: I've changed wherever it was straightforward. We can keep fixing anything that looks ugly when we run into it. As it does not affect our tests in any way, should be fine for now.

e2e/test/PnpServiceTests.cs
```diff
@@ -0,0 +1,83 @@
...
+            // Setup
+
+            // Create a device.
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
+            // Send model ID with MQTT connect packet to make the device plug and play.
+            var options = new ClientOptions
+            {
+                ModelId = TestModelId,
+            };
+            using var deviceClient = DeviceClient.CreateFromConnectionString(testDevice.ConnectionString, Client.TransportType.Mqtt_Tcp_Only, options);
```
bikamani, line 39: hey, I am adding test case for x509 auth as well in addition to this, just talked to Tim about it that we should test that flow as well.

vinagesh: Sure. In this specific case it probably makes sense though it would be testing their SDK.

PR #1467, opened 2020-07-15
feat(iot-device): Implement model Id support over Amqp

abhipsaMisra:
From #1397 :

Currently Edge does not have any support for Plug and Play. This change adds the necessary Id, the Device Capability Model Id (referred to as "model-id").

### DIFF
--- a/iothub/device/src/Message.cs
+++ b/iothub/device/src/Message.cs
@@ -255,6 +255,16 @@ namespace Microsoft.Azure.Devices.Client
         }
 
         /// <summary>
+        /// The DTDL component name from where the telemetry message has originated.
+        /// This is relevant only for plug and play certified devices.
+        /// </summary>
+        public string ComponentName
+        {
+            get => GetSystemProperty<string>(MessageSystemPropertyNames.ComponentName);
+            set => SystemProperties[MessageSystemPropertyNames.ComponentName] = value;
+        }
+
+        /// <summary>
         /// Gets the dictionary of user properties which are set when user send the data.
         /// </summary>
         public IDictionary<string, string> Properties { get; private set; }
--- a/iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
+++ b/iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
@@ -31,22 +31,22 @@ namespace Microsoft.Azure.Devices.Client.Transport.Amqp
         }
 
         public AmqpUnit CreateAmqpUnit(
-            DeviceIdentity deviceIdentity, 
-            Func<MethodRequestInternal, Task> methodHandler, 
-            Action<Twin, string, TwinCollection> twinMessageListener, 
+            DeviceIdentity deviceIdentity,
+            Func<MethodRequestInternal, Task> methodHandler,
+            Action<Twin, string, TwinCollection> twinMessageListener,
             Func<string, Message, Task> eventListener,
             Action onUnitDisconnected)
         {
             if (Logging.IsEnabled) Logging.Enter(this, deviceIdentity, $"{nameof(CreateAmqpUnit)}");
 
-            AmqpUnit amqpUnit = new AmqpUnit(
-                deviceIdentity, 
+            var amqpUnit = new AmqpUnit(
+                deviceIdentity,
                 this,
                 methodHandler,
-                twinMessageListener, 
+                twinMessageListener,
                 eventListener,
                 onUnitDisconnected);
-            lock(_unitsLock)
+            lock (_unitsLock)
             {
                 _amqpUnits.Add(amqpUnit);
             }
--- a/iothub/device/src/Transport/AmqpIoT/AmqpIoTConstants.cs
+++ b/iothub/device/src/Transport/AmqpIoT/AmqpIoTConstants.cs
@@ -40,11 +40,27 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
         internal const string MethodName = "IoThub-methodname";
         internal const string Status = "IoThub-status";
 
+        // The tracking Id identifier, for an Amqp Error returned by the service.
         public static readonly AmqpSymbol TrackingId = Vendor + ":tracking-id";
+
+        // The client version identifier, added to Amqp link settings, while opening a sending link.
+        // The client version is the product info of the device, for which the connection is being established.
         public static readonly AmqpSymbol ClientVersion = Vendor + ":client-version";
+
+        // The API version identifier, added to Amqp link settings, while opening a sending link.
+        // The API version identifies which version of the service API the connection is targeted towards.
         public static readonly AmqpSymbol ApiVersion = Vendor + ":api-version";
+
+        // The correlation Id identifier, added to Amqp link settings, while opening a sending link.
+        // The correlation Id is a guid prefixed with either "methods" or "twin", identifying the target of the sending link.
         public static readonly AmqpSymbol ChannelCorrelationId = Vendor + ":channel-correlation-id";
+
+        // The authentication chain identifier, added to Amqp link settings, while opening a sending link.
+        // The authentication chain is required for enabling nested Edge scenarios.
         public static readonly AmqpSymbol AuthChain = Vendor + ":auth-chain";
+
+        // The digital twin model Id identifier, added to Amqp link settings, while opening a sending link.
+        // The digital twin model Id is required to be sent over the sending link, in order for the service to identify it as a PnP-enabled device.
         public static readonly AmqpSymbol ModelId = Vendor + ":model-id";
 
         internal const string IotHubSasTokenType = CbsConstants.IotHubSasTokenType;
--- a/iothub/device/src/Transport/AmqpIoT/AmqpIoTConstants.cs
+++ b/iothub/device/src/Transport/AmqpIoT/AmqpIoTConstants.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.Azure.Amqp;
+using Microsoft.Azure.Amqp.Encoding;
 using Microsoft.Azure.Amqp.Framing;
 
 namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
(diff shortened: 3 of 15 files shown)


abhipsaMisra:
End to end tests for this will be added into the preview code. This is because, the only way to test this right now is to verify if the returned `Twin` object contains the specified model Id. This functionality is currently available only on preview.

abhipsaMisra:
This has been manually tested with Azure IoT explorer, to work for single device and muxed scenarios. It has also been tested for iot hub module (create from connection string flow).

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -31,22 +31,22 @@ public AmqpConnectionHolder(DeviceIdentity deviceIdentity)
...
+            Action<Twin, string, TwinCollection> twinMessageListener,
             Func<string, Message, Task> eventListener,
             Action onUnitDisconnected)
         {
             if (Logging.IsEnabled) Logging.Enter(this, deviceIdentity, $"{nameof(CreateAmqpUnit)}");
 
-            AmqpUnit amqpUnit = new AmqpUnit(
-                deviceIdentity, 
+            var amqpUnit = new AmqpUnit(
+                deviceIdentity,
```
prmathur-microsoft, line 43: why var?

abhipsaMisra: we do a `new AmqpUnit()` here, so the type can be automatically inferred.

iothub/device/src/Transport/AmqpIoT/AmqpIoTConstants.cs
```diff
@@ -39,6 +40,13 @@ internal static class AmqpIoTConstants
         internal const string MethodName = "IoThub-methodname";
         internal const string Status = "IoThub-status";
 
+        public static readonly AmqpSymbol TrackingId = Vendor + ":tracking-id";
+        public static readonly AmqpSymbol ClientVersion = Vendor + ":client-version";
+        public static readonly AmqpSymbol ApiVersion = Vendor + ":api-version";
+        public static readonly AmqpSymbol ChannelCorrelationId = Vendor + ":channel-correlation-id";
+        public static readonly AmqpSymbol AuthChain = Vendor + ":auth-chain";
```
prmathur-microsoft, line 47: What are line 46 and 47 for? Actually can you explain each one of these with a comment?

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -30,14 +30,6 @@ internal static class AmqpIoTErrorAdapter
...
         public static readonly AmqpSymbol IotHubSuspended = AmqpIoTConstants.Vendor + ":iot-hub-suspended";
 
-        public static readonly AmqpSymbol TrackingId = AmqpIoTConstants.Vendor + ":tracking-id";
-        public static readonly AmqpSymbol ClientVersion = AmqpIoTConstants.Vendor + ":client-version";
-        public static readonly AmqpSymbol ApiVersion = AmqpIoTConstants.Vendor + ":api-version";
-        public static readonly AmqpSymbol ChannelCorrelationId = AmqpIoTConstants.Vendor + ":channel-correlation-id";
-        public static readonly AmqpSymbol AuthChain = AmqpIoTConstants.Vendor + ":auth-chain";
-
-        private const int MaxSizeInInfoMap = 32 * 1024;
-
```
prmathur-microsoft, line 40: is line 39 not needed anymore? if so why?

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -205,7 +197,7 @@ public static Exception ToIotHubClientContract(Error error)
             string message = error.Description;
 
             string trackingId = null;
-            if (error.Info != null && error.Info.TryGetValue(TrackingId, out trackingId))
+            if (error.Info != null && error.Info.TryGetValue(AmqpIoTConstants.TrackingId, out trackingId))
             {
```
prmathur-microsoft, line 201: would "AmqpIoTConstants.TrackingId" always be a valid value?

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
abhipsaMisra: This was required in one of our older "convert amqp outcome to iot hub exception" methods. That method is no longer in use, so this variable was simply lying around, and not being referenced anywhere.

abhipsaMisra: We have updated implementation for "convert amqp outcome to iot hub exception" now, one that no longer uses this `MaxSizeInInfoMap`.

iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
```diff
@@ -203,12 +203,14 @@ TimeSpan timeout
...
             string linkSuffix,
-            string CorrelationId,
+            string correlationId,
             TimeSpan timeout
         )
         {
             if (Logging.IsEnabled) Logging.Enter(typeof(AmqpIoTSession), deviceIdentity, $"{nameof(OpenSendingAmqpLinkAsync)}");
 
+            string serviceApiVersion = ClientApiVersionHelper.ApiVersionString;
+
```
prmathur-microsoft, line 213: How would you set this to preview api version? Does ClientApiVersionHelper give you the right service api version?

iothub/device/src/Transport/AmqpIoT/AmqpIoTConstants.cs
abhipsaMisra: sure, I'll add some comments for each of these.

iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
prmathur-microsoft: nm.. seeing it later

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
abhipsaMisra: It is a const `com.microsoft:tracking-id`.

iothub/device/src/Transport/AmqpIoT/AmqpIoTSession.cs
abhipsaMisra: it is being set to the preview API version below, based on the check if we have model Id set into client options.

iothub/device/src/Transport/DeviceIdentity.cs
```diff
@@ -20,11 +20,15 @@ internal class DeviceIdentity
         internal ProductInfo ProductInfo { get; }
         internal AuthenticationModel AuthenticationModel { get; }
         internal string Audience { get; }
-        internal DeviceIdentity(IotHubConnectionString iotHubConnectionString, AmqpTransportSettings amqpTransportSettings, ProductInfo productInfo)
+        internal ClientOptions Options { get; }
+
+        internal DeviceIdentity(IotHubConnectionString iotHubConnectionString, AmqpTransportSettings amqpTransportSettings, ProductInfo productInfo, ClientOptions options)
         {
```
prmathur-microsoft, line 26: I wish product info can be a part of client options evry where. I know that is a breaking change; but is it possible to make it so in implementation except for public API?

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -1032,6 +1032,7 @@ private static class IotHubWirePropertyNames
             public const string MqttDiagIdKey = "$.diagid";
             public const string MqttDiagCorrelationContextKey = "$.diagctx";
             public const string InterfaceId = "$.ifid";
+            public const string ComponentName = "$.sub";
         }
```
prmathur-microsoft, line 1036: If Message.ComponentName is set for MQTT and not via application property; would it still be set on topic? I don't see that code anywhere.

abhipsaMisra: Yes, it would happen here: https://github.com/Azure/azure-iot-sdk-csharp/blob/081a3ecaa49ecc88c88b9d803fcc10dfd5ab5a83/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs#L1194

We do not need to map each system property explicitly for mqtt, it merges all your application and system properties, and appends them to the topic.

I have also tested this functionality over Mqtt, and the component name is being set as expected.

abhipsaMisra:
> CIL; especially Message.ComponentName translating to MQTT as well.
> Missing e2e ? Will it be a separate PR?

I've answered your questions @prmathur-microsoft .
E2E will be added in a separate PR (it will be added into preview code"

iothub/device/src/Transport/DeviceIdentity.cs
abhipsaMisra: Yeah, we could. 

What do you think about marking `DeviceClient.ProductInfo` as deprecated, and adding `ProductInfo` into `ClientOptions`?

prmathur-microsoft: I would love that but I think we should hold off on breaking changes at this point. So just internal implementation should be good for now

abhipsaMisra: This requires a bit of refactoring, I'll add this in a subsequent PR.

iothub/device/src/Message.cs
```diff
@@ -254,6 +254,16 @@ public string ContentEncoding
             set => SystemProperties[MessageSystemPropertyNames.ContentEncoding] = value;
         }
 
+        /// <summary>
+        /// The digital twin component name, from where the telemetry message has originated.
+        /// This is relevant only for digital twin enabled devices.
```
bikamani, line 259: Is this wording right? "digital twin enabled"? Or certified or may be there's better one

abhipsaMisra: We have been calling them pnp-enabled devices, but yes, I do like "certified".

iothub/device/src/Transport/AmqpIoT/AmqpIoTConstants.cs
```diff
@@ -39,6 +40,29 @@ internal static class AmqpIoTConstants
         internal const string MethodName = "IoThub-methodname";
         internal const string Status = "IoThub-status";
 
+        // The tracking Id identifier, for an Amqp Error returned by the service.
+        public static readonly AmqpSymbol TrackingId = Vendor + ":tracking-id";
```
bikamani, line 44: how will this id be used? Is this introduced as a new field now?

abhipsaMisra: No, this has simply been moved over from `AmqpIoTErrorAdapter.TrackingId ` to `AmqpIoTConstants.TrackingId`.

PR #1470, opened 2020-07-15
fix(vsts): Remove rolling builds for master branch

abhipsaMisra:
Updating description:

Right now, when any change gets merged into preview branch, we trigger 2 rolling builds on our Canary pipeline; one that runs against "master" sourcecode, and another that runs against "preview" sourcecode.
Since we don't want to test the "master" sourcecode in Canary environment, this change specifically removes the "master" rolling build trigger for "preview" branch.

### DIFF
--- a/vsts/vsts.yaml
+++ b/vsts/vsts.yaml
@@ -4,7 +4,6 @@ trigger:
   batch: true
   branches:
     include:
-    - master
     - preview
   paths:
     exclude:


vsts/vsts.yaml
```diff
@@ -4,7 +4,6 @@ trigger:
   batch: true
   branches:
     include:
-    - master
```
drwill-ms, line 7: why?

vsts/vsts.yaml
```diff
@@ -4,7 +4,6 @@ trigger:
...
     - preview
```
bikamani, line 7: +1 on why are we making this change?

vsts/vsts.yaml
abhipsaMisra: Right now, when any change gets merged into preview branch, we trigger 2 rolling builds on our Canary pipeline; one that runs against "master" sourcecode, and another that runs against "preview" sourcecode.
Since we don't want to test the "master" sourcecode in Canary environment, this change specifically removes the "master" rolling build trigger for "preview" branch.

drwill-ms: Is there a way for us to have 2 independent triggers and runs? I don't want to lose the master integration testing just because we're mostly working in a preview branch right now.

abhipsaMisra: I didn't modify the yaml getting picked up from the master branch (for prod pipeline): https://github.com/Azure/azure-iot-sdk-csharp/blob/master/vsts/vsts.yaml

So this change should in no way affect the rolling builds that we have set up on West Central US, to run against `master` sourcecode.
If it does affect that, I will revert this change.

PR #1471, opened 2020-07-15
fix(e2e): Wait longer to avoid throttling exceptions with v2 jobs.

vinagesh:
## Checklist

This fix should mitigate pipeline errors till we find a better solution.

### DIFF
--- a/e2e/test/IoTHubServiceProxyE2ETests.cs
+++ b/e2e/test/IoTHubServiceProxyE2ETests.cs
@@ -7,6 +7,7 @@ using System.Net;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Common.Exceptions;
 using Microsoft.Azure.Devices.Shared;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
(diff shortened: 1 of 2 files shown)


e2e/test/IoTHubServiceProxyE2ETests.cs
```diff
@@ -89,17 +92,30 @@ private async Task RegistryManager_AddDevice(HttpTransportSettings httpTransport
...
+                        string jobId = "JOBSAMPLE" + Guid.NewGuid().ToString();
+                        string query = $"DeviceId IN ['{JobDeviceId}']";
+                        JobResponse createJobResponse = await jobClient.ScheduleTwinUpdateAsync(jobId, query, twin, DateTime.UtcNow, (long)TimeSpan.FromMinutes(2).TotalSeconds).ConfigureAwait(false);
+                        break;
+                    }
+                    // Concurrent jobs can be rejected, so implement a retry mechanism to handle conflicts with other tests
+                    catch (ThrottlingException) when (++tryCount < MaxIterationWait)
+                    {
+                        _log.WriteLine($"ThrottlingException... waiting.");
+                        await Task.Delay(_waitDuration).ConfigureAwait(false);
```
bikamani, line 115: so basically with adding a delay of 5 second before retry, we do not see throttling exception?

abhipsaMisra:
Are we seeing quota exceeded errors for non-import/export jobs as well? afaik these jobs don't have as strict a limitation as import/export jobs (which have a limit of 1 running job, at any given time).

e2e/test/IoTHubServiceProxyE2ETests.cs
vinagesh: No it would probably take more than that. About 30-40 seconds. We will have about 5 jobs running in parallel so depends on the contention.

vinagesh:
> Are we seeing quota exceeded errors for non-import/export jobs as well? afaik these jobs don't have as strict a limitation as import/export jobs (which have a limit of 1 running job, at any given time).

yes according to my testing and error message, the limit is 1 for V2 jobs as well.

PR #1472, opened 2020-07-16
Add logging of SQMMachineId for Linux hosts

SlavyMihov:
Add logging of _SQMMachineId_ (aka known as _LocalDeviceId_) for Linux hosts.
Currently, _SQMMachineId_ is being included in the _UserAgent_ string for Windows hosts. This change is intended to bring Linux on par.

### DIFF
--- a/iothub/device/src/TelemetryMethods.cs
+++ b/iothub/device/src/TelemetryMethods.cs
@@ -6,6 +6,10 @@ using Microsoft.Win32;
 using System;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
+using System.IO;
+using System.Text;
+using System.Security.Cryptography;
+using System.Globalization;
 
 namespace Microsoft.Azure.Devices.Client
 {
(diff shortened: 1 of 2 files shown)


iothub/device/src/TelemetryMethods.cs
```diff
@@ -29,8 +33,77 @@ public static string GetSqmMachineId()
                     if (Logging.IsEnabled) Logging.Error(null, ex, nameof(TelemetryMethods));
                 }
             }
+            else
+            {
+                try
+                {
+                    // get HMAC-SHA256 of the ADHS application ID, keyed by the machine ID
+                    using var sr = new StreamReader("/etc/machine-id");
```
prmathur-microsoft, line 41: What would this resolve to on a non-linux machine?

iothub/device/src/TelemetryMethods.cs
```diff
@@ -29,8 +33,77 @@ public static string GetSqmMachineId()
...
+                    string key = sr.ReadLine();
+                    if (!string.IsNullOrEmpty(key))
+                    {
+                        // ADHS application ID = 586e0ed989c946a794e5de337644d308
+                        return GetUUIDv4("586e0ed989c946a794e5de337644d308", key);
```
prmathur-microsoft, line 46: Can you also point to public doc describing this ID?

iothub/device/src/TelemetryMethods.cs
prmathur-microsoft: perhaps consider adding check that this else block is valid only for linux ?

iothub/device/tests/TelemetryMethodsTests.cs
```diff
@@ -31,8 +33,25 @@ public void GetSqmMachineId_ReturnsExpectedValue()
             }
             else
             {
-                // GetSqmMachineId() should always return null for all other platforms
-                Assert.IsNull(actualValue);
+                if (File.Exists("/etc/machine-id"))
+                {
+                    // check if actualValue is a valid OSF v4 UUID
+                    // https://www.freedesktop.org/software/systemd/man/machine-id.html
```
prmathur-microsoft, line 39: I don't follow why is this check a valid test for the SDK?

iothub/device/src/TelemetryMethods.cs
SlavyMihov: The ADHS application ID = 586e0ed989c946a794e5de337644d308 is a hardcoded value ([here](https://mscodehub.visualstudio.com/Azile/_git/azure-device-health-services?path=%2Fdevice-utils%2Fsrc%2FIdentity.cpp&_a=contents&version=GBmaster)) used by the ADHS client to derive a "device id" with which all telemetry data from a device is tamped with. The "device id" is practically a hash (HMAC-SHA256) of the ADHS application ID, keyed by the machine ID. The latter is a secret that should never be exposed.
ADHS is expected to go public as open-source in the next semester (Cobalt).

iothub/device/tests/TelemetryMethodsTests.cs
SlavyMihov: The _SQMMachineId_ generated by the SDK needs to be a valid OSF v4 UUID. The machine-id [documentation](https://www.freedesktop.org/software/systemd/man/machine-id.html#Relation%20to%20OSF%20UUIDs) calls out the bitwise operations that decode the ID into a valid OSF v4 UUID.
This is what is being tested here - if the _SQMMachineId_ generated by the SDK has the right bits set.

iothub/device/src/TelemetryMethods.cs
SlavyMihov: Added explicit check for Linux.

iothub/device/src/TelemetryMethods.cs
prmathur-microsoft: I'm concerned that without public documentation; this would be something hard to explain/maintain as we are open source. Also I don't see "The "device id" is practically a hash (HMAC-SHA256) of the ADHS application ID, keyed by the machine ID. The latter is a secret that should never be exposed." as an iot SDK concept. So why is this sitting in SDK?

SlavyMihov:
Closing for now, until the ADHS story evolves and future plans clarify.

PR #1474, opened 2020-07-17
feat(e2e): Log to AI for e2e tests.

vinagesh:
## Checklist

## Description of the changes
This change is the first part to start logging to Application Insights. I have not done any renaming of classes/methods in the PR as I wanted to keep the review short. Will make necessary changes in following PRs.

Logging from our E2E tests to Application Insights.
Logging the following properties:
Unique TestRunId
BuildId - for pipeline
TestName
ClassName(calller)
TestResult
TestRunTime
TargetFramework
Language
Service

Currently only two classes are on-boarded to this. I have added a task to onboard every other class. There is also the issue of all helpers initializing their own loggers. In a following PR I will make changes to pass logger from the tests to the helpers instead so that the logs are correctly correlated.

image

### DIFF
--- a/common/test/TestLogging.cs
+++ b/common/test/TestLogging.cs
@@ -2,39 +2,122 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Diagnostics;
-using System.Globalization;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.Versioning;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Azure.Devices.E2ETests
 {
+    /// <summary>
+    /// This class logs to multiple providers - Event source, Application insights
+    /// </summary>
     public class TestLogging
     {
-        private readonly static TestLogging s_instance = new TestLogging();
+        public const string Language = "C#";
+        public const string Service = "IotHub";
+        private static object _initLock = new object();
+
+        // Client to log to application insights
+        private static TelemetryClient _telemetryClient;
+
+        // Unique ID for all tests of a run.
+        private static string _testRunId;
+
+        public TestContext Context { get; set; }
+        public string TargetFramework { get; set; }
 
         private TestLogging()
         {
+            // Thread-safe initialization of the telemetry client and run Id.
+            lock (_initLock)
+            {
+                if (_telemetryClient == null)
+                {
+                    // Instrumentation key is used to connect to Application Insights instance.
+                    // The value is copied from the portal and stored in KeyVault.
+                    // The E2E tests script will load it as an environment variable and the pipelines have a task to do the same.
+                    string intrumentationKey = Environment.GetEnvironmentVariable("E2E_IKEY");
+                    if (!string.IsNullOrWhiteSpace(intrumentationKey))
+                    {
+                        var config = new TelemetryConfiguration
+                        {
+                            InstrumentationKey = intrumentationKey,
+                        };
+                        _telemetryClient = new TelemetryClient(config);
+                        _testRunId = Guid.NewGuid().ToString();
+                    }
+                }
+            }
         }
 
         private const string NullInstance = "(null)";
-        
+
         public static string IdOf(object value) => value != null ? value.GetType().Name + "#" + GetHashCode(value) : NullInstance;
 
         public static int GetHashCode(object value) => value?.GetHashCode() ?? 0;
 
-        public static TestLogging GetInstance()
+        public static TestLogging GetInstance(TestContext testContext = null)
         {
-            return s_instance;
+            var logger = new TestLogging();
+
+            if (_telemetryClient != null)
+            {
+                // Set logger properties for the current test.
+                var targetFramework = (TargetFrameworkAttribute)Assembly
+                    .GetExecutingAssembly()
+                    .GetCustomAttributes(typeof(TargetFrameworkAttribute), false)
+                    .SingleOrDefault();
+                logger.Context = testContext;
+                logger.TargetFramework = targetFramework.FrameworkName;
+            };
+
+            return logger;
         }
 
-        public void WriteLine(string message)
+        public void WriteLine(string message, SeverityLevel severity = SeverityLevel.Information, IDictionary<string, string> propertyBag = null, [CallerMemberName] string caller = null)
         {
+            // Log to event source
             EventSourceTestLogging.Log.TestMessage(message);
+
+            // Log to Application insights
+            if (_telemetryClient != null)
+            {
+                if (propertyBag == null)
+                {
+                    propertyBag = new Dictionary<string, string>();
+                }
+
+                // Add common properties
+                propertyBag.Add(LoggingPropertyNames.TestRunId, _testRunId);
+                propertyBag.Add(LoggingPropertyNames.TestFramework, TargetFramework);
+                propertyBag.Add(LoggingPropertyNames.Language, Language);
+                propertyBag.Add(LoggingPropertyNames.Service, Service);
+                if (Context != null)
+                {
+                    propertyBag.Add(LoggingPropertyNames.TestName, Context.TestName);
+                    propertyBag.Add(LoggingPropertyNames.ClassName, Context.FullyQualifiedTestClassName ?? caller);
+                    propertyBag.Add(LoggingPropertyNames.TestStatus, Context?.CurrentTestOutcome.ToString());
+                }
+
+                // Environment Variable set in Azure pipelines to correlate logs with the build number.
+                propertyBag.Add("BuildId", Environment.GetEnvironmentVariable("BUILD_BUILDID"));
+
+                _telemetryClient.TrackTrace(message, severity, propertyBag);
+            }
         }
 
-        public void WriteLine(string format, params object[] args)
+        public void Flush()
         {
-            string message = string.Format(CultureInfo.InvariantCulture, format, args);
-            EventSourceTestLogging.Log.TestMessage(message);
+            if (_telemetryClient != null)
+            {
+                _telemetryClient.Flush();
+            }
         }
     }
 }
--- a/e2e/test/BulkOperationsE2ETests.cs
+++ b/e2e/test/BulkOperationsE2ETests.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Azure.Devices.E2ETests
     [TestClass]
     [TestCategory("E2E")]
     [TestCategory("IoTHub")]
-    public class BulkOperationsE2ETests : IDisposable
+    public class BulkOperationsE2ETests : E2ETestBase, IDisposable
     {
         private readonly string DevicePrefix = $"E2E_{nameof(BulkOperationsE2ETests)}_";
         private readonly string ModulePrefix = $"E2E_{nameof(BulkOperationsE2ETests)}_";
--- a/e2e/test/E2ETestBase.cs
+++ b/e2e/test/E2ETestBase.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    /// <summary>
+    /// This class creates an instance of the logger for each test and logs the test result along with other useful information.
+    /// </summary>
+    public class E2ETestBase
+    {
+        protected TestLogging Logger { get; set; }
+        private Stopwatch Stopwatch { get; set; }
+        public TestContext TestContext { get; set; }
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            Stopwatch = Stopwatch.StartNew();
+            Logger = TestLogging.GetInstance(TestContext);
+            Logger.WriteLine($"Starting test - {TestContext.TestName}", SeverityLevel.Information);
+        }
+
+        [TestCleanup]
+        public void TestCleanup()
+        {
+            Stopwatch.Stop();
+            double elapsed = Stopwatch.Elapsed.TotalSeconds;
+            var properties = new Dictionary<string, string>
+            {
+                { LoggingPropertyNames.SecondsElapsed, elapsed.ToString() }
+            };
+            Logger.WriteLine($"Finished test - {TestContext.TestName}", SeverityLevel.Information, properties);
+            // As this is not an application that keeps running, explicitly flushing is required to ensure we do not lose any logs.
+            // The recommendation from AI is to wait for 5 seconds after flushing.
+            Logger.Flush();
+            Thread.Sleep(5000);
+        }
+    }
+}
--- a/e2e/test/CombinedClientOperationsPoolAmqpTests.cs
+++ b/e2e/test/CombinedClientOperationsPoolAmqpTests.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Azure.Devices.E2ETests
     [TestClass]
     [TestCategory("E2E")]
     [TestCategory("IoTHub")]
-    public class CombinedClientOperationsPoolAmqpTests : IDisposable
+    public class CombinedClientOperationsPoolAmqpTests : E2ETestBase, IDisposable
     {
         private const string MethodName = "MethodE2ECombinedOperationsTest";
         private readonly string _devicePrefix = $"E2E_{nameof(CombinedClientOperationsPoolAmqpTests)}_";
--- a/e2e/test/E2ETests.csproj
+++ b/e2e/test/E2ETests.csproj
@@ -79,6 +79,7 @@
   <!-- All Platforms -->
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.10.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.14.0" />
     <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="5.2.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
     <PackageReference Include="MSTest.TestAdapter" Version="2.0.0" />
--- a/e2e/test/LoggingPropertyNames.cs
+++ b/e2e/test/LoggingPropertyNames.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    /// <summary>
+    /// The property names to log to telemetry.
+    /// </summary>
+    public class LoggingPropertyNames
+    {
+        public const string TestRunId = "TestRunId";
+        public const string TestName = "TestName";
+        public const string ClassName = "ClassName";
+        public const string TestFramework = "TestFramework";
+        public const string TestStatus = "TestStatus";
+        public const string SecondsElapsed = "SecondsElapsed";
+        public const string Language = "Language";
+        public const string Service = "Service";
+    }
+}


e2e/test/BulkOperationsE2ETests.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Azure.Devices.E2ETests
     [TestClass]
     [TestCategory("E2E")]
     [TestCategory("IoTHub")]
-    public class BulkOperationsE2ETests : IDisposable
+    public class BulkOperationsE2ETests : E2ETestBase, IDisposable
```
vinagesh, line 15: Currently only two classes are on-boarded to this. I have added a task to onboard every other class.

e2e/test/E2ETestBase.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    /// <summary>
+    /// This class creates an instance of the logger for each test and logs the test result along with other useful information.
```
abhipsaMisra, line 13: One issue I had faced in the past, in parallelizing EventSource trace generation, is that the SDK writes all of it's events to the same EventSource providers (for tests running in parallel). The test logging, in that case, was unable to differentiate between the SDK log generated for test 1 vs test 2.

If I understand correctly, since we can now forward these traces to application insights, we should be able to add additional identifiers to each of the log statements being forwarded, and that would help us filter out SDK logs for test 1 vs SDK logs for test 2. Is that correct?

abhipsaMisra: This is, of course, with the assumption that test 1 and test 2 are both running in parallel, under the same process.

e2e/test/TelemetryPropertyNames.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    /// <summary>
+    /// The property names to log to telemetry.
+    /// </summary>
+    public class TelemetryPropertyNames
```
abhipsaMisra, line 12: I would rename this to TestLoggingPropertyNames or something, something that doesn't use the word "telemetry".
"Telemetry" kind-of has an established meaning in the scope of Azure IoT Hub.

abhipsaMisra:
Can we attach a snapshot of what this looks like on AI, just for reference?

e2e/test/E2ETestBase.cs
vinagesh: yes we are doing that in AI right now. I'm not very sure how EventSource works.

vinagesh: FYI - These additional things are being added as properties, not in the message itself. It just makes querying easier. But we can always add all that information to the message as well.

e2e/test/TelemetryPropertyNames.cs
vinagesh: Application Insights calls it telemetry and hence I named it that. It is not a customer facing API so I'm sort of on the fence. This is test telemetry just like IotHub telemetry. Let me think about it.

e2e/test/E2ETestBase.cs
vinagesh: image

e2e/test/TelemetryPropertyNames.cs
abhipsaMisra: True, I meant it might be confusing for us. When I first looked at the class, I was confused as to what it stood for.
How about `ApplicationInsightsTelemetryPropertyNames`?

vinagesh: We can always use different telemetry providers so a generic name is better. Ex: We can log to AI. Kusto etc... We can always change implementation but we should not be changing class names. Maybe LoggerPropertyNames/LoggingPropertyNames?

abhipsaMisra: Yup, like that too: LoggingPropertyNames

common/test/TestLogging.cs
```diff
@@ -2,39 +2,114 @@
...
+            {
+                return;
+            }
+            else
+            {
+                // Thread-safe initialization of the telemetry client and run Id.
+                lock (initLock)
+                {
+                    string iKey = Environment.GetEnvironmentVariable("E2E_IKEY");
+                    if (!string.IsNullOrWhiteSpace(iKey))
```
abhipsaMisra, line 45: maybe log an exception if the key isn't set properly?

common/test/TestLogging.cs
```diff
@@ -2,39 +2,114 @@
...
+                // Add common properties
+                propertyBag.Add(TelemetryPropertyNames.TestRunId, _testRunId);
+                propertyBag.Add(TelemetryPropertyNames.TestName, Context?.TestName);
+                propertyBag.Add(TelemetryPropertyNames.ClassName, Context?.FullyQualifiedTestClassName ?? caller);
+                propertyBag.Add(TelemetryPropertyNames.TestFramework, TargetFramework);
+                propertyBag.Add(TelemetryPropertyNames.TestStatus, Context?.CurrentTestOutcome.ToString());
+                propertyBag.Add(TelemetryPropertyNames.Language, TelemetryConstants.Language);
+                propertyBag.Add(TelemetryPropertyNames.Service, TelemetryConstants.Service);
+
+                // Environment Variable set in Azure pipelines to correlate logs with the build number.
```
abhipsaMisra, line 103: 👍

common/test/TestLogging.cs
```diff
@@ -2,39 +2,114 @@
...
+                propertyBag.Add("BuildId", Environment.GetEnvironmentVariable("BUILD_BUILDID"));
+
+                _telemetryClient.TrackTrace(message, severity, propertyBag);
+            }
         }
 
-        public void WriteLine(string format, params object[] args)
```
abhipsaMisra, line 34: was this method call unused?

vinagesh: it was unused

vinagesh: I want to rename WriteLine to something else in the next PR and if we need one with args, we can add an override with correct name when needed.

common/test/TestLogging.cs
vinagesh: We don't have the key in any prod key-vault yet so it will always log. After it starts working everywhere, I can just remove that check. I will add a task for it.

e2e/test/E2ETestBase.cs
```diff
@@ -0,0 +1,47 @@
...
+            double elapsed = StopWatch.Elapsed.TotalSeconds;
+            var properties = new Dictionary<string, string>
+            {
+                { LoggingPropertyNames.TestRunTime, elapsed.ToString() }
+            };
+            Logger.WriteLine($"Finished test - {TestContext.TestName}", SeverityLevel.Information, properties);
+            // As this is not an application that keeps running, explicitly flushing is required to ensure we do not lose any logs.
+            // The recommendation from AI is to wait for 5 seconds after flushing.
+            Logger.Flush();
+            Thread.Sleep(5000);
```
abhipsaMisra, line 44: why do we need this sleep here? 
if we have 10 tests in a class, and we run all of them in parallel, `TestCleanup` would mark the end of processing for each of those tests, independently.

How would a "sleep" statement here have any impact?

common/test/TestLogging.cs
abhipsaMisra: no, I mean, in the scenario where something in our infra setup changed, and this key was not migrated over, there isn't any logging here to indicate what went wrong/ what is missing.

e2e/test/E2ETestBase.cs
vinagesh: The logger is per test and the object will be gone once the test exits. Before exiting this thread, I want to ensure all the logs are flushed out. Application Insights does some processing in the background after you call Flush and they ask us to wait 5 seconds before exiting. Flush should be sufficient and it works but just to be safe I added this - https://devblogs.microsoft.com/premier-developer/application-insights-use-case-for-telemetryclient-flush-calls/ . This should not cause any issues, just 5 secs extra per test and they all run in parallel, so impact should be very low.

vinagesh: To add more context - The way MsTest works is it creates and instance of the class per test, which implies an instance of logger per tests.

common/test/TestLogging.cs
```diff
@@ -2,39 +2,116 @@
...
         private TestLogging()
         {
+            if (_telemetryClient != null)
+            {
+                return;
+            }
+            else
+            {
+                // Thread-safe initialization of the telemetry client and run Id.
+                lock (initLock)
```
abhipsaMisra, line 44: there needs to be another check for `_telemetryClient != null` inside this lock.

If we have 2 tests accessing this at the same time, both will evaluate `_telemetryClient != null` as false, and try to grab the lock.
suppose test1 succeeds, it will grab the lock, initialize `_telemetryClient`. Once the lock is released, test2 will grab it again, still think `_telemetryClient` is null, and try to reinitialize.

e2e/test/E2ETestBase.cs
abhipsaMisra: So if I understand correctly, the sleep here is to hold on to the thread context for an additional 5secs, after Flush, so that the logs are incorporated completely?

common/test/TestLogging.cs
vinagesh: yes you are right. I had it that way and wanted to optimize but this does not work. I guess we have no other way than locking always.

common/test/TestLogging.cs
vinagesh: There is no logging till a logger is setup anyway :) The only time I can throw an exception is if the value is null/empty (which I cannot do right now as it is not configured). What else are you suggesting to throw on?

e2e/test/E2ETestBase.cs
vinagesh: yes

common/test/TestLogging.cs
abhipsaMisra: What my focus here is, there should be some way to inform us that since AI key isn't set up, nothing will get logged. It doesn't have to be an exception, a simple log statement would help too.

It is only to help us debug when we think that we have the infra set up as expected, but still don't see any records on application insights. Does that make sense?

vinagesh: Ok then we are on the same page. Once I remove the if null check, AI will automatically throw the correct exception back if the key is null or empty, which will bubble up. We don't need to throw a different exception. 
It will throw this:
Initialization method Microsoft.Azure.Devices.E2ETests.BulkOperationsE2ETests.TestInitialize threw exception. System.ArgumentNullException: Value cannot be null.
Parameter name: InstrumentationKey.

However there is no way to know if you put in an incorrect iKey. This is better done by setting up an alert on AI itself. We can say send an alert if you see 0 logs from this service in a day.

vinagesh:
> Can we attach a snapshot of what this looks like on AI, just for reference?

I had attached it on a comment but I can attach it here. I also plan to do a hands on session where everyone can run it and learn it.

common/test/TestLogging.cs
```diff
@@ -2,39 +2,115 @@
...
 {
+    /// <summary>
+    /// This class logs to multiple providers - Event source, Application insights
+    /// </summary>
     public class TestLogging
     {
-        private readonly static TestLogging s_instance = new TestLogging();
+        public const string Language = "C#";
+        public const string Service = "IotHub";
+        private readonly object initLock = new object();
```
drwill-ms, line 24: _initLock

common/test/TestLogging.cs
```diff
@@ -2,39 +2,115 @@
...
+        public string TargetFramework { get; set; }
 
         private TestLogging()
         {
+            // Thread-safe initialization of the telemetry client and run Id.
+            lock (initLock)
+            {
+                if (_telemetryClient == null)
+                {
+                    string iKey = Environment.GetEnvironmentVariable("E2E_IKEY");
```
drwill-ms, line 42: If you do the check outside of the lock, you can skip the lock all together for runs that don't have an AI Key.

common/test/TestLogging.cs
```diff
@@ -2,39 +2,115 @@
...
+        public static TestLogging GetInstance(TestContext testContext = null)
         {
-            return s_instance;
+            var logger = new TestLogging();
+
+            if (_telemetryClient != null)
+            {
+                // Set logger properties for the current test.
+                var targetFramework = (TargetFrameworkAttribute)Assembly.GetExecutingAssembly()
+                .GetCustomAttributes(typeof(TargetFrameworkAttribute), false)
```
drwill-ms, line 70: tab in

common/test/TestLogging.cs
```diff
@@ -2,39 +2,115 @@
...
 
         private const string NullInstance = "(null)";
-        
+
         public static string IdOf(object value) => value != null ? value.GetType().Name + "#" + GetHashCode(value) : NullInstance;
 
         public static int GetHashCode(object value) => value?.GetHashCode() ?? 0;
 
-        public static TestLogging GetInstance()
+        public static TestLogging GetInstance(TestContext testContext = null)
```
drwill-ms, line 62: This ties us to a specific testing framework ,doesn't it? Perhaps break out the parameters you specifically want.

common/test/TestLogging.cs
```diff
@@ -2,39 +2,115 @@
...
+            // Log to Application insights
+            if (_telemetryClient != null)
+            {
+                if (propertyBag == null)
+                {
+                    propertyBag = new Dictionary<string, string>();
+                }
+
+                // Add common properties
+                propertyBag.Add(LoggingPropertyNames.TestRunId, _testRunId);
```
drwill-ms, line 93: Why not have a persistent property bag you merge with, like we did in DTL?

e2e/test/E2ETestBase.cs
```diff
@@ -0,0 +1,47 @@
...
+        protected Stopwatch StopWatch { get; set; }
+        public TestContext TestContext { get; set; }
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            StopWatch = new Stopwatch();
+            StopWatch.Start();
+            Logger = TestLogging.GetInstance(TestContext);
+            Logger.WriteLine($"Starting test - {TestContext.TestName}", SeverityLevel.Information);
```
drwill-ms, line 27: Consider logging an event, instead.

e2e/test/E2ETestBase.cs
drwill-ms: How much slower are tests?

e2e/test/E2ETestBase.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        [TestCleanup]
+        public void TestCleanup()
+        {
+            string testStatus = TestContext.CurrentTestOutcome.ToString();
+            StopWatch.Stop();
+            double elapsed = StopWatch.Elapsed.TotalSeconds;
+            var properties = new Dictionary<string, string>
+            {
+                { LoggingPropertyNames.TestRunTime, elapsed.ToString() }
```
drwill-ms, line 38: Suggest putting the test status as a field in properties too, that way it can be easily queried for.

e2e/test/TelemetryPropertyNames.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    /// <summary>
+    /// The property names to log to telemetry.
+    /// </summary>
+    public class LoggerPropertyNames
```
drwill-ms, line 12: Isn't this a duplicate of Logging PropertyNames?

e2e/test/E2ETestBase.cs
vinagesh: 5 seconds total as they all run in parallel.

vinagesh: I do have a dimension to monitor the time it takes as well so I could do some analysis on the pipeline. Plus we can compare run time in the pipeline with respect to what it takes today. I'm not sure if there is a better solution today. I did see that if we exit without flushing, we were losing logs.

common/test/TestLogging.cs
vinagesh: It does tie us to a framework. I did have it like that previously and it became more than one parameter and decided to pass the whole object. But I do agree it is probably better not to tie it to a specific framework. Will send required parameters.

e2e/test/TelemetryPropertyNames.cs
vinagesh: I thought I renamed this. Not sure why they both exist. Let me take a look. It was renamed from TelemetryPropertyName -> LoggerPropertyNames.

e2e/test/E2ETestBase.cs
vinagesh: Start and Stop could be events.

e2e/test/E2ETestBase.cs
vinagesh: TestStatus is a customDimension too. Is that what you are asking about?

common/test/TestLogging.cs
vinagesh: I had it that way originally but there were hardly any constant properties so decided to create one on the fly instead of having another dictionary. Do you think having a persistent property bag is better? In this case, every test has its own property bag other than 2-3 dimensions.

common/test/TestLogging.cs
vinagesh: I did not choose to do that because every test does not have to Get the EnvVariable. And this way, it happens only once. And after I update all pipelines, this EnvVariable will always be there so the check might not be very useful. This is more to guard against errors till I onboard every pipeline.

common/test/TestLogging.cs
```diff
@@ -2,39 +2,116 @@
...
+    /// </summary>
     public class TestLogging
     {
-        private readonly static TestLogging s_instance = new TestLogging();
+        public const string Language = "C#";
+        public const string Service = "IotHub";
+        private readonly object _initLock = new object();
+
+        // Client to log to application insights
+        private static TelemetryClient _telemetryClient;
```
drwill-ms, line 27: I'm worried that this and _testRunId are static, but _initLock and other fields are not.

common/test/TestLogging.cs
drwill-ms: We can always change it later, I guess.

common/test/TestLogging.cs
drwill-ms: If you recall for DTL, we had a property bag for the initial logger.

Then each operation (for here that would be a test class or a test) would clone that, and add its own persistent properties.

Then each call to Log or Event could add additional properties, but that is rare.

So in the log method, you'd check if there are any custom properties, if so you'd merge persistent and transient into a new dictionary, and log that. If not, you'd just pass the persistent properties. That means, in most log cases, you'd not create a dictionary each time.

common/test/TestLogging.cs
```diff
@@ -2,39 +2,116 @@
...
+                propertyBag.Add(LoggingPropertyNames.TestRunId, _testRunId);
+                propertyBag.Add(LoggingPropertyNames.TestName, Context?.TestName);
+                propertyBag.Add(LoggingPropertyNames.ClassName, Context?.FullyQualifiedTestClassName ?? caller);
+                propertyBag.Add(LoggingPropertyNames.TestFramework, TargetFramework);
+                propertyBag.Add(LoggingPropertyNames.TestStatus, Context?.CurrentTestOutcome.ToString());
+                propertyBag.Add(LoggingPropertyNames.Language, Language);
+                propertyBag.Add(LoggingPropertyNames.Service, Service);
+
+                // Environment Variable set in Azure pipelines to correlate logs with the build number.
+                propertyBag.Add("BuildId", Environment.GetEnvironmentVariable("BUILD_BUILDID"));
```
drwill-ms, line 103: What if the variable is empty, because I ran locally?

e2e/test/E2ETestBase.cs
```diff
@@ -0,0 +1,47 @@
...
+            StopWatch = new Stopwatch();
+            StopWatch.Start();
+            Logger = TestLogging.GetInstance(TestContext);
+            Logger.WriteLine($"Starting test - {TestContext.TestName}", SeverityLevel.Information);
+        }
+
+        [TestCleanup]
+        public void TestCleanup()
+        {
+            string testStatus = TestContext.CurrentTestOutcome.ToString();
```
drwill-ms, line 33: remove

e2e/test/E2ETestBase.cs
drwill-ms, line 38: ToString on a time, you should specify the format.

e2e/test/E2ETestBase.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    /// <summary>
+    /// This class creates an instance of the logger for each test and logs the test result along with other useful information.
+    /// </summary>
+    public class E2ETestBase
+    {
+        protected TestLogging Logger { get; set; }
+        protected Stopwatch StopWatch { get; set; }
```
drwill-ms, line 18: Stopwatch

and private set?

common/test/TestLogging.cs
vinagesh: Good point, fixing it.

common/test/TestLogging.cs
vinagesh: It will be null - No value. I thought that would be an indicator that this is not running in the pipeline. AI did not really crash with null values but i can add a check to be very safe.

common/test/TestLogging.cs
vinagesh: True but other than the RunId, everything else is per test. To keep it simple and not have to clone, I just created it this way. We can surely do that if we think this is complicated or if it makes the code more generic.

e2e/test/E2ETestBase.cs
vinagesh: I renamed this as it seemed like a DateTime. This is just the number of seconds elapsed. It is a double, not DateTime.

PR #1476, opened 2020-07-17
fix(vsts): Fix script that determines which tests should be run

abhipsaMisra:


vsts/determine_tests_to_run.ps1
```diff
@@ -13,11 +13,6 @@ function IsPullRequestBuild
 
 function ShouldSkipIotHubTests 
 {	
-	if (IsPullRequestBuild)
```
vinagesh, line 16: Were we not running tests earlier?

vinagesh: Ahh draft PR, missed my eye :)

abhipsaMisra: we were actually doing the opposite, running tests irrespective of changes.

PR #1478, opened 2020-07-21

abhipsaMisra:
SDK code changes:
- pnp support over Amqp
- support for DTDL component name to be passed via message system properties.

### DIFF
--- a/e2e/test/E2ETestBase.cs
+++ b/e2e/test/E2ETestBase.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    /// <summary>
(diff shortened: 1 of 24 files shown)


e2e/test/E2ETestBase.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            Stopwatch.Stop();
+            double elapsed = Stopwatch.Elapsed.TotalSeconds;
+            var properties = new Dictionary<string, string>
+            {
+                { LoggingPropertyNames.SecondsElapsed, elapsed.ToString() }
+            };
+            Logger.WriteLine($"Finished test - {TestContext.TestName}", SeverityLevel.Information, properties);
+            // As this is not an application that keeps running, explicitly flushing is required to ensure we do not lose any logs.
+            // The recommendation from AI is to wait for 5 seconds after flushing.
```
azabbasi, line 40: in that case, isn't it better if the Logger.Flush method does the waiting?

azabbasi: so we don't have to always remember to do Thread.Wait every time we Flush

azabbasi: and this is a test logger ... so we should be fine.

abhipsaMisra: good idea, @vinagesh Can we add the sleep in here directly: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/common/test/TestLogging.cs#L115

PR #1480, opened 2020-07-22
Add E2E tests for supporting multiplexed pnp devices over Amqp

abhipsaMisra:


e2e/test/PlugAndPlayTests.cs
```diff
@@ -0,0 +1,332 @@
...
+    public class PlugAndPlayTests
+    {
+        private const string DevicePrefix = "plugAndPlayDevice";
+        private const string ModulePrefix = "plugAndPlayModule";
+        private const string TestModelId = "dtmi:com:example:testModel;1";
+
+        private static TestLogging s_log = TestLogging.GetInstance();
+        private readonly ConsoleEventListener _listener = TestConfig.StartEventListener();
+
+        [TestMethod]
```
abhipsaMisra, line 31: Decided against using `DataTestMethod`, since data driven tests cannot be parallelized using MS test: https://github.com/Microsoft/testfx-docs/blob/master/RFCs/004-In-Assembly-Parallel-Execution.md

e2e/test/PnpServiceTests.cs
```diff
@@ -1,114 +0,0 @@
...
-namespace Microsoft.Azure.Devices.E2ETests
-{
-    /// <summary>
-    /// Test class containing all tests to be run for plug and play.
-    /// </summary>
-    [TestClass]
-    [TestCategory("E2E")]
-    [TestCategory("IoTHub")]
-    [TestCategory("PlugAndPlay")]
-    public class PnpServiceTests
```
abhipsaMisra, line 22: I renamed this file, and added to the existing tests.

e2e/test/PlugAndPlayTests.cs
```diff
@@ -0,0 +1,326 @@
...
+                {
+                    AmqpConnectionPoolSettings = new AmqpConnectionPoolSettings()
+                    {
+                        MaxPoolSize = unchecked((uint)PoolingOverAmqp.MultipleConnections_PoolSize),
+                        Pooling = true
+                    }
+                }
+            };
+
+            // With PoolingOverAmqp.MultipleConnections_DevicesCount = 4, and PoolingOverAmqp.MultipleConnections_PoolSize = 2:
```
azabbasi, line 212: I am not sure about this comment, what if those constants change? if we want to test with specific values shouldn't we just not use constants here?

abhipsaMisra: There was no way for me to reference these constant values in these dev comments. The public summary comments can reference them directly by using <see cref=""/>, but that doesn't work for inline dev comments.
If these values change, the tests will still be valid, they'll just be tested against a different batch size.

The same logic is used for other multiplexing/ pooling tests as well.

PR #1482, opened 2020-07-22
fix(e2e): Rename files

vinagesh:
## Checklist

## Description of the changes
Added files only used in E2E tests to the E2E project. If this needs to be shared in the future, we can create a common project and add a project reference to it. Adding links to files is not the best option and makes it hard to do things like refactoring from  VS. Also keeping it in a common place like that might lead us to accidentally take dependency on something that just belongs to the tests.

Renamed the following:

WriteLine -> Trace
TestLogging -> TestLogger
VerboseTestLogging -> VerboseTestLogger
EventSourceTestLogging -> EventSourceTestLogger

Moved the following to E2E project:
Following files were added as a compile link in the .csproj in the E2E project and not used in any other project.

Common\CodeAnalysisOverrides.cs
Common\Configuration.cs
Common\Configuration.IoTHub.cs
Common\Configuration.Provisioning.cs
Common\Configuration.Provisioning.cs
Common\Configuration.Storage.cs
Common\Configuration.AzureSecurityCenterForIoTLogAnalytics.cs
Common\EventSourceTestLogging.cs
Common\TestModule.cs
Common\TestLogging.cs
Common\TestDevice.cs
Common\VerboseTestLogging.cs
Common\CustomWebProxy.cs
Common\ConsoleEventListener.cs


e2e/test/E2ETests.csproj
```diff
@@ -14,49 +14,6 @@
     <CommonTest>$(RootDir)\common\test</CommonTest>
```
abhipsaMisra, line 14: this can now be removed

e2e/test/Common/CodeAnalysisOverrides.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.CodeAnalysis;
+
+// Required by Configuration.*
+[assembly: SuppressMessage("Microsoft.Design", "CA1724:TypeNamesShouldNotMatchNamespaces")]
```
abhipsaMisra, line 7: do you know which classes are using this?

e2e/test/Common/ConsoleEventListener.cs
```diff
@@ -0,0 +1,84 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Globalization;
+using System.Linq;
+
+namespace System.Diagnostics.Tracing
+{
+    public sealed class ConsoleEventListener : EventListener
```
abhipsaMisra, line 9: FYI - this class is referenced in our readme here: https://github.com/Azure/azure-iot-sdk-csharp/tree/master/tools/CaptureLogs#console-logging
Make sure this path is updated as well

e2e/test/Common/EventSourceTestLogger.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    // f7ac322b-77f1-5a2d-0b56-ec79a41e82a2
+    [EventSource(Name = "Microsoft-Azure-Devices-TestLogging")]
+    internal class EventSourceTestLogger : EventSource
```
abhipsaMisra, line 10: don't unit test files use this class as well?

e2e/test/Common/CodeAnalysisOverrides.cs
vinagesh: I have not checked. I just copied everything the project was referring to.

e2e/test/Common/ConsoleEventListener.cs
vinagesh: Thanks for the reference. I'll update it.

e2e/test/Common/EventSourceTestLogger.cs
vinagesh: Everything builds after removing it. Plus I looked at all the .csproj for the links and did not find anything.

PR #1483, opened 2020-07-23
Reorganize E2E tests

abhipsaMisra:
The PR does the following:
- re-organize the E2E tests folder structure, to improve PR gate performance.
- Have all tests inherit from `E2EMsTestBase`.
    - This allows us to initialize the `ConsoleEventListener` in a per test basis. It does not have an effect on generated logs (based on local runs). If vsts produces erroneous results, we can rethink this approach.
    - Allows us to define default `Dispose()` behavior at the base class level.


e2e/test/config/TestConfig.cs
```diff
@@ -5,7 +5,7 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 //Workers = 0 makes the test engine use one worker per available core. It does not mean to run serially.
-[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.ClassLevel)]
+//[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.ClassLevel)]
```
abhipsaMisra, line 8: This is commented out right now, to leverage parallelization at the base test class level. If it works as expected, the goal is to remove this class altogether.

vinagesh, line 8: Delete?

e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs
```diff
@@ -2,28 +2,21 @@
...
         private readonly string ModulePrefix = $"E2E_{nameof(ConnectionStatusChangeHandlerTests)}";
-        private static TestLogger _log = TestLogger.GetInstance();
-
-        private readonly ConsoleEventListener _listener;
-
-        public ConnectionStatusChangeHandlerTests()
-        {
-            _listener = TestConfig.StartEventListener();
-        }
+        private static TestLogger s_log = TestLogger.GetInstance();
```
vinagesh, line 19: Just remove this s_log. The base class has 'Logger'. Use that below.

e2e/test/iothub/NoRetryE2ETests.cs
```diff
@@ -3,21 +3,21 @@
...
     [TestClass]
     [TestCategory("E2E")]
     [TestCategory("IoTHub")]
-    public class NoRetryE2ETests : IDisposable
+    public class NoRetryE2ETests : E2EMsTestBase
     {
         private readonly string DevicePrefix = $"E2E_{nameof(NoRetryE2ETests)}_";
-        private static TestLogger _log = TestLogger.GetInstance();
-        private readonly ConsoleEventListener _listener = TestConfig.StartEventListener();
+        private static TestLogger s_log = TestLogger.GetInstance();
```
vinagesh, line 20: Remove s_log and use Logger from base class.

e2e/test/iothub/messaging/MessageFeedbackE2ETests.cs
```diff
@@ -7,31 +7,26 @@
...
     {
         private const int MESSAGE_COUNT = 5;
 
         private static readonly string s_devicePrefix = $"E2E_{nameof(MessageFeedbackE2ETests)}_";
         private static readonly TimeSpan TIMESPAN_ONE_MINUTE = TimeSpan.FromMinutes(1);
         private static readonly TimeSpan TIMESPAN_FIVE_SECONDS = TimeSpan.FromSeconds(5);
 #pragma warning disable CA1823
-        private static readonly TestLogger _log = TestLogger.GetInstance();
-        private readonly ConsoleEventListener _listener;
+        private static readonly TestLogger s_log = TestLogger.GetInstance();
```
vinagesh, line 27: Remove s_log and use Logger from base class.

e2e/test/iothub/messaging/MessageReceiveE2EPoolAmqpTests.cs
```diff
@@ -6,23 +6,19 @@
...
     {
         private readonly string DevicePrefix = $"E2E_{nameof(MessageReceiveE2EPoolAmqpTests)}_";
-        private readonly ConsoleEventListener _listener;
-        private static TestLogger _log = TestLogger.GetInstance();
-
-        public MessageReceiveE2EPoolAmqpTests()
-        {
-            _listener = TestConfig.StartEventListener();
-        }
+        private static TestLogger s_log = TestLogger.GetInstance();
```
vinagesh, line 21: Remove s_log and use Logger

e2e/test/iothub/messaging/MessageReceiveFaultInjectionTests.cs
```diff
@@ -5,28 +5,24 @@
...
     [TestCategory("FaultInjection")]
-    public partial class MessageReceiveFaultInjectionTests : IDisposable
+    public partial class MessageReceiveFaultInjectionTests : E2EMsTestBase
     {
         private readonly string DevicePrefix = $"E2E_{nameof(MessageReceiveFaultInjectionTests)}_";
 
 #pragma warning disable CA1823
-        private static TestLogger _log = TestLogger.GetInstance();
-        private readonly ConsoleEventListener _listener;
+        private static TestLogger s_log = TestLogger.GetInstance();
```
vinagesh, line 23: Remove s_log and use Logger

e2e/test/iothub/messaging/MessageSendE2ETests.cs
```diff
@@ -10,28 +10,22 @@
...
         private static string ProxyServerAddress = Configuration.IoTHub.ProxyServerAddress;
-        private static TestLogger _log = TestLogger.GetInstance();
-
-        private readonly ConsoleEventListener _listener;
-
-        public MessageSendE2ETests()
-        {
-            _listener = TestConfig.StartEventListener();
-        }
+        private static TestLogger s_log = TestLogger.GetInstance();
```
vinagesh, line 28: Remove s_log and use Logger

e2e/test/iothub/messaging/MessageSendFaultInjectionTests.cs
```diff
@@ -6,21 +6,22 @@
...
     [TestCategory("FaultInjection")]
-    public partial class MessageSendFaultInjectionTests : IDisposable
+    public partial class MessageSendFaultInjectionTests : E2EMsTestBase
     {
         private readonly string _devicePrefix = $"E2E_{nameof(MessageSendFaultInjectionTests)}_";
 
 #pragma warning disable CA1823
-        private static TestLogger _log = TestLogger.GetInstance();
-        private readonly ConsoleEventListener _listener = TestConfig.StartEventListener();
+        private static TestLogger s_log = TestLogger.GetInstance();
```
vinagesh, line 24: I believe there are lots of this. Would be great if you can just update to using the base logger. But if it is too much work let me know and we can do it as a separate task.

e2e/test/E2EMsTestBase.cs
```diff
@@ -44,6 +53,20 @@ public void TestCleanup()
             Logger.Event(TestFinishedEventName, extraProperties);
             // As this is not an application that keeps running, explicitly flushing is required to ensure we do not lose any logs.
             Logger.SafeFlush();
+
+            _listener.Dispose();
```
vinagesh, line 57: Call Dispose here and move _listener.Dispose() to Dispose()

e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs
abhipsaMisra: awesome, good point!

e2e/test/E2EMsTestBase.cs
abhipsaMisra: we do want to dispose the console test listener for each test run, and not when the base test class is disposed.

e2e/test/config/TestConfig.cs
abhipsaMisra: yes, this class will be deleted once E2EMsBaseTest can handle logging as well. I'll add that change once a run is complete on VSTS

e2e/test/E2EMsTestBase.cs
```diff
@@ -44,6 +53,20 @@ public void TestCleanup()
...
+        }
+
+        // Dispose artifacts that are in the scope of the entire E2E test suite, and not of each individual test run.
+        // Test run specific artifacts are cleaned up in TestCleanup() above.
+        public void Dispose()
```
vinagesh, line 62: Why is this needed? Can you not add this to the virtual function below?

e2e/test/E2EMsTestBase.cs
```diff
@@ -1,21 +1,28 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.Tracing;
 using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
+//Workers = 0 makes the test engine use one worker per available core. It does not mean to run serially.
```
drwill-ms, line 11: Thanks for this.

e2e/test/E2EMsTestBase.cs
vinagesh, line 62: Instead of following this approach, why don't we add everything needed to the E2E base including - _logAnalyticsClient. They can be initialized on condition. For dispose, you can just check if this is null or not and then call dispose on it.

e2e/test/iothub/messaging/MessageSendFaultInjectionTests.cs
abhipsaMisra: I've removed all direct references to `TestLogger`, and added references tp `MsTestLogger`, instead.

abhipsaMisra: This is turning out to be more of renamings; you can take this up if you are already working on that code, i'll revert my changes :D

e2e/test/E2EMsTestBase.cs
vinagesh: And that should happen if you call Dispose() here and just move the _listener.Dispose() to the Dispose method. Ideally if this class is IDisposable, it should dispose everything disposable in the Dispose method and not in the TestCleanup.

e2e/test/E2EMsTestBase.cs
```diff
@@ -1,21 +1,28 @@
...
+[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.ClassLevel)]
+
```
vinagesh, line 13: I read about ExecutionScope.ClassLevel - This means tests within a class do not run in parallel right? This would interfere with the existing Logger implementation. We would then have to create a logger per test and not per class. Why are we not doing ExecutionScope.MethodLevel?

e2e/test/E2EMsTestBase.cs
abhipsaMisra: Is there an advantage of doing that? Why not have individual test classes manage their own resources, and shared resources be manages in the base class.
If we keep adding all resources refernce to the base case, it will end up getting all bloated up, and will be difficult to manage.

e2e/test/E2EMsTestBase.cs
abhipsaMisra: > Ideally if this class is IDisposable, it should dispose everything disposable in the Dispose method and not in the TestCleanup.

I agree with this, but how would we manage the scenario where you want to dispose some resources once you are done with test1, and some resources that you want to dispose once you are done with the entire test suite.
Wouldn't managing the lifecycle of the test specific resource inside `TestCleanup` make more sense, then?

e2e/test/E2EMsTestBase.cs
abhipsaMisra: Yes, that is correct. The plan is to have tests run at `ExecutionScope.MethodLevel` eventually, but we had issues with logging and also saw some test failures. (test failures were mostly related to event hub listener that we have - for validating message flow).
I will ignore the logging failures for now, since the AI changes you are making will take a different approach to logging, and remove the need for us relying on console logs on vsts. So let me run it at method level and see if we have test failures.

e2e/test/iothub/messaging/MessageSendFaultInjectionTests.cs
vinagesh: Probably just extend from base class but don't do any other renaming. Will be easier to do it in the following PR.

abhipsaMisra: yes, agreed.

e2e/test/E2EMsTestBase.cs
vinagesh: What is the difference? The only place we are initializing the listener is in the TestInitialize. What I was suggesting is have a generic dispose method you can call and based on if something is initialized or not, it will be disposed in the TestCleanup

e2e/test/E2EMsTestBase.cs
vinagesh: In case another tests needs it, it knows where to get it and then it knows there is a method to call to dispose it.

abhipsaMisra: yes, if there is resource required by >1 class, then we can treat is as a shared resource. There is only one test class using logAnalyticsClient, so I don't see the point of bring that into the base test class.

timtay-microsoft:
> re-organize the E2E tests folder structure, **to improve PR gate performance**.

Can you expand on this a bit? Why does changing the folder structure help our gate performance?

abhipsaMisra:
> > re-organize the E2E tests folder structure, **to improve PR gate performance**.
> 
> Can you expand on this a bit? Why does changing the folder structure help our gate performance?

Sure, currently a change to any file under the "e2e" folder will run the entire test suite. This includes changes to "prereq folder" - infra setup scripts, which don't require us running the entire test suite at the gate.
Also, any change in dps tests do not require us running the iothub tests; but our existing folder structure did not make it possible for us to differentiate between iothub tests and dps tests.

PR #1488, opened 2020-07-27
Parameterize direct method tests

timtay-microsoft:
I'm working on addressing the test gaps in our direct method tests, but this is the first step in making that feasible. Pretty much all of these test suites had copy/pasted versions of the same test but with one parameter different.

The only difference in testing that this PR brings is the use of x509 devices to the basic direct method tests. Previously, all direct method tests used SAS based devices. Now the test suite will test SAS and x509 devices for coverage.

Notably missing from direct method tests are invoking methods on modules. I'll address that in another PR to keep this one simpler

### DIFF
--- a/e2e/test/iothub/method/FaultInjectionPoolAmqpTests.MethodFaultInjectionPoolAmqpTests.cs
+++ b/e2e/test/iothub/method/FaultInjectionPoolAmqpTests.MethodFaultInjectionPoolAmqpTests.cs
@@ -18,704 +18,142 @@ namespace Microsoft.Azure.Devices.E2ETests
         private const string MethodDevicePrefix = "E2E_MethodFaultInjectionPoolAmqpTests";
         private const string MethodName = "MethodE2EFaultInjectionPoolAmqpTests";
 
+        [DataTestMethod]
         // TODO: #943 - Honor different pool sizes for different connection pool settings.
-        [Ignore]
-        [TestMethod]
-        public async Task Method_DeviceSak_DeviceMethodTcpConnRecovery_SingleConnection_Amqp()
+        //[DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_WebSocket_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.IoTHub, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_WebSocket_Only, ConnectionStringAuthScope.IoTHub, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        [DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.MultipleConnections_PoolSize, PoolingOverAmqp.MultipleConnections_DevicesCount)]
+        [DataRow(Client.TransportType.Amqp_WebSocket_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.MultipleConnections_PoolSize, PoolingOverAmqp.MultipleConnections_DevicesCount)]
+        [DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.IoTHub, PoolingOverAmqp.MultipleConnections_PoolSize, PoolingOverAmqp.MultipleConnections_DevicesCount)]
+        [DataRow(Client.TransportType.Amqp_WebSocket_Only, ConnectionStringAuthScope.IoTHub, PoolingOverAmqp.MultipleConnections_PoolSize, PoolingOverAmqp.MultipleConnections_DevicesCount)]
+        public async Task Method_DeviceMethodTcpConnRecovery(Client.TransportType transportType, ConnectionStringAuthScope authScope, int poolSize, int devicesCount)
         {
             await SendMethodAndRespondRecoveryPoolOverAmqpAsync(
                         TestDeviceType.Sasl,
-                        Client.TransportType.Amqp_Tcp_Only,
-                        PoolingOverAmqp.SingleConnection_PoolSize,
-                        PoolingOverAmqp.SingleConnection_DevicesCount,
+                        transportType,
+                        poolSize,
+                        devicesCount,
                         MethodE2ETests.SetDeviceReceiveMethodAsync,
                         FaultInjection.FaultType_Tcp,
-                        FaultInjection.FaultCloseReason_Boom)
+                        FaultInjection.FaultCloseReason_Boom,
+                        authScope: authScope)
                 .ConfigureAwait(false);
         }
 
+        [DataTestMethod]
         // TODO: #943 - Honor different pool sizes for different connection pool settings.
-        [Ignore]
-        [TestMethod]
-        public async Task Method_DeviceSak_DeviceMethodTcpConnRecovery_SingleConnection_AmqpWs()
+        //[DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_WebSocket_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.IoTHub, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_WebSocket_Only, ConnectionStringAuthScope.IoTHub, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        [DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.MultipleConnections_PoolSize, PoolingOverAmqp.MultipleConnections_DevicesCount)]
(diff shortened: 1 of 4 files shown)


e2e/test/iothub/method/FaultInjectionPoolAmqpTests.MethodFaultInjectionPoolAmqpTests.cs
```diff
@@ -20,702 +20,142 @@ public partial class FaultInjectionPoolAmqpTests
...
-                    FaultInjection.FaultCloseReason_Boom,
-                    authScope: ConnectionStringAuthScope.IoTHub)
-                .ConfigureAwait(false);
-        }
-
-        // TODO: #943 - Honor different pool sizes for different connection pool settings.
-        [Ignore]
-        [TestMethod]
-        public async Task Method_IoTHubSak_DeviceMethodAmqpConnLostRecovery_SingleConnection_AmqpWs()
+        //[DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
```
timtay-microsoft, line 48: Note that the single connection variants of many of these tests were disabled previously. There is no great way to ignore individual data rows other than to comment them out. The TODO's nearby explain why they are disabled by linking to github issues

timtay-microsoft:
(From the link above)

> Execution of data driven tests will not be parallelized - i.e. parallelizing over DataRow attributes is not supported.

I see, that is an issue. I thought this was okay based on the [send messages tests using it](https://github.com/Azure/azure-iot-sdk-csharp/blob/11fb2273a2f101884f6d1f6ac79399da7b0e0b94/e2e/test/iothub/messaging/MessageSendE2ETests.cs#L235)

timtay-microsoft:
Is this something we could accomplish if we switched over to XUnit or NUnit?

vinagesh:
> Is this something we could accomplish if we switched over to XUnit or NUnit?

It won't change with XUnit. There the parallelism is only across classes, not methods.

vinagesh:
> > Is this something we could accomplish if we switched over to XUnit or NUnit?
> 
> It won't change with XUnit. There the parallelism is only across classes, not methods. However, we should evaluate how long does it take for all the tests to run. Even if they are run serially and it takes less time than the longest time taking tests, [DataRow] is a good option for code cleanliness. Hopefully the logs I am adding should help determine that.

vinagesh:
> > > Is this something we could accomplish if we switched over to XUnit or NUnit?
> > 
> > 
> > It won't change with XUnit. There the parallelism is only across classes, not methods. However, we should evaluate how long does it take for all the tests to run. Even if they are run serially and it takes less time than the longest running test, [DataRow] is a good option for code cleanliness. I don't think we should write off [DataRow] for all tests. In fact it might help with thread starvation. If we do every single thing in parallel, we might see that the tests take longer than usual. Hopefully the logs I am adding should help determine that.

e2e/test/iothub/method/MethodE2EPoolAmqpTests.cs
```diff
@@ -19,203 +19,45 @@ public class MethodE2EPoolAmqpTests : E2EMsTestBase
...
-        {
-            await SendMethodAndRespondPoolOverAmqp(
-                    Client.TransportType.Amqp_WebSocket_Only,
-                    PoolingOverAmqp.SingleConnection_PoolSize,
-                    PoolingOverAmqp.SingleConnection_DevicesCount,
+        //[DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_WebSocket_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.IoTHub, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        //[DataRow(Client.TransportType.Amqp_WebSocket_Only, ConnectionStringAuthScope.IoTHub, PoolingOverAmqp.SingleConnection_PoolSize, PoolingOverAmqp.SingleConnection_DevicesCount)]
+        [DataRow(Client.TransportType.Amqp_Tcp_Only, ConnectionStringAuthScope.Device, PoolingOverAmqp.MultipleConnections_PoolSize, PoolingOverAmqp.MultipleConnections_DevicesCount)]
```
vinagesh, line 28: Evaluate the time this takes running in serial to determine if you want to use this or not.

vinagesh: If you give me a day, we should be able to get that data :)

timtay-microsoft: I'm fairly certain these tests take a long time to run in serial. I'm with Abhipsa on this one. I'm annoyed that MSTest isn't smart enough to parallelize these, though

vinagesh: I don't have context on the time this takes but we should definitely consider thread starvation. We cannot run everything in parallel and we need to group some tests (may not be these). Also, adding so many tests is hard to really maintain/figure out what combinations are being tested. The other option is to combine these tests if possible. Again I have not looked at these tests to know if that is a good solution for this situation or nor. Plus currently we only have parallelism across classes and not methods 'Scope = ExecutionScope.ClassLevel'. Abhipsa tried to switch it to per test and it seemed much much slower. We should definitely investigate further if there is any advantage not to add [DataRow] if we really cannot switch to 'Scope = ExecutionScope.MethodLevel' or not.

abhipsaMisra:
> I see, that is an issue. I thought this was okay based on the [send messages tests using it](https://github.com/Azure/azure-iot-sdk-csharp/blob/11fb2273a2f101884f6d1f6ac79399da7b0e0b94/e2e/test/iothub/messaging/MessageSendE2ETests.cs#L235)

Yes, I should have updated those tests to not use DataRow. I had started migrating the tests to DataRow, but then soon saw an increase in our test runtime. I highlighted this issue in the next PR and abandoned that effort, but forgot to migrate this sub-section of tests back to using TestMethod, my bad! 

> Is this something we could accomplish if we switched over to XUnit or NUnit?

XUnit doesn't seem to support this either: https://github.com/xunit/xunit/issues/648
It looks like NUnit does support parallelizing data driven tests: https://developers.perfectomobile.com/display/PD/Parallel+test+execution+in+Visual+Studio+using+NUnit

> Evaluate the time this takes running in serial to determine if you want to use this or not.

You are correct, we currently do not have tests running in parallel within a test class, but I am reluctant to make code changes that actively prohibits that.

> I don't have context on the time this takes but we should definitely consider thread starvation. We cannot run everything in parallel and we need to group some tests (may not be these)

Wouldn't the test framework handle thread scheduling? If we have a 100 test classes, and we parallelize at class level, we don't expect all 100 of them to get kicked off together, but rather be optimized based on the thread pool size available. What kind of issue do you anticipate here, with parallelization?

vinagesh:
> > I see, that is an issue. I thought this was okay based on the [send messages tests using it](https://github.com/Azure/azure-iot-sdk-csharp/blob/11fb2273a2f101884f6d1f6ac79399da7b0e0b94/e2e/test/iothub/messaging/MessageSendE2ETests.cs#L235)
> 
> Yes, I should have updated those tests to not use DataRow. I had started migrating the tests to DataRow, but then soon saw an increase in our test runtime. I highlighted this issue in the next PR and abandoned that effort, but forgot to migrate this sub-section of tests back to using TestMethod, my bad!
> 
> > Is this something we could accomplish if we switched over to XUnit or NUnit?
> 
> XUnit doesn't seem to support this either: [xunit/xunit#648](https://github.com/xunit/xunit/issues/648)
> It looks like NUnit does support parallelizing data driven tests: https://developers.perfectomobile.com/display/PD/Parallel+test+execution+in+Visual+Studio+using+NUnit
> 
> > Evaluate the time this takes running in serial to determine if you want to use this or not.
> 
> You are correct, we currently do not have tests running in parallel within a test class, but I am reluctant to make code changes that actively prohibits that.
> 
> > I don't have context on the time this takes but we should definitely consider thread starvation. We cannot run everything in parallel and we need to group some tests (may not be these)
> 
> Wouldn't the test framework handle thread scheduling? If we have a 100 test classes, and we parallelize at class level, we don't expect all 100 of them to get kicked off together, but rather be optimized based on the thread pool size available. What kind of issue do you anticipate here, with parallelization?

Based on your trials a few days back, we clearly saw that parallalizing at method level was performing much worse than class level. Either we have written code that deadlocks or the framework is not great at handling the thread scheduling and is doing something weird. The other side to plain parallalization code maintainability. Think about when we need to update or change these tests to reflect changes in the service or SDK. We have to make multiple edits and this would not be easily maintainable. We should just measure the time it takes for the test and make a data driven decision than guessing. Given we cannot run all methods in parallel, we should probably investigate why before we conclude on not using [DataRow]. If we never get to method level parallalization, we will neither have parallalization nor compact code.

vinagesh:
> > > I see, that is an issue. I thought this was okay based on the [send messages tests using it](https://github.com/Azure/azure-iot-sdk-csharp/blob/11fb2273a2f101884f6d1f6ac79399da7b0e0b94/e2e/test/iothub/messaging/MessageSendE2ETests.cs#L235)
> > 
> > 
> > Yes, I should have updated those tests to not use DataRow. I had started migrating the tests to DataRow, but then soon saw an increase in our test runtime. I highlighted this issue in the next PR and abandoned that effort, but forgot to migrate this sub-section of tests back to using TestMethod, my bad!
> > > Is this something we could accomplish if we switched over to XUnit or NUnit?
> > 
> > 
> > XUnit doesn't seem to support this either: [xunit/xunit#648](https://github.com/xunit/xunit/issues/648)
> > It looks like NUnit does support parallelizing data driven tests: https://developers.perfectomobile.com/display/PD/Parallel+test+execution+in+Visual+Studio+using+NUnit
> > > Evaluate the time this takes running in serial to determine if you want to use this or not.
> > 
> > 
> > You are correct, we currently do not have tests running in parallel within a test class, but I am reluctant to make code changes that actively prohibits that.
> > > I don't have context on the time this takes but we should definitely consider thread starvation. We cannot run everything in parallel and we need to group some tests (may not be these)
> > 
> > 
> > Wouldn't the test framework handle thread scheduling? If we have a 100 test classes, and we parallelize at class level, we don't expect all 100 of them to get kicked off together, but rather be optimized based on the thread pool size available. What kind of issue do you anticipate here, with parallelization?
> 
> Based on your trials a few days back, we clearly saw that parallalizing at method level was performing much worse than class level. Either we have written code that deadlocks or the framework is not great at handling the thread scheduling and is doing something weird. The other side to plain parallalization code maintainability. Think about when we need to update or change these tests to reflect changes in the service or SDK. We have to make multiple edits and this would not be easily maintainable. We should just measure the time it takes for the test and make a data driven decision than guessing. Given we cannot run all methods in parallel, we should probably investigate why before we conclude on not using [DataRow]. If we never get to method level parallalization, we will neither have parallalization nor compact code. Again this test might really be long running and benefit from not using [DataRow] so just do a quick time check and conclude.

timtay-microsoft:
Closing this PR since this discussion is getting bigger than just parameterizing a few tests. If we switch to NUnit, this effort may work, but this PR isn't really enough by itself

PR #1490, opened 2020-07-28
feat(e2e): Extend all tests from the new attribute to be able to log …

vinagesh:
…test error.

## Checklist

## Description of the changes
Adding the extended attribute to all E2E tests to help logging test errors.


timtay-microsoft:
What kind of information gets logged with this new attribute that wasn't logged before?

vinagesh:
> What kind of information gets logged with this new attribute that wasn't logged before?

I am sending all logs to Application Insights (there is a story on the board) so that we can easily query and find patterns with test results/failures. As part of that,  I added this attribute to extract any exception in the test and log it to AI. This is not for console logging, only Application insights logging. This way each test does not need to worry about this.

PR #1491, opened 2020-07-28
Remove event hub validation from telemetry e2e tests

abhipsaMisra:
This PR removes the event hub side validation of things, from the telemetry E2E tests flow. The SDK will rely on the `SendEventAsync()` call to complete successfully, in order to validate that the call was successful.
The forwarding of message from IoT hub to any integrated endpoint is under the responsibility of the service, and the SDK doesn't have a role to play there.

This PR also:
- parallelizes test execution to occur at method level (classes and methods are run in parallel).
- adds some additional logging for mqtt implementation.


iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -315,6 +315,8 @@ private async Task Connect(IChannelHandlerContext context)
                     usernameString += $"&{AuthChainParam}={Uri.EscapeDataString(this.mqttTransportSettings.AuthenticationChain)}";
                 }
 
+                if (Logging.IsEnabled) Logging.Info(this, $"{nameof(usernameString)}={usernameString}", nameof(Connect));
```
abhipsaMisra, line 318: This is unrelated to the PR, but will help in debugging mqtt logs

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -415,6 +417,8 @@ private static async void PingServer(object ctx)
                 {
                     // We've been idle for too long, send a ping!
                     await WriteMessageAsync(context, PingReqPacket.Instance, ShutdownOnWriteErrorHandler).ConfigureAwait(true);
+
+                    if (Logging.IsEnabled) Logging.Info(context, $"Idle time was {idleTime}, so ping request was sent.", nameof(PingServer));
```
abhipsaMisra, line 421: This is unrelated to the PR, but will help in debugging mqtt logs

e2e/test/config/Configuration.IoTHub.cs
```diff
@@ -13,12 +13,6 @@ public static partial class IoTHub
         {
             public static string ConnectionString => GetValue("IOTHUB_CONN_STRING_CSHARP");
 
-            public static string EventHubString => GetValue("IOTHUB_EVENTHUB_CONN_STRING_CSHARP", string.Empty);
-
-            public static string EventHubCompatibleName => GetValue("IOTHUB_EVENTHUB_COMPATIBLE_NAME", string.Empty);
-
-            public static string EventHubConsumerGroup => GetValue("IOTHUB_EVENTHUB_CONSUMER_GROUP", "$Default");
```
timtay-microsoft, line 20: Don't forget to clean up these values in keyvault, too

e2e/test/iothub/messaging/MessageSendE2EPoolAmqpTests.cs
```diff
@@ -132,9 +132,6 @@ public async Task Message_IoTHubSak_DeviceSendSingleMessage_MultipleConnections_
                 (Client.Message testMessage, string payload, string p1Value) = MessageSendE2ETests.ComposeD2cTestMessage();
                 s_log.Trace($"{nameof(MessageSendE2EPoolAmqpTests)}.{testDevice.Id}: messageId='{testMessage.MessageId}' payload='{payload}' p1Value='{p1Value}'");
                 await deviceClient.SendEventAsync(testMessage).ConfigureAwait(false);
```
timtay-microsoft, line 134: Is there some logic built into the SDK that makes this method throw an exception if IotHub responds to this message with a non-200 status code?

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
timtay-microsoft: Does this logging framework have lower log levels? This seems like a trace level log, if the framework supports it.

vsts/test-release-nuget.yaml
```diff
@@ -89,6 +86,7 @@ jobs:
           LA_AAD_APP_ID: $(LA-AAD-APP-ID)
           LA_AAD_APP_CERT_BASE64: $(LA-AAD-APP-CERT-BASE64)
           LA_WORKSPACE_ID: $(LA-WORKSPACE-ID)
+          E2E_IKEY: $(E2E-IKEY)
```
timtay-microsoft, line 89: What is this used for? I don't see the e2e tests looking for this value

e2e/test/E2EMsTestBase.cs
```diff
@@ -9,7 +9,7 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 //Workers = 0 makes the test engine use one worker per available core. It does not mean to run serially.
-[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.ClassLevel)]
+[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.MethodLevel)]
```
prmathur-microsoft, line 13: 👍

e2e/test/iothub/messaging/AzureSecurityCenterForIoTSecurityMessageE2ETests.cs
```diff
@@ -174,23 +174,17 @@ private async Task TestSecurityMessageModuleAsync(Client.TransportType transport
...
         private async Task ValidateEventAsync(
             string deviceId,
-            Client.Message message,
             string eventId,
-            string payload,
-            string p1Value,
             AzureSecurityCenterForIoTLogAnalyticsClient logAnalticsTestClient)
         {
-            bool isReceivedEventHub = EventHubTestListener.VerifyIfMessageIsReceived(deviceId, message, payload, p1Value, TimeSpan.FromSeconds(10));
-            Assert.IsFalse(isReceivedEventHub, "Security message received in customer event hub.");
```
prmathur-microsoft, line 193: For this pattern of verification, is there a place where we are still checking for ack from iothub?

e2e/test/iothub/messaging/MessageSendE2EPoolAmqpTests.cs
abhipsaMisra: If the `.SendEventAsync()` call throws an exception, the SDK will retry based on it's retry policy, or if non-retryable, will forward the exception to the calling application.

vsts/test-release-nuget.yaml
abhipsaMisra: This is being used to hook up our tests to application insights: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/e2e/test/Helpers/logging/TestLogger.cs#L40

e2e/test/iothub/messaging/AzureSecurityCenterForIoTSecurityMessageE2ETests.cs
abhipsaMisra: Not explicitly, no. If the `SemdEventAsync()` call completes successfully, then the call was implicitly ack'ed by the service. If the response is not a success, then either the application retries (based on the retry policy), or forwards the exception to the calling application.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
abhipsaMisra: The `Info` event is something that we use to log additional information: https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/removeEventHub/common/src/Logging.Common.cs#L195. 
Afaik EventSource doesn't have defined log lovels (like log4j does), but we could always add an additional event called `Trace` and use that for more granular logging.

PR #1494, opened 2020-07-30
Clean up test logging instance references

abhipsaMisra:
This PR removes the test logger instance initialization from each test class, and instead, references them from the base class directly.

This does not solve the increase in test suite running time, but is a first step in isolating the logging changes.

### DIFF
--- a/e2e/test/Helpers/CustomWebProxy.cs
+++ b/e2e/test/Helpers/CustomWebProxy.cs
@@ -9,8 +9,13 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
 {
     public class CustomWebProxy : IWebProxy
     {
-        private static readonly TestLogger s_testLog = TestLogger.GetInstance();
+        private readonly TestLogger s_testLog;
         private long _counter = 0;
+
+        public CustomWebProxy(TestLogger logger)
+        {
+            s_testLog = logger;
+        }
 
         public ICredentials Credentials { get; set; }
 
--- a/e2e/test/Helpers/TestDevice.cs
+++ b/e2e/test/Helpers/TestDevice.cs
@@ -24,8 +24,9 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
     public class TestDevice
     {
         private const int DelayAfterDeviceCreationSeconds = 0;
-        private static readonly TestLogger s_log = TestLogger.GetInstance();
         private static readonly SemaphoreSlim s_semaphore = new SemaphoreSlim(1, 1);
+
+        private static TestLogger s_logger;
 
         private TestDevice(Device device, Client.IAuthenticationMethod authenticationMethod)
         {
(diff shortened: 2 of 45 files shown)


e2e/test/Helpers/TestDevice.cs
```diff
@@ -38,16 +39,17 @@ private TestDevice(Device device, Client.IAuthenticationMethod authenticationMet
         /// </summary>
         /// <param name="namePrefix">The prefix to apply to your device name</param>
         /// <param name="type">The way the device will authenticate</param>
-        public static async Task<TestDevice> GetTestDeviceAsync(string namePrefix, TestDeviceType type = TestDeviceType.Sasl)
+        public static async Task<TestDevice> GetTestDeviceAsync(TestLogger logger, string namePrefix, TestDeviceType type = TestDeviceType.Sasl)
```
abhipsaMisra, line 42: The calling test method will pass in it's `Logger` instance to all helper methods.

e2e/test/Helpers/logging/TestLogger.cs
```diff
@@ -89,14 +76,14 @@ public void Event(string eventName, IDictionary<string, string> extraProperties
...
-        public void SafeFlush()
+        public async Task SafeFlushAsync()
         {
             if (_telemetryClient != null)
             {
                 _telemetryClient.Flush();
 
                 // Hold on to the thread context to allow flushing to complete.
-                Thread.Sleep(5000);
+                await Task.Delay(5000).ConfigureAwait(false);
```
abhipsaMisra, line 86: I added a task delay instead of a thread sleep. @vinagesh - is this ok, or does this need to be a thread sleep?

vinagesh: No we should not use Thread.Sleep at all. I missed it. It should always be Task.Delay.

e2e/test/Helpers/logging/MsTestLogger.cs
```diff
@@ -30,10 +30,5 @@ private MsTestLogger(TestContext testContext) : base()
             Properties.Add(LoggingPropertyNames.TargetFramework, targetFramework.FrameworkName);
             Properties.Add(LoggingPropertyNames.OsInfo, operatingSystem);
         }
-
-        public static MsTestLogger GetInstance(TestContext testContext)
```
abhipsaMisra, line 34: This does not need to be removed, I had modifies this as a part of my internal testing, and forgot to revert.

e2e/test/Helpers/CustomWebProxy.cs
```diff
@@ -9,9 +9,14 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
 {
     public class CustomWebProxy : IWebProxy
     {
-        private static readonly TestLogger s_testLog = TestLogger.GetInstance();
+        private readonly TestLogger s_testLog;
```
vinagesh, line 12: rename to _logger

e2e/test/Helpers/logging/TestLogger.cs
```diff
@@ -27,28 +25,22 @@ public class TestLogger
 
         public IDictionary<string, string> Properties { get; } = new Dictionary<string, string>();
 
-        protected TestLogger()
+        // Static constructor is called at most one time, before any instance constructor is invoked or member is accessed.
+        static TestLogger()
```
abhipsaMisra, line 29: I preferred using a static constructor over adding locks to an instance constructor, since the implementation requires this block to be executed only once, and this was an easy way to achieve this.

e2e/test/Helpers/logging/TestLogger.cs
```diff
@@ -58,11 +50,6 @@ protected TestLogger()
 
         public static int GetHashCode(object value) => value?.GetHashCode() ?? 0;
 
-        public static TestLogger GetInstance()
```
abhipsaMisra, line 61: Again, had removed for testing, but this is not the cause for the delay.

PR #1497, opened 2020-07-30
refactor(e2e): Add tests for invoking methods on modules

timtay-microsoft:
Our existing methods tests are largely fine, but we didn't have any tests for invoking methods on modules, so I added them. When reviewing, please point out any scenarios that are worth adding tests for as this is all I could find that was missing for our "Address Gaps" effort

### DIFF
--- a/e2e/test/iothub/method/MethodE2ETests.cs
+++ b/e2e/test/iothub/method/MethodE2ETests.cs
@@ -107,7 +107,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Methods
                 HttpProxy = new WebProxy(Configuration.IoTHub.ProxyServerAddress)
             };
 
-            await SendMethodAndRespondAsync(Client.TransportType.Mqtt_Tcp_Only, SetDeviceReceiveMethodAsync, serviceClientTransportSettings).ConfigureAwait(false);
+            await SendMethodAndRespondAsync(Client.TransportType.Mqtt_Tcp_Only, SetDeviceReceiveMethodAsync, serviceClientTransportSettings: serviceClientTransportSettings).ConfigureAwait(false);
         }
 
         [LoggedTestMethod]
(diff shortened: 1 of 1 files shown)


e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -133,7 +134,7 @@ public async Task Method_ServiceInvokeDeviceMethodWithUnknownDeviceThrows()
             try
             {
                 // Invoke the direct method asynchronously and get the response from the simulated device.
-                await serviceClient.InvokeDeviceMethodAsync("MyDummyDevice", methodInvocation);
+                await serviceClient.InvokeDeviceMethodAsync("MyDummyDevice", methodInvocation).ConfigureAwait(false);
```
abhipsaMisra, line 137: - I know you didn't add this, but maybe change `MyDummyDevice` to something else? 😄 
- We don't really need to `ConfigureAwait()` in a .netcoreapp/ .netfm application. The application's synchronization context would handle the task scheduling.

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -145,14 +146,89 @@ public async Task Method_ServiceInvokeDeviceMethodWithUnknownDeviceThrows()
...
+            // act
+            ErrorCode actualErrorCode = ErrorCode.InvalidErrorCode;
+            try
+            {
+                // Invoke the direct method asynchronously and get the response from the simulated device.
+                await serviceClient.InvokeDeviceMethodAsync(testDevice.Id, "someNonExistantModuleOnAnExistingDevice", methodInvocation).ConfigureAwait(false);
+            }
+            catch (DeviceNotFoundException ex)
+            {
+                // Although the exception is called "Device" not found, it is used for all 404's, including the 404010 that denotes a module was not found
```
abhipsaMisra, line 215: 👍

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -145,14 +146,89 @@ public async Task Method_ServiceInvokeDeviceMethodWithUnknownDeviceThrows()
...
+        public async Task Method_ServiceInvokeDeviceMethodWithUnknownModuleThrows()
+        {
+            // setup
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, "ModuleNotFoundTest").ConfigureAwait(false);
+            using var serviceClient = ServiceClient.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+            var methodInvocation = new CloudToDeviceMethod("SetTelemetryInterval");
+            methodInvocation.SetPayloadJson("10");
+
+            // act
+            ErrorCode actualErrorCode = ErrorCode.InvalidErrorCode;
```
abhipsaMisra, line 207: why do we need to initialize this to InvalidErrorCode?

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -163,14 +239,33 @@ public static async Task ServiceSendMethodAndVerifyResponseAsync(string deviceNa
             await serviceClient.CloseAsync().ConfigureAwait(false);
         }
 
-        public static async Task ServiceSendMethodAndVerifyResponseAsync(string deviceName, string methodName, string respJson, string reqJson, TimeSpan responseTimeout, ServiceClientTransportSettings serviceClientTransportSettings, TestLogger logger)
+        public static async Task ServiceSendMethodAndVerifyResponseAsync(string deviceId, string methodName, string respJson, string reqJson, TimeSpan responseTimeout, ServiceClientTransportSettings serviceClientTransportSettings, TestLogger logger)
```
abhipsaMisra, line 242: why have this and `public static async Task ServiceSendMethodAndVerifyResponseAsync(string deviceId, string methodName, string respJson, string reqJson, TestLogger logger)` both? Why not set optional params to default, and have just one method?

e2e/test/iothub/method/MethodE2ETests.cs
timtay-microsoft: So that the value of the field can be compared outside of the try/catch block.

timtay-microsoft: If the catch doesn't execute, then the comparison will fail

e2e/test/iothub/method/MethodE2ETests.cs
timtay-microsoft: Are you referring to the methods that take deviceId vs the methods that take deviceId and moduleId? I did try combining them like that, but the entire method ended being one big if/else block on the condition that the moduleId is present, so it didn't simplify much

abhipsaMisra: There are also some methods that take in `TimeSpan responseTimeout, ServiceClientTransportSettings serviceClientTransportSettings` as optional. Those seem like they can be combined. (line 224 vs line 242).

PR #1499, opened 2020-07-31
Ensuring that logging does not take too much time

vinagesh:
## Checklist

Addressing the issue of tests timing out on enabling AI on our CI and nightly pipelines
1) Added a global static logger to use for logging. This is ensure we don't have to flush after each test. We will only flush at the end of test suite to ensure all logs are really sent to AI.
2) Removed logging events as they seem to take double the time as traces (I will try to follow up with AI to understand why there is a performance impact using their service). 
As a result of these changes, the CI and nightly will take the same time as they took earlier. If we observe differently after a few days of trial, I will investigate further.

### DIFF
--- a/e2e/test/Helpers/CustomWebProxy.cs
+++ b/e2e/test/Helpers/CustomWebProxy.cs
@@ -9,12 +9,12 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
 {
     public class CustomWebProxy : IWebProxy
     {
-        private readonly TestLogger s_testLog;
+        private readonly MsTestLogger _logger;
         private long _counter = 0;
 
-        public CustomWebProxy(TestLogger logger)
+        public CustomWebProxy(MsTestLogger logger)
         {
-            s_testLog = logger;
+            _logger = logger;
         }
 
         public ICredentials Credentials { get; set; }
@@ -29,7 +29,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         public bool IsBypassed(Uri host)
         {
             Interlocked.Increment(ref _counter);
-            s_testLog.Trace($"{nameof(CustomWebProxy)}.{nameof(IsBypassed)} Uri = {host}");
+            _logger.Trace($"{nameof(CustomWebProxy)}.{nameof(IsBypassed)} Uri = {host}");
             return false;
         }
     }
--- a/e2e/test/Helpers/TestDevice.cs
+++ b/e2e/test/Helpers/TestDevice.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         private const int DelayAfterDeviceCreationSeconds = 0;
         private static readonly SemaphoreSlim s_semaphore = new SemaphoreSlim(1, 1);
 
-        private static TestLogger s_logger;
+        private static MsTestLogger _logger;
 
         private TestDevice(Device device, Client.IAuthenticationMethod authenticationMethod)
         {
@@ -39,9 +39,9 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         /// </summary>
         /// <param name="namePrefix">The prefix to apply to your device name</param>
         /// <param name="type">The way the device will authenticate</param>
-        public static async Task<TestDevice> GetTestDeviceAsync(TestLogger logger, string namePrefix, TestDeviceType type = TestDeviceType.Sasl)
+        public static async Task<TestDevice> GetTestDeviceAsync(MsTestLogger logger, string namePrefix, TestDeviceType type = TestDeviceType.Sasl)
         {
-            s_logger = logger;
+            _logger = logger;
             string prefix = namePrefix + type + "_";
 
             try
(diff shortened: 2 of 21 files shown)


e2e/test/E2ETests.csproj
```diff
@@ -26,7 +26,7 @@
   <!-- All Platforms -->
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.10.0" />
```
vinagesh, line 28: Updated AI in case it helps

e2e/test/Helpers/CustomWebProxy.cs
```diff
@@ -9,12 +9,12 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
 {
     public class CustomWebProxy : IWebProxy
     {
-        private readonly TestLogger s_testLog;
+        private readonly MsTestLogger _logger;
```
abhipsaMisra, line 12: Any reason this needs to be the child class instance, rather than the parent class `TestLogger`? Isn't the `.Trace()` API call available on the parent class directly?

e2e/test/Helpers/TestDevice.cs
```diff
@@ -26,7 +26,7 @@ public class TestDevice
         private const int DelayAfterDeviceCreationSeconds = 0;
         private static readonly SemaphoreSlim s_semaphore = new SemaphoreSlim(1, 1);
 
-        private static TestLogger s_logger;
+        private static MsTestLogger _logger;
```
abhipsaMisra, line 29: static variable needs `s_` prefix

abhipsaMisra: question about using `TestLogger` vs `MsTestLogger` instance, here and below.

e2e/test/Helpers/logging/MsTestLogger.cs
```diff
@@ -1,20 +1,33 @@
...
     /// <summary>
     /// Ms Test framework specific logging.
+    /// Every test will have its own instance of this logger with properties specific to the test.
+    /// This logger uses the global logger to write logs but has its own set of properties.
+    /// We cannot inherit and log using this because this logger object gets disposed at the end of the test
+    /// and we will lose logs without doing a flush. Flush is expensive and should be done only once when
+    /// we exit the test suite.
     /// </summary>
-    public class MsTestLogger : TestLogger
+    public class MsTestLogger
```
abhipsaMisra, line 22: I did not understand the bit about the problem with having this class inherit from `TestLogger`; could you explain that part a bit more?

e2e/test/Helpers/logging/MsTestLogger.cs
```diff
@@ -25,10 +38,16 @@ internal MsTestLogger(TestContext testContext) : base()
...
-            Properties.Add(LoggingPropertyNames.TestClassName, testContext.FullyQualifiedTestClassName);
-            Properties.Add(LoggingPropertyNames.TargetFramework, targetFramework.FrameworkName);
-            Properties.Add(LoggingPropertyNames.OsInfo, operatingSystem);
+            MsTestProperties.Add(LoggingPropertyNames.TestName, testContext.TestName);
+            MsTestProperties.Add(LoggingPropertyNames.TestClassName, testContext.FullyQualifiedTestClassName);
+            MsTestProperties.Add(LoggingPropertyNames.TargetFramework, targetFramework.FrameworkName);
+            MsTestProperties.Add(LoggingPropertyNames.OsInfo, operatingSystem);
+        }
+
+        public void Trace(string message, SeverityLevel severity = SeverityLevel.Information, IDictionary<string, string> extraProperties = null)
```
abhipsaMisra, line 47: I am confused about how this method call works, as opposed to how `TestLogger.Trace` works

vsts/vsts.yaml
```diff
@@ -17,15 +17,107 @@ resources:
...
+      - task: PublishTestResults@2
+        displayName: "Publish Test Results **/*.trx"
+        inputs:
+          testRunner: VSTest
+          testRunTitle: "Linux Tests ($(FRAMEWORK))"
+          testResultsFiles: "**/*.trx"
+
+        condition: always()
+  ### Linux build 2 ###
+  - job: LINUX2
```
abhipsaMisra, line 115: oh, why are we separating this matrix into different jobs? each framework in the matrix should be assigned its own hosted agent to run on, so what benefit does this bring in?

vsts/vsts.yaml
```diff
@@ -216,6 +304,104 @@ jobs:
 
         condition: always()
 
+  ### Windows build 2 ###
```
abhipsaMisra, line 307: same question here as well

vinagesh:
Adding explanation general for everyone -

1) We want to have a singleton instance of Logger that does not get disposed so that we can call Flush on it in the end only once -> This means we need a static class
2) We want to have an instance of properties per test class (this is not really a logger but has a reference to the global logger plus test specific properties)
3) Every test will create an instance of its logger but call GlobalLogger.Trace.

vinagesh:
Discussed all comments offline and resolved.

PR #1503, opened 2020-08-04
Provide validation while setting Twin property to ensure they respect naming restrictions

tomkerkhove:
## Checklist

## Description of the changes
Provide validation while setting Twin property to ensure they respect naming restrictions.
> **Keys:** All keys in JSON objects are UTF-8 encoded, case-sensitive, and up-to 1 KB in length. Allowed characters exclude UNICODE control characters (segments C0 and C1), and ., $, and SP.

Learn more on https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins.

I did not add, however, validation for UNICODE characters given I'm not sure if you want to go that far. In terms of "SP" I'm not sure what that means so didn't add that either.

## Why?

Yesterday I was using a property along the lines of `Property.Sample` where the SDK threw an exception along the lines of `Service did not provide a response` and it took a while to notice that the name was not allowed.

If the validation would happen during setting the name it would be a major help for people who are new to twins.

None, but happy to open in case of need.

### DIFF
--- a/shared/src/TwinCollection.cs
+++ b/shared/src/TwinCollection.cs
@@ -154,7 +154,29 @@ namespace Microsoft.Azure.Devices.Shared
                     throw new ArgumentOutOfRangeException(nameof(propertyName));
                 }
             }
-            set => TrySetMemberInternal(propertyName, value);
+            set
+            {
+                if (propertyName == null)
+                {
+                    throw new ArgumentNullException(nameof(propertyName),"Property name cannot be null");
+                }
+#pragma warning disable CA1307 // Specify StringComparison
+                else if (propertyName.Contains("."))
+                {
+                    throw new ArgumentException(
+                        "Property name cannot contain '.' in its name. Learn more on https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins");
+                }
+                else if (propertyName.Contains("$"))
+                {
+                    throw new ArgumentException(
+                        "Property name cannot contain '$' in its name. Learn more on https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins");
+                }
+#pragma warning restore CA1307 // Specify StringComparison
+                else
+                {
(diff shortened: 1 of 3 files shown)


shared/src/TwinCollection.cs
```diff
@@ -154,7 +154,29 @@ public int Count
...
                 }
             }
-            set => TrySetMemberInternal(propertyName, value);
+            set
+            {
+                if (propertyName == null)
+                {
+                    throw new ArgumentNullException(nameof(propertyName),"Property name cannot be null");
+                }
+#pragma warning disable CA1307 // Specify StringComparison
```
tomkerkhove, line 163: The suggested overload with `StringComparison` does not exist

abhipsaMisra:
Thank you for the thorough explanation, @tomkerkhove !
We really appreciate your PR (and thanks for adding tests as well); however, we avoid adding client-side validations for these scenarios, and instead rely on the service throwing the correct exception with appropriate details.

The reason for this is, any time these requirements change at the service end, they would necessitate client code changes as well; which are an overhead here. This validation information communicated over exception message would be more efficient in this case.

Can you share the complete exception message that you received from the client (on passing invalid input)? We can certainly inspect the client to ensure that we've not truncated any relevant information.

tomkerkhove:
That's what I was thinking as well but in often cases the info is being passed around and only the exceptions happen later on but that's ok.

I don't have the exact exception anymore but it was when calling [`UpdateReportedPropertiesAsync`](https://github.com/Azure/azure-iot-sdk-csharp/blob/da91a363fa770fdcccc57921adde2902ac9baeb8/iothub/device/src/DeviceClient.cs#L594) where it threw an `InvalidOperationException`.

I'll create a separate issue for this.

tomkerkhove:
@abhipsaMisra So I presume it's best to close this PR then?

abhipsaMisra:
Yes, I see you've opened a github issue for this; let's use that for tracking purposes.

PR #1504, opened 2020-08-04
Update exception handling logic (and documentation) for amqp

abhipsaMisra:
The readme is here: https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/exceptionReadMe/iothub/device/devdoc/amqpTransportExceptions.md

This PR has since grown to add a couple of other changes - 
- Mark `AmqpErrorCode.DecodeError` and `AmqpErrorCode.FrameSizeTooSmall` as non-retryable exceptions -> mapped to `IotHubException`.
- Remove `DeviceAlreadyExistsException` - service team confirmed that they would never throw this over the device client API.
- Remove `DeviceDisabledException` - this was not referenced anywhere in our code -> there is no error code from service that was mapped to this exception on the client.
- Add description to each `ConnectionStatusChangeReason` returned by the client.


iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -77,7 +77,7 @@ public static Exception ToIotHubClientContract(AmqpException amqpException)
             }
             else if (Equals(AmqpErrorCode.DecodeError, amqpSymbol))
             {
-                return new AmqpIoTResourceException(message, amqpException);
+                return new IotHubException(message, amqpException);
```
abhipsaMisra, line 80: `DecodeError` is not something that can be fixed on retry; should be mapped to `IotHubException`

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -113,7 +113,7 @@ public static Exception ToIotHubClientContract(AmqpException amqpException)
             }
             else if (Equals(AmqpErrorCode.FrameSizeTooSmall, amqpSymbol))
             {
-                return new AmqpIoTResourceException(message, amqpException);
+                return new IotHubException(message, amqpException);
```
abhipsaMisra, line 116: `FrameSizeTooSmall` is not something that can be fixed on retry; should be mapped to `IotHubException`

iothub/device/devdoc/amqpTransportExceptions.md
```diff
@@ -1,35 +1,58 @@
...
+| com.microsoft:message-lock-lost | DeviceMessageLockLostException | The device client attempted to complete/reject/abandon a received cloud-to-device message, but the lock token was expired (took > 1min after receiving the message to complete/reject/abandon it) | Call `ReceiveAsync()` again to retrieve an updated lock token, and then complete/reject/abandon the message. De-duplication logic wil need to be implemented at the application level |
+| amqp:transaction :unknown-id | IotHubException | InnerException: AmqpException.Error.Condition = AmqpSymbol.TransactionUnknownId | Inspect the exception details, collect logs and contact service |
+| com.microsoft:argument-error | ArgumentException | | Inspect the exception details, collect logs and contact service |
+| com.microsoft:argument-out-of-range | ArgumentOutOfRangeException | | Inspect the exception details, collect logs and contact service |
+| com.microsoft:device-already-exists | DeviceAlreadyExistsException | | Inspect the exception details, collect logs and contact service |
```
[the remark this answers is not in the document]
abhipsaMisra: I was surprised to see this in our device SDK as well, if you can confirm that service will never return this error code to the device client, then I can remove this mapping.

iothub/device/devdoc/amqpTransportExceptions.md
```diff
@@ -1,35 +1,58 @@
...
+| amqp:session:handle-in-use | AmqpIoTResourceException | SDK will retry, this exception will not bubble up to the application directly |  |
+| amqp:session:unattached-handle | AmqpIoTResourceException | SDK will retry, this exception will not bubble up to the application directly |  |
+| amqp:session:window-violation | AmqpIoTResourceException | SDK will retry, this exception will not bubble up to the application directly |  |
+| amqp:internal-error | IotHubCommunicationException | SDK will retry, this exception will not bubble up to the application directly |  |
+| amqp:transaction :rollback | IotHubCommunicationException | SDK will retry, this exception will not bubble up to the application directly |  |
+| amqp:transaction :timeout | IotHubCommunicationException | SDK will retry, this exception will not bubble up to the application directly |  |
+| com.microsoft:timeout | TimeoutException | SDK will retry, this exception will not bubble up to the application directly |  |
+| com.microsoft:device-container-throttled | IotHubThrottledException | SDK will retry, this exception will not bubble up to the application directly | Always generated from an AmqpOutcome |
```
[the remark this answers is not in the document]
abhipsaMisra: The default retry policy is based on exponential backoff, that is what will be used here; so it should fit the requirement

iothub/device/devdoc/amqpTransportExceptions.md
```diff
@@ -1,35 +1,58 @@
...
+| amqp:illegal-state | IotHubException | InnerException: AmqpException.Error.Condition = AmqpSymbol.IllegalState | Inspect the exception details, collect logs and contact service |
+| amqp:not-allowed | InvalidOperationException	| InnerException: AmqpException.Error.Condition = AmqpSymbol.NotAllowed | Inspect the exception details, collect logs and contact service |
+| amqp:not-found | DeviceNotFoundException | This is the exception thrown when the error is received as an AmqpException | Verify that your device exits on your IoT hub instance |
+| amqp:not-found | DeviceMessageLockLostException | This is the exception thrown when the error is received as a Rejected outcome for an Amqp `DisposeMessageAsync()` operation | The device client attempted to complete/reject/abandon a received cloud-to-device message, but the lock token was expired (possible cause is that device reconnected after receiving the c2d message). <br/> Call `ReceiveAsync()` again to retrieve an updated lock token, and then complete/reject/abandon the message. De-duplication logic wil need to be implemented at the application level |
```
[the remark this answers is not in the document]
abhipsaMisra: This is what the current logic is based on - if a Dispose operation returns a Rejected outcome and the associated error is amqp:not-found, then the SDK maps it to DeviceMessageLockLostException. This is separate from the SDK receiving `com.microsoft:message-lock-lost` on an amqp outcome directly.
This bring us back to the question that I asked yesterday - in what scenarios does the client receive an outcome of rejected, vs an exception on the link directly?

Could you confirm if this mapping is true from the service's perspective as well?

iothub/device/devdoc/amqpTransportExceptions.md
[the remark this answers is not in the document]
ailn: Yes, I confirm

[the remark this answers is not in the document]
abhipsaMisra: Cool, I'll remove this mapping from our device client then.

iothub/device/devdoc/amqpTransportExceptions.md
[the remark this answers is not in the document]
ailn: I cannot confirm this. In our code I don not see this. We do respond with Rejected outcome on Message dispose but we do not set amqp:not-found.
... I'd need to run some tests to understand better why you end up getting amqp:not-found.
The only theory I have is: we close the link on MessageLockLost. Probably this is the cause. But unlikely..

iothub/device/devdoc/amqpTransportExceptions.md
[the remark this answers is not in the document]
ailn: Yes, I confirm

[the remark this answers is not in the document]
abhipsaMisra: @ailn  This exception is also mapped to status code 409 (conflict) on Http. Can we also confirm that service doesn't throw 409 to device client?

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -25,7 +25,6 @@ internal static class AmqpIoTErrorAdapter
         public static readonly AmqpSymbol IotHubNotFoundError = AmqpIoTConstants.Vendor + ":iot-hub-not-found-error";
         public static readonly AmqpSymbol ArgumentError = AmqpIoTConstants.Vendor + ":argument-error";
         public static readonly AmqpSymbol ArgumentOutOfRangeError = AmqpIoTConstants.Vendor + ":argument-out-of-range";
-        public static readonly AmqpSymbol DeviceAlreadyExists = AmqpIoTConstants.Vendor + ":device-already-exists";
```
abhipsaMisra, line 28: Service team has confirmed that they never throw this exception to the device client

iothub/device/src/Common/Exceptions/DeviceAlreadyExistsException.cs
```diff
@@ -1,56 +0,0 @@
...
-using System.Runtime.Serialization;
-using Microsoft.Azure.Devices.Client.Extensions;
-
-namespace Microsoft.Azure.Devices.Client.Exceptions
-{
-    /// <summary>
-    /// The exception that is thrown when an attempt to create a device fails because it already exists.
-    /// </summary>
-    [Serializable]
-    public sealed class DeviceAlreadyExistsException : IotHubException
```
abhipsaMisra, line 14: This type was not referenced anywhere in the SDK -> no exception returned by service was mapped to this type.
@prmathur-microsoft , @sharmasejal This is a public type that we are removing.

iothub/device/src/Common/Exceptions/DeviceDisabledException.cs
```diff
@@ -1,79 +0,0 @@
...
-
-namespace Microsoft.Azure.Devices.Client.Exceptions
-{
-    // TODO: #707 - This exception is not thrown by any protocol.
-
-    /// <summary>
-    /// The exception that is thrown when the device was disabled from IoT Hub.
-    /// </summary>
-    [Serializable]
-    public sealed class DeviceDisabledException : IotHubException
```
abhipsaMisra, line 16: This type was not referenced anywhere in the SDK -> no exception returned by service was mapped to this type.
@prmathur-microsoft , @sharmasejal This is a public type that we are removing.

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -719,12 +719,6 @@ private void HandleConnectionStatusExceptions(IotHubException hubException)
             {
                 status = ConnectionStatusChangeReason.Bad_Credential;
             }
-            else if (hubException is DeviceDisabledException)
```
abhipsaMisra, line 722: There was a pending TODO to remove this exception mapping

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,55 @@ namespace Microsoft.Azure.Devices.Client
...
+        /// This connection status change reason is returned with a connection status of <see cref="ConnectionStatus.Connected"/>.
+        /// </summary>
         Connection_Ok,
+
+        /// <summary>
+        /// This signifies that the SAS token associated with the client has expired, and cannot be renewed.
+        /// The supplied credentials need to be fixed before a connection can be established.
+        /// NOTE: This status change reason is currently not used in the client library.
+        /// </summary>
         Expired_SAS_Token,
```
abhipsaMisra, line 22: This status change reason is currently unused in our SDK

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,55 @@ namespace Microsoft.Azure.Devices.Client
...
+        /// This signifies that the client was disconnected due to a transient exception, but the retry policy expired before a connection could be re-established.
+        /// This connection status change reason is returned with a connection status of <see cref="ConnectionStatus.Disconnected"/>
+        /// </summary>
         Retry_Expired,
+
+        /// <summary>
+        /// This signifies that the client was disconnected due to loss of network.
+        /// NOTE: This status change reason is currently not used in the client library.
+        /// </summary>
         No_Network,
```
abhipsaMisra, line 48: This status change reason is currently unused in our SDK

iothub/device/src/ConnectionStatus.cs
```diff
@@ -20,12 +20,10 @@ public enum ConnectionStatus
         /// </summary>
         Connected,
 
-#pragma warning disable CA1707 // Identifiers should not contain underscores - Reason: public API cannot be changed.
         /// <summary>
         /// The device is attempting to reconnect.
         /// </summary>
         Disconnected_Retrying,
-#pragma warning restore CA1707 // Identifiers should not contain underscores - Reason: public API cannot be changed.
```
prmathur-microsoft, line 28: Why was this removed? Don't we get warning on this anymore?

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,55 @@ namespace Microsoft.Azure.Devices.Client
...
+        /// <summary>
+        /// This signifies that the client is connected, and ready to be used.
+        /// This connection status change reason is returned with a connection status of <see cref="ConnectionStatus.Connected"/>.
+        /// </summary>
         Connection_Ok,
+
+        /// <summary>
+        /// This signifies that the SAS token associated with the client has expired, and cannot be renewed.
+        /// The supplied credentials need to be fixed before a connection can be established.
+        /// NOTE: This status change reason is currently not used in the client library.
```
prmathur-microsoft, line 20: Even when user supplied sas token is expired? What do we throw then?

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,55 @@ namespace Microsoft.Azure.Devices.Client
...
+        /// <summary>
+        /// This signifies that the client was disconnected due to loss of network.
+        /// NOTE: This status change reason is currently not used in the client library.
+        /// </summary>
         No_Network,
+
+        /// <summary>
+        /// This connection status change reason when returned with a connection status of <see cref="ConnectionStatus.Disconnected_Retrying"/> signifies that the client is trying to recover from a disconnect due to a transient exception.
+        /// This connection status change reason when returned with a connection status of <see cref="ConnectionStatus.Disconnected"/> signifies that client is disconnected due to a non-retryable exception.
+        /// </summary>
```
prmathur-microsoft, line 53: Can you also add what action should user take when they end up in this state for both connection status and this reason?

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,55 @@ namespace Microsoft.Azure.Devices.Client
...
         Communication_Error,
+
+        /// <summary>
+        /// This signifies that the client has been closed gracefully.
+        /// This connection status change reason is returned with a connection status of <see cref="ConnectionStatus.Disabled"/>
+        /// </summary>
```
prmathur-microsoft, line 59: Same here as well. Can you add user action on this status:reason combination nation?

iothub/device/src/GlobalSuppressions.cs
```diff
@@ -5,5 +5,9 @@
 // Project-level suppressions either have no target or are given
 // a specific target and scoped to a namespace, type, member, etc.
 
-[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Globalization", "CA1303:Do not pass literals as localized parameters", Justification = "Not localizing", Scope = "module")]
-[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "SDK hides non-actionable errors from user", Scope = "module")]
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Globalization", "CA1303:Do not pass literals as localized parameters", Justification = "Not localizing", Scope = "module")]
+[assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "SDK hides non-actionable errors from user", Scope = "module")]
+[assembly: SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "Public API cannot be changed.", Scope = "type", Target = "~T:Microsoft.Azure.Devices.Client.ConnectionStatusChangeReason")]
```
prmathur-microsoft, line 12: Ignore previous. Saw this later.

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
prmathur-microsoft: What would happen if you create two clients with same device ID?

iothub/device/src/Transport/RetryDelegatingHandler.cs
prmathur-microsoft: This was a specific ask from mcvp. Can you double check if this is not used by them?

bikamani:
> The readme is here: https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/exceptionReadMe/iothub/device/devdoc/amqpTransportExceptions.md

Can you update descriptions to elaborate these changes such as if it is cleaning up unused exceptions etc?

iothub/device/src/Common/Exceptions/DeviceDisabledException.cs
```diff
@@ -1,79 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-using System;
-using System.Runtime.Serialization;
-using Microsoft.Azure.Devices.Client.Extensions;
-
-namespace Microsoft.Azure.Devices.Client.Exceptions
-{
-    // TODO: #707 - This exception is not thrown by any protocol.
```
bikamani, line 10: Can we remove it when it's public? Also what is this todo for?

iothub/device/src/ConnectionStatusChangeReason.cs
bikamani: so what's the reason behind keeping this or the Expired_SAS_Token?

iothub/device/src/Transport/RetryDelegatingHandler.cs
abhipsaMisra: Sure, I will check with them.

iothub/device/src/ConnectionStatusChangeReason.cs
abhipsaMisra: Since this is a publicly exposed enum, removing these would break anyone who has referenced these types in their application code (it would still be dead code for them, but the reference would still exist now).

iothub/device/src/ConnectionStatusChangeReason.cs
abhipsaMisra: I would think Bad_credential, but let me run a quick test to confirm this.

iothub/device/src/ConnectionStatusChangeReason.cs
abhipsaMisra: I can add it the summary, sure. I did not want it becoming too verbose, but I can see code comment being a better reference than a sample/doc.

iothub/device/src/ConnectionStatusChangeReason.cs
abhipsaMisra: sure

iothub/device/src/Common/Exceptions/DeviceDisabledException.cs
abhipsaMisra: The TODO was probably added to ensure that either the mapping is correct (what service throws vs what the sdk throws), or to remove this type.

abhipsaMisra: Ideally we shouldn't be removing any public types/ APIs, but in this case, this was unreferenced code. 
Removing such unreferenced types is going to be more of a case-by-case decision - in this case, since the SDK didn't throw this exception to the calling application, I didn't see a huge impact (a user would have had to scrub through our code to discover this exception, and add a reference to this in their application).

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
abhipsaMisra: They would create the ping-pong situation - one client would create a connection -> other client would want to create another connection -> client 1 gets disconnected -> client 2 gets connection -> and so on.

iothub/device/src/ConnectionStatusChangeReason.cs
abhipsaMisra: Yes, it does get mapped to bad_credential.

iothub/device/devdoc/amqpTransportExceptions.md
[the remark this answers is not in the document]
abhipsaMisra: Confirmed from service that device client API should not throw `DeviceAlreadyExistsException`.

iothub/device/src/ConnectionStatusChangeReason.cs
bikamani: But if this is a dead code, they should be updating their application right? How will it help them keeping for example a switch case for this status which is never set/returned by our sdk?
Do we have any documentation/policy which tells us to keep public enums as well which are not used at all? OR any way to deprecate these so can be removed in future?

abhipsaMisra:
Closing this PR; the changes will be handled in #1506 (non-breaking changes) and #1507 (removing public unreferenced types).

iothub/device/src/ConnectionStatusChangeReason.cs
abhipsaMisra: > But if this is a dead code, they should be updating their application right? How will it help them keeping for example a switch case for this status which is never set/returned by our sdk?

This is, and has been dead code for a while now. The "Note:" added will inform them that the SDK will not return this enum value.

> Do we have any documentation/policy which tells us to keep public enums as well which are not used at all? OR any way to deprecate these so can be removed in future?

Ideally, anything that was being used, and we want to deprecate, we would mark them as `[Obsolete]`. However, for this case, this is more of a bug on the SDK, than this feature being deprecated. 
The SDK SHOULD be able to differentiate between sas token invalid vs sas token expired (especially if the service is able to provide different exceptions for both the cases). 
For the "NO_NETWORK" enum, we SHOULD be using it to inform the user that the SDK is retrying, because of a transient network exception.
This is what the java sdk does currently.

PR #1506, opened 2020-08-07
Update doc for exception handling over amqp, add more summary descriptions etc

abhipsaMisra:
The readme is here: https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/updateDesc/iothub/device/devdoc/amqpTransportExceptions.md

- Mark `AmqpErrorCode.DecodeError` and `AmqpErrorCode.FrameSizeTooSmall` as non-retryable exceptions -> mapped to `IotHubException`.
- Remove `com.microsoft:device-already-exists` error mapping over Amqp - service team confirmed that they would never throw this over the device client API.
- Add description to each `ConnectionStatus` and `ConnectionStatusChangeReason` returned by the client.


iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,59 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public enum ConnectionStatusChangeReason
     {
+        /// <summary>
```
bikamani, line 11: Aren't you already making these changes in your other PR?

abhipsaMisra: Yes, I was; but as we discussed we want to put out non-breaking changes first -> which can then go into the LTS branch.
As a result, #1504 will now be broken into 2 PRs - #1506 - with no public API changes, and another one with public API changes (that I'll put up next).

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -25,7 +25,6 @@ internal static class AmqpIoTErrorAdapter
         public static readonly AmqpSymbol IotHubNotFoundError = AmqpIoTConstants.Vendor + ":iot-hub-not-found-error";
         public static readonly AmqpSymbol ArgumentError = AmqpIoTConstants.Vendor + ":argument-error";
         public static readonly AmqpSymbol ArgumentOutOfRangeError = AmqpIoTConstants.Vendor + ":argument-out-of-range";
-        public static readonly AmqpSymbol DeviceAlreadyExists = AmqpIoTConstants.Vendor + ":device-already-exists";
```
prmathur-microsoft, line 28: You have a confirmation from partners if this is fine to remove?

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -113,7 +112,7 @@ public static Exception ToIotHubClientContract(AmqpException amqpException)
             }
             else if (Equals(AmqpErrorCode.FrameSizeTooSmall, amqpSymbol))
             {
-                return new AmqpIoTResourceException(message, amqpException);
+                return new IotHubException(message, amqpException);
```
bikamani, line 115: so with AmqpIoTResourceException earlier did we keep retrying if this was the error?

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -77,7 +76,7 @@ public static Exception ToIotHubClientContract(AmqpException amqpException)
             }
             else if (Equals(AmqpErrorCode.DecodeError, amqpSymbol))
             {
-                return new AmqpIoTResourceException(message, amqpException);
+                return new IotHubException(message, amqpException);
```
bikamani, line 79: Just wondering when does this happen? 

In the docs I see this - Mismatch between AMQP message sent by client and received by service; collect logs and contact service

What does mismatch mean here?

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
abhipsaMisra: We have confirmation from the service team that they never issue an error of `com.microsoft:device-already-exists` over Amqp, on the device client API: https://github.com/Azure/azure-iot-sdk-csharp/pull/1504#discussion_r465380144

abhipsaMisra: Also, any error code not explicitly mapped by our SDK would be translated to a non-retryable `IotHubException`.

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
abhipsaMisra: It means that the data that the client encodes and sends over the amqp channel is not able to be decoded by the service (assuming that the same algorithm is used for encoding and decoding). This exception will likely originate from the amqp library directly, and not from the service.

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
abhipsaMisra: yes

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
bikamani: So does it always mean an error on client encoding?

iothub/device/src/ConnectionStatusChangeReason.cs
bikamani: got it

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
abhipsaMisra: yes; this is one of the error codes that the amqp lib defines, but we don't expect to see this exception. Also, if we ever see this exception for one operation, we would see it for all operations as well (since this is an encoding-decoding based error code).

iothub/device/devdoc/amqpTransportExceptions.md
```diff
@@ -1,35 +1,57 @@
-Below is the behavior of the SDK on receiving an exception from the AMQP library being used underneath. If the exception is marked as retryable, the SDK will implement the default retry-policy and attempt to reconnect. For exceptions not marked as retryable, it is advised to inspect the exception details and perform the necessary action as indicated below.
+Below is the behavior of the SDK on receiving an exception over Amqp transport protocol. If the exception is marked as retryable, the exception will get translated to an `AmqpIoTResourceException` internally, and the SDK will implement the default retry-policy and attempt to reconnect. For exceptions not marked as retryable, it is advised to inspect the exception details and perform the necessary action as indicated below.
```
vinagesh, line 1: Do we need to specify the implementation details on what exception it gets converted to here? This might get out of sync if we change the name of the exception. Might be better to keep it generic.

iothub/device/devdoc/amqpTransportExceptions.md
```diff
@@ -1,35 +1,57 @@
...
+
+* NOTE - For exceptions marked as retryable, the SDK will implement its retry policy internally, and you do not need to take any action. For non-retryable exceptions, or after the SDK retry policy has expired, you should inspect both the connection status change handler and the returned exception details to determine the next step.
 
-* NOTE - For exceptions marked as retryable, the SDK will implement its retry policy internally, and you do not need to take any action. For non-retryable exceptions, you can inspect the exception details, and if a reconnection makes sense, you should dispose of the existing `DeviceClient` instance and then initialize a new client (initializing a new `DeviceClient` instance without disposing the previously used instance will cause them to fight for the same connection resources).
\ No newline at end of file
+* If the device is in `Connected` state, you can perform subsequent operations on the same client instance.
+* If the device is in `Disconnected_Retrying` state, then the SDK is retrying to recover its connection. Wait until device recovers and reports a `Connected` state, and then perform subsequent operations.
```
vinagesh, line 56: the skd is ~~retrying~~ trying

iothub/device/devdoc/amqpTransportExceptions.md
```diff
@@ -1,35 +1,57 @@
...
+* If the device is in `Disconnected` or `Disabled` state, then the underlying transport layer has been disposed. You should dispose of the existing `DeviceClient` instance and then initialize a new client (initializing a new `DeviceClient` instance without disposing the previously used instance will cause them to fight for the same connection resources).
```
vinagesh, line 57: You should dispose ~~of~~ the

vinagesh: will cause ~~them to fight~~ contention for same connection resources

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,59 @@ namespace Microsoft.Azure.Devices.Client
...
+        /// This signifies that the client is connected, and ready to be used.
```
vinagesh, line 12: I don't think you need "this signifies". It can be similiar to the enum defined in ConnectionStatus.cs - /// <para>The client is connected, and ready to be used.</para>

vinagesh: Same for all other enum values in this file.

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,59 @@ namespace Microsoft.Azure.Devices.Client
...
+        /// <para>This connection status change reason is returned with a connection status of <see cref="ConnectionStatus.Connected"/>.</para>
+        /// </summary>
         Connection_Ok,
+
+        /// <summary>
+        /// This signifies that the SAS token associated with the client has expired, and cannot be renewed.
+        /// The supplied credentials need to be fixed before a connection can be established.
+        /// <para>NOTE: This status change reason is currently not used in the client library.</para>
```
vinagesh, line 20: I would suggest to keep it simple and emphasize it is not used - This is unused in the client library.

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,59 @@ namespace Microsoft.Azure.Devices.Client
...
+        /// </summary>
         Expired_SAS_Token,
+
+        /// <summary>
+        /// This signifies that the device/ module has been deleted or marked as disabled (on your hub instance).
```
vinagesh, line 25: we are trying to use IotHub for consistency everywhere.

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -8,13 +8,59 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public enum ConnectionStatusChangeReason
     {
+        /// <summary>
+        /// This signifies that the client is connected, and ready to be used.
+        /// <para>This connection status change reason is returned with a connection status of <see cref="ConnectionStatus.Connected"/>.</para>
```
vinagesh, line 13: What does this mean? I don't think I followed. Might want to revisit this description for users that have not used IotHub a lot.

iothub/device/devdoc/amqpTransportExceptions.md
abhipsaMisra: When our retry policy expires/ an operation is carried out when the client is disconnected, then the inner exception has this `AmqpIoTResourceException` referenced. In the past customers have asked about this internal exception, so I would like to have this mentioned here.

iothub/device/src/ConnectionStatusChangeReason.cs
abhipsaMisra: I can update this to `IoT Hub instance`, sure

iothub/device/src/ConnectionStatusChangeReason.cs
abhipsaMisra: which part of the description is unclear?

vinagesh: This I did not get - This connection status change reason is returned with a connection status of <see cref="ConnectionStatus.Connected"/>.</para>

abhipsaMisra: The client will receive the connection status change reason as a part of the connection status change callback, the signature of which is:
```csharp
public delegate void ConnectionStatusChangesHandler(ConnectionStatus status, ConnectionStatusChangeReason reason);
```
So, the connection status change reason is always returned with an associated connection status.

Issue #707, opened 2018-11-19
Ensure all fault modes are implemented by all IoT Hub DeviceClient


CIPop:
The following exceptions are not thrown by any Transport:
- `DeviceDisabledException`
- `IotHubNotFoundException`

CIPop:
@yzhong94 @prmathur-microsoft  @tameraw FYI

vinagesh:
These exceptions are not being used in the SDK and can be used if the service throws the corresponding errors back. As these are public and un-used, we are having a conversation to see if it can be removed from our code base instead. The service plans to fix the errors returned in v2 and we can use those in our track2 SDK.

vinagesh:
We will be removing the un-referenced exceptions as we have got the approval for it.

az-iot-builder-01:
@CIPop, @vinagesh, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1507, opened 2020-08-07
fix(iot-device): Remove unreferenced exception mappings

abhipsaMisra:
- Remove `DeviceAlreadyExistsException` - service team confirmed that this exception should never be thrown over the device client API.
- Remove `DeviceDisabledException` - this was not referenced anywhere in our code -> there is no error code from service that was mapped to this exception on the client.
- Remove `IotHubNotFoundException` - this was not referenced anywhere in our code -> there is no error code from service that was mapped to this exception on the client.

Fix for #707


abhipsaMisra:
@vinagesh - FYI: I've added `IotHubNoFoundException` to the list of exception types getting removed from the device client.

vinagesh (approved):

Issue #1502, opened 2020-08-03
SetMethodHandlerAsync throws SemaphoreFullException 


arabelaa:
# Context

- **OS, version, SKU and CPU architecture used:** x64, Windows 10
- **Application's .NET Target Framework :** netcoreapp3.1
- **Device:** Intel NUC
- **SDK version used:** 1.26.0

## Description of the issue
Exception is sometimes thrown when registering direct methods. This doesn't happen all the time. I am afraid something is wrong in a different thread causing the executing thread to release the semaphore twice. I can't reproduce the issue locally, unfortunately. 

```
 private async Task RegisterDirectMethodsAsync()
        {
           //_directMethodRegistrations is a dictionary that contains the method name and the callback 
            await Task.WhenAll(_directMethodRegistrations.Registrations.Select(pair =>
                _deviceClient.SetMethodHandlerAsync(pair.Key, pair.Value, null)));

            await _deviceClient.SetMethodDefaultHandlerAsync(OnDefaultHandlerCalled, null);
        }

```

```System.AggregateException:
System.Threading.SemaphoreFullException:
   at System.Threading.SemaphoreSlim.Release (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at Microsoft.Azure.Devices.Client.InternalClient+<SetMethodHandlerAsync>d__74.MoveNext (Microsoft.Azure.Devices.Client, Version=1.23.2.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at Microsoft.Azure.Devices.Client.InternalClient+<SetMethodHandlerAsync>d__73.MoveNext (Microsoft.Azure.Devices.Client, Version=1.23.2.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
[...]

PR #1508, opened 2020-08-07
fix(iot-dev): Fix issue where SemaphoreFullException was thrown unexpectedly

timtay-microsoft:
By putting the ```semaphore.wait()``` call inside a try block and the ```semaphore.release()``` within the finally block, the assumption is that the semaphore should always be released regardless of if the semaphore protected operations succeed or not.

The issue here was that there are cases where the finally block is reached without ever successfully grabbing the semaphore in the first place. For instance, if the task calling into setMethodHandlerAsync is cancelled while waiting to grab the semaphore, the finally block will always throw SemaphoreFullException upon release because it never owned the semaphore. 

#1502

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -839,6 +839,7 @@ namespace Microsoft.Azure.Devices.Client
             try
             {
                 if (Logging.IsEnabled) Logging.Enter(this, methodName, methodHandler, userContext, nameof(SetMethodHandlerAsync));
+                cancellationToken.ThrowIfCancellationRequested();
                 await _methodsDictionarySemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
 
                 if (methodHandler != null)
@@ -928,7 +929,7 @@ namespace Microsoft.Azure.Devices.Client
             try
             {
                 if (Logging.IsEnabled) Logging.Enter(this, methodHandler, userContext, nameof(SetMethodDefaultHandlerAsync));
-
+                cancellationToken.ThrowIfCancellationRequested();
                 await _methodsDictionarySemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
                 if (methodHandler != null)
                 {
@@ -1495,6 +1496,7 @@ namespace Microsoft.Azure.Devices.Client
             ValidateModuleTransportHandler("SetInputMessageHandlerAsync for a named output");
             try
             {
+                cancellationToken.ThrowIfCancellationRequested();
                 await _receiveSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
 
                 if (messageHandler != null)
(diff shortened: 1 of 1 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -879,7 +879,18 @@ public async Task SetMethodHandlerAsync(string methodName, MethodCallback method
...
             {
-                _methodsDictionarySemaphore.Release();
+                try
+                {
+                    _methodsDictionarySemaphore.Release();
+                }
+                catch (SemaphoreFullException e)
+                {
+                    // this semaphore is typically grabbed at the start of the method, but if
+                    // this method is cancelled while waiting to grab the semaphore, then this semaphore.release
```
abhipsaMisra, line 889: Good catch!

A question - 
"this method is cancelled while waiting to grab the semaphore" -> does this mean that the cancellation token passed expires before grabbing the semaphore?
would that not throw an OperationCancelledException, or something along those lines?

Also, wouldn't it be better to do a cancellationToken.ThrowIfCancellationRequested instead?

timtay-microsoft: > does this mean that the cancellation token passed expires before grabbing the semaphore?

Yes, or the cancellation token expires during the wait for the semaphore

> would that not throw an OperationCancelledException...?

Regardless of what it throws, these finally blocks will execute. So the idea is that the finally blocks should try to release the semaphore, but shouldn't throw if they didn't have the semaphore to begin with

timtay-microsoft: > wouldn't it be better to do a cancellationToken.ThrowIfCancellationRequested instead?

I don't follow. We give the cancellation token to the semaphore during ```semaphore.WaitAsync(cancellationToken)``` so what control do we have over that?

timtay-microsoft: The added catch statement for SemaphoreFullException basically just allows for the ```semaphore.release()``` call to "always work." If the semaphore was grabbed, then the release is successful and the catch block isn't hit. If the semaphore wasn't grabbed, then the release will throw this SemaphoreFullException and this catch block will prevent it from reaching the user, but importantly the release wasn't necessary anyways since the semaphore wasn't grabbed

abhipsaMisra: so there are 3 possible broad scenarios:
- cancellation token was valid throughout the duration of the method call -> the semaphore was acquired and then released as expceted
- cancellation token expired during method processing -> semaphore acquired, threw OperationCancelledException when the cancellation token expired, semaphore released.
- cancellation token was already expired -> semaphore not acquired, so the release had nothing to release -> threw exception.

My question is around scenario 3 -> if the cancellation token was already expired, wouldn't it make more sense to call ThrowIfCancellationRequested before we use that cancellation token (be it either in passing it into the semaphore, or into any other subsequent method calls).

timtay-microsoft: > wouldn't it make more sense to call ThrowIfCancellationRequested before we use that cancellation token

Just as an addition to these changes? Sure that sounds reasonable. I'll add it

Issue #1496, opened 2020-07-30
[Bug Report] ModuleClient.SetInputMessageHandlerAsync: No connection retry if a module client fails to authenticate after disconnection


and-rewsmith:
# Context

- **OS, version, SKU and CPU architecture used:** <VERSION> (Windows 10 Desktop x64, Ubuntu 15.04 x86, Windows 10 IoT Core arm32, etc.)
Ubuntu 16.04
- **Application's .NET Target Framework :** <NETVERSION> (See https://docs.microsoft.com/en-us/dotnet/standard/frameworks. E.g. netcoreapp2.1, net451, uap10.0, xamarin)
netcoreapp3.1
- **Device:** <DEVICETYPE> (Laptop, Raspberry PI3, Android APIv25 etc.)
Ubuntu 18.04 VM
- **SDK version used:** <VERSION> (Please include the NuGet package version for all involved components)
1.27.3

## Description of the issue

I am a member of the iotedge team. We are developing a custom mqtt broker and have encountered the following issue:
- ModuleClient connects to EdgeHub over mqtt
- ModuleClient.SetInputMessageHandlerAsync is called and messages are processed as expected
- EdgeHub restarts (so ModuleClient loses connection)
- ModuleClient reconnects, but fails auth when sending the connect packet because EdgeHub auth is not ready (this will be fixed in EdgeHub)
- ModuleClient **never tries to reconnect again** and is left in a state where it hangs, appearing to be connected, with no exception thrown

This doesn't block us because we can avoid the original auth error, but I wanted to call this behavior out because it seems like a bug.

## Code sample exhibiting the issue
https://github.com/Azure/iotedge/blob/master/test/modules/TemperatureFilter/Program.cs

timtay-microsoft:
Is this a 100% repro? And could you provide the module client logs for when this happens?

and-rewsmith:
Yes this is a 100% repro. The logs from the module aren't very interesting because it will hang and not throw an exception.
```
[2020-07-30 19:25:04 : Starting Temperature Filter
[2020-07-30 19:25:05.166 INF] TemperatureFilter Main() started.
[2020-07-30 19:25:05.249 INF] Trying to initialize module client using transport type [Mqtt_Tcp_Only].
[2020-07-30 19:25:06.333 INF] Successfully initialized module client of transport type [Mqtt_Tcp_Only].
[2020-07-30 19:25:06.342 DBG] Waiting on shutdown handler to trigger
[...]

PR #1510, opened 2020-08-07
Update MqttIoTHubAdapter to handle different Connect response codes

abhipsaMisra:
We currently treat all Connect return codes which are not "Accepted" as Unauthorized exceptions, and do not retry on them.
This PR modifies the connect ack code process logic to the following:
- `ConnectReturnCode.RefusedServerUnavailable` - should be retried. This is mapped to `IotHubCommunicationException`. 
- `ConnectReturnCode.RefusedNotAuthorized` and `ConnectReturnCode.RefusedBadUsernameOrPassword` - mapped to `UnauthorizedException` and not retried.
- `ConnectReturnCode.RefusedIdentifierRejected` and `ConnectReturnCode.RefusedUnacceptableProtocolVersion` - mapped to `IotHubException` and no retried.

This does modify the application behavior -> previously for all status code != Accepted, we would thrown an `UnauthorizedException` (non-retryable). Now, we would throw either of `UnauthorizedException` (non-retryable), `IotHubException` (non-retryable), or `IotHubCommunicationException` (retryable).

Fix for #1496

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -40,9 +40,13 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             Closed = 16
         }
 
-        private const string DeviceCommandTopicFilterFormat = "devices/{0}/messages/devicebound/#";
+        // Topic names for sending cloud-to-device messages
         private const string DeviceTelemetryTopicFormat = "devices/{0}/messages/events/";
         private const string ModuleTelemetryTopicFormat = "devices/{0}/modules/{1}/messages/events/";
+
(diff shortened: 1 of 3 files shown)


iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -43,7 +43,6 @@ private enum StateFlags
         private const string DeviceCommandTopicFilterFormat = "devices/{0}/messages/devicebound/#";
         private const string DeviceTelemetryTopicFormat = "devices/{0}/messages/events/";
         private const string ModuleTelemetryTopicFormat = "devices/{0}/modules/{1}/messages/events/";
-        private const string ApiVersionParam = "api-version";
```
abhipsaMisra, line 46: This variable was unreferenced in code.

PR #1512, opened 2020-08-11
Trying to fix unit test failure

vinagesh:
## Checklist

## Description of the changes
Trying to use the same CancellationToken object in the mock and the calls in this test. This could be causing the test to be finding a wrong mock at times. This is definitely safer than earlier code.

### DIFF
--- a/iothub/device/tests/RetryDelegatingHandlerImplicitOpenTests.cs
+++ b/iothub/device/tests/RetryDelegatingHandlerImplicitOpenTests.cs
@@ -83,14 +83,15 @@ namespace Microsoft.Azure.Devices.Client.Test
         {
             var contextMock = Substitute.For<IPipelineContext>();
             var innerHandlerMock = Substitute.For<IDelegatingHandler>();
-            innerHandlerMock.OpenAsync(CancellationToken.None).Returns(t => TaskHelpers.CompletedTask);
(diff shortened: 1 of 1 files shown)


iothub/device/tests/RetryDelegatingHandlerImplicitOpenTests.cs
```diff
@@ -83,14 +83,15 @@ public async Task Retry_CloseAsyncSuccessDisposesHandler_Throws()
...
+            innerHandlerMock.CloseAsync(cancellationToken).Returns(t => TaskHelpers.CompletedTask);
 
             var sut = new RetryDelegatingHandler(contextMock, innerHandlerMock);
-            await sut.OpenAsync(CancellationToken.None).ConfigureAwait(false);
-            await sut.CloseAsync(CancellationToken.None).ConfigureAwait(false);
+            await sut.OpenAsync(cancellationToken).ConfigureAwait(false);
+            await sut.CloseAsync(cancellationToken).ConfigureAwait(false);
 
-            var cancellationToken = new CancellationToken();
             await ((Func<Task>)(() => sut.OpenAsync(cancellationToken))).ExpectedAsync<ObjectDisposedException>().ConfigureAwait(false);
```
vinagesh, line 95: Previously we were trying to call the mock with a new object. Trying to use the same object so that the test definitely uses the correct mock.

PR #1516, opened 2020-08-11
Fix for fault injection tests running concurrently

vinagesh:
## Checklist

### DIFF
--- a/e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
+++ b/e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
@@ -181,7 +181,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
                         else
                         {
                             // 2 is the minimum notification count: connect, disable.
-                            Assert.IsTrue(amqpConnectionStatuses[i].ConnectionStatusChangeCount == 2, $"The expected connection status change count for {testDevices[i].Id}  should be 2 but was {amqpConnectionStatuses[i].ConnectionStatusChangeCount}");
+                            Assert.IsTrue(amqpConnectionStatuses[i].ConnectionStatusChangeCount >= 2, $"The expected connection status change count for {testDevices[i].Id}  should be 2 but was {amqpConnectionStatuses[i].ConnectionStatusChangeCount}");
                         }
                     }
                     Assert.AreEqual(ConnectionStatus.Disabled, amqpConnectionStatuses[i].LastConnectionStatus, $"The expected connection status should be {ConnectionStatus.Disabled} but was {amqpConnectionStatuses[i].LastConnectionStatus}");


e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
```diff
@@ -175,12 +175,12 @@ public static class FaultInjectionPoolingOverAmqp
                     {
                         if (FaultInjection.FaultShouldDisconnect(faultType))
                         {
-                            // 4 is the minimum notification count: connect, fault, reconnect, disable.
-                            Assert.IsTrue(amqpConnectionStatuses[i].ConnectionStatusChangeCount >= 4, $"The expected connection status change count for {testDevices[i].Id} should equals or greater than 4 but was {amqpConnectionStatuses[i].ConnectionStatusChangeCount}");
+                            // Notifications: connect, fault, reconnect, disable.
```
vinagesh, line 178: I am not sure why we were checking >= earlier. I think we should be checking for an exact number here. Let me know if there was a reason to have >=.

e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
```diff
@@ -51,7 +51,7 @@ public static class FaultInjectionPoolingOverAmqp
             logger.Trace($">>> {nameof(FaultInjectionPoolingOverAmqp)} Initializing Device Clients for multiplexing test.");
             for (int i = 0; i < devicesCount; i++)
             {
-                TestDevice testDevice = await TestDevice.GetTestDeviceAsync(logger, $"{devicePrefix}_{i}_").ConfigureAwait(false);
+                TestDevice testDevice = await TestDevice.GetTestDeviceAsync(logger, $"{devicePrefix}_{Guid.NewGuid()}").ConfigureAwait(false);
```
vinagesh, line 54: This should create unique devices across parallel tests, which should help fix the concurrency issues.

abhipsaMisra: this is just the prefix to the device Id that we send here, the index `i` is to help identify which of the multiplexed devices got faulted. 
`TestDevice.GetTestDeviceAsync()` implementation already appends a prefix to the device name, so uniqueness is automatically ensured.

This change needs to be reverted.

e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
abhipsaMisra: The reason we are checking >= count is because the device could get faulted due to some network instability etc (prior to any fault being injected). This ensures that the min no of connection change events take place.

This change needs to be reverted as well.

vinagesh: Thanks for this clarification.

e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
vinagesh: hmmm. Then maybe I don't understand something here. How do we ensure we only check the amqpConnectionStatuses for the specific devices ? Or is that not per device? How can I ensure we only check the amqpConnectionStatuses for this run and set of devices?

e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
vinagesh: By that logic, do we need >= 2 in the check below then? Is that the reason for failure and not really device uniqueness?

abhipsaMisra: It's possible; there could always be some amount of instability due to network disconnects, and I remember we had a retry logic within these tests (at some point). I am not sure if we still have that logic, but personally, I would have preferred to have these tests retry, rather than assert with a >= check.

e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
abhipsaMisra: you use the list of test devices `testDevices`, which are unique to each test run.

PR #1520, opened 2020-08-13
feat(e2e-tests): Add fault injection tests with proxy enabled

abhipsaMisra:


iothub/device/src/ITransportSettings.cs
```diff
@@ -20,5 +21,10 @@ public interface ITransportSettings
         /// The default receive timeout.
         /// </summary>
         TimeSpan DefaultReceiveTimeout { get; }
+
+        /// <summary>
+        /// The proxy settings to be used for communication.
+        /// </summary>
+        IWebProxy Proxy { get; set; }
```
abhipsaMisra, line 28: `Proxy` is already defined in each of the transport settings that implement `ITransportSettings`. I am defining it in the interface as well, so that it can be easily initialized and tested.
This does not change the API surface.

PR #1531, opened 2020-08-21
Update preview with changes from master

abhipsaMisra:
There have been a couple of changes made to the e2e test project in `master`, this PR should bring those changes over to the `preview` branch as well.

Commit b979f5dd007acb8785cdb1b81622dfc1d709d763 (the last commit) has been manually added to disable twin array tests for reported property updates - The device client API version used on preview branch (2019-07-01-preview) is lower than that used on master branch (2019-10-01).
Array support for twin properties is available on API version >= 2019-10-01.


abhipsaMisra:
> The device client twin tests with sending reported property updates as an array are failing: Twin_DeviceSetsReportedPropertyArrayAndGetsItBack_Mqtt.
> This will need to be investigated.

The device client API version used on preview branch (2019-07-01-preview) is lower than that used on master branch (2019-10-01). 
Array support for twin properties is available on API version >= 2019-10-01.

bikamani:
> There have been a couple of changes made to the e2e test project in `master`, this PR should bring those changes over to the `preview` branch as well.

So we are doing merge and not rebase here? Are there only specific commits we are trying to bring from master

abhipsaMisra:
> > There have been a couple of changes made to the e2e test project in `master`, this PR should bring those changes over to the `preview` branch as well.
> 
> So we are doing merge and not rebase here? Are there only specific commits we are trying to bring from master

Yes, since we diverged from rebase sometime in the past, we have been doing merges for bringing changes from master to preview.

bikamani:
> Yes, since we diverged from rebase sometime in the past, we have been doing merges for bringing changes from master to preview.

Ok but are we getting all the commits or filtering out some?

abhipsaMisra:
> > Yes, since we diverged from rebase sometime in the past, we have been doing merges for bringing changes from master to preview.
> 
> Ok but are we getting all the commits or filtering out some?

All commits.

abhipsaMisra:
FYI - device client being able to receive a twin array as a property update/ on a getTwin() call - even tough it is on an older API version, is expected:
If a customer uses a newer API version on the service side to set twin array, but then uses an older API version with the device client -> they can still receive the updated twin (on a get twin call, or as desired property update), but won't be able to send an array as a reported property update.

This is because we should not remove customer's data on a GET, like in this instance, where they create array values using a new api-version, but do a GET using an old api-version.

PR #1532, opened 2020-08-21
feat(e2e): Ensure AI logging is setup for E2E tests

vinagesh:
…pipelines.

## Checklist

## Description of the changes
We needed this so that the pipelines do not crash. Now that all pipelines are on-boarded, we can remove this. The advantage or removing is that in case any pipeline is missing this, it will throw and error and we will know to fix it.

Note: Everyone will need to run the script to get their own instance of AI to log to while running tests.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -187,6 +187,13 @@ Function RunTests($message, $framework = "*", $filterTestCategory = "*")
     Set-Location $rootDir
 
     Write-Host "Invoking expression: $runTestCmd ----------"
+
+    # Ensure the E2E tests can log to Application Insights
+    if ([string]::IsNullOrWhiteSpace($env:E2E_IKEY))
+    {
+        throw "Application Insights is not configured for logging in E2E tests."
+    }
+
     Invoke-Expression $runTestCmd
 
     if ($LASTEXITCODE -ne 0)


abhipsaMisra:
I think we should keep the dependency on application insights optional. The reason is, it will make utilizing our E2E test project easier for external customers, and we do have external customers who run the test suite.

vinagesh:
> I think we should add setting up an AI instance in the script, but still have tests treat it as optional. Then we could set this up for our pipelines, but external customers are free to run the tests with whatever minimum setup their scenario requires.

I agree with that but we need to be able to ensure that the pipelines are setup for it. It is easy to just think it is setup if it is optional. One option could be to add a unit test to ensure the iKey is setup. I just don't want to make it options without having a way to ensure it is not disabled on pipelines.

vinagesh:
> > I think we should add setting up an AI instance in the script, but still have tests treat it as optional. Then we could set this up for our pipelines, but external customers are free to run the tests with whatever minimum setup their scenario requires.
> 
> I agree with that but we need to be able to ensure that the pipelines are setup for it. It is easy to just think it is setup if it is optional. One option could be to add a unit test to ensure the iKey is setup. I just don't want to make it optional without having a way to ensure it is not disabled on pipelines. And I believe when I originally had the PR out, you wanted the same outcome as well - have some failure message pop up if AI is not setup.

PR #1537, opened 2020-08-25
refactor(e2e): Address gap in registry manager

timtay-microsoft:
Bulk device and twin APIs were missing an e2e test

### DIFF
--- a/e2e/test/iothub/service/RegistryManagerE2ETests.cs
+++ b/e2e/test/iothub/service/RegistryManagerE2ETests.cs
@@ -92,12 +92,10 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
         public async Task RegistryManager_BulkLifecycle()
         {
             int bulkCount = 50;
-            string[] deviceIds = new string[bulkCount];
             List<Device> devices = new List<Device>();
             for (int i = 0; i < bulkCount; i++)
             {
-                deviceIds[i] = _devicePrefix + Guid.NewGuid();
-                devices.Add(new Device(deviceIds[i]));
+                devices.Add(new Device(_devicePrefix + Guid.NewGuid()));
             }
 
             var registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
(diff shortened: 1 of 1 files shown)


e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -87,6 +88,62 @@ public async Task RegistryManager_AddDeviceWithTwinWithDeviceCapabilities()
             }
         }
 
+        [LoggedTestMethod]
+        public async Task RegistryManager_BulkLifecycle()
+        {
+            int bulkCount = 50;
+            string[] deviceIds = new string[bulkCount];
```
vinagesh, line 95: This is not really needed. You can just iterate over devices and get the deviceId from that in the for loop below.

timtay-microsoft: True

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -87,6 +88,62 @@ public async Task RegistryManager_AddDeviceWithTwinWithDeviceCapabilities()
...
+            var registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+
+            // Test that you can create devices in bulk
+            var bulkAddResult = await registryManager.AddDevices2Async(devices).ConfigureAwait(false);
+            Assert.IsTrue(bulkAddResult.IsSuccessful);
+
+            foreach (Device device in devices)
+            {
+                // After a bulk add, every device should be able to be retrieved
+                Assert.IsNotNull(await registryManager.GetDeviceAsync(device.Id).ConfigureAwait(false));
```
vinagesh, line 112: This is totally fine but if we are following fluent assertions in general, we should be consistent and use it. Ex: .Should().Be()

timtay-microsoft: We don't really use that pattern in this repo for whatever reason. I'd rather keep the pattern here

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -87,6 +88,62 @@ public async Task RegistryManager_AddDeviceWithTwinWithDeviceCapabilities()
...
+
+            // Test that you can delete device identities in bulk
+            var bulkDeleteResult = await registryManager.RemoveDevices2Async(devices, true, default).ConfigureAwait(false);
+
+            Assert.IsTrue(bulkDeleteResult.IsSuccessful);
+
+            foreach (Device device in devices)
+            {
+                // After a bulk delete, every device should not be found
+                Assert.IsNull(await registryManager.GetDeviceAsync(device.Id).ConfigureAwait(false));
```
vinagesh, line 143: Will this not throw 404?

timtay-microsoft: Surprisingly, it doesn't. I was expecting it to, but it just returns null

timtay-microsoft: I'm not looking to change the behavior of the API at this point, though

vinagesh: Ohh really? That is surprising to me :) Ofcourse we cannot change the behavior of the API but I was expecting a 404.

PR #1526, opened 2020-08-18
to address issue 1183

chewcw:
## Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/.github/CONTRIBUTING.md).
- [ ] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
- [x] This pull-request is submitted against the `master` branch.

## Description of the changes
- New API for getting Azure Provisioning Service Individual Enrollment Attestation
- New API for getting Azure Provisioning Service Group Enrollment Attestation

## Reference/Link to the issue solved with this PR (if any)

#1183

PR #1539, opened 2020-09-02
Update preview branch to fix GetFileUploadSasURi and Attestation APIs

azabbasi:
fix(bug): Fix get file upload sas uri using certificates. #1534
API for getting DPS enrollment attestation mechanism #1526


bikamani:
> _No description provided._

We usually put a description of the last commit we are bring changes from. Lgtm otherwise :)

PR #1542, opened 2020-09-03
Added SetAsDistributedTracingMessage method on Message

algorni:
This enable distributing tracing also with the C# SDK.  It still miss the logic to automatically inject that system property as per cloud configuration.

## Checklist

### DIFF
--- a/iothub/device/src/Message.cs
+++ b/iothub/device/src/Message.cs
@@ -396,7 +396,7 @@ namespace Microsoft.Azure.Devices.Client
         /// </summary>
         public void SetAsDistributedTracingMessage()
         {
-            Int32 unixTimestamp = (Int32)(DateTime.UtcNow.Subtract(DateTime.UnixEpoch)).TotalSeconds;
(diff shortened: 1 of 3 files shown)


abhipsaMisra:
@algorni Thank you for submitting this PR. Is there a particular scenario for which you need this feature support? 
Distributed tracing is currently a preview feature, and has scope for bringing in breaking changes before it goes GA. For this reason, the use-case needs to be investigated before making any kind of code changes to support this. 
Could you detail the user-scenario for which this support is required? This will help our PM teams prioritize and make a decision on this.

/cc @sharmasejal FYI

algorni:
Hi Misra, two thread here.

1) Having another SDK supporting distributed tracing would be great benefit for the adoption of this feature from a broader audience.

2) The use of Distributed Tracing help me on haivng the information that a device is sending data into log analytics. 
That open to additional monitoring rules (e.g. discover that a device, even if it is connected, is not sending data at all!)

That said, i'm still working o the fork to enable the DeviceClient to react on the Twin updates and sent a DT messages according to the frequency pushed down from IoT Hub if this would help :)

algorni:
@ebertrams is there any way to have this merged?

PR #1544, opened 2020-09-04
LTS_PATCH- fix(bug): Fix get file upload sas uri using certificates. (#1534)

azabbasi:
Currently there is a bug on the Device client when it's instantiated using the x509 authentication.
The ClientCertificate is set on the internal client after the client has been initialized and if the user doesn't provide the clientOption to specify the FileUloadTransferSetting, we will be using the default Http transport handler which will not contain the certificate.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -224,7 +224,7 @@ try
 {
     if ($sign)
     {
-        if ($localPackagesAvailableForTesting)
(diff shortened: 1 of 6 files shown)


abhipsaMisra (approved):

bikamani (approved):
lgtm

Issue #1535, opened 2020-08-23
AMQP DeviceClient takes very long to detect a lost connection


kiranpradeep:
- **OS, version, SKU and CPU architecture used:** Ubuntu 18.04 amd64
- **Application's .NET Target Framework:** netcoreapp3.1
- **Device:** Laptop
- **SDK version used:** 1.28.0

## Description of the issue
Unlike MQTT DeviceClient (3 minutes), AMQP DeviceClient takes long duration (24 minutes)  to detecting a lost network connection. Docker is used only to simulate network loss and also for consistent repro. Otherwise, it has no other intended relation to the issue.

```console
cd DeviceReconnectionSample
dotnet publish -c Release -o /tmp/out
docker run -it --rm --name=test-sdk --entrypoint=/bin/bash -v /tmp/out:/app mcr.microsoft.com/dotnet/core/runtime:3.1-bionic
$ cd /app
$ export DEBUG=*
$ export IOTHUB_DEVICE_CONN_STRING='HostName=************.azure-devices.net;DeviceId=testub20;SharedAccessKey=***************************************='
$ dotnet DeviceReconnectionSample.dll
```

```console
// after initial connection establishment by the sample (10 seconds), open a second terminal and disconnect docker network 
docker network disconnect bridge test-sdk

```

## Code sample exhibiting the issue
https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/183408/iot-hub/Samples/device/DeviceReconnectionSample

## Console log of the issue

> roots@roots-pc:/tmp/dc$ docker run -it --rm --name=test-sdk --entrypoint=/bin/bash -v /tmp/out:/app mcr.microsoft.com/dotnet/core/runtime:3.1-bionic
> root@81a52de9c59f:/# cd /app
> root@81a52de9c59f:/app# export DEBUG=*
> root@81a52de9c59f:/app# export IOTHUB_DEVICE_CONN_STRING='HostName=************.azure-devices.net;DeviceId=testub20;SharedAccessKey=***************************************='
> root@81a52de9c59f:/app# dotnet DeviceReconnectionSample.dll
> 2020-08-23T15:25:32.4724197>> Debug - Attempting to initialize the client instance, current status=Disconnected
> 2020-08-23T15:25:32.5317887>> Debug - Initialized the client instance.
> 2020-08-23T15:25:32.5354899>> Information - Device waiting for C2D messages from the hub - for 00:00:05...
[...]

PR #1545, opened 2020-09-04
fix(iot-device): Set a default idle timeout of 2mins over amqp

abhipsaMisra:
The device client library currently only sets the idle timeout over Amqp if the user supplies a valid Timespan via AmpqTransportSettings.IdleTimeout. This PR updates the behavior to set a default IdleTimeout of 2 minutes.
Without this, if the user doesn't set an IdleTimeout, the client will always rely on either the server or the underlying OS to inform it of disconnection (which can potentially have delays).

Related #1535

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             Closed = 16
         }
 
-        // Topic names for sending cloud-to-device messages
+        // Topic names for sending device-to-cloud messages
         private const string DeviceTelemetryTopicFormat = "devices/{0}/messages/events/";
         private const string ModuleTelemetryTopicFormat = "devices/{0}/modules/{1}/messages/events/";
 
@@ -397,6 +397,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
         private static async void PingServer(object ctx)
         {
+            if (Logging.IsEnabled) Logging.Enter(ctx, $"Scheduled check to see if a ping request should be sent.", nameof(PingServer));
+
             var context = (IChannelHandlerContext)ctx;
             try
             {
@@ -408,6 +410,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                 }
 
                 TimeSpan idleTime = DateTime.UtcNow - self._lastChannelActivityTime;
+                if (Logging.IsEnabled) Logging.Info(context, $"Idle time currently is {idleTime}.", nameof(PingServer));
 
                 if (idleTime > self._pingRequestInterval)
                 {
@@ -434,6 +437,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             {
                 ShutdownOnError(context, ex);
             }
+
+            if (Logging.IsEnabled) Logging.Exit(ctx, $"Scheduled check to see if a ping request should be sent.", nameof(PingServer));
         }
 
         private async Task ProcessConnectAckAsync(IChannelHandlerContext context, ConnAckPacket packet)
--- a/iothub/device/tests/TransportSettingsTests.cs
+++ b/iothub/device/tests/TransportSettingsTests.cs
@@ -2,8 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Security.Cryptography.X509Certificates;
-using Microsoft.Azure.Devices.Client.ApiTest;
 using Microsoft.Azure.Devices.Client.Transport.Mqtt;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
@@ -146,7 +144,7 @@ namespace Microsoft.Azure.Devices.Client.Test
         }
 
         [TestMethod]
-        public void AmqpTransportSettings_TimeoutDefaultsAndOverrides()
+        public void AmqpTransportSettings_SetsDefaultTimeout()
         {
             // act
             var transportSetting = new AmqpTransportSettings(TransportType.Amqp_WebSocket_Only, 200);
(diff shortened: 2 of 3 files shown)


iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -40,7 +40,7 @@ private enum StateFlags
             Closed = 16
         }
 
-        // Topic names for sending cloud-to-device messages
+        // Topic names for sending device-to-cloud messages
```
abhipsaMisra, line 43: This comment description was incorrect

kiranpradeep:
@abhipsaMisra Isn't 1 minute an aggresive option which might result in too many timeouts? It look like azure-iot-sdk-c uses 4 [minutes](https://github.com/Azure/azure-iot-sdk-c/blob/cf2b33b6888c2322647737852499a0d7e80fcde7/iothub_client/inc/iothub_client_options.h#L50). That library also had some [ratio](https://github.com/Azure/azure-iot-sdk-c/blob/cf2b33b6888c2322647737852499a0d7e80fcde7/iothub_client/inc/iothub_client_options.h#L61) to be considered.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
drwill-ms: yikes

drwill-ms (approved):

abhipsaMisra:
> @abhipsaMisra Isn't 1 minute an aggresive option which might result in too many timeouts? It look like azure-iot-sdk-c uses 4 [minutes](https://github.com/Azure/azure-iot-sdk-c/blob/cf2b33b6888c2322647737852499a0d7e80fcde7/iothub_client/inc/iothub_client_options.h#L50). That library also had some [ratio](https://github.com/Azure/azure-iot-sdk-c/blob/cf2b33b6888c2322647737852499a0d7e80fcde7/iothub_client/inc/iothub_client_options.h#L61) to be considered.

Thanks for the inputs @kiranpradeep , this is definitely something for us to consider.

drwill-ms:
> The device client library currently only sets the idle timeout over Amqp if the user supplies a valid Timespan via AmpqTransportSettings.IdleTimeout. This PR updates the behavior to set a dfault IdleTimeout of 1 minutes.
> Without this, if the user doesn't set an IdleTimeout, the client will always rely on either the server or the underlying OS to inform it of disconnection (which can potentially have delays).
> 
> Related #1535

So what are the effective defaults before this change then? Note this is a behavior change, which we need to treat as a breaking change. I'll ping you on Teams to discuss how we go about choosing to do that or not.

abhipsaMisra:
The previous behavior was the the idle timeout would be set only if the user sets this property on transport settings. If the idle timeout is not set, then the client no longer monitors for connection loss. This was resulting in the following behavior where ~20mins would go by without the disconnection having been detected.

On setting a default idle timeout of 2 mins, the client will now exhibit the following behavior:
- the client informs the service that it's idle timeout is 2 mins.
- as per amqp specs, the service now needs to respond with a ping at least once every 2 mins.
- in the event that 2mins go by without any communication, the amqp library at the client side will trigger a disconnect.
- the sdk will see a disconnection, and retry based on it's retry policy.

The reason for selecting an idle timeout of 2 mins is to ensure that a disconnection event is triggered before the default operation timeout is reached (4 minutes). Otherwise, if we set the default idle timeout for > 4minutes, an operation would timeout before a disconnection event was detected (i.e. in the event of a disconnection, the operation would be stuck for 4 minutes and then timeout, and then return the disconnection event). On setting it to < 4 minutes, a disconnection event would first be detected, and then the operation would timeout.

Note that this is different from the heartbeat protocol followed by Mqtt, where on establishing a keep-alive of 5mins from the client to the service, the client was responsible for sending across a ping request (which it would send if it had been idle for > (5mins/4 = 75 secs) and monitoring the ping response (wait for a max of 30 secs)). This resulted in the client taking about 2-3mins to detect a disconnection.

@kiranpradeep  We are unable to follow the c-sdk in taking into account this ratio (ping interval is 1/2 of the keep-alive interval). This is because the ping request-response logic is encapsulated by the azure-amqp library being used underneath, and those hooks are not available to us to modify.

**The effect of this change would be that the device client would be more proactive in detecting connection losses over Amqp.**
- If disconnection-reconnection duration is short enough (less than 2mins) = does not report any change in status = no change in behavior.
- If the disconnection-reconnection duration is > 2mins and < 20 mins (on Linux) = previously the device would simply be stuck in disconnected state, but not have reported it to the application. Now the device will return the "disconnected_retrying" status to the connection status change callback handler (after about 2-3 minutes). The device client will still have the responsibility of retrying and re-establishing the connection.
- If the disconnection-reconnection duration is > 20 mins = reports disconnection event = no change in behavior.

iothub/device/tests/TransportSettingsTests.cs
```diff
@@ -146,14 +144,37 @@ public void AmqpTransportSettings_TimeoutPropertiesSet()
         }
 
         [TestMethod]
-        public void AmqpTransportSettings_TimeoutDefaultsAndOverrides()
+        public void AmqpTransportSettings_TimeoutDefaults()
```
bikamani, line 147: I would rename this to AmqpTransportSettings_SetsDefaultTimeout

iothub/device/tests/TransportSettingsTests.cs
```diff
@@ -146,14 +144,37 @@ public void AmqpTransportSettings_TimeoutPropertiesSet()
...
 
             // assert
             Assert.AreEqual(AmqpTransportSettings.DefaultOpenTimeout, transportSetting.OpenTimeout, "Default OpenTimeout not set correctly");
             Assert.AreEqual(AmqpTransportSettings.DefaultOperationTimeout, transportSetting.OperationTimeout, "Default OperationTimeout not set correctly");
+            Assert.AreEqual(AmqpTransportSettings.DefaultIdleTimeout, transportSetting.IdleTimeout, "Default IdleTimeout not set correctly");
+            Assert.AreEqual(AmqpTransportSettings.DefaultOperationTimeout, transportSetting.DefaultReceiveTimeout, "Default DefaultReceiveTimeout not set correctly");
+        }
+
+        [TestMethod]
+        public void AmqpTransportSettings_TimeoutOverrides()
```
bikamani, line 160: and this one to AmqpTransportSettings_OverridesDefaultTimeout

iothub/device/tests/TransportSettingsTests.cs
abhipsaMisra: I can do that, sure!

abhipsaMisra:
Git stopped updating the status, so I had to close it. On attempting to reopen the PR, git complained that there had been a force-push to the feature branch, so it didn't allow me to reopen the PR.
This is now being tracked in #1608

PR #1546, opened 2020-09-04
Fix all constructors to make sure ClientOptions is setup correctly.

azabbasi:
We found instances of the clientFactory constructors that do not ensure options is setup correctly.

### DIFF
--- a/iothub/device/src/ClientFactory.cs
+++ b/iothub/device/src/ClientFactory.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Azure.Devices.Client
 
             IotHubConnectionStringBuilder connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
 
-            // If the file upload transport settings hasn't been specified, we will create one using the client certificate on the connection string
+            // Make sure client options is initialized with the correct transport setting.
             EnsureOptionsIsSetup(connectionStringBuilder.Certificate, ref options);
 
             if (authenticationMethod is DeviceAuthenticationWithX509Certificate)
@@ -137,7 +137,7 @@ namespace Microsoft.Azure.Devices.Client
 
             var connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
 
-            // If the file upload transport settings hasn't been specified, we will create one using the client certificate on the connection string if applicable.
+            // Make sure client options is initialized with the correct transport setting.
             EnsureOptionsIsSetup(connectionStringBuilder.Certificate, ref options);
 
             if (authenticationMethod is DeviceAuthenticationWithX509Certificate)
@@ -387,6 +387,7 @@ namespace Microsoft.Azure.Devices.Client
                 }
             }
 
+            // Make sure client options is initialized with the correct transport setting.
             EnsureOptionsIsSetup(builder.Certificate, ref options);
 
             pipelineBuilder = pipelineBuilder ?? BuildPipeline();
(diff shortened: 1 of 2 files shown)


iothub/device/src/ClientFactory.cs
```diff
@@ -80,27 +81,16 @@ public static InternalClient Create(string hostname, string gatewayHostname, IAu
 
             IotHubConnectionStringBuilder connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
 
+            // If the file upload transport settings hasn't been specified, we will create one using the client certificate on the connection string
+            EnsureOptionsIsSetup(connectionStringBuilder.Certificate, ref options);
```
abhipsaMisra, line 85: This step is valid only if the device auth mechanism is x509 certs, it should be inside this block: `if (authenticationMethod is DeviceAuthenticationWithX509Certificate)`

iothub/device/src/ClientFactory.cs
```diff
@@ -80,27 +81,16 @@ public static InternalClient Create(string hostname, string gatewayHostname, IAu
 
             IotHubConnectionStringBuilder connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
 
+            // If the file upload transport settings hasn't been specified, we will create one using the client certificate on the connection string
```
abhipsaMisra, line 84: The client cert isn't a part of the connection string, it is supplied via the client's authentication mechanism provider.

iothub/device/src/ClientFactory.cs
```diff
@@ -146,6 +136,10 @@ public static InternalClient Create(string hostname, IAuthenticationMethod authe
             }
 
             var connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
+
+            // If the file upload transport settings hasn't been specified, we will create one using the client certificate on the connection string if applicable.
+            EnsureOptionsIsSetup(connectionStringBuilder.Certificate, ref options);
```
abhipsaMisra, line 141: here, and below: this operation is required only when the client's auth mechanism is x509 certs, it wouldn't be relevant for sas token based authentication flows.

iothub/device/src/ClientFactory.cs
azabbasi: we are going to need the options initialized regardless of the auth method.

iothub/device/src/ClientFactory.cs
```diff
@@ -385,6 +379,17 @@ internal static ITransportSettings[] GetTransportSettings(TransportType transpor
                 }
             }
 
+            if (authenticationMethod is DeviceAuthenticationWithX509Certificate)
+            {
+                if (builder.Certificate == null)
+                {
+                    throw new ArgumentException("certificate must be present in DeviceAuthenticationWithX509Certificate");
```
abhipsaMisra, line 386: nit: can we make the exception message descriptive, instead of referencing the `DeviceAuthenticationWithX509Certificate` class type :)

iothub/device/src/ClientFactory.cs
```diff
@@ -394,6 +399,26 @@ internal static ITransportSettings[] GetTransportSettings(TransportType transpor
             return client;
         }
 
+        /// <summary>
+        /// Ensures that the ClientOptions is configured and initialized.
+        /// If a certificate is provided, the fileUploadTransportSettings will use it during initialization.
```
abhipsaMisra, line 404: 👍

abhipsaMisra (approved):

PR #1553, opened 2020-09-10
Add support for DPS provisioning flow for pnp samples

abhipsaMisra:
This PR currently has the DPS flow for the Thermostat sample. Once I get a sign-off on the approach, I'll implement the same for the TemperatureController sample as well.

Passing in env var via command line -> defaults to env variable:
image

### DIFF
--- a/iothub/device/samples/PnpDeviceSamples/Thermostat/Parameter.cs
+++ b/iothub/device/samples/PnpDeviceSamples/Thermostat/Parameter.cs
@@ -1,0 +1,76 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using CommandLine;
+
+namespace Thermostat
+{
+    /// <summary>
+    /// Parameters for the application supplied via command line arguments. It default to environment variables..
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            's',
+            "DeviceProvisioningType",
+            HelpText = "(Required) The provisioning type that will be used for provisioning the device for the sample. Possible values include dps, hubConnectionString (case-insensitive)." +
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_SECURITY_TYPE\".")]
+        public string DeviceProvisioningType { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_SECURITY_TYPE");
+
+        [Option(
+            'p',
+            "PrimaryConnectionString",
+            HelpText = "(Required if DeviceProvisioningType is \"hubConnectionString\"). \nThe primary connection string for the device to simulate." +
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_CONNECTION_STRING\".")]
+        public string PrimaryConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_CONNECTION_STRING");
+
+        [Option(
+            'e',
+            "DpsEndpoint",
+            HelpText = "(Required if DeviceProvisioningType is \"dps\"). \nThe DPS endpoint to use during device provisioning." +
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_DPS_ENDPOINT\".")]
+        public string DpsEndpoint { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_DPS_ENDPOINT");
+
+        [Option(
+            'i',
+            "DpsIdScope",
+            HelpText = "(Required if DeviceProvisioningType is \"dps\"). \nThe DPS ID Scope to use during device provisioning." +
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_DPS_ID_SCOPE\".")]
+        public string DpsIdScope { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_DPS_ID_SCOPE");
+
+        [Option(
+            'd',
+            "DeviceId",
+            HelpText = "(Required if DeviceProvisioningType is \"dps\"). \nThe device registration Id to use during device provisioning." +
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_DPS_DEVICE_ID\".")]
+        public string DeviceId { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_DPS_DEVICE_ID");
+
+        [Option(
+            'k',
+            "DeviceSymmetricKey",
+            HelpText = "(Required if DeviceProvisioningType is \"dps\"). \nThe device symmetric key to use during device provisioning." +
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_DPS_DEVICE_KEY\".")]
+        public string DeviceSymmetricKey { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_DPS_DEVICE_KEY");
+
+        public bool Validate()
+        {
+            if (string.IsNullOrWhiteSpace(DeviceProvisioningType))
+            {
+                throw new ArgumentNullException(nameof(DeviceProvisioningType), "Device provisioning type needs to be specified, please set the environment variable \"IOTHUB_DEVICE_SECURITY_TYPE\"" +
+                    "or pass in \"-s | --DeviceProvisioningType\" through command line.");
+            }
+
+            return (DeviceProvisioningType.ToLowerInvariant()) switch
+            {
+                "dps" => !string.IsNullOrWhiteSpace(DpsEndpoint)
+                                       && !string.IsNullOrWhiteSpace(DpsIdScope)
+                                       && !string.IsNullOrWhiteSpace(DeviceId)
+                                       && !string.IsNullOrWhiteSpace(DeviceSymmetricKey),
+                "hubconnectionstring" => !string.IsNullOrWhiteSpace(PrimaryConnectionString),
+                _ => throw new ArgumentException($"Unrecognized value for device provisioning received: {DeviceProvisioningType}." +
+                        $" It should be either \"dps\" or \"hubConnectionString\" (case-insensitive)."),
+            };
+        }
+    }
+}
--- a/iothub/device/samples/PnpDeviceSamples/Thermostat/Parameter.cs
+++ b/iothub/device/samples/PnpDeviceSamples/Thermostat/Parameter.cs
@@ -13,63 +13,63 @@ namespace Thermostat
     {
         [Option(
             's',
-            "DeviceProvisioningType",
-            HelpText = "(Required) The provisioning type that will be used for provisioning the device for the sample. Possible values include dps, hubConnectionString (case-insensitive)." +
+            "DeviceSecurityType",
+            HelpText = "(Required) The provisioning type that will be used for provisioning the device for the sample. Possible values include dps, connectionString (case-insensitive)." +
             "\nDefaults to environment variable \"IOTHUB_DEVICE_SECURITY_TYPE\".")]
-        public string DeviceProvisioningType { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_SECURITY_TYPE");
+        public string DeviceSecurityType { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_SECURITY_TYPE");
 
         [Option(
             'p',
             "PrimaryConnectionString",
-            HelpText = "(Required if DeviceProvisioningType is \"hubConnectionString\"). \nThe primary connection string for the device to simulate." +
+            HelpText = "(Required if DeviceSecurityType is \"connectionString\"). \nThe primary connection string for the device to simulate." +
             "\nDefaults to environment variable \"IOTHUB_DEVICE_CONNECTION_STRING\".")]
         public string PrimaryConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_CONNECTION_STRING");
 
         [Option(
             'e',
             "DpsEndpoint",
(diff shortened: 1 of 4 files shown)


iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -27,7 +28,21 @@ public class Program
         // DTDL interface used: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/Thermostat.json
         private const string ModelId = "dtmi:com:example:Thermostat;1";
 
+        // This environment variable indicates if DPS or IoT Hub connection string will be used to provision the device.
+        // Expected values: (case-insensitive)
```
bikamani, line 32: shouldn't we take care of casing?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -7,10 +7,11 @@
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Provisioning.Client;
```
bikamani, line 10: Are you planning to do temperature controller sample changes in a separate PR?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: no, it doesn't matter; this is only a switch to indicate which flow is to be demonstrated.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: No, i'll it to this PR, once I get a green from the certification flow.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
bikamani: you are already doing it actually.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
bikamani: 👍

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -27,7 +28,21 @@ public class Program
         // DTDL interface used: https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/Thermostat.json
         private const string ModelId = "dtmi:com:example:Thermostat;1";
 
+        // This environment variable indicates if DPS or IoT Hub connection string will be used to provision the device.
+        // Expected values: (case-insensitive)
+        // "DPS" - The sample will use DPS to provision the device.
+        // "connectionString" - The sample will use IoT Hub connection string to provision the device.
+        private static readonly string s_deviceSecurityType = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_SECURITY_TYPE");
```
abhipsaMisra, line 35: Since it is a requirement to have a similar usage experience across all sdks, all languages are using environment variables, and not language specific command line parsers (even though they are definitely more convenient).

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -27,7 +28,21 @@ public class Program
...
+
+        // Required if IOTHUB_DEVICE_SECURITY_TYPE is set to "connectionString".
         private static readonly string s_deviceConnectionString = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_CONNECTION_STRING");
```
drwill-ms, line 38: Any interest in using the CommandLineParser nuget, like I changed up recently in our samples repo? It isn't that big of a change. https://github.com/Azure-Samples/azure-iot-samples-csharp/commit/6097ef17fe76875c40241136cd45439ca187f21c

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -66,8 +81,36 @@ private static async Task RunSampleAsync()
...
+                        var authMethod = new DeviceAuthenticationWithRegistrySymmetricKey(dpsRegistrationResult.DeviceId, s_deviceSymmetricKey);
+                        InitializeDeviceClientAsync(dpsRegistrationResult.AssignedHub, authMethod);
+                        break;
+                    }
+                    throw new ArgumentException("Required environment variables are not set for DPS flow, please recheck your environment.");
+
+                case "connectionstring":
+                    if (ValidateArgsForIotHubFlow())
+                    {
+                        InitializeDeviceClientAsync(s_deviceConnectionString);
```
drwill-ms, line 105: no await?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: Yes, I did want to do that, but maintaining similar usage experience across all languages was important here. So the decision was taken for all languages to use environment variables.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -66,8 +81,36 @@ private static async Task RunSampleAsync()
...
             s_logger.LogDebug($"Initialize the device client.");
-            InitializeDeviceClientAsync();
+            switch (s_deviceSecurityType.ToLowerInvariant())
+            {
+                case "dps":
+                    if (ValidateArgsForDpsFlow())
+                    {
+                        DeviceRegistrationResult dpsRegistrationResult = await ProvisionDeviceAsync();
+                        var authMethod = new DeviceAuthenticationWithRegistrySymmetricKey(dpsRegistrationResult.DeviceId, s_deviceSymmetricKey);
+                        InitializeDeviceClientAsync(dpsRegistrationResult.AssignedHub, authMethod);
```
drwill-ms, line 97: no await?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: That is bad naming, this method doesn't do anything async; I'll fix this.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: same here as well, shouldn't be called async. I'll fix this

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -66,8 +81,36 @@ private static async Task RunSampleAsync()
...
+                        break;
+                    }
+                    throw new ArgumentException("Required environment variables are not set for DPS flow, please recheck your environment.");
```
abhipsaMisra, line 100: I am thinking about changing this to log error and then exit (as opposed to throwing an exception). Thoughts?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
drwill-ms: Who set that requirement?

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
drwill-ms: We have argument handling all mixed in with the rest of the sample. This is bad form and makes it hard to read. Can you do that checking at program init?

drwill-ms: Here's a solution to have both https://github.com/commandlineparser/commandline/issues/677

drwill-ms: Perhaps not, it seems.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: This was set when we first started writing the samples this summer. The idea was that, if someone where to give the python sample a try, and then decide to try the .NET sample, the on-boarding time should be min, and experience should be similar.
We know that java and .NET have similar experience with these command line parsers, I'll sync up with scripting team to see if they have a similar option. If not, we might have to stick to env var for now.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: I agree, it does make the sample cluttered. I'll decouple the args validation from the main sample.

drwill-ms (approved):

iothub/device/samples/PnpDeviceSamples/Thermostat/Parameter.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    /// <summary>
+    /// Parameters for the application supplied via command line arguments. It default to environment variables..
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            's',
+            "DeviceProvisioningType",
+            HelpText = "(Required) The provisioning type that will be used for provisioning the device for the sample. Possible values include dps, hubConnectionString (case-insensitive)." +
```
abhipsaMisra, line 17: Cannot set the option as "required", since then we cannot dynamically resolve the value if it needs to be sent via environment variables. This is a trade-off in the sample supporting both env var and command line args.

iothub/device/samples/PnpDeviceSamples/Thermostat/Parameter.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (string.IsNullOrWhiteSpace(DeviceProvisioningType))
+            {
+                throw new ArgumentNullException(nameof(DeviceProvisioningType), "Device provisioning type needs to be specified, please set the environment variable \"IOTHUB_DEVICE_SECURITY_TYPE\"" +
+                    "or pass in \"-s | --DeviceProvisioningType\" through command line.");
+            }
+
+            return (DeviceProvisioningType.ToLowerInvariant()) switch
+            {
+                "dps" => !string.IsNullOrWhiteSpace(DpsEndpoint)
+                                       && !string.IsNullOrWhiteSpace(DpsIdScope)
```
drwill-ms, line 67: excessive indenting

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -2,46 +2,61 @@
...
+                    parameters = parsedParams;
+                })
+                .WithNotParsed(errors =>
+                {
+                    Environment.Exit(1);
+                });
+
+            if (!parameters.Validate())
+            {
+                s_logger.LogError("Required parameters are not set, please recheck required variables by running \"dotnet run -- --help\"");
```
drwill-ms, line 39: period after `are not set`, as the next bit is its own sentence.

drwill-ms, line 39: If someone has the compiled sample, and not the csproj, `dotnet run` doesn't help them.

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
```diff
@@ -52,179 +67,82 @@ public static async Task Main(string[] _)
...
-                        // Generate a random value between 5.0°C and 45.0°C for the current temperature reading.
-                        s_temperature = Math.Round(s_random.NextDouble() * 40.0 + 5.0, 1);
-                        temperatureReset = false;
-                    }
-
-                    await SendTemperatureAsync();
-                    await Task.Delay(5 * 1000);
-                }
-            });
+                JsonData = $"{{ \"modelId\": \"{ModelId}\" }}",
```
drwill-ms, line 107: When do we plan on addressing the needing to know this format and hard code?

iothub/device/samples/PnpDeviceSamples/Thermostat/ThermostatSample.cs
```diff
@@ -0,0 +1,200 @@
...
+            }
+        }
+
+        // Send temperature update over telemetry.
+        private async Task SendTemperatureTelemetryAsync()
+        {
+            string telemetryName = "temperature";
+
+            string telemetryPayload = $"{{ \"{telemetryName}\": {_temperature} }}";
+            var message = new Message(Encoding.UTF8.GetBytes(telemetryPayload))
```
drwill-ms, line 176: using?

iothub/device/samples/PnpDeviceSamples/Thermostat/Parameter.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace Thermostat
+{
+    /// <summary>
+    /// Parameters for the application supplied via command line arguments. It default to environment variables..
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            's',
+            "DeviceProvisioningType",
```
drwill-ms, line 16: DpsSecurityType?

drwill-ms (approved):

iothub/device/samples/PnpDeviceSamples/Thermostat/Program.cs
abhipsaMisra: I'll be adding this into a pnp helper that will go into our dps assembly. This helper will be referenced in the `TemperatureController` sample.

iothub/device/samples/PnpDeviceSamples/Thermostat/Parameter.cs
abhipsaMisra: Updated to DeviceSecurityType -> in line with the env var name.

PR #1556, opened 2020-09-15
Add support for DPS provisioning flow for pnp TemperatureController samples

abhipsaMisra:
image

Todo: Moving the helper class into the client library - will do this in the next PR.


iothub/device/samples/PnpDeviceSamples/PnpHelper/PnpHelper.cs
```diff
@@ -150,6 +150,16 @@ public static string CreatePnpCommandName(string commandName, string componentNa
...
         }
 
+        /// <summary>
+        /// Create the DPS payload to provision a device as plug and play.
+        /// </summary>
+        /// <param name="modelId">The model Id corresponding to the device.</param>
+        /// <returns>The DPS payload to provision a device as plug and play.</returns>
+        public static string CreateDpsPayload(string modelId)
+        {
+            return $"{{ \"modelId\": \"{modelId}\" }}";
```
drwill-ms, line 160: parameter checking? Not just null or whitespace, but it should follow a certain format too.

drwill-ms, line 160: You can drop the spaces for extra, on-the-wire savings!

iothub/device/samples/PnpDeviceSamples/PnpHelper/PnpHelper.cs
```diff
@@ -150,6 +150,16 @@ public static string CreatePnpCommandName(string commandName, string componentNa
             return string.IsNullOrWhiteSpace(componentName) ? commandName : $"{componentName}*{commandName}";
         }
 
+        /// <summary>
+        /// Create the DPS payload to provision a device as plug and play.
+        /// </summary>
+        /// <param name="modelId">The model Id corresponding to the device.</param>
```
drwill-ms, line 156: The Id of the model the device adheres to for properties, telemetry, and commands.

iothub/device/samples/PnpDeviceSamples/PnpHelper/PnpHelper.cs
```diff
@@ -150,6 +150,16 @@ public static string CreatePnpCommandName(string commandName, string componentNa
             return string.IsNullOrWhiteSpace(componentName) ? commandName : $"{componentName}*{commandName}";
         }
 
+        /// <summary>
+        /// Create the DPS payload to provision a device as plug and play.
```
drwill-ms, line 154: Any article links to PnP we can add in a remarks section?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// <summary>
+    /// Parameters for the application supplied via command line arguments.
+    /// If the parameter is not supplied via command line args, it will look for it in environment variables.
+    /// </summary>
+    internal class Parameters
+    {
+        [Option(
+            's',
+            "DeviceSecurityType",
+            HelpText = "(Required) The provisioning type that will be used for provisioning the device for the sample. Possible values include dps, connectionString (case-insensitive)." +
```
drwill-ms, line 18: Possible case-insensitive values include: dps, connectionString.

drwill-ms, line 18: ...that will be used for **connecting** the device...

Otherwise we use the word provisioning twice, which sounds weird. Also, if I use a connection string to the hub, it isn't really provisioning. Might think that through a bit.

drwill-ms:
For a readme, these might be good links to add:

https://docs.microsoft.com/en-us/azure/iot-dps/about-iot-dps
https://docs.microsoft.com/en-us/azure/iot-dps/how-to-manage-enrollments

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,316 @@
...
+            _maxTemp[Thermostat2] = 0d;
+
+            await Task.Run(async () =>
+            {
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    if (temperatureReset)
+                    {
+                        // Generate a random value between 5.0°C and 45.0°C for the current temperature reading for each "Thermostat" component.
+                        _temperature[Thermostat1] = Math.Round(_random.NextDouble() * 40.0 + 5.0, 1);
```
drwill-ms, line 94: DRY

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,316 @@
...
+            {
+                _logger.LogDebug($"Property: Update - component=\"{componentName}\", received an update which is not associated with a valid property.");
+            }
+        }
+
+        // Send working set of device memory over telemetry.
+        private async Task SendDeviceMemoryAsync(CancellationToken cancellationToken)
+        {
+            string telemetryName = "workingSet";
+            long workingSet = Environment.WorkingSet / 1024;
```
drwill-ms, line 255: turns out workingset is not a great measure of memory usage as it includes shared memory used by other processes.

private bytes is better. https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.privatememorysize64?view=netcore-3.1

drwill-ms (approved):

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
```diff
@@ -0,0 +1,77 @@
...
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_SECURITY_TYPE\".")]
+        public string DeviceSecurityType { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_SECURITY_TYPE");
+
+        [Option(
+            'p',
+            "PrimaryConnectionString",
```
vinagesh, line 24: Just ConnectionString? It should not matter if you provide primary or secondary.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
```diff
@@ -0,0 +1,77 @@
...
+        [Option(
+            'i',
+            "DpsIdScope",
+            HelpText = "(Required if DeviceSecurityType is \"dps\"). \nThe DPS ID Scope to use during device provisioning." +
+            "\nDefaults to environment variable \"IOTHUB_DEVICE_DPS_ID_SCOPE\".")]
+        public string DpsIdScope { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_DEVICE_DPS_ID_SCOPE");
+
+        [Option(
+            'd',
+            "DeviceId",
```
vinagesh, line 45: Do we really need DeviceId when we have DeviceConnectionString? Can we not extract it from the DeviceConnection string? It will lead to fewer user errors.

vinagesh (approved):

iothub/device/samples/PnpDeviceSamples/PnpHelper/PnpHelper.cs
abhipsaMisra: I'll add a check for null or whitespace. The format of DTDL model Id is being validated by the service.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
abhipsaMisra: Yeah, I am still looking for the proper term to use here - it isn't really a security type, is it?
I do like connection here - maybe we can rename it to "deviceConnectionType" and use "dps" and "iothub" as possible values? Although, even if you connect _via_ dps, you will still connect to iothub finally.

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: We get the next random double, and then round it to 1 decimal place. Is there a better way to do this?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
abhipsaMisra: Since this is a device sample, I don't see there being a confusion about this being a hub connection string. Also, I went with **Primary**ConnectionString since that is what we've decided to use in other samples as well, and keeping sync is imp.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
abhipsaMisra: They are required in different flows - the dps flow requires device Id, while the iothub flow requires the device connection string.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
vinagesh: I still think being explicit is useful. I myself have sometimes stumbled because I am new to this. It would be good to be explicit which connection string we are expecting. Though this is a device sample, we know that it needs the device connection string as we have implemented it. As a new user one might not be sure. End of the day the device communicates to the IoTHub. For consistency, are we talking of samples across SDKs or just our samples? One can easily use the SecondaryConnectionString and succeed. This just feels a little misleading. Might be better to be precise rather than follow other samples.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
vinagesh: Is the DeviceId going to be different from the one we can extract from the connection string? If not, can we just not extract it in our code so user does not have to have yet another input? And also if the deviceId in the connection string is not the same as the deviceId specified here, will the sample work?

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
abhipsaMisra: For this being connectionString vs deviceConnectionString - this info is described in the helptext: "The primary connection string for the device to simulate.". Does it still seem confusing?

For this being primaryConnectionString vs connectionSTring - see https://github.com/Azure-Samples/azure-iot-samples-csharp/pull/132/files#r486537250.

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
abhipsaMisra: The input required for hub flow is device connection string.
The input required for dps flow is device Id, dps Id scope, dps endpoint, device symmetric key.

For dps flow, the user would need to supply the device Id, but not the device connection string. So there is no way for the sample to extract it from elsewhere.

```
  -s, --DeviceSecurityType         (Required) The provisioning type that will be used for provisioning the device for the sample. Possible values include dps,
                                                  connectionString (case-insensitive).
                                                  Defaults to environment variable "IOTHUB_DEVICE_SECURITY_TYPE".

  -p, --PrimaryConnectionString    **(Required if DeviceSecurityType is "connectionString").**
                                                      The primary connection string for the device to simulate.
                                                      Defaults to environment variable "IOTHUB_DEVICE_CONNECTION_STRING".

  -e, --DpsEndpoint                (Required if DeviceSecurityType is "dps").
                                               The DPS endpoint to use during device provisioning.
                                               Defaults to environment variable "IOTHUB_DEVICE_DPS_ENDPOINT".

  -i, --DpsIdScope                 (Required if DeviceSecurityType is "dps").
                                             The DPS ID Scope to use during device provisioning.
                                             Defaults to environment variable "IOTHUB_DEVICE_DPS_ID_SCOPE".

  -d, --DeviceId                   **(Required if DeviceSecurityType is "dps").**
                                           The device registration Id to use during device provisioning.
                                           Defaults to environment variable "IOTHUB_DEVICE_DPS_DEVICE_ID".

  -k, --DeviceSymmetricKey         (Required if DeviceSecurityType is "dps").
                                                    The device symmetric key to use during device provisioning.
                                                    Defaults to environment variable "IOTHUB_DEVICE_DPS_DEVICE_KEY".

```

iothub/device/samples/PnpDeviceSamples/TemperatureController/Parameter.cs
vinagesh: I think naming is more important than text. People tend to not read things

PR #1573, opened 2020-09-23
Update device and service SDK to refer to API version 2020-09-30, bring model Id changes to twin (from preview)

abhipsaMisra:
- update the device and service SDK to refer to API version 2020-09-30.
- remove the if-else block added for using different API versions on device client, depending on if model Id was set or not.
- bring model Id changes to twin (from preview)

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -297,7 +297,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                     Debug.Assert(_mqttTransportSettings.ClientCertificate != null);
                 }
 
-                string usernameString = $"{this._iotHubHostName}/{id}/?{ClientApiVersionHelper.ApiVersionQueryStringLatest}&{DeviceClientTypeParam}={Uri.EscapeDataString(this._productInfo.ToString())}";
+                string usernameString = $"{_iotHubHostName}/{id}/?{ClientApiVersionHelper.ApiVersionQueryStringLatest}&{DeviceClientTypeParam}={Uri.EscapeDataString(_productInfo.ToString())}";
 
                 if (!_mqttTransportSettings.AuthenticationChain.IsNullOrWhiteSpace())
                 {
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -297,24 +297,18 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                     Debug.Assert(_mqttTransportSettings.ClientCertificate != null);
                 }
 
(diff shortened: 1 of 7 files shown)


abhipsaMisra:
Pipeline run: https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=68714&view=results

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -297,26 +297,20 @@ private async Task ConnectAsync(IChannelHandlerContext context)
...
-                {
-                    serviceParams = ClientApiVersionHelper.ApiVersionQueryStringLatest;
-                }
-                else
-                {
-                    serviceParams = $"{ClientApiVersionHelper.ApiVersionQueryStringPreview}&{ModelIdParam}={_options.ModelId}";
-                }
-
-                string usernameString = $"{this._iotHubHostName}/{id}/?{serviceParams}&{DeviceClientTypeParam}={Uri.EscapeDataString(this._productInfo.ToString())}";
+                string usernameString = $"{this._iotHubHostName}/{id}/?{ClientApiVersionHelper.ApiVersionQueryStringLatest}&{DeviceClientTypeParam}={Uri.EscapeDataString(this._productInfo.ToString())}";
```
drwill-ms, line 300: `this.`?

shared/src/Twin.cs
```diff
@@ -48,6 +48,13 @@ public Twin(TwinProperties twinProperties)
         /// </summary>
         public string DeviceId { get; set; }
 
+        /// <summary>
+        /// The DTDL model id of the device.
```
drwill-ms, line 52: Id

shared/src/Twin.cs
```diff
@@ -48,6 +48,13 @@ public Twin(TwinProperties twinProperties)
...
+        /// The value will be null for a non-pnp device.
```
drwill-ms, line 53: Seems like these extra statements belong in `<remarks>`

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
abhipsaMisra: yeah, can be removed for sure. I'll update this.

drwill-ms (approved):
My only concern is updating all our APIs to the latest API version. We haven't done that in a long time, and such a change should be considered WRT our test coverage. We really do need to get to the point we can update to the latest API version. If this is that point in time, great.

abhipsaMisra:
> My only concern is updating all our APIs to the latest API version. We haven't done that in a long time, and such a change should be considered WRT our test coverage. We really do need to get to the point we can update to the latest API version. If this is that point in time, great.

Depending on Srivani's response, we can decide if we want to hold off on enabling the new service API version for all code paths. If there aren't any breaking changes that they highlight, we can go forward with this change.

abhipsaMisra:
I am merging this in for now, so that we can generate the private nugets for the bug-bash (more tests with new API version). 
If we get a response from Srivani asking us not to make this update, I will revert/ fix in another PR.

PR #1583, opened 2020-09-25
Update digital twin update operations utility, add readme demonstrating DelegatingHandler usage

abhipsaMisra:


iothub/service/src/DigitalTwin/Serialization/UpdateOperationsUtility.cs
```diff
@@ -21,11 +22,34 @@ public class UpdateOperationsUtility
         private readonly List<Dictionary<string, object>> _ops = new List<Dictionary<string, object>>();
 
         /// <summary>
-        /// Include an add operation.
+        /// Include an add property operation.
+        /// Learn more about managing digital twins here <see href="https://docs.microsoft.com/en-us/azure/iot-pnp/howto-manage-digital-twin"/>.
         /// </summary>
+        /// <remarks>
```
vinagesh, line 28: This is good. We should add in java too.

iothub/service/src/DigitalTwin/readme.md
```diff
@@ -1,20 +1,91 @@
-#### Autorest config file for generating Digital Twin Service Client models and protocol layer
+*NOTE: The `DigitalTwinClient` is available only on .NET framework 4.7.2 and .NET Standard 2.0+.
```
vinagesh, line 1: There is no space beween * and Note so it will just show up as *Note:......
You can either add a space or use blockquotes - https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#blockquotes

abhipsaMisra: I like blockquotes, I'll use that!

iothub/service/src/DigitalTwin/readme.md
```diff
@@ -1,20 +1,91 @@
-#### Autorest config file for generating Digital Twin Service Client models and protocol layer
+  > The `DigitalTwinClient` is available only on .NET framework 4.7.2 and .NET Standard 2.0+.
 
-#### Protocol layer generation:
-- Run the powershell script `generateCode.ps`. It will pick up the config below and output the results into this folder. It will also make a few automated changes to the generated protocol layer, that are required for this client library.
+### Protocol layer generation:
```
vinagesh, line 3: We need to add this for java as well if we explain how to generate your own SDK.

iothub/service/src/DigitalTwin/readme.md
```diff
@@ -1,20 +1,91 @@
...
+- Run the powershell script [generateCode.ps](./generateCode.ps1). It will pick up the autorest config and output the results into this folder. It will also make a few automated changes to the generated protocol layer, that are required for this client library.
```
vinagesh, line 4: add link to readme.md ?

iothub/service/src/DigitalTwin/readme.md
```diff
@@ -1,20 +1,91 @@
...
-input-file: DigitalTwin.json
+### /Authentication
+
+The code for generating shared access signature tokens, used for authentication Http requests against Azure IoT Hub service.
+
+### /Generated
+
+The code generated by autorest using the swagger file defined under [DigitalTwin.json](./DigitalTwin.json).
+
+### /src/Customized
```
vinagesh, line 20: This has been renamed to models in this PR right?

vinagesh: You only moved a few to Models so ignore above comment.

iothub/service/src/DigitalTwin/readme.md
```diff
@@ -1,20 +1,91 @@
...
+
+The customzied code written to override the following behavior of auto-generated code:
+
+- Rename some of the generated types, eg. [DigitalTwinInvokeCommandHeaders](./Customized/DigitalTwinInvokeCommandHeaders.cs)
```
vinagesh, line 24: Should we really give all these details here? This is implementation details for what we generate. Customers need to adhere to this if they want to generate their own SDK. The things below does not tell why we do it so it might look confusing. Thoughts?

iothub/service/src/DigitalTwin/readme.md
```diff
@@ -1,20 +1,91 @@
...
+}
+catch (HttpOperationException ex) when (ex.Response.StatusCode == HttpStatusCode.NotFound)
+{
+    // Service could not reach the device, make sure the device is online.
+}
+```
+
+## Using delegating handlers with DigitalTwinClient
+
+The `DigitalTwinClient.CreateFromConnectionString()` factory method takes in an optional list of `System.Net.Http.DelegatingHandler`, which are added to the Http client pipeline. You can use these delegating handler to specify your custom policies for tracing, retry, routing through a proxy, etc.
```
vinagesh, line 57: do we need to say etc ? It becomes open ended.

vinagesh (approved):

iothub/service/src/DigitalTwin/Extensions.cs
```diff
@@ -38,5 +39,17 @@ internal static void ThrowIfNullOrWhiteSpace(this string argumentValue, string a
                 throw new ArgumentNullException(argumentName, errorMessage);
             }
         }
+
+        /// <summary>
+        /// Append the "$metadata" identifier to the property values, which helps the service identify the patch as a component update.
+        /// </summary>
+        /// <param name="propertyKeyValuePairs">The dictionary of property key values pairs to update to.</param>
+        internal static void AddComponentUpdateIdentifier(this Dictionary<string, object> propertyKeyValuePairs)
```
vinagesh, line 47: It feels like a overkill to add extensions just for this. Extensions are less discoverable.

iothub/service/src/DigitalTwin/readme.md
```diff
@@ -1,20 +1,91 @@
...
+The customzied code written to override the following behavior of auto-generated code:
+
+- Rename some of the generated types, eg. [DigitalTwinInvokeCommandHeaders](./Customized/DigitalTwinInvokeCommandHeaders.cs)
+- Declare some of the generated types as **internal**, instead of the autorest default of **public**.
+- Declare some methods to accept input parameters as **strings** instead of **objects**.
+- Declare some methods to return the response as **strings** instead of **objects**.
+
+### /src/Models
+
+Model classes useful for use with the Digital Twins client operations.
```
barustum, line 31: should we stick with singular 'Twin'? per the naming meeting with had this morning?

vinagesh: ha ha the million $ question :) I think yes we should.

barustum (approved):

iothub/service/src/DigitalTwin/readme.md
abhipsaMisra: I don't understand, which readme? This doc is the readme.

abhipsaMisra: Do you mean the autorest config?

iothub/service/src/DigitalTwin/readme.md
abhipsaMisra: I followed the format that we have in track 2 repo.
I don't mind putting this info in, as it find of explains why we have a "customization" on top. If they look at the "generateCode" script, they find the same info = what is being done, but not why.

iothub/service/src/DigitalTwin/readme.md
abhipsaMisra: It is open-ended though, right. I could have any no. of delegating handler as I like.

iothub/service/src/DigitalTwin/readme.md
abhipsaMisra: yup, good catch. I copied this over from the track 2 repo, and forgot to rename this.

iothub/service/src/DigitalTwin/Extensions.cs
abhipsaMisra: discussed offline, will leave as is, unless this goes against coding guidelines/ is inefficient.

iothub/service/src/DigitalTwin/readme.md
abhipsaMisra: I assume you mean the autorest config yaml, so I'll add a link to that.

PR #1584, opened 2020-09-26
Add unit tests for PnpHelper and UpdateOperationsUtility

abhipsaMisra:
Final API surface: https://github.com/Azure/iot-sdks-internals/pull/204/files

### DIFF
--- a/iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
+++ b/iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
@@ -1,0 +1,164 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using FluentAssertions;
+using Microsoft.Azure.Devices.Client.PlugAndPlay;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Devices.Client.Tests.PlugAndPlay
+{
+    /// <summary>
+    /// Ensures the PnP convention helpers produce content to spec and using different data types
+    /// </summary>
+    [TestClass]
+    [TestCategory("Unit")]
+    public class PnpHelperTests
+    {
+        [TestMethod]
+        public void CreateRootPropertyPatch()
+        {
+            // Format:
+            //  {
+            //      "samplePropertyName": 20
+            //  }
+
+            const string propertyName = "someName";
+            const int propertyValue = 10;
+
+            string patch = PnpHelper.CreatePropertyPatch(propertyName, JsonConvert.SerializeObject(propertyValue));
+            var jObject = JObject.Parse(patch);
+
+            jObject.Count.Should().Be(1, "there should be a single property added");
+            jObject.Value<int>(propertyName).Should().Be(propertyValue);
+        }
+
+        [TestMethod]
+        public void CreateComponentPropertyPatch()
+        {
+            // Format:
+            //  {
+            //      "sampleComponentName": {
+            //          "__t": "c",
+            //          "samplePropertyName"": 20
+            //      }
+            //  }
+
+            const string componentName = "someComponent";
+            const string propertyName = "someName";
+            const int propertyValue = 10;
+
+            string patch = PnpHelper.CreatePropertyPatch(propertyName, JsonConvert.SerializeObject(propertyValue), componentName);
+
+            var jObject = JObject.Parse(patch);
+            JObject component = jObject.Value<JObject>(componentName);
+
+            component.Count.Should().Be(2, "there should be two properties added - the above property and a component identifier {\"__t\": \"c\"}");
+            component.Value<int>(propertyName).Should().Be(propertyValue);
+            component[PnpHelper.PropertyComponentIdentifierKey].Should().NotBeNull();
+            ((string)component[PnpHelper.PropertyComponentIdentifierKey]).Should().Be(PnpHelper.PropertyComponentIdentifierValue);
+        }
+
+        [TestMethod]
+        public void CreateRootPropertyEmbeddedValuePatch()
+        {
+            // Format:
+            //  {
+            //      "samplePropertyName": {
+            //          "value": 20,
+            //          "ac": 200,
+            //          "av": 5,
+            //          "ad": "The update was successful."
+            //      }
+            //  }
+
+            const string propertyName = "someName";
+            const int propertyValue = 10;
+            const int ackCode = 200;
+            const long ackVersion = 2;
+
+            string patch = PnpHelper.CreatePropertyEmbeddedValuePatch(propertyName, JsonConvert.SerializeObject(propertyValue), ackCode, ackVersion);
+
+            var jObject = JObject.Parse(patch);
+            EmbeddedPropertyPatch actualPatch = jObject.ToObject<EmbeddedPropertyPatch>();
+
+            // The property patch object should have "value", "ac" and "av" properties set. Since we did not supply an "ackDescription", "ad" should be null.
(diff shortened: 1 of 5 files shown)


iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
```diff
@@ -0,0 +1,161 @@
...
+using Microsoft.Azure.Devices.Client.PlugAndPlay;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Devices.Client.Tests.PlugAndPlay
+{
+    [TestClass]
+    [TestCategory("Unit")]
+    public class PnpHelperTests
```
vinagesh, line 17: xml comments for class

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
```diff
@@ -0,0 +1,161 @@
...
+        [TestMethod]
+        public void CreateRootPropertyPatch()
+        {
+            // Format:
+            //  {
+            //      "samplePropertyName": 20
+            //  }
+
+            string propertyName = "someName";
+            int propertyValue = 10;
```
vinagesh, line 28: If you declare this as a string, you won't need the extra serialization in line 30. Same with other tests.

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            string patch = PnpHelper.CreatePropertyPatch(propertyName, JsonConvert.SerializeObject(propertyValue));
+            var jObject = JObject.Parse(patch);
```
vinagesh, line 31: JObject jObject as per coding guidelines.

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
```diff
@@ -0,0 +1,161 @@
...
+            JObject component = jObject.Value<JObject>(componentName);
+
+            // There should be two properties added to the component- the above property and a component identifier "__t": "c".
+            component.Count.Should().Be(2);
+            component[PnpHelper.PropertyComponentIdentifierKey].Should().NotBeNull();
+            ((string)component[PnpHelper.PropertyComponentIdentifierKey]).Should().Be(PnpHelper.PropertyComponentIdentifierValue);
+
+            // The property patch object should have "value", "ac", "ad" properties set. Since we did not supply an "ackDescription", "ad" should be null.
+            EmbeddedPropertyPatch actualPatch = component.ToObject<EmbeddedPropertyPatch>();
```
vinagesh, line 133: What the helper really returns is a string. So we should just test for the exact string returned instead of doing a deserialization to the class defined below.

iothub/service/tests/DigitalTwin/UpdateOperationsUtilityTests.cs
```diff
@@ -0,0 +1,158 @@
...
+        private const string Add = "add";
+        private const string Replace = "replace";
+        private const string Remove = "remove";
+        private const string Value = "value";
+        private const string Metadata = "$metadata";
+
+        [TestMethod]
+        public void UpdateUtilityAppendsAddPropertyOp()
+        {
+            UpdateOperationsUtility op = new UpdateOperationsUtility();
```
vinagesh, line 28: Use var here as per coding guidelines as the right hand side has the type.

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
abhipsaMisra: for unit test classes, why?

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
abhipsaMisra: I don't see an issue in using "int" here and serializing it below, do you?

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
abhipsaMisra: @vinagesh  I thought it was intuitive that `ABC.Parse()` is going to return `ABC`. If it isn't, let me know; I can specify the return type explicitly.

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
abhipsaMisra: I wouldn't rely on json string matching to do validations. There could be extra spaces, ordering differences etc; which will give rise to false negatives. 
It is always better to deserialize it into a json type, and then do validations.

iothub/service/tests/DigitalTwin/UpdateOperationsUtilityTests.cs
abhipsaMisra: good catch, will update.

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
drwill-ms: Line 35 checks it against the value, so we're making sure the type is maintained round-trip. I say leave it as it is.

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
drwill-ms: Seems fine to me. It isn't as obvious as `var foo = new Foo()`, but still pretty obvious.

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
```diff
@@ -0,0 +1,161 @@
...
+            //      "samplePropertyName": 20
+            //  }
+
+            string propertyName = "someName";
+            int propertyValue = 10;
+
+            string patch = PnpHelper.CreatePropertyPatch(propertyName, JsonConvert.SerializeObject(propertyValue));
+            var jObject = JObject.Parse(patch);
+
+            // There should be a single property added - the above property.
```
drwill-ms, line 33: instead of a comment, how about a failure message instead?

`jobject.Count.Should().Be(1, "there should be a single property added");`

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
drwill-ms, line 28: const

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        [TestMethod]
+        public void CreateRootPropertyPatch()
+        {
+            // Format:
+            //  {
+            //      "samplePropertyName": 20
+            //  }
+
+            string propertyName = "someName";
```
drwill-ms, line 27: const

iothub/service/tests/Microsoft.Azure.Devices.Tests.csproj
```diff
@@ -11,11 +11,16 @@
     <RunAnalyzersDuringLiveAnalysis>false</RunAnalyzersDuringLiveAnalysis>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net451' ">
```
drwill-ms, line 14: Why only for net451?

drwill-ms (approved):

iothub/service/tests/Microsoft.Azure.Devices.Tests.csproj
drwill-ms: Never mind. The verb is remove.

iothub/device/tests/PlugAndPlay/PnpHelperTests.cs
drwill-ms: How about "Ensures the PnP convention helpers produce content to spec and using different data types"

vinagesh (approved):

PR #1585, opened 2020-09-28
fix(service client): Update the IsFatal logic to consider NullRef exceptions as fatal

azabbasi:


iothub/service/src/AmqpServiceClient.cs
```diff
@@ -113,18 +104,12 @@ public async override Task SendAsync(string deviceId, Message message, TimeSpan?
...
-                            .ConfigureAwait(false);
-                    }
-                    else
-                    {
-                        outcome = await sendingLink
-                            .SendMessageAsync(amqpMessage, IotHubConnection.GetNextDeliveryTag(ref _sendingDeliveryTag), AmqpConstants.NullBinary, OperationTimeout)
-                            .ConfigureAwait(false);
-                    }
+
+                    TimeSpan calculatedTimeout = timeout != null ? (TimeSpan)timeout : OperationTimeout;
```
drwill-ms, line 108: `(TimeSpan)timeout ?? OperationTimeout;` ?

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -272,6 +257,11 @@ private async Task<SendingAmqpLink> GetSendingLinkAsync()
                 sendingLink = await _faultTolerantSendingLink.GetOrCreateAsync(OpenTimeout).ConfigureAwait(false);
             }
 
+            if (sendingLink == null)
+            {
+                throw new ArgumentNullException(nameof(sendingLink), "Failed to get or create a sending link");
```
drwill-ms, line 262: The sending link isn't an argument. Misleading exception type.

drwill-ms: @abhipsaMisra @timtay-microsoft what exception type would be best to throw here?

timtay-microsoft: I don't follow why the above method (getOrCreate) would ever return null for a singleton. We have customer's reporting stacktraces that indicate that this instance is null though? This may be worth asking the Microsoft.Azure.Amqp folks. I couldn't say what type of exception is appropriate to throw here unless I understood why the link would be null

abhipsaMisra: I have a similar question to what Tim asked above, under what scenario would `_faultTolerantSendingLink.GetOrCreateAsync()`  return a null link? What does that state signify - from the Amqp library's perspective? Is it worth retrying at that point? What is the next step?

For the question about the type of exception to be thrown - unless we can map it to a [specific category](https://github.com/Azure/azure-iot-sdk-csharp/tree/master/iothub/service/src/Common/Exceptions), the default exception type tends to be [IotHubException](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/service/src/Common/Exceptions/IotHubException.cs).

drwill-ms: Do we agree that ArgumentNullException is the wrong type?

abhipsaMisra: For the Amqp questions - Xin should be able to give some recommendation

abhipsaMisra: my opinion - the sending link being null doesn't really fit "an argument is null" criteria, for me. I would rather have it throw an "IotHubException" and have the exception message specify what went wrong and how to recover (since service client has no built-in retry, so the responsibility is now on the user).

drwill-ms: Looks like we have more investigation to do, mainly looking at the Amqp library, to understand why a sending link might be null, and how it might be recoverable.

azabbasi: I agree that ArgumentNullException doesn't fit here. I will ping Xin to see what his input is on this issue.

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -113,18 +104,12 @@ public async override Task SendAsync(string deviceId, Message message, TimeSpan?
...
-                            .ConfigureAwait(false);
-                    }
-                    else
-                    {
-                        outcome = await sendingLink
-                            .SendMessageAsync(amqpMessage, IotHubConnection.GetNextDeliveryTag(ref _sendingDeliveryTag), AmqpConstants.NullBinary, OperationTimeout)
-                            .ConfigureAwait(false);
-                    }
+
+                    TimeSpan calculatedTimeout = timeout ?? OperationTimeout;
```
drwill-ms, line 108: s_defaultOperationTimeout or OperationTimeout? Why do we have two?

drwill-ms (approved):

bikamani (approved):

PR #1591, opened 2020-10-01
Fix access modifier, bump release version

barustum:


provisioning/device/src/Extensions/CommonExtensions.cs
```diff
@@ -22,7 +22,7 @@ internal static class CommonExtensions
         /// </summary>
         /// <param name="input">The string to be formatted.</param>
         /// <returns>The input string, with extra white space removed. </returns>
-        public static string TrimWhiteSpace(this string input)
+        internal static string TrimWhiteSpace(this string input)
```
timtay-microsoft, line 25: Is this a breaking change? Or are we assuming that no one uses this API?

timtay-microsoft (approved):

PR #1594, opened 2020-10-02
Fix(device): Fix issue with .NET standard 5 preview -> Not supported exceptions. Fixes #1592

### LINKED ISSUE #1592: Microsoft.Azure.Devices.Client incorrectly determines .NET Standard support
This issue was [reported by a customer](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/comment-page-3/#comment-7383).

1. Use .NET 5 preview.
2. Install `Microsoft.Azure.Devices.Client`

### Expectation

Stuff works.

### Actuality

The customer gets an exception:

```
Unhandled exception. System.NotSupportedException: Specified method is not supported.
   at Microsoft.Azure.Devices.Client.Common.ReadOnlyMergeDictionary`2.System.Collections.Generic.ICollection<System.Collections.Generic.KeyValuePair>.get_Count()
```

It seems the root cause is a fragile check in the MSBuild targets:

```XML
  <PropertyGroup>
    <SupportsNetStandard20AndAbove Condition="'$(TargetFramework)' == 'netstandard2.1' Or '$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'net472'">true</SupportsNetStandard20AndAbove>
  </PropertyGroup>
```

What are you trying to achieve by this method?

azabbasi:
The latest .NET standard 5 preview accesses the Count() method on the dictionary and since we haven't implemented it , it throws a Not supported exception.
Instead of implementing it, I removed it and replaced it with a 5liner.


azabbasi:
Fixes #1592

timtay-microsoft (approved):

iothub/device/src/Common/Utils.cs
```diff
@@ -106,5 +109,20 @@ public static void ValidateDataIsEmptyOrJson(byte[] data)
...
+            if (dictionaries == null || dictionaries.Any(item => item == null))
+            {
+                throw new ArgumentNullException(nameof(dictionaries), "Provided dictionaries should not be null");
+            }
+
+            Dictionary<TKey, TValue> result = dictionaries.SelectMany(dict => dict)
+                .ToLookup(pair => pair.Key, pair => pair.Value)
+                .ToDictionary(group => group.Key, group => group.First());
+
+            return new ReadOnlyDictionary<TKey, TValue>(result);
```
drwill-ms, line 125: What does it do if the dictionaries have duplicate keys?

Unit test?

drwill-ms: Answered in the file below. Nice!

iothub/device/tests/UtilsTests.cs
```diff
@@ -35,5 +36,64 @@ public void ConvertDeliveryAckTypeToStringInvalidValueFail()
...
+        {
+            var dict1 = new Dictionary<string, string> { { "1", "2" }, { "2", "4" }, { "3", "5" } };
+
+            Utils.MergeDictionaries(new Dictionary<string, string>[] { dict1, null });
+        }
+
+        [TestMethod]
+        public void MergeDictionaries_TwoDictionaries_NoOverlap()
+        {
+            var dict1 = new Dictionary<string, string> { { "1", "2" }, { "2", "4" }, { "3", "5" } };
```
drwill-ms, line 52: For readability, I'd suggest changing the value to something not a number.

You could do `{ 1, "one" }` for instance, and then later for the dupe tests use "one1" and "one2" as the values.

drwill-ms (approved):

iothub/device/tests/UtilsTests.cs
azabbasi: done

PR #1595, opened 2020-10-05
fix(device) : Fix exception mapping in Device Client

azabbasi:
## Checklist

### DIFF
--- a/iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
+++ b/iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
@@ -239,7 +239,12 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIoT
             }
             else if (error.Condition.Equals(AmqpErrorCode.ResourceLimitExceeded))
             {
-                retException = new QuotaExceededException(message);
+                // Note: The DeviceMaximumQueueDepthExceededException is not supposed to be thrown here as it is being mapped to the incorrect error code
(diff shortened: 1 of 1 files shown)


timtay-microsoft (approved):

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
```diff
@@ -239,7 +239,7 @@ public static Exception ToIotHubClientContract(Error error)
             }
             else if (error.Condition.Equals(AmqpErrorCode.ResourceLimitExceeded))
             {
-                retException = new DeviceMaximumQueueDepthExceededException(message);
+                retException = new QuotaExceededException(message);
```
drwill-ms, line 242: We need to fully vet this breaking change, run it by Mark, and bring alternatives.

drwill-ms (changes requested):
Needs more vetting.

iothub/device/src/Transport/AmqpIoT/AmqpIoTErrorAdapter.cs
azabbasi: Yes. Starting the conversation now. Will hold on to this PR until we have a resolution.

abhipsaMisra: Under what scenarios does the service throw `amqp:resource-limit-exceeded` on amqp? 
You can update the doc [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/amqpTransportExceptions.md#not-retried-by-the-sdk), so that it carries an explanation for `amqp:resource-limit-exceeded` (right now it says - "Inspect the exception details, collect logs and contact service"; but if we have more info, it would help to be more descriptive.)

drwill-ms (approved):

Issue #1562, opened 2020-09-17
System.NotSupportedException: Specified method is not supported on .NET5.0 RC1


darenm:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64 v2004 19041.508
- **Application's .NET Target Framework :** net5.0
- **Device:** Desktop PC
- **SDK version used:** 1.30.0

## Description of the issue
A console application was created with `dotnet new console` and targeted `net5.0`.  When sending an event `await deviceClient.SendEventAsync(message);`, the following exception is thrown `System.NotSupportedException: Specified method is not supported`. Changing the console target to `netcoreapp3.1` results in a successful execution.

## Code sample exhibiting the issue
```csharp
deviceClient = DeviceClient.CreateFromConnectionString(connectionString, TransportType.Mqtt);
var messageString = CreateMessageString(currentTemperature, currentHumidity);

 // create a byte array from the message string using ASCII encoding
var message = new Message(Encoding.ASCII.GetBytes(messageString));

// Add a custom application property to the message.
// An IoT hub can filter on these properties without access to the message body.
message.Properties.Add("temperatureAlert", (currentTemperature > 30) ? "true" : "false");

// Send the telemetry message
await deviceClient.SendEventAsync(message);
```

## Console log of the issue
```
Unhandled exception. System.NotSupportedException: Specified method is not supported.
   at Microsoft.Azure.Devices.Client.Common.ReadOnlyMergeDictionary`2.System.Collections.Generic.ICollection<System.Collections.Generic.KeyValuePair<TKey,TValue>>.get_Count()
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source)
   at Microsoft.Azure.Devices.Client.Common.UrlEncodedDictionarySerializer.Serialize(IEnumerable`1 properties)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.PopulateMessagePropertiesFromMessage(String topicName, Message message)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.ComposePublishPacketAsync(IChannelHandlerContext context, Message message, QualityOfService qos, String topicName)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.SendMessageAsync(IChannelHandlerContext context, Message message)
[...]

PR #1599, opened 2020-10-07
Fix cross target framework dependency management

abhipsaMisra:
Since net451 is the only tfm < ns2.0 that we target, Instead of adding an explicit check for each supported tfm greater than ns2.0, we can simply check if the tfm != net451.

Fix for #1562 , #1592

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -49,10 +49,6 @@
     <common>$(RootDir)\common\src</common>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <SupportsNetStandard20AndAbove Condition="'$(TargetFramework)' == 'netstandard2.1' Or '$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'net472'">true</SupportsNetStandard20AndAbove>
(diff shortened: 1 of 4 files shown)


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -95,15 +91,15 @@
   </ItemGroup>
 
   <!-- NetStandard 2.1, NetStandard 2.0 and net472 -->
-  <ItemGroup Condition="'$(SupportsNetStandard20AndAbove)' == 'true'">
+  <ItemGroup Condition=" '$(TargetFramework)' != 'net451' ">
```
barustum, line 94: The space between the double-quote and single quote is intentional?

barustum (approved):

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
abhipsaMisra: Interestingly, yes. It was the format followed in the csproj everywhere, so I kept it consistent. I am sure there is no scientific reason for it, I just wanted to have it be consistent with the pre-existing code.

abhipsaMisra: I think it makes it readable

PR #1601, opened 2020-10-08
build: Exclude running device stream tests in prod pipelines

barustum:
## Checklist

## Description of the changes

- Extract the device streaming unit tests into a separate test file
- Add a new test category for the device streaming tests
- Wrote logic in the build scripts to exclude the test category when running against the prod pipeline

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -368,6 +368,7 @@ try
         $verbosity = "normal"
         $testCategory = "TestCategory=E2E"
         
+        # Skip the Device Streaming tests when running on the production pipeline. Device Streaming features are not yet available in all regions.
         if ($skipDeviceStreamTests) 
         {
             $testCategory += "&TestCategory!=DeviceStreaming"
--- a/build.ps1
+++ b/build.ps1
@@ -171,7 +171,7 @@ Function RunTests($message, $framework = "*", $filterTestCategory = "*")
     Write-Host
     Write-Host -ForegroundColor Cyan $label
 
-    $runTestCmd = "dotnet test -t -s test.runsettings --verbosity $verbosity --configuration $configuration --logger trx"
+    $runTestCmd = "dotnet test -s test.runsettings --verbosity $verbosity --configuration $configuration --logger trx"
 
     if ($noBuildBeforeTesting)
     {
--- a/build.ps1
+++ b/build.ps1
@@ -171,7 +171,7 @@ Function RunTests($message, $framework = "*", $filterTestCategory = "*")
     Write-Host
     Write-Host -ForegroundColor Cyan $label
 
-    $runTestCmd = "dotnet test -s test.runsettings --verbosity $verbosity --configuration $configuration --logger trx"
+    $runTestCmd = "dotnet test -t -s test.runsettings --verbosity $verbosity --configuration $configuration --logger trx"
 
     if ($noBuildBeforeTesting)
     {
--- a/build.ps1
+++ b/build.ps1
@@ -72,7 +72,8 @@ Param(
     [switch] $skipIotHubTests,
     [switch] $skipDPSTests,
     [switch] $skipPnPTests,
-    [switch] $noBuildBeforeTesting
+    [switch] $noBuildBeforeTesting,
+    [switch] $skipDeviceStreamTests
 )
 
 Function CheckSignTools()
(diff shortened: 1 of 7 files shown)


build.ps1
```diff
@@ -274,7 +275,14 @@ try
             Write-Host -ForegroundColor Cyan "Unit Test execution"
             Write-Host
 
-            RunTests "Unit tests" -filterTestCategory "TestCategory=Unit" -framework $framework
+            $testCategory = "TestCategory=Unit";
+
+            if ($skipDeviceStreamTests)
```
drwill-ms, line 280: By definition, unit tests shouldn't hit the service. We don't seem to have other exclusions in this if block, so should this be removed?

drwill-ms: Or do the other exclusions belong here too?

iothub/device/tests/DeviceStreamingTests.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NSubstitute;
+
+namespace Microsoft.Azure.Devices.Client.Test
+{
+    [TestClass]
+    [TestCategory("DeviceStreaming")]
+    [TestCategory("Unit")]
```
drwill-ms, line 14: These might not be unit tests. Perhaps they were just miscategorized E2E tests?

drwill-ms: Otherwise, I don't see how we've put real service calls in West Central US, causing this problem.

drwill-ms: According to the email, we should be looking for these test names:

> ModuleNotFoundTestSasl...
> DeviceClient_Amqp_TimesOutSasl_...

drwill-ms (changes requested):
E2E tests need categorizing

build.ps1
barustum: Good catch. Removed this from the unittests section as they hit a fake endpoint (mock tests).

iothub/device/tests/DeviceStreamingTests.cs
barustum: me and @abhipsaMisra  are trying to figure out where these device names are coming from. We can't seem to find them in our code:
ModuleNotFoundTestSasl...
DeviceClient_Amqp_TimesOutSasl_...

Will report ASAP

barustum: It seems like the tests that are problematic are the ones that start with E2E_DeviceStreamingTests_Sasl*, not as previously reported by the service team.

e2e/test/iothub/DeviceStreamingE2ETests.cs
```diff
@@ -1,33 +1,32 @@
...
+using Microsoft.Azure.Devices.Client.Exceptions;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 using ClientDeviceStreamingRequest = Microsoft.Azure.Devices.Client.DeviceStreamRequest;
 using ServiceDeviceStreamingRequest = Microsoft.Azure.Devices.DeviceStreamRequest;
 
 namespace Microsoft.Azure.Devices.E2ETests
 {
     [TestClass]
     [TestCategory("IoTHub-E2E")]
```
drwill-ms, line 20: I think this should be broken out to E2E and IoTHub

drwill-ms (approved):

build.ps1
```diff
@@ -360,8 +366,14 @@ try
         # Override verbosity to display individual test execution.
         $oldVerbosity = $verbosity
         $verbosity = "normal"
+        $testCategory = "TestCategory=E2E"
+        
+        if ($skipDeviceStreamTests)
```
abhipsaMisra, line 371: I would add a note here stating that we are intentionally skipping e2e tests for batched/CI builds for certian scenarios.

vsts/determine_tests_to_run.ps1
```diff
@@ -39,6 +39,14 @@ function ShouldSkipPnPTests
 	return $false
 }
 
+function ShouldSkipDeviceStreamTests
+{
+	return ($env:SOURCE_BRANCH_NAME -and `
+		$env:SOURCE_BRANCH_NAME.toLower() -eq 'preview' -and `
+		$env:PIPELINE_ENVIRONMENT -and `
+		$env:PIPELINE_ENVIRONMENT.toLower() -eq 'prod')
```
abhipsaMisra, line 47: q - what are possible values for this env variable?

abhipsaMisra (approved):

timtay-microsoft (approved):

vsts/determine_tests_to_run.ps1
barustum: Good idea. Done.

PR #1607, opened 2020-10-09
Fix(device-Service-Provisioning) Upgrade AMQP libary nuget package - Re-enable tests

azabbasi:
- Upgrade Microsoft.Azure.Amqp Library from 2.4.4 to 2.4.7
- Re-enable the flaky test

We have an e2e test that has been failing on Linux for a while. The issue seems to be that the underlying AMQP library is not throwing timeout exceptions sometimes during 
`await tcpInitiator.ConnectTaskAsync(timeout).ConfigureAwait(false);`

Looking into the release notes from the Amqp library, it appears they have issued a fix for this problem: https://github.com/Azure/azure-amqp/releases/tag/v2.4.7

### DIFF
--- a/e2e/test/provisioning/ProvisioningE2ETests.cs
+++ b/e2e/test/provisioning/ProvisioningE2ETests.cs
@@ -757,8 +757,6 @@ namespace Microsoft.Azure.Devices.E2ETests.Provisioning
         }
 
         // Note: This test takes 3 minutes.
-        // TODO: azabbasi - Investigate the failure reason and re-enable test.
-        [Ignore("Skipping test due to failures in .net core/ Linux")]
         [LoggedTestMethod]
         [TestCategory("LongRunning")]
         public async Task ProvisioningDeviceClient_InvalidGlobalAddress_Register_Amqp_Fail()
@@ -795,10 +793,10 @@ namespace Microsoft.Azure.Devices.E2ETests.Provisioning
             using SecurityProvider security = await CreateSecurityProviderFromName(attestationType, enrollmentType, groupId, null, AllocationPolicy.Hashed, null, null).ConfigureAwait(false);
 
             ProvisioningDeviceClient provClient = ProvisioningDeviceClient.Create(
-InvalidGlobalAddress,
-Configuration.Provisioning.IdScope,
-security,
-transport);
+                InvalidGlobalAddress,
+                Configuration.Provisioning.IdScope,
+                security,
+                transport);
 
             using var cts = new CancellationTokenSource(FailingTimeoutMiliseconds);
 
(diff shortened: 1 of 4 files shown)


e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -795,10 +793,10 @@ public async Task ProvisioningDeviceClient_InvalidGlobalAddress_Register_MqttWs_
             using SecurityProvider security = await CreateSecurityProviderFromName(attestationType, enrollmentType, groupId, null, AllocationPolicy.Hashed, null, null).ConfigureAwait(false);
 
             ProvisioningDeviceClient provClient = ProvisioningDeviceClient.Create(
-InvalidGlobalAddress,
-Configuration.Provisioning.IdScope,
-security,
-transport);
+                InvalidGlobalAddress,
```
bikamani, line 796: Is it tabs here? shows up like spaces

bikamani (approved):

e2e/test/provisioning/ProvisioningE2ETests.cs
drwill-ms: looks correct here, and incorrect in former file.

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -52,7 +52,7 @@
   <ItemGroup>
     <PackageReference Include="DotNetty.Codecs.Mqtt" Version="0.6.0" />
     <PackageReference Include="DotNetty.Handlers" Version="0.6.0" />
-    <PackageReference Include="Microsoft.Azure.Amqp" Version="2.4.4" />
+    <PackageReference Include="Microsoft.Azure.Amqp" Version="2.4.7" />
```
drwill-ms, line 55: Is this update still compatible with all the frameworks we target?

What fixes come with it?

drwill-ms (approved):

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
azabbasi: Yes. I ran through the pipelines and everything seems to be in place. 
Here are the fixes we will be bringing in: 
- Implemented timeout when creating a tcp transport. On some platforms the tcp connect can take up to 90 seconds to error out when the remote peer is down.
- Fails to open an AMQPS connection on .NET 5.0.
- Add netstandard2.0 target.
- Dispose timer objects explicitly instead of relying on finalizer.
Notes: for netstandard2.0 target, serialization of a field uses reflection instead of IL emit. If you have fields marked as [AmqpMember], please convert them to property members. It only affect code that uses AmqpContractSerializer.

azabbasi: https://github.com/Azure/azure-amqp/releases

e2e/test/provisioning/ProvisioningE2ETests.cs
azabbasi: I think we should never have tabs. I always make sure I have untabified files.

abhipsaMisra (approved):

PR #1611, opened 2020-10-13
Add comments for DeviceClient, ModuleClient, MqttTransportSettings

abhipsaMisra:
This PR adds/updates the comments in :
- `DeviceClient`, marking code blocks with appropriate tags.
- `ModuleClient`, explaining the use-case for `InvokeMethod` APIs (since they are edge-specific). Usually we would expect these APIs to be present as a part of `ServiceClient` (they exist in the `ServiceClient` as well, currently).
- `MqttTransportSettings` - there are a bunch of properties here that are unused in the client library. This PR call them out as such. The long-term plan is to either hook them to the library (if applicable), or mark them as deprecated.

### DIFF
--- a/iothub/device/src/ModuleClient.cs
+++ b/iothub/device/src/ModuleClient.cs
@@ -12,6 +12,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client.Edge;
+using Microsoft.Azure.Devices.Client.Extensions;
 using Microsoft.Azure.Devices.Client.Transport;
 using Microsoft.Azure.Devices.Shared;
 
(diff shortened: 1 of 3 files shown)


iothub/device/src/ModuleClient.cs
```diff
@@ -577,47 +578,57 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
             InternalClient.SetMessageHandlerAsync(messageHandler, userContext, cancellationToken);
 
         /// <summary>
-        /// Interactively invokes a method on a device
+        /// Interactively invokes a method from an edge module to an edge device.
```
abhipsaMisra, line 581: @ancaantochi Let me know if this sounds good

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -18,38 +18,22 @@ public class MqttTransportSettings : ITransportSettings
     {
         private readonly TransportType _transportType;
 
-        /// <summary>
-        /// Used by Edge runtime to specify an authentication chain for Edge-to-Edge connections
-        /// </summary>
-        internal string AuthenticationChain { get; set; }
```
abhipsaMisra, line 24: moved this below private variable declarations

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -18,38 +18,22 @@ public class MqttTransportSettings : ITransportSettings
...
         private const QualityOfService DefaultPublishToServerQoS = QualityOfService.AtLeastOnce;
         private const QualityOfService DefaultReceivingQoS = QualityOfService.AtLeastOnce;
-        private static readonly TimeSpan DefaultConnectArrivalTimeout = TimeSpan.FromSeconds(10);
-        private static readonly TimeSpan DefaultDeviceReceiveAckTimeout = TimeSpan.FromSeconds(300);
-
-        /// <summary>
-        /// To enable certificate revocation check. Default to be false.
-        /// </summary>
-#pragma warning disable IDE1006 // Naming Styles
-        public bool CertificateRevocationCheck
```
abhipsaMisra, line 42: moved this below the constructor

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -86,48 +70,141 @@ public MqttTransportSettings(TransportType transportType)
...
+            set => TlsVersions.Instance.CertificateRevocationCheck = value;
+        }
+#pragma warning restore IDE1006 // Naming Styles
+
+        /// <summary>
+        /// Indicates if a device can timeout while waiting for a acknowledgment from service.
+        /// The default value is <c>false</c>.
+        /// </summary>
+        /// <remarks>
+        /// This property is currently unused.
```
abhipsaMisra, line 92: As of today, this property is unused in the client library

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -86,48 +70,141 @@ public MqttTransportSettings(TransportType transportType)
...
+        /// </remarks>
         public bool DeviceReceiveAckCanTimeout { get; set; }
 
+        /// <summary>
+        /// The time a device will wait, for an acknowledgment from service.
+        /// The default is 5 minutes.
+        /// </summary>
+        /// <remarks>
+        /// This property is currently unused.
```
abhipsaMisra, line 101: same here, unused.

abhipsaMisra: here, and below

iothub/device/src/ModuleClient.cs
abhipsaMisra: this, and the 3 API signatures below

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
azabbasi: Why keep them then?

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -86,48 +70,141 @@ public MqttTransportSettings(TransportType transportType)
...
-        /// Whether the transport has a will message
+        /// Indicates whether the transport has a will message.
         /// </summary>
+        /// <remarks>
+        /// Setting a will message is a way for clients to notify other subscribed clients about ungraceful disconnects in an appropriate way.
+        /// In response to the ungraceful disconnect, the service will send the last-will message to the configured telemetry channel.
+        /// The telemetry channel can be either the default Events endpoint or a custom endpoint defined by IoT Hub routing.
+        /// For more details, refer to https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-mqtt-protocol-directly-as-a-device.
+        /// </remarks>
         public bool HasWill { get; set; }
```
azabbasi, line 191: It is too late to change this property to HasWillMessage ? is this publicly exposed?

azabbasi (approved):
LGTM

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
abhipsaMisra: We cannot delete them. We have to either hook them up, or mark them as deprecated.
I am still waiting for confirmation from the service team on if these properties are required. This is to indicate to customer that setting this property right now has no effect.

abhipsaMisra: Yeah, `HasWill` is already public

iothub/device/src/ModuleClient.cs
abhipsaMisra: Confirmed with Anca that invoking methods from a module client only works if you are connected to an edge hub.

PR #1615, opened 2020-10-14
refactor(iot-device): Clean up class structure for code readability

abhipsaMisra:
This PR moves the code blocks around as per StyleCop ordeing rules [SA1201](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md)  and [SA1202](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1202.md).

However, I've deviated a bit from this rule in grouping related method calls together (public, internal and private method calls for a single operation are grouped together).
This made debugging the code easier for me, and it is also the structure being followed in track 2 clients (eg [AppConfiguration](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs)).

Readme update: https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/methodsArrange/readme.md


iothub/device/src/InternalClient.cs
```diff
@@ -26,21 +26,49 @@
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// Delegate for desired property update callbacks.  This will be called
-    /// every time we receive a PATCH from the service.
+    /// Delegate for connection status changed.
```
abhipsaMisra, line 29: delegates should be before enums

iothub/device/src/InternalClient.cs
```diff
@@ -199,25 +204,6 @@ public int DiagnosticSamplingPercentage
         // Codes_SRS_DEVICECLIENT_28_002: [This property shall be defaulted to 240000 (4 minutes).]
         public uint OperationTimeoutInMilliseconds { get; set; } = DeviceClient.DefaultOperationTimeoutInMilliseconds;
 
-        internal X509Certificate2 Certificate { get; set; }
-
-        internal IDelegatingHandler InnerHandler { get; set; }
-
-        internal Task SendMethodResponseAsync(MethodResponseInternal methodResponse, CancellationToken cancellationToken)
```
abhipsaMisra, line 206: method should be after properties

iothub/device/src/InternalClient.cs
```diff
@@ -1349,44 +1217,151 @@ internal void OnReportedStatePatchReceived(TwinCollection patch)
             _desiredPropertyUpdateCallback(patch, _twinPatchCallbackContext);
         }
 
-        private Task EnableMethodAsync(CancellationToken cancellationToken)
+        #region Device Specific API
```
abhipsaMisra, line 1220: There are some methods which are used in both Device and Module clients, and some which are client specific. I have separated the client specific methods into regions.

abhipsaMisra:
Note: this PR isn't against `master` right now, since I am waiting for https://github.com/Azure/azure-iot-sdk-csharp/pull/1611 to be merged in. Once it is in, I'll update the base branch to `master`.

iothub/device/src/InternalClient.cs
```diff
@@ -1790,12 +1765,42 @@ internal void ValidateModuleTransportHandler(string apiName)
 
         #endregion Module Specific API
 
+        public void Dispose()
```
drwill-ms, line 1768: publics above privates

iothub/device/src/InternalClient.cs
```diff
@@ -1790,12 +1765,42 @@ internal void ValidateModuleTransportHandler(string apiName)
...
+        {
+            InnerHandler?.Dispose();
+            _methodsDictionarySemaphore?.Dispose();
+            _receiveSemaphore?.Dispose();
+            _fileUploadHttpTransportHandler?.Dispose();
+        }
+
+        internal bool IsE2EDiagnosticSupportedProtocol()
```
drwill-ms, line 1776: internal above private

iothub/device/src/InternalClient.cs
abhipsaMisra: does this mean the private methods in the regions above?

iothub/device/src/DeviceClient.cs
```diff
@@ -287,21 +288,27 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+        /// <see cref="AbandonAsync(Message, CancellationToken)"/>, or <see cref="RejectAsync(Message, CancellationToken)"/>, and then dispose the message.
         /// </summary>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <exception cref="OperationCanceledException">Thrown when the operation has been canceled.</exception>
         /// <returns>The receive message or null if there was no message until CancellationToken Expired</returns>
         public Task<Message> ReceiveAsync(CancellationToken cancellationToken) => InternalClient.ReceiveAsync(cancellationToken);
 
         /// <summary>
-        /// Receive a message from the device queue with the specified timeout. After handling a received message, a client should call Complete, Abandon, or Reject, and then dispose the message.
+        /// Receive a message from the device queue using the cancellation token.
```
bikamani, line 309: Does this description apply to all protocols?

abhipsaMisra: that is a very good point, you cannot complete/reject/abandon messages over mqtt. I'll add a note for this.
I'll also check our device client comments to see what we say there - since that is the one customers will actually see

drwill-ms (commented):
I, for one, am not a fan of regions. They don't make the code more readable for me. I'm not approving this PR, but I'm not prohibiting it either. You probably should ask the team how we feel about using them, and then do it.

abhipsaMisra:
> I, for one, am not a fan of regions. They don't make the code more readable for me. I'm not approving this PR, but I'm not prohibiting it either. You probably should ask the team how we feel about using them, and then do it.

@drwill-ms : Fair enough! Do you sign-off on the rest of the code changes (minus the `#regions`)?
Calling on @azabbasi , @timtay-microsoft , @vinagesh , @barustum , @bikamani for your comments.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -632,7 +632,12 @@ private async Task OpenAsyncInternal(CancellationToken cancellationToken)
             if (IsProxyConfigured())
             {
                 //No need to do a DNS lookup since we have the proxy address already
+
+#if NET451
                 _serverAddresses = new IPAddress[0];
+#else
+                _serverAddresses = Array.Empty<IPAddress>();
```
abhipsaMisra, line 639: CA182 - Avoid unnecessary zero-length array allocations.  Use `Array.Empty<IPAddress>()` instead. (not available in net451)

iothub/device/src/InternalClient.cs
timtay-microsoft: In the Java, SDK, inheritance is used to make this a bit cleaner. File upload APIs live in device client since module clients can't do it, and sendEventAsync lives in internal client since both module and device client can do it. It's a bigger ask, but should we move away from composition and into inheritance (i.e. device client extends internal client, not device client has an instance of internal client)?

A switch like that would make this layer significantly less confusing

drwill-ms: Not sure I'd use inheritance, but partial classes would give you the same effect. They bring their own evil, but that'd be closer to what we should have had: sub clients.

abhipsaMisra: I agree, that was our plan at one point, where we would have the device client/ module client inherit from internal client, but we haven't gotten around to actually doing that. Our InternalClient is already "internal", so we aren't exposing those APIs to begin with, so public API surface wise, it wouldn't have a big impact.
It being a bigger work-item though, I am not sure on the prioritization for it. 

To me this felt like a good workaround for the time being.
That being said, if we prefer to give it a complete overhaul, I am all up for it.

bikamani (commented):
> > I, for one, am not a fan of regions. They don't make the code more readable for me. I'm not approving this PR, but I'm not prohibiting it either. You probably should ask the team how we feel about using them, and then do it.
> 
> @drwill-ms : Fair enough! Do you sign-off on the rest of the code changes (minus the `#regions`)?
> Calling on @azabbasi , @timtay-microsoft , @vinagesh , @barustum , @bikamani for your comments.

I like regions in this particular scenario due to two main reasons. 
1. This a very large file so it is good to be able to collapse unnecessary regions
2. It is not evident from a method name whether it applies to just modules or devices/both. It makes it easier to separate it using regions.

iothub/device/src/DeviceClient.cs
```diff
@@ -538,7 +582,7 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
 
         /// <summary>
         /// Registers a new delegate for the connection status changed callback. If a delegate is already associated,
-        /// it will be replaced with the new delegate. Note that this callback will never be called if the client is configured to use HTTP as that protocol is stateless
+        /// it will be replaced with the new delegate. Note that this callback will never be called if the client is configured to use HTTP, as that protocol is stateless.
```
bikamani, line 585: minor: "I" capital

abhipsaMisra: for "...it will be replaced with the new delegate"? It is a continuation from the line above.

readme.md
```diff
@@ -89,11 +89,11 @@ For details on OS support see the following resources:
...
+| [Send device-to-cloud message](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-d2c)     | :heavy_check_mark:* | :heavy_check_mark:* | :heavy_check_mark:  | :heavy_check_mark:  | :heavy_check_mark:  | Send device-to-cloud messages (max 256KB) to IoT Hub with the option to add custom properties and system properties, and batch send.  *IoT Hub only supports batch send over AMQP and HTTPS at the moment. The MQTT implementation loops over the batch and sends each message individually.                                                              |
+| [Receive cloud-to-device messages](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-c2d) | :heavy_check_mark:* | :heavy_check_mark:* | :heavy_check_mark:  | :heavy_check_mark:  | :heavy_check_mark:  | Receive cloud-to-device messages and read associated custom and system properties from IoT Hub, with the option to complete/reject/abandon C2D messages.  *IoT Hub supports the option to reject/abandon C2D messages over HTTPS and AMQP only at the moment. |
```
abhipsaMisra, line 93: you can complete messages over mqtt

drwill-ms:
> Do you sign-off on the rest of the code changes (minus the #regions)?

Yep

iothub/device/src/DeviceClient.cs
bikamani: Aha didn't realize that. It appeared as a new line

iothub/device/src/InternalClient.cs
abhipsaMisra: > Our InternalClient is already "internal", so we aren't exposing those APIs to begin with, so public API surface wise, it wouldn't have a big impact.

So, this statement by me was actually incorrect. InternalClient is internal, but if we were to have DeviceClient inherit from it, we would need to make InternalClient public. I am not particularly inclined to do that, since it is entirely an implementation detail, and it shouldn't be accessible to users.

timtay-microsoft (approved):

Issue #1289, opened 2020-03-26
[API query] DeviceClient ReceiveAsync does not have delegate/callback handler 


kiranpradeep:
- **OS, version, SKU and CPU architecture used:** Ubuntu 18.04 x86
- **Application's .NET Target Framework :** netcoreapp2.1
- **Device:** Laptop
- **SDK version used:** 1.20.0

## Description of the issue:
This is more of a query and API request. In DeviceClient[[1](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.client.deviceclient.receiveasync?view=azure-dotnet)], `ReceiveAsync` does not have delegate/callback handler API equivalent to SettMethodHandlerAsync[[2](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.client.deviceclient.setmethoddefaulthandlerasync?view=azure-dotnet)]/SetDesiredPropertyUpdateCallbackAsync/SetConnectionStatusChangesHandle API. 

This forces the lib user to poll for checking C2D messages (harder if there are multiple devices to manage). Is it intentional that every other method has a delegate/callback and not `ReceiveAsync`? If not, is it possible to provide delegate/callback API for receiving C2D messages?

AB#7288824

prmathur-microsoft:
We will add this to our backlog and plan to address this in future.

timtay-microsoft:
This API was designed this way likely because HTTP/MQTT/AMQPS all need to support receiving cloud to device messages. AMQPS/MQTT could theoretically have a callback like you want, but HTTP doesn't fit well with this approach since HTTP. Even if we tried to make HTTP fit into an API like this, the SDK would have to implicitly poll iothub to get those messages which seems messy.

@abhipsaMisra how feasible would it be to implement this suggested API with the caveat that it only works for amqp/mqtt?

abhipsaMisra:
@timtay-microsoft Yes, that is how I see it; I would add callbacks for ReceiveAsync() that would be supported only over AMQP and MQTT, being in parity with how callback based method and twin APIs are also supported only over AMQP and MQTT.

abhipsaMisra:
This has now been released: https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2020-11-2

az-iot-builder-01:
[...]

PR #1618, opened 2020-10-16
Add an API to set a callback for receiving C2D messages (mqtt)

abhipsaMisra:
Currently we can receive C2D messages on a device client only by calling `ReceiveAsync()` and waiting on the Task.
This PR adds another API using which you can set a callback to be triggered whenever the device receives C2D  messages.

This PR adds the implementation for MQTT, the implementation for AMQP will follow in the next PR.

Issue: #1289 

```csharp
public Task SetReceiveMessageHandlerAsync(ReceiveMessageCallback messageHandler, object userContext, CancellationToken cancellationToken = default(CancellationToken));
public delegate Task ReceiveMessageCallback(Message message, object userContext);
```

Behavior over different transport protocols:
- Http -
 ```csharp
throw new NotSupportedException("HTTP protocol does not support setting callbacks for receiving messages." +
                " You can either call DeviceClient.ReceiveAsync() to wait and receive messages," +
                " or set the callback over MQTT or AMQP.");
```
- Mqtt - 
    - enable callback - subscribe to `devices/{device_id}/messages/devicebound/#` as a Topic Filter.
    - disable callback - unsubscribe from `devices/{device_id}/messages/devicebound/#` topic.
    - once callback is enabled, `public Task<Message> ReceiveAsync();` will not work anymore 
        - it will always return null.
        - we will log the following statement:
```csharp
Logging.Error(this, "Callback handler set for receiving c2d messages, ReceiveAsync() will now always return null", nameof(ReceiveAsync));
```
   - returning `null` instead of throwing an exception will make writing application code easier for our users, else they will need to keep track of their transport protocol used (http only supports `ReceiveAsync()`), and also the latest state (whether they are polling, or have subscribed to a callback). The drawback is that they will need to maintain the processing code for received messages in 2 places.

### DIFF
--- a/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
+++ b/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using FluentAssertions;
 using Microsoft.Azure.Devices.Client;
 using Microsoft.Azure.Devices.E2ETests.Helpers;
 using Microsoft.Azure.Devices.E2ETests.Helpers.Templates;
@@ -148,54 +149,6 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
         }
 
         [LoggedTestMethod]
-        public async Task Message_DeviceReceiveMessageWithLessTimeout_Amqp()
-        {
-            await ReceiveMessageWithTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Amqp_Tcp_Only, s_oneSecond).ConfigureAwait(false);
-        }
-
-        [LoggedTestMethod]
-        public async Task Message_DeviceReceiveMessageWithMoreTimeout_Amqp()
-        {
-            await ReceiveMessageWithTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Amqp_Tcp_Only, s_twentySeconds).ConfigureAwait(false);
-        }
-
-        [LoggedTestMethod]
-        public async Task Message_DeviceReceiveMessageWithLessTimeout_AmqpWs()
-        {
-            await ReceiveMessageWithTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Amqp_WebSocket_Only, s_oneSecond).ConfigureAwait(false);
-        }
-
-        [LoggedTestMethod]
-        public async Task Message_DeviceReceiveMessageWithMoreTimeout_AmqpWs()
-        {
-            await ReceiveMessageWithTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Amqp_WebSocket_Only, s_twentySeconds).ConfigureAwait(false);
-        }
-
-        [LoggedTestMethod]
-        public async Task Message_DeviceReceiveMessageWithLessTimeout_Mqtt()
-        {
-            await ReceiveMessageWithTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_Tcp_Only, s_oneSecond).ConfigureAwait(false);
-        }
-
-        [LoggedTestMethod]
-        public async Task Message_DeviceReceiveMessageWithMoreTimeout_Mqtt()
-        {
-            await ReceiveMessageWithTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_Tcp_Only, s_twentySeconds).ConfigureAwait(false);
-        }
-
-        [LoggedTestMethod]
-        public async Task Message_DeviceReceiveMessageWithLessTimeout_MqttWs()
-        {
-            await ReceiveMessageWithTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_WebSocket_Only, s_oneSecond).ConfigureAwait(false);
-        }
-
-        [LoggedTestMethod]
-        public async Task Message_DeviceReceiveMessageWithMoreTimeout_MqttWs()
-        {
-            await ReceiveMessageWithTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_WebSocket_Only, s_twentySeconds).ConfigureAwait(false);
-        }
-
-        [LoggedTestMethod]
         public async Task Message_DeviceReceiveMessageOperationTimeout_Amqp()
         {
             await ReceiveMessageInOperationTimeoutAsync(TestDeviceType.Sasl, Client.TransportType.Amqp_Tcp_Only).ConfigureAwait(false);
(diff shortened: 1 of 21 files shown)


e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
```diff
@@ -501,32 +482,89 @@ private static async Task ReceiveMessageWithoutTimeoutCheckAsync(DeviceClient dc
             }
         }
 
-        private static async Task ReceiveMessageWithTimeoutCheckAsync(DeviceClient dc, TimeSpan timeout)
```
abhipsaMisra, line 504: This test only asserted that the transport handler returns the ReceiveAsync() call within the timeout specified. It did not actually send any c2d message via the service client, to test the e2e flow. For this reason, I modified this to be a unit test instead.

iothub/device/src/InternalClient.cs
```diff
@@ -1541,7 +1682,7 @@ public async Task SetInputMessageHandlerAsync(string inputName, MessageHandler m
             try
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                await _receiveSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                await _moduleReceiveMessageSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 1685: renamed this to differentiate between the receiving semaphore for devices (newly added) vs the one for modules (pre-existing).

iothub/device/src/Message.cs
```diff
@@ -85,9 +83,12 @@ internal Message(Stream stream, bool ownStream)
         /// + {'-', ':', '/', '\', '.', '+', '%', '_', '#', '*', '?', '!', '(', ')', ',', '=', '@', ';', '$', '''}.
         /// Non-alphanumeric characters are from URN RFC.
         /// </summary>
+        /// <remarks>
+        /// If this value is not supplied by the user, the device client will set this to a new GUID.
+        /// </remarks>
         public string MessageId
         {
-            get => GetSystemProperty<string>(MessageSystemPropertyNames.MessageId);
+            get => GetSystemProperty<string>(MessageSystemPropertyNames.MessageId) ?? Guid.NewGuid().ToString();
```
abhipsaMisra, line 91: This is a change in behavior. Currently the MessageId will be null, if unspecified by the user. This however, makes tracing the message very difficult.

iothub/device/src/Message.cs
```diff
@@ -103,9 +104,9 @@ public string To
         /// <summary>
         /// [Optional] The time when this message is considered expired
         /// </summary>
-        public DateTimeT ExpiryTimeUtc
+        public DateTime ExpiryTimeUtc
```
abhipsaMisra, line 107: not sure why `using DateTimeT = System.DateTime;` alias directive was used.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -256,77 +261,135 @@ public override async Task SendEventAsync(IEnumerable<Message> messages, Cancell
 
         public override async Task<Message> ReceiveAsync(CancellationToken cancellationToken)
         {
-            cancellationToken.ThrowIfCancellationRequested();
+            if (_deviceReceiveMessageCallbackSet)
```
abhipsaMisra, line 264: In case we have set a callback for receiving messages, the call to ReceiveAsync() will always return null.

iothub/service/src/Message.cs
```diff
@@ -101,127 +101,87 @@ internal Message(Stream stream, bool ownStream)
         /// + {'-', ':', '/', '\', '.', '+', '%', '_', '#', '*', '?', '!', '(', ')', ',', '=', '@', ';', '$', '''}.
         /// Non-alphanumeric characters are from URN RFC.
         /// </summary>
+        /// <remarks>
+        /// If this value is not supplied by the user, the service client will set this to a new GUID.
+        /// </remarks>
```
abhipsaMisra, line 106: This is also a "change" in behavior, to assist in better message tracking.

iothub/service/src/Message.cs
```diff
@@ -101,127 +101,87 @@ internal Message(Stream stream, bool ownStream)
...
+            get => GetSystemProperty<DateTime>(MessageSystemPropertyNames.ExpiryTimeUtc);
+            set => SystemProperties[MessageSystemPropertyNames.ExpiryTimeUtc] = value;
         }
 
         /// <summary>
-        /// Used in message responses and feedback
+        /// A string property in a response message that typically contains the MessageId of the request, in request-reply patterns.
         /// </summary>
+        /// <remarks>
+        /// If this value is not supplied by the user, the service client will set this to <see cref="MessageId"/>.
```
abhipsaMisra, line 135: This is based on the ms doc description for c2d message correlation Id: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct#system-properties-of-c2d-iot-hub-messages

iothub/device/src/DeviceClient.cs
```diff
@@ -325,6 +325,18 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
         /// <returns>The receive message or null if there was no message until the specified time has elapsed</returns>
         public Task<Message> ReceiveAsync(TimeSpan timeout) => InternalClient.ReceiveAsync(timeout);
 
+        /// <summary>
+        /// Registers a new delegate for receiving a message from the device queue using the default timeout.
```
bikamani, line 329: can we use a word "Sets" here

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -226,6 +227,90 @@ public override async Task<Message> ReceiveAsync(TimeoutHelper timeoutHelper)
...
 
+        public override async Task EnableReceiveMessageAsync(CancellationToken cancellationToken)
+        {
+            try
+            {
+                if (Logging.IsEnabled)
+                {
+                    Logging.Enter(this, cancellationToken, nameof(EnableReceiveMessageAsync));
+                }
+
```
azabbasi, line 238: I see inconsistent behavior when it comes to 
`cancellationToken.ThrowIfCancellationRequested();`
In methods that take cancellationToken as a parameter.
I think we have to have that statement at the beginning and in between retry attempts

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -226,6 +227,90 @@ public override async Task<Message> ReceiveAsync(TimeoutHelper timeoutHelper)
...
+                await _internalRetryPolicy
+                    .ExecuteAsync(
+                        async () =>
+                        {
+                            // Ensure that the connection has been opened, before enabling the callback for receiving messages.
+                            await EnsureOpenedAsync(cancellationToken).ConfigureAwait(false);
+
+                            // Wait to acquire the _handlerLock. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
+                            await _handlerLock.WaitAsync(cancellationToken).ConfigureAwait(false);
```
azabbasi, line 247: I would rename _handlerLock to _handlerSemaphore to make it clear for the reader that this is a semaphore ... just like what you did in InternalClient

iothub/device/src/Transport/RetryDelegatingHandler.cs
azabbasi: after looking at it, it does seem like all the methods we call already throw if cancellation is requested, so we may not need it after all.

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -226,6 +227,90 @@ public override async Task<Message> ReceiveAsync(TimeoutHelper timeoutHelper)
...
+                            try
+                            {
+                                // The telemetry downlink needs to be enabled only for the first time that the callback is set.
+                                Debug.Assert(!_deviceReceiveMessageEnabled);
+                                await base.EnableReceiveMessageAsync(cancellationToken).ConfigureAwait(false);
+                                _deviceReceiveMessageEnabled = true;
+                            }
+                            finally
+                            {
+                                _handlerLock.Release();
```
azabbasi, line 257: could this throw a SemaphoreFullException?

iothub/service/src/Message.cs
```diff
@@ -1,40 +1,40 @@
...
     public sealed class Message : IDisposable, IReadOnlyIndicator
     {
-        private readonly object messageLock = new object();
-        private volatile Stream bodyStream;
-        private AmqpMessage serializedAmqpMessage;
-        private bool disposed;
-        private bool ownsBodyStream;
-        private int getBodyCalled;
-        private long sizeInBytesCalled;
+        private readonly object _messageLock = new object();
```
azabbasi, line 21: I know this is existing code, but isn't it better if we use semaphores instead of a random object to maintain locks? 
It also help stay consistent in thread safety practices.

iothub/device/src/DeviceClient.cs
```diff
@@ -325,6 +325,18 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+        /// After handling a received message, a client should call <see cref="CompleteAsync(Message, CancellationToken)"/>,
```
bikamani, line 330: Add documentation on how to unset this callback.

iothub/device/src/InternalClient.cs
```diff
@@ -1273,6 +1290,130 @@ public Task<Message> ReceiveAsync(CancellationToken cancellationToken)
             }
         }
 
+        /// <summary>
```
bikamani, line 1293: same comment as above here and everywhere

iothub/device/src/InternalClient.cs
```diff
@@ -1273,6 +1290,130 @@ public Task<Message> ReceiveAsync(CancellationToken cancellationToken)
...
+            {
+                return;
+            }
+
+            Tuple<ReceiveMessageCallback, object> callbackContextTuple = null;
+
+            try
+            {
+                // Wait to acquire the _deviceReceiveMessageSemaphore. This ensures that we get a reference to the latest callback set on the device client.
+                await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
```
bikamani, line 1402: why not use lock?

iothub/device/src/InternalClient.cs
bikamani: Is the underlying implementation different for both?

iothub/device/src/DeviceClient.cs
abhipsaMisra: I followed the wording that we've used for connection status change callback, method callback etc., so I used "Registers" here.

iothub/device/src/InternalClient.cs
```diff
@@ -1273,6 +1290,130 @@ public Task<Message> ReceiveAsync(CancellationToken cancellationToken)
...
+        public async Task SetReceiveMessageHandlerAsync(ReceiveMessageCallback messageHandler, object userContext, CancellationToken cancellationToken)
+        {
+            try
+            {
+                if (Logging.IsEnabled)
+                {
+                    Logging.Enter(this, messageHandler, userContext, nameof(SetReceiveMessageHandlerAsync));
+                }
+
+                cancellationToken.ThrowIfCancellationRequested();
```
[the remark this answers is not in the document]
abhipsaMisra: Well, kind-of yes and no;.
Yes, because yeah, we should not perform big chunks of operations and then only check for cancellations at the beginning.
No, because then we'd be calling it after every other line. 

The way I see it, the Tasks should be broken into logical sections that can be executed as a single block, and then when you enter a method, you check if the cancellation token has been cancelled, if not, then perform the next step of operations.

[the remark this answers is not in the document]
abhipsaMisra: For this scenario, we check if cancellation has been requested, if not, then call enable/disable the c2d subscription, and in there, delegate the cancellation token check to the pipeline/ transport handler.

iothub/device/src/InternalClient.cs
```diff
@@ -1708,14 +1849,14 @@ internal async Task OnReceiveEventMessageCalledAsync(string input, Message messa
                 }
                 finally
                 {
-                    _receiveSemaphore.Release();
+                    _moduleReceiveMessageSemaphore.Release();
```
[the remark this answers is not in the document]
abhipsaMisra: The semaphore.Wait call is outside this try block, so ideally, you would never call Release without having called Wait first. So no, I don't see a scenario where we would release more semaphores than the max count.
Do you see a scenario where we might run into semaphore full exception?

iothub/device/src/Transport/RetryDelegatingHandler.cs
abhipsaMisra: right.

iothub/device/src/Transport/RetryDelegatingHandler.cs
abhipsaMisra: sure, I can do that

iothub/device/src/Transport/RetryDelegatingHandler.cs
abhipsaMisra: no, it shouldn't. We grab the semaphore outside of the try block, so we wouldn't hit this finally block without having grabbed the semaphore first.

iothub/service/src/Message.cs
abhipsaMisra: sure, I can do that

iothub/device/src/DeviceClient.cs
abhipsaMisra: It is present on L332

iothub/device/src/InternalClient.cs
abhipsaMisra: which comment?

iothub/device/src/InternalClient.cs
abhipsaMisra: I don't have a really strong argument for this, other than semaphores being easier to track and maintain, hence my preference.

azabbasi (approved):

PR #1619, opened 2020-10-17
Update TestDeviceCallbackHandler to initialize the callback semaphore correctly

abhipsaMisra:
The callback semaphores being initialized in `TestDeviceCallbackHandler` were not being initialized correctly - the semaphore is released once the callback has been invoked, but because of the way the semaphore was initialized (initial count of 1), it resulted in the semaphore being available to be grabbed even though it had not been released from the callback. 
This will result in the task completing, before any test assertion was carried out within the callback.


e2e/test/Helpers/TestDeviceCallbackHandler.cs
```diff
@@ -2,28 +2,29 @@
...
 
 namespace Microsoft.Azure.Devices.E2ETests.Helpers
 {
     public class TestDeviceCallbackHandler
     {
-        private DeviceClient _deviceClient;
+        private readonly DeviceClient _deviceClient;
+        private readonly MsTestLogger _logger;
 
+        private readonly SemaphoreSlim _methodCallbackSemaphore = new SemaphoreSlim(0, 1);
```
drwill-ms, line 21: Needs disposing

drwill-ms (approved):
Should ensure those get properly disposed.

drwill-ms (approved):

Issue #1026, opened 2019-08-08
No way to remove IoTHub subscriptions once they have been set


varunpuranik:
## Description of the issue:
Currently the C# SDK does not provide a way to remove subscriptions for DirectMethods, DesiredPropertyUpdates or C2D messages once they have been set. 

The EdgeHub has this requirement where it sets up subscriptions to IoTHub on behalf of the client (module or device) when the client sets up those subscriptions to EdgeHub. That allows one to trigger those subscriptions from the cloud.
But if the client disconnects, then EdgeHub needs to remove those subscriptions, so that they cannot be triggered from IoTHub. (Note that it cannot simply drop the connection for the client since there might be D2C telemetry that EdgeHub needs to send on behalf of the client)

AB#7304414

prmathur-microsoft:
Thanks for filing request. We will keep this in our backlog for now and will add this feature in future.

RubiaL:
Thanks for the feedback.This is currently in our backlog and will consider it as part of our planning cycle

az-iot-builder-01:
@prmathur-microsoft, @RubiaL, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

and-rewsmith:
@RubiaL Is this still on the backlog?

philipktlin:
Please re-open this issue and provide an update about this bug in the backlog.  thanks.

and-rewsmith:
@prmathur-microsoft Is there a rough estimate for when this will get prioritized?

timtay-microsoft:
@bikamani is taking a look at this now. From the look of things, IotHub itself does not respect MQTT unsubscribe messages, so it may take a while for the SDK to support this scenario.

abhipsaMisra:
Turns out that we can unsubscribe from MQTT topics, so this fix has been added for both MQTT (will unsubscribe from topic) and AMQP (will close the links).
This has now been released: https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2020-11-2

PR #1621, opened 2020-10-21
feat(device): Add support for disabling callbacks for properties and methods

bikamani:
## Checklist

## Description of the changes

- Add support for disabling callbacks for properties and methods

- https://github.com/Azure/azure-iot-sdk-csharp/issues/1026


iothub/device/src/InternalClient.cs
```diff
@@ -1225,26 +1225,26 @@ public async Task SetDesiredPropertyUpdateCallbackAsync(DesiredPropertyUpdateCal
...
+            try
             {
-                try
+                if (callback != null && !_patchSubscribedWithService)
                 {
                     await InnerHandler.EnableTwinPatchAsync(cancellationToken).ConfigureAwait(false);
                     _patchSubscribedWithService = true;
                 }
-                catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
+                else if(callback == null && _patchSubscribedWithService)
```
drwill-ms, line 1237: missing a space: `if (`

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -528,6 +528,45 @@ private async Task OpenMethodsReceiverLinkAsync(AmqpIoTSession amqpIoTSession, s
             }
         }
 
+        public async Task DisableTwinPatchAsync(TimeSpan timeout)
```
drwill-ms, line 531: presumably the timeout is for the entire operation, but we're using it in every method call below, which would mean the total timeout would be multiples of `timeout`, right?

Seems we should be decrementing timeout as we go.

iothub/device/src/InternalClient.cs
```diff
@@ -1225,26 +1225,26 @@ public async Task SetDesiredPropertyUpdateCallbackAsync(DesiredPropertyUpdateCal
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         public async Task SetDesiredPropertyUpdateCallbackAsync(DesiredPropertyUpdateCallback callback, object userContext, CancellationToken cancellationToken)
```
drwill-ms, line 1226: Interesting choice of user passing `null` to this, instead of another method to clear the update callback. Passing null seems less obvious to me. Thoughts?

bikamani: We do this for Events already - DisableEventReceiveAsync(). Also we discussed with team on both approaches and this is the one we agreed upon. If we decide to do otherwise - we will have at least 4 new APIs unsubscribe for events, methods, properties, c2d. 

I agree it is not obvious and documenting this would be necessary here.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
bikamani: Timeout is for the underlying amqp operations. We just need to pass it down and the amqp library takes care of rest.

iothub/device/src/InternalClient.cs
drwill-ms: Let's chat offline.

drwill-ms: Makes sense. I agree this is an acceptable approach, and prefer not to over-invest in track1.

iothub/device/src/InternalClient.cs
```diff
@@ -1225,26 +1225,26 @@ public async Task SetDesiredPropertyUpdateCallbackAsync(DesiredPropertyUpdateCal
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
```
drwill-ms, line 1225: Let's update this in a few ways:

1. Summary sets you can use it to set or clear/remove the callback.
2. For the second sentence, move it to `<remarks>`.

iothub/device/tests/DeviceClientTwinApiTests.cs
```diff
@@ -59,6 +59,26 @@ await innerHandler.
...
+            client.InnerHandler = innerHandler;
+            DesiredPropertyUpdateCallback myCallback = (p, c) => TaskHelpers.CompletedTask;
+            var context = new object();
+
+            // act
+            await client.SetDesiredPropertyUpdateCallbackAsync(myCallback, context).ConfigureAwait(false);
+            await client.SetDesiredPropertyUpdateCallbackAsync(null, null).ConfigureAwait(false);
+
+            // assert
+            await innerHandler.
```
drwill-ms, line 77: preferred syntax:

```csharp
await innerHandler
    .Received(1)
    .DisableTwinPatchAsync(...)
    .ConfigureAwait(false);
```

iothub/device/tests/DeviceClientTests.cs
```diff
@@ -157,6 +157,28 @@ public async Task DeviceClient_OperationTimeoutInMilliseconds_Equals_0_Receive()
...
+        [TestMethod]
+        public async Task DeviceClient_OnMethodCalled_Unsubscribe()
+        {
+            // arrange
+            DeviceClient deviceClient = DeviceClient.CreateFromConnectionString(fakeConnectionString);
+            var innerHandler = Substitute.For<IDelegatingHandler>();
+            deviceClient.InnerHandler = innerHandler;
+
+            // act
+            await deviceClient.SetMethodHandlerAsync("TestMethodName", (payload, context) =>
```
drwill-ms, line 169: preferred syntax:

```csharp
await deviceClient
    .SetMethodHandlerAsync(
        "TestMethodName",
        (payload, context) =>Task.FromResult(new MethodResponse(Encoding.UTF8.GetBytes("{\"name\":\"ABC\"}"), 200));
        "custom data")
    .ConfigureAwait(false);
```

drwill-ms (approved):

iothub/device/src/DeviceClient.cs
```diff
@@ -529,6 +529,7 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
         /// <summary>
         /// Registers a new delegate for the named method. If a delegate is already associated with
         /// the named method, it will be replaced with the new delegate.
+        /// A method hanler can be unset by passing null.
```
drwill-ms, line 532: handler

iothub/device/src/DeviceClient.cs
```diff
@@ -606,19 +610,24 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
 
         /// <summary>
         /// Set a callback that will be called whenever the client receives a state update
-        /// (desired or reported) from the service.  This has the side-effect of subscribing
-        /// to the PATCH topic on the service.
+        /// (desired or reported) from the service.
         /// </summary>
+        /// <remarks>
+        /// This has the side-effect of subscribing to the PATCH topic on the service.
+        /// Set callback value to null to clear.
```
drwill-ms, line 629: I'd put this line in summary.

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -279,6 +337,24 @@ public static async Task ServiceSendMethodAndVerifyResponseAsync(string deviceId
             serviceClient.Dispose();
         }
 
+        public static async Task<Task> SubscribeAndUnsubscribeMethodAsync(DeviceClient deviceClient, string methodName, MsTestLogger logger)
+        {
+            var methodCallReceived = new TaskCompletionSource<bool>();
+
+            await deviceClient.SetMethodHandlerAsync(methodName,
```
drwill-ms, line 344: syntax here could use some clean up, like reported elsewhere

e2e/test/iothub/twin/TwinE2ETests.cs
```diff
@@ -379,7 +403,42 @@ public static async Task RegistryManagerUpdateDesiredPropertyAsync(string device
...
+                        // After unsubscribing it should never reach here
+                        Assert.IsNull(patch);
+
+                        return Task.FromResult<bool>(true);
+                    },
+                    null)
+                .ConfigureAwait(false);
+
+            // Unsubscribe
+            await deviceClient.
```
drwill-ms, line 431: syntax here could use some clean up, like reported elsewhere

iothub/device/tests/DeviceClientTests.cs
```diff
@@ -157,6 +157,29 @@ public async Task DeviceClient_OperationTimeoutInMilliseconds_Equals_0_Receive()
...
+        {
+            // arrange
+            DeviceClient deviceClient = DeviceClient.CreateFromConnectionString(fakeConnectionString);
+            var innerHandler = Substitute.For<IDelegatingHandler>();
+            deviceClient.InnerHandler = innerHandler;
+
+            // act
+            await deviceClient
+                .SetMethodHandlerAsync(
+                "TestMethodName",
```
drwill-ms, line 171: tab in params

iothub/device/tests/DeviceClientTests.cs
```diff
@@ -157,6 +157,29 @@ public async Task DeviceClient_OperationTimeoutInMilliseconds_Equals_0_Receive()
...
+            await deviceClient
+                .SetMethodHandlerAsync(
+                "TestMethodName", 
+                (payload, context) => Task.FromResult(new MethodResponse(Encoding.UTF8.GetBytes("{\"name\":\"ABC\"}"), 200)), "custom data")
+                .ConfigureAwait(false);
+
+            await deviceClient.SetMethodHandlerAsync("TestMethodName", null, null).ConfigureAwait(false);
+
+            // assert
+            await innerHandler.
```
drwill-ms, line 178: syntax here could use some clean up, like reported elsewhere

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -421,6 +500,18 @@ public static async Task<Task> SetModuleReceiveMethodDefaultHandlerAsync(ModuleC
...
+        private async Task SendMethodAndUnsubscribeAsync(Client.TransportType transport, Func<DeviceClient, string, MsTestLogger, Task<Task>> subscribeAndUnsubscribeMethod, TimeSpan responseTimeout = default, ServiceClientTransportSettings serviceClientTransportSettings = default)
+        {
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix).ConfigureAwait(false);
+            using var deviceClient = DeviceClient.CreateFromConnectionString(testDevice.ConnectionString, transport);
+
+            await subscribeAndUnsubscribeMethod(deviceClient, MethodName, Logger).ConfigureAwait(false);
+
+            await ServiceSendMethodAndVerifyNotReceivedAsync(testDevice.Id, MethodName, Logger, responseTimeout, serviceClientTransportSettings).ConfigureAwait(false);
+
+            await deviceClient.CloseAsync().ConfigureAwait(false);
```
abhipsaMisra, line 512: what do you think about testing the following flow:
- subscribe to method -> invoke method -> pass -> unsubscribe -> invoke method -> fail.

Currently we subscribe and unsubscribe without any invocation in between, so even if subscribe was a no-op call, we wouldn't know that. I know that there are other tests that test subscribe operation specifically, but it seemed like a gap to me.
Thoughts?

Also, instead of annotating the test method with a "throws DeviceNotFoundException", we would specifically assert that `ServiceSendMethodAndVerifyNotReceivedAsync` throws DeviceNotFoundException. Otherwise, right now we have multiple API calls within this test method, and either of them could throw this exception (not the API call that we are targetting) and the test would still pass.

e2e/test/iothub/twin/TwinE2ETests.cs
```diff
@@ -379,7 +403,46 @@ public static async Task RegistryManagerUpdateDesiredPropertyAsync(string device
...
+                .ConfigureAwait(false);
+
+            await RegistryManagerUpdateDesiredPropertyAsync(testDevice.Id, propName, propValue)
+                .ConfigureAwait(false);
+
+            await deviceClient
+                .SetDesiredPropertyUpdateCallbackAsync(null, null)
+                .ConfigureAwait(false);
+
+            await deviceClient.CloseAsync().ConfigureAwait(false);
```
abhipsaMisra, line 442: same comment here about the flow -> subscribe -> test that it is received -> unsubscribe -> test not received.

Also, why is
```csharp
await deviceClient
                .SetDesiredPropertyUpdateCallbackAsync(null, null)
                .ConfigureAwait(false);
```
called twice (L431 and L438)?

iothub/device/src/DeviceClient.cs
```diff
@@ -527,8 +527,9 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
             InternalClient.CompleteFileUploadAsync(notification, cancellationToken);
 
         /// <summary>
-        /// Registers a new delegate for the named method. If a delegate is already associated with
+        /// Sets a new delegate for the named method. If a delegate is already associated with
```
abhipsaMisra, line 530: Oh, you are updating the text here to say "Sets" instead of "Registers"! Now I understand your comment on the ReceiveAsync() PR! 😉 

Do you prefer "set and unset" vs "register and deregister"?

iothub/device/src/InternalClient.cs
```diff
@@ -1071,37 +1072,39 @@ public async Task SetDesiredPropertyUpdateCallbackAsync(DesiredPropertyUpdateCal
...
-
             // Codes_SRS_DEVICECLIENT_18_003: `SetDesiredPropertyUpdateCallbackAsync` shall call the transport to register for PATCHes on it's first call.
             // Codes_SRS_DEVICECLIENT_18_004: `SetDesiredPropertyUpdateCallbackAsync` shall not call the transport to register for PATCHes on subsequent calls
-            if (!_twinPatchSubscribedWithService)
+            try
             {
-                try
+                if (callback != null && !_twinPatchSubscribedWithService)
                 {
                     await InnerHandler.EnableTwinPatchAsync(cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 1092: shouldn't you grab a lock before enabling the twin patches/ disabling them? If you have parallel calls that both set and unset the callback, what is the internal state going to be then?

iothub/device/src/InternalClient.cs
```diff
@@ -1071,37 +1072,39 @@ public async Task SetDesiredPropertyUpdateCallbackAsync(DesiredPropertyUpdateCal
...
                     _twinPatchSubscribedWithService = true;
                 }
-                catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
+                else if (callback == null && _twinPatchSubscribedWithService)
                 {
-                    cancellationToken.ThrowIfCancellationRequested();
-                    throw;
+                    await InnerHandler.DisableTwinPatchAsync(cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 1097: It looks like `EnableTwinPatchAsync` and `DisableTwinPatchAsync` both delegate the operation to the pipeline/ transport layer directly. Is that where you check if enabling/ disabling is required (enable only for the first time, disable only if it has been previously enabled)?

iothub/device/src/ModuleClient.cs
```diff
@@ -399,17 +401,19 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
             InternalClient.SetMethodHandlerAsync(methodName, methodHandler, userContext, cancellationToken);
 
         /// <summary>
-        /// Registers a new delegate that is called for a method that doesn't have a delegate registered for its name.
+        /// Sets a new delegate that is called for a method that doesn't have a delegate registered for its name.
         /// If a default delegate is already registered it will replace with the new delegate.
+        /// A method handler can be unset by passing null.
```
abhipsaMisra, line 406: what do you think about elaborating this statement a bit more:
If a null delegate is passed, it will disable triggering any callback on method invocation from the service.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -528,6 +528,45 @@ private async Task OpenMethodsReceiverLinkAsync(AmqpIoTSession amqpIoTSession, s
...
+            {
+                ICollection<Task> tasks = new List<Task>();
+                if (_twinReceivingLink != null)
+                {
+                    tasks.Add(_twinReceivingLink.CloseAsync(timeout));
+                }
+
+                if (_twinSendingLink != null)
+                {
+                    tasks.Add(_twinSendingLink.CloseAsync(timeout));
```
abhipsaMisra, line 551: Is there a reason why we are closing the sending link as well? Disabling desired property updates should not have any impact on sending links.

Even for receiving links, I believe we are using the same receiving link to receive desired property updates and receive `GetTwinAsync()` on, right? Do we have proper locks to ensure that calling unsubscribe on twin updates and calling GetTwin() in parallel doesn't cause our internal state to get messed up?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -808,6 +808,14 @@ public override async Task EnableTwinPatchAsync(CancellationToken cancellationTo
             await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(TwinPatchTopicFilter, QualityOfService.AtMostOnce))).ConfigureAwait(true);
         }
 
+        public override async Task DisableTwinPatchAsync(CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
```
abhipsaMisra, line 813: These would benefit from some log statements here, at Enter and Exit

iothub/device/tests/DeviceClientTests.cs
```diff
@@ -157,6 +157,32 @@ public async Task DeviceClient_OperationTimeoutInMilliseconds_Equals_0_Receive()
...
+        [TestMethod]
+        public async Task DeviceClient_OnMethodCalled_Unsubscribe()
+        {
+            // arrange
+            DeviceClient deviceClient = DeviceClient.CreateFromConnectionString(fakeConnectionString);
+            var innerHandler = Substitute.For<IDelegatingHandler>();
+            deviceClient.InnerHandler = innerHandler;
+
+            // act
+            await deviceClient
```
abhipsaMisra, line 169: this call can be skipped for this test, right?

iothub/device/tests/DeviceClientTwinApiTests.cs
```diff
@@ -59,6 +59,27 @@ await innerHandler.
...
+        {
+            // arrange
+            var innerHandler = Substitute.For<IDelegatingHandler>();
+            var client = DeviceClient.CreateFromConnectionString(fakeConnectionString);
+            client.InnerHandler = innerHandler;
+            DesiredPropertyUpdateCallback myCallback = (p, c) => TaskHelpers.CompletedTask;
+            var context = new object();
+
+            // act
+            await client.SetDesiredPropertyUpdateCallbackAsync(myCallback, context).ConfigureAwait(false);
```
abhipsaMisra, line 73: same here, this call can also be skipped.

abhipsaMisra:
Left some questions about support for parallel calls and links being closed.

e2e/test/iothub/method/MethodE2ETests.cs
bikamani: Can you explain what you mean by "if subscribe was a no-op call"?

Currently the flow we have is testing unsubscribe specifically. I do agree we should assert specifically for devicenotfound. I will add that  change.

e2e/test/iothub/twin/TwinE2ETests.cs
bikamani: Good catch. I'll remove duplicate code.

iothub/device/src/DeviceClient.cs
bikamani: David thinks that's much simpler and easy to understand :)

iothub/device/src/InternalClient.cs
bikamani: Agree. It is something I have thought about, not addressing in this PR though. Let's discuss this offline.

iothub/device/src/InternalClient.cs
bikamani: correct. we have a bool for this check.

iothub/device/src/ModuleClient.cs
bikamani: I had it like that and feedback was to keep it short and simple. Is it unclear to you? If so, we should chat offline.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
bikamani: we open both sending and receiving link as part of enabling twin patch call so I suppose disabling should close both. 

Yes have a todo to make this thread safe.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
bikamani: Totally, will add to my list of todos.

iothub/device/tests/DeviceClientTests.cs
bikamani: It is required since we need to enabled before disabling. I'll check on this though.

iothub/device/tests/DeviceClientTwinApiTests.cs
bikamani: No we need to set before unset.

e2e/test/iothub/method/MethodE2ETests.cs
abhipsaMisra: what I mean about it being a no-op call is, if subscribe wasn't actually subscribing to anything (intentionally, or there was a bug).

iothub/device/src/DeviceClient.cs
abhipsaMisra: Ok then, would you mind scanning through the rest of the comments as well, and make this change.

iothub/device/src/InternalClient.cs
abhipsaMisra: Sure, we can discuss this offline.

iothub/device/src/InternalClient.cs
abhipsaMisra: got it!

iothub/device/src/ModuleClient.cs
abhipsaMisra: I would prefer if we specified whic argument is expected to be null:
A method handler can be unset by passing a null `MethodCallback`.

iothub/device/src/DeviceClient.cs
bikamani: I had done it for all but may be the newer one may not be following this, is that what you meant?

bikamani: Changed it for new callback you added

iothub/device/src/Transport/Amqp/AmqpUnit.cs
abhipsaMisra: Twins would be a special case in that the same twin sending and receiving links are used for sending reported property updates, receiving desired property updated and retrieving the twin object.
If we close links on unsubscribe, we have to be very careful and put locks before we do that. That way it won't conflict with a separate twin operation attempting to open the links.
The other alternative is to simply stop processing unsolicited response.

What is the current behavior if we call GetTwinAsync() -> which will open both sending and receiving links, but don't subscribe to desired property updates. If I update a property from service, does the client receive it but then choses not to process it?

iothub/device/src/InternalClient.cs
```diff
@@ -1058,17 +1062,14 @@ private Task EnableMethodAsync(CancellationToken cancellationToken)
...
-        {
-            // TODO # 890.
-            return TaskHelpers.CompletedTask;
-        }
-
         /// <summary>
-        /// Set a callback that will be called whenever the client receives a state update
-        /// (desired or reported) from the service.  This has the side-effect of subscribing
-        /// to the PATCH topic on the service.
+        /// Set and clear a callback that will be called whenever the client receives a state update
```
drwill-ms, line 1066: Set **or** clear

iothub/device/src/Transport/Amqp/AmqpUnit.cs
bikamani: Are you talking about Enabling Twin which in turn opens links? It already exists at both these places so should be the same flow.

bikamani: For unsubscribe - we will be closing sending and receiving links. Next call to use those links will open them.

iothub/device/src/InternalClient.cs
bikamani: Addressed this.

iothub/device/src/InternalClient.cs
bikamani: done

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -528,16 +528,22 @@ private async Task OpenMethodsReceiverLinkAsync(AmqpIoTSession amqpIoTSession, s
...
             if (Logging.IsEnabled)
             {
-                Logging.Enter(this, timeout, $"{nameof(DisableTwinPatchAsync)}");
+                Logging.Enter(this, timeout, $"{nameof(DisableTwinLinksAsync)}");
             }
 
             Debug.Assert(_twinSendingLink != null);
             Debug.Assert(_twinReceivingLink != null);
 
+            bool gain = await _twinLinksSemaphore.WaitAsync(timeout).ConfigureAwait(false);
```
abhipsaMisra, line 541: what is the timeout here?

bikamani: It's operation timeout same as what is used with _messageReceivingLinkSemaphore

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -221,6 +247,39 @@ public async Task Method_ServiceInvokeDeviceMethodWithUnknownModuleThrows()
...
+            try
+            {
+                CloudToDeviceMethodResult response =
+                    await serviceClient.InvokeDeviceMethodAsync(
+                        deviceId,
+                        new CloudToDeviceMethod(methodName, methodTimeout).SetPayloadJson(null)).ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                if (!(ex is DeviceNotFoundException))
```
abhipsaMisra, line 273: this should always mandatorily throw an exception correct, so if response is returned then that would be a "fail" case.

you can instead do:
```suggestion
               await Assert.ThrowsExceptionAsync<DeviceNotFoundException>(async () =>
            {
                await serviceClient.InvokeDeviceMethodAsync(
                        deviceId,
                        new CloudToDeviceMethod(methodName, methodTimeout).SetPayloadJson(null)).ConfigureAwait(false);
            });
```

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -279,6 +338,27 @@ public static async Task ServiceSendMethodAndVerifyResponseAsync(string deviceId
...
+        public static async Task<Task> SubscribeAndUnsubscribeMethodAsync(DeviceClient deviceClient, string methodName, MsTestLogger logger)
+        {
+            var methodCallReceived = new TaskCompletionSource<bool>();
+
+            await deviceClient
+                .SetMethodHandlerAsync(
+                methodName,
+                (request, context) =>
+                {
+                    logger.Trace($"{nameof(SubscribeAndUnsubscribeMethodAsync)}: DeviceClient method: {request.Name} {request.ResponseTimeout}.");
```
abhipsaMisra, line 350: nit: none of this code is required, since this callback is never triggered by the test device

bikamani: We still need to set a callback though. Set -> Unset sequence

e2e/test/iothub/method/MethodE2ETests.cs
bikamani: I did not get it. We can discuss offline

iothub/device/src/InternalClient.cs
```diff
@@ -716,8 +716,9 @@ public Task SendEventBatchAsync(IEnumerable<Message> messages, CancellationToken
         }
 
         /// <summary>
-        /// Registers a new delegate for the named method. If a delegate is already associated with
+        /// Sets a new delegate for the named method. If a delegate is already associated with
```
abhipsaMisra, line 719: Sets or clears?

iothub/device/src/ModuleClient.cs
```diff
@@ -433,19 +437,24 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
 
         /// <summary>
         /// Set a callback that will be called whenever the client receives a state update
-        /// (desired or reported) from the service.  This has the side-effect of subscribing
```
abhipsaMisra, line 436: Why did we chose to remove this explanation?

abhipsaMisra (approved):
We need to add locks before we enable (and set) or disable (and remove) the callback on the `InternalClient` (to ensure we get the latest reference).
Other than that, some minor comments, but nothing blocking.

e2e/test/iothub/method/MethodE2ETests.cs
abhipsaMisra: Yes, you need to set the callback, but it doesn't need any of these log statements (hence the "nit" tag 😄 )

iothub/device/src/ModuleClient.cs
bikamani: We did not remove. It is part of remarks

e2e/test/iothub/method/MethodE2ETests.cs
abhipsaMisra: sure, give me call when you want to discuss this

PR #1622, opened 2020-10-22
Add logging framework for ServiceClient

abhipsaMisra:
This PR enables ETW logs for the service client.
It also adds a trace provider for the AMQP library, enabling us to write the AMQP logs into our library's log provider (ETW).


iothub/service/src/AmqpServiceClient.cs
```diff
@@ -50,6 +52,9 @@ public AmqpServiceClient(IotHubConnectionString iotHubConnectionString, bool use
                 ExceptionHandlingHelper.GetDefaultErrorMapping(),
```
azabbasi, line 52: nit: strange indentation here

azabbasi (approved):

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: these are args for the constructor, hence the extra tab :)

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -12,6 +12,8 @@
 using Microsoft.Azure.Devices.Common;
 using Microsoft.Azure.Devices.Common.Data;
 using Microsoft.Azure.Devices.Common.Exceptions;
+using Microsoft.Azure.Devices.Shared;
+using AmqpTrace = Microsoft.Azure.Amqp.AmqpTrace;
```
drwill-ms, line 16: Was this alias necessary?

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -88,6 +93,11 @@ public async override Task CloseAsync()
 
         public async override Task SendAsync(string deviceId, Message message, TimeSpan? timeout = null)
         {
+            if (Logging.IsEnabled)
+            {
+                Logging.Enter(this, $"Sending message with Id [{message.MessageId}] for device [{deviceId}]", nameof(SendAsync));
```
drwill-ms, line 98: What if message is null?

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -106,6 +116,11 @@ public async override Task SendAsync(string deviceId, Message message, TimeSpan?
             {
                 SendingAmqpLink sendingLink = await GetSendingLinkAsync().ConfigureAwait(false);
 
+                if (Logging.IsEnabled)
+                {
+                    Logging.Info(this, $"Retrieved SendingAmqpLink [{sendingLink.Name}]", nameof(SendAsync));
```
drwill-ms, line 121: What if sendingLink is null?

iothub/service/src/AmqpTransportLog.cs
```diff
@@ -0,0 +1,289 @@
...
+            }
+
+            base.AmqpAbortThrowingException(exception);
+        }
+
+        public override void AmqpAddSession(object source, object session, ushort localChannel, ushort remoteChannel)
+        {
+            if (Logging.IsEnabled)
+            {
+                Logging.Associate(source, session, "AmqpAddSession:", $"{" source:"}{source}.{" session:"}{session}.{" localChannel:"}.{localChannel}.{" remoteChannel:"}.{remoteChannel}");
```
drwill-ms, line 28: What does this string interpolation mean:
`{" source:"}`

iothub/service/src/AmqpTransportLog.cs
```diff
@@ -0,0 +1,289 @@
...
+            }
+
+            base.AmqpAddSession(source, session, localChannel, remoteChannel);
+        }
+
+        public override void AmqpAttachLink(object connection, object session, object link, uint localHandle, uint remoteHandle, string linkName, string role, object source, object target)
```
drwill-ms, line 34: long list of parameters, best break up 1 per line

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: Yeah, we have an "empty" `AmqpTrace` in our commond directory, which was creating a conflict: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/service/src/Common/AmqpTrace.cs

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: good point, I'll check if we do allow message to be null, and if yes, then update this

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: It cannot be, this is a result of `_faultTolerantSendingLink.GetOrCreateAsync()` which will give us a link, or throw an exception (unless there is a bug in the amqp lib). But this is a good point, I'll add a null check here and also I'll be adding more logs in our service client that will log the state for these link creation as well.

iothub/service/src/AmqpTransportLog.cs
abhipsaMisra: yeah, that is weird, I'll fix it.

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: Interestingly message can be null, i'll update the log statement

iothub/service/src/AmqpServiceClient.cs
drwill-ms: Yikes. I know it would be pulling a type out, but is it worth evaluating whether we should do that? What would the negative impact be on a customer?

iothub/service/src/AmqpServiceClient.cs
drwill-ms: Logging should never cause an exception, so we have to be sure it is always safe.

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: Interestingly enough, it is an internal class. I say, yes, let's remove it.

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: yes, I agree 100%.

PR #1625, opened 2020-10-23
Add logs to AmqpServiceClient

azabbasi:
Our log statements do not require a check for IsEnabled since that is already checked in the first implementation layer of the Logging class.

This is PR is just a sample on how the rest of the service client is going to look like when we are done with log instrumentations.

I've also updated the MessageId default behavior implementation in this PR -> we now take in a flag `setDefaultMessageId` in `Message` constructors. If set to true, the client library will set the MessageId to a GUID.

### DIFF
--- a/common/src/Logging.Common.cs
+++ b/common/src/Logging.Common.cs
@@ -62,10 +62,11 @@ namespace Microsoft.Azure.Devices.Shared
         [NonEvent]
         public static void Enter(object thisOrContextObject, FormattableString formattableString = null, [CallerMemberName] string memberName = null)
         {
-            DebugValidateArg(thisOrContextObject);
-            DebugValidateArg(formattableString);
             if (IsEnabled)
             {
+                DebugValidateArg(thisOrContextObject);
+                DebugValidateArg(formattableString);
+
                 Log.Enter(IdOf(thisOrContextObject), memberName, formattableString != null ? Format(formattableString) : NoParameters);
             }
         }
@@ -79,10 +80,11 @@ namespace Microsoft.Azure.Devices.Shared
         [NonEvent]
         public static void Enter(object thisOrContextObject, object arg0, [CallerMemberName] string memberName = null)
         {
-            DebugValidateArg(thisOrContextObject);
-            DebugValidateArg(arg0);
             if (IsEnabled)
             {
+                DebugValidateArg(thisOrContextObject);
+                DebugValidateArg(arg0);
+
                 Log.Enter(IdOf(thisOrContextObject), memberName, $"({Format(arg0)})");
             }
         }
@@ -95,11 +97,12 @@ namespace Microsoft.Azure.Devices.Shared
         [NonEvent]
         public static void Enter(object thisOrContextObject, object arg0, object arg1, [CallerMemberName] string memberName = null)
         {
-            DebugValidateArg(thisOrContextObject);
-            DebugValidateArg(arg0);
-            DebugValidateArg(arg1);
             if (IsEnabled)
             {
+                DebugValidateArg(thisOrContextObject);
+                DebugValidateArg(arg0);
+                DebugValidateArg(arg1);
+
                 Log.Enter(IdOf(thisOrContextObject), memberName, $"({Format(arg0)}, {Format(arg1)})");
             }
         }
@@ -113,12 +116,13 @@ namespace Microsoft.Azure.Devices.Shared
         [NonEvent]
         public static void Enter(object thisOrContextObject, object arg0, object arg1, object arg2, [CallerMemberName] string memberName = null)
         {
-            DebugValidateArg(thisOrContextObject);
-            DebugValidateArg(arg0);
-            DebugValidateArg(arg1);
-            DebugValidateArg(arg2);
             if (IsEnabled)
             {
+                DebugValidateArg(thisOrContextObject);
+                DebugValidateArg(arg0);
+                DebugValidateArg(arg1);
+                DebugValidateArg(arg2);
+
                 Log.Enter(IdOf(thisOrContextObject), memberName, $"({Format(arg0)}, {Format(arg1)}, {Format(arg2)})");
             }
         }
@@ -140,10 +144,11 @@ namespace Microsoft.Azure.Devices.Shared
         [NonEvent]
         public static void Exit(object thisOrContextObject, FormattableString formattableString = null, [CallerMemberName] string memberName = null)
         {
-            DebugValidateArg(thisOrContextObject);
-            DebugValidateArg(formattableString);
             if (IsEnabled)
             {
+                DebugValidateArg(thisOrContextObject);
+                DebugValidateArg(formattableString);
+
                 Log.Exit(IdOf(thisOrContextObject), memberName, formattableString != null ? Format(formattableString) : NoParameters);
             }
         }
@@ -157,10 +162,11 @@ namespace Microsoft.Azure.Devices.Shared
         [NonEvent]
         public static void Exit(object thisOrContextObject, object arg0, [CallerMemberName] string memberName = null)
         {
-            DebugValidateArg(thisOrContextObject);
-            DebugValidateArg(arg0);
             if (IsEnabled)
             {
+                DebugValidateArg(thisOrContextObject);
+                DebugValidateArg(arg0);
+
                 Log.Exit(IdOf(thisOrContextObject), memberName, Format(arg0).ToString());
             }
         }
@@ -173,11 +179,12 @@ namespace Microsoft.Azure.Devices.Shared
         [NonEvent]
         public static void Exit(object thisOrContextObject, object arg0, object arg1, [CallerMemberName] string memberName = null)
         {
-            DebugValidateArg(thisOrContextObject);
-            DebugValidateArg(arg0);
-            DebugValidateArg(arg1);
             if (IsEnabled)
             {
+                DebugValidateArg(thisOrContextObject);
+                DebugValidateArg(arg0);
+                DebugValidateArg(arg1);
+
                 Log.Exit(IdOf(thisOrContextObject), memberName, $"{Format(arg0)}, {Format(arg1)}");
             }
         }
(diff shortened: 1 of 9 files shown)


abhipsaMisra (approved):

iothub/device/src/Message.cs
```diff
@@ -51,28 +59,27 @@ public Message(Stream stream)
...
         /// we can clean up the stream.
         /// </summary>
-        /// <param name="stream"></param>
-        /// <param name="ownStream"></param>
-        internal Message(Stream stream, bool ownStream)
-            : this(stream)
+        /// <param name="stream">A stream which will be used as body stream.</param>
+        /// <param name="ownStream">A flag indicating that the client library will dispose the stream when the Message is disposed.</param>
+        /// <param name="setDefaultMessageId">A flag indicating if the MessageId should be set to a random GUID.</param>
+        internal Message(Stream stream, bool ownStream, bool setDefaultMessageId = false)
```
abhipsaMisra, line 81: This can potentially cause confusion with
```csharp
public Message(Stream stream, bool setDefaultMessageId = false)
```
 but this method is internal; so we can be careful about its usage.
This does come across to me as bad code though.

iothub/device/src/Message.cs
```diff
@@ -28,21 +28,29 @@ public sealed class Message : IReadOnlyIndicator, IDisposable
         /// <summary>
         /// Default constructor with no body data
         /// </summary>
-        public Message()
+        /// <param name="setDefaultMessageId">A flag indicating if the MessageId should be set to a random GUID.</param>
+        public Message(bool setDefaultMessageId = false)
```
abhipsaMisra, line 32: @drwill-ms The messageId changes, as we discussed

common/src/Logging.Common.cs
```diff
@@ -62,10 +62,11 @@ public static class Keywords
         [NonEvent]
         public static void Enter(object thisOrContextObject, FormattableString formattableString = null, [CallerMemberName] string memberName = null)
         {
-            DebugValidateArg(thisOrContextObject);
-            DebugValidateArg(formattableString);
             if (IsEnabled)
             {
+                DebugValidateArg(thisOrContextObject);
```
abhipsaMisra, line 67: We need to validate args only if logging is enabled.

iothub/device/src/Message.cs
```diff
@@ -83,12 +90,9 @@ internal Message(Stream stream, bool ownStream)
         /// + {'-', ':', '/', '\', '.', '+', '%', '_', '#', '*', '?', '!', '(', ')', ',', '=', '@', ';', '$', '''}.
         /// Non-alphanumeric characters are from URN RFC.
         /// </summary>
-        /// <remarks>
-        /// If this value is not supplied by the user, the device client will set this to a new GUID.
-        /// </remarks>
         public string MessageId
         {
-            get => GetSystemProperty<string>(MessageSystemPropertyNames.MessageId) ?? Guid.NewGuid().ToString();
```
abhipsaMisra, line 91: This was just wrong!

iothub/device/src/Message.cs
azabbasi, line 81: `ownStream` doesn't really describe what the impact is. I talked to @abhipsaMisra offline and we came up with `shouldDisposeStream` as an alternative since the impact of setting this flag to `true` is that the client will dispose the stream ... thoughts?

common/src/Logging.Common.cs
azabbasi: Thank you for fixing all that

iothub/device/src/Message.cs
```diff
@@ -51,28 +59,27 @@ public Message(Stream stream)
...
-            byte[] byteArray)
-            : this(new MemoryStream(byteArray))
+        /// <remarks>User should treat the input byte array as immutable when sending the message.</remarks>
+        /// <param name="byteArray">A byte array which will be used to form the body stream.</param>
+        /// <param name="setDefaultMessageId">A flag indicating if the MessageId should be set to a random GUID.</param>
+        public Message(byte[] byteArray, bool setDefaultMessageId = false)
+            : this(new MemoryStream(byteArray), setDefaultMessageId)
         {
-            // reset the owning of the steams
+            // reset the owning of the streams
```
drwill-ms, line 70: stream singular?

iothub/device/src/Message.cs
drwill-ms: yikes

iothub/device/tests/MessageTests.cs
```diff
@@ -169,66 +169,118 @@ public void SettingMessageAsSecurityMessageTest()
...
-                        clonedContent = reader.ReadToEnd();
-                    }
-                
-                    Assert.AreEqual("Cloned version", clonedContent);
-                }
-            }
+                ContentEncoding = contentEncoding,
+                ContentType = contentType,
+                UserId = userId,
+            };
```
drwill-ms, line 186: FWIW, I think you can still do this inline as:

```csharp
using var originalMessage = new Message(...)
{
    // ...
    Properties =
    {
        { propName1, propValue1 },
        { propName2, propValue2 ),
    }
};
```

iothub/device/tests/MessageTests.cs
```diff
@@ -169,66 +169,118 @@ public void SettingMessageAsSecurityMessageTest()
...
+            originalMessage.Properties[propName2] = propValue2;
+
+            // act
+            string clonedMessageContent = "Cloned version";
+            using var clonedMessage = originalMessage.CloneWithBody(Encoding.UTF8.GetBytes(clonedMessageContent));
+
+            // assert
+            clonedMessage.Properties[propName1].Should().Be(propValue1, "Cloned message should have the original message's properties.");
+            clonedMessage.Properties[propName2].Should().Be(propValue2, "Cloned message should have the original message's properties.");
+            clonedMessage.Properties.Count.Should().Be(2);
```
drwill-ms, line 197: should validate count before dereferencing the dictionary

iothub/device/src/Message.cs
abhipsaMisra: I know :(

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -123,10 +128,7 @@ public async override Task SendAsync(string deviceId, Message message, TimeSpan?
                     .SendMessageAsync(amqpMessage, IotHubConnection.GetNextDeliveryTag(ref _sendingDeliveryTag), AmqpConstants.NullBinary, timeout.Value)
                     .ConfigureAwait(false);
 
-                if (Logging.IsEnabled)
-                {
-                    Logging.Info(this, $"Outcome was: {outcome?.DescriptorName}", nameof(SendAsync));
-                }
+                Logging.Info(this, $"Outcome was: {outcome?.DescriptorName}", nameof(SendAsync));
```
drwill-ms, line 131: surround with square brackets in case it is null

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -284,10 +319,7 @@ public override async Task SendAsync(string deviceId, string moduleId, Message m
                         OperationTimeout)
                     .ConfigureAwait(false);
 
-                if (Logging.IsEnabled)
-                {
-                    Logging.Info(this, $"Outcome was: {outcome?.DescriptorName}", nameof(SendAsync));
-                }
+                Logging.Info(this, $"Outcome was: {outcome?.DescriptorName}", nameof(SendAsync));
```
drwill-ms, line 322: square brackets in case of null

iothub/device/src/Message.cs
drwill-ms: I envisioned it being a client options value, not one passed to each Message instance. Thoughts?

drwill-ms (commented):
Waiting for comments review

iothub/device/tests/MessageTests.cs
abhipsaMisra: Dereferencing or referencing? It doesn't really matter though, right?

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: Actually null is already logged with braces, so I left it as is: https://github.com/Azure/azure-iot-sdk-csharp/blob/azabbasi/AmqpServiceClientLogs/common/src/Logging.Common.cs#L44

iothub/device/src/Message.cs
drwill-ms: Yeah, I don't love that.

iothub/device/tests/MessageTests.cs
drwill-ms: dereferencing, meaning we have a a pointer (the dictionary) and we index into it

iothub/service/src/AmqpServiceClient.cs
drwill-ms: interesting

iothub/device/src/Message.cs
abhipsaMisra: The way I see it, client options are used to configure the client instance itself (proxy, model Id, http client etc,), whereas this is a configuration for a message. It isn't really related to a client per se.
Also, I might want some set of messages to have a library-set-guid for message Id, while others to have messageId as null. Taking this configuration at a message level will allow to user the option to configure it like that. (anything we set at client options level will apply to all messages).

iothub/device/tests/MessageTests.cs
abhipsaMisra: got it

iothub/device/src/Message.cs
abhipsaMisra: I've used named args wherever `internal Message(Stream stream, bool ownStream, bool setDefaultMessageId = false)` is currently used, but I don't see a way around this.

iothub/device/src/Message.cs
drwill-ms: I see why you might see it that way.

The downsides are that:
- As a user, I have to pass true to each and every message. That is tedious.
- It makes the Message ctor signature messier.

I see it as, I tell the client I want it to behave a certain way when it does something, that could even extend to putting a message Id on each message, if not otherwise applied.

iothub/device/src/Message.cs
abhipsaMisra: Updated to use an enum

abhipsaMisra:
Closing this PR, this issue is now handled in #1630 (service client logging) and #1631 (Message.MessageId default behavior)

PR #1629, opened 2020-10-28
fix(iot-service): Fix Message disposal behavior

abhipsaMisra:
Our `Message` class on the service client contained a reference to the actual `AmqpMessage` which is sent when `SentAsync()` was called. Since the `AmqpMessage` is disposed after each Send attempt, resending the same IoT Hub message (on network failure etc) would result in us attempting to resend the same disposed Amqp message, resulting in an `ObjectDisposedException`. 

This PR removes the AmqpMessage from being a property of IoT Hub message, resulting in us being able to manage their lifetimes independently.

Is is very surprising however that no one has reported this issue yet. With our current code, if you attempt to resend the same message instance multiple times, you will always get an `ObjectDisposedException`.

You can use the sample [here](https://github.com/Azure-Samples/azure-iot-samples-csharp/pull/155) to test this behavior (simulate network disconnect to have the sdk resend the same Message multiple times).

Even this fails:
```csharp
private async Task SendSameMessage()
{
    var message = new Message(Encoding.UTF8.GetBytes("test"));

    for (int i = 0; i < 5; i++)
    {
        await _serviceClient.SendAsync(_deviceId, message);
        Console.WriteLine($"done {i}");
    }
}
```

### DIFF
--- a/e2e/test/Helpers/templates/FaultInjection.cs
+++ b/e2e/test/Helpers/templates/FaultInjection.cs
@@ -67,9 +67,9 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
 
         public static bool FaultShouldDisconnect(string faultType)
         {
-            return (faultType != FaultType_Auth) &&
-                (faultType != FaultType_Throttle) &&
-                (faultType != FaultType_QuotaExceeded);
+            return faultType != FaultType_Auth
+                && faultType != FaultType_Throttle
+                && faultType != FaultType_QuotaExceeded;
         }
 
         // Fault timings:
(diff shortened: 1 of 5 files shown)


abhipsaMisra:
We cannot combine the disposal of `AmqpMessage` with disposal of IoT Hub `Message` either, since an `AmqpMessage` sent (or attempted) over the link once cannot be resent.
So we need to dispose the `AmqpMessage` after every send attempt.

```
[Microsoft-Azure-Devices-Service-Client-ErrorMessage] (AmqpServiceClient#32106157, SendAsync, SendAsync threw an exception: System.InvalidOperationException: A message cannot be sent because it is either received from a link or is already sent over a link.
   at Microsoft.Azure.Amqp.SendingAmqpLink.BeginSendMessage(AmqpMessage message, ArraySegment`1 deliveryTag, ArraySegment`1 txnId, TimeSpan timeout, AsyncCallback callback, Object state)
   at Microsoft.Azure.Amqp.SendingAmqpLink.<>c__DisplayClass11_0.<SendMessageAsync>b__0(AsyncCallback c, Object s)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl(Func`3 beginMethod, Func`2 endFunction, Action`1 endAction, Object state, TaskCreationOptions creationOptions)
   at Microsoft.Azure.Amqp.SendingAmqpLink.SendMessageAsync(AmqpMessage message, ArraySegment`1 deliveryTag, ArraySegment`1 txnId, TimeSpan timeout)
   at Microsoft.Azure.Devices.AmqpServiceClient.SendAsync(String deviceId, Message message, Nullable`1 timeout) in C:\MyData\SourceCode\csharp\azure\iothub\service\src\AmqpServiceClient.cs:line 135).
```

drwill-ms (approved):

PR #1630, opened 2020-10-28
Add additional logging to service client

abhipsaMisra:
From #1625 
Our log statements do not require a check for IsEnabled since that is already checked in the first implementation layer of the Logging class.

This is PR is just a sample on how the rest of the service client is going to look like when we are done with log instrumentations.

### DIFF
--- a/iothub/service/src/AmqpServiceClient.cs
+++ b/iothub/service/src/AmqpServiceClient.cs
@@ -81,26 +81,31 @@ namespace Microsoft.Azure.Devices
         // This call is executed over AMQP.
         public override async Task OpenAsync()
         {
+            Logging.Enter(this, $"Opening AmqpServiceClient", nameof(OpenAsync));
+
             await _faultTolerantSendingLink.OpenAsync(OpenTimeout).ConfigureAwait(false);
             await _feedbackReceiver.OpenAsync().ConfigureAwait(false);
+
+            Logging.Exit(this, $"Opening AmqpServiceClient", nameof(OpenAsync));
         }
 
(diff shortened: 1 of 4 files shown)


azabbasi (approved):

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -159,12 +155,26 @@ public override Task<PurgeMessageQueueResult> PurgeMessageQueueAsync(string devi
...
+                {
+                    { HttpStatusCode.NotFound, responseMessage => Task.FromResult((Exception)new DeviceNotFoundException(deviceId)) }
+                };
 
-            return _httpClientHelper.DeleteAsync<PurgeMessageQueueResult>(GetPurgeMessageQueueAsyncUri(deviceId), errorMappingOverrides, null, cancellationToken);
+                return _httpClientHelper.DeleteAsync<PurgeMessageQueueResult>(GetPurgeMessageQueueAsyncUri(deviceId), errorMappingOverrides, null, cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                Logging.Error(this, $"PurgeMessageQueueAsync threw an exception: {ex}", nameof(PurgeMessageQueueAsync));
```
barustum, line 171: nit: maybe change PurgeMessageQueueAsync to {nameof(PurgeMessageQueueAsync)}

barustum (approved):

iothub/service/src/AmqpServiceClient.cs
abhipsaMisra: sure!

PR #1631, opened 2020-10-28
Update MessageId default behavior

abhipsaMisra:
Enum for MessageId default behaviour:
```csharp
/// <summary>
/// The different configurations available for setting a default value of MessageId on an IoT Hub Message.
/// </summary>
public enum MessageIdDefault
{
    /// <summary>
    /// MessageId is settable only by the user.
    /// </summary>
    Null,

    /// <summary>
    /// If MessageId is not set by the user, the client library will set to a random GUID before sending the message.
    /// </summary>
    RandomGuid
}
```

DeviceClient - Set it using `ClientOptions.MessageIdDefault`.

ServiceClient - Set using `ServiceClientOptions.MessageIdDefault`.
[Note: `ServiceClientOptions` is newly added on `Microsoft.Azure.Devices` lib]


e2e/test/iothub/service/ServiceClientE2ETests.cs
```diff
@@ -63,5 +64,71 @@ private async Task TestTimeout(TimeSpan? timeout)
                 Logger.Trace($"Testing ServiceClient SendAsync(): exiting test after time={sw.Elapsed}; ticks={sw.ElapsedTicks}");
             }
         }
+
+        [LoggedTestMethod]
+        public async Task MessageIdDefaultNotSet_SendEventDoesNotSetMessageId()
```
abhipsaMisra, line 69: Unfortunately the way `AmqpServiceClient` is implemented, it makes mocking the required amqp types difficult (the amqp types are private members of the class, and cannot be set from any public/ internal API). This is making mocking the amqp behavior difficult.
To get by that, I am testing this in the E2E flow, even though this is a unit test scenario.

iothub/device/src/ClientOptions.cs
```diff
@@ -21,5 +23,11 @@ public class ClientOptions
         /// in the transport settings set for the non-file upload operations.
         /// </summary>
         public Http1TransportSettings FileUploadTransportSettings { get; set; } = new Http1TransportSettings();
+
+        /// <summary>
+        /// The default configuration of MessageId for every message sent by the device or module client instance.
+        /// The default behavior is that MessageId is settable only by the user, by calling <see cref="Message.MessageId"/>.
+        /// </summary>
+        public MessageIdDefault MessageIdDefault { get; set; } = MessageIdDefault.Null;
```
abhipsaMisra, line 31: Suggestions for renaming `MessageIdDefault`?

iothub/device/src/ClientOptions.cs
```diff
@@ -21,5 +23,11 @@ public class ClientOptions
         /// in the transport settings set for the non-file upload operations.
         /// </summary>
         public Http1TransportSettings FileUploadTransportSettings { get; set; } = new Http1TransportSettings();
+
+        /// <summary>
+        /// The default configuration of MessageId for every message sent by the device or module client instance.
+        /// The default behavior is that MessageId is settable only by the user, by calling <see cref="Message.MessageId"/>.
```
drwill-ms, line 29: "by setting" not "by calling", as it isn't a method.

iothub/device/src/ClientOptions.cs
drwill-ms:
```csharp
public enum MessageIdDefaultValue
{
    Unset,
    AssignedRandomGuidBySdk,
};
```

iothub/device/src/Message.cs
```diff
@@ -10,14 +10,20 @@
 
 namespace Microsoft.Azure.Devices.Client
 {
+    internal enum StreamDisposalOwnership
```
drwill-ms, line 13: Might use the word "lifetime" instead of disposal, or the word "responsibility" instead of ownership.

So `StreamLifetimeOwnership` or `StreamDisposalResponsibility`.

iothub/device/src/Message.cs
```diff
@@ -10,14 +10,20 @@
...
+    {
+        User,
```
drwill-ms, line 15: I was thinking `App` and `Sdk`, maybe. You think of it as the user, but the actor is really the device application.

shared/src/MessageIdDefault.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// The different configurations available for setting a default value of MessageId on an IoT Hub Message.
+    /// </summary>
+    public enum MessageIdDefault
+    {
+        /// <summary>
+        /// MessageId is settable only by the user.
```
drwill-ms, line 12: set, not settable. It isn't that the SDK can't, it just won't.

shared/src/MessageIdDefault.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// </summary>
+        Null,
+
+        /// <summary>
+        /// If MessageId is not set by the user, the client library will set to a random GUID before sending the message.
```
drwill-ms, line 17: This is really a hybrid solution then. We only set it when null.

Rather than looking at this enum as what will be set by the SDK (null isn't true), think of it as the behavior the SDK will have. Do we set it? When? To what?

drwill-ms (approved):
Minor naming ideas to consider.

iothub/device/src/Message.cs
abhipsaMisra: That makes sense, I'll rename this to `StreamDisposalResponsibility` with values `App` and `Sdk`.

shared/src/MessageIdDefault.cs
abhipsaMisra: That's true

shared/src/MessageIdDefault.cs
abhipsaMisra: Yes, we don't want to override the value if the user already sets it to something. Only for the case when it isn't set by the user, the sdk will set it to a guid, right before sending the message.

abhipsaMisra: I didn't really get your comment @drwill-ms ; was it a question, or did you have a concern with the behavior?

drwill-ms: I agree with the comment, but don't think the enum names match it. It requires a comment to get across what the behavior will be. So I'm suggesting changing the enum value name from **what** the property will be set to to **when** and **what** it'll be set to.

iothub/service/src/ServiceClientOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Options that allow configuration of the service client instance during initialization.
+    /// </summary>
+    public class ServiceClientOptions
+    {
+        /// <summary>
+        /// The default configuration of MessageId for every message sent by the device or module client instance.
```
drwill-ms, line 14: I'm quibbling with the word default here. At first I thought we meant the default value on this options, but of course it can be changed so it is no longer default.

Then I realized you meant default value on `Message`, or at least I think that, so I think it should be clearly stated.

I'd suggest: `The default value of <see cref="Message.MessageId"/> for every message sent by the device or module client instance.`

shared/src/MessageIdDefault.cs
drwill-ms: You've updated it and I don't have a concern here anymore.

iothub/service/src/ServiceClientOptions.cs
abhipsaMisra: I see your point, but this isn't really the default value of the MessageId, but more of the default "configuration" - as it determines if the library will set a value for the messageId or not.

It is essentially this:
```csharp
public enum ShouldSdkAssignMessagaIdIfNotSet
{
    Yes,
    No
}
```

abhipsaMisra: This is what we want to get represented by the enum (nd matching description). Suggestions?

drwill-ms: My only issue here is on the doc comment

abhipsaMisra: Resolved:
```csharp
public enum SdkAssignsMessageId
{
    /// <summary>
    /// MessageId is set only by the user.
    /// </summary>
    Never,

    /// <summary>
    /// If MessageId is not set by the user, the client library will set it to a random GUID before sending the message.
    /// </summary>
    WhenUnset
}
```

PR #1633, opened 2020-10-29
Add an API to set a callback for receiving C2D messages (amqp)

abhipsaMisra:
Currently we can receive C2D messages on a device client only by calling `ReceiveAsync()` and waiting on the Task.
This PR adds another API using which you can set a callback to be triggered whenever the device receives C2D  messages.

The implementation for MQTT has been added already, this PR adds the implementation for AMQP.

Issue: #1289 

```csharp
public Task SetReceiveMessageHandlerAsync(ReceiveMessageCallback messageHandler, object userContext, CancellationToken cancellationToken = default(CancellationToken));
public delegate Task ReceiveMessageCallback(Message message, object userContext);
```

Behavior over different transport protocols:
- Amqp - 
     - enable callback - open the amqp message receiving link `/devices/<deviceID>/messages/devicebound link`.
    - disable callback - closethe amqp message receiving link `/devices/<deviceID>/messages/devicebound link`.
    - once callback is enabled, `public Task<Message> ReceiveAsync();` will not work anymore 
        - it will always return null.
        - we will log the following statement:
```csharp
Logging.Error(this, "Callback handler set for receiving c2d messages, ReceiveAsync() will now always return null", nameof(ReceiveAsync));
```
   - returning `null` instead of throwing an exception will make writing application code easier for our users, else they will need to keep track of their transport protocol used (http only supports `ReceiveAsync()`), and also the latest state (whether they are polling, or have subscribed to a callback). The drawback is that they will need to maintain the processing code for received messages in 2 places.

### DIFF
--- a/e2e/test/iothub/messaging/MessageReceiveE2EPoolAmqpTests.cs
+++ b/e2e/test/iothub/messaging/MessageReceiveE2EPoolAmqpTests.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
+using FluentAssertions;
 using Microsoft.Azure.Devices.Client;
 using Microsoft.Azure.Devices.E2ETests.Helpers;
 using Microsoft.Azure.Devices.E2ETests.Helpers.Templates;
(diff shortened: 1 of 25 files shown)


iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -18,20 +18,24 @@ internal class AmqpUnit : IDisposable
...
 
-        private readonly SemaphoreSlim _sessionLock = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim _sessionSemaphore = new SemaphoreSlim(1, 1);
 
         private AmqpIoTSendingLink _messageSendingLink;
         private AmqpIoTReceivingLink _messageReceivingLink;
         private readonly SemaphoreSlim _messageReceivingLinkSemaphore = new SemaphoreSlim(1, 1);
 
+        private readonly SemaphoreSlim _messageReceivingCallbackSemaphore = new SemaphoreSlim(1, 1);
+        private bool _deviceReceiveMessageCallbackSet = false;
```
vinagesh, line 37: rename to isDeviceReceiveMessageCallbackSet

abhipsaMisra: sure!

e2e/test/iothub/messaging/MessageReceiveE2EPoolAmqpTests.cs
```diff
@@ -102,6 +104,98 @@ public async Task Message_IoTHubSak_DeviceReceiveSingleMessage_MultipleConnectio
                 ConnectionStringAuthScope.IoTHub).ConfigureAwait(false);
         }
 
+        [LoggedTestMethod]
+        public async Task Message_DeviceSak_DeviceReceiveSingleMessageUsingCallback_MultipleConnections_Amqp()
+        {
+            await
```
drwill-ms, line 110: Why is await on its own line?

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -339,20 +295,109 @@ public async Task<Message> ReceiveMessageAsync(TimeSpan timeout)
...
+            Logging.Enter(this, timeout, $"{nameof(EnableReceiveMessageAsync)}");
+
+            try
+            {
+                // Wait to grab the semaphore, and then open the telemetry receiving link and set the callback,
+                // and set _isDeviceReceiveMessageCallbackSet  to true.
+                // Once _isDeviceReceiveMessageCallbackSet  is set to true, all received c2d messages will be returned on the callback,
+                // and not via the polling ReceiveAsync() call.
+                bool gain = await _messageReceivingCallbackSemaphore.WaitAsync(timeout).ConfigureAwait(false);
+                if (!gain)
```
drwill-ms, line 318: I don't know what gain means. gainedSemaphore? receivedSemaphore?

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -339,20 +295,109 @@ public async Task<Message> ReceiveMessageAsync(TimeSpan timeout)
...
+                {
+                    throw new TimeoutException();
```
drwill-ms, line 320: Perhaps a meaningful exception message here.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -364,10 +409,7 @@ public async Task<AmqpIoTOutcome> DisposeMessageAsync(string lockToken, AmqpIoTD
                 await EnableEventReceiveAsync(timeout).ConfigureAwait(false);
                 disposeOutcome = await _eventReceivingLink.DisposeMessageAsync(lockToken, AmqpIoTResultAdapter.GetResult(disposeAction), timeout).ConfigureAwait(false);
             }
-            if (Logging.IsEnabled)
-            {
-                Logging.Exit(this, lockToken, $"{nameof(DisposeMessageAsync)}");
-            }
+            Logging.Exit(this, lockToken, $"{nameof(DisposeMessageAsync)}");
```
drwill-ms, line 412: Out of curiosity, why not just `nameof(DisposeMessageAsync)`? Why wrap it in a $""?

drwill-ms (approved):

e2e/test/iothub/messaging/MessageReceiveE2EPoolAmqpTests.cs
abhipsaMisra: this was the only way I could get the tabs to line up and look ok. Is it too weird?
Otherwise it ends up like this:
```csharp
await ReceiveMessageUsingCallbackPoolOverAmqpAsync(
	Client.TransportType.Amqp_Tcp_Only,
	PoolingOverAmqp.MultipleConnections_PoolSize,
	PoolingOverAmqp.MultipleConnections_DevicesCount)
	.ConfigureAwait(false);
```

I don't like how `.ConfigureAwait()` is not tabbed properly.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
abhipsaMisra: yeah, gained semaphore. I can rename this.

abhipsaMisra: `enteredSemaphore`

iothub/device/src/Transport/Amqp/AmqpUnit.cs
abhipsaMisra: No reason at all, `nameof()` will give the result as a string, so it will the same. 
My guess, this might have been copied over from some other interpolated string log statement and then modified.

abhipsaMisra: I'll remove the occurrences that I find

e2e/test/iothub/messaging/MessageReceiveE2EPoolAmqpTests.cs
drwill-ms: I'd have done:

```csharp
    await ReceiveMessageUsingCallbackPoolOverAmqpAsync(
        Client.TransportType.Amqp_Tcp_Only,
        PoolingOverAmqp.MultipleConnections_PoolSize,
        PoolingOverAmqp.MultipleConnections_DevicesCount)
    .ConfigureAwait(false);
```

abhipsaMisra: I can update it in subsequent PRs, sure!

PR #1641, opened 2020-11-03
Cleanup device streaming API surface

drwill-ms:
Small updates to property names like `Url` to `Uri`, code style, and collapsing overloads where possible.

### DIFF
--- a/e2e/test/DeviceStreamingE2ETests.cs
+++ b/e2e/test/DeviceStreamingE2ETests.cs
@@ -5,11 +5,8 @@ using Microsoft.Azure.Devices.Client;
 using Microsoft.Azure.Devices.Client.Exceptions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.Diagnostics.Tracing;
 using System.Linq;
-using System.Net;
 using System.Net.WebSockets;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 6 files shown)


e2e/test/DeviceStreamingE2ETests.cs
```diff
@@ -289,7 +289,7 @@ private async Task TestDeviceStreamingAsync(TestDeviceType type, ITransportSetti
 
                 Assert.IsNotNull(clientRequest, "Received an unexpected null device streaming request");
 
-                _log.WriteLine("Device streaming request received (name=" + clientRequest.Name + "; url=" + clientRequest.Url + "; authToken=" + clientRequest.AuthorizationToken + ")");
+                _log.WriteLine("Device streaming request received (name=" + clientRequest.Name + "; url=" + clientRequest.Uri + "; authToken=" + clientRequest.AuthorizationToken + ")");
```
vinagesh, line 292: string interpolation?

e2e/test/DeviceStreamingE2ETests.cs
```diff
@@ -342,7 +342,7 @@ private async Task TestModuleStreamingAsync(TestDeviceType type, ITransportSetti
...
```
vinagesh, line 345: string interpolation?

vinagesh (approved):

e2e/test/DeviceStreamingE2ETests.cs
drwill-ms: I agree, but I'm trying to limit how much I clean up in this branch, rather than doing it in master.

This change was just from a rename, so I prefer to leave the rest of the file alone.

drwill-ms: On second thought, this is a device-streaming specific file, so I'll clean it up.

iothub/device/src/DeviceStreamResponse.cs
```diff
@@ -1,37 +1,30 @@
...
 
-
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// The data structure represent the stream response that is used for interacting with IotHub.
+    /// The data structure to represent the stream response that is used for interacting with Iot Hub.
     /// </summary>
-    public sealed class DeviceStreamResponse
+    internal sealed class DeviceStreamResponse
```
abhipsaMisra, line 9: Is this class of any value to the user?

iothub/device/src/DeviceClient.cs
```diff
@@ -540,87 +539,64 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
         /// <summary>
-        /// Waits for an incoming Cloud-to-Device Stream request.
+        /// Waits for an incoming device stream request.
         /// </summary>
         /// <param name="cancellationToken">Token used for cancelling this operation.</param>
-        /// <returns>A stream request when received</returns>
-        public Task<DeviceStreamRequest> WaitForDeviceStreamRequestAsync(CancellationToken cancellationToken)
-            => this.internalClient.WaitForDeviceStreamRequestAsync(cancellationToken);
+        /// <returns>A stream request when received.</returns>
+        public Task<DeviceStreamRequest> WaitForDeviceStreamRequestAsync(CancellationToken cancellationToken = default)
```
abhipsaMisra, line 580: this is a good change for all of our current APIs!

abhipsaMisra (approved):
Is the `DeviceStreaming` branch up-to-date? 
@timtay-microsoft might have some context to add here, he added some fixes for device streaming APIs recently.

iothub/device/src/DeviceClient.cs
drwill-ms: I'd agree. We can do that during our tech debt work, if you like.

drwill-ms:
> Is the `DeviceStreaming` branch up-to-date?
> @timtay-microsoft might have some context to add here, he added some fixes for device streaming APIs recently.

I'm be worried if it wasn't. Isn't it supposed to be the only true source of our device streaming code?

iothub/device/src/DeviceStreamResponse.cs
drwill-ms: It isn't exposed in our convenience layer, so I figured no.

abhipsaMisra:
> > Is the `DeviceStreaming` branch up-to-date?
> > @timtay-microsoft might have some context to add here, he added some fixes for device streaming APIs recently.
> 
> I'm be worried if it wasn't. Isn't it supposed to be the only true source of our device streaming code?

That was the plan, yes; but I am not aware if we covered that during the handoff that happened after device streaming went preview (after Ewerton handed off this code to us).

PR #1644, opened 2020-11-04
Fix issue where setting handler for receiving messages wouldn't work if telemetry receiving link was already open (amqp)

abhipsaMisra:
There is a bug in the current code where if you try to set a callback for receiving c2d messages after the telemetry link has been already opened, the callback was not getting registered.

This PR also adds some E2E tests for ensuring that the device client always honors the latest handler set for the receive c2d callback APIs.

### DIFF
--- a/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
+++ b/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
@@ -207,25 +207,25 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
         [LoggedTestMethod]
         public async Task DeviceReceiveMessageUsingCallbackAndUnsubscribe_Mqtt()
         {
-            await ReceiveSingleMessageUsingCallbackAndUnsubscribeAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_Tcp_Only);
+            await ReceiveMessageUsingCallbackAndUnsubscribeAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_Tcp_Only);
         }
 
         [LoggedTestMethod]
         public async Task DeviceReceiveMessageUsingCallbackAndUnsubscribe_MqttWs()
         {
-            await ReceiveSingleMessageUsingCallbackAndUnsubscribeAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_WebSocket_Only);
+            await ReceiveMessageUsingCallbackAndUnsubscribeAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_WebSocket_Only);
         }
 
         [LoggedTestMethod]
         public async Task X509_DeviceReceiveMessageUsingCallbackAndUnsubscribe_Mqtt()
         {
-            await ReceiveSingleMessageUsingCallbackAndUnsubscribeAsync(TestDeviceType.X509, Client.TransportType.Mqtt_Tcp_Only);
+            await ReceiveMessageUsingCallbackAndUnsubscribeAsync(TestDeviceType.X509, Client.TransportType.Mqtt_Tcp_Only);
         }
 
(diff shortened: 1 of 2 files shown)


e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
```diff
@@ -606,21 +667,84 @@ private async Task ReceiveSingleMessageUsingCallbackAndUnsubscribeAsync(TestDevi
...
+            using Client.Message receivedThirdMessage = await deviceClient.ReceiveAsync(s_tenSeconds).ConfigureAwait(false);
+            receivedThirdMessage.MessageId.Should().Be(thirdMessage.MessageId);
             receiveMessageOverCallback.Should().Throw<OperationCanceledException>();
+            await deviceClient.CompleteAsync(receivedThirdMessage).ConfigureAwait(false);
 
             firstMessage.Dispose();
             secondMessage.Dispose();
+            thirdMessage.Dispose();
+
+            await deviceClient.CloseAsync().ConfigureAwait(false);
```
drwill-ms, line 688: Should we be disposing, not closing?

e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
```diff
@@ -606,21 +667,84 @@ private async Task ReceiveSingleMessageUsingCallbackAndUnsubscribeAsync(TestDevi
...
+                    firstHandlerSemaphore.Release();
+                },
+                deviceClient);
+
+            // The C2D message should be received over the first callback handler, releasing the corresponding semaphore.
+            using var firstCts = new CancellationTokenSource(s_oneMinute);
+            (Message firstMessage, _, _) = ComposeC2dTestMessage(Logger);
+            Logger.Trace($"Sending C2D message from service, messageId={firstMessage.MessageId}");
+            await Task
+                    .WhenAll(
```
drwill-ms, line 716: This block is tabbed in 1 tab too many

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -231,12 +231,12 @@ private async Task EnsureMessageReceivingLinkAsync(TimeSpan timeout, bool enable
                     {
                         amqpIoTSession.SafeClose();
                     };
-                    if (enableCallback)
-                    {
-                        _messageReceivingLink.RegisterReceiveMessageListener(OnDeviceMessageReceived);
-                    }
                     Logging.Associate(this, this, _messageReceivingLink, nameof(EnsureMessageReceivingLinkAsync));
                 }
+                if (enableCallback)
```
drwill-ms, line 236: newline ahead of this new block, to make it visually distinct from the previous one (unlike try/catch/finally/if/else if/else)

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -231,12 +231,12 @@ private async Task EnsureMessageReceivingLinkAsync(TimeSpan timeout, bool enable
                     {
```
drwill-ms, line 231: Seems like this method name should be `EnsureMessageReceivingLinkIsOpenAsync`

drwill-ms (approved):

e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
abhipsaMisra: Since we are not reusing this client again, technically both should be fine. The device client will anyway be disposed once it is out of scope.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
abhipsaMisra: This is fair, yes; I'll rename it.

PR #1653, opened 2020-11-10
New, clean device streaming port

drwill-ms:
I did a little code cleanup as I went, but for the most part this is the implementation that is in the preview branch (looking at diffs on master). I don't want to put too much effort into it, so I'll probably skip nits.

### DIFF
--- a/common/src/device/provisioning/transport/ClientApiVersionHelper.cs
+++ b/common/src/device/provisioning/transport/ClientApiVersionHelper.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
     {
         // TODO: Split ApiVersionName as it is only used by HTTP and AMQP.
         public const string ApiVersionName = "api-version";
+
         public const string ApiVersion = "2019-03-31";
     }
 }
--- a/e2e/test/iothub/DeviceStreamingE2ETests.cs
+++ b/e2e/test/iothub/DeviceStreamingE2ETests.cs
@@ -1,0 +1,392 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Linq;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Client.Exceptions;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using ClientDeviceStreamingRequest = Microsoft.Azure.Devices.Client.DeviceStreamRequest;
+using ServiceDeviceStreamingRequest = Microsoft.Azure.Devices.DeviceStreamRequest;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    [TestCategory("DeviceStreaming")]
+    public partial class DeviceStreamingTests : E2EMsTestBase
+    {
+        private static readonly string _devicePrefix = $"{nameof(DeviceStreamingTests)}_";
+        private static readonly string _modulePrefix = $"{nameof(DeviceStreamingTests)}_";
+        private static readonly string s_proxyServerAddress = Configuration.IoTHub.ProxyServerAddress;
+
+        [TestMethod]
+        public async Task DeviceStreaming_RequestAccepted_Sas_Amqp()
+        {
+            var amqpTransportSettings = new AmqpTransportSettings(Client.TransportType.Amqp_Tcp_Only);
+            var transportSettings = new ITransportSettings[] { amqpTransportSettings };
+
+            await TestDeviceStreamingAsync(TestDeviceType.Sasl, transportSettings, true).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task DeviceStreaming_RequestAccepted_Sas_AmqpWs()
+        {
+            var amqpTransportSettings = new AmqpTransportSettings(Client.TransportType.Amqp_WebSocket_Only);
+            var transportSettings = new ITransportSettings[] { amqpTransportSettings };
+
+            await TestDeviceStreamingAsync(TestDeviceType.Sasl, transportSettings, true).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task DeviceStreaming_RequestAccepted_Sas_Mqtt()
+        {
+            var mqttTransportSettings = new Client.Transport.Mqtt.MqttTransportSettings(Client.TransportType.Mqtt_Tcp_Only);
+            var transportSettings = new ITransportSettings[] { mqttTransportSettings };
+
+            await TestDeviceStreamingAsync(TestDeviceType.Sasl, transportSettings, true).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task DeviceStreaming_RequestAccepted_Sas_MqttWs()
+        {
+            var mqttTransportSettings = new Client.Transport.Mqtt.MqttTransportSettings(Client.TransportType.Mqtt_WebSocket_Only);
+            var transportSettings = new ITransportSettings[] { mqttTransportSettings };
+
+            await TestDeviceStreamingAsync(TestDeviceType.Sasl, transportSettings, true).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task DeviceStreaming_RequestAccepted_x509_Amqp()
+        {
+            var amqpTransportSettings = new AmqpTransportSettings(Client.TransportType.Amqp_Tcp_Only);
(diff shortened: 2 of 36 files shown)


e2e/test/iothub/DeviceStreamingE2ETests.cs
```diff
@@ -0,0 +1,392 @@
...
+using System.Linq;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Client.Exceptions;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using ClientDeviceStreamingRequest = Microsoft.Azure.Devices.Client.DeviceStreamRequest;
```
drwill-ms, line 14: I don't love that we have the same type name in the device and service clients. Sure, normally someone won't be using both together in an application, but there are cases, especially in tests and samples.

I'd propose we name them differently, so they don't conflict. Thoughts on names?

Perhaps on the service it can be `DeviceStreamingRequest/Response` but on the device it'd just be `StreamingRequest/Response`.

common/src/device/provisioning/transport/ClientApiVersionHelper.cs
```diff
@@ -7,6 +7,10 @@ internal class ClientApiVersionHelper
     {
         // TODO: Split ApiVersionName as it is only used by HTTP and AMQP.
         public const string ApiVersionName = "api-version";
+
         public const string ApiVersion = "2019-03-31";
+
+        // The preview branch uses a GA api version because Device Streaming is already in public preview and is available on certain Azure regions.
```
abhipsaMisra, line 13: really nitty - shouldn't the comment be above the var declaration?

iothub/device/src/ModuleClient.cs
```diff
@@ -702,5 +703,42 @@ private static Uri GetModuleMethodUri(string deviceId, string moduleId)
...
 
         #endregion Module Specific API
+
+        #region Device Streaming
+
+        /// <summary>
+        /// Waits for an incoming Cloud-to-Device Stream request.
+        /// </summary>
+        /// <returns>A stream request when received</returns>
+        public Task<DeviceStreamRequest> WaitForDeviceStreamRequestAsync()
```
abhipsaMisra, line 713: this one can be removed, right?

e2e/test/iothub/DeviceStreamingE2ETests.cs
timtay-microsoft: "DeviceClientStreamingRequest" vs "ServiceClientStreamingRequest"
"DeviceToServiceStreamingRequest" vs "ServiceToDeviceStreamingRequest"

These extra distinctions feel a bit clunkier then just using the same names in the device and service client

iothub/device/tests/Amqp/AmqpTransportHandlerTests.cs
```diff
@@ -1,4 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
drwill-ms, line 2: fixed

iothub/device/tests/Mqtt/MqttTransportHandlerTests.cs
```diff
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
drwill-ms, line 2: fixed

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -77,13 +80,15 @@ internal sealed class MqttTransportHandler : TransportHandler, IMqttIotHubEventH
 
         // Topic names for receiving cloud-to-device messages.
         private const string DeviceBoundMessagesTopicFilter = "devices/{0}/messages/devicebound/#";
+
```
abhipsaMisra, line 83: why a new line here? the comment is relevant for both of these variables

drwill-ms: VS does it. It thinks a comment should only apply to the thing below it. The only way to fix that is to put a newline after the comment.

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -812,7 +926,7 @@ private async Task HandleDisconnectAsync()
 
                     await base.OpenAsync(cancellationToken).ConfigureAwait(false);
 
-                    var tasks = new List<Task>(3);
+                    var tasks = new List<Task>(5);
```
abhipsaMisra, line 929: good catch!

iothub/service/src/ServiceClient.cs
```diff
@@ -205,5 +205,24 @@ public static ServiceClient CreateFromConnectionString(string connectionString,
...
+        /// <param name="deviceStreamRequest">Configuration needed for initiating a cloud-to-device stream.</param>
+        /// <param name="cancellationToken">Token used for controlling the termination of the asynchronous call.</param>
+        /// <returns>The result of the cloud-to-device stream request, or null of the request itself could not be completed.</returns>
+        public abstract Task<DeviceStreamResponse> CreateStreamAsync(string deviceId, DeviceStreamRequest deviceStreamRequest, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Initiates a new cloud-to-device stream.
+        /// </summary>
+        /// <param name="deviceId">Device Id</param>
+        /// <param name="moduleId">Module ID</param>
```
timtay-microsoft, line 222: device Id and module Id could have a bit more helpful descriptions here and above. For instance: "The Id of the device to create the stream to"

timtay-microsoft: I figure these are just being copied from the old branch, but we could spruce it up a bit

timtay-microsoft (approved):

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -360,6 +364,86 @@ public override async Task SendAsync(string deviceId, string moduleId, Message m
...
+        /// <summary>
+        /// Initiates a new cloud-to-device stream.
+        /// </summary>
+        /// <param name="deviceId">Device Id</param>
+        /// <param name="deviceStreamRequest">Configuration needed for initiating a cloud-to-device stream.</param>
+        /// <param name="cancellationToken">Token used for controlling the termination of the asynchronous call.</param>
+        /// <returns>The result of the cloud-to-device stream request, or null of the request itself could not be completed.</returns>
+        public override Task<DeviceStreamResponse> CreateStreamAsync(string deviceId, DeviceStreamRequest deviceStreamRequest, CancellationToken cancellationToken = default)
+        {
+            return this.CreateStreamAsync(GetDeviceStreamUri(deviceId, deviceStreamRequest.StreamName), deviceStreamRequest, cancellationToken);
```
abhipsaMisra, line 376: don't need `.this`

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -360,6 +364,86 @@ public override async Task SendAsync(string deviceId, string moduleId, Message m
...
+        {
+            return this.CreateStreamAsync(GetModuleStreamUri(deviceId, moduleId, deviceStreamRequest.StreamName), deviceStreamRequest, cancellationToken);
+        }
+
+        internal async Task<DeviceStreamResponse> CreateStreamAsync(
+            Uri uri,
+            DeviceStreamRequest deviceStreamRequest,
+            CancellationToken cancellationToken)
+        {
+            TimeSpan timeout = GetInitiateStreamOperationTimeout(deviceStreamRequest);
```
abhipsaMisra, line 397: our device streaming code on preview branch probably didn't have logging enabled, we could add it now.

e2e/test/iothub/DeviceStreamingE2ETests.cs
abhipsaMisra: My preference:
1. `DeviceToServiceStreamingRequest` and `ServiceToDeviceStreamingRequest`
2. `Devices.Client.DeviceStreamRequest` and `Devices.DeviceStreamRequest` (or alias -> what we have now)
3. `DeviceClientStreamingRequest` and `ServiceClientStreamingRequest` (I don't particularly like putting the term "client" in there)

Also, a note on the existing names - we call it "device" streaming, but I am seeing that we support it on the module client as well :(

Should we call it `ClientToServiceStreamingRequest`/`Response`?

abhipsaMisra (approved):

iothub/service/src/AmqpServiceClient.cs
drwill-ms: I'm not going to invest in this code that much. I just wanted something that would be up-to-date on master that we could refer to when we need to.

e2e/test/iothub/DeviceStreamingE2ETests.cs
drwill-ms: I'll just leave it as is for now.

PR #1654, opened 2020-11-12
Fix a smattering of project warnings

drwill-ms:


common/src/service/ExceptionTrace.cs
```diff
@@ -22,24 +25,24 @@ public ExceptionTrace(string eventSourceName)
             _eventSourceName = eventSourceName;
         }
 
-        public Exception AsError(Exception exception, EventTraceActivity activity = null)
```
drwill-ms, line 25: Unused parameters

common/src/service/ExceptionTrace.cs
```diff
@@ -95,24 +99,16 @@ public void TraceHandled(Exception exception, string catchLocation, EventTraceAc
                 exception.ToStringSlim()));
 #endif
 
-            ////MessagingClientEtwProvider.Provider.HandledExceptionWithFunctionName(
```
drwill-ms, line 98: Several cases of removing dead code

common/src/service/ExceptionTrace.cs
```diff
@@ -95,24 +99,16 @@ public void TraceHandled(Exception exception, string catchLocation, EventTraceAc
...
-            ////    activity, catchLocation, exception.ToStringSlim(), string.Empty);
-
             BreakOnException(exception);
         }
 
-        public void TraceUnhandled(Exception exception)
-        {
-            ////MessagingClientEtwProvider.Provider.EventWriteUnhandledException(this.eventSourceName + ": " + exception.ToStringSlim());
```
drwill-ms, line 106: Removing empty methods and call chain to them

common/src/service/Fx.cs
```diff
@@ -153,10 +145,13 @@ public static Exception AssertAndFailFastService(string description)
...
                     // Throwing FatalException since our service does not morph/eat up fatal exceptions
                     // We should find the tracking bug in Azure for this issue, and remove the workaround when fixed by Azure
-                    Thread failFastWorkaroundThread = new Thread(delegate ()
-                    {
-                        throw new FatalException(failFastMessage);
-                    });
+                    var failFastWorkaroundThread = new Thread(
+                        delegate ()
+                        {
+#pragma warning disable CA2219 // Do not raise exceptions in finally clauses
```
drwill-ms, line 151: I'm not sure wat the right fix is for this. Inside a finally you could be running code just before an exception gets thrown, so it makes sense.

I don't know what most of the Fx.cs class is for. Much of the code looks highly dubious.

vinagesh (approved):

PR #1656, opened 2020-11-12
Fix CA1001 in ClientWebSocketTransport

drwill-ms:
It seems we're already disposing there, but this ensures that.

### DIFF
--- a/iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
+++ b/iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
@@ -13,17 +13,20 @@ using Microsoft.Azure.Amqp.Transport;
 
 namespace Microsoft.Azure.Devices
 {
-    internal sealed class ClientWebSocketTransport : TransportBase
+    internal sealed class ClientWebSocketTransport : TransportBase, IDisposable
     {
         private static readonly AsyncCallback s_onReadComplete = OnReadComplete;
         private static readonly AsyncCallback s_onWriteComplete = OnWriteComplete;
         private static readonly TimeSpan s_closeTimeout = TimeSpan.FromSeconds(30);
 
-        private readonly ClientWebSocket _webSocket;
         private readonly EndPoint _localEndPoint;
         private readonly EndPoint _remoteEndPoint;
-        private volatile CancellationTokenSource _writeCancellationTokenSource;
-        private bool _disposed;
+
+        // Disposables
+
+        private readonly ClientWebSocket _webSocket;
+        private readonly CancellationTokenSource _writeCancellationTokenSource;
+        private bool _isDisposed;
 
         public ClientWebSocketTransport(ClientWebSocket webSocket, EndPoint localEndpoint, EndPoint remoteEndpoint)
             : base("clientwebsocket")
@@ -36,10 +39,7 @@ namespace Microsoft.Azure.Devices
 
         public override string LocalEndPoint => _localEndPoint.ToString();
 
-        public override string RemoteEndPoint
-        {
-            get { return _remoteEndPoint.ToString(); }
-        }
+        public override string RemoteEndPoint => _remoteEndPoint.ToString();
 
         public override bool RequiresCompleteFrames => true;
 
@@ -58,7 +58,7 @@ namespace Microsoft.Azure.Devices
             Fx.AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
             args.Exception = null; // null out any exceptions
 
-            Task taskResult = WriteAsyncCore(args);
+            Task taskResult = WriteImplAsync(args);
             if (WriteTaskDone(taskResult, args))
             {
                 return false;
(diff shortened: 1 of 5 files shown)


iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -58,7 +58,7 @@ public override bool WriteAsync(TransportAsyncCallbackArgs args)
             Fx.AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
             args.Exception = null; // null out any exceptions
 
-            Task taskResult = WriteAsyncCore(args);
+            Task taskResult = WriteAsyncCoreAsync(args);
```
vinagesh, line 61: Just WriteAsync?

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -68,7 +68,44 @@ public override bool WriteAsync(TransportAsyncCallbackArgs args)
...
+        }
+
+        public void Dispose()
+        {
+            if (!_isDisposed)
+            {
+                if (_webSocket.State != WebSocketState.Aborted)
+                {
+                    _webSocket.Abort();
+                    _webSocket.Dispose();
```
vinagesh, line 99: I don't know the context but should we not be disposing the websocket irrespective of the state when we call dispose? I can see that we only abort if the state is no Aborted.

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
barustum: there is already WriteAsync..maybe WriteCoreAsync

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
barustum: My only guess is that when it's aborted, it's already disposed. I could be wrong :)

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -255,24 +258,22 @@ private static void HandleReadComplete(IAsyncResult result)
...
                 args.Exception = taskResult.Exception;
                 return true;
             }
-            else if (taskResult.IsCompleted)
+
+            if (taskResult.IsCompleted)
             {
                 args.BytesTransfered = taskResult.Result;
                 args.CompletedSynchronously = result.CompletedSynchronously;
                 return true;
```
vinagesh, line 273: Are these returns needed? return taskResult.IsCanceled should be sufficient right?

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -295,25 +296,21 @@ private static void HandleWriteComplete(IAsyncResult result)
...
-                return true;
-            }
-            else if (taskResult.IsCanceled)  // This should not happen since TaskCanceledException is handled in WriteAsyncCore.
-            {
+                args.CompletedSynchronously = ((IAsyncResult)taskResult).CompletedSynchronously;
                 return true;
             }
 
-            return false;
+            return taskResult.IsCanceled;
```
vinagesh, line 313: Are these returns needed - line 310 and line 303? return taskResult.IsCanceled should be sufficient right?

vinagesh (approved):

iothub/service/src/IotHubConnection.cs
```diff
@@ -296,6 +293,12 @@ private async Task<AmqpSession> CreateSessionAsync(TimeSpan timeout)
             }
             finally
             {
+                if (transport != null
+                    && transport is ClientWebSocketTransport clientWebSocketTransport)
+                {
+                    clientWebSocketTransport.Dispose();
```
drwill-ms, line 299: @abhipsaMisra is this safe to do? I'm not entirely sure what this method is doing, but it doesn't seem to save off any of the things it has created here, so it feels like they should only be kept for the duration of the method.

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -68,7 +68,41 @@ public override bool WriteAsync(TransportAsyncCallbackArgs args)
...
+
+            taskResult.ToAsyncResult(s_onReadComplete, args);
+            return true;
+        }
+
+        public void Dispose()
+        {
+            if (!_isDisposed)
+            {
+                _webSocket.Abort();
```
abhipsaMisra, line 96: do you think it'll be an issue if we cann websocket.Abort() on an aborted socket? the doc only says that a websocket cannot be reused once aborted

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -219,18 +233,7 @@ private void CancelPendingWrite()
 
         protected override void AbortInternal()
         {
-            if (!_disposed)
```
abhipsaMisra, line 222: funny that we had all this code, but didn't make the type disposable

iothub/service/src/IotHubConnection.cs
abhipsaMisra: I don't think we should dispose it here, we are giving a reference of this web socket transport to the amqp library, which it will hold on to until the connection is active. I think we should dispose the transport object when we dispose the iot hub connection

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
drwill-ms: yeah

iothub/service/src/IotHubConnection.cs
drwill-ms: Let's do a call later today, because I'm not seeing where the objects are being saved off.

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
drwill-ms: I don't know. I worry about changing it too much, though.

abhipsaMisra (approved):
Looks good

barustum (approved):

PR #1660, opened 2020-11-13
Fix project warnings

azabbasi:
https://msazure.visualstudio.com/One/_workitems/edit/8792109
https://msazure.visualstudio.com/One/_workitems/edit/8792111
https://msazure.visualstudio.com/One/_workitems/edit/8792122

### DIFF
--- a/iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
+++ b/iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
@@ -23,55 +23,55 @@ using System;
 namespace Microsoft.Azure.Devices.Client.TransientFaultHandling
 {
     /// <summary>
-    /// A retry strategy with backoff parameters for calculating the exponential delay between retries.
+    /// A retry strategy with back-off parameters for calculating the exponential delay between retries.
     /// </summary>
     internal class ExponentialBackoff : RetryStrategy
     {
-        private readonly int retryCount;
+        private readonly int _retryCount;
 
-        private readonly TimeSpan minBackoff;
+        private readonly TimeSpan _minBackoff;
 
-        private readonly TimeSpan maxBackoff;
+        private readonly TimeSpan _maxBackoff;
 
-        private readonly TimeSpan deltaBackoff;
+        private readonly TimeSpan _deltaBackoff;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="T:Microsoft.Azure.Devices.Client.TransientFaultHandling.ExponentialBackoff" /> class.
+        /// Initializes a new instance of the <see cref="ExponentialBackoff" /> class.
         /// </summary>
-        public ExponentialBackoff() : this(RetryStrategy.DefaultClientRetryCount, RetryStrategy.DefaultMinBackoff, RetryStrategy.DefaultMaxBackoff, RetryStrategy.DefaultClientBackoff)
+        public ExponentialBackoff() : this(DefaultClientRetryCount, DefaultMinBackoff, DefaultMaxBackoff, DefaultClientBackoff)
         {
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="T:Microsoft.Azure.Devices.Client.TransientFaultHandling.ExponentialBackoff" /> class with the specified retry settings.
+        /// Initializes a new instance of the <see cref="ExponentialBackoff" /> class with the specified retry settings.
         /// </summary>
         /// <param name="retryCount">The maximum number of retry attempts.</param>
-        /// <param name="minBackoff">The minimum backoff time</param>
-        /// <param name="maxBackoff">The maximum backoff time.</param>
+        /// <param name="minBackoff">The minimum back-off time</param>
+        /// <param name="maxBackoff">The maximum back-off time.</param>
         /// <param name="deltaBackoff">The value that will be used to calculate a random delta in the exponential delay between retries.</param>
         public ExponentialBackoff(int retryCount, TimeSpan minBackoff, TimeSpan maxBackoff, TimeSpan deltaBackoff) : this(null, retryCount, minBackoff, maxBackoff, deltaBackoff, RetryStrategy.DefaultFirstFastRetry)
         {
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="T:Microsoft.Azure.Devices.Client.TransientFaultHandling.ExponentialBackoff" /> class with the specified name and retry settings.
+        /// Initializes a new instance of the <see cref="ExponentialBackoff" /> class with the specified name and retry settings.
         /// </summary>
         /// <param name="name">The name of the retry strategy.</param>
         /// <param name="retryCount">The maximum number of retry attempts.</param>
-        /// <param name="minBackoff">The minimum backoff time</param>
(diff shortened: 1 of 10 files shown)


timtay-microsoft (approved):

iothub/service/src/PurgeMessageQueueResult.cs
```diff
@@ -21,6 +21,6 @@ public sealed class PurgeMessageQueueResult
         /// The total number of messages that were purged from the device's queue.
         /// </summary>
         [JsonProperty(PropertyName = "totalMessagesPurged", Required = Required.Always)]
-        public int TotalMessagesPurged;
```
timtay-microsoft, line 24: Is this a breaking change?

azabbasi: I don't believe so, public fields are both gettable and settable

iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
```diff
@@ -23,55 +23,55 @@
...
-    /// A retry strategy with backoff parameters for calculating the exponential delay between retries.
+    /// A retry strategy with back-off parameters for calculating the exponential delay between retries.
     /// </summary>
     internal class ExponentialBackoff : RetryStrategy
     {
-        private readonly int retryCount;
+        private readonly int _retryCount;
 
-        private readonly TimeSpan minBackoff;
+        private readonly TimeSpan _minBackoff;
```
azabbasi, line 32: should we rename these to `_minBackOff` to camelcase the O ??

abhipsaMisra: https://en.wikipedia.org/wiki/Exponential_backoff
all lowercase

iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
```diff
@@ -23,55 +23,55 @@
 namespace Microsoft.Azure.Devices.Client.TransientFaultHandling
 {
     /// <summary>
-    /// A retry strategy with backoff parameters for calculating the exponential delay between retries.
+    /// A retry strategy with back-off parameters for calculating the exponential delay between retries.
```
abhipsaMisra, line 26: hyphenated?

azabbasi: my spell checker doesn't like backoff as a single word ... if we all agree that it is a single word I will change these back.

iothub/service/src/ExportImportDevice.cs
```diff
@@ -11,27 +11,51 @@
...
     public sealed class ExportImportDevice
     {
         private string _eTag;
         private string _twinETag;
 
         /// <summary>
-        /// making default ctor public
+        /// Property container
+        /// </summary>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1034:Nested types should not be visible", Justification = "Publicly property. No behavior changes allowed.")]
```
barustum, line 24: publicly -> public

iothub/service/src/ExportImportDevice.cs
```diff
@@ -11,27 +11,51 @@
...
+        /// <summary>
+        /// Create an ExportImportDevice
         /// </summary>
         public ExportImportDevice()
         {
         }
 
         /// <summary>
-        /// ctor which takes a Device object along with import mode
+        /// Create an ExportImportDevice
```
barustum, line 48: use <see cref="ExportImportDevice" />

iothub/service/src/PurgeMessageQueueResult.cs
```diff
@@ -2,10 +2,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // ---------------------------------------------------------------
 
+using Newtonsoft.Json;
```
barustum, line 5: why did you need to add this using?

abhipsaMisra (approved):

iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
drwill-ms: compound adjectives should be hyphenated

drwill-ms: when used as a verb, compound phrases do not use hyphens

iothub/device/src/TransientFaultHandling/ExponentialBackoff.cs
drwill-ms: Wikipedia is not the authority on grammar and spelling tho

drwill-ms: You could change the non-public variables, but I noticed there are parameters with those names and that would be a breaking change, so you might as well leave it the way it is so it is consistent.

iothub/service/src/ExportImportDevice.cs
drwill-ms, line 24: Why did the class get moved?

iothub/service/src/ExportImportDevice.cs
```diff
@@ -87,32 +111,35 @@ public string ETag
         [JsonProperty(PropertyName = "authentication")]
         public AuthenticationMechanism Authentication { get; set; }
 
+        /// <summary>
+        /// Twin ETag
```
drwill-ms, line 115: These need to be more meaningful. See our related comments in ADT for examples.

drwill-ms (approved):

iothub/service/src/ExportImportDevice.cs
azabbasi: I just moved it to the top of the wrapper class.

iothub/service/src/PurgeMessageQueueResult.cs
azabbasi: I didn't. Moved it from inside the namespace to outside

iothub/service/src/ExportImportDevice.cs
azabbasi: That will be a pointer to the same file .. what's the benefit ?

barustum: True. It's just easier to make sure the comment is not outdated if we rename the class. I also see you used `<see cref="ExportImportDevice" />,` in the ExportImportDevice.cs class, so I thought maybe keep same pattern.

barustum (approved):

iothub/service/src/ExportImportDevice.cs
```diff
@@ -11,27 +11,51 @@
...
+        /// <summary>
+        /// Create an ExportImportDevice <see cref="ExportImportDevice" />
         /// </summary>
         public ExportImportDevice()
         {
         }
 
         /// <summary>
-        /// ctor which takes a Device object along with import mode
+        /// Create an ExportImportDevice <see cref="ExportImportDevice" />
```
barustum, line 48: @azabbasi Wouldn't this show the 'ExportImportDevice' twice? I think it should be `Create an <see cref="ExportImportDevice" />`

drwill-ms: Yes, it will.

PR #1661, opened 2020-11-13
Update XML comments, code cleanup and add missing internal constructors for Exceptions

abhipsaMisra:
CA1032: Implement standard exception constructors
Cause: A type extends System.Exception but doesn't declare all the required constructors.

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1032

Also added XML comments for public methods and properties (CS1591).

### DIFF
--- a/iothub/service/src/Common/Exceptions/DeviceAlreadyExistsException.cs
+++ b/iothub/service/src/Common/Exceptions/DeviceAlreadyExistsException.cs
@@ -6,30 +6,62 @@ using System.Runtime.Serialization;
 
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
+    /// <summary>
+    /// The exception that is thrown when an attempt is made to create a device that already exists in the hub.
+    /// </summary>
     [Serializable]
     public sealed class DeviceAlreadyExistsException : IotHubException
     {
+        /// <summary>
+        /// Creates an instance of <see cref="DeviceAlreadyExistsException"/> with the specified device Id, an empty error message
+        /// and marks it as non-transient.
+        /// </summary>
+        /// <param name="deviceId">The Id of the device that is already registered on IoT Hub.</param>
         public DeviceAlreadyExistsException(string deviceId)
             : this(deviceId, string.Empty)
         {
         }
 
+        /// <summary>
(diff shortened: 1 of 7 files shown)


iothub/service/src/Common/Exceptions/DeviceAlreadyExistsException.cs
```diff
@@ -6,30 +6,62 @@
...
         {
         }
 
-        public DeviceAlreadyExistsException(SerializationInfo info, StreamingContext context)
+        internal DeviceAlreadyExistsException()
+            : base()
+        {
+        }
+
+        private DeviceAlreadyExistsException(SerializationInfo info, StreamingContext context)
```
abhipsaMisra, line 64: Since this exception is a sealed type, the serialization constructor is declared as private: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2229?view=vs-2019#cause

iothub/service/src/Common/Exceptions/DeviceInvalidResultCountException.cs
```diff
@@ -1,37 +1,57 @@
...
+using System.Runtime.Serialization;
+
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
-    using System;
-    using System.Runtime.Serialization;
-
+    /// <summary>
+    /// The exception that is thrown when the count of device results exceeds the specified maximum value.
+    /// Note: This exception is currently not thrown by the client library.
```
abhipsaMisra, line 11: Not sure what the intention/ use-case is behind this exception. It is currently unreferenced in our code.

drwill-ms: I'm pretty sure I removed an exception type before that was unreferenced, but that was before the "breaking changes" rule. You could check with Mark about removing it, though. We've removed something else unreferenced since then too.

drwill-ms (approved):

iothub/service/src/Common/Exceptions/DeviceInvalidResultCountException.cs
abhipsaMisra: As I clean up the rest of the exception types, I'll keep on adding the unreferenced ones to a list and we can discuss their future

PR #1662, opened 2020-11-13
CS4014: No unwanted, unawaited calls

drwill-ms:


iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -189,9 +189,10 @@ protected override bool OpenInternal()
 
         protected override bool CloseInternal()
         {
-            if (_webSocket.State != WebSocketState.Closed && _webSocket.State != WebSocketState.Aborted)
+            if (_webSocket.State != WebSocketState.Closed
+                && _webSocket.State != WebSocketState.Aborted)
             {
-                CloseInternalAsync(s_closeTimeout);
+                CloseInternalAsync(s_closeTimeout).GetAwaiter().GetResult();
```
drwill-ms, line 195: Call signature of `CloseInternal` mandated by base class from AMQP assembly, so we just need to make this synchronous.

azabbasi (approved):

iothub/service/src/Common/Amqp/ClientWebSocketTransport.cs
barustum: I see a similar pattern in this file, where for example the synchronous method WriteAsync would call into the async method WriteImplAsync. In that case though, we used Task.ToAsyncResult to make it synchronous. Can we not use the same pattern here?

drwill-ms: I don't know what `Task.ToAsyncResult` is. I mean, I've looked at it, but I don't get why it is there. It is a custom class in our project that has several code warnings of its own.

I will say that in the track2 library when they want a sync method to call an async method, they use `GetAwaiter().GetResult()` but also tell you to avoid doing it.

barustum (approved):

bikamani (approved):

PR #1664, opened 2020-11-16
Fix 1034 - 1035 - 1037

azabbasi:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1304?view=vs-2019
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1305?view=vs-2019
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1307?view=vs-2019

### DIFF
--- a/iothub/device/src/Transport/HttpClientHelper.cs
+++ b/iothub/device/src/Transport/HttpClientHelper.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
                 }
 
                 handler.ClientCertificates.Add(clientCert);
-                this._usingX509ClientCert = true;
+                _usingX509ClientCert = true;
             }
 
             if (proxy != DefaultWebProxySettings.Instance)
@@ -85,7 +85,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
                 handler.Proxy = proxy;
             }
 
-            this._httpClientObj = handler != null ? new HttpClient(handler) : new HttpClient();
+            _httpClientObj = handler != null ? new HttpClient(handler) : new HttpClient();
 #else
             if (httpClientHandler == null)
             {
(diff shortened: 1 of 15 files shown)


iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -70,7 +71,7 @@ IWebProxy proxy
                 }
 
                 handler.ClientCertificates.Add(clientCert);
-                this.usingX509ClientCert = true;
+                this._usingX509ClientCert = true;
```
drwill-ms, line 74: remove `this.`?

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -180,8 +185,8 @@ internal async Task UploadToBlobAsync(string blobName, Stream source, Cancellati
             try
             {
                 // 2. Use SAS URI to send data to Azure Storage Blob (PUT)
-                CloudBlockBlob blob = new CloudBlockBlob(new Uri(putString));
-                var uploadTask = blob.UploadFromStreamAsync(source);
+                var blob = new CloudBlockBlob(new Uri(putString));
+                Task uploadTask = blob.UploadFromStreamAsync(source);
                 await uploadTask.ConfigureAwait(false);
```
drwill-ms, line 190: Any reason why the former statement is not just awaited, rather than being split up like this?

azabbasi: I actually changed it initially, there is some logic below that I didn't change that relies on the task variable.
I can look into it and see if there is any value in having them around.

drwill-ms (approved):

barustum (approved):

PR #1666, opened 2020-11-17
refactor(iot-device): Add xml comments to exceptions, code cleanup

abhipsaMisra:
Last set of:
CA1032: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1032
CA2229: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2229

### DIFF
--- a/iothub/device/src/Common/Exceptions/IotHubException.cs
+++ b/iothub/device/src/Common/Exceptions/IotHubException.cs
@@ -19,104 +19,93 @@ namespace Microsoft.Azure.Devices.Client.Exceptions
         private const string TrackingIdValueSerializationStoreName = "IotHubException-TrackingId";
 
         /// <summary>
-        /// Gets a value indicating if the error is transient.
-        /// </summary>
-        public bool IsTransient { get; private set; }
-
-        /// <summary>
-        /// Gets or sets the Azure IoT service-side Tracking ID in Support Requests.
-        /// </summary>
-        public string TrackingId { get; set; }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="IotHubException"/> class.
+        /// Creates an instance of <see cref="IotHubException"/> an empty error message.
         /// </summary>
         public IotHubException() : base()
         {
         }
 
-        internal IotHubException(bool isTransient) : base()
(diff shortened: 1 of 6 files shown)


iothub/device/src/Common/Exceptions/IotHubException.cs
```diff
@@ -139,11 +129,26 @@ protected IotHubException(SerializationInfo info, StreamingContext context)
...
+        public bool IsTransient { get; private set; }
+
+        /// <summary>
+        /// The service returned tracking Id associated with this particular error.
+        /// </summary>
+        public string TrackingId { get; set; }
+
         /// <summary>
-        /// Sets the SerializationInfo object.
+        /// Sets the <see cref="IsTransient"/> and <see cref="TrackingId"/> information to the <see cref="SerializationInfo"/>.
```
drwill-ms, line 148: I don't quite get this sentence.

abhipsaMisra: What I meant here is, using this method you can set the tracking Id and transient flag into the serialization info for this exception. 
I can see how it can be confusing, esp since the method is called "Getxxx" and we perform a set operation within it: https://docs.microsoft.com/en-us/dotnet/api/system.exception.getobjectdata?view=net-5.0#System_Exception_GetObjectData_System_Runtime_Serialization_SerializationInfo_System_Runtime_Serialization_StreamingContext_

How about:
`Sets the <see cref="IsTransient"/> and <see cref="TrackingId"/> information to the <see cref="SerializationInfo"/> and performs serialization.`

drwill-ms: Wow the method really is very poorly named, but it is specified by `System.Exception`.

Why not just copy the doc comment from the base method, but indicate that is how one would include the additional properties defined in this exception: `IsTransient` and `TrackingId`?

abhipsaMisra: sure

drwill-ms (approved):

PR #1669, opened 2020-11-17
Fix warnings: 1067, 1710, 1716

azabbasi:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1067
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1710
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1716

### DIFF
--- a/iothub/device/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
+++ b/iothub/device/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
@@ -55,24 +55,32 @@ namespace Microsoft.Azure.Devices.Client
             return Equals(obj as SharedAccessSignatureAuthorizationRule);
         }
 
-        [System.Diagnostics.CodeAnalysis.SuppressMessage(
-            "Globalization", "CA1307:Specify StringComparison",
-            Justification = "string.GetHashCode(StringComparison) is not supported for .net standard 2.0")]
-        // https://docs.microsoft.com/en-us/dotnet/api/system.string.gethashcode?view=net-5.0#System_String_GetHashCode_System_StringComparison_
-        public int GetHashCode(SharedAccessSignatureAuthorizationRule rule)
+        private int GetHashCode(SharedAccessSignatureAuthorizationRule rule)
         {
             if (rule == null)
             {
                 return 0;
             }
 
-            int hashKeyName = rule.KeyName == null ? 0 : rule.KeyName.GetHashCode();
+            int hashKeyName, hashPrimaryKey, hashSecondaryKey, hashRights;
 
-            int hashPrimaryKey = rule.PrimaryKey == null ? 0 : rule.PrimaryKey.GetHashCode();
+#if NETSTANDARD2_0 || NET451 || NET472
+            hashKeyName = rule.KeyName == null ? 0 : rule.KeyName.GetHashCode();
 
-            int hashSecondaryKey = rule.SecondaryKey == null ? 0 : rule.SecondaryKey.GetHashCode();
+            hashPrimaryKey = rule.PrimaryKey == null ? 0 : rule.PrimaryKey.GetHashCode();
 
-            int hashRights = rule.Rights.GetHashCode();
+            hashSecondaryKey = rule.SecondaryKey == null ? 0 : rule.SecondaryKey.GetHashCode();
+
+            hashRights = rule.Rights.GetHashCode();
+#else
+            hashKeyName = rule.KeyName == null ? 0 : rule.KeyName.GetHashCode(StringComparison.InvariantCultureIgnoreCase);
+
+            hashPrimaryKey = rule.PrimaryKey == null ? 0 : rule.PrimaryKey.GetHashCode(StringComparison.InvariantCultureIgnoreCase);
+
+            hashSecondaryKey = rule.SecondaryKey == null ? 0 : rule.SecondaryKey.GetHashCode(StringComparison.InvariantCultureIgnoreCase);
+
+            hashRights = rule.Rights.GetHashCode();
+#endif
 
             return hashKeyName ^ hashPrimaryKey ^ hashSecondaryKey ^ hashRights;
         }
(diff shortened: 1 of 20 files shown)


iothub/device/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -51,6 +50,13 @@ public bool Equals(SharedAccessSignatureAuthorizationRule other)
             return equals;
         }
 
+        public override bool Equals(object obj)
+        {
+            return obj is SharedAccessSignatureAuthorizationRule
```
barustum, line 55: if obj is of type SharedAccessSignatureAuthorizationRule, wouldn't the other 'Equals' method be called instead:
`public bool Equals(SharedAccessSignatureAuthorizationRule other)`

If so, that means this Equals will always throw because it will always be called when the passed in object is not SharedAccessSignatureAuthorizationRule.

I could be way off here and my cold symptoms are speaking :D

barustum: I guess it depends on the caller calls this method. Whether he passes in an object type (but underneath is of type SharedAccessSignatureAuthorizationRule), or passes in a SharedAccessSignatureAuthorizationRule type.

if that's the case, should line 56 be `Equals(obj as SharedAccessSignatureAuthorizationRule)`?

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -54,5 +54,37 @@ public bool Equals(SharedAccessSignatureAuthorizationRule other)
 
             return equals;
         }
+
+        public override bool Equals(object obj)
+        {
+            return Equals(obj as SharedAccessSignatureAuthorizationRule);
```
barustum, line 60: The implementation of this method is different than the one in device.

iothub/device/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
azabbasi: You are not wrong, the only caviat is that if you pass in an instance of the SharedAccessSignatureAuthorizationRule as an object, it will hi that path. 
The main reason we now have this method override is to basically get rid of the warning

azabbasi: I opted to not throw the exception and always call the implemented model with the SharedAccessSignatureAuthorizationRule as an input.

iothub/device/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -72,5 +76,10 @@ public int GetHashCode(SharedAccessSignatureAuthorizationRule rule)
 
             return hashKeyName ^ hashPrimaryKey ^ hashSecondaryKey ^ hashRights;
         }
+
+        public override int GetHashCode()
```
drwill-ms, line 80: What exactly does this do?

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -54,5 +54,37 @@ public bool Equals(SharedAccessSignatureAuthorizationRule other)
...
+        }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage(
```
drwill-ms, line 63: Rather than suppress the whole thing, what about doing it conditionally based on the framework?

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -54,5 +54,37 @@ public bool Equals(SharedAccessSignatureAuthorizationRule other)
...
+            "Globalization", "CA1307:Specify StringComparison",
+            Justification = "string.GetHashCode(StringComparison) is not supported for .net standard 2.0")]
+        // https://docs.microsoft.com/en-us/dotnet/api/system.string.gethashcode?view=net-5.0#System_String_GetHashCode_System_StringComparison_
+        public int GetHashCode(SharedAccessSignatureAuthorizationRule rule)
```
drwill-ms, line 67: Should this be public?

drwill-ms (approved):
Ping me if you have any questions about my comments, if you don't see a response from me from the email notification before you are ready to merge.

iothub/device/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
azabbasi: Directly from the .NET documentation:
Serves as the default hash function.

for complex objects, you can override how the hash code is generated.
All Equatable objects require this method to be implemented by the class.

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
azabbasi: it doesn't seem to be referenced anywhere else and it's an internal class. so no, changing it to private

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
azabbasi: let me give that a shot and see how it would look like. I wanted to avoid adding more conditional logic based on framework but I guess it's better than supressing the whole method.

iothub/device/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
drwill-ms: I know what GetHashCode is and why it exists, I didn't understand what calling the same method with `this` as a parameter did, but I saw later in another file it is delegated to another (now private) method.

Why not just put that implementation in here?

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
azabbasi: done

PR #1672, opened 2020-11-18
Code cleanup in MqttIotHubAdapter: style, async/await

drwill-ms:


iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -386,66 +342,51 @@ private async Task ConnectAsync(IChannelHandlerContext context)
                 }
                 _stateFlags = StateFlags.Connecting;
 
-                ScheduleCheckConnectTimeout(context);
+                await ScheduleCheckConnectTimeoutAsync(context).ConfigureAwait(true);
```
jamdavi, line 345: **nit:** All of these ConfigureAwait(true) calls can be removed to make the code less verbose. Unless of course FxCop yells about it the I guess it's okay.

see: https://devblogs.microsoft.com/dotnet/configureawait-faq/#why-would-i-want-to-use-configureawaittrue

drwill-ms: Not a nit at all, actually. Typically this would be very bad.

However, and it is my fault for not pointing this out, at the top of this file there is this comment:

```csharp
     // Note on ConfigureAwait: dotNetty is using a custom TaskScheduler that binds Tasks to the corresponding
    // EventLoop. To limit I/O to the EventLoopGroup and keep Netty semantics, we are going to ensure that the	    // EventLoop. To limit I/O to the EventLoopGroup and keep Netty semantics, we are going to ensure that the
    // task continuations are executed by this scheduler using ConfigureAwait(true).	    // task continuations are executed by this scheduler using ConfigureAwait(true).
```

drwill-ms: As a rule, we use `ConfigureAwait(false)` on all library async calls. This appears to be an exception.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -53,15 +54,15 @@ private enum StateFlags
...
         private const string IotHubTrueString = "true";
         private const string SegmentSeparator = "/";
         private const int MaxPayloadSize = 0x3ffff;
         private const int MaxTopicNameLength = 0xffff;
 
+        private static readonly char[] s_wildcardChars = { MultiSegmentWildcardChar, SingleSegmentWildcardChar };
         private static readonly Action<object> s_pingServerCallback = PingServer;
         private static readonly Action<object> s_checkConnAckTimeoutCallback = ShutdownIfNotReady;
-        private static readonly Func<IChannelHandlerContext, Exception, bool> s_shutdownOnWriteErrorHandler = (ctx, ex) => { ShutdownOnError(ctx, ex); return false; };
+        private static readonly Func<IChannelHandlerContext, Exception, Task<bool>> s_shutdownOnWriteErrorHandler = async (ctx, ex) => { await ShutdownOnErrorAsync(ctx, ex).ConfigureAwait(true); return false; };
```
barustum, line 65: isn't using ConfigureAwait(true) here not needed? As in, removing it is like having the same effect.

barustum: aaah..it could be because you're trying to suppress a warning

drwill-ms: It is a function that will be called in a method below, so it needs to maintain the same async strategy used elsewhere in the file.

Before, it was unawaited.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -125,39 +124,32 @@ private enum StateFlags
...
         {
-            if (Logging.IsEnabled)
-            {
-                Logging.Enter(this, context.Name, nameof(ChannelActive));
-            }
+            Logging.Enter(this, context.Name, nameof(ChannelActive));
 
             _stateFlags = StateFlags.NotConnected;
 
-            await ConnectAsync(context).ConfigureAwait(true);
```
abhipsaMisra, line 143: is this to avoid fire-and-forget of async void?

drwill-ms: Yes

PR #1674, opened 2020-11-19
CA1056 Error Cleanup

jamdavi:
Ignoring this error to remove warnings

### DIFF
--- a/iothub/service/src/FileNotification.cs
+++ b/iothub/service/src/FileNotification.cs
@@ -15,10 +15,8 @@ namespace Microsoft.Azure.Devices
         public string DeviceId { get; set; }
 
         [JsonProperty(PropertyName = "blobUri", DefaultValueHandling = DefaultValueHandling.Ignore)]
-#pragma warning disable CA1056 // Uri properties should not be strings, changing this would break the API
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1056: Uri properties should not be strings.", Justification = "Changing this would break the API.")]
         public string BlobUri { get; set; }
-
-#pragma warning restore CA1056 // Uri properties should not be strings, changing this would break the API
 
         [JsonProperty(PropertyName = "blobName", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string BlobName { get; set; }
--- a/iothub/service/src/FileNotification.cs
+++ b/iothub/service/src/FileNotification.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Azure.Devices
     using Newtonsoft.Json;
 
     /// <summary>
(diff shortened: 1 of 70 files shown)


timtay-microsoft (approved):

iothub/service/src/FileNotification.cs
```diff
@@ -7,16 +7,19 @@ namespace Microsoft.Azure.Devices
...
-    /// Contains properties for file upload notifications 
+    /// Contains properties for file upload notifications
     /// </summary>
     public class FileNotification
     {
         [JsonProperty(PropertyName = "deviceId", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string DeviceId { get; set; }
 
         [JsonProperty(PropertyName = "blobUri", DefaultValueHandling = DefaultValueHandling.Ignore)]
+#pragma warning disable CA1056 // Uri properties should not be strings, changing this would break the API
```
abhipsaMisra, line 18: nit: how do you feel about suppressing this via an attribute instead:
`[System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1056:Uri properties should not be strings", Justification = "changing this would break the API")]`

1 line of code vs 2!

abhipsaMisra (approved):

iothub/service/src/FileNotification.cs
drwill-ms, line 18: punctuation: technically these are two sentences, so I'd use a period or a semi-colon (because they are related)

drwill-ms (approved):

iothub/service/src/FileNotification.cs
jamdavi: I like this MUCH better.

jamdavi: Fiixed with SupressMessage suggestion

iothub/service/src/FileNotification.cs
```diff
@@ -7,14 +7,15 @@ namespace Microsoft.Azure.Devices
...
-    /// Contains properties for file upload notifications 
+    /// Contains properties for file upload notifications
     /// </summary>
     public class FileNotification
     {
         [JsonProperty(PropertyName = "deviceId", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string DeviceId { get; set; }
 
         [JsonProperty(PropertyName = "blobUri", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1056: Uri properties should not be strings.", Justification = "Changing this would break the API.")]
```
azabbasi, line 18: maybe: "Public facing types cannot change as they are considered a breaking change"

azabbasi (approved):
LGTM

PR #1679, opened 2020-11-20
Reduce memory footprint of CertificateInstaller

drwill-ms:


provisioning/device/src/CertificateInstaller.cs
```diff
@@ -1,68 +1,31 @@
...
-using System;
-using System.Collections.Generic;
 using System.Security.Cryptography.X509Certificates;
+using Microsoft.Azure.Devices.Shared;
 
 namespace Microsoft.Azure.Devices.Provisioning.Client
 {
     internal static class CertificateInstaller
     {
-        private static readonly HashSet<string> s_installedCertificates = new HashSet<string>();
```
drwill-ms, line 13: By saving all the thumbprints installed in the user's cert store in a static member of a static class, it will never be released for the duration of the program.

At first, I thought about making the class an instance class so the thumbprint list could be used over and over, but then I looked at usage and we really just use it once, on construction of the DPS client when using a cert.

So I figured we might as well do less with this code. It shouldn't be any less performant.

iothub/service/tests/ClientWebSocketTransportTests.cs
```diff
@@ -105,6 +105,7 @@ public async Task ClientWebSocketTransportReadWithoutConnectTest()
         // The following tests can only be run in Administrator mode
         [TestMethod]
         [Timeout(10000)] // if it is going to fail, fail fast. Otherwise it can go on for 4+ minutes. :(
+        [DoNotParallelize]
```
drwill-ms, line 108: This test always fails for me locally when run with the other tests in this class, so I thought we'd try this.

provisioning/device/src/CertificateInstaller.cs
```diff
@@ -1,68 +1,31 @@
...
-                            {
-                                Logging.Info(null, $"{nameof(CertificateInstaller)} adding {certificate.Thumbprint}");
-                            }
+                    Logging.Info(null, $"{nameof(CertificateInstaller)} adding {certificate.Thumbprint}");
 
-                            store.Add(certificate);
-                            s_installedCertificates.Add(certificate.Thumbprint);
-                        }
-                    }
+                    store.Add(certificate);
```
jamdavi, line 28: Why are we adding this back to the store? Seems like we're just logging the existence of this.

drwill-ms: The code is only adding it to the store when it wasn't already in the store (the original code captured all the installed cert thumbprints and used that to identify).

As to why it adds it, I'm not 100% sure. I was going to ask our local security guy (CrisPop) for advice about it.

provisioning/device/src/CertificateInstaller.cs
```diff
@@ -1,68 +1,31 @@
...
-        }
-
         public static void EnsureChainIsInstalled(X509Certificate2Collection certificates)
         {
             if (certificates == null)
             {
                 return;
             }
 
-            lock (s_lock)
```
CIPop, line 47: I think the lock should stay. That said, it won't fix the case where multiple processes try to do the operation at the same time.
I remember having this discussion with Schannel: sometimes the certificate storage files can get in an odd state on Windows if two processes try to add the same cert at the same time.
Likely the issue exists in Linux/OpenSSL where files are created on disk.

Ideally, customers should have the Intermediates pre-installed and not use this API. (Tool available here: https://gist.github.com/CIPop/baf530337cc6b29e3d6a88047661f69c and on Windows `certmgr.msc` can be used.)

provisioning/device/src/CertificateInstaller.cs
```diff
@@ -1,68 +1,31 @@
...
-                using (var store = new X509Store(StoreName.CertificateAuthority, StoreLocation.CurrentUser))
-                {
-                    store.Open(OpenFlags.ReadOnly);
-                    foreach (X509Certificate2 certificate in store.Certificates)
-                    {
-                        s_installedCertificates.Add(certificate.Thumbprint);
-                    }
-                }
-            }
-            catch (Exception ex)
```
CIPop, line 29: The exception handling is important: if certs fail to install you would want to get the information in logs. 
It could mean that either the process can't access the stores (device won't authenticate) or that some corruption happened with the store. I remember seeing the last one in SslStream testing (quite rare but very hard to debug).

provisioning/device/src/CertificateInstaller.cs
CIPop: TLS RFC requires that the client sends the client certificate _and_ the chain up to (but not including) the issuer trusted by the server. (In this case the issuer is not necessary a CA root - can be any issuer that the server is aware of.)
The server's responsibility is to take the partial chain offered by the client and form a chain with any of the certificates it trusts. If a valid chain is formed then the client is trusted and the TLS handshake resumes.

Some interesting notes:
1. The certificate trusted by the store doesn't need to be sent by the client. This is an optimization that makes the TLS Client Hello smaller.
2. The client cert (or chain) doesn't need to be trusted by the client.

To summarize: 
* it is mandatory to have the intermediates not known by the server within the Intermediate Store (which is called `StoreName.CertificateAuthority` in .NET, _not_ to be confused with `Root`! (https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.storename?view=net-5.0).
* it is not recommended to use this functionality because of race conditions. Instead, it's best to store _all_ certificates in stores (including the client certificate) or in Hardware Security Modules rather than files (HSM drivers on Windows would expose the certs through the store).

provisioning/device/src/CertificateInstaller.cs
```diff
@@ -1,68 +1,31 @@
...
-
-                    foreach (X509Certificate2 certificate in certificates)
-                    {
-                        if (!s_installedCertificates.Contains(certificate.Thumbprint))
-                        {
-                            if (Logging.IsEnabled)
-                            {
-                                Logging.Info(null, $"{nameof(CertificateInstaller)} adding {certificate.Thumbprint}");
-                            }
+                    Logging.Info(null, $"{nameof(CertificateInstaller)} adding {certificate.Thumbprint}");
```
CIPop, line 26: Perf: (I copied logging from .NET Core.)
If (Logging.IsEnabled) is preventing the string expansion (slow) from happening if the log is not enabled. Removing it will perform the string formatter _always_.

The recommendation is to add the check whenever the logging line has parameters that end up taking CPU cycles (all except passing already evaluated primitive types). 

E.g. I think Log.Associate below doesn't require the if check (it does save one if branch but I don't see that as important as string formatting for example).

provisioning/device/src/ProvisioningDeviceClient.cs
```diff
@@ -110,14 +103,11 @@ public Task<DeviceRegistrationResult> RegisterAsync(CancellationToken cancellati
...
-            if (Logging.IsEnabled)
-            {
-                Logging.RegisterAsync(this, _globalDeviceEndpoint, _idScope, _transport, _security);
-            }
+            Logging.RegisterAsync(this, _globalDeviceEndpoint, _idScope, _transport, _security);
 
             var request = new ProvisioningTransportRegisterMessage(_globalDeviceEndpoint, _idScope, _security, data?.JsonData)
             {
-                ProductInfo = ProductInfo
+                ProductInfo = ProductInfo,
```
CIPop, line 110: Nit: the reason I add `,` (and why it's supported in the language) in a list is that a future addition doesn't touch the line above (and the diff is one line smaller).

CIPop (commented):
LGTM:
Perf-wise this is going to be slower (how much should be assessed at scale) since it performs system calls for each client performing client auth with group certs. I would not though hold back the change - instead I recommend our customers to pre-install the chains and avoid calling the API entirely.

provisioning/device/src/CertificateInstaller.cs
drwill-ms: I assumed the lock was here because the HashMap was static. Now that we don't have that, I don't see the point.

This class is only used in 1 place, in a DPS client factory method.

provisioning/device/src/ProvisioningDeviceClient.cs
drwill-ms: That's why I add it too. I don't understand your nit, though if you like having it, and I added it.

provisioning/device/src/CertificateInstaller.cs
CIPop: If more than one DPS clients make the call the factory method will run multiple times. The only concern is when more than one DPS clients exist and they all run the code at the same time. AFAIK, the cert installer function is not thread safe.

jamdavi (approved):
LGTM, question was answered by CIPop

PR #1683, opened 2020-11-20
CA1062 and bumping FxCop to get better suppression rules

jamdavi:
## Bumping FxCop to 3.3.1

I ended up bumping the FxCop version to get some additional exclusions. Namely the `dotnet_code_quality.CA1062.exclude_extension_method_this_parameter = true` exclusion.

**However I also had to convert a few errors to warnings and you can see the addition behind the flag.**

## CA1062 Fixes
Bumping the version and adding the exclusion reduced the number of CA1062s from 74 to 2 (yes ... 2). So I'd like to keep this in instead of littering our code with a number of supression rules

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -298,3 +298,19 @@ dotnet_naming_style.end_in_async.required_prefix =
 dotnet_naming_style.end_in_async.required_suffix = Async
 dotnet_naming_style.end_in_async.capitalization = pascal_case
 dotnet_naming_style.end_in_async.word_separator = 
+
+# Exclude extension method 'this' parameter
+# https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062
+dotnet_code_quality.CA1062.exclude_extension_method_this_parameter = true
+
+# CA2016: Forward the 'CancellationToken' parameter to methods that take one
+dotnet_diagnostic.CA2016.severity = warning
+
+# CA1805: Do not initialize unnecessarily
+dotnet_diagnostic.CA1805.severity = warning
+
+# CA1834: Consider using 'StringBuilder.Append(char)' when applicable
+dotnet_diagnostic.CA1834.severity = warning
+
+# CA2000: Dispose objects before losing scope
+dotnet_diagnostic.CA2000.severity = warning
--- a/.editorconfig
+++ b/.editorconfig
@@ -300,17 +300,26 @@ dotnet_naming_style.end_in_async.capitalization = pascal_case
 dotnet_naming_style.end_in_async.word_separator = 
 
 # Exclude extension method 'this' parameter
-# https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062
+#
+# For all methods the default behaivor of the analyzer is to check to see if any argument is being 
+# used without a null check (this is good). However, for extension methods, this means that 
+# the "this" argument to the extension method is checked alongside of the other arguments.
+#
+# This is not great behavior as an extension method should behave like an instance method and we should
+# allow a NullRefException be generated if we attempt to access the "this" object.
+#
+# This flag ignores all of the "this" arguments but still reports the others.
 dotnet_code_quality.CA1062.exclude_extension_method_this_parameter = true
 
+
+# All of these severity changes below were because of an upgrade to the 3.3.1 analyzers which turned them
+# from warnings to errors. As we fix these warnings we should remove these lines to allow the default behaivor
+# and break the build.
 # CA2016: Forward the 'CancellationToken' parameter to methods that take one
 dotnet_diagnostic.CA2016.severity = warning
-
 # CA1805: Do not initialize unnecessarily
 dotnet_diagnostic.CA1805.severity = warning
-
 # CA1834: Consider using 'StringBuilder.Append(char)' when applicable
 dotnet_diagnostic.CA1834.severity = warning
-
 # CA2000: Dispose objects before losing scope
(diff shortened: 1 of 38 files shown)


.editorconfig
```diff
@@ -298,3 +298,19 @@ dotnet_naming_style.end_in_async.required_prefix =
 dotnet_naming_style.end_in_async.required_suffix = Async
 dotnet_naming_style.end_in_async.capitalization = pascal_case
 dotnet_naming_style.end_in_async.word_separator = 
+
+# Exclude extension method 'this' parameter
```
drwill-ms, line 302: Would you speak to these new settings in more detail?

iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -164,7 +164,7 @@
 
   <ItemGroup>
     <!-- FXCop -->
-    <PackageReference Condition=" '$(Configuration)' == 'Debug' " Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.9.8">
+    <PackageReference Condition=" '$(Configuration)' == 'Debug' " Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.1">
```
drwill-ms, line 167: Track2 libraries are only on 2.6.2. Do you mind starting a convo with Pavel Krymets about why they haven't updated to the latest?

drwill-ms: https://github.com/Azure/azure-sdk-for-net/blob/master/eng/Packages.Data.props#L66

iothub/service/src/Common/TrackingHelper.cs
```diff
@@ -155,6 +155,11 @@ public static string GetGatewayId(this AmqpLink link)
 
         public static ErrorCode GetErrorCodeFromAmqpError(Error ex)
         {
+            if (ex == null)
+            {
+                throw new ArgumentNullException(nameof(ex), "The Error property of the Ampq exception is null.");
```
barustum, line 160: AMQP

iothub/service/src/Microsoft.Azure.Devices.csproj
jamdavi: These have been updated

.editorconfig
jamdavi: Sure:
```dotnet_code_quality.CA1062.exclude_extension_method_this_parameter = true```
This removed the false positive when an extension method wasn't checking the `this` parameter. Otherwise we'd have had to suppress ALL of these instances and there were quite a few.

```
# CA2016: Forward the 'CancellationToken' parameter to methods that take one
dotnet_diagnostic.CA2016.severity = warning

# CA1805: Do not initialize unnecessarily
dotnet_diagnostic.CA1805.severity = warning

# CA1834: Consider using 'StringBuilder.Append(char)' when applicable
dotnet_diagnostic.CA1834.severity = warning

# CA2000: Dispose objects before losing scope
dotnet_diagnostic.CA2000.severity = warning
```
These were all **warnings** in the original package analyzers but has been upgraded to an **error**. I plan to take these in other PRs to remove the need for these lines to be configured and we'll put them back as errors.

Comment (drwill-ms) [reply .editorconfig:302]: I suggest you add a comment for them to that effect.

**Comment by drwill-ms** [comment]
> dotnet_code_quality.CA1062.exclude_extension_method_this_parameter = true

Why not just fix our code to check for null?

**Comment by jamdavi** [comment]
> > dotnet_code_quality.CA1062.exclude_extension_method_this_parameter = true
> 
> Why not just fix our code to check for null?

That is an option, or just suppress the message. 

But to the end user of the API there is no "this" parameter they are only calling the extension method with the correct arguments. Considering the code below. Do we want to say that the `this stringToDoWorkOn string` paramter is an ArgumentNull or a NullRef (spoiler alert it's a NullRef)? Since it's a NullRef we should truly let the runtime decide that and let it throw the correct exception and let the developer they are trying to call this method on something null.

```
public static string JamesFunc(this string stringToDoWorkOn, string appendedString) {...}
string s = null;
string b = s.JamesFunction("append this")
```

**Comment by drwill-ms** [comment]
Extension methods can generate an NRE. If I have a var of the right type that is set to null, the extension method is still valid to call, but the `this` pointer will be null.

We shouldn't suppress it. We should validate.

**Review by drwill-ms — APPROVED** [review]

**Review by barustum — APPROVED** [review]

**Review by abhipsaMisra — APPROVED** [review]

File: iothub/service/src/RegistryManager.cs [hunk 4]
```diff
@@ -31,6 +31,10 @@ public static RegistryManager CreateFromConnectionString(string connectionString
         /// <returns> An RegistryManager instance. </returns>
         public static RegistryManager CreateFromConnectionString(string connectionString, HttpTransportSettings transportSettings)
         {
+            if (transportSettings == null)
```
azabbasi, line 34: this one is fine since the caller creates a new TransportSetting before calling into this method.

iothub/service/src/JobClient/JobClient.cs
```diff
@@ -31,6 +31,10 @@ public static JobClient CreateFromConnectionString(string connectionString)
         /// <returns> A JobClient instance. </returns>
         public static JobClient CreateFromConnectionString(string connectionString, HttpTransportSettings transportSettings)
         {
+            if (transportSettings == null)
```
azabbasi, line 34: this is fine since the caller at line 21 creates a new transport setting before calling into this method.

PR #1684, opened 2020-11-20
Supress CA2211

jamdavi:
Supress the public static field


iothub/service/src/Common/TrackingHelper.cs
```diff
@@ -16,7 +16,9 @@ namespace Microsoft.Azure.Devices.Common
 
     public static class TrackingHelper
     {
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA2211:Non-constant fields should not be visible", Justification = "This property may be used by others so it is not safe to modify it.")]
         public static string GatewayId;
```
drwill-ms, line 20: The potential problem here is that someone can set it directly, bypassing the checks in `CheckAndAddGatewayIdToTrackingId`. We may want to change this to a property with a private setter. I know it is a breaking change. We'd just need to surface it to Mark with a case as to why we'd do it.

drwill-ms: Ack, I realized it isn't specified by that method. I need to see more context in this file to know if the customer changing it is acceptable.

drwill-ms: Can you look into that?

drwill-ms (changes requested):
Waiting on call on `GatewayId` mutability by clients.

iothub/service/src/Common/TrackingHelper.cs
jamdavi: I updated the description of this PR.

After looking at the code and how we intend to use it, it would seem that we are allowing the customer to store the GatewayId here instead of in their own global context as a convenience. Many of the methods check for this field and will add it to the TrackingId if it's got non-null or empty value. Otherwise it just puts a 0 there.

iothub/service/src/Common/TrackingHelper.cs
```diff
@@ -16,7 +16,9 @@ namespace Microsoft.Azure.Devices.Common
 
     public static class TrackingHelper
     {
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA2211:Non-constant fields should not be visible", Justification = "This property may be used by others so it is not safe to modify it.")]
```
drwill-ms, line 19: Move namespace to using

iothub/service/src/Common/TrackingHelper.cs
drwill-ms, line 20: Where's the obsolete attribute? Are we going to add a property with a private setter to replace it?

jamdavi: No, there would be no need. 

If we add a private setter we'll still need to have some public mutator method that just sets the value to something since the idea is this field is to be reused across many threads.

drwill-ms (approved):

PR #1685, opened 2020-11-21
Remove redundant call for Enabling Twin links

bikamani:
EnableTwinLinksAsync is called as part of SendTwinMessageAsync
EnableTwinLinksAsync is a re-entrant and idempotent

Removing unnecessary call for EnableTwinLinksAsync and this does not  change any behavior.


iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -317,7 +317,6 @@ public override async Task EnableTwinPatchAsync(CancellationToken cancellationTo
             try
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                await _amqpUnit.EnableTwinLinksAsync(_operationTimeout).ConfigureAwait(false);
```
barustum, line 320: Not too familiar with this code, but where is the redundant part? Do we call EnableTwinLinksAsync before this instance, and that's why we're removing it?

azabbasi: Agreed, This method should be idempotent and re-entrant.

bikamani: I should have been more specific in the description.

1. EnableTwinLinksAsync is called as part of SendTwinMessageAsync
2.  EnableTwinLinksAsync is a re-entrant and idempotent

I just found this is being called twice while investigating another issue so just removing this statement, it doesn't change any behavior as of today.

bikamani: Apologize for late response on this since I was OOF.

jamdavi (approved):

PR #1686, opened 2020-11-24
Fix MQTT impl to deliver pending messages when connected with MQTT CleanSession set to false

abhipsaMisra:
You can read about MQTT CleanSession behavior for C2D messages here: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#receiving-cloud-to-device-messages

### DIFF
--- a/iothub/device/src/IDelegatingHandler.cs
+++ b/iothub/device/src/IDelegatingHandler.cs
@@ -34,6 +34,10 @@ namespace Microsoft.Azure.Devices.Client
 
         Task EnableReceiveMessageAsync(CancellationToken cancellationToken);
 
+        // This is to ensure that if device connects over MQTT with CleanSession flag set to false,
+        // then any message sent while the device was disconnected is delivered on the callback.
+        Task EnsurePendingMessagesAreDelivered(CancellationToken cancellationToken);
+
         Task DisableReceiveMessageAsync(CancellationToken cancellationToken);
 
         Task RejectAsync(string lockToken, CancellationToken cancellationToken);
--- a/iothub/device/src/IDelegatingHandler.cs
+++ b/iothub/device/src/IDelegatingHandler.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Azure.Devices.Client
 
         // This is to ensure that if device connects over MQTT with CleanSession flag set to false,
         // then any message sent while the device was disconnected is delivered on the callback.
-        Task EnsurePendingMessagesAreDelivered(CancellationToken cancellationToken);
+        Task EnsurePendingMessagesAreDeliveredAsync(CancellationToken cancellationToken);
 
         Task DisableReceiveMessageAsync(CancellationToken cancellationToken);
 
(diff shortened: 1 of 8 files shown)


iothub/device/src/IDelegatingHandler.cs
```diff
@@ -34,6 +34,10 @@ internal interface IDelegatingHandler : IContinuationProvider<IDelegatingHandler
 
         Task EnableReceiveMessageAsync(CancellationToken cancellationToken);
 
+        // This is to ensure that if device connects over MQTT with CleanSession flag set to false,
+        // then any message sent while the device was disconnected is delivered on the callback.
+        Task EnsurePendingMessagesAreDelivered(CancellationToken cancellationToken);
```
drwill-ms, line 39: Async suffix?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -360,14 +367,16 @@ private Message ProcessMessage(Message message, bool hasMessage)
...
                     lock (_syncRoot)
                     {
-                        _messageQueue.TryDequeue(out message);
-                        message.LockToken = message.LockToken;
-                        if (_qos == QualityOfService.AtLeastOnce)
+                        if (_messageQueue.TryDequeue(out message))
                         {
-                            _completionQueue.Enqueue(message.LockToken);
-                        }
+                            message.LockToken = message.LockToken;
```
drwill-ms, line 372: What does this do?

iothub/device/src/IDelegatingHandler.cs
abhipsaMisra: we don't have any impl in here, just the signature

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -360,14 +372,16 @@ private Message ProcessMessage(Message message, bool hasMessage)
...
+                            if (_qos == QualityOfService.AtLeastOnce)
+                            {
+                                _completionQueue.Enqueue(message.LockToken);
```
drwill-ms, line 380: What does enqueuing the lock token mean?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: good point, this is unnecessary

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: this is used while marking a c2d message as complete. (MQTT supports only completing messages, while on AMQP we can reject and abandon as well)

drwill-ms: I mean, I generally think of enqueuing objects to be picked up by something on the other side. In this case, we're enqueuing the lock token, not an actually object (message).

Is this how we communicate back to the service that we complete a message, by enqueuing the lock token?

drwill-ms (approved):

PR #1689, opened 2020-11-30
Fix 1802 and 1805 project warnings

azabbasi:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1802
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1805

### DIFF
--- a/iothub/device/src/AuthenticationWithTokenRefresh.cs
+++ b/iothub/device/src/AuthenticationWithTokenRefresh.cs
@@ -12,14 +12,15 @@ namespace Microsoft.Azure.Devices.Client
     /// <summary>
     /// Authentication method that uses a shared access signature token and allows for token refresh.
     /// </summary>
-    public abstract class AuthenticationWithTokenRefresh : IAuthenticationMethod
+    public abstract class AuthenticationWithTokenRefresh : IAuthenticationMethod, IDisposable
     {
         private readonly int _suggestedTimeToLiveSeconds;
         private readonly int _timeBufferPercentage;
 
         private int _bufferSeconds;
-        private readonly SemaphoreSlim _lock = new SemaphoreSlim(1);
+        private SemaphoreSlim _lock = new SemaphoreSlim(1);
         private string _token;
+        private bool _isDisposed;
 
         /// <summary>
         /// Gets a snapshot of the UTC token expiry time.
@@ -133,6 +134,33 @@ namespace Microsoft.Azure.Devices.Client
         private void UpdateTimeBufferSeconds(int ttl)
         {
             _bufferSeconds = (int)(ttl * ((float)_timeBufferPercentage / 100));
+        }
+
+        /// <summary>
+        /// Releases the unmanaged resources used by the Component and optionally releases the managed resources.
+        /// </summary>
+        /// <param name="disposing">true to release both managed and unmanaged resources; false to release only unmanaged resources.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_isDisposed)
+            {
+                if (disposing)
+                {
+                    _lock?.Dispose();
+                    _lock = null;
+                }
+
+                _isDisposed = true;
+            }
+        }
+
+        /// <summary>
+        /// Dispose resources
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(disposing: true);
+            GC.SuppressFinalize(this);
         }
     }
 }
--- a/iothub/device/src/AuthenticationWithTokenRefresh.cs
+++ b/iothub/device/src/AuthenticationWithTokenRefresh.cs
@@ -18,8 +18,8 @@ namespace Microsoft.Azure.Devices.Client
         private readonly int _timeBufferPercentage;
 
         private int _bufferSeconds;
-        private SemaphoreSlim _lock = new SemaphoreSlim(1);
-        private string _token = null;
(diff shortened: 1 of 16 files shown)


iothub/device/src/AuthenticationWithTokenRefresh.cs
```diff
@@ -18,8 +18,8 @@ public abstract class AuthenticationWithTokenRefresh : IAuthenticationMethod
         private readonly int _timeBufferPercentage;
 
         private int _bufferSeconds;
-        private SemaphoreSlim _lock = new SemaphoreSlim(1);
-        private string _token = null;
+        private readonly SemaphoreSlim _lock = new SemaphoreSlim(1);
```
drwill-ms, line 21: Is this disposed with the class/client? I ask because often we dispose them and set them to null but if this is readonly then we wouldn't be setting it to null.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -482,7 +484,7 @@ private static async void PingServer(object ctx)
 
                     if (Logging.IsEnabled)
                     {
-                        Logging.Info(context, $"Idle time was {idleTime}, so ping request was sent.", nameof(PingServer));
+                        Logging.Info(context, $"Idle time was {idleTime}, so ping request was sent.", nameof(PingServerAsync));
```
drwill-ms, line 487: Would you mind putting the `if (Logging.IsEnabled) ` prefix here?

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -491,26 +493,26 @@ private static async void PingServer(object ctx)
                     {
                         if (Logging.IsEnabled)
```
drwill-ms, line 494: Feel free to make this into 1 line, if you wish

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
```diff
@@ -21,10 +21,10 @@ internal class ProtocolRoutingDelegatingHandler : DefaultDelegatingHandler
         /// <summary>
         /// After we've verified that we could open the transport for any operation, we will stop attempting others in the list.
         /// </summary>
-        private bool _transportSelectionComplete = false;
-        private int _nextTransportIndex = 0;
+        private bool _transportSelectionComplete;
+        private int _nextTransportIndex;
 
-        private SemaphoreSlim _handlerLock = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim _handlerLock = new SemaphoreSlim(1, 1);
```
drwill-ms, line 27: some with this one - does it get disposed?

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
azabbasi: you mean make them into a single line? line 485 already has the check .

drwill-ms (approved):

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -82,16 +82,18 @@ private enum StateFlags
...
 
         private readonly ProductInfo _productInfo;
         private readonly ClientOptions _options;
 
-        private ushort _packetId = 0;
-        private SpinLock _packetIdLock = new SpinLock();
+        // default value for ushort is 0.
+        private ushort _packetId;
+
+        private SpinLock _packetIdLock;
```
abhipsaMisra, line 96: is this initialization removal intentional?

azabbasi: yes, this is a struct and it automatically gets initialized with all its properties (default value for all properties)

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
abhipsaMisra, line 27: same comment that drwill had added above -> when this class is disposed, ideally we should dispose this semaphore and set it to null

iothub/device/src/AuthenticationWithTokenRefresh.cs
azabbasi: That is an excellent point, and by that I mean this class should be disposable since it has disposable fields. Changing.

iothub/device/src/Transport/ProtocolRoutingDelegatingHandler.cs
azabbasi: it should, but it doesn't implement dispose. I will change it.

azabbasi: yap, I will go over these classes and implement dispose on all of them.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
drwill-ms: My mistake. OOps!

abhipsaMisra (approved):

PR #1693, opened 2020-12-01
Fixing CA1081

jamdavi:


build.ps1
```diff
@@ -120,7 +120,7 @@ Function BuildProject($path, $message)
         }
     }
 
-    & dotnet build $projectPath --verbosity $verbosity --configuration $configuration
+    & dotnet build $projectPath --verbosity $verbosity --configuration $configuration -logger:"C:\Users\jamdavi\source\repos\BasicFileLogger\BasicFileLogger\bin\Debug\BasicFileLogger.dll;out.log;CA1801"
```
drwill-ms, line 123: I don't think we want your path in this file

jamdavi: Doh

iothub/device/src/ModernDotNet/Common/IOThreadTimerSlim.cs
```diff
@@ -17,6 +18,7 @@ private void CreateTimer()
             this.timer = new Timer((obj) => callback(obj), callbackState, TimeSpan.FromMilliseconds(-1), TimeSpan.FromMilliseconds(-1));
         }
 
+        [SuppressMessage("Usage", "CA1801:Review unused parameters", Justification = "Unknown if this is used by customers so we should leave as is to preserve the API.")]
```
drwill-ms, line 21: internal class => not used by customers

iothub/device/src/Transport/AmqpIoT/AmqpIoTMessageConverter.cs
```diff
@@ -46,6 +47,7 @@ public static Message AmqpMessageToMessage(AmqpMessage amqpMessage)
             return message;
         }
 
+        [SuppressMessage("Usage", "CA1801:Review unused parameters", Justification = "Unknown if this is used by customers so we should leave as is.")]
```
drwill-ms, line 50: same, internal class => not used by customers

iothub/device/src/Common/ExceptionTrace.cs
```diff
@@ -208,16 +205,6 @@ internal void BreakOnException(Exception exception)
 #endif
         }
 
-        [MethodImpl(MethodImplOptions.NoInlining)]
```
jamdavi, line 211: Does nothing, not used anywhere...

iothub/device/src/Common/Fx.cs
```diff
@@ -258,27 +259,6 @@ private static bool TryGetDebugSwitch(string name, out object value)
 
 #endif // DEBUG
 
-        [SuppressMessage(FxCop.Category.Design, FxCop.Rule.DoNotCatchGeneralExceptionTypes,
```
jamdavi, line 261: After removing TraceUnhandled (which did nothing). This method did nothing...

iothub/device/src/Common/ExceptionTrace.cs
```diff
@@ -90,18 +91,13 @@ public void TraceHandled(Exception exception, string catchLocation, EventTraceAc
             this.BreakOnException(exception);
         }
 
-        public void TraceUnhandled(Exception exception)
```
jamdavi, line 93: Does nothing

iothub/device/src/Common/Singleton.cs
```diff
@@ -39,7 +40,7 @@ public Task OpenAsync(TimeSpan timeout, CancellationToken cancellationToken)
             return GetOrCreateAsync(timeout, cancellationToken);
         }
 
-        public Task CloseAsync(CancellationToken cancellationToken)
+        public Task CloseAsync()
```
drwill-ms, line 43: Why not remove this impl?

drwill-ms (approved):

iothub/device/src/Common/Singleton.cs
jamdavi: Even though this really isn't async all the way down it seems better to be consistent we have OpenAsync, GetOrCreateAsync, etc..

iothub/device/src/Common/ExceptionTrace.cs
```diff
@@ -73,6 +73,7 @@ public ObjectDisposedException ObjectDisposed(string message)
             return TraceException<ObjectDisposedException>(new ObjectDisposedException(null, message), TraceEventType.Error);
         }
 
+        [SuppressMessage("Usage", "CA1801:Review unused parameters", Justification = "Unused parameters are inside of the DEBUG compilation flag.")]
```
abhipsaMisra, line 76: nit: worth mentioning here what the unused parameter is.

iothub/device/src/Common/Singleton.cs
abhipsaMisra, line 43: Q - why not throw if cancellation was requested? Does the upstream call not have a cancellation token available?

jamdavi: One thing I thought about here was we only Dispose the object and return a task complete. If we cancel this should it be after the dispose? If so do we really want to throw an exception if we're disposing the object?

abhipsaMisra: You are correct, throwing an exception after disposing makes no sense.
My idea is to to check for cancellation before performing any processing operation; that way, we do not dispose if a cancellation has been requested.

jamdavi: Upon further review this is not used by anyone and it's not public so it's gone.

drwill-ms (approved):

PR #1694, opened 2020-12-01
feat(iot-device): Add additional logging to MQTT transport layer

abhipsaMisra:


iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -136,9 +137,7 @@ private enum StateFlags
         public override async void ChannelActive(IChannelHandlerContext context)
         {
             if (Logging.IsEnabled)
-            {
                 Logging.Enter(this, context.Name, nameof(ChannelActive));
```
abhipsaMisra, line 140: Updating the log statements to this format was just easier, let me know if you guys prefer a different formatting style.

drwill-ms: We talked about it last week, but knew we didn't have a quorum. I think the going notion was that we were okay without braces, just for these logging methods, but some preferred it all on one line. Personally, I'm fine without braces and fine with a broken line (to limit line length).

drwill-ms (approved):

PR #1696, opened 2020-12-03
Fix warnings: 1824, 1835

azabbasi:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1824
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1835


iothub/service/src/IotHubClientWebSocket.cs
```diff
@@ -229,24 +229,19 @@ public async Task<int> ReceiveAsync(byte[] buffer, int offset, int size, TimeSpa
                 {
                     // Ignore pong frame and start over
                     totalBytesRead = 0;
-                    do
```
azabbasi, line 232: This while loop is unnecessary as totalBytesRead always resets back to 0 and we only read 2 bytes from the stream.

iothub/service/src/IotHubClientWebSocket.cs
```diff
@@ -229,24 +229,19 @@ public async Task<int> ReceiveAsync(byte[] buffer, int offset, int size, TimeSpa
...
-                    {
-                        bytesRead = await WebSocketStream.ReadAsync(header, totalBytesRead, header.Length - totalBytesRead).ConfigureAwait(false);
-                        if (bytesRead == 0)
-                        {
-                            throw new IOException(FramingPrematureEOF, new InvalidDataException("IotHubClientWebSocket was expecting more bytes"));
-                        }
+                    totalBytesRead = await ReadFromStreamAsync(WebSocketStream, header).ConfigureAwait(false);
```
azabbasi, line 232: Since there are multiple instances of this warning across this class and I didn't want to clutter the class with #if statements, I created methods that would contain the framework conditions in there.

drwill-ms (approved):

abhipsaMisra (approved):

PR #1697, opened 2020-12-03
Fix warning CA2000 - PART 1

azabbasi:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2000

I did not include any other changes with this PR besides the CA2000 warnings to make sure I can get eyes on all the important changes with no distractions.


iothub/device/src/Edge/InstalledCertificateValidator.cs
```diff
@@ -33,9 +33,9 @@ private void SetupCertificateValidation()
         {
             Debug.WriteLine("InstalledCertificateValidator.SetupCertificateValidation()");
 
-            foreach (var cert in this._certs)
```
azabbasi, line 36: we don't need to create a new store object on every single cert

iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -87,26 +87,32 @@ IWebProxy proxy
 
             _httpClientObj = handler != null ? new HttpClient(handler) : new HttpClient();
 #else
+
+
```
barustum, line 91: extra space

barustum (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -357,8 +354,9 @@ public override async Task<Message> ReceiveAsync(TimeoutHelper timeoutHelper)
             }
         }
 
-        private Message ProcessMessage(Message message, bool hasMessage)
+        private Message ProcessMessage(bool hasMessage)
```
azabbasi, line 357: This method was always called with null as the input for `message`. There is no reason to take it as an input since it can be super confusing. Refactored.

iothub/device/src/Edge/InstalledCertificateValidator.cs
abhipsaMisra: I don't see anyone referencing this private variable `store`, is it even required? or is the only validation that the certificate can be added to **_a_** store?

iothub/device/src/HsmAuthentication/HttpClientHelper.cs
```diff
@@ -33,7 +33,8 @@ public static HttpClient GetHttpClient(Uri providerUri)
 #if !NET451
             if (providerUri.Scheme.Equals(UnixScheme, StringComparison.OrdinalIgnoreCase))
             {
-                client = new HttpClient(new HttpUdsMessageHandler(providerUri));
+                using var httpUdsMessageHandler = new HttpUdsMessageHandler(providerUri);
+                client = new HttpClient(httpUdsMessageHandler);
```
abhipsaMisra, line 37: shouldn't this `HttpUdsMessageHandler` be disposed only when we dispose the `HttpClient`?

azabbasi: it will be, as long as it's in scope, it won't be disposed.

iothub/device/src/ModernDotNet/HsmAuthentication/Transport/HttpUdsMessageHandler.cs
```diff
@@ -20,8 +20,8 @@ public HttpUdsMessageHandler(Uri providerUri)
 
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
-            Socket socket = await GetConnectedSocketAsync().ConfigureAwait(false);
-            HttpBufferedStream stream = new HttpBufferedStream(new NetworkStream(socket, true));
+            using Socket socket = await GetConnectedSocketAsync().ConfigureAwait(false);
+            using var stream = new HttpBufferedStream(new NetworkStream(socket, true));
```
abhipsaMisra, line 24: Q - does the returned `HttpResponseMessage` contain a reference to the `HttpBufferedStream`? If yes, then we shouldn't dispose it here.

iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -87,26 +87,31 @@ IWebProxy proxy
...
+                }
+
+                if (proxy != DefaultWebProxySettings.Instance)
+                {
+                    httpClientHandler.UseProxy = proxy != null;
+                    httpClientHandler.Proxy = proxy;
+                }
 
-            _httpClientObj = new HttpClient(httpClientHandler);
+                _httpClientObj = new HttpClient(httpClientHandler);
```
drwill-ms, line 113: now you've saved this to the class but disposed the handler that you passed to it. That's not good. I think you need to save the handler to the class as a field as well, and dispose it with the `HttpClient` object

drwill-ms (changes requested):
Need to address one comment

azabbasi:
Closing this PR since there are multiple test failures and I need to start over to ensure I am not missing anything.

PR #1699, opened 2020-12-04
Fixing CA2000

azabbasi:


provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -78,14 +78,7 @@ public ProvisioningTransportHandlerHttp()
 
                 Logging.Associate(authStrategy, this);
 
-                var httpClientHandler = new HttpClientHandler
-                {
-                    // Cannot specify a specific protocol here, as desired due to an error:
```
drwill-ms, line 83: These are all notes I left for the Jedi work I did this year, that we're going to have to finish this next year. I would prefer you leave them here.

iothub/device/src/Edge/InstalledCertificateValidator.cs
```diff
@@ -32,10 +32,13 @@ public static InstalledCertificateValidator Create(IList<X509Certificate2> certs
...
         {
             Debug.WriteLine("InstalledCertificateValidator.SetupCertificateValidation()");
-
-            foreach (var cert in this._certs)
+#if NET451
+            var store = new X509Store(StoreName.Root, StoreLocation.CurrentUser);
+#else
+            using var store = new X509Store(StoreName.Root, StoreLocation.CurrentUser);
+#endif
+            foreach (X509Certificate2 cert in _certs)
```
abhipsaMisra, line 40: What do you think we intend to do here? The `store` looks to be a local `X509Store` variable, that we don't reference anywhere. Do we simply want to validate that the certificates can be added to **_any_** store? Do we think it's relevant if we add them to a new store or to the same store?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -357,8 +354,10 @@ public override async Task<Message> ReceiveAsync(TimeoutHelper timeoutHelper)
             }
         }
 
-        private Message ProcessMessage(Message message, bool hasMessage)
+        private Message ProcessMessage(bool hasMessage)
         {
+            Message message = null;
```
abhipsaMisra, line 359: 👍

iothub/device/src/Edge/InstalledCertificateValidator.cs
azabbasi: I think this just installs certificates in the cert store , it's the current user's root store btw
https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509store?view=netframework-4.7.2

drwill-ms (approved):

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -78,11 +78,11 @@ public ProvisioningTransportHandlerHttp()
...
                 Logging.Associate(authStrategy, this);
 
-                var httpClientHandler = new HttpClientHandler
+                using var httpClientHandler = new HttpClientHandler()
                 {
                     // Cannot specify a specific protocol here, as desired due to an error:
-                    //   ProvisioningDeviceClient_ValidRegistrationId_AmqpWithProxy_SymmetricKey_RegisterOk_GroupEnrollment failing for me with System.PlatformNotSupportedException: Operation is not supported on this platform.
-                    // When revisiting TLS12 work for DPS, we should figure out why. Perhaps the service needs to support it.
+                    //   ProvisioningDeviceClient_ValidRegistrationId_AmqpWithProxy_SymmetricKey_RegisterOk_GroupEnrollment failing for me with System.PlatformNotSupportedException: Operation is not supported on this platform.	
+                    // When revisiting TLS12 work for DPS, we should figure out why. Perhaps the service needs to support it.
```
jamdavi, line 85: RE on this: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.sslprotocols?view=net-5.0

Short answer is it's only supported on >= .NET4.7.1

Issue #1620, opened 2020-10-20
DeviceClient.SendEventAsync fails in versions  1.21.3 and newer of Microsoft.Azure.Devices.Client


mrtaunus:
# Context

- **OS, version, SKU and CPU architecture used:** <VERSION>Windows 10 and Azure Function App
- **Application's .NET Target Framework :** <NETVERSION> netcoreapp3.1
- **Device:** <DEVICETYPE> Laptop + Cloud
- **SDK version used:** <VERSION> Microsoft.Azure.Devices 1.27.3, Microsoft.Azure.Devices.Client 1.21.2 - 1.32.0

## Description of the issue
I have had a working Function App sending messages to Azure IoT Hub via DeviceClient.SendEventAsync. After upgrading referenced packages, that stopped working. After testing several packages, I found that upgrading Microsoft.Azure.Devices.Client to anything newer than 1.21.2 breaks my function in both local testing in Visual Studio as well as running in the cloud in an Azure Function App. The error is shown below. After downgrading back to 1.21.2 everything starts to work again.

## Code sample exhibiting the issue
```
var iotHubConnectionString = await KeyVaultFacade.GetIotHubConnectionString( context );

iotHubConnectionString += $";DeviceId={deviceId};";

var deviceClient = DeviceClient.CreateFromConnectionString( iotHubConnectionString );

// This call fails
await deviceClient.SendEventAsync( new Message( Encoding.UTF8.GetBytes( requestData.ToString() ) ) );
```

## Console log of the issue
Exception: 
`Microsoft.Azure.Devices.Client.Exceptions.DeviceNotFoundException: 'error(condition:amqp:not-found,description:{"errorCode":404001,"trackingId":"5a7c2af5076f4f499fa51faffb2f2c9e-TimeStamp:10/20/2020 09:47:48-G:47-TimeStamp:10/20/2020 09:47:48","message":"MyDeviceId;","timestampUtc":"2020-10-20T09:47:48.0903707Z"},info:[com.microsoft:is-filtered:True])'`

abhipsaMisra:
The exception message indicates that the device was not registered on Azure; however, from your explanation I understand that on simply downgrading the device client package version fixes this error for you (with no other change in the infrastructure setup).
[...]

Issue #583, opened 2018-08-15
IoTHub DeviceClient ConnectionString Unit Tests are disabled


CIPop:
This is tracking tests disabled within iothub\device\tests\ConnectionString\DeviceClientConnectionStringTests.cs

drwill-ms:
Updated the tests.

az-iot-builder-01:
@CIPop, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #1700, opened 2020-12-08
fix(devices-client): Add a comment to clarify why semi-colons is allowed in the SDK

barustum:
At some point, IoT Hub allowed semicolons in device names as part of the connection strings. Based on that, this PR was made:
https://github.com/Azure/azure-iot-sdk-csharp/commit/9dde3fee9e88fbc53ef1b9a0dbeee7dcfa9e7e6b

Since then, IoT reverted on this change. And now IoT does not allow semicolons. The SDK did not make the change to reflect that. Here's the change in documentation on when this happened:
https://github.com/MicrosoftDocs/azure-docs/commit/237b82518c01bc4cd9e3d98fc3d6b96ae2c3f487

This PR addresses this issue:
#1620 

The logic change in this change is similar to how the service client does it. I am not sure why the original PR did not make the change on the service side.

### DIFF
--- a/iothub/device/src/Common/Extensions/CommonExtensions.cs
+++ b/iothub/device/src/Common/Extensions/CommonExtensions.cs
@@ -61,8 +61,8 @@ namespace Microsoft.Azure.Devices.Client.Extensions
                 throw new ArgumentException("Malformed Token");
             }
 
-            // This regex allows semi-colons to be part of the allowed characters for device names. Although new devices are not allowed to have semi-colons in the name,
-            // some legacy devices still have them and so this name validation cannot be changed.
+            // This regex allows semi-colons to be part of the allowed characters for device names. Although new devices are not
+            // allowed to have semi-colons in the name, some legacy devices still have them and so this name validation cannot be changed.
             IEnumerable<string[]> parts = new Regex($"(?:^|{kvpDelimiter})([^{kvpDelimiter}{kvpSeparator}]*){kvpSeparator}")
                 .Matches(valuePairString)
                 .Cast<Match>()
--- a/iothub/device/src/Common/Extensions/CommonExtensions.cs
+++ b/iothub/device/src/Common/Extensions/CommonExtensions.cs
@@ -61,6 +61,8 @@ namespace Microsoft.Azure.Devices.Client.Extensions
                 throw new ArgumentException("Malformed Token");
             }
 
+            // This regex allows semi-colons to be part of the allowed characters for device names. Although new devices are not allowed to have semi-colons in the name,
+            // some legacy devices still have them and so this name validation cannot be changed.
             IEnumerable<string[]> parts = new Regex($"(?:^|{kvpDelimiter})([^{kvpDelimiter}{kvpSeparator}]*){kvpSeparator}")
                 .Matches(valuePairString)
                 .Cast<Match>()
--- a/iothub/device/src/Common/Extensions/CommonExtensions.cs
+++ b/iothub/device/src/Common/Extensions/CommonExtensions.cs
@@ -61,10 +61,15 @@ namespace Microsoft.Azure.Devices.Client.Extensions
                 throw new ArgumentException("Malformed Token");
             }
 
-            IEnumerable<string[]> parts = valuePairString
-                .Split(kvpDelimiter)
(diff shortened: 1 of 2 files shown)


timtay-microsoft:
Is it possible that there are registered devices on older IoT hubs that still have devices with semicolons in their names? If so, this would break those users

iothub/device/src/Common/Extensions/CommonExtensions.cs
```diff
@@ -61,15 +61,10 @@ public static string GetValueOrDefault(this IDictionary<string, string> map, str
...
-                .Select(m => new string[] {
-                    m.Result("$1"),
-                    valuePairString.Substring(
-                        m.Index + m.Value.Length,
-                        (m.NextMatch().Success ? m.NextMatch().Index : valuePairString.Length) - (m.Index + m.Value.Length))
-                });
+            IEnumerable<string[]> parts = valuePairString
+                .Split(kvpDelimiter)
+                .Where(p => p.Trim().Length > 0)
+                .Select(p => p.Split(new char[] { kvpSeparator }, 2));
```
drwill-ms, line 67: You can save off this character array as as `static readonly` so it is not allocated each time

drwill-ms (approved):

iothub/device/src/Common/Extensions/CommonExtensions.cs
barustum: > le that there are registered devices on older IoT hubs that still have devices with semicolons in their names? If so, this would break those users

Good point. Any suggestion on how we can figure this out? Would checking Kusto service logs be sufficient? Or should I talk with someone on the Hub team?

barustum:
> Is it possible that there are registered devices on older IoT hubs that still have devices with semicolons in their names? If so, this would break those users

Good point. Any suggestion on how I can find this out? I can take a look at Kusto service logs and query for any devices with a semicolon in it. Or talk with someone on the hub team if there are indeed devices with semicolons.

timtay-microsoft:
I'd ask the service folks over email/teams if it is possible. When they deliberately removed the semicolon from the set of allowable characters, they must have talked about if they would have to delete devices that were previously registered with semicolons in their name, or if they would allow them to stay registered

timtay-microsoft:
The risk/reward of this PR seems iffy to me if the service folks confirm that there could be devices floating around with semicolons in the name. Especially since we're trying to avoid breaking changes lately. The issue that spurred this PR seems to be resolved just by providing a connection string without a semicolon at the end, which seems like a reasonable expectation to put on the user.

If the service team confirms that it would be impossible for any legacy devices to have a semicolon in them, then I'm all for this change.

azabbasi (approved):
As we have already discussed that this is a good change if there are no more devices out there that will be broken with this change.

iothub/device/tests/ConnectionString/DeviceClientConnectionStringTests.cs
```diff
@@ -96,6 +96,15 @@ public void DeviceClientConnectionStringX509CertTest()
             var deviceClient = DeviceClient.CreateFromConnectionString(connectionString);
         }
 
+        [TestMethod]
```
vinagesh, line 99: As you are on this file, any chance you can take a look at the ignored tests? There are many and there is a TODO to reenable them. We should have good test coverage for this. Also I recently found it was tracked here - https://github.com/Azure/azure-iot-sdk-csharp/issues/583.

If you feel it is way out of the current scope, we can make a task for it. Let me know.

barustum:
I synched up with the Hub team. It turns out that there are still devices out there with semicolons. So this is a breaking change. The issue that spurred this PR seems to be resolved just by providing a connection string without a semicolon at the end, which seems like a reasonable expectation to put on the user.

iothub/device/tests/ConnectionString/DeviceClientConnectionStringTests.cs
barustum: Ended up abandoning this PR :)

barustum:
I re-opened the PR to add a comment in the code to clarify why we're doing what we're doing.

drwill-ms (approved):

PR #1702, opened 2020-12-09
Fix CA 2000 Part 2

azabbasi:


iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -63,51 +62,50 @@ IWebProxy proxy
...
             {
-                if (handler == null)
+                if (_httpClientHandler == null)
                 {
-                    handler = new WebRequestHandler();
+                    _httpClientHandler = new WebRequestHandler();
                 }
 
-                handler.ClientCertificates.Add(clientCert);
+                (_httpClientHandler as WebRequestHandler).ClientCertificates.Add(clientCert);
```
drwill-ms, line 74: Why not specify WebRequestHandler as the type on line 41?

iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -63,51 +62,50 @@ IWebProxy proxy
...
                 }
 
-                handler.UseProxy = (proxy != null);
-                handler.Proxy = proxy;
+                _httpClientHandler.UseProxy = (proxy != null);
+                _httpClientHandler.Proxy = proxy;
             }
 
-            _httpClientObj = handler != null ? new HttpClient(handler) : new HttpClient();
+            _httpClientObj = _httpClientHandler != null ? new HttpClient(_httpClientHandler) : new HttpClient();
```
drwill-ms, line 89: It can't be null, right?

drwill-ms (approved):

iothub/device/src/Transport/HttpClientHelper.cs
azabbasi: we use WebRequestHandler only for NET451
for the other frameworks, it's a HttpClientHandler

iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -532,6 +530,14 @@ public void Dispose()
                     _httpClientObj = null;
                 }
 
+                // HttpClientHandler that is used to create HttpClient will automatically be disposed when HttpClient is disposed
+                // But in case the client handler didn't end up being used by the HttpClient, we explicitly dispose it here.
+                if (_httpClientHandler != null)
+                {
+                    _httpClientHandler?.Dispose();
```
barustum, line 537: nit: question mark here is redundant since we already check if it's not null

barustum (approved):

azabbasi (commented):
comments disabled.

iothub/device/src/Transport/HttpClientHelper.cs
azabbasi: I will change it in an upcoming PR since I have the gates passed already and this is safe.

PR #1705, opened 2020-12-11
Update ClientFactory.cs

azabbasi:


iothub/device/src/ClientFactory.cs
```diff
@@ -23,7 +23,7 @@ internal class ClientFactory
         /// <summary>
         /// Create an Amqp InternalClient from individual parameters
         /// </summary>
-        /// <param name="hostname">The fully-qualified DNS hostname of IoT Hub</param>
+        /// <param name="hostname">The fully-qualified DNS host name of IoT Hub</param>
```
barustum, line 26: The other overloaded methods also have hostname.

azabbasi: just a test to run the pipeline. will close afterwards.

PR #1708, opened 2020-12-14
Fix warnings 2008, 2000, 1834, 2016

azabbasi:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2008

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -975,7 +975,7 @@ namespace Microsoft.Azure.Devices.Client
                 return;
             }
 
-            MethodResponseInternal methodResponseInternal;
+            MethodResponseInternal methodResponseInternal = null;
             byte[] requestData = methodRequestInternal.GetBytes();
 
             await _methodsDictionarySemaphore.WaitAsync().ConfigureAwait(false);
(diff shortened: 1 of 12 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -999,6 +999,9 @@ internal async Task OnMethodCalledAsync(MethodRequestInternal methodRequestInter
                 methodResponseInternal = new MethodResponseInternal(methodRequestInternal.RequestId, (int)MethodResponseStatusCode.BadRequest);
 
                 await SendMethodResponseAsync(methodResponseInternal, methodRequestInternal.CancellationToken).ConfigureAwait(false);
+
+                methodResponseInternal?.Dispose();
```
drwill-ms, line 1003: Won't this already happen on 1016 in the finally clause?

iothub/device/src/InternalClient.cs
```diff
@@ -1058,6 +1062,8 @@ internal async Task OnMethodCalledAsync(MethodRequestInternal methodRequestInter
 
             await SendMethodResponseAsync(methodResponseInternal, methodRequestInternal.CancellationToken).ConfigureAwait(false);
 
+            methodResponseInternal?.Dispose();
```
drwill-ms, line 1065: If `SendMethodResponseAsync` throws, this won't be disposed.

drwill-ms (approved):

iothub/device/src/InternalClient.cs
azabbasi: Correct. Fixed.

PR #1710, opened 2020-12-14
Fix(warnings) : CA2213 - CA2215

azabbasi:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2215
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2213


iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -557,7 +559,6 @@ protected override void Dispose(bool disposing)
                 if (disposing)
                 {
                     _closed = true;
-                    OnTransportClosedGracefully();
```
azabbasi, line 560: This is called during `base.Dispose()`

drwill-ms (approved):

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -625,13 +625,13 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
 
                 _handleDisconnectCts.Cancel();
                 await base.CloseAsync(cancellationToken).ConfigureAwait(false);
-                Dispose(true);
```
azabbasi, line 628: Since in the finally statement we are releasing the semaphore, we need to dispose the object after the semaphore has been released to avoid nullref.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
abhipsaMisra: I am concerned about the ordering of the transport closure calls - previously we would call `OnTransportClosedGracefully` only if `disposing` was true, but now we will call it anytime Dispose() is called.

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -389,6 +389,7 @@ private Task<SendingAmqpLink> CreateSendingLinkAsync(TimeSpan timeout)
         /// <inheritdoc/>
         protected override void Dispose(bool disposing)
         {
+            base.Dispose(disposing);
```
abhipsaMisra, line 392: q - what operation is associated with this base.dispose() call?

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -117,7 +117,8 @@ public ProvisioningTransportHandlerHttp()
                     .RegisterDeviceAsync(
                         registrationId,
                         message.IdScope,
-                        deviceRegistration)
+                        deviceRegistration,
+                        cancellationToken: cancellationToken)
```
abhipsaMisra, line 121: ooh, cancellation token was not being passed through, not good! :(

iothub/service/src/AmqpServiceClient.cs
azabbasi: There is nothing, the pattern should exist though regardless of what happens in the base class.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
azabbasi: I don't think that is the case. We call base.Dispose(disposing) so the order for the transport closure is intact.

iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -640,6 +640,13 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
         /// </summary>
         private async Task EnsureOpenedAsync(CancellationToken cancellationToken)
         {
+            // If this object has already been disposed, we will throw an exception indicating that.
```
abhipsaMisra, line 643: nit: maybe add a note specifying that this is because the current behavior doesn't allow open -> close -> open

abhipsaMisra (approved):
Looks good!
Since changing dispose ~class~ calls makes me a bit queasy, how do you feel about running the entire test suite on this branch first (with fault injection) and then merging it in?

PR #1713, opened 2020-12-15
Fix(e2e): Fix file upload bug (bug introduced before any release happened. release note not required)

azabbasi:
The commit that broke the functionality:
https://github.com/Azure/azure-iot-sdk-csharp/commit/220acb35b07a1e8537f27376e1075e3504af1545

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -1543,20 +1543,7 @@ namespace Microsoft.Azure.Devices.Client
                     throw Fx.Exception.Argument(nameof(blobName), "Path segment count cannot exceed 254");
                 }
 
-                var context = new PipelineContext();
-                context.Set(_productInfo);
-
-                var transportSettings = new Http1TransportSettings();
-
-                //We need to add the certificate to the fileUpload httpTransport if DeviceAuthenticationWithX509Certificate
-                if (Certificate != null)
-                {
-                    transportSettings.ClientCertificate = Certificate;
-                }
-
-                using var httpTransport = new HttpTransportHandler(context, IotHubConnectionString, transportSettings);
-
-                return httpTransport.UploadToBlobAsync(blobName, source, cancellationToken);
+                return _fileUploadHttpTransportHandler.UploadToBlobAsync(blobName, source, cancellationToken);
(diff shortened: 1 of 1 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -1543,20 +1543,7 @@ public Task UploadToBlobAsync(string blobName, Stream source, CancellationToken
                     throw Fx.Exception.Argument(nameof(blobName), "Path segment count cannot exceed 254");
                 }
 
-                var context = new PipelineContext();
```
azabbasi, line 1546: we don't need to create a new transport handler as the internal client is initialized with one in the first place.

abhipsaMisra: was this entire block added recently?

abhipsaMisra: Could you add a bit of context on what the code was previously, and what was changed?

abhipsaMisra: Or maybe just link the commits :)

drwill-ms (approved):

iothub/device/src/InternalClient.cs
```diff
@@ -1543,20 +1543,7 @@ public Task UploadToBlobAsync(string blobName, Stream source, CancellationToken
...
-                context.Set(_productInfo);
-
-                var transportSettings = new Http1TransportSettings();
```
drwill-ms, line 1549: The difference between this one and the ctor one, is that the ctor could be any other transport, and this is specifically Http.

iothub/device/src/InternalClient.cs
azabbasi: spoke to Abhipsa offline.

iothub/device/src/InternalClient.cs
azabbasi: file upload is always over HTTP though

drwill-ms: Never mind.

azabbasi: including the one in the ctor.

PR #1714, opened 2020-12-15
CS4014

drwill-ms:
<https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs4014>
Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the await operator to the result of the call.

### DIFF
--- a/iothub/device/src/net451/Common/LegacyClientWebSocketTransport.cs
+++ b/iothub/device/src/net451/Common/LegacyClientWebSocketTransport.cs
@@ -1,64 +1,57 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+using System.IO;
+using System.Net;
+using System.Net.WebSockets;
+using System.Threading.Tasks;
+using Microsoft.Azure.Amqp;
+using Microsoft.Azure.Amqp.Transport;
+using Microsoft.Azure.Devices.Shared;
+
 namespace Microsoft.Azure.Devices.Client
 {
-    using System;
-    using System.IO;
-    using System.Net;
-    using System.Net.WebSockets;
-    using System.Threading.Tasks;
-    using Microsoft.Azure.Amqp;
-    using Microsoft.Azure.Amqp.Transport;
-    using Microsoft.Azure.Devices.Shared;
-
-    sealed class LegacyClientWebSocketTransport : TransportBase
+    internal sealed class LegacyClientWebSocketTransport : TransportBase
     {
-        const string ClientWebSocketTransportReadBufferTooSmall = "LegacyClientWebSocketTransport Read Buffer too small.";
-        const int MaxReadBufferSize = 256 * 1024; // Max Read buffer size is hard-coded to 256k
-        static readonly AsyncCallback onWriteComplete = OnWriteComplete;
+        private const string ClientWebSocketTransportReadBufferTooSmall = "LegacyClientWebSocketTransport Read Buffer too small.";
+        private const int MaxReadBufferSize = 256 * 1024; // Max Read buffer size is hard-coded to 256k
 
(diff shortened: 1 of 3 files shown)


iothub/device/src/net451/Common/LegacyClientWebSocketTransport.cs
```diff
@@ -68,51 +57,45 @@ public override void SetMonitor(ITransportMonitor usageMeter)
...
             if (Logging.IsEnabled)
-            {
                 Logging.Exit(this, args.Transport, $"{nameof(LegacyClientWebSocketTransport)}.{nameof(WriteAsync)}");
-            }
 
             return true;
         }
 
-        async Task WriteAsyncCore(TransportAsyncCallbackArgs args)
+        private async Task WriteCoreAsync(TransportAsyncCallbackArgs args)
```
drwill-ms, line 82: rename and explicitly add accessibility modifier

iothub/device/src/net451/Common/LegacyClientWebSocketTransport.cs
```diff
@@ -245,50 +220,40 @@ protected override bool CloseInternal()
...
-                }
 
-                var webSocketState = this.webSocket.State;
-                if (webSocketState != IotHubClientWebSocket.WebSocketState.Closed && webSocketState != IotHubClientWebSocket.WebSocketState.Aborted)
+                IotHubClientWebSocket.WebSocketState webSocketState = _webSocket.State;
+                if (webSocketState != IotHubClientWebSocket.WebSocketState.Closed
+                    && webSocketState != IotHubClientWebSocket.WebSocketState.Aborted)
                 {
-                    this.CloseInternalAsync();
+                    _ = CloseInternalAsync();
```
drwill-ms, line 229: Should this be awaited? For now, I'm explicitly indicating we don't care about waiting for it.

iothub/device/src/net451/Common/LegacyClientWebSocketTransport.cs
```diff
@@ -136,26 +119,22 @@ async Task WriteAsyncCore(TransportAsyncCallbackArgs args)
...
                 if (Logging.IsEnabled)
-                {
-                    Logging.Exit(this, args.Transport, $"{nameof(LegacyClientWebSocketTransport)}.{nameof(WriteAsyncCore)}");
-                }
+                    Logging.Exit(this, args.Transport, $"{nameof(LegacyClientWebSocketTransport)}.{nameof(WriteCoreAsync)}");
             }
         }
 
-        async Task<int> ReadAsyncCore()
+        async Task<int> ReadCoreAsync()
```
barustum, line 129: do we need to add private access modifier?

iothub/device/src/net451/Common/LegacyClientWebSocketTransport.cs
```diff
@@ -245,50 +220,40 @@ protected override bool CloseInternal()
...
                 if (Logging.IsEnabled)
-                {
                     Logging.Enter(this, $"{nameof(LegacyClientWebSocketTransport)}.{nameof(CloseInternalAsync)}");
-                }
 
-                await this.webSocket.CloseAsync().ConfigureAwait(false);
+                await _webSocket.CloseAsync().ConfigureAwait(false);
             }
-            catch (Exception e)
+            catch (Exception e) when (!Fx.IsFatal(e))
```
barustum, line 249: Before, we were rethrowing when it's a fatal exception. After, we don't attempt to rethrow if it's a fatal exception. Is this expected?

drwill-ms: Now we only catch when it isn't fatal. Meaning, if it is fatal, the exception is not caught.

Before, we'd rethrow if it was fatal.

So unless I've made a grievous error, it should be behaving the same.

barustum (approved):

PR #1716, opened 2020-12-16
Use CultureInvariant for validating device connection string values

drwill-ms:
The test I added fails without the additional regex option, and passes with it.

### DIFF
--- a/iothub/device/tests/ConnectionString/IotHubConnectionStringBuilderTests.cs
+++ b/iothub/device/tests/ConnectionString/IotHubConnectionStringBuilderTests.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Globalization;
+using System.Threading;
 using FluentAssertions;
 using Microsoft.Azure.Devices.Client;
 using Microsoft.Azure.Devices.Client.ApiTest;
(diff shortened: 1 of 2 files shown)


iothub/device/tests/ConnectionString/IotHubConnectionStringBuilderTests.cs
```diff
@@ -41,6 +43,32 @@ public void IotHubConnectionStringBuilder_ParamConnectionString_ParsesDeviceId()
...
+
+            try
+            {
+                foreach (CultureInfo culture in allCultures)
+                {
+                    Console.WriteLine($"Testing culture {culture}");
+                    Thread.CurrentThread.CurrentCulture = culture;
+                    var connectionString = $"HostName={HostName};DeviceId={deviceId};SharedAccessKey={SharedAccessKey}";
+                    var csBuilder = IotHubConnectionStringBuilder.Create(connectionString);
+                    csBuilder.DeviceId.Should().Be(deviceId);
```
jamdavi, line 63: If it fails will we know what culture caused it?

jamdavi (approved):
I like it. Since we're enumerating all cultures we should have some way to know pass/fail per culture otherwise we won't know what's broke.

barustum (approved):

iothub/device/tests/ConnectionString/IotHubConnectionStringBuilderTests.cs
drwill-ms: Good question. Yes because of the Console.WriteLine above, but I can add a comment to the assert to make it easier to find.

abhipsaMisra (approved):

arunkaithayil (approved):

PR #1717, opened 2020-12-16
fix(warnings):CS1570

barustum:
CS1570: XML comment on 'construct' has badly formed XML

### DIFF
--- a/iothub/device/src/ModuleClient.cs
+++ b/iothub/device/src/ModuleClient.cs
@@ -44,9 +44,7 @@ namespace Microsoft.Azure.Devices.Client
             }
 
             if (Logging.IsEnabled)
-            {
                 Logging.Associate(this, this, internalClient, nameof(ModuleClient));
-            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


abhipsaMisra (approved):

iothub/device/src/ModuleClient.cs
```diff
@@ -44,9 +44,7 @@ internal ModuleClient(InternalClient internalClient, ICertificateValidator certV
             }
 
             if (Logging.IsEnabled)
-            {
```
jamdavi, line 47: Did you intend to remove these? I haven't seen us choose one way or another in the code so not sure if you're confirming to a code style

iothub/service/src/DigitalTwin/Serialization/WritableProperty.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.Azure.Devices.Serialization
     /// <para>
```
drwill-ms, line 13: `<para>` tags not needed

drwill-ms (approved):

jamdavi (approved):

azabbasi (approved):

iothub/device/src/ModuleClient.cs
barustum: Yes. I believe that's the agreed upon style for the Logging.IsEnabled block that we discussed. Looking at recent PRs, that's what we've been doing.

drwill-ms: We decided for the ubiquitous logging only, to reduce how many lines of code it takes.

PR #1718, opened 2020-12-16
fix(warnings): Fix for CS1573

barustum:
Parameter 'parameter' has no matching param tag in the XML comment for 'parameter' (but other parameters do).

And other minor styling fixes.

### DIFF
--- a/iothub/service/src/RegistryManager.cs
+++ b/iothub/service/src/RegistryManager.cs
@@ -68,38 +68,38 @@ namespace Microsoft.Azure.Devices
         /// <summary>
         /// Register a new device with the system
         /// </summary>
-        /// <param name="device">The Device object to be registered.</param>
-        /// <returns>echoes back the Device object with the generated keys and ETags.</returns>
+        /// <param name="device">The Device object being registered.</param>
+        /// <returns>The Device object with the generated keys and ETags.</returns>
         public abstract Task<Device> AddDeviceAsync(Device device);
 
         /// <summary>
         /// Register a new device with the system
         /// </summary>
-        /// <param name="device">The Device object to be registered.</param>
+        /// <param name="device">The Device object being registered.</param>
         /// <param name="cancellationToken">The token which allows the operation to be canceled.</param>
-        /// <returns>echoes back the Device object with the generated keys and ETags.</returns>
+        /// <returns>The Device object with the generated keys and ETags.</returns>
         public abstract Task<Device> AddDeviceAsync(Device device, CancellationToken cancellationToken);
 
         /// <summary>
         /// Register a new module with device in the system
         /// </summary>
-        /// <param name="iotHubModule">The Module object to be registered.</param>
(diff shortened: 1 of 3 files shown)


iothub/service/src/RegistryManager.cs
```diff
@@ -68,161 +69,134 @@ public void Dispose()
         /// Register a new device with the system
         /// </summary>
         /// <param name="device">The Device object to be registered.</param>
-        /// <returns>echoes back the Device object with the generated keys and ETags</returns>
+        /// <returns>echoes back the Device object with the generated keys and ETags.</returns>
```
abhipsaMisra, line 72: nit: Uppercase for ~e~Echoes

iothub/service/src/RegistryManager.cs
```diff
@@ -68,161 +69,134 @@ public void Dispose()
...
         public abstract Task<Module> AddModuleAsync(Module iotHubModule, CancellationToken cancellationToken);
 
         /// <summary>
         /// Adds a Device with Twin information
         /// </summary>
-        /// <param name="device"></param>
-        /// <param name="twin"></param>
-        /// <returns></returns>
+        /// <param name="device">The device to add.</param>
+        /// <param name="twin">The twin information for the device to be added.</param>
```
abhipsaMisra, line 102: nit: "The twin information for the device ~to be~ being added"?

iothub/service/src/RegistryManager.cs
abhipsaMisra: or just: "The Device object with the generated keys and ETags"

iothub/service/src/RegistryManager.cs
```diff
@@ -68,161 +69,134 @@ public void Dispose()
...
+        /// <returns>The result of the add operation.</returns>
         public abstract Task<BulkRegistryOperationResult> AddDeviceWithTwinAsync(Device device, Twin twin);
```
abhipsaMisra, line 104: Weird that adding a single device returns a "Bulk"RegistryOperationResult, right?

iothub/service/src/RegistryManager.cs
```diff
@@ -68,161 +69,134 @@ public void Dispose()
...
 
         /// <summary>
         /// Register a list of new devices with the system
         /// </summary>
-        /// <param name="devices">
-        /// The Device objects to be registered.
-        /// </param>
-        /// <returns>returns a string array of error messages</returns>
+        /// <param name="devices">The Device objects to be registered.</param>
+        /// <returns>returns a string array of error messages.</returns>
```
abhipsaMisra, line 119: nit: case

iothub/service/src/RegistryManager.cs
barustum: Good point. I believe that's because even though we're adding a single device, the underlying operations puts that device in a list and does a Bulk device operation. And that returns a bulk result.

drwill-ms (approved):

PR #1722, opened 2020-12-21
Fixing CS0168

jamdavi:
Removed a couple of obsolete properties from our tests and also removed a set of tests for methods that have been set as obsolete.


e2e/test/iothub/FileUploadE2ETests.cs
```diff
@@ -23,14 +23,6 @@ public class FileUploadE2ETests : E2EMsTestBase
         private const int FileSizeSmall = 10 * 1024;
         private const int FileSizeBig = 5120 * 1024;
 
-        [LoggedTestMethod]
```
drwill-ms, line 26: Just because the SDK method is marked as obsolete, doesn't mean a customer won't still use it and we need to ensure it is tested.

We shouldn't remove these tests.

drwill-ms (changes requested):
This is one to suppress.

e2e/test/iothub/FileUploadE2ETests.cs
jamdavi: Yeah good call, I'll fix that.

drwill-ms: Also, please update the PR title because when I see updates for it, I have a small heart attack each time. :)

e2e/test/iothub/FileUploadE2ETests.cs
```diff
@@ -25,6 +26,7 @@ public class FileUploadE2ETests : E2EMsTestBase
 
         [LoggedTestMethod]
         [TestCategory("LongRunning")]
+        [Obsolete]
```
drwill-ms, line 29: I'd rather suppress with a reason for the benefit of future code readers.

jamdavi: The CS suppression will need to be inline with #pragmas. I opted for this as it keeps us from having to push/pop these warning checks on the preprocessor stack.

That being said, I'll add some comments around the methods.

e2e/test/iothub/FileUploadE2ETests.cs
```diff
@@ -49,17 +53,23 @@ public async Task FileUpload_GetFileUploadSasUri_Http_x509_NoFileTransportSettin
...
             await UploadFileAsync(Client.TransportType.Http1, bigFile).ConfigureAwait(false);
         }
 
         [LoggedTestMethod]
         [TestCategory("LongRunning")]
+        [Obsolete]
         public async Task FileUpload_X509_SmallFile_Http()
         {
             string smallFile = await GetTestFileNameAsync(FileSizeSmall).ConfigureAwait(false);
+            // UploadFileAsync is marked obsolete due to a call to UploadToBlobAsync being obsolete
```
drwill-ms, line 71: ...however that method must remain in the SDK to avoid breaking changes...

drwill-ms (approved):

abhipsaMisra (approved):

e2e/test/iothub/FileUploadE2ETests.cs
jamdavi: Title updated, heart attack avoided. :)

PR #1726, opened 2020-12-22
Fix regression on AsyncResult

azabbasi:
in a recent change, while removing dead code, we caused an unreachable code warning
https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0162

I have pointed out the state of the code before the regression in the comments.

### DIFF
--- a/iothub/service/src/Common/AsyncResult.cs
+++ b/iothub/service/src/Common/AsyncResult.cs
@@ -272,19 +272,19 @@ namespace Microsoft.Azure.Devices.Common
             }
 
             callback = null;
-            if (result.CompletedSynchronously)
+            if (!result.CompletedSynchronously)
             {
+                return false;
+            }
 #if NET451
+            else
+            {
                 if (ReferenceEquals(result, _deferredTransactionalResult))
                 {
                     _deferredTransactionalResult = null;
                 }
+            }
 #endif
-            }
-            else
-            {
-                return false;
-            }
 
             callback = GetNextCompletion();
             if (callback == null)
--- a/iothub/service/src/Common/AsyncResult.cs
+++ b/iothub/service/src/Common/AsyncResult.cs
@@ -281,6 +281,7 @@ namespace Microsoft.Azure.Devices.Common
             }
             else
 #endif
+            if (!result.CompletedSynchronously)
             {
                 return false;
             }
--- a/iothub/service/src/Common/AsyncResult.cs
+++ b/iothub/service/src/Common/AsyncResult.cs
@@ -272,19 +272,16 @@ namespace Microsoft.Azure.Devices.Common
             }
 
             callback = null;
-
-#if NET451
-            if (!result.CompletedSynchronously && ReferenceEquals(result, _deferredTransactionalResult))
+            if (result.CompletedSynchronously)
             {
-                // Use deferredTransactionalResult to see if we are supposed to execute a post-transaction callback.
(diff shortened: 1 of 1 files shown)


iothub/service/src/Common/AsyncResult.cs
```diff
@@ -281,6 +281,7 @@ private bool TryContinueHelper(IAsyncResult result, out AsyncCompletion callback
             }
             else
 #endif
+            if (!result.CompletedSynchronously)
```
azabbasi, line 284: if you expand the file to see the method `TryContinueHelper` you will notice that we will have an unreachable code for all frameworks other than NET451. 
This method used to look like this:

`        private bool TryContinueHelper(IAsyncResult result, out AsyncCompletion callback)
        {
            if (result == null)
            {
                throw Fx.Exception.AsError(new InvalidOperationException(CommonResources.InvalidNullAsyncResult));
            }

            callback = null;

            if (result.CompletedSynchronously)
            {
#if NET451
                // Once we pass the check, we know that we own forward progress, so transactionContext is correct. Verify its state.
                if (_transactionContext != null)
                {
                    if (_transactionContext.State != TransactionSignalState.Completed)
                    {
                        ThrowInvalidAsyncResult("Check/SyncContinue cannot be called from within the PrepareTransactionalCall using block.");
                    }
                    else if (_transactionContext.IsSignalled)
                    {
                        // This is most likely to happen when result.CompletedSynchronously registers differently here and in the callback, which
                        // is the fault of 'result'.
                        ThrowInvalidAsyncResult(result);
                    }
                }
#endif
            }
#if NET451
            else if (ReferenceEquals(result, _deferredTransactionalResult))
            {
                // The transactionContext may not be current if forward progress has been made via the callback. Instead,
                // use deferredTransactionalResult to see if we are supposed to execute a post-transaction callback.
                //
                // Once we pass the check, we know that we own forward progress, so transactionContext is correct. Verify its state.
                if (_transactionContext == null || !_transactionContext.IsSignalled)
                {
                    ThrowInvalidAsyncResult(result);
                }
                _deferredTransactionalResult = null;
            }
#endif
            else
            {
                return false;
            }

            callback = GetNextCompletion();
            if (callback == null)
            {
                ThrowInvalidAsyncResult("Only call Check/SyncContinue once per async operation (once per PrepareAsyncCompletion).");
            }
            return true;
        }`

azabbasi: As you can see, the `if (result.CompletedSynchronously)` statement was outside of the net framework condition so the `else` after `#endif` made sense. 
but now it will always return false for all frameworks other than 451 regardless.

drwill-ms: Good point. Do we need to worry about the `else if` clause too?

azabbasi: excellent question, I didn't think about that part. let me revert this code and re-do the whole method.

drwill-ms: Sorry for creating this problem in the first place.

azabbasi: no worries at all !

azabbasi: ok, with _transactionContext gone, I think I got parity with the original functionality.

drwill-ms (approved):

PR #1727, opened 2020-12-23
Code clean up of ClientWebSocketChannel files

drwill-ms:
More incremental changes leading up the the CA1001 fix, to track down what is causing test run failures in the pipeline.

These classes look awfully similar but are used in 2 different projects, and one is public (why? I don't know).

### DIFF
--- a/common/src/device/transport/mqtt/ClientWebSocketChannel.cs
+++ b/common/src/device/transport/mqtt/ClientWebSocketChannel.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Contracts;
 using System.Net;
 using System.Net.WebSockets;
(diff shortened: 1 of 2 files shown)


common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -13,24 +14,20 @@
 
 namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 {
-#pragma warning disable CA1001 // Types that own disposable fields should be disposable - WS is owned by the caller.
-
+    //[SuppressMessage("Design", "CA1001:Types that own disposable fields should be disposable", Justification = "WS is owned by the caller.")]
```
drwill-ms, line 17: I have this ready, but will disable it until actually dealing with `CancellationTokenSource`

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -13,24 +14,20 @@
...
     internal class ClientWebSocketChannel : AbstractChannel
-#pragma warning restore CA1001 // Types that own disposable fields should be disposable
     {
-        private readonly ClientWebSocket _webSocket;
-        private readonly CancellationTokenSource _writeCancellationTokenSource;
-        private bool _active;
-
-        internal bool ReadPending { get; set; }
```
drwill-ms, line 25: These weren't referenced outside this class

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -84,7 +75,67 @@ protected override void Flush0()
             }
         }
 
-        protected override bool IsCompatible(IEventLoop eventLoop) => true;
+        public ClientWebSocketChannel Option<T>(ChannelOption<T> option, T value)
```
drwill-ms, line 78: Moved up from below for publics above protected/internal/private

azabbasi (approved):

PR #1728, opened 2020-12-23
Make DPS' ClientWebSocketChannel disposable

drwill-ms:


provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -334,7 +333,7 @@ private static DeviceRegistrationResult ConvertToProvisioningRegistrationResult(
             var uriBuilder = new UriBuilder(WsScheme, message.GlobalDeviceEndpoint, Port);
             Uri websocketUri = uriBuilder.Uri;
 
-            using var websocket = new ClientWebSocket();
```
drwill-ms, line 337: ClientWebSocketChannel should own the lifetime of ClientWebSocket. It is weird to pass something to it, but dispose it while the passed object's lifetime has not ended.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -333,11 +344,6 @@ private int CompareTo(ClientWebSocketChannel other)
         private void Abort()
         {
             _webSocket?.Abort();
-            _webSocket?.Dispose();
```
drwill-ms, line 336: I moved dispose behavior to the Dispose method

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
```diff
@@ -19,108 +18,93 @@ public class ProvisioningCertificateValidationE2ETest : E2EMsTestBase
         [LoggedTestMethod]
         public async Task ProvisioningServiceClient_QueryInvalidServiceCertificateHttp_Fails()
         {
-            using (var provisioningServiceClient =
```
drwill-ms, line 22: Mostly this is switching to the simple `using` style.

provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -347,38 +336,32 @@ private static DeviceRegistrationResult ConvertToProvisioningRegistrationResult(
...
-                {
                     Logging.Error(this, $"{nameof(ProvisionOverWssUsingX509CertificateAsync)} PlatformNotSupportedException thrown as .NET Core 2.0 doesn't support proxy");
-                }
             }
 
             await websocket.ConnectAsync(websocketUri, cancellationToken).ConfigureAwait(false);
 
-            var clientChannel = new ClientWebSocketChannel(null, websocket);
-            clientChannel
+            _webSocketChannel = new ClientWebSocketChannel(null, websocket);
```
drwill-ms, line 363: Save off the instance so it can be disposed when `this` is disposed.

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -83,6 +82,15 @@ public ClientWebSocketChannel Option<T>(ChannelOption<T> option, T value)
             return this;
         }
 
+        public void Dispose()
+        {
+            _webSocket?.Dispose();
+            _webSocket = null;
```
vinagesh, line 88: Do we really need to set to null after dispose?

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
```diff
@@ -19,108 +18,93 @@ public class ProvisioningCertificateValidationE2ETest : E2EMsTestBase
...
-                ProvisioningServiceClient.CreateFromConnectionString(
-                    Configuration.Provisioning.ConnectionStringInvalidServiceCertificate))
-            {
-                Query q = provisioningServiceClient.CreateEnrollmentGroupQuery(
-                    new QuerySpecification("SELECT * FROM enrollmentGroups"));
+            using var provisioningServiceClient = ProvisioningServiceClient.CreateFromConnectionString(
+                Configuration.Provisioning.ConnectionStringInvalidServiceCertificate);
+            Query q = provisioningServiceClient.CreateEnrollmentGroupQuery(
```
vinagesh, line 23: I know this was already here but I think one letter variables are not the best choice. Rename q to query?

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
drwill-ms: Do we have to? No, not in this case. I do it as a best practice for any disposable object so we never accidentally try to dispose twice.

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
drwill-ms: I agree the variable name could be better. I'm not going too deep into this file clean up.

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -559,51 +559,50 @@ public async Task ProvisioningDeviceClient_ValidRegistrationId_MqttWsWithProxy_S
...
                 {
-                    try
-                    {
-                        await DeleteCreatedEnrollmentAsync(enrollmentType, provisioningServiceClient, security, groupId).ConfigureAwait(false);
-                    }
-                    catch (Exception ex)
-                    {
-                        Console.WriteLine($"Cleanup of enrollment failed due to {ex}");
-                    }
+                    Console.WriteLine($"Cleanup of enrollment failed due to {ex}");
```
vinagesh, line 605: This was also already there but curious why not Logger.Trace. We should probably update.

drwill-ms: I thought Logger.Trace was only for SDK code, not tests. In a test, Console.WriteLine goes to test output.

drwill-ms: Oh, is that the test logger?

vinagesh (approved):

Issue #1667, opened 2020-11-17
[Bug Report] NullreferenceException when accessing twin metadata retrieved via DeviceClient


AieatAssam:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64
- **Application's .NET Target Framework :** netcoreapp3.1
- **Device:** Laptop
- **SDK version used:** Microsoft.Azure.Devices.Client 1.33.1

## Description of the issue
Upon retrieving TwinCollection device-side (not registry-side) via GetTwinAsync(), an attempt to call any metadata-related methods generates a meaningless NullReferenceException. Examples of methods are **GetLastUpdated**(), **GetLastUpdatedVersion**(), **GetMetadata**().

## Code sample exhibiting the issue
```csharp
using var deviceClient = DeviceClient.CreateFromConnectionString("<SNIP>");
	
var twin = await deviceClient.GetTwinAsync();
var lastUpdated = twin.Properties.Reported.GetLastUpdated();
```

## Cause
A number of methods inside of TwinCollection.cs retrieve data from _metadata field without checking it for null.

## Suggestion
Since the workflow here is indeed flawed (metadata for twin is available via RegistryManager and not DeviceClient), the system should generate an error. However, allowing NullreferenceException is just plain wrong. Instead, null check should be done and a meaningfull exception thrown instead. Alternatively, the methods in question should return nullable types and null values when metadata is nto present.

bikamani:
@AieatAssam thanks for submitting this bug. I will discuss with rest of the team and get back on this soon.

bikamani:
@AieatAssam Since adding a different exception here would change the behavior and break customers. For now, we have added appropriate [documentation ](https://github.com/Azure/azure-iot-sdk-csharp/pull/1734/files)(xml comments) and next release will have this change.

AieatAssam:
Thank you. That is a very valid point. However, the xml documentation here does not actually explain why the metadata can be null (i.e. when used from the wrong side of the IoT Hub access - i.e. DeviceCLient rather than RegistryManager). Should this be added to the XML docs as well? I.e. highlight that it is not meant to be used from device side?

bikamani:
[...]

PR #1731, opened 2020-12-29
Fix for NRE when accessing twin metadata

bikamani:
This is a fix for github issue [1667](https://github.com/Azure/azure-iot-sdk-csharp/issues/1667#)

This change will change behavior when metadata is null. Prior to this you would see a NRE being thrown, going forward this will throw argument null exception. Added xml comments for the same.

### DIFF
--- a/shared/src/TwinCollection.cs
+++ b/shared/src/TwinCollection.cs
@@ -176,9 +176,9 @@ namespace Microsoft.Azure.Devices.Shared
         /// Gets the LastUpdated time for this property
         /// </summary>
         /// <returns>DateTime instance representing the LastUpdated time for this property</returns>
-        public DateTime GetLastUpdated()
+        public DateTime? GetLastUpdated()
         {
-            return (DateTime)_metadata[LastUpdatedName];
+            return (DateTime?)_metadata[LastUpdatedName];
         }
 
         /// <summary>
--- a/shared/src/TwinCollection.cs
+++ b/shared/src/TwinCollection.cs
@@ -176,7 +176,8 @@ namespace Microsoft.Azure.Devices.Shared
         /// Gets the LastUpdated time for this property
         /// </summary>
         /// <returns>DateTime instance representing the LastUpdated time for this property</returns>
-        /// <exception cref="System.ArgumentNullException">Thrown when _metadata is null</exception>
+        /// <exception cref="System.ArgumentNullException">Thrown when the TwinCollection metadata is null. 
+        /// An example would be when the TwinCollection class is created with the default constructor</exception>
         public DateTime GetLastUpdated()
         {
             return (DateTime)_metadata?[LastUpdatedName];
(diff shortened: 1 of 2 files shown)


shared/src/TwinCollection.cs
```diff
@@ -176,9 +176,9 @@ public Metadata GetMetadata()
         /// Gets the LastUpdated time for this property
         /// </summary>
         /// <returns>DateTime instance representing the LastUpdated time for this property</returns>
-        public DateTime GetLastUpdated()
+        public DateTime? GetLastUpdated()
         {
-            return (DateTime)_metadata[LastUpdatedName];
+            return (DateTime?)_metadata[LastUpdatedName];
```
jamdavi, line 181: DateTime isn't a nullable type and we should check for the property here rather than return a null. Changing this to return null could break more people who already depend on this property.

For example, if a customer is expecting a DateTIme object (which is never null), the NRE will be moved to their code.

```
using var deviceClient = DeviceClient.CreateFromConnectionString("<SNIP>");
	
var twin = await deviceClient.GetTwinAsync();

// Assume this normally has a value and we check for it...
var lastUpdated = twin.Properties.Reported.GetLastUpdated();

if (lastUpdate > DateTime.Now.AddHours(-1))
{
    // Do something...
}

```

jamdavi (changes requested):

jamdavi (approved):
Approved. We should note in code or in the document comments about this returning an ArgumentNullException when the metadata has not been initialized instead of a NullReferenceException.

shared/src/TwinCollection.cs
```diff
@@ -176,6 +176,7 @@ public Metadata GetMetadata()
         /// Gets the LastUpdated time for this property
         /// </summary>
         /// <returns>DateTime instance representing the LastUpdated time for this property</returns>
+        /// <exception cref="System.ArgumentNullException">Thrown when _metadata is null</exception>
```
jamdavi, line 179: _metadata is a private field that customers won't see. I'd put something to the effect of "Thrown when the TwinCollection metadata is null. An example would be when the TwinCollection class is created with the default constructor."

shared/src/TwinCollection.cs
```diff
@@ -176,9 +176,11 @@ public Metadata GetMetadata()
         /// Gets the LastUpdated time for this property
         /// </summary>
         /// <returns>DateTime instance representing the LastUpdated time for this property</returns>
+        /// <exception cref="System.ArgumentNullException">Thrown when the TwinCollection metadata is null.
```
drwill-ms, line 179: I don't see this exception being thrown anywhere in this method.

drwill-ms: IOW, should we remove the doc comment that says this could return this exception?

bikamani: This exception gets thrown if metadata is null and someone calls this method so the comment is still valid.

For example a customer calls default TwinCollection constructor with no parameters, metadata would be null and then calling this method would return this exception

drwill-ms: The code as I see it returns null, not an exception.

drwill-ms: I see, it happens because casting a null to DateTime throws that exception.

Changing the exception thrown is also a breaking change. We can't do that.

PR #1734, opened 2021-01-04
Revert change for GetLastUpdated

bikamani:
We do not want to change behavior of this API hence reverting change and added xml comments.

### DIFF
--- a/shared/src/TwinCollection.cs
+++ b/shared/src/TwinCollection.cs
@@ -28,6 +28,7 @@ namespace Microsoft.Azure.Devices.Shared
 
         /// <summary>
         /// Creates instance of <see cref="TwinCollection"/>.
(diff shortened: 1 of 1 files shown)


drwill-ms (approved):

shared/src/TwinCollection.cs
```diff
@@ -176,11 +177,11 @@ public Metadata GetMetadata()
         /// Gets the LastUpdated time for this property
         /// </summary>
         /// <returns>DateTime instance representing the LastUpdated time for this property</returns>
-        /// <exception cref="System.ArgumentNullException">Thrown when the TwinCollection metadata is null. 
+        /// <exception cref="System.NullReferenceException">Thrown when the TwinCollection metadata is null.
```
bikamani, line 180: Reverting this change so no more behavior changes. It will still throw nre as before, we are documenting it.

PR #1736, opened 2021-01-06
Fix issue where retrieving a twin array from registry manager returned null

abhipsaMisra:
On retrieving a twin property by index/ key, our sdk will try to return it as a `JObject` within a `TwinCollection`. For twin arrays, since they are represented as `JArray`, they cannot be converted to a `JObject` (`JArray` is not inherited from `JObject`), and are thus returned as `null`. 
This PR adds a new `TwinCollectionArray` type, one that is inherited from a `JArray`, and returns that instead`.

Note: this only affects the registry manager retuned twin. The device-client returned twin is always returned as the base `JToken` type: https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/twinArrayNull/shared/src/TwinCollection.cs#L255

### DIFF
--- a/shared/src/TwinCollectionArray.cs
+++ b/shared/src/TwinCollectionArray.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.Devices.Shared
         internal TwinCollectionArray(JArray jArray, JObject metadata)
             : base(jArray)
         {
-            _metadata = metadata;
+            _metadata = metadata ?? throw new ArgumentNullException(nameof(metadata));
         }
 
         /// <summary>
--- a/shared/src/TwinCollectionArray.cs
+++ b/shared/src/TwinCollectionArray.cs
@@ -11,10 +11,6 @@ namespace Microsoft.Azure.Devices.Shared
     /// </summary>
     public class TwinCollectionArray : JArray
     {
-        private const string MetadataName = "$metadata";
-        private const string LastUpdatedName = "$lastUpdated";
-        private const string LastUpdatedVersionName = "$lastUpdatedVersion";
-
         private readonly JObject _metadata;
 
         internal TwinCollectionArray(JArray jArray, JObject metadata)
@@ -27,27 +23,27 @@ namespace Microsoft.Azure.Devices.Shared
         /// Gets the value for the given property name
         /// </summary>
         /// <param name="propertyName">Property Name to lookup</param>
-        /// <returns>Value if present</returns>
+        /// <returns>Property value, if present</returns>
         public dynamic this[string propertyName]
         {
             get
             {
-                if (propertyName == MetadataName)
+                if (propertyName == TwinCollection.MetadataName)
                 {
                     return GetMetadata();
                 }
 
-                if (propertyName == LastUpdatedName)
+                if (propertyName == TwinCollection.LastUpdatedName)
                 {
                     return GetLastUpdated();
                 }
 
-                if (propertyName == LastUpdatedVersionName)
+                if (propertyName == TwinCollection.LastUpdatedVersionName)
                 {
                     return GetLastUpdatedVersion();
                 }
 
-                throw new ArgumentException($"'Newtonsoft.Json.Linq.JArray' does not contain a definition for '{propertyName}'.");
+                throw new ArgumentException($"{nameof(TwinCollectionArray)} does not contain a definition for '{propertyName}'.");
             }
         }
 
@@ -66,7 +62,7 @@ namespace Microsoft.Azure.Devices.Shared
         /// <returns>DateTime instance representing the LastUpdated time for this property</returns>
         public DateTime GetLastUpdated()
         {
-            return (DateTime)_metadata[LastUpdatedName];
+            return (DateTime)_metadata[TwinCollection.LastUpdatedName];
         }
 
         /// <summary>
@@ -75,7 +71,7 @@ namespace Microsoft.Azure.Devices.Shared
         /// <returns>LastUpdatdVersion if present, null otherwise</returns>
         public long? GetLastUpdatedVersion()
         {
-            return (long?)_metadata[LastUpdatedVersionName];
+            return (long?)_metadata[TwinCollection.LastUpdatedVersionName];
         }
     }
 }
--- a/shared/src/TwinCollectionArray.cs
+++ b/shared/src/TwinCollectionArray.cs
@@ -7,7 +7,7 @@ using Newtonsoft.Json.Linq;
 namespace Microsoft.Azure.Devices.Shared
 {
     /// <summary>
-    /// Represents a property array in <see cref="TwinCollection"/>
+    /// Represents a property array in a <see cref="TwinCollection"/>
     /// </summary>
     public class TwinCollectionArray : JArray
     {
--- a/shared/src/TwinCollectionArray.cs
+++ b/shared/src/TwinCollectionArray.cs
@@ -1,0 +1,81 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// Represents a property array in <see cref="TwinCollection"/>
+    /// </summary>
+    public class TwinCollectionArray : JArray
+    {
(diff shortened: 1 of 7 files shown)


shared/src/TwinCollectionArray.cs
```diff
@@ -0,0 +1,81 @@
...
+            : base(jArray)
+        {
+            _metadata = metadata;
+        }
+
+        /// <summary>
+        /// Gets the value for the given property name
+        /// </summary>
+        /// <param name="propertyName">Property Name to lookup</param>
+        /// <returns>Value if present</returns>
```
azabbasi, line 30: nit: `Property value if present`

shared/src/TwinCollectionArray.cs
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    return GetLastUpdated();
+                }
+
+                if (propertyName == LastUpdatedVersionName)
+                {
+                    return GetLastUpdatedVersion();
+                }
+
+                throw new ArgumentException($"'Newtonsoft.Json.Linq.JArray' does not contain a definition for '{propertyName}'.");
```
azabbasi, line 50: is it really the Newtonsoft object that doesn't have a definition for the property? or the twin object?

shared/src/TwinCollectionArray.cs
drwill-ms: And add a comma before "if present"

shared/src/TwinCollectionArray.cs
```diff
@@ -0,0 +1,81 @@
...
+        private const string MetadataName = "$metadata";
+        private const string LastUpdatedName = "$lastUpdated";
+        private const string LastUpdatedVersionName = "$lastUpdatedVersion";
+
+        private readonly JObject _metadata;
+
+        internal TwinCollectionArray(JArray jArray, JObject metadata)
+            : base(jArray)
+        {
+            _metadata = metadata;
```
drwill-ms, line 23: null check?

shared/src/TwinCollectionArray.cs
```diff
@@ -0,0 +1,81 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// Represents a property array in <see cref="TwinCollection"/>
+    /// </summary>
+    public class TwinCollectionArray : JArray
+    {
+        private const string MetadataName = "$metadata";
```
drwill-ms, line 14: Can you find a way to not have to duplicate these constants across files?

shared/src/TwinCollectionValue.cs
```diff
@@ -17,9 +17,9 @@ namespace Microsoft.Azure.Devices.Shared
         Justification = "Uses default JValue comparison, equality and hashing implementations.")]
     public class TwinCollectionValue : JValue
     {
-        const string metadataName = "$metadata";
-        const string lastUpdatedName = "$lastUpdated";
-        const string lastUpdatedVersionName = "$lastUpdatedVersion";
+        private const string metadataName = "$metadata";
```
drwill-ms, line 20: Field casing is wrong for constants

drwill-ms (changes requested):
Waiting on responses to code comments.

shared/src/TwinCollectionArray.cs
abhipsaMisra: will update.

shared/src/TwinCollectionArray.cs
abhipsaMisra: yeah, it should be ~TwinCollectionArray` actually, i'll update it

shared/src/TwinCollectionArray.cs
abhipsaMisra: metadata is a JObject, which will default to null anyway. Is there an advantage to adding an explicit null check here?

shared/src/TwinCollectionArray.cs
abhipsaMisra: sure, will do

shared/src/TwinCollectionArray.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                throw new ArgumentException($"{nameof(TwinCollectionArray)} does not contain a definition for '{propertyName}'.");
+            }
+        }
+
+        /// <summary>
+        /// Gets the Metadata for this property
+        /// </summary>
+        /// <returns>Metadata instance representing the metadata for this property</returns>
+        public Metadata GetMetadata()
```
abhipsaMisra, line 54: this, and below: they are common for both TwinCollectionArray and TwinCollectionValue; but with multiple inheritance not supported in c#, I've kept them repeated. I could define an interface with implementation, and use that, but it seemed too much of work for this.
What do you think?

shared/src/TwinCollectionArray.cs
drwill-ms: There is no guarding against NRE later in this file.

So either it can be null, in which case we we should conditionally derereference it, or we should check for null and throw.

shared/src/TwinCollectionArray.cs
```diff
@@ -0,0 +1,77 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// Represents a property array in <see cref="TwinCollection"/>
```
drwill-ms, line 10: in a

shared/src/TwinCollectionArray.cs
drwill-ms: I don't feel too much code is duplicated. If you want to pursue a common base class, that'd be fine, but I don't expect it.

abhipsaMisra: I am going to leave it as-is. It isn't a whole lot of repeated code and I don't expect it to grow in the future either.

shared/src/TwinCollectionArray.cs
abhipsaMisra: I see, I can throw an ArgumentNullException during initialization, sure. The use-case does require the metadata to be available (and this is an internal constructor), so makes sense.

azabbasi (approved):

shared/src/TwinCollectionArray.cs
```diff
@@ -0,0 +1,77 @@
...
+            return new Metadata(GetLastUpdated(), GetLastUpdatedVersion());
+        }
+
+        /// <summary>
+        /// Gets the LastUpdated time for this property
+        /// </summary>
+        /// <returns>DateTime instance representing the LastUpdated time for this property</returns>
+        public DateTime GetLastUpdated()
+        {
+            return (DateTime)_metadata[TwinCollection.LastUpdatedName];
```
drwill-ms, line 65: I think the return value needs to change to `DateTime?` and the return statement to `return (DateTime?)_metadata?[TwinCollection.LastUpdatedName];`

shared/src/TwinCollectionArray.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        /// <summary>
+        /// Gets the LastUpdatedVersion for this property
+        /// </summary>
+        /// <returns>LastUpdatdVersion if present, null otherwise</returns>
+        public long? GetLastUpdatedVersion()
+        {
+            return (long?)_metadata[TwinCollection.LastUpdatedVersionName];
```
drwill-ms, line 74: should be `return (long?)_metadata?[TwinCollection.LastUpdatedVersionName];`

shared/src/TwinCollectionArray.cs
drwill-ms: Would you add a couple of unit tests to validate that this methods down throw when initialized with null?

drwill-ms (approved):

barustum (approved):

ThomasPe:
any ETA when this fix will be released to nuget?

abhipsaMisra:
We are in the process of finalizing our next release. We should have it out by end of next week.
Does this issue currently block you, @ThomasPe ?

ThomasPe:
Indeed it does, the earlier the better. Appreciate your efforts 👍

abhipsaMisra:
Sure thing, I'll keep you updated!

Issue #1737, opened 2021-01-06
MqttIotHubAdapter can throw unobserved NullReferenceException when reconnecting


ccruden-shookiot:
# Context

- **OS, version, SKU and CPU architecture used:** Windows Server 2016 x64
- **Application's .NET Target Framework :** .NET framework 4.8
- **Device:** Cloud hosted VM
- **SDK version used:** Microsoft.Azure.Devices.Client 1.31.1

## Description of the issue
Our application is a service hosted on a cloud VM that continually writes data to an MQTT IOT hub. We use an SAS token to authenticate the connection, so the connection is dropped and refreshed every 65 minutes. It appears that when that refresh happens, if something goes wrong, MqttIotHubAdapter.ShutdownOnError (now ShutdownOnErrorAsync) can handle an exception which results in a NullReferenceException being thrown on a task / thread which is unobserved, and so bubbles up to the top level and causes the application itself to exit.

Event viewer provided the following information:
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.NullReferenceException
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.OrderedTwoPhaseWorkQueue`2[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089],[System.__Canon, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]].Abort(System.Exception)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter+<ShutdownOnError>d__69.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()

[...]

PR #1738, opened 2021-01-07
Code cleanup in OrderedTwoPhaseWorkQueue

drwill-ms:
Initial look at code referenced in (#1737). Cleaning up code, and providing one possible NRE protection.

Also added parameter checking to provide more info in the even of an NRE. This is an internal class, and when it does throw an NRE the customer is helpless so better to provide more info even if it is a slight behavior change - I don't believe it is a meaningful behavior change that would affect a device app that could continue to run (it can't).

### DIFF
--- a/iothub/device/src/Transport/Mqtt/ClientWebSocketChannel.cs
+++ b/iothub/device/src/Transport/Mqtt/ClientWebSocketChannel.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             _writeCancellationTokenSource = null;
         }
 
-        protected class WebSocketChannelUnsafe : AbstractUnsafe
+        private class WebSocketChannelUnsafe : AbstractUnsafe
         {
             public WebSocketChannelUnsafe(AbstractChannel channel)
                 : base(channel)
--- a/iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
+++ b/iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using DotNetty.Transport.Channels;
 using Microsoft.Azure.Devices.Client.Exceptions;
@@ -15,8 +16,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         {
             public IncompleteWorkItem(TWorkId id, TWork workItem)
             {
-                this.WorkItem = workItem;
-                this.Id = id;
+                WorkItem = workItem ?? throw new ArgumentNullException(nameof(workItem));
+                Id = id ?? throw new ArgumentNullException(nameof(id));
             }
 
             public TWork WorkItem { get; }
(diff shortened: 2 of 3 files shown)


iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -15,71 +16,79 @@ private class IncompleteWorkItem
...
+
+            if (stateBefore != State
+                && State == States.Aborted)
             {
-                Queue<IncompleteWorkItem> queue = this.incompleteQueue;
-                while (queue.Count > 0)
+                while (_incompleteQueue.Any())
                 {
-                    TWork workItem = queue.Dequeue().WorkItem;
+                    var workItem = _incompleteQueue.Dequeue()?.WorkItem as ICancellable;
```
drwill-ms, line 83: If the event more than one thread was aborting (despite the protection supposedly offered in the lines above), even if we thought there was work in the queue and then tried to dequeue it, we might as well protect against an empty queue.

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -15,71 +16,79 @@ private class IncompleteWorkItem
...
-            if (this.incompleteQueue.Count == 0)
+            if (!_incompleteQueue.Any())
             {
                 throw new IotHubException("Nothing to complete.", isTransient: false);
             }
-            IncompleteWorkItem incompleteWorkItem = this.incompleteQueue.Peek();
-            if (incompleteWorkItem.Id.Equals(workId))
+
+            IncompleteWorkItem incompleteWorkItem = _incompleteQueue.Peek();
+            if (incompleteWorkItem != null
```
drwill-ms, line 50: No guarantee that `Peek` returns an item, right?

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -15,71 +16,79 @@ private class IncompleteWorkItem
...
+                && incompleteWorkItem.Id.Equals(workId))
             {
-                this.incompleteQueue.Dequeue();
-                return this.completeWork(context, incompleteWorkItem.WorkItem);
+                _incompleteQueue.Dequeue();
```
drwill-ms, line 53: I don't love that we peek an item, validate it, and then assume when we call dequeue we get the same item. Perhaps that is a safe assumption, but I'd prefer our code not make such assumptions. I didn't change this, though - just noting I don't care for it.

Ideas?

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
jamdavi: From the docs it will throw an exception if the queue is empty. But it will return what ever is there.

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1.peek?view=netcore-3.1

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
jamdavi: Either we lock around this and make it a transaction of sorts, or we dequeue, check and requeue. So that means either we steal work or we steal threads.

iothub/device/src/Transport/Mqtt/ClientWebSocketChannel.cs
```diff
@@ -65,7 +65,7 @@ public void Dispose()
             _writeCancellationTokenSource = null;
         }
 
-        protected class WebSocketChannelUnsafe : AbstractUnsafe
+        private class WebSocketChannelUnsafe : AbstractUnsafe
```
drwill-ms, line 68: When building I saw a warning here that the parent class is sealed, so this type being protected made no sense.

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
drwill-ms: I'd like to understand the larger context around this class before making too many meaningful changes.

jamdavi: Yeah, agreed. I can see a few race conditions in this code but we'd need to sit down to see what the outcome of such a thing is

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -15,71 +16,78 @@ private class IncompleteWorkItem
...
-            this.getWorkId = getWorkId;
-            this.completeWork = completeWork;
+            _getWorkId = getWorkId ?? throw new ArgumentNullException(nameof(getWorkId));
+            _completeWorkAsync = completeWorkAsync ?? throw new ArgumentNullException(nameof(completeWorkAsync));
         }
 
         public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
         {
-            if (this.incompleteQueue.Count == 0)
+            if (!_incompleteQueue.Any())
```
drwill-ms, line 44: Okay, so `Peek` and `Dequeue` will throw a `InvalidOperationException` if the queue is empty, not return null as I expected. So that can't be a source of the customer-reported NRE.

However, we have this check about there being items in the queue before we peek or dequeue, so unless multiple threads are working against it, there won't be a race condition or a chance at that exception.

I toyed with adding try/catch around it to guard against that, but as it isn't the exception the customer saw, I don't see a reason to add that protection. Again, more insight into the greater context of this class should be known before we muck with it too much.

jamdavi: I thought I put the InvalidOp exception in one of our conversations. Anyway, it's correct that there won't be a race if it's single threaded.

jamdavi (approved):
This looks good, I think this should avoid NREs when Abort() is called as it shows in the issue.

PR #1740, opened 2021-01-08
Add deprecated API back to avoid breaking changes.

azabbasi:
Based on a conversation in the release PR here: 
https://github.com/Azure/iot-sdks-internals/pull/220

Bringing back the old API and mark it as deprecated.

### DIFF
--- a/e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
+++ b/e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
@@ -235,7 +235,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                 return _tokenRefreshSemaphore.WaitAsync(cancellationToken);
             }
 
-            protected override Task<string> SafeCreateNewTokenAsync(string iotHub, int suggestedTimeToLive)
+            protected override Task<string> SafeCreateNewToken(string iotHub, int suggestedTimeToLive)
             {
                 _logger.Trace($"[{DateTime.UtcNow}] Refresher: Creating new token.");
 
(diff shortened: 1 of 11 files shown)


timtay-microsoft (approved):

e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
```diff
@@ -272,6 +272,12 @@ protected override Task<string> SafeCreateNewTokenAsync(string iotHub, int sugge
 
                 return Task.FromResult(token);
             }
+
+            [Obsolete("This method has been deprecated due to lack of the asynchronous suffix in the method name. Please use SafeCreateNewTokenAsync instead.")]
+            protected override Task<string> SafeCreateNewToken(string iotHub, int suggestedTimeToLive)
```
abhipsaMisra, line 277: we don't really need this here, do we? this is a test class :)

abhipsaMisra (approved):

e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
azabbasi: it's implementing the same abstract class so it has to implement this method as well

abhipsaMisra: Oh, that is right. This is definitely not going to give a good experience to our users - we're going to ask them to mandatorily implement an obsolete API.
What do you think about changing this obsolete method to a `virtual` method (with the default implementation calling `SafeCreateNewTokenAsync`? That way, people who've already implemented the method will still have their code work, while new users will not be forced to implement it.

azabbasi: to complicate things even further, people who implement this abstract class will be broken as we are adding a new abstract method to it. what is our stance regarding that type of situation?

azabbasi: We decided that we will revert this change to add the Async suffix to a public facing API as there is no straight forward way to not break customers.

PR #1742, opened 2021-01-13
feat(device client): Install the device chain certificates using the SDK

vinagesh:
## Checklist

1) Added option in X509Auth for users to pass in the chain information as well
2) Added code to install the chain certs if specified
3) Added integration tests for this scenario
4) Updated script and pipeline to setup to run these tests
5) Upgraded ARM template to bicep files for easy maintenance
6) TODO: Add a sample (Next PR)


iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
```diff
@@ -40,6 +53,13 @@ public string DeviceId
         /// </summary>
         public X509Certificate2 Certificate { get; set; }
 
+        /// <summary>
+        /// Full chain of certificates from the one used to sign the device certificate to the one uploaded to the service. Private keys are not required for these certificates.
+        /// This is only supported on AMQP_Tcp_Only and Mqtt_Tcp_Only
+        /// See <see href="TODO (Sample ref) - Sindhu"/>
```
vinagesh, line 59: Next PR

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
```diff
@@ -19,10 +19,12 @@ public sealed class DeviceAuthenticationWithX509Certificate : IAuthenticationMet
         /// </summary>
         /// <param name="deviceId">Device Identifier.</param>
         /// <param name="certificate">X.509 Certificate.</param>
-        public DeviceAuthenticationWithX509Certificate(string deviceId, X509Certificate2 certificate)
+        /// <param name="chainCertificates">Certificates in the device certificate chain.</param>
+        public DeviceAuthenticationWithX509Certificate(string deviceId, X509Certificate2 certificate, X509Certificate2Collection chainCertificates = null)
```
vinagesh, line 23: Another option is to add a new constructor. I chose to add it as it felt like less code and more discoverable. I'm open to suggestions.

e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -141,6 +143,77 @@ public async Task X509_Enable_CertificateRevocationCheck_Amqp_WebSocket()
             await SendMessageTest(transportSetting).ConfigureAwait(false);
         }
 
+        [LoggedTestMethod]
+        public async Task X509_Cert_Chain_Install_Test_MQTT()
```
abhipsaMisra, line 147: small nit: you might want to rename the test method to say "mqtt_tcp", since "mqtt" by itself is usually interpreted in our test project to be "tcp with fallback to ws".

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -393,6 +418,51 @@ $systemIdentityPrincipal = az resource list -n $iotHubName --query [0].identity.
...
+    Export-Certificate -cert $verificationCert -filePath $verificationCertPath -Type Cert | Out-Null
+    $etag = az iot hub certificate show -g $ResourceGroup --hub-name $iotHubName --name $hubUploadCertificateName --query 'etag'
+    az iot hub certificate verify -g $ResourceGroup --hub-name $iotHubName --name $hubUploadCertificateName -e $etag --path $verificationCertPath --output none
+}
+
+##################################################################################################################################
+# Create device in IoTHub that uses a certificate signed by intermediate certificate
+##################################################################################################################################
+
+$iotHubCertChainDevice = az iot hub device-identity list -g vinageshe2etests --hub-name vinageshe2etests-hub --query "[?deviceId=='$iotHubCertChainDeviceCommonName'].deviceId" --output tsv
```
abhipsaMisra, line 458: I think this is leftover test code?

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -393,6 +418,51 @@ $systemIdentityPrincipal = az resource list -n $iotHubName --query [0].identity.
...
+        "-Signer"                        = $rootCACert;
+    }
+    $verificationCert = New-SelfSignedCertificate @verificationCertArgs
+    Export-Certificate -cert $verificationCert -filePath $verificationCertPath -Type Cert | Out-Null
+    $etag = az iot hub certificate show -g $ResourceGroup --hub-name $iotHubName --name $hubUploadCertificateName --query 'etag'
+    az iot hub certificate verify -g $ResourceGroup --hub-name $iotHubName --name $hubUploadCertificateName -e $etag --path $verificationCertPath --output none
+}
+
+##################################################################################################################################
+# Create device in IoTHub that uses a certificate signed by intermediate certificate
```
abhipsaMisra, line 455: is this done on-the-fly by the tests (as required), or do we need to create a device identity in advance?

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        private static readonly object s_certOperationsLock = new object();
+
+        /// <summary>
+        /// Ensures the specified certs (presumably in a chain) are installed in the cert store.
+        /// </summary>
+        /// <remarks>
+        /// Because Intermediate Authorities may have been issued by the uploaded CA, the application must present the full chain of
+        /// certificates from the one used during authentication to the one uploaded to the service.
+        /// See <see href="TODO (Sample ref) - Sindhu"/>
```
abhipsaMisra, line 20: nit: we might want to remove this before checking it in :)

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,70 @@
...
+                    }
+                }
+                catch (Exception ex)
+                {
+                    if (Logging.IsEnabled) Logging.Error(null, $"{nameof(CertificateInstaller)} failed to read or write to cert store due to: {ex}");
+                }
+                finally
+                {
+#if !NET451
+                    store?.Dispose();
```
abhipsaMisra, line 64: q - how is the store disposed for net451? is it when the process ends?

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
abhipsaMisra: so: 
for self signed certs -> pass only certificate
for certs with a single parent -> pass only certificate = device cert, and upload the signing parent cert on hub
for certs with multiple certs in their hierarchy -> certificate = device cert, chainCertificates = chain from device cert to the point in the hierarchy whichever is upload to iot hub.

is this correct?

abhipsaMisra: hehe, looks like you've answered this right below!

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
```diff
@@ -40,6 +42,13 @@ public string DeviceId
         /// </summary>
         public X509Certificate2 Certificate { get; set; }
 
+        /// <summary>
+        /// Full chain of certificates from the one used to sign the device certificate to the one uploaded to the service. Private keys are not required for these certificates.
```
abhipsaMisra, line 46: 👍🏽

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
```diff
@@ -40,6 +42,13 @@ public string DeviceId
...
+        /// This is only supported on AMQP_Tcp_Only and Mqtt_Tcp_Only
```
abhipsaMisra, line 47: q - what was the restriction for ws? was it that the .net ws implementation wasn't forwarding the chain certificate info to the socket layer?

iothub/device/src/ClientFactory.cs
```diff
@@ -79,6 +79,14 @@ public static InternalClient Create(string hostname, string gatewayHostname, IAu
                 throw new ArgumentNullException(nameof(authenticationMethod));
             }
 
+            if (transportType != TransportType.Amqp_Tcp_Only
+                && transportType != TransportType.Mqtt_Tcp_Only
+                && authenticationMethod is DeviceAuthenticationWithX509Certificate
+                && ((DeviceAuthenticationWithX509Certificate)authenticationMethod).ChainCertificates != null)
+            {
+                throw new ArgumentException("Certificate chains are only supported on Amqp_Tcp_Only and Mqtt_Tcp_Only");
```
azabbasi, line 87: so certificate chains are not supported in HTTPS and websockets? is that a hub restriction ?

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// certificates from the one used during authentication to the one uploaded to the service.
+        /// See <see href="TODO (Sample ref) - Sindhu"/>
+        /// for more information.
+        /// </remarks>
+        /// <param name="certificates">The certificate chain to ensure is installed.</param>
+        public static void EnsureChainIsInstalled(X509Certificate2Collection certificates)
+        {
+            if (certificates == null || certificates.Count == 0)
+            {
+                if (Logging.IsEnabled) Logging.Info(null, $"{nameof(CertificateInstaller)} parameter 'certificates' was null or empty.");
```
azabbasi, line 28: nit: use `nameof(certificates)` instead

azabbasi: so should we just peacefully return here? no exceptions? do we expect this to be called with a null value?

iothub/device/src/Common/Security/CertificateInstaller.cs
azabbasi: it's not disposable in .net451

azabbasi: so it'll get garbage collected.

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,70 @@
...
+                        {
+                            if (Logging.IsEnabled) Logging.Info(null, $"{nameof(CertificateInstaller)} adding cert with thumbprint {certificate.Thumbprint} to X509 store.");
+
+                            store.Add(certificate);
+                        }
+                    }
+                }
+                catch (Exception ex)
+                {
+                    if (Logging.IsEnabled) Logging.Error(null, $"{nameof(CertificateInstaller)} failed to read or write to cert store due to: {ex}");
```
azabbasi, line 59: we are catching the exception and swallowing it, isn't that considered a broken scenario? do we just continue?

e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
vinagesh: I see. Will update.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
vinagesh: I did not add this in test as we run the tests in parallel and the device certificate is for a specific device only. Our tests can become flake if all tests try to create the same device.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
vinagesh: Answered above.

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: I thought it was ok to leave it and update it after my next PR. Not sure if it is ok to leave it or not. What does everyone follow?

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: yupp what Azad said.

vinagesh: It is unfortunate though as code looks bad :(

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
vinagesh: No the socket layer never forwards it. It somehow does not work for ws. I did not dive deep as that code is out of our control and no customer was asking for it in particular.

iothub/device/src/ClientFactory.cs
vinagesh: It was surely not working in websockets. I did not see HTTPS used for our device client. It seemed like it was only for file uploads so I did not test that a lot. If we use HTTPS for device client, let me know and I can do more testing and see if I can include in the next PR.

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: The caller ensures this is not called when null. However we can choose to throw an exception too but I wanted to keep the code same in DPS and DeviceClient so it is easy to maintain if we make changes. It would have been good to have it common but this file is already part of DPS so I just copied it here as well.

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: I am sure the auth will fail going forward. But good point. I was trying to keep the same code between DPS and DeviceClient but this might be worth looking into. I am not sure why we swallow it in the DPS scenario.

azabbasi: however, getting unauthorize without understanding why the provided certs didn't work is confusing, I'd rather expose the exception here or throw and catch it in the caller and throw a meaningful exception.

iothub/device/src/Common/Security/CertificateInstaller.cs
azabbasi: so with that logic this whole `if` statement is unnecessary. 
I would argue that this class should be self contained. I think an exception should be thrown here. 
If we want to reuse the same code, we can keep it in shared.

iothub/device/src/ClientFactory.cs
azabbasi: @abhipsaMisra  to comment.

iothub/device/src/Common/Security/CertificateInstaller.cs
azabbasi: I would say if your next PR follows this one immediately, it's ok to leave it here.

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: But we cannot move it out of our existing DPS package so we will have two copies anyway. I am in for updating this file here though. I am inclined towards not having the if statement here. This is a helper so all it should do is install if there is something in the list. The caller should take care of exception of not calling when null.

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: I think catching it in the caller is a good idea. I will do that.

iothub/device/src/ClientFactory.cs
abhipsaMisra: Yes, we do use https for a small subset of device client operations - send d2c message, receive c2d message and file uploads.

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: But because this is a helper, it should not really throw exceptions. I'll see if I can do something in the caller. The best would be a generic catch only.

iothub/device/src/ClientFactory.cs
abhipsaMisra: Methods and twin operations are not supported over https since these operations require a persistent connection to receive service initiated requests (for methods, twin updates etc), and https does not maintain persistent connections.
Technically we could have updated reported properties and fetched the twin propeties over https, but I guess the service team felt that the effort was not worth it.

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// Ensures the specified certs (presumably in a chain) are installed in the cert store.
+        /// </summary>
+        /// <remarks>
+        /// Because Intermediate Authorities may have been issued by the uploaded CA, the application must present the full chain of
+        /// certificates from the one used during authentication to the one uploaded to the service.
+        /// See <see href="TODO (Sample ref) - Sindhu"/>
+        /// for more information.
+        /// </remarks>
+        /// <param name="certificates">The certificate chain to ensure is installed.</param>
+        public static void EnsureChainIsInstalled(X509Certificate2Collection certificates)
```
barustum, line 24: probably change access modifier to internal, since the class itself is internal

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            if (certificates == null || certificates.Count == 0)
```
barustum, line 26: does X509Certificate2Collection have a Any method? If so, more performant than Count.

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    store.Open(OpenFlags.ReadWrite);
+
+                    foreach (X509Certificate2 certificate in certificates)
+                    {
+                        X509Certificate2Collection results = store.Certificates.Find(
+                            X509FindType.FindByThumbprint,
+                            certificate.Thumbprint,
+                            false);
+                        if (results.Count == 0)
```
barustum, line 49: .Any if supported.

iothub/device/src/ClientFactory.cs
vinagesh: I will see what can be done with HTTPS after I complete this and a sample. If it is possible, I'll add support for it.

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: No it does not have one.

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: Not support with the X509Collection

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
```
bikamani, line 4: xml comments

barustum (approved):

iothub/device/src/Common/Security/CertificateInstaller.cs
```diff
@@ -0,0 +1,59 @@
...
+                        if (results.Count == 0)
+                        {
+                            if (Logging.IsEnabled) Logging.Info(null, $"{nameof(CertificateInstaller)} adding cert with thumbprint {certificate.Thumbprint} to X509 store.");
+
+                            store.Add(certificate);
+                        }
+                    }
+                }
+#if !NET451
+                store?.Dispose();
```
abhipsaMisra, line 54: this could be in a finally block

abhipsaMisra (approved):

iothub/device/src/Common/Security/CertificateInstaller.cs
vinagesh: There is no try anymore.

vinagesh: got it, fixed.

azabbasi (approved):

PR #1746, opened 2021-01-15
docs: Fix a few CS1591 warnings

barustum:


iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -11,17 +11,29 @@
...
+        /// <summary>
+        /// Initializes a new instance of <see cref="SharedAccessSignatureBuilder"/> class
+        /// </summary>
         public SharedAccessSignatureBuilder()
         {
             TimeToLive = TimeSpan.FromMinutes(60);
         }
 
+        /// <summary>
+        /// Name of the authorization rule
```
abhipsaMisra, line 30: Since we call this a "shared access policy" in azure terms, how about we call this something like - "The shared access policy name"?

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -11,17 +11,29 @@
...
+        /// </summary>
         public string KeyName { get; set; }
 
+        /// <summary>
+        /// The key used to sign the request string
+        /// </summary>
```
abhipsaMisra, line 36: Similarly - "the shared access key value"?

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -33,10 +45,20 @@ public string Key
             }
         }
 
+        /// <summary>
+        /// The resource being accessed
+        /// </summary>
         public string Target { get; set; }
 
+        /// <summary>
+        /// Token expiry instant
```
abhipsaMisra, line 54: do you think the term "instant" might be misleading, since this is a timespan?

also, kind-of reminds me of this java type: https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html 😆

iothub/service/src/Common/Security/ISharedAccessSignatureCredential.cs
```diff
@@ -6,17 +6,38 @@ namespace Microsoft.Azure.Devices.Common.Security
...
+
+    /// <summary>
+    /// Credential interface used for authentication and authorization
+    /// </summary>
     public interface ISharedAccessSignatureCredential
     {
+        /// <summary>
+        /// Indicates if the token has expired
+        /// </summary>
         bool IsExpired();
```
abhipsaMisra, line 18: q - in .NET, do all methods in a public interface have public visibility? should we declare it as such explicitly?

abhipsaMisra (approved):
just some minor comments

bikamani (approved):

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -33,10 +45,20 @@ public string Key
             }
         }
 
+        /// <summary>
+        /// The resource being accessed
```
vinagesh, line 49: Probably resource Id? An example might help here. Ex: iothubname/devices/deviceId

iothub/service/src/Common/Security/CryptoKeyGenerator.cs
```diff
@@ -92,11 +106,21 @@ public static Guid GenerateGuid()
...
         {
             return GeneratePassword(DefaultPasswordLength, false);
         }
 
+        /// <summary>
+        /// Generate a unique password with a specific length and a flag to indicate whether to encode the password
+        /// </summary>
+        /// <param name="length">Desired length of the password</param>
+        /// <param name="base64Encoding">Encode the password if set to True. False otherwise</param>
+        /// <returns>A generated key</returns>
```
vinagesh, line 123: Returns a password?

vinagesh (approved):

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
barustum: Good point. How does "The time before the signature expires" sound?

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -11,17 +11,29 @@
 
 namespace Microsoft.Azure.Devices.Client
 {
+    /// <summary>
+    /// A helper class to build Shared Access Signature tokens
```
drwill-ms, line 15: How about `Builds Shared Access Signature (SAS) tokens`

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -11,17 +11,29 @@
...
+    /// </summary>
     public class SharedAccessSignatureBuilder
     {
         private string _key;
 
+        /// <summary>
+        /// Initializes a new instance of <see cref="SharedAccessSignatureBuilder"/> class
```
drwill-ms, line 22: Should include punctuation so it shows up proper on the web.

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
drwill-ms: Preferred notation for **example** would be E.G.

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
drwill-ms: Why not `The time the token expires.`?

iothub/service/src/Common/Security/CryptoKeyGenerator.cs
```diff
@@ -73,6 +83,10 @@ public static string GenerateKeyInHex(int keySize)
             return BitConverter.ToString(keyBytes).Replace("-", "");
         }
 
+        /// <summary>
+        /// Generate a unique GUID
```
drwill-ms, line 87: It isn't just a unique GUID, because otherwise it would be `Guid.NewGuid()`. What kind of GUID does it create?

iothub/service/src/Common/Security/SecurityConstants.cs
```diff
@@ -3,6 +3,9 @@
 
 namespace Microsoft.Azure.Devices.Common.Security
 {
+    /// <summary>
+    /// Security related constants
```
drwill-ms, line 7: `Security-related` is a compound adjective, so it should be hyphenated.

iothub/service/src/Common/Security/SecurityHelper.cs
```diff
@@ -6,8 +6,16 @@
 
 namespace Microsoft.Azure.Devices.Common.Security
 {
+    /// <summary>
+    /// A helper class that performs security related operations
```
drwill-ms, line 10: `security-related` also hyphenated

iothub/service/src/Common/Security/SharedAccessSignature.cs
```diff
@@ -110,6 +128,11 @@ public static SharedAccessSignature Parse(string iotHubName, string rawToken)
                 encodedAudience);
         }
 
+        /// <summary>
+        /// Helper function to validate whether a string token is a valid SAS token
```
drwill-ms, line 132: `Validates whether a string token is a valid SAS token.`

iothub/service/src/Common/Security/SharedAccessSignature.cs
```diff
@@ -171,11 +198,19 @@ public void Authenticate(SharedAccessSignatureAuthorizationRule sasAuthorization
             throw new UnauthorizedAccessException("The specified SAS token has an invalid signature. It does not match either the primary or secondary key.");
         }
 
+        /// <summary>
+        /// Authorize access to the IoT Hub
+        /// </summary>
+        /// <param name="iotHubHostName">IoT Hub host to authorize against</param>
```
drwill-ms, line 204: authorize to?

iothub/service/src/Common/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -1,56 +1,77 @@
...
 {
-    using System;
-    using System.Collections.Generic;
-    using System.Globalization;
-    using System.Net;
-    using System.Security.Cryptography;
-    using System.Text;
-
+    /// <summary>
+    /// A helper class to build Shared Access Signature tokens
```
drwill-ms, line 14: Again, you can get rid of the meta "a helper class to". You can just say `Builds Shared Access Signature (SAS) tokens.`

iothub/service/src/Common/WebApi/CustomHeaderConstants.cs
```diff
@@ -3,6 +3,9 @@
 
 namespace Microsoft.Azure.Devices.Common.WebApi
 {
+    /// <summary>
+    /// A static class encapsulating constants that can be used in custom headers
```
drwill-ms, line 7: Lose the meta "A static class encapsulating"

iothub/service/src/Common/WebApi/CustomHeaderConstants.cs
```diff
@@ -89,12 +92,18 @@ public static class CustomHeaderConstants
         public const string HttpAppPropertyPrefix = "iothub-app-";
 
         /// <summary>
-        /// Internal HTTP custom headers
+        /// A constant representing iothub-name custom header
         /// </summary>
         public const string HttpIotHubName = "iothub-name";
 
+        /// <summary>
+        /// A constant representing iothub-metadata custom header
```
drwill-ms, line 100: The iothub-metadata custom header key.

iothub/service/src/Common/Security/ISharedAccessSignatureCredential.cs
barustum: according to https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers#other-types


"Interface members are public by default because the purpose of an interface is to enable other types to access a class or struct."

iothub/device/src/Common/Security/SharedAccessSignatureBuilder.cs
drwill-ms: Everywhere.

drwill-ms (changes requested):
I've left some comments, many of which apply throughout the PR.

iothub/service/src/Common/Security/CryptoKeyGenerator.cs
barustum: This method generates a guid using bytes generated from the RNGCryptoServiceProvider class. Will update to reflect that.

iothub/service/src/Common/Security/CryptoKeyGenerator.cs
```diff
@@ -73,6 +83,10 @@ public static string GenerateKeyInHex(int keySize)
             return BitConverter.ToString(keyBytes).Replace("-", "");
         }
 
+        /// <summary>
+        /// Generate a unique GUID.
```
drwill-ms, line 87: I still feel this isn't specific enough. It creates a special kind of GUID, right? Can we delve into that?

iothub/service/src/Common/Security/ISharedAccessSignatureCredential.cs
```diff
@@ -6,17 +6,38 @@ namespace Microsoft.Azure.Devices.Common.Security
     using System;
 
     using Microsoft.Azure.Devices.Common.Data;
-      
+
+    /// <summary>
+    /// Credential interface used for authentication and authorization
```
drwill-ms, line 11: Punctuation needed from here on down.

iothub/service/src/Common/Security/SecurityHelper.cs
```diff
@@ -6,8 +6,16 @@
 
 namespace Microsoft.Azure.Devices.Common.Security
 {
+    /// <summary>
+    /// A helper class that performs security-related operations.
```
drwill-ms, line 10: Remove the meta "a helper class that"

drwill-ms (approved):

iothub/service/src/Common/Security/CryptoKeyGenerator.cs
barustum: oops, I changed the '<return>' comment, but not the summary one. Changing now.

PR #1747, opened 2021-01-15
fix(device-client) - Fixed xml comment warnings in device/src

vinagesh:
## Checklist

### DIFF
--- a/iothub/device/src/DeviceAuthenticationWithTpm.cs
+++ b/iothub/device/src/DeviceAuthenticationWithTpm.cs
@@ -18,6 +18,12 @@ namespace Microsoft.Azure.Devices.Client
     {
         private SecurityProviderTpm _securityProvider;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DeviceAuthenticationWithTpm"/> class with default
+        /// time to live of 1 hour and default buffer percentage value of 15.
+        /// </summary>
+        /// <param name="deviceId">Device Identifier.</param>
+        /// <param name="securityProvider">Device Security Provider settings for TPM Hardware Security Modules.</param>
         public DeviceAuthenticationWithTpm(
             string deviceId,
             SecurityProviderTpm securityProvider) : base(deviceId)
@@ -25,6 +31,14 @@ namespace Microsoft.Azure.Devices.Client
             _securityProvider = securityProvider ?? throw new ArgumentNullException(nameof(securityProvider));
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DeviceAuthenticationWithTpm"/> class.
+        /// </summary>
+        /// <param name="deviceId">Device Identifier.</param>
+        /// <param name="securityProvider">Device Security Provider settings for TPM Hardware Security Modules.</param>
+        /// <param name="suggestedTimeToLiveSeconds">Token time to live suggested value.</param>
+        /// <param name="timeBufferPercentage">Time buffer before expiry when the token should be renewed expressed as percentage of
+        /// the time to live. EX: If you want a SAS token to live for 85% of life before proactive renewal, this value should be 15.</param>
         public DeviceAuthenticationWithTpm(
             string deviceId,
             SecurityProviderTpm securityProvider,
--- a/iothub/device/src/Http1TransportSettings.cs
+++ b/iothub/device/src/Http1TransportSettings.cs
@@ -15,20 +15,36 @@ namespace Microsoft.Azure.Devices.Client
     {
         private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromSeconds(60);
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Http1TransportSettings"/> class.
+        /// </summary>
         public Http1TransportSettings()
         {
             Proxy = DefaultWebProxySettings.Instance;
         }
 
+        /// <summary>
+        /// Gets the transport type for this settings class.
+        /// </summary>
+        /// <returns>HyperText Transfer Protocol transport type. <see cref="TransportType.Http1"/></returns>
         public TransportType GetTransportType()
         {
             return TransportType.Http1;
         }
 
+        /// <summary>
+        /// Device certificate used for authentication.
+        /// </summary>
         public X509Certificate2 ClientCertificate { get; set; }
 
+        /// <summary>
+        /// The time to wait for a receive operation. The default value is 1 minute.
+        /// </summary>
         public TimeSpan DefaultReceiveTimeout => s_defaultOperationTimeout;
 
+        /// <summary>
+        /// Proxy information.
+        /// </summary>
         public IWebProxy Proxy { get; set; }
     }
 }
--- a/iothub/device/src/MethodRequest.cs
+++ b/iothub/device/src/MethodRequest.cs
@@ -6,20 +6,46 @@ using System.Text;
 
 namespace Microsoft.Azure.Devices.Client
 {
+    /// <summary>
+    /// The data structure that represents a method request.
+    /// </summary>
     public sealed class MethodRequest
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MethodRequest"/> class with without any method data
+        /// and extra time for device to connect and send a response.
+        /// </summary>
+        /// <param name="name">The method name.</param>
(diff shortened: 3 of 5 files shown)


iothub/device/src/MethodResponseStatusCode.cs
```diff
@@ -8,8 +8,11 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public enum MethodResponseStatusCode
     {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
vinagesh, line 11: Adding comments here seemed redundant. If someone has a better idea, let me know.

iothub/device/src/MethodRequest.cs
```diff
@@ -6,20 +6,46 @@
 
 namespace Microsoft.Azure.Devices.Client
 {
+    /// <summary>
+    /// The data structure that represents a method request.
+    /// </summary>
     public sealed class MethodRequest
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MethodRequest"/> class with without any method data and extra time for
```
vinagesh, line 15: This is what I found by looking in code. Someone with more context please validate.

iothub/device/src/DeviceAuthenticationWithTpm.cs
```diff
@@ -18,13 +18,26 @@ public sealed class DeviceAuthenticationWithTpm : DeviceAuthenticationWithTokenR
     {
         private SecurityProviderTpm _securityProvider;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DeviceAuthenticationWithTpm"/> class with default
+        /// time to live of 1 minute and default buffer percentage value of 15.
```
abhipsaMisra, line 23: isn't the default ttl 1 hour?

iothub/device/src/Http1TransportSettings.cs
```diff
@@ -15,20 +15,36 @@ public sealed class Http1TransportSettings : ITransportSettings
     {
         private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromSeconds(60);
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Http1TransportSettings"/> class with default proxy settings.
```
abhipsaMisra, line 19: I wouldn't specify that we initialize it with any default proxy setting

iothub/device/src/DeviceAuthenticationWithTpm.cs
```diff
@@ -18,13 +18,26 @@ public sealed class DeviceAuthenticationWithTpm : DeviceAuthenticationWithTokenR
...
             _securityProvider = securityProvider ?? throw new ArgumentNullException(nameof(securityProvider));
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DeviceAuthenticationWithTpm"/> class.
+        /// </summary>
+        /// <param name="deviceId">Device Identifier.</param>
+        /// <param name="securityProvider">Device Security Provider settings for TPM Hardware Security Modules.</param>
+        /// <param name="suggestedTimeToLiveSeconds">Token time to live suggested value.</param>
+        /// <param name="timeBufferPercentage">Time buffer before expiry when the token should be renewed expressed as percentage of the time to live.</param>
```
timtay-microsoft, line 40: The wording here could use an example to really clarify that if they want a SAS token to live for 85% of life before proactive renewal, then this value should be 15.

iothub/device/src/Http1TransportSettings.cs
timtay-microsoft, line 19: Does "Default proxy settings" mean that no proxy will be used?

iothub/device/src/Http1TransportSettings.cs
```diff
@@ -15,20 +15,36 @@ public sealed class Http1TransportSettings : ITransportSettings
...
+        /// </summary>
         public Http1TransportSettings()
         {
             Proxy = DefaultWebProxySettings.Instance;
         }
 
+        /// <summary>
+        /// Gets the transport type for this settings class.
+        /// </summary>
+        /// <returns>HyperText Transfer Protocol transport type.</returns>
```
timtay-microsoft, line 29: Maybe link the enum value directly here instead?

iothub/device/src/MethodRequest.cs
```diff
@@ -6,20 +6,46 @@
...
         {
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MethodRequest"/> class.
+        /// </summary>
+        /// <param name="name">The method name.</param>
+        /// <param name="data">The method data.</param>
+        /// <param name="responseTimeout">The method timeout value.</param>
+        /// <param name="connectionTimeout">The device connection timeout value.</param>
```
timtay-microsoft, line 48: It's worth spelling out what these response/connect timeouts actually do. You can read about it in the paragraph in [this doc](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-direct-methods#method-invocation)

iothub/device/src/MethodRequest.cs
```diff
@@ -28,18 +54,33 @@ public MethodRequest(string name, byte[] data, TimeSpan? responseTimeout, TimeSp
             ConnectionTimeout = connectionTimeout;
         }
 
+        /// <summary>
+        /// The method name.
+        /// </summary>
         public string Name { get; private set; }
 
+        /// <summary>
+        /// The method data.
```
timtay-microsoft, line 63: It's worth noting that this data must be serialized json

timtay-microsoft (approved):
Just a few small comments, otherwise looks good

iothub/device/src/MethodRequest.cs
timtay-microsoft: Also worth adding that there are service side limits to these values, also found in the above doc

iothub/device/src/Http1TransportSettings.cs
abhipsaMisra: default proxy behavior - whatever is specified at your OS level

iothub/device/src/MethodRequest.cs
```diff
@@ -28,18 +54,37 @@ public MethodRequest(string name, byte[] data, TimeSpan? responseTimeout, TimeSp
...
         [System.Diagnostics.CodeAnalysis.SuppressMessage(
             "Performance",
             "CA1819:Properties should not return arrays",
             Justification = "Cannot change property types on public classes.")]
         public byte[] Data { get; private set; }
 
+        /// <summary>
+        /// The method response timeout value. This value is the amount of time that IoT Hub service must await for
+        /// completion of a direct method execution on a device. The minimum and maximum values are 5 and 300 seconds.
+        /// </summary>
```
abhipsaMisra, line 74: it is worth specifying here that these values for `MethodRequest` are relevant only when invoking methods from one edge module to another. For device client operations the user will receive the `MethodRequest` as a part of the direct method callback invoked, so they wouldn't really be setting these timeout values.

iothub/device/src/MethodResponseStatusCode.cs
abhipsaMisra: The description for 500 = UserCodeException is a bit confusing to me; isn't 500 always a service error?

for the warning - how about we add it to the suppressions file instead, might look cleaner. It looks like we have one for device client: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/GlobalSuppressions.cs

iothub/device/src/MethodRequest.cs
abhipsaMisra: the default value for timeouts here should be null -> not specified -> 30secs for response timeout and 0s for connect: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-direct-methods#method-invocation

abhipsaMisra (approved):
just the one comment about method request timeout value, everything else looks good 👍🏽

iothub/device/src/MethodResponseStatusCode.cs
azabbasi: What if we mention that these values are interchangable with actual http status codes?
Anyways, if you decide not to add comments for them, you can supress them in a single attribute on the class

azabbasi (approved):
Just a small comment on the MethodResponseStatusCode file.

PR #1749, opened 2021-01-19
fix(device-client): Add missing xml documentation - CS1591

vinagesh:
## Checklist


iothub/device/src/Common/ErrorCode.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.Azure.Devices.Client.Errors
     /// </remarks>
     public enum ErrorCode
     {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
vinagesh, line 14: I could not find a way to add this to the global file so had to do it in place. Also if anyone has good context on any of these errors that would be good to include as comments, let me know.

timtay-microsoft (approved):

PR #1750, opened 2021-01-20
Mark ClientWebsocketChannel class as internal

azabbasi:
We have had some back and forth about this class (and honestly, many other classes) that should have never been public to begin with. 
The argument here is that there is no value in terms of our SDK's functionality whether or not the user can instantiate a new ClientWebsocketChannel. 
This type is not exposed in any of our public API signatures and should be hidden from the user.


abhipsaMisra (approved):

drwill-ms (approved):
mark as in MarkMyd? I guess he's internal.

PR #1753, opened 2021-01-21
fix(service-client): Add missing xml documentation - CS1591

vinagesh:
## Checklist

### DIFF
--- a/iothub/service/src/Common/Exceptions/ErrorCode.cs
+++ b/iothub/service/src/Common/Exceptions/ErrorCode.cs
@@ -3,12 +3,17 @@
 
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
+    /// <summary>
+    /// Error Codes for common IoT hub exceptions.
+    /// </summary>
     public enum ErrorCode
     {
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
         InvalidErrorCode = 0,
 
-        //BadRequest - 400
+        // BadRequest - 400
         InvalidProtocolVersion = 400001,
+
         DeviceInvalidResultCount = 400002,
         InvalidOperation = 400003,
         ArgumentInvalid = 400004,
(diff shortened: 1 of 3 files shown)


iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -3,12 +3,19 @@
 
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
+    /// <summary>
```
vinagesh, line 6: I don't have any more context that I can add to these error codes and it seemed self-explanatory to an extent. If anyone has any insights to share, I can include them.

iothub/service/src/Common/Extensions/CommonExtensions.cs
```diff
@@ -17,12 +17,12 @@
 
 namespace Microsoft.Azure.Devices.Common
 {
-    public delegate bool TryParse<TInput, TOutput>(TInput input, bool ignoreCase, out TOutput output);
+    internal delegate bool TryParse<TInput, TOutput>(TInput input, bool ignoreCase, out TOutput output);
```
vinagesh, line 20: We were using this helper only in the IotHubConnectionStringBuilder class. It has functions like IsNullOrWhiteSpace, RemoveWhitespace etc... They definitely don't need any xml comments. However this seems like a very good candidate to make internal. There is an off chance a customer is dependent on this but looks very highly unlikely to me. I want to bring this up here so that I can get some insight from others as well who have more context.

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -23,47 +30,75 @@ public enum ErrorCode
         BulkRegistryOperationFailure = 400013,
         CannotRegisterModuleToModule = 400301,
 
-        //Unauthorized - 401
+        /// <summary>
+        /// Unauthorized - 401
+        /// </summary>
         IotHubNotFound = 401001,
```
timtay-microsoft, line 36: Since the naming of this enum is questionable at best, I'd add to the comment that "This error code signals that the cloud to device operation failed because the device is either not registered, or is disabled"

https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-troubleshoot-error-404001-devicenotfound

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -23,47 +30,75 @@ public enum ErrorCode
...
+
         IotHubUnauthorizedAccess = 401002,
         IotHubUnauthorized = 401003,
```
timtay-microsoft, line 39: Again, the naming of this enum isn't doing itself any favors. It could use a comment about "This error code signals that the cloud to device operation failed because the device is not currently connected to IoT Hub"

https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-troubleshoot-error-404103-devicenotonline

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -23,47 +30,75 @@ public enum ErrorCode
...
 
-        // Forbidden - 403
+        /// <summary>
+        /// Forbidden - 403
+        /// </summary>
         IotHubSuspended = 403001,
+
         IotHubQuotaExceeded = 403002,
```
timtay-microsoft, line 46: "The daily message quota for the IoT hub is exceeded." Just so that users know the quota is in regard to messages, not to jobs/desired property updates/methods or whatever else they think may have a quota

timtay-microsoft: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-troubleshoot-error-403002-iothubquotaexceeded

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -23,47 +30,75 @@ public enum ErrorCode
...
         JobQuotaExceeded = 403003,
         DeviceMaximumQueueDepthExceeded = 403004,
         IotHubMaxCbsTokenExceeded = 403005,
 
-        // NotFound - 404
```
timtay-microsoft, line 38: I think these comments were intentionally not xml comments. They were a sort of grouping for our eyes, since the split up the http error codes by the 3 digit prefix. I'd be inclined to leave them alone rather than convert them to xml comments on a specific error code like you did here

timtay-microsoft (approved):
Looks good, other than the ErrorCode.cs file. The links I gave should help you give better descriptions for at least some of these error codes

iothub/service/src/Common/Exceptions/ErrorCode.cs
vinagesh: Cool I think a link to that will also be good to add.

iothub/service/src/Common/Exceptions/ErrorCode.cs
vinagesh: Thanks for the link. It has details on a lot of errors. I will include all the useful ones and update my PR.

iothub/service/src/Common/Exceptions/ErrorCode.cs
azabbasi: I still think it might be worth having them for just not having to suppress them. also, the enum string might be nice to have, I would remove the actual http error code though:
```
///
// Not Found
///
```

azabbasi: unless there is a code that is not representing its HTTP counterpart. Which I couldn't find

iothub/service/src/Common/Extensions/CommonExtensions.cs
abhipsaMisra: These being extension methods on types such as strings and dictionaries, there might be a slight chance that users could have taken a dependency on these, for some logic in their application code. It wouldn't hurt to leave it as-is.

azabbasi (approved):

iothub/service/src/Common/Extensions/CommonExtensions.cs
vinagesh: I feel like customers might not be using this and should be safe. Is is good for us to not provide such extensions. This class is truly internal. But I definitely don't want to break customers if it is not acceptable. @drwill-ms what are your thoughts? Can we use our discretion and make such changes or not?

iothub/service/src/Common/Exceptions/ErrorCode.cs
vinagesh: > Looks good, other than the ErrorCode.cs file. The links I gave should help you give better descriptions for at least some of these error codes

I have added all the explanation available. Not everything was documented so I had to add the suppress in the end.

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -23,47 +28,101 @@ public enum ErrorCode
...
 
-        //RequestEntityTooLarge - 413
+        // RequestEntityTooLarge - 413
         MessageTooLarge = 413001,
+
         TooManyDevices = 413002,
         TooManyModulesOnDevice = 413003,
 
-        //Throttling Exception 
+        // Throttling Exception
```
bikamani, line 96: Nit: We can remove these one line comments now since the comments you added is lot more detailed :)

bikamani (approved):

iothub/service/src/Common/Exceptions/ErrorCode.cs
vinagesh: These are just grouping the different exceptions. I don't have details for all comments so I just left the grouping there.

bikamani: Ok

iothub/service/src/Common/Extensions/CommonExtensions.cs
drwill-ms: I don't see a compelling enough reason to possibly break customers. Why take the chance?

drwill-ms: In the very least, we'd mark it as deprecated and eventually make it internal.

vinagesh: That is good. Let me mark it as deprecated. We should surely give the impression that we will help support any kind of custom string extensions.

PR #1754, opened 2021-01-21
fix(release-pipeline): Adding the new infra needed to run auth tests …

vinagesh:
…in nuget test pipeline

## Checklist


vsts/test-release-nuget.yaml
```diff
@@ -86,6 +86,11 @@ jobs:
           LA_AAD_APP_ID: $(LA-AAD-APP-ID)
           LA_AAD_APP_CERT_BASE64: $(LA-AAD-APP-CERT-BASE64)
           LA_WORKSPACE_ID: $(LA-WORKSPACE-ID)
+          IOTHUB_X509_CHAIN_DEVICE_NAME: $(IOTHUB-X509-CHAIN-DEVICE-NAME)
+          HUB_CHAIN_DEVICE_PFX_CERTIFICATE: $(HUB-CHAIN-DEVICE-PFX-CERTIFICATE)
```
azabbasi, line 90: So it looks like they are both targetting the same infra environment variable group, it should suffice I think ! Thanks Sindhu.

azabbasi (approved):
LGTM

abhipsaMisra (approved):

vsts/test-release-nuget.yaml
vinagesh: yes this should be sufficient

PR #1756, opened 2021-01-26
Add xml comments and update visibility of unused interface

abhipsaMisra:


iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -1,18 +1,25 @@
...
 using System;
 using Newtonsoft.Json;
 
 namespace Microsoft.Azure.Devices.Common.Data
 {
-    public sealed class SharedAccessSignatureAuthorizationRule : IEquatable<SharedAccessSignatureAuthorizationRule>
+    /// <summary>
+    /// A shared access signature based authorization rule for authenticating requests against an IoT Hub.
+    /// </summary>
+    internal sealed class SharedAccessSignatureAuthorizationRule : IEquatable<SharedAccessSignatureAuthorizationRule>
```
abhipsaMisra, line 12: I made this class internal since it was not being referenced anywhere in our public APIs -> users could have initialized this class, but there was never any way that they could pass this to our client.

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -1,18 +1,25 @@
...
     {
         private string _primaryKey;
         private string _secondaryKey;
 
         [JsonProperty(PropertyName = "keyName")]
```
abhipsaMisra, line 17: I cannot figure out what this field is supposed to signify. KeyName usually represents the shared access policy name but we already get that info via the `Rights` property

abhipsaMisra: Also, I have no idea why these properties have json annotations -> this object isn't something we'd generally serialialize or deserialize.

iothub/service/src/Common/Security/ISharedAccessSignatureCredential.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.Devices.Common.Security
     /// <summary>
     /// Credential interface used for authentication and authorization.
     /// </summary>
-    public interface ISharedAccessSignatureCredential
+    internal interface ISharedAccessSignatureCredential
```
abhipsaMisra, line 12: again, this interface is not referenced anywhere in our public API, so I made it internal

iothub/service/src/Common/Security/SharedAccessSignature.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Devices.Common.Security
     /// <summary>
     /// A shared access signature, which can be used for authorization to an IoT Hub.
     /// </summary>
-    public sealed class SharedAccessSignature : ISharedAccessSignatureCredential
+    internal sealed class SharedAccessSignature : ISharedAccessSignatureCredential
```
abhipsaMisra, line 17: same here

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -37,9 +47,17 @@ public string SecondaryKey
...
+        /// The name of the shared access policy that will be used to grant permission to IoT Hub endpoints.
+        /// </summary>
         [JsonProperty(PropertyName = "rights")]
         public AccessRights Rights { get; set; }
 
+        /// <summary>
+        /// Evaluates if two <see cref="SharedAccessSignatureAuthorizationRule"/> objects have the same shared access policies and keys.
+        /// </summary>
+        /// <param name="other">The <see cref="SharedAccessSignatureAuthorizationRule"/> object to compare the current <see cref="SharedAccessSignatureAuthorizationRule"/> to.</param>
+        /// <returns></returns>
```
drwill-ms, line 60: fill in or remove

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -37,9 +47,17 @@ public string SecondaryKey
...
+        /// <summary>
+        /// The name of the shared access policy that will be used to grant permission to IoT Hub endpoints.
+        /// </summary>
         [JsonProperty(PropertyName = "rights")]
         public AccessRights Rights { get; set; }
 
+        /// <summary>
+        /// Evaluates if two <see cref="SharedAccessSignatureAuthorizationRule"/> objects have the same shared access policies and keys.
+        /// </summary>
+        /// <param name="other">The <see cref="SharedAccessSignatureAuthorizationRule"/> object to compare the current <see cref="SharedAccessSignatureAuthorizationRule"/> to.</param>
```
drwill-ms, line 59: Seems redundant

iothub/service/src/Common/Data/AccessRights.cs
```diff
@@ -1,16 +1,16 @@
...
 {
-    using System;
-    using System.Collections.Generic;
-    using Newtonsoft.Json;
-    using Newtonsoft.Json.Converters;
-
     /// <summary>
     /// Shared access policy permissions of IoT hub.
-    /// For more information, <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#iot-hub-permissions"/>.
+    /// For more information, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#iot-hub-permissions"/>.
```
bikamani, line 13: I see "see" twice :) Is that intentional?

iothub/service/src/Common/Data/AccessRights.cs
```diff
@@ -19,28 +19,28 @@ public enum AccessRights
...
+        /// It grants permission to receive device-to-cloud messages, send cloud-to-device messages, retrieve delivery acknowledgments for sent messages
+        /// and file uploads, retrieve desired and reported properties on device twins, update tags and desired properties on device twins,
         /// and run queries.
         /// </summary>
         ServiceConnect = 4,
 
         /// <summary>
         /// Grants access to device facing endpoints.
-        /// It grants permission to send and receive device-to-cloud messages, perform file upload from a device, receive device twin
+        /// It grants permission to send device-to-cloud messages and receive cloud-to-device, perform file upload from a device, receive device twin
```
bikamani, line 43: add  "messages" following receive cloud-to-device. But to me however the previous one seems enough as well.

provisioning/service/src/Auth/SharedAccessSignature.cs
```diff
@@ -33,51 +29,27 @@ private SharedAccessSignature(string shareAccessSignatureName, DateTime expiresO
...
-        public string ShareAccessSignatureName
-        {
-            get
-            {
-                return _shareAccessSignatureName;
-            }
-        }
+        public string ShareAccessSignatureName { get; private set; }
 
         public DateTime ExpiresOn
```
drwill-ms, line 42: format this one similarly?

drwill-ms (approved):

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
abhipsaMisra: oops, I missed this, will fill

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
abhipsaMisra: how would you suggest I phrase it?

iothub/service/src/Common/Data/AccessRights.cs
abhipsaMisra: yes, the first "see" is a part of the statement, the second "see" is a part of the xml tag

abhipsaMisra: this is so that the statement is rendered as:
"For more information, see https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#iot-hub-permissions"
instead of
"For more information, https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#iot-hub-permissions"

iothub/service/src/Common/Data/AccessRights.cs
abhipsaMisra: we had previously stated that this access rights grants send and receive permission for d2c devices; whereas it actually grants access rights to send d2c message and receive c2d message (notice that the source and target are different for each operation)

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
drwill-ms: You don't need to add links to the type twice, for one. I'd be fine with that.

iothub/service/src/Common/Data/AccessRights.cs
drwill-ms: Yep, much better.

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
barustum: I see it being used here:
https://github.com/Azure/iotedge/blob/2d26602a08c787551ce9f6dc977bab3169048a7d/edge-hub/core/src/Microsoft.Azure.Devices.Edge.Hub.CloudProxy/authenticators/DeviceScopeTokenAuthenticator.cs

barustum: actually, this might be a different SharedAccessSignatureAuthorizationRule...

abhipsaMisra: that is a good find!
This being a service client class, I'm surprised that the edge team is taking a dependency on this, but I'll definitely verify it with them before merging the PR in.

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
abhipsaMisra: So I ended up removing the doc comments, since it seemed self-explanatory and I had also modified the type to be internal.

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
bikamani: Interesting, best to check with service team.

provisioning/service/src/Auth/SharedAccessSignature.cs
```diff
@@ -33,51 +29,23 @@ private SharedAccessSignature(string shareAccessSignatureName, DateTime expiresO
                 throw new UnauthorizedAccessException("The specified SAS token is expired");
             }
 
-            _shareAccessSignatureName = shareAccessSignatureName;
-            _signature = signature;
-            _audience = WebUtility.UrlDecode(encodedAudience);
+            ShareAccessSignatureName = shareAccessSignatureName;
```
bikamani, line 32: where is this defined?

provisioning/service/src/Auth/SharedAccessSignature.cs
```diff
@@ -33,51 +29,23 @@ private SharedAccessSignature(string shareAccessSignatureName, DateTime expiresO
...
         }
 
-        public string ShareAccessSignatureName
-        {
-            get
-            {
-                return _shareAccessSignatureName;
-            }
-        }
+        public string ShareAccessSignatureName { get; private set; }
```
bikamani, line 40: Is this per standard? public fields below constructor?

provisioning/service/src/Auth/SharedAccessSignature.cs
```diff
@@ -33,51 +29,23 @@ private SharedAccessSignature(string shareAccessSignatureName, DateTime expiresO
                 throw new UnauthorizedAccessException("The specified SAS token is expired");
             }
 
-            _shareAccessSignatureName = shareAccessSignatureName;
-            _signature = signature;
-            _audience = WebUtility.UrlDecode(encodedAudience);
+            ShareAccessSignatureName = shareAccessSignatureName;
+            Signature = signature;
+            Audience = WebUtility.UrlDecode(encodedAudience);
             _encodedAudience = encodedAudience;
```
bikamani, line 35: How are these being referenced/used? There is no get/set right?

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
vinagesh: This is sooo weird. How can we move these to the Edge team :D We should not have helpers for their code. I saw a similar thing on my PR.

provisioning/service/src/Auth/SharedAccessSignature.cs
abhipsaMisra: yes, since this is a public property, it would be placed below any constructors

provisioning/service/src/Auth/SharedAccessSignature.cs
abhipsaMisra: Is there a particular field you're asking about?
`_encodedAudience` in particular is being used while computing the signature (https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/serviceInternal/provisioning/service/src/Auth/SharedAccessSignature.cs#L164)

iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
abhipsaMisra: So, they fork our repo and add some stuff on top of it, which they then version it as the "Nested-Edge-xxx" in their code (this is not publicly available). I am planning to send across an API surface diff to Varun to let him know of our intended changes. If you'd like I can generate the diff for your PR files as well @vinagesh .

provisioning/service/src/Auth/SharedAccessSignature.cs
bikamani: Fields which do not have getter/setters

bikamani: I see. Looked a bit weird on first glance since others are public now and this one is not.

barustum (approved):

PR #1757, opened 2021-01-26
fix(service-client): Add missing xml documentation - CS1591

vinagesh:
## Checklist

### DIFF
--- a/iothub/service/src/Common/Data/AmqpErrorMapper.cs
+++ b/iothub/service/src/Common/Data/AmqpErrorMapper.cs
@@ -183,7 +183,9 @@ namespace Microsoft.Azure.Devices.Common.Exceptions
             }
 
             error.Info.TryGetValue(IotHubAmqpProperty.TrackingId, out string trackingId);
+#pragma warning disable CS0618 // Type or member is obsolete only for external dependency.
             trackingId = TrackingHelper.CheckAndAddGatewayIdToTrackingId(trackingId);
+#pragma warning restore CS0618 // Type or member is obsolete only for external dependency.
             error.Info[IotHubAmqpProperty.TrackingId] = trackingId;
 
             return error;
--- a/iothub/service/src/Common/PerfectHash.cs
+++ b/iothub/service/src/Common/PerfectHash.cs
@@ -7,8 +7,16 @@ using System.Text;
 
 namespace Microsoft.Azure.Devices.Common
 {
+    /// <summary>
+    /// Generates hash values of a string.
+    /// </summary>
     public static class PerfectHash
     {
+        /// <summary>
+        /// Computes a long hash from the input data.
+        /// </summary>
+        /// <param name="data">Input data.</param>
+        /// <returns>Long hash computed from the input data.</returns>
         public static long HashToLong(string data)
         {
             if (data == null) { throw new ArgumentNullException(nameof(data), "The data string cannot be null."); }
@@ -17,12 +25,17 @@ namespace Microsoft.Azure.Devices.Common
 
             string upper = data.ToUpper(CultureInfo.InvariantCulture);
 
-            PerfectHash.ComputeHash(ASCIIEncoding.ASCII.GetBytes(upper), seed1: 0, seed2: 0, hash1: out hash1, hash2: out hash2);
+            ComputeHash(Encoding.ASCII.GetBytes(upper), seed1: 0, seed2: 0, hash1: out hash1, hash2: out hash2);
             long hashedValue = ((long)hash1 << 32) | (long)hash2;
 
             return hashedValue;
         }
 
+        /// <summary>
+        /// Computes a short hash from the input data.
+        /// </summary>
+        /// <param name="data">Input data.</param>
+        /// <returns>Short hash computed from the input data.</returns>
         public static short HashToShort(string data)
         {
             if (data == null) { throw new ArgumentNullException(nameof(data), "The data string cannot be null."); }
@@ -32,13 +45,13 @@ namespace Microsoft.Azure.Devices.Common
 
             string upper = data.ToUpper(CultureInfo.InvariantCulture);
 
-            PerfectHash.ComputeHash(ASCIIEncoding.ASCII.GetBytes(upper), seed1: 0, seed2: 0, hash1: out hash1, hash2: out hash2);
+            ComputeHash(Encoding.ASCII.GetBytes(upper), seed1: 0, seed2: 0, hash1: out hash1, hash2: out hash2);
             long hashedValue = hash1 ^ hash2;
 
             return (short)hashedValue;
         }
 
-        // Perfect hashing implementation. source: distributed cache team
+        // Perfect hashing implementation. Source: distributed cache team.
         private static void ComputeHash(byte[] data, uint seed1, uint seed2, out uint hash1, out uint hash2)
         {
             uint a, b, c;
--- a/iothub/service/src/Common/TrackingHelper.cs
+++ b/iothub/service/src/Common/TrackingHelper.cs
@@ -13,10 +13,18 @@ namespace Microsoft.Azure.Devices.Common
     using Microsoft.Azure.Devices.Common.Client;
     using Microsoft.Azure.Devices.Common.Exceptions;
 
-    //TrackingId format is GUID[-G:<GatewayId>][-B:<BackendId>][-P:<PartitionId>]-TimeStamp:<Timestamp>
-
+    /// <summary>
+    /// Generates tracking Id with gateway, backend, partition and timestamp data.
+    /// </summary>
+    /// <remarks>
+    /// Tracking Id format is [GUID][-G:GatewayId][-B:BackendId][-P:PartitionId][-TimeStamp:Timestamp]
+    /// </remarks>
+    [Obsolete("This is for internal use only. SDK will not support this for external usage.")]
     public static class TrackingHelper
     {
+        /// <summary>
+        /// Gateway Id.
+        /// </summary>
         [SuppressMessage("Usage", "CA2211:Non-constant fields should not be visible", Justification = "This property may be used by others so it is not safe to modify it.")]
         public static string GatewayId;
 
@@ -25,17 +33,32 @@ namespace Microsoft.Azure.Devices.Common
         private const string PartitionPrefix = "-P:";
         private const string TimeStampPrefix = "-TimeStamp:";
 
+        /// <summary>
+        /// Generates a tracking Id. Not used.
+        /// </summary>
+        /// <returns>A tracking Id.</returns>
         public static string GenerateTrackingId()
         {
             return GenerateTrackingId(string.Empty, string.Empty);
         }
 
+        /// <summary>
+        /// Generates a tracking Id. Not used.
+        /// </summary>
+        /// <returns>A tracking Id.</returns>
         public static string GenerateTrackingId(string backendId, string partitionId)
         {
             string gatewayId = GatewayId;
             return GenerateTrackingId(gatewayId, backendId, partitionId);
         }
 
+        /// <summary>
+        /// Generates a unique tracking Id with gateway, backend, partition and timestamp information.
+        /// </summary>
+        /// <returns>A tracking Id.</returns>
+        /// <remarks>
+        /// Tracking Id format is [GUID][-G:GatewayId][-B:BackendId][-P:PartitionId][-TimeStamp:Timestamp]
+        /// </remarks>
         public static string GenerateTrackingId(string gatewayId, string backendId, string partitionId)
         {
             string trackingId = Guid.NewGuid().ToString("N", CultureInfo.InvariantCulture);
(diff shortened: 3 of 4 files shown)


iothub/service/src/Common/Data/AmqpErrorMapper.cs
```diff
@@ -194,7 +194,9 @@ public static Error ToAmqpError(Exception exception, bool includeStackTrace)
 
             string trackingId;
             error.Info.TryGetValue(IotHubAmqpProperty.TrackingId, out trackingId);
+#pragma warning disable CS0618 // Type or member is obsolete only for external dependency.
```
vinagesh, line 197: TrackingHelper file should be internal only but it is not. To avoid breaking customers, it has been marked as Obsolete so we can ignore this warning here.

iothub/service/src/Common/EventHubPartitionKeyResolver.cs
```diff
@@ -1,38 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
```
vinagesh, line 1: This is not used anywhere and does not seem like something a customer would be using. If anyone has other thoughts, let me know.

iothub/service/src/Common/TrackingHelper.cs
```diff
@@ -13,10 +13,18 @@ namespace Microsoft.Azure.Devices.Common
...
 
-    //TrackingId format is GUID[-G:<GatewayId>][-B:<BackendId>][-P:<PartitionId>]-TimeStamp:<Timestamp>
-
+    /// <summary>
+    /// Helper class to generate tracking Id
+    /// </summary>
+    /// <remarks>
+    /// Tracking Id format is [GUID][-G:GatewayId][-B:BackendId][-P:PartitionId][-TimeStamp:Timestamp]
+    /// </remarks>
+    [Obsolete("This is for internal use only. SDK will not support this for external usage.")]
```
vinagesh, line 22: Other than 1 function, no functions in this class are being used in our SDK. This is also something that should have been internal so I've marked it as Obsolete.

iothub/service/src/MessageSystemPropertyNames.cs
```diff
@@ -36,6 +36,9 @@ public static class MessageSystemPropertyNames
         /// </summary>
         public const string To = "to";
 
+        /// <summary>
```
vinagesh, line 39: For everything in this class, I got inputs from the IoT hub team.

timtay-microsoft (approved):

iothub/service/src/Common/EventHubPartitionKeyResolver.cs
barustum: I do see it being used in a few places. Searching in the Azure Organization, seems like it's being used by edge team:
https://github.com/Azure/iotedge/blob/master/test/modules/TestResultCoordinator/Services/TestResultEventReceivingService.cs

iothub/service/src/Common/PerfectHash.cs
```diff
@@ -7,8 +7,16 @@
 
 namespace Microsoft.Azure.Devices.Common
 {
+    /// <summary>
+    /// This is a helper class to generate hash values of a string.
```
barustum, line 11: Something I learned on my recent PR: remove sentences such as "This is a helper class" or "Helper method". And rather stick with format like "Generates hash values of a string."

iothub/service/src/Common/PerfectHash.cs
```diff
@@ -32,13 +45,14 @@ public static short HashToShort(string data)
...
-            PerfectHash.ComputeHash(ASCIIEncoding.ASCII.GetBytes(upper), seed1: 0, seed2: 0, hash1: out hash1, hash2: out hash2);
+            ComputeHash(Encoding.ASCII.GetBytes(upper), seed1: 0, seed2: 0, hash1: out hash1, hash2: out hash2);
             long hashedValue = hash1 ^ hash2;
 
             return (short)hashedValue;
         }
 
-        // Perfect hashing implementation. source: distributed cache team
+        // Perfect hashing implementation. Source: distributed cache team.
+
```
barustum, line 55: extra new line

iothub/service/src/Common/TrackingHelper.cs
```diff
@@ -13,10 +13,18 @@ namespace Microsoft.Azure.Devices.Common
     using Microsoft.Azure.Devices.Common.Client;
     using Microsoft.Azure.Devices.Common.Exceptions;
 
-    //TrackingId format is GUID[-G:<GatewayId>][-B:<BackendId>][-P:<PartitionId>]-TimeStamp:<Timestamp>
-
+    /// <summary>
+    /// Helper class to generate tracking Id
```
barustum, line 17: dot at end of sentence

iothub/service/src/Common/TrackingHelper.cs
```diff
@@ -25,17 +33,32 @@ public static class TrackingHelper
         private const string PartitionPrefix = "-P:";
         private const string TimeStampPrefix = "-TimeStamp:";
 
+        /// <summary>
+        /// Generates a tracking Id. Not used.
```
barustum, line 37: The method is "Not used"? This will show up in intellisense for one who wants to consume this method. Do we need the "Not used"?

barustum: Same comment for GenerateTrackingId

barustum: ... and other methods below :)

iothub/service/src/Common/EventHubPartitionKeyResolver.cs
vinagesh: Wow! That is some crazy dependency! I will restore it then. We are not using it in any of our code. Does it even make sense for them to use this?

vinagesh: Are you sure they are using our implementation or they have their own?

iothub/service/src/Common/TrackingHelper.cs
abhipsaMisra: It will show up as having 0 references in intellisense (if unused), any reason why we want to call it out in comments as well?
Also, +1 to not having our docs say "Not used" :)

Just for reference: we had a bunch of unused properties in transport settings. I had called those out as "unused" in our docs, just so that users are aware that even if they set those properties, the sdk will not reference/ use it anywhere. (not ideal, but a workaround until we either implement them or get rid of them).

iothub/service/src/MessageSystemPropertyNames.cs
abhipsaMisra: does DeliveryCount hold the max allowed value, or the current count of the no of times the message has transitioned between enqueued and invisible?
It might also help to add a link to the ms doc, so that users can visualize the entire lifecycle: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-c2d#the-cloud-to-device-message-life-cycle

iothub/service/src/MessageSystemPropertyNames.cs
```diff
@@ -36,6 +36,9 @@ public static class MessageSystemPropertyNames
...
+        /// The maximum number of times a message can transition between the Enqueued and Invisible states.
+        /// After the maximum number of transitions, the IoT hub sets the state of the message to dead lettered.
         /// </summary>
         public const string DeliveryCount = "iothub-deliverycount";
```
abhipsaMisra, line 43: does DeliveryCount hold the max allowed value, or the current count of the no of times the message has transitioned between enqueued and invisible?
It might also help to add a link to the ms doc, so that users can visualize the entire lifecycle: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-c2d#the-cloud-to-device-message-life-cycle

iothub/service/src/MessageSystemPropertyNames.cs
```diff
@@ -98,8 +101,14 @@ public static class MessageSystemPropertyNames
         /// </summary>
         public const string ConnectionDeviceGenerationId = "iothub-connection-auth-generation-id";
 
+        /// <summary>
+        /// The connection authentication method value is ignored for cloud-to-device messages.
+        /// </summary>
         public const string ConnectionAuthMethod = "iothub-connection-auth-method";
```
abhipsaMisra, line 107: can we add a note specifying what this auth method signifies?

iothub/service/src/Common/EventHubPartitionKeyResolver.cs
barustum: So looking at how it's used in their file, seems like it is using the Microsoft.Azure.Devices.Common namespace. However, I am looking at the csproj that is referencing out package, and I see this:
<PackageReference Include="Microsoft.Azure.Devices" Version="1.28.1-NestedEdge" />

This is in https://github.com/Azure/iotedge/blob/df7d717b21eb6ed5b9e8792902c4b985789e9506/test/modules/TestResultCoordinator/TestResultCoordinator.csproj#L23

So maybe they are using their own NuGet? which is a fork of our package?

iothub/service/src/Common/Data/AmqpErrorMapper.cs
abhipsaMisra: sad that these warnings cannot be moved to a global suppressions file; it would be so much more cleaner (although one can argue that having the suppression description inline is more descriptive and easier to understand)!

iothub/service/src/Common/EventHubPartitionKeyResolver.cs
vinagesh: Possible. This is just very weird usage though for a helper. I'll see if I can ping someone and figure this out.

iothub/service/src/Common/TrackingHelper.cs
vinagesh: I marked this as Obsolete and wanted to make it clear that this is not being used and should not be used. And ideally we would remove it after sometime as it is marked obsolete.

iothub/service/src/Common/TrackingHelper.cs
```diff
@@ -25,17 +33,32 @@ public static class TrackingHelper
...
         public static string GenerateTrackingId(string backendId, string partitionId)
         {
             string gatewayId = GatewayId;
             return GenerateTrackingId(gatewayId, backendId, partitionId);
         }
 
+        /// <summary>
+        /// Generates a unique tracking Id with gateway, backend, partition and timestamp information.
+        /// </summary>
+        /// <returns>Tracking Id.</returns>
```
drwill-ms, line 58: Proper English would be "A tracking Id." or "The tracking Id." In this case, I think 'a' is the right article.

iothub/service/src/MessageSystemPropertyNames.cs
vinagesh: I think it should not be the maximum thinking about it. I should just be the count and the logic to compare against the max should be in the hub.

iothub/service/src/MessageSystemPropertyNames.cs
```diff
@@ -36,6 +36,9 @@ public static class MessageSystemPropertyNames
         /// </summary>
         public const string To = "to";
 
+        /// <summary>
+        /// The maximum number of times a message can transition between the Enqueued and Invisible states.
```
bikamani, line 40: Did they mention what is invisible state? Just wondering.

iothub/service/src/MessageSystemPropertyNames.cs
```diff
@@ -36,6 +36,9 @@ public static class MessageSystemPropertyNames
...
+        /// After the maximum number of transitions, the IoT hub sets the state of the message to dead lettered.
```
drwill-ms, line 41: Compound adjectives are hyphenated: dead-lettered.

drwill-ms (commented):
Waiting on update.

iothub/service/src/MessageSystemPropertyNames.cs
vinagesh: I'm pretty sure the invisible state is for avoiding race conditions. Basically the message will not be available for any action as long as it is invisible.

iothub/service/src/MessageSystemPropertyNames.cs
vinagesh: I have no idea about this and this is ignored for c2d messages (based on my conversation with hub team) so there is no need to add extra info.

iothub/service/src/Common/Data/AmqpErrorMapper.cs
vinagesh: yeah. Ideally this should go away if we can delete the obsolete stuff sometime.

iothub/service/src/Common/EventHubPartitionKeyResolver.cs
vinagesh: I will restore this file till I can figure it out.

iothub/service/src/Common/Data/AmqpErrorMapper.cs
drwill-ms: After a major version, probably.

drwill-ms (approved):

barustum (approved):

PR #1758, opened 2021-01-27
fix(vsts): Update anti-malware task to treat stale signatures as warning

abhipsaMisra:


vinagesh (approved):

barustum (approved):

drwill-ms (approved):
Perhaps add a comment as to why we set it as warning

PR #1759, opened 2021-01-28
fix(iot-service): Add xml comments to public types and methods

abhipsaMisra:


iothub/service/src/Common/Extensions/EnumerableExtensions.cs
```diff
@@ -1,40 +1,32 @@
...
         public static IList<T> ToListSlim<T>(this IEnumerable<T> enumerable, bool requireMutable = false)
         {
-            var list = enumerable as IList<T>;
-            if (list != null && !(requireMutable && list.IsReadOnly))
-            {
-                return list;
-            }
-
-            return enumerable.ToList();
+            return enumerable is IList<T> list && !(requireMutable && list.IsReadOnly)
```
drwill-ms, line 27: if enumable is null, will it fail the "is IList<T>" check? If not, we lost our NRE guard.

iothub/service/src/Common/Extensions/ExceptionExtensions.cs
```diff
@@ -121,16 +162,16 @@ public static string ToStringSlim(this Exception exception)
             return exception.ToString();
         }
 
+        /// <summary>
+        /// Gets the exception ID from the given exception.
```
drwill-ms, line 166: Prefer casing of Id

drwill-ms (approved):

iothub/service/src/Common/Extensions/EnumerableExtensions.cs
abhipsaMisra: this being an extension method "enumerable" will never be null, so I think we're good here

iothub/service/src/Common/Extensions/ExceptionExtensions.cs
abhipsaMisra: will do

iothub/service/src/Common/Extensions/ExceptionExtensions.cs
```diff
@@ -121,16 +162,16 @@ public static string ToStringSlim(this Exception exception)
...
         {
-            if (exception.Data != null && exception.Data.Contains(ExceptionIdentifierName))
-            {
-                return (string)exception.Data[ExceptionIdentifierName];
-            }
-            else
-            {
-                return null;
-            }
+            return exception.Data != null && exception.Data.Contains(ExceptionIdentifierName)
```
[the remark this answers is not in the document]
abhipsaMisra: this is also an extension method, so the null check for "exception" wouldn't add value

PR #1762, opened 2021-01-29
Fix(service):Removed unused field

barustum:
Remove a field not being used in the solution.

### DIFF
--- a/iothub/service/src/IoTHubExceptionResult.cs
+++ b/iothub/service/src/IoTHubExceptionResult.cs
@@ -12,8 +12,5 @@ namespace Microsoft.Azure.Devices
     {
         [JsonProperty("Message")]
         internal string _message;
-
-        [JsonProperty("ExceptionMessage")]
-        internal string _exceptionMessage;
     }
 }


iothub/service/src/IoTHubExceptionResult.cs
```diff
@@ -12,8 +12,5 @@ internal class IoTHubExceptionResult
     {
         [JsonProperty("Message")]
         internal string _message;
-
-        [JsonProperty("ExceptionMessage")]
-        internal string _exceptionMessage;
```
abhipsaMisra, line 17: do we ever serialize/ deserialize this type?

drwill-ms: I asked Basel about that, and he's following up.

drwill-ms: @barustum I just didn't know if this field was included in the response, and if so would it be valuable?

barustum: I don't see us serializing/deserializing this type. I looked at the service logs on Kusto on what is populated here. It's not something useful for the user in my opinion. It's a tracking ID guide concatenated with a time stamp of some sort.

drwill-ms: Is the tracking Id something that would help a customer (giving it to CSS, CSS to service team, service team using it to quickly identify relevant logs)?

abhipsaMisra: do we have an example of what the response from service looks like?

abhipsaMisra: While we do deserialize the response into this type, I see that we never access the "ExceptionMessage": https://github.com/Azure/azure-iot-sdk-csharp/blob/master/common/src/service/ExceptionHandlingHelper.cs#L127
I wonder if that is useful info that we could add to the exception that we return to the calling application.

barustum: I might have lied. I looked closely at the Kusto logs, and the response we get from Visual Studio does not match what they log in the service. I am seeing ExceptionMessage (the field I removed) as empty during my testing. However, I do not know whether it'll always be empty. I am raeching out to someone in the Hub team to confirm whether this ExceptionMessage is populated at a specific type of exception.

bikamani: That's interesting, if it is always empty I see it is safe to remove. Would be good to add a task if that's something for an enhancement later based on what you find from service team.

drwill-ms: I just don't know where we got the notion of that field from, but I don't want us to drop it until we know it isn't useful. If it is useful, we should make sure it makes it's way into the customer-facing exception.

barustum: I checked with Andrei. Seems like safe to ignore. Here's his response:

> According to the code below it’s always empty unless ‘isDebugEnabled’ which is not enabled in Prod:
> ```
> httpError.Message = JsonConvert.SerializeObject(hubError);
> httpError.ExceptionMessage = isDebugEnabled ? exception.ToString() : string.Empty;
> 
> ```
> So, I think you should ignore this property

drwill-ms (approved):

abhipsaMisra (approved):

PR #1763, opened 2021-02-02
Add a log statement to indicate a NoRetry retry policy was enabled

abhipsaMisra:
The ask from the CSS team is as below:
In certain situations customers raise support issues stating that the client does not recover on disconnection, only to discover on inspection of customer code that they have explicitly disabled the inbuilt retry policy in the client.
In order to avoid long delays in identifying the issue, the CSS team wants us to log a statement in case the retry policy is disabled.

This PR adds the following statement if a customer chooses a `NoRetry` retry policy:
`NOTE: A no-retry retry policy has been enabled, the device client will not perform any retries on disconnection.`

In addition, since the sdk has a default retry policy of retrying indefinitely(!), the CSS team can be advised to look for the following information:
1. Advise customers to always set a `ConnectionStatusChangeHandler`. 
   If the customer has disabled retries, or has enabled a custom retry policy with a limited no of retry attempts; they will seea connection status change event returned with a reason `Retry_Expired`: 
   i. `NoRetry` policy:
      Connection status changed: status=Connected, reason=Connection_Ok
      Disconnection
      Connection status changed: status=Disconnected, reason=Retry_Expired
      
   ii. Custom retry policy: (eg.: single retry attempt)
      Connection status changed: status=Connected, reason=Connection_Ok
      Disconnection
      Retry policy's ShouldRetry is = True
      Connection status changed: status=Disconnected_Retrying, reason=Communication_Error
      Retry policy's ShouldRetry is = False
      Connection status changed: status=Disconnected, reason=Retry_Expired
      
2. If, for some reason the customer cannot monitor this status change callback, look for the newly added log statement.

### DIFF
--- a/e2e/test/Helpers/templates/FaultInjection.cs
+++ b/e2e/test/Helpers/templates/FaultInjection.cs
@@ -39,17 +39,17 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
         public const string FaultCloseReason_Boom = "boom";
         public const string FaultCloseReason_Bye = "byebye";
 
-        public const int DefaultDelayInSec = 5; // Time in seconds after service initiates the fault.
-        public const int DefaultDurationInSec = 5; // Duration in seconds
-        public const int LatencyTimeBufferInSec = 10; // Buffer time waiting fault occurs or connection recover
+        public static readonly TimeSpan DefaultFaultDelay = TimeSpan.FromSeconds(5); // Time in seconds after service initiates the fault.
+        public static readonly TimeSpan DefaultFaultDuration = TimeSpan.FromSeconds(5); // Duration in seconds
+        public static readonly TimeSpan LatencyTimeBuffer = TimeSpan.FromSeconds(10); // Buffer time waiting fault occurs or connection recover
 
-        public const int WaitForDisconnectMilliseconds = 3 * DefaultDelayInSec * 1000;
-        public const int WaitForReconnectMilliseconds = 2 * DefaultDurationInSec * 1000;
-        public const int ShortRetryInMilliSec = (int)(DefaultDurationInSec / 2.0 * 1000);
+        public static readonly TimeSpan WaitForDisconnectDuration = TimeSpan.FromTicks(DefaultFaultDelay.Ticks * 3);
+        public static readonly TimeSpan WaitForReconnectDuration = TimeSpan.FromTicks(DefaultFaultDuration.Ticks * 2);
+        public static readonly TimeSpan ShortRetryDuration = TimeSpan.FromTicks(DefaultFaultDuration.Ticks / 2);
 
-        public const int RecoveryTimeMilliseconds = 5 * 60 * 1000;
+        public static readonly TimeSpan RecoveryTime = TimeSpan.FromMinutes(5);
 
-        public static Client.Message ComposeErrorInjectionProperties(string faultType, string reason, int delayInSecs, int durationInSecs)
+        public static Client.Message ComposeErrorInjectionProperties(string faultType, string reason, TimeSpan delayInSecs, TimeSpan durationInSecs)
         {
             string dataBuffer = Guid.NewGuid().ToString();
 
@@ -59,8 +59,8 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
                 {
                     ["AzIoTHub_FaultOperationType"] = faultType,
                     ["AzIoTHub_FaultOperationCloseReason"] = reason,
-                    ["AzIoTHub_FaultOperationDelayInSecs"] = delayInSecs.ToString(CultureInfo.InvariantCulture),
-                    ["AzIoTHub_FaultOperationDurationInSecs"] = durationInSecs.ToString(CultureInfo.InvariantCulture)
+                    ["AzIoTHub_FaultOperationDelayInSecs"] = delayInSecs.Seconds.ToString(CultureInfo.InvariantCulture),
+                    ["AzIoTHub_FaultOperationDurationInSecs"] = durationInSecs.Seconds.ToString(CultureInfo.InvariantCulture)
                 }
             };
         }
@@ -76,9 +76,9 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
         // --------------------------------------------------------------------------------------------------------------------------------------------------------------------
         //  --- device in normal operation --- | FaultRequested | --- <delayInSec> --- | --- Device in fault mode for <durationInSec> --- | --- device in normal operation ---
         // --------------------------------------------------------------------------------------------------------------------------------------------------------------------
-        public static async Task ActivateFaultInjectionAsync(Client.TransportType transport, string faultType, string reason, int delayInSec, int durationInSec, DeviceClient deviceClient, MsTestLogger logger)
+        public static async Task ActivateFaultInjectionAsync(Client.TransportType transport, string faultType, string reason, TimeSpan delayInSec, TimeSpan durationInSec, DeviceClient deviceClient, MsTestLogger logger)
         {
-            logger.Trace($"{nameof(ActivateFaultInjectionAsync)}: Requesting fault injection type={faultType} reason={reason}, delay={delayInSec}s, duration={FaultInjection.DefaultDurationInSec}s");
+            logger.Trace($"{nameof(ActivateFaultInjectionAsync)}: Requesting fault injection type={faultType} reason={reason}, delay={delayInSec}s, duration={DefaultFaultDuration}s");
 
             uint oldTimeout = deviceClient.OperationTimeoutInMilliseconds;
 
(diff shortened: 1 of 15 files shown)


iothub/device/src/RetryPolicies/NoRetry.cs
```diff
@@ -10,6 +11,16 @@ namespace Microsoft.Azure.Devices.Client
...
     public class NoRetry : IRetryPolicy
     {
+        /// <summary>
+        /// Create an instance of a retry policy that perfrms no retries.
+        /// </summary>
+        public NoRetry()
+        {
+            if (Logging.IsEnabled)
+                Logging.Info(this, $"NOTE: A no-retry retry policy has been enabled," +
+                    $" the device client will not perform any retries on disconnection.", nameof(NoRetry));
```
abhipsaMisra, line 21: This is the only line of code being added

e2e/test/Helpers/templates/FaultInjection.cs
```diff
@@ -39,17 +39,17 @@ public static class FaultInjection
         public const string FaultCloseReason_Boom = "boom";
         public const string FaultCloseReason_Bye = "byebye";
 
-        public const int DefaultDelayInSec = 5; // Time in seconds after service initiates the fault.
-        public const int DefaultDurationInSec = 5; // Duration in seconds
-        public const int LatencyTimeBufferInSec = 10; // Buffer time waiting fault occurs or connection recover
+        public static readonly TimeSpan DefaultDelayInSec = TimeSpan.FromSeconds(5); // Time in seconds after service initiates the fault.
```
abhipsaMisra, line 42: Although tangent to the retry policy related changes, I am also updating our E2E tests to use TimeSpan instead of int, to represent time. (I spent longer than I should have trying to identify a test failure, only to realize the test was setting time time in ms, while the service was expecting the time in secs.

e2e/test/iothub/NoRetryE2ETests.cs
```diff
@@ -55,81 +59,100 @@ await FaultInjection
...
+            lastConnectionStatusChangeReason.Should().Be(ConnectionStatusChangeReason.Retry_Expired, $"Expected device to be {ConnectionStatusChangeReason.Retry_Expired} but was {lastConnectionStatusChangeReason}.");
+            connectionStatusChanges.Should().NotContainKey(ConnectionStatus.Disconnected_Retrying, $"Shouldn't get {ConnectionStatus.Disconnected_Retrying} status change.");
             int connected = connectionStatusChanges[ConnectionStatus.Connected];
-            Assert.AreEqual(1, connected, $"Should get {ConnectionStatus.Connected} once but was {connected}.");
+            connected.Should().Be(1, $"Should get {ConnectionStatus.Connected} once but got it {connected} times.");
             int disconnected = connectionStatusChanges[ConnectionStatus.Disconnected];
-            Assert.AreEqual(1, disconnected, $"Should get {ConnectionStatus.Disconnected} once but was {disconnected}.");
+            disconnected.Should().Be(1, $"Should get {ConnectionStatus.Disconnected} once but got it {disconnected} times.");
```
abhipsaMisra, line 81: I wanted to have a similar test for a custom retry policy, however, since we cannot simulate device-client-status-changing-transient-faults for a "duration", we cannot have an E2E test for this. It has been verified through a sample, though.

iothub/device/src/RetryPolicies/NoRetry.cs
timtay-microsoft: nit: wouldn't a module client also hit this code path? Maybe just "... the ~~device~~ client will not..."

timtay-microsoft (approved):

iothub/device/src/RetryPolicies/NoRetry.cs
abhipsaMisra: That's true, I'll update it!

e2e/test/Helpers/templates/FaultInjection.cs
```diff
@@ -39,17 +39,17 @@ public static class FaultInjection
...
 
-        public const int WaitForDisconnectMilliseconds = 3 * DefaultDelayInSec * 1000;
-        public const int WaitForReconnectMilliseconds = 2 * DefaultDurationInSec * 1000;
-        public const int ShortRetryInMilliSec = (int)(DefaultDurationInSec / 2.0 * 1000);
+        public static readonly TimeSpan WaitForDisconnectSeconds = TimeSpan.FromSeconds(15); // 3 * DefaultDelayInSec
+        public static readonly TimeSpan WaitForReconnectSeconds = TimeSpan.FromSeconds(10); // 2 * DefaultDurationInSec
+        public static readonly TimeSpan ShortRetryInSec = TimeSpan.FromSeconds(2.5); // DefaultDurationInSec / 2
 
-        public const int RecoveryTimeMilliseconds = 5 * 60 * 1000;
+        public static readonly TimeSpan RecoveryTimeMilliseconds = TimeSpan.FromMinutes(5);
```
vinagesh, line 50: Should this be renamed to minutes or should the timespan be in milliseconds?

abhipsaMisra: hehe, you are right, this should be renamed!

e2e/test/Helpers/templates/FaultInjection.cs
```diff
@@ -223,19 +230,23 @@ await deviceClient
                 {
                     logger.Trace($"{nameof(FaultInjection)}: Performing test operation while fault injection is being activated.");
                     // Perform the test operation for the faulted device multi times.
-                    for (int i = 0; i < LatencyTimeBufferInSec; i++)
+                    int counter = 0;
+                    var sw = new Stopwatch();
+                    sw.Start();
```
vinagesh, line 235: There is also a one-liner you can use Stopwatch.StartNew()

e2e/test/Helpers/templates/FaultInjection.cs
```diff
@@ -39,17 +39,17 @@ public static class FaultInjection
...
-        public const int DefaultDurationInSec = 5; // Duration in seconds
-        public const int LatencyTimeBufferInSec = 10; // Buffer time waiting fault occurs or connection recover
+        public static readonly TimeSpan DefaultDelayInSec = TimeSpan.FromSeconds(5); // Time in seconds after service initiates the fault.
+        public static readonly TimeSpan DefaultDurationInSec = TimeSpan.FromSeconds(5); // Duration in seconds
+        public static readonly TimeSpan LatencyTimeBufferInSec = TimeSpan.FromSeconds(10); // Buffer time waiting fault occurs or connection recover
 
-        public const int WaitForDisconnectMilliseconds = 3 * DefaultDelayInSec * 1000;
-        public const int WaitForReconnectMilliseconds = 2 * DefaultDurationInSec * 1000;
-        public const int ShortRetryInMilliSec = (int)(DefaultDurationInSec / 2.0 * 1000);
+        public static readonly TimeSpan WaitForDisconnectSeconds = TimeSpan.FromSeconds(15); // 3 * DefaultDelayInSec
```
drwill-ms, line 46: You could do the TimeSpan math if you really wanted to, right?

e2e/test/Helpers/templates/FaultInjection.cs
drwill-ms: Rename it to remove the unit

e2e/test/Helpers/templates/FaultInjection.cs
```diff
@@ -183,15 +183,18 @@ await deviceClient
...
 
                 // For disconnect type faults, the device should disconnect and recover.
                 if (FaultShouldDisconnect(faultType))
                 {
                     logger.Trace($"{nameof(FaultInjection)}: Confirming fault injection has been activated.");
                     // Check that service issued the fault to the faulting device
                     bool isFaulted = false;
-                    for (int i = 0; i < LatencyTimeBufferInSec; i++)
+
+                    var sw = new Stopwatch();
```
drwill-ms, line 195: `Stopwatch.StartNew()`

e2e/test/iothub/NoRetryE2ETests.cs
```diff
@@ -16,32 +18,34 @@ namespace Microsoft.Azure.Devices.E2ETests
     [TestCategory("IoTHub")]
     public class NoRetryE2ETests : E2EMsTestBase
     {
-        private readonly string DevicePrefix = $"E2E_{nameof(NoRetryE2ETests)}_";
+        private readonly string _devicePrefix = $"E2E_{nameof(NoRetryE2ETests)}_";
```
drwill-ms, line 21: static?

drwill-ms (approved):

iothub/device/src/RetryPolicies/NoRetry.cs
```diff
@@ -10,6 +11,16 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public class NoRetry : IRetryPolicy
     {
+        /// <summary>
+        /// Create an instance of a retry policy that perfrms no retries.
+        /// </summary>
+        public NoRetry()
+        {
+            if (Logging.IsEnabled)
+                Logging.Info(this, $"NOTE: A no-retry retry policy has been enabled," +
```
vinagesh, line 20: you can just says 'A no-retry policy has been enabled'

vinagesh (approved):

e2e/test/Helpers/templates/FaultInjection.cs
abhipsaMisra: I actually didn't even think of that!

iothub/device/src/RetryPolicies/NoRetry.cs
abhipsaMisra: I preferred specifying what kind of policy a "no-retry" policy was; do you think it seems redundant?

azabbasi (approved):

PR #1768, opened 2021-02-02
(fix): Build error due to missing dispose

vinagesh:
## Checklist

## Description of the changes
Some implementations of IAuthenticationMethod  have objects that need to be disposed. And we were not disposing because the interface did not have IDisposable and the implementations did not dispose anything. 
This change makes the interface disposable and hence any implementation that has disposable objects, dispose them. It is a no-op for the rest.

This also means, the public IotHubConnectionStringBuilder is now IDisposable. This could be a breaking change but something necessary.

### DIFF
--- a/iothub/device/src/ClientFactory.cs
+++ b/iothub/device/src/ClientFactory.cs
@@ -100,7 +100,9 @@ namespace Microsoft.Azure.Devices.Client
                     throw new ArgumentException("No certificate was found. To use certificate authentication certificate must be present.");
                 }
 
+#pragma warning disable CA2000 // This is returned to client so cannot be disposed here.
                 InternalClient dc = CreateFromConnectionString(connectionStringBuilder.ToString(), PopulateCertificateInTransportSettings(connectionStringBuilder, transportType), options);
+#pragma warning restore CA2000
                 dc.Certificate = connectionStringBuilder.Certificate;
 
                 // Install all the intermediate certificates in the chain if specified.
--- a/iothub/device/src/DeviceAuthenticationWithRegistrySymmetricKey.cs
+++ b/iothub/device/src/DeviceAuthenticationWithRegistrySymmetricKey.cs
@@ -69,8 +69,8 @@ namespace Microsoft.Azure.Devices.Client
                 throw new ArgumentNullException(nameof(iotHubConnectionStringBuilder));
             }
 
-            iotHubConnectionStringBuilder.DeviceId = this.DeviceId;
-            iotHubConnectionStringBuilder.SharedAccessKey = this.KeyAsBase64String;
+            iotHubConnectionStringBuilder.DeviceId = DeviceId;
+            iotHubConnectionStringBuilder.SharedAccessKey = KeyAsBase64String;
             iotHubConnectionStringBuilder.SharedAccessKeyName = null;
             iotHubConnectionStringBuilder.SharedAccessSignature = null;
 
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -1974,11 +1974,17 @@ namespace Microsoft.Azure.Devices.Client
         public void Dispose()
         {
             InnerHandler?.Dispose();
+            InnerHandler = null;
             _methodsDictionarySemaphore?.Dispose();
             _moduleReceiveMessageSemaphore?.Dispose();
             _fileUploadHttpTransportHandler?.Dispose();
             _deviceReceiveMessageSemaphore?.Dispose();
             _twinDesiredPropertySemaphore?.Dispose();
+#if !NET451
+            Certificate?.Dispose();
+            Certificate = null;
+#endif
+            IotHubConnectionString?.TokenRefresher?.Dispose();
         }
 
         internal bool IsE2EDiagnosticSupportedProtocol()
--- a/iothub/device/src/Edge/EdgeModuleClientFactory.cs
+++ b/iothub/device/src/Edge/EdgeModuleClientFactory.cs
@@ -93,7 +93,9 @@ namespace Microsoft.Azure.Devices.Client.Edge
                 }
 
                 ISignatureProvider signatureProvider = new HttpHsmSignatureProvider(edgedUri, DefaultApiVersion);
+#pragma warning disable CA2000 // This is disposed when the client is disposed.
                 var authMethod = new ModuleAuthenticationWithHsm(signatureProvider, deviceId, moduleId, generationId);
+#pragma warning restore CA2000
 
                 Debug.WriteLine("EdgeModuleClientFactory setupTrustBundle from service");
 


iothub/device/src/IAuthenticationMethod.cs
```diff
@@ -1,12 +1,14 @@
...
 
+using System;
+
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// Authentication interface to use for device communications.
     /// </summary>
-    public interface IAuthenticationMethod
+    public interface IAuthenticationMethod : IDisposable
```
vinagesh, line 11: Some of the implementations of this interface are disposable. There was no other better way to enforce disposing objects.

iothub/device/src/IotHubConnectionStringBuilder.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.Devices.Client
     /// <summary>
     /// Builds a connection string for the IoT Hub service based on the properties populated by the user.
     /// </summary>
-    public sealed class IotHubConnectionStringBuilder
+    public sealed class IotHubConnectionStringBuilder : IDisposable
```
vinagesh, line 19: This is now IDisposable and could be breaking change but something necessary. I want to call this out so that we can make a decision otherwise if necessary.

iothub/device/src/DeviceAuthenticationWithRegistrySymmetricKey.cs
```diff
@@ -106,5 +106,12 @@ private void SetDeviceId(string deviceId)
 
             _deviceId = deviceId;
         }
+
+        /// <summary>
```
drwill-ms, line 110: Change to `<inheritdoc />`

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
```diff
@@ -48,6 +48,16 @@ public string DeviceId
         /// </summary>
         public X509Certificate2Collection ChainCertificates { get; }
 
+        /// <summary>
+        /// Dispose necessary objects.
+        /// </summary>
+        public void Dispose()
+        {
+#if !NET451
+            Certificate?.Dispose();
```
drwill-ms, line 57: prefer this approach to avoid disposing twice:

```csharp
Certificate?.Dispose();
Certificate = null;
```

iothub/device/src/IotHubConnectionStringBuilder.cs
drwill-ms: It is an additive change.

drwill-ms (approved):

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
vinagesh: Yes thanks for reminding of the best practice. Will do it everywhere.

iothub/device/src/ClientFactory.cs
```diff
@@ -100,7 +100,7 @@ public static InternalClient Create(string hostname, string gatewayHostname, IAu
                     throw new ArgumentException("No certificate was found. To use certificate authentication certificate must be present.");
                 }
 
-                InternalClient dc = CreateFromConnectionString(connectionStringBuilder.ToString(), PopulateCertificateInTransportSettings(connectionStringBuilder, transportType), options);
+                using InternalClient dc = CreateFromConnectionString(connectionStringBuilder.ToString(), PopulateCertificateInTransportSettings(connectionStringBuilder, transportType), options);
```
abhipsaMisra, line 103: this client shouldn't be disposed here, it is returned to the caller, and they should dispose it

vinagesh: Makes sense. So I guess we need to ignore the warning in this case.

abhipsaMisra:
We'll also need to update AuthenticationWithTokenRefresh to have it not implement IDisposable again

vinagesh:
> AuthenticationWithTokenRefresh

Will remove

azureiot.sln
```diff
@@ -12,6 +12,9 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "security", "security", "{96531499-C8F4-4F77-88DD-9F9585B5E565}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "e2e", "e2e", "{662CE01D-7A7F-48D8-80CE-5DC9C60677E8}"
+	ProjectSection(SolutionItems) = preProject
```
vinagesh, line 15: Not sure how this came here. I will take a look.

azureiot.sln
```diff
@@ -12,6 +12,9 @@ EndProject
...
+		iothub\device\src\DeviceAuthenticationWithRegistrySymmetricKey.cs = iothub\device\src\DeviceAuthenticationWithRegistrySymmetricKey.cs
+	EndProjectSection
```
abhipsaMisra, line 17: this looks like a testing change that went in accidentally

vinagesh: Oh I commented saying it might have come in accidentally and has been fixed since.

iothub/device/src/ClientFactory.cs
```diff
@@ -88,7 +88,7 @@ public static InternalClient Create(string hostname, string gatewayHostname, IAu
                 throw new ArgumentException("Certificate chains are only supported on Amqp_Tcp_Only and Mqtt_Tcp_Only");
             }
 
-            IotHubConnectionStringBuilder connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
+            using IotHubConnectionStringBuilder connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
```
abhipsaMisra, line 91: I don't think disposing this object here is going to work:
- we have an instance of IotHubConnectionStringBuilder, which contains a disposable property "Certificate".
- we set internalClient.Certificate to point to the same instance of Certificate as in the IotHubConnectionStringBuilder.
- we return the InternalClient instance to the caller, but dispose the IotHubConnectionStringBuilder once this method exits (=> the Certificate is also disposed).
- Now we have an internal client, which contains a reference to a Certificate that is already disposed.

iothub/device/src/ModuleAuthenticationWithRegistrySymmetricKey.cs
```diff
@@ -128,5 +128,11 @@ private void SetModuleId(string moduleId)
 
             _moduleId = moduleId;
         }
+
+        /// <inheritdoc />
```
bikamani, line 132: 👍  something new for me to learn :)

bikamani (approved):

iothub/device/src/ClientFactory.cs
vinagesh: Hmm interesting. I looked at who was using the builder but not the certificate. Who can dispose the builder then! The customer surely won't. Maybe needs to be disposed by the InternalClient and we hold a ref to it?

vinagesh: Ok this was way more complicated than this. Worked offline with @abhipsaMisra and now we have something that seems correct.

vinagesh: Will be testing this against the full pipeline before checking in.

vinagesh:
Tested on the main pipeline and everything works well now.

drwill-ms (approved):

iothub/device/src/InternalClient.cs
```diff
@@ -1974,11 +1974,17 @@ internal void ValidateModuleTransportHandler(string apiName)
...
             _methodsDictionarySemaphore?.Dispose();
             _moduleReceiveMessageSemaphore?.Dispose();
             _fileUploadHttpTransportHandler?.Dispose();
             _deviceReceiveMessageSemaphore?.Dispose();
             _twinDesiredPropertySemaphore?.Dispose();
+#if !NET451
+            Certificate?.Dispose();
+            Certificate = null;
+#endif
+            IotHubConnectionString?.TokenRefresher?.Dispose();
```
vinagesh, line 1987: I decided to dispose whatever is needed here. Making the IotHubConnectionString class did not seem right as we really do not want that dispose to be called anywhere else.

PR #1769, opened 2021-02-03
(service-client): Design for IoT hub AAD authentication

vinagesh:
## Checklist

## Description of the changes
The final design after getting inputs from SDK board. I will remove the design.md file after implementation.

**SDK board inputs:**

We have recently added a class, AzureSasCredential in Azure.Core for service clients  which support sas token authentication to use. You’ll see that the type is fairly simple, really just a string “Signature” with a method to allow updating it without recreating the client. The intended usage of this would be to add overloaded constructors which take the AzureSasCredential in addition to those added for TokenCredentia support. 

Note that these are distinct classes so that the client can be prescriptive through it’s constructors about what types of credentials can authenticate the client. Also this serves the purpose of informing the client what type of credential the user is passing so it can properly authenticate the call with. 

You would have two constructors on the client:

ServiceClient(hostName, TokenCredential credential)

ServiceClient(hostName, AzureSasCredential credential)

### DIFF
--- a/iothub/service/design.md
+++ b/iothub/service/design.md
@@ -1,0 +1,119 @@
+# RBAC API Design for IoT hub
+
+## Service Client
+
+```csharp
+/// <summary>
+/// Creates a <see cref="ServiceClient"/> using AAD token credentials and the specified transport type.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">AAD <see cref="TokenCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportType">Specifies whether Amqp or Amqp_WebSocket_Only transport is used.</param>
+/// <param name="transportSettings">Specifies the AMQP_WS and HTTP proxy settings for service client.</param>
+/// <param name="options">The options that allow configuration of the service client instance during initialization.</param>
+/// <returns>An instance of <see cref="ServiceClient"/>.</returns>
+public static ServiceClient Create(
+    string hostName,
+    TokenCredential credential,
+    TransportType transportType,
+    ServiceClientTransportSettings transportSettings = default,
+    ServiceClientOptions options = default)
+
+/// <summary>
+/// Creates a see <see cerf="ServiceClient"/> using SAS token credentials and the specified transport type.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential"><see cref="AzureSasCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportType">Specifies whether Amqp or Amqp_WebSocket_Only transport is used.</param>
+/// <param name="transportSettings">Specifies the AMQP_WS and HTTP proxy settings for service client.</param>
+/// <param name="options">The options that allow configuration of the service client instance during initialization.</param>
+/// <returns>An instance of <see cref="ServiceClient"/>.</returns>
+public static ServiceClient Create(
+    string hostName,
+    AzureSasCredential credential,
+    TransportType transportType,
+    ServiceClientTransportSettings transportSettings = default,
+    ServiceClientOptions options = default)
+```
+
+## Registry Manager
+
+```csharp
+/// <summary>
+/// Creates an instance of <see cref="RegistryManager"/>.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">AAD <see cref="TokenCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportSettings">The HTTP transport settings.</param>
+/// <returns>An instance of <see cref="RegistryManager"/>.</returns>
+public static RegistryManager Create(
+    string hostName,
+    TokenCredential credential,
+    HttpTransportSettings transportSettings = default)
+
+/// <summary>
+/// Creates an instance of <see cref="RegistryManager"/>.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential"><see cref="AzureSasCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportSettings">The HTTP transport settings.</param>
+/// <returns>An instance of <see cref="RegistryManager"/>.</returns>
+public static RegistryManager Create(
+    string hostName,
+    AzureSasCredential credential,
+    HttpTransportSettings transportSettings = default)
+```
+
+## Job Client
+
+```csharp
+/// <summary>
+/// Creates an instance of <see cref="JobClient"/>.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">AAD <see cref="TokenCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportSettings">The HTTP transport settings.</param>
+/// <returns>An instance of <see cref="JobClient"/>.</returns>
+public static JobClient Create(
+    string hostName,
+    TokenCredential credential,
+    HttpTransportSettings transportSettings = default)
+
+/// <summary>
+/// Creates an instance of <see cref="JobClient"/>.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential"><see cref="TokenCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportSettings">The HTTP transport settings.</param>
+/// <returns>An instance of <see cref="JobClient"/>.</returns>
+public static JobClient Create(
+    string hostName,
+    AzureSasCredential credential,
+    HttpTransportSettings transportSettings = default)
+```
+
+## Digital Twin Client
+
+```csharp
+/// <summary>
+/// Initializes a new instance of the <see cref="DigitalTwinClient"/> class.</summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">AAD <see cref="TokenCredential"> to authenticate with IoT hub.</param>
+/// <param name="handlers">The delegating handlers to add to the http client pipeline. You can add handlers for tracing, implementing a retry strategy, routing requests through a proxy, etc.</param>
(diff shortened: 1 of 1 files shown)


iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
...
+/// <returns>An instance of <see cref="ServiceClient"/>.</returns>
+public static ServiceClient Create(
+    string hostName,
+    TokenCredential credential,
+    TransportType transportType,
+    ServiceClientTransportSettings transportSettings = default,
+    ServiceClientOptions options = default)
+
+/// <summary>
+/// Create ServiceClient using AAD token credentials and the specified Transport Type.
```
abhipsaMisra, line 23: Create ServiceClient using ~AAD~ SAS token credentials and the specified Transport Type.

abhipsaMisra: A note for future - we'll need to add a sample that shows people how to generate sas tokens from a connection string (similar to what we use internally in the client)

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
...
+/// <param name="transportSettings">Specifies the AMQP and HTTP proxy settings for Service Client.</param>
+/// <param name="options">The options that allow configuration of the service client instance during initialization.</param>
+/// <returns>An instance of <see cref="ServiceClient"/>.</returns>
+public static ServiceClient Create(
+    string hostName,
+    AzureSasCredential credential,
+    TransportType transportType,
+    ServiceClientTransportSettings transportSettings = default,
+    ServiceClientOptions options = default)
+```
```
abhipsaMisra, line 37: Q - What about the "ConnectionStringCredential" factory method (the one that'll let people specify the ttl for their sas tokens)? Is that a stretch goal?

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
...
+    TokenCredential credential,
+    TransportType transportType,
+    ServiceClientTransportSettings transportSettings = default,
+    ServiceClientOptions options = default)
+
+/// <summary>
+/// Create ServiceClient using AAD token credentials and the specified Transport Type.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">SAS <see cref="AzureSasCredential"> to authenticate with IoT hub.</param>
```
abhipsaMisra, line 26: do we need to specify "sas" again (it is a part of the type)?

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
...
+## Service Client
+
+```csharp
+/// <summary>
+/// Create ServiceClient using AAD token credentials and the specified Transport Type.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">AAD <see cref="TokenCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportType">Specifies whether Amqp or Amqp_WebSocket_Only transport is used.</param>
+/// <param name="transportSettings">Specifies the AMQP and HTTP proxy settings for Service Client.</param>
```
abhipsaMisra, line 12: I know that this is from existing code comment, but do we want to specify which operations use http on the service client? This is only because in the transportType above we ask about amqp, but then in proxy settings we talk about both amqp and http.
I know that this is such a bad design, and looking at it for the first time confused me; so I wanted to know what you think about it.

iothub/service/design.md
vinagesh: I was on the fence on this one. It might be redundant but want to see what others feel.

iothub/service/design.md
vinagesh: I think it might be too much to specify that info here in comments. I'm thinking how about we create a .md that talks about it in detail and just link it here?

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
+# RBAC API Design for IoT hub
+
+## Service Client
+
+```csharp
+/// <summary>
+/// Create ServiceClient using AAD token credentials and the specified Transport Type.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">AAD <see cref="TokenCredential"> to authenticate with IoT hub.</param>
```
abhipsaMisra, line 10: TokenCredential being an abstract class, we should specify that the user needs to pass in an implementation of a TokenCredentail

abhipsaMisra: From our ADT client: "The <see cref="TokenCredential"/> implementation which will be used to request for the authentication token."

iothub/service/design.md
vinagesh: That should be user defined in their implementation. If they have a one time ttl, the token will only be valid for a while. If they have implementation to renew the token, our SDK will automatically pick up the updated token. Apart from that, our existing ConnectionString input support continues to exist.

iothub/service/design.md
vinagesh: I can add that as a task on my card.

abhipsaMisra: Works for me, it might be a lot of info for doc comments, true. 
small nit: update AMQP to AMQP_WS

iothub/service/design.md
vinagesh: It is obvious as it is an abstract class so I did not add that.

abhipsaMisra (approved):
Lgtm! We should make a note to provide a sample on how customers should use AzureSasCredentail from connection string (and use it to cache and renew tokens), but otherwise looks good!

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
...
+/// <summary>
+/// Creates an instance of <see cref="RegistryManager"/>.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">AAD <see cref="TokenCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportSettings">The HTTP transport settings.</param>
+/// <returns>An instance of <see cref="RegistryManager"/>.</returns>
+public static RegistryManager Create(
+    string hostName,
+    TokenCredential credential,
```
timtay-microsoft, line 51: Maybe nitpicky, but to me this argument is more of a "credentialProvider" or "authenticationTokenProvider" rather than a "credential". What do you think though?

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
...
+/// Creates an instance of <see cref="RegistryManager"/>.
+/// </summary>
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential"><see cref="AzureSasCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportSettings">The HTTP transport settings.</param>
+/// <returns>An instance of <see cref="RegistryManager"/>.</returns>
+public static RegistryManager Create(
+    string hostName,
+    AzureSasCredential credential,
+    HttpTransportSettings transportSettings)
```
timtay-microsoft, line 64: Is there a default HttpTransportSettings we could have here? For users who don't use proxies, it would simplify their code a bit

timtay-microsoft (approved):

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
+# RBAC API Design for IoT hub
+
+## Service Client
+
+```csharp
+/// <summary>
+/// Create ServiceClient using AAD token credentials and the specified Transport Type.
```
drwill-ms, line 7: you can use lower case when referring to things generically, like transport type.

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
...
+/// <param name="hostName">IoT hub host name.</param>
+/// <param name="credential">AAD <see cref="TokenCredential"> to authenticate with IoT hub.</param>
+/// <param name="transportType">Specifies whether Amqp or Amqp_WebSocket_Only transport is used.</param>
+/// <param name="transportSettings">Specifies the AMQP and HTTP proxy settings for Service Client.</param>
+/// <param name="options">The options that allow configuration of the service client instance during initialization.</param>
+/// <returns>An instance of <see cref="ServiceClient"/>.</returns>
+public static ServiceClient Create(
+    string hostName,
+    TokenCredential credential,
+    TransportType transportType,
```
drwill-ms, line 18: Did we find because of the types of TokenCredentials available that we won't need this enum anymore?

iothub/service/design.md
drwill-ms: If that's what it is, why is `TokenCredential` named as it is?

In our track2 ADT code we call this `credential` too. Why not stick with that standard?

iothub/service/design.md
timtay-microsoft: This is a different enum than the TokenCredentialType that you are thinking of. This transport type is to choose between AMQP and AMQP_WS

iothub/service/design.md
timtay-microsoft: > If that's what it is, why is TokenCredential named as it is?

No clue. I can't say that I like it's name, to be honest.

> In our track2 ADT code we call this credential too. Why not stick with that standard?

I can get on board with this point. Since we plan on using implementations from the Azure Identity library anyways, we may as well follow their naming patterns

iothub/service/design.md
vinagesh: Good point. I don't know so I'll find out.

iothub/service/design.md
drwill-ms: Oops, you are right. :)

drwill-ms (approved):

vinagesh:
> Lgtm! We should make a note to provide a sample on how customers should use AzureSasCredentail from connection string (and use it to cache and renew tokens), but otherwise looks good!

Yes ofcourse! Those are already tasks on my story.

iothub/service/design.md
vinagesh: Yes there is, so made it optional. Thanks for calling this out.

bikamani (approved):

PR #1771, opened 2021-02-04
Amqp connections are not getting disposed

### LINKED ISSUE #1770: DeviceClient.Dispose() does not close socket connections
I'm using Microsoft.Azure.Devices.Client 1.34.0 (the latest). It looks like this issue is back: https://github.com/Azure/azure-iot-sdk-csharp/issues/1098

To put it short: when DeviceClient.Dispose() is called, underlying socket connections remain open. Those are to our IoT Hub, port 6571.

This started to happen after updating from 1.33.1 and is wreaking havoc on our service. I see that 1.34.0 changelog has a bunch of "dispose"-related fixes, so one of those might be to blame here.

jamdavi:
Fixes #1770 by reverting regression introduced in [this change](https://github.com/Azure/azure-iot-sdk-csharp/commit/911923977c7ce279e2c73740d7160b3104231a3c#diff-305c99963184ff9db1704e4c76120b5c6cad3d959d8aa959f6787b25c8ec5381R550).

### DIFF
--- a/iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
+++ b/iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
@@ -548,17 +548,12 @@ namespace Microsoft.Azure.Devices.Client.Transport.Amqp
             {
                 base.Dispose(disposing);
 
-                if (_disposed)
-                {
-                    return;
-                }
-
                 Logging.Info(this, $"{nameof(disposing)}");
 
                 if (disposing)
                 {
                     _closed = true;
-                    AmqpUnitManager.GetInstance().RemoveAmqpUnit(_amqpUnit);
+                    AmqpUnitManager.GetInstance()?.RemoveAmqpUnit(_amqpUnit);
                     _disposed = true;
                 }
             }
--- a/iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
+++ b/iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
@@ -546,6 +546,12 @@ namespace Microsoft.Azure.Devices.Client.Transport.Amqp
         {
             lock (_lock)
             {
+
+                if (_disposed)
+                {
+                    return;
+                }
+
                 base.Dispose(disposing);
 
                 Logging.Info(this, $"{nameof(disposing)}");


iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -548,17 +548,12 @@ protected override void Dispose(bool disposing)
             {
                 base.Dispose(disposing);
 
-                if (_disposed)
```
drwill-ms, line 551: Why is this a problem?

drwill-ms (approved):

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -548,17 +548,12 @@ protected override void Dispose(bool disposing)
...
-                }
-
                 Logging.Info(this, $"{nameof(disposing)}");
 
                 if (disposing)
                 {
                     _closed = true;
-                    AmqpUnitManager.GetInstance().RemoveAmqpUnit(_amqpUnit);
+                    AmqpUnitManager.GetInstance()?.RemoveAmqpUnit(_amqpUnit);
                     _disposed = true;
```
azabbasi, line 557: if we never use the _disposed field, we should just remove it.

azabbasi:
Fixes #1770

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
abhipsaMisra: +1. We seem to set _disposed to true only after removing the amqp unit, so it shouldn't cause leaks, right?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
azabbasi: nvm. Talked to James offline and he explained it to me. 
Based on our conversation, we can still keep the if(_disposed) check if we move it above the base.Dispose() call.

azabbasi (approved):

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
vinagesh: @abhipsaMisra - Are you not OOF :D Take rest!

azabbasi: _disposed is a protected field from the base class. That's what's throwing us all off

azabbasi: when we call base.Dispose(), the _disposed is getting set to false, then we bail out prematurely.

jamdavi: I moved the dispose to below the if statement to match the pattern of the other handlers.

vinagesh (approved):

vinagesh:
Can we add anything to catch this ? E2E test? Not this PR but after this.

jamdavi:
> Can we add anything to catch this before customers? E2E test? Not this PR but after this.

We could add some monitoring to long haul. To do this in an integration test we'd need to mark the `_amqpUnit` member as internal to check to see if it's still part of the `AmqpUnitManager` and if it's disposed. 

I'll put a PR out for that to see what everyone thinks.

vinagesh:
> > Can we add anything to catch this before customers? E2E test? Not this PR but after this.
> 
> We could add some monitoring to long haul. To do this in an integration test we'd need to mark the `_amqpUnit` member as internal to check to see if it's still part of the `AmqpUnitManager` and if it's disposed.
> 
> I'll put a PR out for that to see what everyone thinks.

Awesome! We can see how the test looks but we should definitely add some monitoring to long haul as we did not catch it when we released this time and this seems to be something that we should cover as a basic functionality.

bikamani (approved):

abhipsaMisra (approved):
Everyone is on board! 😄

PR #1772, opened 2021-02-05
Update MqttTransportHandler to not use SemaphoreSlimWaitAsync(TimeSpan, CancellationToken)

abhipsaMisra:
It is generally not a good idea to combine Timeouts and cancellation tokens when used with Semaphores. This is because either the semaphore is gained/ not gained but timed out (returns false)/ not gained but cancelled (throws OperationCencelledException). All three scenarios will need to be handled.

For our code, since the only behavior we needs is:
- cancellation token valid -> wait to acquire the semaphore
- cancellation token expired -> exit

it makes more sense to use the SemaphoreSlim.WaitAsync(CancellationToken) API.
This PR doesn't solve a bug but rather encourages usage of a "better suited" API.

Note, since our code intends for the semaphore to wait indefinitely (-1):
- we don't need to handle the "false" returning scenario
- the above issue won't cause "race" conditions for us. Else, using both timeouts and cancellation tokens for awaiting for a semaphore will always give rise to "race" conditions => if the timeout and cancellation token both end (or expire) at the same time, we result could either be "false" or an exception, depending on the internal state of the thread pool.

Another fix is in this PR is:
Dotnetty uses a custom task scheduler to bind tasks to the corresponding EventLoop. To limit I/O to the EventLoop, it requires that tasks scheduled are resumed by the same thread => using ConfigureAwait(true). 
For this reason, calls to dotnetty library are made with ConfigureAwait(true)*. Other tasks are configured with ConfigureAwait(false), thus allowing completed tasks to be resumed on any available thread.
* = getting clarification on this from Max.

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
     // EventLoop. To limit I/O to the EventLoopGroup and keep Netty semantics, we are going to ensure that the
     // task continuations are executed by this scheduler using ConfigureAwait(true).
     //
-    // All await calls that happen within dotnetty's pipeline should be ConfigureAwait(true).
+    // All awaited calls that happen within dotnetty's pipeline should be ConfigureAwait(true).
     //
     internal sealed class MqttIotHubAdapter : ChannelHandlerAdapter
     {
--- a/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
@@ -29,6 +29,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
     // EventLoop. To limit I/O to the EventLoopGroup and keep Netty semantics, we are going to ensure that the
     // task continuations are executed by this scheduler using ConfigureAwait(true).
     //
+    // All await calls that happen within dotnetty's pipeline should be ConfigureAwait(true).
+    //
     internal sealed class MqttIotHubAdapter : ChannelHandlerAdapter
     {
         [Flags]
@@ -475,7 +477,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                         Logging.Info(context, $"Idle time was {idleTime}, so ping request was sent.", nameof(PingServerAsync));
 
                     // Wait to capture the ping response semaphore, which is released when a PINGRESP packet is received.
-                    bool receivedPingResponse = await s_pingResponseSemaphore.WaitAsync(s_pingResponseTimeout).ConfigureAwait(false);
+                    bool receivedPingResponse = await s_pingResponseSemaphore.WaitAsync(s_pingResponseTimeout).ConfigureAwait(true);
                     if (!receivedPingResponse)
                     {
                         if (Logging.IsEnabled)
@@ -1162,7 +1164,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
             int length = (int)streamLength;
             IByteBuffer buffer = context.Channel.Allocator.Buffer(length, length);
-            await buffer.WriteBytesAsync(payloadStream, length).ConfigureAwait(false);
+            await buffer.WriteBytesAsync(payloadStream, length).ConfigureAwait(true);
             Contract.Assert(buffer.ReadableBytes == length);
 
             packet.Payload = buffer;
@@ -1361,7 +1363,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         {
             try
             {
-                await context.WriteAndFlushAsync(message).ConfigureAwait(false);
+                await context.WriteAndFlushAsync(message).ConfigureAwait(true);
             }
             catch (Exception ex)
             {
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -281,10 +281,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                         await SubscribeCloudToDeviceMessagesAsync().ConfigureAwait(true);
                     }
 
-                    bool hasMessage = await ReceiveMessageArrivalAsync(cancellationToken).ConfigureAwait(true);
-                    Message message = ProcessMessage(hasMessage);
-
-                    return message;
+                    await WaitUntilC2dMessageArrivesAsync(cancellationToken).ConfigureAwait(false);
+                    return ProcessMessage();
                 }
                 finally
                 {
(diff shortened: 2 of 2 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -281,10 +281,8 @@ public override async Task<Message> ReceiveAsync(CancellationToken cancellationT
                         await SubscribeCloudToDeviceMessagesAsync().ConfigureAwait(true);
                     }
 
-                    bool hasMessage = await ReceiveMessageArrivalAsync(cancellationToken).ConfigureAwait(true);
-                    Message message = ProcessMessage(hasMessage);
-
-                    return message;
+                    await WaitUntilC2dMessageArrivesAsync(cancellationToken).ConfigureAwait(false);
```
drwill-ms, line 284: `_ =` if we don't care about the return value.

Why don't we care about it?

drwill-ms, line 284: ConfigureAwait true or false?

drwill-ms (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: I've updated the calls to dotnetty to have ConfigureAwait(true), and rest others following ConfigureAwait(false). I've also reached out to Max to get a confirmation on the reasoning behind this.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -363,20 +362,20 @@ private Message ProcessMessage(bool hasMessage)
...
             cancellationToken.ThrowIfCancellationRequested();
             CancellationToken disconnectToken = _disconnectAwaitersCancellationSource.Token;
             EnsureValidState();
 
             using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, disconnectToken);
 
-            // -1 millisecond represents for SemaphoreSlim to wait indefinitely until either of the linked cancellation tokens have been canceled.
-            return await _receivingSemaphore.WaitAsync(TimeSpan.FromMilliseconds(-1), linkedCts.Token).ConfigureAwait(true);
+            // Wait until either of the linked cancellation tokens have been canceled.
+            await _receivingSemaphore.WaitAsync(linkedCts.Token).ConfigureAwait(false);
```
bikamani, line 378: Does the behavior change here? If so, how? Or is this just a better way of waiting until token expires?

abhipsaMisra: The behavior doesn't change here, the semaphore still waits until it receives a message or either of the cancellation tokens are cancelled. I've explained the behavior in the PR description, let me know if it is unclear and I can update it.

abhipsaMisra: The OperationCancelledException is handled a bit differently based on where it is thrown from. For all protocols, if the OperationCancelledException is thrown as a result of cancellation of the passed cancellation token, it will get thrown back to the caller. For cancellation of the cancellation token generated from the default operation timeout (i.e. from the timeout API or no param API), the API will return a "null" message (this is because it wouldn't make sense to throw an OperationCancelledException from an API that doesn't accept a cancellation token.)

As far as the semaphore on MQTT is concerned, it doesn't really know if this is a user supplied cancellation token or the cancellation token generated from the default operation timeout; so it will simply throw the OperationCancelledException on cancellation, and delegate the handling of the exception to its caller.

bikamani: Got it, in that case the behavior of ReceiveAsync taking timeout as a parameter is changed correct?

bikamani: Synced offline, as per which the behavior of public API remains same.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -29,6 +29,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
     // EventLoop. To limit I/O to the EventLoopGroup and keep Netty semantics, we are going to ensure that the
     // task continuations are executed by this scheduler using ConfigureAwait(true).
     //
+    // All await calls that happen within dotnetty's pipeline should be ConfigureAwait(true).
```
drwill-ms, line 32: awaited

drwill-ms (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -333,28 +337,25 @@ public override async Task<Message> ReceiveAsync(TimeoutHelper timeoutHelper)
...
-        private Message ProcessMessage(bool hasMessage)
+        private Message ProcessMessage()
         {
             Message message = null;
 
             try
             {
                 if (Logging.IsEnabled)
-                    Logging.Enter(this, message, $"hasMessage={hasMessage}", nameof(ProcessMessage));
+                    Logging.Enter(this, message, $"Will begin processing received C2D message, queue size={_messageQueue.Count}", nameof(ProcessMessage));
```
bikamani, line 347: cloud-to-device

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -363,20 +364,20 @@ private Message ProcessMessage(bool hasMessage)
             finally
             {
                 if (Logging.IsEnabled)
-                    Logging.Exit(this, message, $"hasMessage={hasMessage}", nameof(ProcessMessage));
+                    Logging.Exit(this, message, $"Processed received C2D message with Id={message?.MessageId}", nameof(ProcessMessage));
```
bikamani, line 367: cloud-to-device

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -554,8 +555,12 @@ private async Task HandleIncomingMessagesAsync()
             if (Logging.IsEnabled)
                 Logging.Enter(this, "Process C2D message via callback", nameof(HandleIncomingMessagesAsync));
 
-            Message message = ProcessMessage(true);
-            await (_deviceMessageReceivedListener?.Invoke(message) ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
+            Message message = ProcessMessage();
+
+            // We are intentionally not awaiting _deviceMessageReceivedListener callback.
+            // This is a user-supplied callback that isn't required to be awaited by us. We can simply invoke it and continue.
+            _ = _deviceMessageReceivedListener?.Invoke(message);
```
bikamani, line 562: can discard assignment and rather just _deviceMessageReceivedListener?.Invoke(message);

Correct me if I am missing something.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: this is an internal log, so I don't actually mind leaving it as is

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: you could, yes, but this is to highlight that the function "_deviceMessageReceivedListener" returns a task but we intentionally don't await it

PR #1774, opened 2021-02-05
Fix issue where we accidentally disposed the user supplied certificates

abhipsaMisra:
The X509 certificate being passed during client initialization should be disposed by the user. If we dispose it when the client is disposed, they won't be able to reuse the previously created IAuthenticationMethod (which in turn has a reference to these certificates).
Note: IAuthenticationMethod should probably have been IDisposable, but isn't, and updating it _might_ require a bunch of refactoring. Certificates being the only IDisposable field in there, it feels like a reasonable solution to ask users to manage its lifecycle.

Edit: Spoke with @vinagesh offline, making the X509 specific implementation of IAuthenticationMethod requires much less of a refactor while still indicating to the user that it has IDisposable types within it.

### DIFF
--- a/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
+++ b/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
@@ -383,8 +383,50 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
         [LoggedTestMethod]
         public async Task DeviceDoesNotReceivePendingMessageUsingCallback_MqttWs()
         {
-            var settings = new ITransportSettings[] { new MqttTransportSettings(Client.TransportType.Mqtt_Tcp_Only) { CleanSession = true } };
+            var settings = new ITransportSettings[] { new MqttTransportSettings(Client.TransportType.Mqtt_WebSocket_Only) { CleanSession = true } };
             await DoNotReceiveMessagesSentBeforeSubscriptionAsync(TestDeviceType.Sasl, settings).ConfigureAwait(false);
+        }
+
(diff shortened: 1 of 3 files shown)


e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
```diff
@@ -387,6 +387,48 @@ public async Task DeviceDoesNotReceivePendingMessageUsingCallback_MqttWs()
...
+        public async Task X509_DeviceDoesNotReceivePendingMessageUsingCallback_Mqtt()
+        {
+            var settings = new ITransportSettings[] { new MqttTransportSettings(Client.TransportType.Mqtt_Tcp_Only) { CleanSession = true } };
+            await DoNotReceiveMessagesSentBeforeSubscriptionAsync(TestDeviceType.X509, settings).ConfigureAwait(false);
+        }
+
+        [LoggedTestMethod]
+        public async Task X509_DeviceDoesNotReceivePendingMessageUsingCallback_MqttWs()
+        {
+            var settings = new ITransportSettings[] { new MqttTransportSettings(Client.TransportType.Mqtt_Tcp_Only) { CleanSession = true } };
```
barustum, line 414: Mqtt_WebSocket_Only

barustum (approved):

e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
abhipsaMisra: Good catch!

vinagesh (approved):

PR #1777, opened 2021-02-09
[DO_NOT_REVIEW](fix): Refactor properties on service client IotHubconnectionString to add a new auth type

vinagesh:
## Checklist

### DIFF
--- a/common/src/service/IotHubConnectionString.cs
+++ b/common/src/service/IotHubConnectionString.cs
@@ -11,12 +11,15 @@ using Microsoft.Azure.Devices.Common.Security;
 
 namespace Microsoft.Azure.Devices
 {
-    internal sealed class IotHubConnectionString : IAuthorizationHeaderProvider, ICbsTokenProvider
+    /// <summary>
+    /// The IoT hub connection properties and methods required for authenticating using connection strings.
+    /// </summary>
+    internal sealed class IotHubConnectionString : IotHubCredential
     {
-        private static readonly TimeSpan s_defaultTokenTimeToLive = TimeSpan.FromHours(1);
         private const char UserSeparator = '@';
 
         public IotHubConnectionString(IotHubConnectionStringBuilder builder)
+            : base(builder.HostName)
         {
             if (builder == null)
             {
@@ -24,25 +27,13 @@ namespace Microsoft.Azure.Devices
             }
 
             Audience = builder.HostName;
-            HostName = string.IsNullOrEmpty(builder.GatewayHostName) ? builder.HostName : builder.GatewayHostName;
             SharedAccessKeyName = builder.SharedAccessKeyName;
             SharedAccessKey = builder.SharedAccessKey;
             SharedAccessSignature = builder.SharedAccessSignature;
-            IotHubName = builder.IotHubName;
-            HttpsEndpoint = new UriBuilder("https", HostName).Uri;
-            AmqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, builder.HostName, AmqpConstants.DefaultSecurePort).Uri;
             DeviceId = builder.DeviceId;
             ModuleId = builder.ModuleId;
             GatewayHostName = builder.GatewayHostName;
         }
-
-        public string IotHubName { get; private set; }
-
-        public string HostName { get; private set; }
-
-        public Uri HttpsEndpoint { get; private set; }
-
-        public Uri AmqpEndpoint { get; private set; }
 
         public string Audience { get; private set; }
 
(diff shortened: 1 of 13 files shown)


common/src/service/IotHubConnectionString.cs
```diff
@@ -11,9 +11,8 @@
 
 namespace Microsoft.Azure.Devices
 {
-    internal sealed class IotHubConnectionString : IAuthorizationHeaderProvider, ICbsTokenProvider
+    internal sealed class IotHubConnectionString : IotHubConnectionBase
```
vinagesh, line 14: The new design is that all the service  clients will only need what is there in the base class. And we will have our different auth types derive from it. The naming here is not great but I don't want to make this a daunting PR so want to keep them for another PR. Ideally, this class (because internal) can be named IotHubConnectionStringCredential". Feel free to leave your inputs.

common/src/service/IotHubConnectionString.cs
```diff
@@ -11,9 +11,8 @@
...
     {
-        private static readonly TimeSpan s_defaultTokenTimeToLive = TimeSpan.FromHours(1);
```
vinagesh, line 16: moved to base class

common/src/service/IotHubConnectionString.cs
```diff
@@ -31,19 +30,8 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
             IotHubName = builder.IotHubName;
             HttpsEndpoint = new UriBuilder("https", HostName).Uri;
             AmqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, builder.HostName, AmqpConstants.DefaultSecurePort).Uri;
-            DeviceId = builder.DeviceId;
```
vinagesh, line 34: This was copied over from DeviceClient and is not needed in ServiceClient. We don't make use of it anyway. I am running tests to ensure nothing breaks.

common/src/service/IotHubConnectionString.cs
```diff
@@ -31,19 +30,8 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
...
-            ModuleId = builder.ModuleId;
-            GatewayHostName = builder.GatewayHostName;
         }
 
-        public string IotHubName { get; private set; }
```
vinagesh, line 39: These are moved to base class

common/src/service/IotHubConnectionString.cs
```diff
@@ -85,12 +67,12 @@ public string GetPassword()
             return password;
         }
 
-        public string GetAuthorizationHeader()
+        public override string GetAuthorizationHeader()
```
vinagesh, line 70: The base class now only has a abstract method and any auth implementation needs to provide the implementation. This is how we will be able to send a base class to all the clients but have specific way to get the tokens based on the auth type.

common/src/service/IotHubConnectionString.cs
```diff
@@ -108,16 +90,6 @@ Task<CbsToken> ICbsTokenProvider.GetTokenAsync(Uri namespaceAddress, string appl
             return Task.FromResult(token);
         }
 
-        public Uri BuildLinkAddress(string path)
```
vinagesh, line 111: moved to base class

common/src/service/IotHubConnectionString.cs
```diff
@@ -130,17 +102,10 @@ private string BuildToken(out TimeSpan ttl)
             {
                 KeyName = SharedAccessKeyName,
                 Key = SharedAccessKey,
-                TimeToLive = s_defaultTokenTimeToLive,
+                TimeToLive = _defaultTokenTimeToLive,
                 Target = Audience
             };
 
-            if (DeviceId != null)
```
vinagesh, line 137: Not needed in ServiceClient

iothub/service/src/DigitalTwin/Authentication/SharedAccessKeyCredentials.cs
```diff
@@ -7,20 +7,18 @@
...
     /// <summary>
-    /// Allows authentication to the API using a Shared Access Key
+    /// Allows authentication to the API using a Shared Access Key.
     /// </summary>
     internal class SharedAccessKeyCredentials : IotServiceClientCredentials
     {
-        // Time buffer before expiry when the token should be renewed, expressed as a percentage of the time to live.
-        // The token will be renewed when it has 15% or less of the sas token's lifespan left.
+        // The time buffer before expiry when the token should be renewed, expressed as a percentage of the time to live.
+        // The token will be renewed when it has 15% or less of the SAS token's lifespan left.
```
vinagesh, line 15: What is the convention here? SAS or sas in comments?

common/src/service/IotHubConnectionString.cs
```diff
@@ -52,24 +38,6 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
 
         public string SharedAccessSignature { get; private set; }
 
-        public string DeviceId { get; private set; }
```
vinagesh, line 55: These are removed

PR #1781, opened 2021-02-10
(service-client: Refactor and add implementation for token credential…

vinagesh:
… input)

## Checklist

### DIFF
--- a/common/src/service/IotHubConnectionString.cs
+++ b/common/src/service/IotHubConnectionString.cs
@@ -11,12 +11,14 @@ using Microsoft.Azure.Devices.Common.Security;
 
 namespace Microsoft.Azure.Devices
 {
-    internal sealed class IotHubConnectionString : IAuthorizationHeaderProvider, ICbsTokenProvider
+    /// <summary>
+    /// The properties required for authentication to IoT hub using a connection string.
+    /// </summary>
+    internal sealed class IotHubConnectionString : IotHubCredential
     {
-        private static readonly TimeSpan s_defaultTokenTimeToLive = TimeSpan.FromHours(1);
-        private const char UserSeparator = '@';
+        private static readonly TimeSpan _tokenTimeToLive = TimeSpan.FromHours(1);
 
-        public IotHubConnectionString(IotHubConnectionStringBuilder builder)
+        public IotHubConnectionString(IotHubConnectionStringBuilder builder) : base(builder.HostName)
         {
             if (builder == null)
             {
@@ -24,25 +26,13 @@ namespace Microsoft.Azure.Devices
             }
 
             Audience = builder.HostName;
-            HostName = string.IsNullOrEmpty(builder.GatewayHostName) ? builder.HostName : builder.GatewayHostName;
             SharedAccessKeyName = builder.SharedAccessKeyName;
             SharedAccessKey = builder.SharedAccessKey;
             SharedAccessSignature = builder.SharedAccessSignature;
-            IotHubName = builder.IotHubName;
-            HttpsEndpoint = new UriBuilder("https", HostName).Uri;
-            AmqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, builder.HostName, AmqpConstants.DefaultSecurePort).Uri;
             DeviceId = builder.DeviceId;
             ModuleId = builder.ModuleId;
             GatewayHostName = builder.GatewayHostName;
         }
-
-        public string IotHubName { get; private set; }
-
-        public string HostName { get; private set; }
-
-        public Uri HttpsEndpoint { get; private set; }
-
-        public Uri AmqpEndpoint { get; private set; }
 
         public string Audience { get; private set; }
 
@@ -58,39 +48,17 @@ namespace Microsoft.Azure.Devices
 
         public string GatewayHostName { get; private set; }
 
-        public string GetUser()
-        {
-            var stringBuilder = new StringBuilder();
-            stringBuilder.Append(SharedAccessKeyName);
-            stringBuilder.Append(UserSeparator);
-            stringBuilder.Append("sas.");
-            stringBuilder.Append("root.");
-            stringBuilder.Append(IotHubName);
-
-            return stringBuilder.ToString();
-        }
-
         public string GetPassword()
         {
-            string password;
-            if (string.IsNullOrWhiteSpace(SharedAccessSignature))
-            {
-                password = BuildToken(out _);
-            }
-            else
-            {
-                password = SharedAccessSignature;
-            }
-
-            return password;
+            return string.IsNullOrWhiteSpace(SharedAccessSignature) ? BuildToken(out _) : SharedAccessSignature;
         }
 
-        public string GetAuthorizationHeader()
+        public override string GetAuthorizationHeader()
         {
             return GetPassword();
         }
 
-        Task<CbsToken> ICbsTokenProvider.GetTokenAsync(Uri namespaceAddress, string appliesTo, string[] requiredClaims)
+        public override Task<CbsToken> GetTokenAsync(Uri namespaceAddress, string appliesTo, string[] requiredClaims)
         {
             string tokenValue;
             CbsToken token;
@@ -108,16 +76,6 @@ namespace Microsoft.Azure.Devices
             return Task.FromResult(token);
         }
 
-        public Uri BuildLinkAddress(string path)
-        {
-            var builder = new UriBuilder(AmqpEndpoint)
-            {
-                Path = path,
-            };
-
-            return builder.Uri;
-        }
-
         public static IotHubConnectionString Parse(string connectionString)
         {
             var builder = IotHubConnectionStringBuilder.Create(connectionString);
@@ -130,7 +88,7 @@ namespace Microsoft.Azure.Devices
             {
                 KeyName = SharedAccessKeyName,
                 Key = SharedAccessKey,
-                TimeToLive = s_defaultTokenTimeToLive,
+                TimeToLive = _tokenTimeToLive,
                 Target = Audience
             };
 
--- a/iothub/service/src/IotHubConnection.cs
+++ b/iothub/service/src/IotHubConnection.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Azure.Devices
         private IOThreadTimer _refreshTokenTimer;
 #endif
 
-        public IotHubConnection(IotHubConnectionString connectionString, AccessRights accessRights, bool useWebSocketOnly, ServiceClientTransportSettings transportSettings)
+        public IotHubConnection(IotHubCredential credential, AccessRights accessRights, bool useWebSocketOnly, ServiceClientTransportSettings transportSettings)
         {
 #if !NET451
             _refreshTokenTimer = new IOThreadTimerSlim(s => ((IotHubConnection)s).OnRefreshTokenAsync(), this);
@@ -59,7 +59,7 @@ namespace Microsoft.Azure.Devices
             _refreshTokenTimer = new IOThreadTimer(s => ((IotHubConnection)s).OnRefreshTokenAsync(), this, false);
 #endif
 
-            ConnectionString = connectionString;
+            Credential = credential;
             _accessRights = accessRights;
             _faultTolerantSession = new FaultTolerantAmqpObject<AmqpSession>(CreateSessionAsync, CloseConnection);
             _useWebSocketOnly = useWebSocketOnly;
@@ -71,7 +71,7 @@ namespace Microsoft.Azure.Devices
             _faultTolerantSession = new FaultTolerantAmqpObject<AmqpSession>(onCreate, onClose);
         }
 
-        internal IotHubConnectionString ConnectionString { get; private set; }
+        internal IotHubCredential Credential { get; private set; }
 
         public Task OpenAsync(TimeSpan timeout)
         {
@@ -114,7 +114,7 @@ namespace Microsoft.Azure.Devices
                     session = await _faultTolerantSession.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
                 }
 
-                Uri linkAddress = ConnectionString.BuildLinkAddress(path);
+                Uri linkAddress = Credential.BuildLinkAddress(path);
 
                 var linkSettings = new AmqpLinkSettings
                 {
@@ -156,7 +156,7 @@ namespace Microsoft.Azure.Devices
                     session = await _faultTolerantSession.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
                 }
 
-                Uri linkAddress = ConnectionString.BuildLinkAddress(path);
+                Uri linkAddress = Credential.BuildLinkAddress(path);
 
                 var linkSettings = new AmqpLinkSettings
                 {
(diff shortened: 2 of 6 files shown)


common/src/service/IotHubConnectionString.cs
```diff
@@ -11,9 +11,11 @@
 
 namespace Microsoft.Azure.Devices
 {
-    internal sealed class IotHubConnectionString : IAuthorizationHeaderProvider, ICbsTokenProvider
+    /// <summary>
+    /// The properties required for authentication to IoT hub using a connection string.
+    /// </summary>
+    internal sealed class IotHubConnectionString : IotHubCredential
```
vinagesh, line 17: This is not the best name for this class but I do not want to make this a daunting PR so I will make the naming changes in a different PR. Feel free to leave suggestions though.
This PR moves a part of the properties in this class that are common to auth to a base class called - IotHubCredential. Each of the auth types will derive from it and add implementation to get tokens.

common/src/service/IotHubConnectionString.cs
```diff
@@ -24,26 +26,14 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
             }
 
             Audience = builder.HostName;
-            HostName = string.IsNullOrEmpty(builder.GatewayHostName) ? builder.HostName : builder.GatewayHostName;
```
vinagesh, line 27: All of this has been moved to a base class.

iothub/service/src/IotHubCredential.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Amqp;
+using Microsoft.Azure.Devices.Common;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// The properties required for authentication to IoT hub that are independent of the authentication type.
+    /// </summary>
+    internal abstract class IotHubCredential
```
vinagesh, line 15: The abstract base class that all specific auth types will derive from and implement.

iothub/service/src/IotHubTokenCredential.cs
```diff
@@ -0,0 +1,60 @@
...
+using Azure.Core;
+
+#endif
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// The properties required for authentication to IoT hub using a token credential.
+    /// </summary>
+    internal class IotHubTokenCredential : IotHubCredential
```
vinagesh, line 21: The new TokenCredential class. Will be adding AzureSasCredential in a subsequent PR.

common/src/service/IotHubConnectionString.cs
```diff
@@ -58,18 +48,6 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
 
         public string GatewayHostName { get; private set; }
 
-        public string GetUser()
```
vinagesh, line 61: Unused in internal class so removed it

iothub/service/src/IotHubConnection.cs
```diff
@@ -114,7 +114,7 @@ public async Task<SendingAmqpLink> CreateSendingLinkAsync(string path, TimeSpan
                     session = await _faultTolerantSession.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
                 }
 
-                Uri linkAddress = ConnectionString.BuildLinkAddress(path);
+                Uri linkAddress = Credential.BuildLinkAddress(path);
```
timtay-microsoft, line 117: I'm going to guess that this would be difficult, and probably not be worth the effort, but is it possible for building the link address to be done somewhere else? This line reads a bit funnily to me. Credentials shouldn't know or care about CBS specific link addresses

iothub/service/src/IotHubCredential.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public Uri HttpsEndpoint { get; protected set; }
+
+        public Uri AmqpEndpoint { get; protected set; }
+
+        public abstract string GetAuthorizationHeader();
+
+        public abstract Task<CbsToken> GetTokenAsync(Uri namespaceAddress, string appliesTo, string[] requiredClaims);
+
+        // Azure.Core (used in IotHubTokenCredential) is not available in NET451.
```
timtay-microsoft, line 32: So does that mean NET451 users won't be able to do RBAC auth? We may want to pass this along to our PMs in case our customers would like NET451 support

iothub/service/src/IotHubCredential.cs
```diff
@@ -0,0 +1,68 @@
...
+        protected IotHubCredential()
+        {
+        }
+
+        protected IotHubCredential(string hostName)
+        {
+            HostName = hostName;
+            IotHubName = GetIotHubName(hostName);
+            AmqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, HostName, AmqpConstants.DefaultSecurePort).Uri;
+            HttpsEndpoint = new UriBuilder("https", HostName).Uri;
```
timtay-microsoft, line 43: "https" could be a const

iothub/service/src/IotHubConnection.cs
timtay-microsoft: This may be a moot point if you rename the Credential type though. The type seems to include connection details (hostname) on top of credentials.

common/src/service/IotHubConnectionString.cs
```diff
@@ -108,16 +86,6 @@ Task<CbsToken> ICbsTokenProvider.GetTokenAsync(Uri namespaceAddress, string appl
             return Task.FromResult(token);
         }
 
-        public Uri BuildLinkAddress(string path)
```
abhipsaMisra, line 111: public methods in internal classes always gets me confused! :D

common/src/service/IotHubConnectionString.cs
```diff
@@ -11,39 +11,29 @@
...
 
             Audience = builder.HostName;
-            HostName = string.IsNullOrEmpty(builder.GatewayHostName) ? builder.HostName : builder.GatewayHostName;
             SharedAccessKeyName = builder.SharedAccessKeyName;
             SharedAccessKey = builder.SharedAccessKey;
             SharedAccessSignature = builder.SharedAccessSignature;
-            IotHubName = builder.IotHubName;
-            HttpsEndpoint = new UriBuilder("https", HostName).Uri;
-            AmqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, builder.HostName, AmqpConstants.DefaultSecurePort).Uri;
             DeviceId = builder.DeviceId;
```
abhipsaMisra, line 32: nit: you wanted to remove these device specific properties, right?

iothub/service/src/IotHubCredential.cs
```diff
@@ -0,0 +1,68 @@
...
+        public Uri HttpsEndpoint { get; protected set; }
+
+        public Uri AmqpEndpoint { get; protected set; }
+
+        public abstract string GetAuthorizationHeader();
+
+        public abstract Task<CbsToken> GetTokenAsync(Uri namespaceAddress, string appliesTo, string[] requiredClaims);
+
+        // Azure.Core (used in IotHubTokenCredential) is not available in NET451.
+        // So we need this constructor for the build to pass.
```
abhipsaMisra, line 33: Interesting! 
Q - Is there an advantage to the adding this to the base class instead of adding it to IotHubTokenCredential directly?

common/src/service/IotHubConnectionString.cs
vinagesh: yes that is in progress. The PR tests failed for some reason so trying to sort that out before I remove something.

iothub/service/src/IotHubTokenCredential.cs
```diff
@@ -0,0 +1,62 @@
...
+
+#endif
+
+        public override string GetAuthorizationHeader()
+        {
+#if !NET451
+            AccessToken token = _credential.GetToken(new TokenRequestContext(), new CancellationToken());
+            return $"Bearer {token.Token}";
+#else
+            throw new InvalidOperationException($"{nameof(GetAuthorizationHeader)} is not supported on NET451");
```
abhipsaMisra, line 41: Q - is it worth adding the class name into the exception message => signifying that TokenCredential is the type that isn't supported in net451?

common/src/service/IotHubConnectionString.cs
vinagesh: The point is that you can always just change the access modifier of the class to easily make things public and private and internal. But this specific method can be private in the base class.

iothub/service/src/IotHubCredential.cs
abhipsaMisra: this = the no-args protected constructor

iothub/service/src/IotHubCredential.cs
vinagesh: Nope they won't be. Good call. I'll bring it up with them tomorrow.

iothub/service/src/IotHubCredential.cs
vinagesh: Sort of. The problem is if I add it to the IotHubTokenCredential class, I need to pass something to the base constructor for 'connectionString'. This will lead to a nullref exception. Infact it reminds me I need to add a constructor that throws an exception in the IotHubTokenCredential class (which I missed somehow). The base class constructor always gets executed before derived so I cannot avoid the nullref exception.

iothub/service/src/IotHubCredential.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    /// <summary>
+    /// The properties required for authentication to IoT hub that are independent of the authentication type.
+    /// </summary>
+    internal abstract class IotHubCredential
+        : IAuthorizationHeaderProvider, ICbsTokenProvider
+    {
+        private const string HostNameSeparator = ".";
+
+        public string IotHubName { get; protected set; }
```
drwill-ms, line 20: properties after the ctor

drwill-ms: Even if the ctor is not public

common/src/service/IotHubConnectionString.cs
vinagesh: Actually I take it back. I tried to do that and remembered why it did not work. It is used in a public method and there is a chain so it does not work.

iothub/service/src/IotHubConnection.cs
vinagesh: Maybe IothubConnectionProperties?

iothub/service/src/IotHubTokenCredential.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    /// <summary>
+    /// The properties required for authentication to IoT hub using a token credential.
+    /// </summary>
+    internal class IotHubTokenCredential : IotHubCredential
+    {
+#if !NET451
+
+        private const string _tokenType = "jwt";
+        private TokenCredential _credential;
```
drwill-ms, line 26: readonly?

iothub/service/src/IotHubTokenCredential.cs
```diff
@@ -0,0 +1,62 @@
...
+#endif
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// The properties required for authentication to IoT hub using a token credential.
+    /// </summary>
+    internal class IotHubTokenCredential : IotHubCredential
+    {
+#if !NET451
```
drwill-ms, line 23: Would you mind inverting the if? I find it easier to read the positive case.

common/src/service/IotHubConnectionString.cs
drwill-ms: `IotHubConnectionStringCredential`?

common/src/service/IotHubConnectionString.cs
```diff
@@ -58,18 +48,6 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
...
-            stringBuilder.Append("sas.");
-            stringBuilder.Append("root.");
-            stringBuilder.Append(IotHubName);
-
-            return stringBuilder.ToString();
-        }
-
         public string GetPassword()
         {
             string password;
```
drwill-ms, line 53: This would work well with a ternary

iothub/service/src/IotHubTokenCredential.cs
vinagesh: I believe it should not even get here. I am now throwing an exception in the constructor itself that TokenCredential is not supported on NET451.

drwill-ms (approved):

iothub/service/src/IotHubTokenCredential.cs
vinagesh: It can be for now. But we might need to add caching eventually and update it whenever it is about to expire.

common/src/service/IotHubConnectionString.cs
vinagesh: Sounds reasonable. Have conversation below with Tim on another class. We should look at it together to find good names. I will do one PR just to get the names right so that noone has to worry about logic. Just that so many names are already taken, I'm finding it hard to name these well :(

timtay-microsoft (approved):

iothub/service/src/IotHubCredential.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            var builder = new UriBuilder(AmqpEndpoint)
+            {
+                Path = path,
+            };
+
+            return builder.Uri;
+        }
+
+        private static string GetIotHubName(string hostName)
```
barustum, line 57: nit: Can we think of a way to unit test this method? Because it's private, we might have to change the access modifier. 

String parsing logic is something that I try to add coverage for because input can vary greatly, especially if 'hostName' is something passed from the user.

common/src/service/IotHubConnectionString.cs
```diff
@@ -11,39 +11,29 @@
...
+    internal sealed class IotHubConnectionString : IotHubCredential
     {
-        private static readonly TimeSpan s_defaultTokenTimeToLive = TimeSpan.FromHours(1);
-        private const char UserSeparator = '@';
+        private static readonly TimeSpan _tokenTimeToLive = TimeSpan.FromHours(1);
 
-        public IotHubConnectionString(IotHubConnectionStringBuilder builder)
+        public IotHubConnectionString(IotHubConnectionStringBuilder builder) : base(builder.HostName)
         {
             if (builder == null)
```
barustum, line 23: we're checking for builder null after the 'base(builder.HostName)' call. Maybe move the base(builder.HostName) call after the null check.

common/src/service/IotHubConnectionString.cs
```diff
@@ -58,39 +48,17 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
...
-            {
-                password = SharedAccessSignature;
-            }
-
-            return password;
+            return string.IsNullOrWhiteSpace(SharedAccessSignature) ? BuildToken(out _) : SharedAccessSignature;
         }
 
-        public string GetAuthorizationHeader()
+        public override string GetAuthorizationHeader()
```
barustum, line 56: Do these need to be public? since the class is internal?

common/src/service/IotHubConnectionString.cs
vinagesh: I will send out another PR with these changes and see what everyone thinks.

iothub/service/src/IotHubConnection.cs
vinagesh: Will send out these changes in the next PR.

iothub/service/src/IotHubCredential.cs
vinagesh: Sure let me take a look. I'll add it as a task and address it in my next PR. The class is internal so I think it should be doable. But the test won't really test based on what user passes. The test can only test if the user sends in the correct hostname, we parse it correctly. There is only one format for the hostname.

common/src/service/IotHubConnectionString.cs
vinagesh: Makes sense. Fixing it in next PR itself as it almost ready to go. I don't want to run the gates again :D

PR #1786, opened 2021-02-10
(service-client): Refactor and add sas credential

vinagesh:
## Checklist

### DIFF
--- a/common/src/service/IotHubConnectionString.cs
+++ b/common/src/service/IotHubConnectionString.cs
@@ -14,11 +14,12 @@ namespace Microsoft.Azure.Devices
     /// <summary>
     /// The properties required for authentication to IoT hub using a connection string.
     /// </summary>
-    internal sealed class IotHubConnectionString : IotHubCredential
+    internal sealed class IotHubConnectionString
+        : IotHubConnectionProperties
     {
         private static readonly TimeSpan _tokenTimeToLive = TimeSpan.FromHours(1);
 
-        public IotHubConnectionString(IotHubConnectionStringBuilder builder) : base(builder.HostName)
+        public IotHubConnectionString(IotHubConnectionStringBuilder builder) : base(builder?.HostName)
         {
             if (builder == null)
             {
@@ -29,9 +30,6 @@ namespace Microsoft.Azure.Devices
             SharedAccessKeyName = builder.SharedAccessKeyName;
             SharedAccessKey = builder.SharedAccessKey;
             SharedAccessSignature = builder.SharedAccessSignature;
-            DeviceId = builder.DeviceId;
-            ModuleId = builder.ModuleId;
-            GatewayHostName = builder.GatewayHostName;
         }
 
         public string Audience { get; private set; }
@@ -41,12 +39,6 @@ namespace Microsoft.Azure.Devices
         public string SharedAccessKey { get; private set; }
 
         public string SharedAccessSignature { get; private set; }
-
-        public string DeviceId { get; private set; }
-
-        public string ModuleId { get; private set; }
-
-        public string GatewayHostName { get; private set; }
 
         public string GetPassword()
         {
@@ -91,13 +83,6 @@ namespace Microsoft.Azure.Devices
                 TimeToLive = _tokenTimeToLive,
                 Target = Audience
             };
-
-            if (DeviceId != null)
-            {
-                builder.Target = string.IsNullOrEmpty(ModuleId)
-                    ? "{0}/devices/{1}".FormatInvariant(Audience, WebUtility.UrlEncode(DeviceId))
-                    : "{0}/devices/{1}/modules/{2}".FormatInvariant(Audience, WebUtility.UrlEncode(DeviceId), WebUtility.UrlEncode(ModuleId));
-            }
 
             ttl = builder.TimeToLive;
 
--- a/iothub/service/src/IoTHubSasCredential.cs
+++ b/iothub/service/src/IoTHubSasCredential.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+#if !NET451
+
+using Azure;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Shared access signature credential used to authenticate with IoT hub.
+    /// </summary>
+    public class IotHubSasCredential
+        : AzureSasCredential
+    {
+        /// <summary>
+        /// Creates an instance of <see cref="IotHubSasCredential"/>.
+        /// </summary>
+        /// <param name="signature">Shared access signature used to authenticate with IoT hub.</param>
+        /// <param name="expiresOnUtc">The shared access signature expiry in UTC.</param>
+        public IotHubSasCredential(string signature, DateTime expiresOnUtc) : base(signature)
+        {
+            ExpiresOnUtc = expiresOnUtc;
+        }
+
+        /// <summary>
+        /// The shared access signature expiry in UTC.
+        /// </summary>
+        public DateTime ExpiresOnUtc { get; private set; }
+
+        /// <summary>
+        /// Updates the shared access signature. This is intended to be used when you've
+        /// regenerated your shared access signature and want to update clients.
+        /// </summary>
+        /// <param name="signature">Shared access signature used to authenticate with IoT hub.</param>
+        /// <param name="expiresOnUtc">The shared access signature expiry in UTC.</param>
+        public void Update(string signature, DateTime expiresOnUtc)
+        {
+            Update(signature);
+            ExpiresOnUtc = expiresOnUtc;
+        }
+    }
+}
+
+#endif
--- a/iothub/service/src/IotHubConnectionProperties.cs
+++ b/iothub/service/src/IotHubConnectionProperties.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Azure.Devices
     /// <summary>
     /// The properties required for authentication to IoT hub that are independent of the authentication type.
     /// </summary>
-    internal abstract class IotHubCredential
+    internal abstract class IotHubConnectionProperties
         : IAuthorizationHeaderProvider, ICbsTokenProvider
     {
         private const string HostNameSeparator = ".";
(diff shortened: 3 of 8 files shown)


common/src/service/IotHubConnectionString.cs
```diff
@@ -14,11 +14,12 @@ namespace Microsoft.Azure.Devices
     /// <summary>
     /// The properties required for authentication to IoT hub using a connection string.
     /// </summary>
-    internal sealed class IotHubConnectionString : IotHubCredential
+    internal sealed class IotHubConnectionString
+        : IotHubConnectionProperties
```
vinagesh, line 18: Renamed the base class to indicate they are connection properties and not credential properties.

common/src/service/IotHubConnectionString.cs
```diff
@@ -29,9 +30,6 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder) : base(buil
             SharedAccessKeyName = builder.SharedAccessKeyName;
             SharedAccessKey = builder.SharedAccessKey;
             SharedAccessSignature = builder.SharedAccessSignature;
-            DeviceId = builder.DeviceId;
```
vinagesh, line 32: These are not needed in the Service Client. It looks like copied over code from Device Client that is unused.

iothub/service/src/IoTHubSasCredential.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
vinagesh, line 1: I had a chat with Scott on this and we landed on adding a derived class from AzureSasCredential. The reason we need this is that the AMQP library we use, takes the expiry date as input. Parsing the signature to find the expiry is not ideal and can cause bugs. So the best way forward for us is to create this derived class and take this as input in our constructors (in the upcoming PR).

iothub/service/src/IotHubConnectionProperties.cs
```diff
@@ -54,7 +59,7 @@ public Uri BuildLinkAddress(string path)
             return builder.Uri;
         }
 
-        private static string GetIotHubName(string hostName)
+        public static string GetIotHubName(string hostName)
```
vinagesh, line 62: Class is internal but this public so that unit tests can be added. This should have no impact.

iothub/service/src/IotHubSasCredentialProperties.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Microsoft.Azure.Devices
+{
+    internal class IotHubSasCredentialProperties : IotHubConnectionProperties
+    {
+#if NET451
+
+        public IotHubSasCredentialProperties()
+        {
+            throw new InvalidOperationException("nameof(AzureSasCredential) is not supported in NET451");
```
timtay-microsoft, line 23: Should the error be "IotHubSasCredentialProperties is not supported in NET451" instead?

vinagesh: I think it should be IotHubSasCredential is not supported in NET451. The IotHubSasCredentialProperties class is internal but the constructors will now expose IotHubSasCredential  instead of AzureSasCredential.

iothub/service/src/IotHubConnectionProperties.cs
drwill-ms: It could be internal then.

iothub/service/src/IoTHubSasCredential.cs
```diff
@@ -0,0 +1,47 @@
...
+#if !NET451
+
+using Azure;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Shared access signature credential used to authenticate with IoT hub.
+    /// </summary>
+    public class IoTHubSasCredential : AzureSasCredential
```
timtay-microsoft, line 16: We have a mix of "IoTHub" vs "IotHub" casing in this project. From what I can tell, "IotHub" is a bit more common. Maybe match that casing here?

iothub/service/src/IotHubTokenCredentialProperties.cs
```diff
@@ -18,19 +18,20 @@ namespace Microsoft.Azure.Devices
...
     /// The properties required for authentication to IoT hub using a token credential.
     /// </summary>
-    internal class IotHubTokenCredential : IotHubCredential
+    internal class IotHubTokenCrendentialProperties
+        : IotHubConnectionProperties
     {
 #if NET451
 
-        public IotHubTokenCredential()
+        public IotHubTokenCrendentialProperties()
```
drwill-ms, line 26: ctor below fields

iothub/service/src/IoTHubSasCredential.cs
vinagesh: It should be IotHub for classes. This one should be changed.

iothub/service/src/IoTHubSasCredential.cs
```diff
@@ -0,0 +1,47 @@
...
+        public void Update(string signature, DateTime expiresOnUtc)
+        {
+            Update(signature);
+            ExpiresOnUtc = expiresOnUtc;
+        }
+
+        /// <summary>
+        /// The shared access signature expiry in UTC.
+        /// </summary>
+        public DateTime ExpiresOnUtc { get; private set; }
```
drwill-ms, line 43: properties above methods

iothub/service/src/IotHubTokenCredentialProperties.cs
vinagesh: ctor below both fields and properties?

iothub/service/src/IoTHubSasCredential.cs
timtay-microsoft, line 43: The one qualm I have with this is that the signature already has this this information baked in as one of the key value pairs, so it feels a bit redundant. I get that parsing it out and converting seconds to a DateTime field leave us open to bugs, but I'm curious to hear everyone's opinion on this.

iothub/service/src/IotHubTokenCredentialProperties.cs
vinagesh: When I looked at the code, it is below the fields.

iothub/service/src/IoTHubSasCredential.cs
timtay-microsoft: Also, should this field be a DateTimeOffset, since it is supposed to be UTC?

vinagesh: Can you send me a link to where fields, methods, ctors should be? I can follow it when I write code.

vinagesh: Is It as follows? And how do we order them based on public/private/protected?
Fields
Ctor
Properties
Methods

vinagesh: The AMQP library expects a DateTime in UTC so I kept it as DateTime. We don't need to do unnecessary conversions.

drwill-ms: Standards are here: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/doc/coding-style.md

Typical order is:
- fields (constants, static/readonly, static, other instance fields)
- ctor (public, internal protected, private)
- properties
- methods (public, internal, protected, private)

iothub/service/src/IotHubTokenCredentialProperties.cs
drwill-ms: I see 2 ctors, one for net451 and one for the others. I see fields between them.

vinagesh: The fields are not for NET 451. I can add a separate #if NET451 block.

vinagesh: I can make the update and you can see how it looks. The #if will be inverted though as there is no fields for NET451

iothub/service/src/IoTHubSasCredential.cs
timtay-microsoft: Does this just mean we are leaving the parsing logic to the user? Or do they have this date time field prior to creating the SAS token?

vinagesh: From what I can tell, they should be constructing the SAS using a DateTime. For example: We do that when we get the connection string as input. Atleast for IotHub the portal only gives a connectionstring so I don't see any other way for them to have a connection string and parse it.

timtay-microsoft: The doc that I found (which I suspect customers will also reference) doesn't seem to give the users the expiration time for free:

```csharp
public static string generateSasToken(string resourceUri, string key, string policyName, int expiryInSeconds = 3600)
{
    TimeSpan fromEpochStart = DateTime.UtcNow - new DateTime(1970, 1, 1);
    string expiry = Convert.ToString((int)fromEpochStart.TotalSeconds + expiryInSeconds);

    string stringToSign = WebUtility.UrlEncode(resourceUri) + "\n" + expiry;

    HMACSHA256 hmac = new HMACSHA256(Convert.FromBase64String(key));
    string signature = Convert.ToBase64String(hmac.ComputeHash(Encoding.UTF8.GetBytes(stringToSign)));

    string token = String.Format(CultureInfo.InvariantCulture, "SharedAccessSignature sr={0}&sig={1}&se={2}", WebUtility.UrlEncode(resourceUri), WebUtility.UrlEncode(signature), expiry);

    if (!String.IsNullOrEmpty(policyName))
    {
        token += "&skn=" + policyName;
    }

    return token;
}
```

From [here](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#security-token-structure)

vinagesh: Good find! However I did not understand which way you were inclined for the design. The code snippet above shows that they have constructed a DateTime and appended to the SAS, which means they will have it handy to pass it on to the SDK. I feel parsing the SAS token again in the SDK might be redundant and could lead to bugs.

drwill-ms: I agree - avoid parsing

drwill-ms (approved):

iothub/service/src/IoTHubSasCredential.cs
vinagesh: @timtay-microsoft - Ping me if you want to discuss further. I will resolve this for now.

PR #1788, opened 2021-02-11
(service-client): Add constructors in registry manager to accept aad …

vinagesh:
…and sas tokens.

## Checklist

### DIFF
--- a/iothub/service/src/HttpRegistryManager.cs
+++ b/iothub/service/src/HttpRegistryManager.cs
@@ -54,12 +54,12 @@ namespace Microsoft.Azure.Devices
         private IHttpClientHelper _httpClientHelper;
         private readonly string _iotHubName;
 
-        internal HttpRegistryManager(IotHubConnectionString connectionString, HttpTransportSettings transportSettings)
+        internal HttpRegistryManager(IotHubConnectionProperties connectionProperties, HttpTransportSettings transportSettings)
         {
-            _iotHubName = connectionString.IotHubName;
+            _iotHubName = connectionProperties.IotHubName;
             _httpClientHelper = new HttpClientHelper(
-                connectionString.HttpsEndpoint,
-                connectionString,
+                connectionProperties.HttpsEndpoint,
+                connectionProperties,
                 ExceptionHandlingHelper.GetDefaultErrorMapping(),
                 s_defaultOperationTimeout,
                 transportSettings.Proxy);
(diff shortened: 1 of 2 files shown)


iothub/service/src/HttpRegistryManager.cs
```diff
@@ -54,12 +54,12 @@ internal class HttpRegistryManager : RegistryManager
         private IHttpClientHelper _httpClientHelper;
         private readonly string _iotHubName;
 
-        internal HttpRegistryManager(IotHubConnectionString connectionString, HttpTransportSettings transportSettings)
+        internal HttpRegistryManager(IotHubConnectionProperties connectionProperties, HttpTransportSettings transportSettings)
```
vinagesh, line 57: All these properties are now in a base class, making the HttpRegistryManager independent of the auth type.

timtay-microsoft (approved):

abhipsaMisra (approved):

iothub/service/src/RegistryManager.cs
```diff
@@ -46,6 +53,66 @@ public static RegistryManager CreateFromConnectionString(string connectionString
...
+            }
+
+            if (credential == null)
+            {
+                throw new ArgumentNullException($"{nameof(credential)} is null");
+            }
+
+            TlsVersions.Instance.SetLegacyAcceptableVersions();
+
+            IotHubTokenCrendentialProperties tokenCredentialProperties = new IotHubTokenCrendentialProperties(hostName, credential);
```
drwill-ms, line 82: var

iothub/service/src/RegistryManager.cs
```diff
@@ -46,6 +53,66 @@ public static RegistryManager CreateFromConnectionString(string connectionString
...
+            }
+
+            if (credential == null)
+            {
+                throw new ArgumentNullException($"{nameof(credential)} is null");
+            }
+
+            TlsVersions.Instance.SetLegacyAcceptableVersions();
+
+            IotHubSasCredentialProperties sasCredentialProperties = new IotHubSasCredentialProperties(hostName, credential);
```
drwill-ms, line 110: var

iothub/service/src/RegistryManager.cs
```diff
@@ -46,6 +53,66 @@ public static RegistryManager CreateFromConnectionString(string connectionString
...
+            {
+                throw new ArgumentNullException($"{nameof(hostName)} is null or empty");
+            }
+
+            if (credential == null)
+            {
+                throw new ArgumentNullException($"{nameof(credential)} is null");
+            }
+
+            TlsVersions.Instance.SetLegacyAcceptableVersions();
```
drwill-ms, line 80: Why call this? It does nothing when compiled with .NET 4.5.1, and this code block is in a !NET451.

drwill-ms (approved):

iothub/service/src/RegistryManager.cs
vinagesh: I had not seen the implementation so did not realize that. Will remove it.

PR #1790, opened 2021-02-11
(service-client): Add constructors to accept aad and sas tokens for d…

vinagesh:
…igital twins client.

## Checklist

### DIFF
--- a/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
+++ b/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
@@ -10,6 +10,10 @@ using Microsoft.Azure.Devices.Extensions;
 using Microsoft.Azure.Devices.Generated;
 using Microsoft.Rest;
 using Newtonsoft.Json;
+using Microsoft.Azure.Devices.DigitalTwin.Authentication;
+using Azure;
+using Azure.Core;
+using PnpDigitalTwin = Microsoft.Azure.Devices.Generated.DigitalTwin;
 
 namespace Microsoft.Azure.Devices
 {
@@ -18,8 +22,16 @@ namespace Microsoft.Azure.Devices
     /// </summary>
     public class DigitalTwinClient : IDisposable
     {
+        private const string HttpsEndpointPrefix = "https";
         private readonly IotHubGatewayServiceAPIs _client;
-        private readonly DigitalTwin _protocolLayer;
+        private readonly PnpDigitalTwin _protocolLayer;
+
+        private DigitalTwinClient(string hostName, DigitalTwinServiceClientCredentials credentials, params DelegatingHandler[] handlers)
+        {
+            var httpsEndpoint = new UriBuilder(HttpsEndpointPrefix, hostName).Uri;
+            _client = new IotHubGatewayServiceAPIs(httpsEndpoint, credentials, handlers);
+            _protocolLayer = new PnpDigitalTwin(_client);
+        }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DigitalTwinClient"/> class.</summary>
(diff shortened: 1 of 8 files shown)


iothub/service/src/DigitalTwin/Authentication/IotServiceClientCredentials.cs
```diff
@@ -11,7 +12,7 @@
 
 namespace Microsoft.Azure.Devices.Authentication
 {
-    internal abstract class IotServiceClientCredentials : ServiceClientCredentials
+    internal abstract class IotServiceClientCredentials : ServiceClientCredentials, IAuthorizationHeaderProvider
```
vinagesh, line 15: This is a interface we already have and is generic instead of defining a abstract method called GetSasToken()

iothub/service/src/DigitalTwin/Authentication/PnpConnectionStringCredential.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.Devices.Authentication
     /// <summary>
     /// Allows authentication to the API using a Shared Access Key
     /// </summary>
-    internal class SharedAccessKeyCredentials : IotServiceClientCredentials
+    internal class PnpConnectionStringCredential : IotServiceClientCredentials
```
vinagesh, line 12: Unfortunately all pnp clients have to derive from a specific class that is auto-generated. So instead of using the credential classes we created for IotHub, we have to create new ones. Hence, the prefix 'pnp' to distinguish them.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -30,14 +42,60 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+            if (credential == null)
+            {
+                throw new ArgumentNullException($"{nameof(credential)} is null");
+            }
+
+            var tokenCredential = new PnpTokenCredential(credential);
+            return new DigitalTwinClient(hostName, tokenCredential, handlers);
         }
 
-        private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, params DelegatingHandler[] handlers)
```
vinagesh, line 37: This diff is weird somehow. But this ctor has been moved above the static methods and refactored to work with all credential types..

iothub/service/src/DigitalTwin/Authentication/PnpConnectionStringCredential.cs
drwill-ms: What is special about auth for PnP clients?

vinagesh: Nothing special. We just need to derive from a specific class and override a method to add the token to the header. It is just not the same as the one we pass to other clients. Is is sort of duplicate functionality.

drwill-ms: Oh, duh. Because it is in the DigitalTwinClient, right?

Why can't clients share these types?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -10,6 +10,10 @@
 using Microsoft.Azure.Devices.Generated;
 using Microsoft.Rest;
 using Newtonsoft.Json;
+using Microsoft.Azure.Devices.DigitalTwin.Authentication;
+using Azure;
+using Azure.Core;
+using PnpDigitalTwin = Microsoft.Azure.Devices.Generated.DigitalTwin;
```
vinagesh, line 16: This is because of the namespace conflict with Azure.Core DigitalTwin :D

iothub/service/src/DigitalTwin/Authentication/IotServiceClientCredentials.cs
```diff
@@ -33,6 +33,6 @@ public override Task ProcessHttpRequestAsync(HttpRequestMessage request, Cancell
         /// Return a SAS token
```
timtay-microsoft, line 33: Can you update the header comments to reflect that it isn't just a SAS token anymore?

iothub/service/src/DigitalTwin/Authentication/PnpConnectionStringCredential.cs
vinagesh: Because this is a generated type that is passed to the protocol layer. Our other clients are handwritten. I have do a lot of changes to be able to make the other clients use this type. Plus it will mean DT client will have unnecessary AMQP CBS token implementation. DT is only Http.

timtay-microsoft (approved):

iothub/service/src/DigitalTwin/Authentication/PnpConnectionStringCredential.cs
drwill-ms: All this should be covered in code comments.

drwill-ms, line 12: Given we used DigitalTwin instead of PnP for the client, shouldn't we stick with that prefix? I think it can be confusing for someone else understanding that DT and PnP are the same thing.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
drwill-ms: Why would we have a namespace conflict with our track2 ADT SDK?

iothub/service/src/DigitalTwin/Authentication/PnpConnectionStringCredential.cs
vinagesh: Will add

abhipsaMisra, line 12: Adding the prefix "pnp" to our types could come across as a bit confusing; esp since the namespace and client name both have the term "DigitalTwin". 
I know that calling this digital twin is confusing given that ADT is another service of its own; what do you think about not adding any service-specific prefix to these credential types and instead deriving that info by looking at the namespace under which it lives? Or how about we call it "DigitalTwin"xxx?

iothub/service/src/DigitalTwin/Authentication/PnpSasCredential.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public PnpSasCredential(AzureSasCredential credential)
+        {
+            _credential = credential;
+        }
+
+        public override string GetAuthorizationHeader()
+        {
+            return _credential.Signature;
+        }
```
abhipsaMisra, line 23: correct me if I'm wrong, was there also an "update" functionality with our iot hub sas credential type?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
vinagesh: Because the AzureSasToken is in the "Azure" namespace, which will include everything under that. I don't see why else it complained about a conflict with a namespace.

abhipsaMisra (approved):

iothub/service/src/DigitalTwin/Authentication/PnpConnectionStringCredential.cs
vinagesh: I am always confused about this. DigitalTwin of Pnp. It makes it confusing. But I see your point.

iothub/service/src/DigitalTwin/Authentication/PnpSasCredential.cs
vinagesh: That the user needs to implement to updated the signature. But we always grab the signature.

iothub/service/src/DigitalTwin/Authentication/PnpConnectionStringCredential.cs
vinagesh: Without a prefix, the other credentials will clash with the base types. And IotHub prefix is added to the ones we use in all other clients. The only option is PnP or DigitalTwin. Given David's argument above, we can just add the DigitalTwin prefix.

vinagesh: Given we have gone with DigitalTwin anyway for the client, i'll updated these to DigitalTwin prefix.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
drwill-ms: If we don't have a reference to Azure.DigitalTwins.Core, I don't see how we could have a type conflict.

iothub/service/src/DigitalTwin/Authentication/IotServiceClientCredentials.cs
```diff
@@ -11,10 +11,10 @@
 
 namespace Microsoft.Azure.Devices.Authentication
 {
-    internal abstract class IotServiceClientCredentials : ServiceClientCredentials
+    internal abstract class IotServiceClientCredentials : ServiceClientCredentials, IAuthorizationHeaderProvider
```
drwill-ms, line 14: code class comments, please

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -30,14 +42,60 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+            params DelegatingHandler[] handlers)
+        {
+            if (string.IsNullOrEmpty(hostName))
+            {
+                throw new ArgumentNullException($"{nameof(hostName)},  Parameter cannot be null or empty");
+            }
+
+            if (credential == null)
+            {
+                throw new ArgumentNullException($"{nameof(credential)},  Parameter cannot be null or empty");
```
drwill-ms, line 68: not a string; can't be empty

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -30,14 +42,60 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
-            _client = new IotHubGatewayServiceAPIs(uri, credentials, handlers);
-            _protocolLayer = new DigitalTwin(_client);
+            if (string.IsNullOrEmpty(hostName))
+            {
+                throw new ArgumentNullException($"{nameof(hostName)},  Parameter cannot be null or empty");
+            }
+
+            if (credential == null)
+            {
+                throw new ArgumentNullException($"{nameof(credential)},  Parameter cannot be null or empty");
```
drwill-ms, line 94: same, can't be empty

drwill-ms (approved):

PR #1793, opened 2021-02-12
Update IothubSasCredential to use encapsulation instead of inheritance

vinagesh:
## Checklist


iothub/service/src/IoTHubSasCredential.cs
```diff
@@ -14,33 +14,48 @@ namespace Microsoft.Azure.Devices
     /// Shared access signature credential used to authenticate with IoT hub.
     /// </summary>
     public class IotHubSasCredential
```
vinagesh, line 16: We are moving away from inheritance and using encapsulation here because we don't want to expose the methods in AzureSasCredential.

drwill-ms (approved):

abhipsaMisra (approved):

bikamani (approved):

Issue #1505, opened 2020-08-05
[Bug Report] DeviceClient.UpdateReportedPropertiesAsync does not provide helpful information when property name contains invalid characters


tomkerkhove:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64
- **Application's .NET Target Framework :** netcoreapp3.1
- **Device:** Laptop
- **SDK version used:** 1.27.0

## Description of the issue
`DeviceClient.UpdateReportedPropertiesAsync` does not provide helpful information when the property name contains invalid characters, but instead throws a `InvalidOperationException` with a message along the lines of `Service did not provide a response` and it took a while to notice that the name was not allowed (in my case `Example.Property`).

This is because Azure IoT Hub has some restrictions in the naming:

> **Keys:** All keys in JSON objects are UTF-8 encoded, case-sensitive, and up-to 1 KB in length. Allowed characters exclude UNICODE control characters (segments C0 and C1), and ., $, and SP.

Learn more on https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins.

## More context
Report originated on https://github.com/Azure/azure-iot-sdk-csharp/pull/1503 which has some additional context

timtay-microsoft:
From my initial investigation, this seems to be an issue in our AMQP stack. When I run a basic test for sending reported properties where a property has the name ```Property.Example```, MQTT returns a 400 (bad format) error as expected (albeit without any explanation about why the message was poorly formatted). AMQP does seem to consistently return the exception detailed above.

I'll look into why AMQP is behaving differently here

bikamani:
@tomkerkhove Thanks for filing this, I am investigating this issue further and will soon post update.

PR #1796, opened 2021-02-18
feature,fix (device-client) Handle Twin failures using Amqp

bikamani:
## Checklist

## Description of the changes

- This PR includes following changes
- Add code to handle failures for Twin scenario for Get Twin 
- Fix a bug where SDK was mapping to wrong exception "Service Response is Null" for reported properties ( currently with this code client does not throw any exception due to limitations of update reported property API)
- Improve naming of internal methods

https://github.com/Azure/azure-iot-sdk-csharp/issues/1505

There will be follow up PRs to fix API implementation to handle failure for updating reported properties.


e2e/test/iothub/twin/TwinE2ETests.cs
```diff
@@ -367,6 +366,10 @@ public async Task Twin_ClientHandlesRejectionInvalidPropertyName_MqttWs()
                 .ConfigureAwait(false);
         }
 
+        // These tests worked earlier for Amqp and AmqpWs since it was catching a wrong exception
+        // To Do: Fix Update reported properties method behavior (breaking change) to wait for response
+        // and we should be able to enable these tests then.
+        [Ignore]
```
bikamani, line 372: This tests worked before due to wrong exception and buggy code. workaround can be waiting for the exception but that does not seem a good solution to make the tests work. Open to suggestions.

bikamani: Also later, I do plan to add a changes for modifying updating reported properties behavior to await for service response, that will enable us to leave this tests back in. So ignoring the tests is more of a temporary change.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -324,7 +324,8 @@ public override async Task EnableTwinPatchAsync(CancellationToken cancellationTo
             try
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                await _amqpUnit.SendTwinMessageAsync(AmqpTwinMessageType.Put, Guid.NewGuid().ToString(), null, _operationTimeout).ConfigureAwait(false);
+                string correlationId = AmqpTwinMessageType.Put + Guid.NewGuid().ToString();
```
bikamani, line 327: Adding "Put/Patch/Get" before correlation Id to be able to identify response.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -405,10 +406,6 @@ private async Task<Twin> RoundTripTwinMessage(AmqpTwinMessageType amqpTwinMessag
                     // Consider that the task may have faulted or been canceled.
                     // We re-await the task so that any exceptions/cancellation is rethrown.
                     response = await receivingTask.ConfigureAwait(false);
-                    if (response == null)
```
bikamani, line 408: This is unnecessary since we can get to this point 2 ways
1. Get Twin response
2. Updating reported properties response

The check here was assuming the response coming here can never be null (one more assumption that it is only get) and we were throwing a wrong exception for (2)

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -522,19 +519,23 @@ private async Task DisposeMessageAsync(string lockToken, AmqpIoTDisposeActions o
 
         private void TwinMessageListener(Twin twin, string correlationId, TwinCollection twinCollection)
         {
-            if (correlationId != null)
+            if (correlationId == null)
```
bikamani, line 522: changing logic here to first handle desired property updates and then look at the type of message.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
-                Twin twin = null;
-                TwinCollection twinProperties = null;
+                    string error = null;
+                    using (StreamReader reader = new StreamReader(amqpMessage.BodyStream, System.Text.Encoding.UTF8))
+                    {
+                        error = reader.ReadToEnd();
+                    }
 
-                if (correlationId != null)
+                    if (status >= 500 || status == 429 || status == 408)
```
bikamani, line 287: Any other error codes I may have missed for retry?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
         internal void RegisterTwinListener(Action<Twin, string, TwinCollection> onDesiredPropertyReceived)
         {
             _onDesiredPropertyReceived = onDesiredPropertyReceived;
-            _receivingAmqpLink.RegisterMessageListener(OnDesiredPropertyReceived);
+            _receivingAmqpLink.RegisterMessageListener(OnTwinChangesReceived);
         }
 
-        private void OnDesiredPropertyReceived(AmqpMessage amqpMessage)
+        private void OnTwinChangesReceived(AmqpMessage amqpMessage)
```
bikamani, line 264: Better naming I hope :)

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
         #endregion Twin handling
 
-        internal static bool VerifyResponseMessage(AmqpMessage response)
+        internal static int getStatus(AmqpMessage response)
         {
-            bool retVal = true;
+            int retStatus = -1;
             if (response != null)
             {
                 if (response.MessageAnnotations.Map.TryGetValue(AmqpIoTConstants.ResponseStatusName, out int status))
```
barustum, line 353: is there a chance that MessageAnnotations is null? In other words, do we need to guard against null for MessageAnnotations?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
             }
         }
 
         #endregion Twin handling
 
-        internal static bool VerifyResponseMessage(AmqpMessage response)
+        internal static int getStatus(AmqpMessage response)
         {
-            bool retVal = true;
+            int retStatus = -1;
```
barustum, line 350: retStatus is only being used once to return. Maybe not needed and just do a return -1;

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
+                if (status >= 400)
                 {
+                    // Handle failures
                     _onDesiredPropertyReceived.Invoke(null, correlationId, null);
-                }
-
-                Twin twin = null;
-                TwinCollection twinProperties = null;
+                    string error = null;
+                    using (StreamReader reader = new StreamReader(amqpMessage.BodyStream, System.Text.Encoding.UTF8))
```
barustum, line 282: not sure if required, but do you need to do a:
amqpMessage.BodyStream.Seek(0, SeekOrigin.Begin);

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
barustum, line 287: You can probably replace the if/else block of code with:
throw new IotHubException(error, status >= 500 || status == 429 || status == 408);

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
                         {
                             string body = reader.ReadToEnd();
                             var properties = JsonConvert.DeserializeObject<TwinProperties>(body);
                             twin = new Twin(properties);
                         }
                     }
-                    else
+                    else if (correlationId.StartsWith(AmqpTwinMessageType.Patch.ToString(), StringComparison.OrdinalIgnoreCase))
+                    {
+                        // This can be used to coorelate success response with Updating reported properties
```
barustum, line 322: so no action here? if so, do we need this else block?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
-                else
-                {
-                    // No correlationId, this is a PATCH sent by the sevice so return (set) the TwinCollection
-
-                    using (StreamReader reader = new StreamReader(amqpMessage.BodyStream, System.Text.Encoding.UTF8))
+                    else
                     {
-                        string patch = reader.ReadToEnd();
-                        twinProperties = JsonConvert.DeserializeObject<TwinCollection>(patch);
+                        // This shouldn't happen
```
barustum, line 331: Maybe we need to log a warning here in case it ever happens.

barustum: ... or it should never ever happen? :)

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -375,7 +376,7 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
             try
             {
                 await EnableTwinPatchAsync(cancellationToken).ConfigureAwait(false);
-                Twin twin = await RoundTripTwinMessage(AmqpTwinMessageType.Patch, reportedProperties, cancellationToken).ConfigureAwait(false);
+                await RoundTripTwinMessage(AmqpTwinMessageType.Patch, reportedProperties, cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 379: are we intentionally ignoring the returned twin?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -405,10 +406,6 @@ private async Task<Twin> RoundTripTwinMessage(AmqpTwinMessageType amqpTwinMessag
                     // Consider that the task may have faulted or been canceled.
                     // We re-await the task so that any exceptions/cancellation is rethrown.
                     response = await receivingTask.ConfigureAwait(false);
-                    if (response == null)
-                    {
-                        throw new InvalidOperationException("Service response is null");
-                    }
                 }
                 else
                 {
```
abhipsaMisra, line 411: How about we add some additional context to the TimeoutException being thrown below (and maybe a log statement as well)?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Ideally it shouldn't but you are right adding a log is a better idea anyways.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -522,19 +519,23 @@ private async Task DisposeMessageAsync(string lockToken, AmqpIoTDisposeActions o
...
-                }
+                // This is desired property updates, so call the callback with TwinCollection.
+                _onDesiredStatePatchListener(twinCollection);
             }
             else
             {
-                // It is a PATCH, just call the callback with the TwinCollection
-                _onDesiredStatePatchListener(twinCollection);
+                if(correlationId.StartsWith(AmqpTwinMessageType.Get.ToString(), StringComparison.OrdinalIgnoreCase) ||
+                    correlationId.StartsWith(AmqpTwinMessageType.Patch.ToString(), StringComparison.OrdinalIgnoreCase))
```
abhipsaMisra, line 530: do we get a twin for the Patch operation as well, or is it null? how are the two cases handled when this task completes?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Not that I know of. We also already do this for twin (get twin response).

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Let me look into this.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
abhipsaMisra, line 287: again, extremely nit picky - this might fit into a helper method of sorts, or document which scenarios we'd like to retry for.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: No action here yet. If we go with implementing request response pattern for patch this will be where we'll add code and hence kept the comment.

abhipsaMisra: this could use a log statement here

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
abhipsaMisra: +1 to adding log warning here, if relevant

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
+                        // Do nothing
                     }
+                    _onDesiredPropertyReceived.Invoke(twin, correlationId, twinProperties);
```
abhipsaMisra, line 334: is this a desired property received callback, or a "something received on twin receiver link" callback? from the looks of it I think it is the second, so this could also use a renaming

e2e/test/iothub/twin/TwinE2ETests.cs
bikamani: Test expects an exception right away and that was working because earlier to this change we are throwing a wrong exception. Let me know if that makes sense.

mqtt works differently, it still throws right exception but not with error message. I will be looking into that next, not in scope of this PR

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: I don't really see a need of that but thanks for the suggestion

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: Yes no changes really. This is same as before, we were not doing anything with returned twin

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: This is just for completing a task which we stored in dictionary. The comment earlier was misleading.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: What would you like the comment to be if the error codes are not sufficient?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: I don't like the idea of making it that much longer. Can we overlay it on top of the first group?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -522,19 +519,23 @@ private async Task DisposeMessageAsync(string lockToken, AmqpIoTDisposeActions o
...
-                    task.SetResult(twin);
-                }
+                // This is desired property updates, so call the callback with TwinCollection.
+                _onDesiredStatePatchListener(twinCollection);
             }
             else
             {
-                // It is a PATCH, just call the callback with the TwinCollection
-                _onDesiredStatePatchListener(twinCollection);
+                if(correlationId.StartsWith(AmqpTwinMessageType.Get.ToString(), StringComparison.OrdinalIgnoreCase) ||
```
drwill-ms, line 529: space after `if`

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: I don't see a need for log statement here. we are also not logging for other success cases and really to the current implementation of updating reported properties it is not request response style.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
drwill-ms: Definitely

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
-                    Logging.Exit(this, amqpMessage, $"{nameof(OnDesiredPropertyReceived)}");
+                    Logging.Exit(this, amqpMessage, $"{nameof(OnTwinChangesReceived)}");
                 }
             }
         }
 
         #endregion Twin handling
 
-        internal static bool VerifyResponseMessage(AmqpMessage response)
+        internal static int getStatus(AmqpMessage response)
```
drwill-ms, line 348: Always TitleCase in C#

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
drwill-ms, line 322: lowercase updating

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: yes you are correct, it can use renaming. Do you have any suggestions?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
drwill-ms: Then I think it deserves a comment saying that.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
drwill-ms: Or initialize a variable with the parameter name in a line above, and pass that into the exception. I think that'd read even better.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -522,19 +519,23 @@ private async Task DisposeMessageAsync(string lockToken, AmqpIoTDisposeActions o
...
+                    correlationId.StartsWith(AmqpTwinMessageType.Patch.ToString(), StringComparison.OrdinalIgnoreCase))
+                {
+                    // For Get and Patch, complete the task.
+                    TaskCompletionSource<Twin> task;
+                    if (_twinResponseCompletions.TryRemove(correlationId, out task))
```
drwill-ms, line 534: What should we do if we don't find that correlation Id in the dictionary? Would that warrant logging a warning?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms, line 379: Why doesn't this method have an Async suffix?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
+                    if (status >= 500 || status == 429 || status == 408)
+                    {
+                        throw new IotHubException(error, true);
+                    }
+                    else
+                    {
+                        throw new IotHubException(error, false);
+                    }
+                }
+                else
```
drwill-ms, line 296: If previous code always throws an exception, we don't need an else. The remaining else block can be tabbed out and not be wrapped in a block.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
                 {
-                    if (amqpMessage.BodyStream != null)
+                    Twin twin = null;
+                    TwinCollection twinProperties = null;
+
+                    if (correlationId == null)
                     {
-                        // This a result of a GET TWIN so return (set) the full twin
+                        // Here we are getting desired property update notifications and want to handle it first
+                        using (StreamReader reader = new StreamReader(amqpMessage.BodyStream, System.Text.Encoding.UTF8))
```
drwill-ms, line 304: While you are here:
1. Use the new using style (no body block)
2. `var reader = new StreamReader...`

drwill-ms (changes requested):
Finished first pass.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -258,88 +258,104 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
                             var properties = JsonConvert.DeserializeObject<TwinProperties>(body);
                             twin = new Twin(properties);
                         }
                     }
-                    else
+                    else if (correlationId.StartsWith(AmqpTwinMessageType.Patch.ToString(), StringComparison.OrdinalIgnoreCase))
+                    {
+                        // This can be used to coorelate success response with Updating reported properties
+                    }
+                    else if (correlationId.StartsWith(AmqpTwinMessageType.Put.ToString(), StringComparison.OrdinalIgnoreCase))
```
abhipsaMisra, line 324: we don't seem to be handling the Put scenario inside TwinMessageListener, is this block (the new Twin()) being referenced elsewhere?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
drwill-ms: Looks good

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: There is a timeout based condition in this method RoundTripTwinMessage where it waits until a specific task is completed or timed out, perhaps that's why.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: sure

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: That's why it is a try remove. No logging.
That should never happen though. correlation Id is set by SDK on a request, this is where we are basically completing a task associated with this Id.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Are you suggesting to update comment to say "updating reported property is not a request response style implementation "?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Because it is not needed for Put.

abhipsaMisra: then where is the twin initialization done below used? if we don't reference the twin anywhere, the object initialization in L327 isn't necessary

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: You are right. I followed the same pattern and did not catch this.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: Can you elaborate @drwill-ms ?

I am concerned about the uniqueness of correlation Id if we shorten it so would like to understand what you meant.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
abhipsaMisra: if updating reported property isn't waiting for the service response, then which call is returning the exception thrown here: https://github.com/Azure/azure-iot-sdk-csharp/blob/8a3cd019acbe2fd4eebc568e76f7e9d01c1fc702/iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs#L287-L294

abhipsaMisra: > this could use a log statement here

Or did you mean that it will be implemented in a following PR? If you are implementing in a following PR then yeah, no need to add a log statement for the interim.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Just fyi: This is not a new code I added here but let me check that.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Exception is thrown for any twin message we get over amqp where a failure code is returned in message Annotation

abhipsaMisra: Sure, but as a caller, which operation/ API call would I need to surround with a try-catch to handle this exception?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: You do not need to worry about the uniqueness of guids, even if you clip the first 5 characters with this small of a data set. Do you realize how large and unique a number a guid is?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: Do you know how many bugs have occurred in history when the programmer said "that should never happen"?

You have to consider that not everything will happen as you expect. What does it hurt to throw a log statement in for this event, even if it may never happen? If it does, you'll be happy to have the logs.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: I can't see where in this PR where this method lives, so it seems you didn't add it. However, it isn't named properly. Can you rename it to fix it?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: Yes I do.

So what we need here is 
for example guid is - abcd12345
for Get we want it to be - **get**d12345

Is that correct?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: Yes I certainly did not add this method. 
I am open to any suggestions for renaming, don't have anything in mind which would fit better than what it is.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: It should be every twin operation be it get/patch/put

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: `RoundTripTwinMessageAsync`

bikamani: I get what you are saying now. Thanks for clarifying

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: yep

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: Resolving comments as I go through them in my local branch.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: added logging

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: added this - Retry for Http status code request timeout, Too many requests and server errors

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
abhipsaMisra: So `UpdateReportedPropertiesAsync` does wait for the response from the service (containing either a 204 or error code) before completing, right?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
abhipsaMisra: of course, I was pointing out there there might be a scope for us to avoid an unnecessary object initialization if we don't use the value anywhere

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
abhipsaMisra: how about "onTwinMessageReceived"?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: I mean for every twin response we are handling exception. We can chat if this is confusing offline.

bikamani: update reported properties is not a request response style so no to that. It's confusing I can totally understand.

Howerver I'll be adding a log statement here.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: yeah I like that, will update

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
abhipsaMisra: if the `UpdateReportedPropertiesAsync` call isn't waiting for the service response, then I still don't understand where the try-catch block would be for catching the above exception. Sure, ping me when you're available, we can chat offline.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Got it. 
It is not used apparently and feels safe to remove. Done

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: Does not seem necessary per understanding. If you think otherwise let me know.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: No it won't. try-catch block would not be catching exception in case of patch. OR in other words not until we change implementation

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -376,19 +377,19 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
...
         }
 
-        private async Task<Twin> RoundTripTwinMessage(AmqpTwinMessageType amqpTwinMessageType, TwinCollection reportedProperties, CancellationToken cancellationToken)
+        private async Task<Twin> RoundTripTwinMessageAsync(AmqpTwinMessageType amqpTwinMessageType, TwinCollection reportedProperties, CancellationToken cancellationToken)
         {
-            Logging.Enter(this, cancellationToken, $"{nameof(RoundTripTwinMessage)}");
+            Logging.Enter(this, cancellationToken, $"{nameof(RoundTripTwinMessageAsync)}");
 
-            string correlationId = amqpTwinMessageType + Guid.NewGuid().ToString();
+            string correlationId = amqpTwinMessageType + Guid.NewGuid().ToString().Substring(amqpTwinMessageType.ToString().Length);
```
bikamani, line 392: @drwill-ms If there is a better way to overlay Amqp message type such as "Get" on guid let me know.

drwill-ms: I've thought about it and can't come up with anything better than what you have. Perhaps you should make a private method out of it, though. Also, I wonder about caching the amqpTwinMessageType length so you don't have to call ToString and get the length each time.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -272,67 +272,57 @@ private void OnTwinChangesReceived(AmqpMessage amqpMessage)
...
-                    {
-                        throw new IotHubException(error, false);
-                    }
+                Twin twin = null;
+                TwinCollection twinProperties = null;
+
+                if (correlationId == null)
+                {
+                    // Here we are getting desired property update notifications and want to handle it first
+                    var reader = new StreamReader(amqpMessage.BodyStream, System.Text.Encoding.UTF8);
```
drwill-ms, line 295: Why not use `using` here instead of the dispose call on 298?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -272,67 +272,57 @@ private void OnTwinChangesReceived(AmqpMessage amqpMessage)
...
-                        twin = new Twin();
-                    }
-                    else
-                    {
-                        // This shouldn't happen
-                        // Do nothing
-                    }
-                    _onDesiredPropertyReceived.Invoke(twin, correlationId, twinProperties);
+                    // This shouldn't happen
+                    Logging.Info("Received a correlation Id for Twin operation that does not match Get, Patch or Put request", $"{nameof(OnTwinChangesReceived)}");
```
drwill-ms, line 323: FWIW, the `$"{` and `}"` around `nameof(OnTinChangesReceived)` is not needed. It is a string and you aren't formatting anything else into the string, so it is pointless work.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: I'm tempted to just go back to simple concat to keep it simple, and not worry about 3-5 extra characters.

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: On this line I can use "using" and have made changes to address this.

On line 282 however, it shows warning since we throw an exception at line 285 so that still needs to be disposed

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
abhipsaMisra: As we discussed, we don't want to throw exceptions that the caller cannot catch and that will cause the application to crash; we should add a check that the response is for a "get" call and then throw the exception, and we can add the same behavior for "patch/put" in the next PR.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
bikamani: I think the same for simplicity, will change it back

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -257,89 +257,98 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
-                    else
+                    _onTwinMessageReceived.Invoke(null, correlationId, null);
+                    if (correlationId.StartsWith(AmqpTwinMessageType.Get.ToString(), StringComparison.OrdinalIgnoreCase))
                     {
-                        // This is a desired property ack from the service
-                        twin = new Twin();
+                        // Handle failures
+                        _onTwinMessageReceived.Invoke(null, correlationId, null);
+                        string error = null;
+                        var reader = new StreamReader(amqpMessage.BodyStream, System.Text.Encoding.UTF8);
```
drwill-ms, line 288: using instead of explicit call to dispose (if an exception were to occur, it would not be disposed)

drwill-ms (approved):

barustum (approved):

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -522,19 +519,27 @@ private async Task DisposeMessageAsync(string lockToken, AmqpIoTDisposeActions o
...
+                {
+                    // For Get and Patch, complete the task.
+                    TaskCompletionSource<Twin> task;
+                    if (_twinResponseCompletions.TryRemove(correlationId, out task))
+                    {
+                        task.SetResult(twin);
+                    }
+                    else
+                    {
+                        Logging.Info("Could not remove a correlation Id from a dictionary", nameof(TwinMessageListener));
```
abhipsaMisra, line 540: how about rewording this to explain why the TryRemove could fail, something along the lines of - 
"attempted to complete the task awaiter for a twin operation whose request did not originate from the sdk - correlation ID was not set by the sdk"?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -257,89 +257,98 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
-                        }
-                    }
-                    else
+                    _onTwinMessageReceived.Invoke(null, correlationId, null);
+                    if (correlationId.StartsWith(AmqpTwinMessageType.Get.ToString(), StringComparison.OrdinalIgnoreCase))
                     {
-                        // This is a desired property ack from the service
-                        twin = new Twin();
+                        // Handle failures
+                        _onTwinMessageReceived.Invoke(null, correlationId, null);
```
abhipsaMisra, line 286: is this intended to be invoked twice - L282 and L286?

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
```diff
@@ -257,89 +257,98 @@ private void DisposeDelivery(AmqpMessage amqpMessage, bool settled, Accepted acc
...
+                        string error = null;
+                        var reader = new StreamReader(amqpMessage.BodyStream, System.Text.Encoding.UTF8);
+                        error = reader.ReadToEnd();
+                        reader.Dispose();
+                        // Retry for Http status code request timeout, Too many requests and server errors
```
abhipsaMisra, line 291: 👍🏽

abhipsaMisra (approved):

iothub/device/src/Transport/AmqpIoT/AmqpIoTReceivingLink.cs
bikamani: fixed it

PR #1797, opened 2021-02-18
fix(service-client): Update support for AzureSasCredential for a better user experience

vinagesh:
… experience

## Checklist

## Description of the changes
After having a discussion, we finally concluded that it would be a better user experience to just parse the DateTime from the SAS token instead of asking for the same information in another form from the user of the SDK.

The initial thought process was that the SAS token format might be a strict dependency. However, investigation led to finding out that the pattern is pretty fixed across azure and the signature in the token itself does not contain the delimiters that we are splitting on. So it should be safe to use this and have java and .NET clients work similarly.


iothub/service/src/IotHubSasCredentialProperties.cs
```diff
@@ -48,10 +50,28 @@ public override Task<CbsToken> GetTokenAsync(Uri namespaceAddress, string applie
             throw new InvalidOperationException($"IotHubSasCredential is not supported on NET451");
 
 #else
+            // Parse the SAS token to find the expiration date and time.
+            // SharedAccessSignature sr=ENCODED(dh://myiothub.azure-devices.net/a/b/c?myvalue1=a)&sig=<Signature>&se=<ExpiresOnValue>[&skn=<KeyName>]
+            var tokenParts = _credential.Signature.Split('&').ToList();
+            var expiresAtTokenPart = tokenParts.Where(tokenPart => tokenPart.StartsWith("se=", StringComparison.OrdinalIgnoreCase));
+
+            if (!expiresAtTokenPart.Any())
```
timtay-microsoft, line 58: Pretty cool use of linq here!

timtay-microsoft (approved):

PR #1798, opened 2021-02-19
doc(service-client): Update readme about the differnt client and oper…

vinagesh:
…ations

## Checklist

## Description of the changes
Updating readme with the functionality of our service clients and their operations and linking them in code.

### DIFF
--- a/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
+++ b/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.Devices
 {
     /// <summary>
     /// The Digital Twins Service Client contains methods to retrieve and update digital twin information, and invoke commands on a digital twin device.
+    /// For more information, see <see href="https://github.com/Azure/azure-iot-sdk-csharp#iot-hub-service-sdk"/>
     /// </summary>
     public class DigitalTwinClient : IDisposable
     {
--- a/iothub/service/src/JobClient/JobClient.cs
+++ b/iothub/service/src/JobClient/JobClient.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Azure.Devices
     {
         /// <summary>
         /// Creates a JobClient from the Iot Hub connection string.
+        /// For more information, see <see href="https://github.com/Azure/azure-iot-sdk-csharp#iot-hub-service-sdk"/>
         /// </summary>
         /// <param name="connectionString"> The Iot Hub connection string.</param>
         /// <returns> A JobClient instance. </returns>
--- a/iothub/service/src/RegistryManager.cs
+++ b/iothub/service/src/RegistryManager.cs
@@ -18,6 +18,7 @@ namespace Microsoft.Azure.Devices
 {
     /// <summary>
     /// Contains methods that services can use to perform create, remove, update and delete operations on devices.
+    /// For more information, see <see href="https://github.com/Azure/azure-iot-sdk-csharp#iot-hub-service-sdk"/>
     /// </summary>
     [System.Diagnostics.CodeAnalysis.SuppressMessage(
         "Naming",
--- a/iothub/service/src/ServiceClient.cs
+++ b/iothub/service/src/ServiceClient.cs
@@ -37,6 +37,7 @@ namespace Microsoft.Azure.Devices
 
     /// <summary>
     /// Contains methods that services can use to send messages to devices
+    /// For more information, see <see href="https://github.com/Azure/azure-iot-sdk-csharp#iot-hub-service-sdk"/>
     /// </summary>
     public abstract class ServiceClient : IDisposable
     {


iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.Devices
 {
     /// <summary>
     /// The Digital Twins Service Client contains methods to retrieve and update digital twin information, and invoke commands on a digital twin device.
+    /// For more information, see <see href="https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/RBAC/readme.md"/>
```
vinagesh, line 22: I don't know of a way to make this not tied to a branch. We have to change this after we move it to the preview or master. What process does everyone follow here? The other option is I can keep it as preview as we will most definitely release a preview version and even if we don't update this value, it will be there in the preview branch. The feature branch can get deleted after the work is over. At the moment I just have it as a task on my card to do this update.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.Devices
 {
     /// <summary>
     /// The Digital Twins Service Client contains methods to retrieve and update digital twin information, and invoke commands on a digital twin device.
+    /// For more information, see <see href="https://github.com/Azure/azure-iot-sdk-csharp/tree/previews/RBAC#iot-hub-service-sdk"/>
```
timtay-microsoft, line 22: be careful, that link is to our preview branch. I'd use a link to the master branch instead

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: Since the readme you are linking to should be the same between master and preview, I don't see why you couldn't make this link point towards the readme in the master branch, right?

readme.md
```diff
@@ -103,17 +103,17 @@ For details on OS support see the following resources:
...
+| [File Upload Notifications](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-file-upload)                    | :heavy_check_mark:        | AMQP | ServiceClient | Use your backend app to receive file upload notifications.
+| [IoT Hub Statistics](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-metrics)                          | :heavy_check_mark:        | HTTP | ServiceClient | Use your backend app to get IoT hub identity registry statistics such as total device count for device statistics, and connected device count for service statistics.
+| [Digital Twin Operations](https://docs.microsoft.com/en-us/azure/iot-pnp/overview-iot-plug-and-play)              | :heavy_check_mark:        | HTTP | DigitalTwinClient or RegistryManager | Use your backend app to perform operations on plug and play devices. The operations include get twins, update twins and invoke commands.                                              |
```
abhipsaMisra, line 116: Since the linked doc here points to pnp, we should guide the users to use the DigitalTwinClient here.

abhipsaMisra (approved):
This PR can be created against master directly, I don't think this needs to go into a preview branch first.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
drwill-ms, line 22: This link didn't work for me. Is that because the change hasn't gone into this branch yet?

drwill-ms (approved):

readme.md
```diff
@@ -103,17 +103,17 @@ For details on OS support see the following resources:
...
-| [File Upload](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-file-upload)                    | :heavy_check_mark:        | AMQP | Set up your backend app to receive file upload notifications. 
-| [Digital Twin Client](https://docs.microsoft.com/en-us/azure/iot-pnp/overview-iot-plug-and-play)              | :heavy_check_mark:        | HTTP | Set up your backend app to perform operations on plug and play devices.                                                      |
-| [IoT Hub Statistics](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-metrics)                          | :heavy_check_mark:        | HTTP | Get IoT Hub identity registry statistics; such as total device count for device statistics, and connected device count for service statistics. |
+| Features                                                                                                      | Support                  | Transport protocol used underneath | Client to use | Description                                                                                                              |
```
bikamani, line 106: Github shows it very differently in the PR, makes it hard to review. But I really appreciate you adding these changes! Super helpful for customers.

bikamani (approved):

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
vinagesh: Makes sense. I could just update the readme in master itself.

readme.md
vinagesh: True but we need to give both options right? I was going to write that one is preferred but it seemed to wordy for the table.

readme.md
vinagesh: Yes and I don't know how to make that better.

vinagesh:
> This PR can be created against master directly, I don't think this needs to go into a preview branch first.

Agreed. I'll update the readme directly in master and keep the links in the constructors.

readme.md
vinagesh: You can see the rich diff in github which makes it much better.

readme.md
vinagesh: I've added a note in the description in the PR against master directly.

PR #1803, opened 2021-02-25
Fix MqttTransportHandler to not await on user supplied C2D callback

abhipsaMisra:


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -555,7 +555,9 @@ private async Task HandleIncomingMessagesAsync()
                 Logging.Enter(this, "Process C2D message via callback", nameof(HandleIncomingMessagesAsync));
 
             Message message = ProcessMessage(true);
-            await (_deviceMessageReceivedListener?.Invoke(message) ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
```
abhipsaMisra, line 558: we don't need to await the completion of the user supplied callback, we can simply invoke it and move-on (we are not dependent on the completion of the callback to dispose any resources)

drwill-ms: Does it generate a warning that needs to be suppressed?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -555,7 +555,9 @@ private async Task HandleIncomingMessagesAsync()
...
+            _deviceMessageReceivedListener?.Invoke(message);
```
drwill-ms, line 558: If we can await it, I suppose it returns a task. In which case, if we want to signal to others we don't mean to wait on it, we should:

1. Comment that we intentionally do not await it
2. Receive the task with a discard `_ = _deviceMessageReceivedListener?.Invoke(message);` to signal we don't care about the return value

abhipsaMisra: sure thing, will add

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: a warning for...?

drwill-ms: unawaited async calls

azabbasi (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: it does not, no

abhipsaMisra: we are invoking a function, which returns a Task. there are no warnings regarding us not awaiting the returned task

PR #1806, opened 2021-02-25
feat(e2e) - Sas token E2E tests

vinagesh:
## Checklist

## Description of the changes
None of these tests run or work yet but once we have a working hub, it can be validated.

### DIFF
--- a/e2e/test/config/Configuration.IoTHub.cs
+++ b/e2e/test/config/Configuration.IoTHub.cs
@@ -3,10 +3,15 @@
 
 using System;
 using System.Security.Cryptography.X509Certificates;
+using System.Net;
+using System.Globalization;
+using System.Text;
+using System.Security.Cryptography;
 
 #if !NET451
 
 using Azure.Identity;
+using Azure;
 
 #endif
 
@@ -35,6 +40,16 @@ namespace Microsoft.Azure.Devices.E2ETests
                     GetValue("IOTHUB_TENANT_ID"),
                     GetValue("IOTHUB_CLIENT_ID"),
                     GetValue("IOTHUB_CLIENT_SECRET"));
+            }
+
+            public static string GetIotHubSharedAccessSignature(TimeSpan timeToLive)
+            {
+                ConnectionStringParser connectionString = new ConnectionStringParser(ConnectionString);
+                return GenerateSasToken(
+                    connectionString.IotHubHostName,
+                    connectionString.SharedAccessKey,
+                    timeToLive,
+                    connectionString.SharedAccessKeyName);
             }
 
 #endif
@@ -90,6 +105,39 @@ namespace Microsoft.Azure.Devices.E2ETests
             /// </summary>
             public const string InvalidProxyServerAddress = "127.0.0.1:1234";
 
+#if !NET451
+
+            private static string GenerateSasToken(string resourceUri, string sharedAccessKey, TimeSpan timeToLive, string policyName = default)
+            {
+                DateTime epochTime = new DateTime(1970, 1, 1);
+                DateTime expiresOn = DateTime.UtcNow.Add(timeToLive);
+                TimeSpan secondsFromEpochTime = expiresOn.Subtract(epochTime);
+                long seconds = Convert.ToInt64(secondsFromEpochTime.TotalSeconds, CultureInfo.InvariantCulture);
+                string expiry = Convert.ToString(seconds, CultureInfo.InvariantCulture);
+
+                string stringToSign = WebUtility.UrlEncode(resourceUri) + "\n" + expiry;
+
+                HMACSHA256 hmac = new HMACSHA256(Convert.FromBase64String(sharedAccessKey));
+                string signature = Convert.ToBase64String(hmac.ComputeHash(Encoding.UTF8.GetBytes(stringToSign)));
+
+                // SharedAccessSignature sr=ENCODED(dh://myiothub.azure-devices.net/a/b/c?myvalue1=a)&sig=<Signature>&se=<ExpiresOnValue>[&skn=<KeyName>]
+                string token = string.Format(
+                    CultureInfo.InvariantCulture,
+                    "SharedAccessSignature sr={0}&sig={1}&se={2}",
+                    WebUtility.UrlEncode(resourceUri),
+                    WebUtility.UrlEncode(signature),
+                    expiry);
+
+                if (!string.IsNullOrWhiteSpace(policyName))
+                {
+                    token += "&skn=" + policyName;
+                }
+
+                return token;
+            }
+
+#endif
+
             public class ConnectionStringParser
             {
                 public ConnectionStringParser(string connectionString)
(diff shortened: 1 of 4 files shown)


e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -0,0 +1,225 @@
...
+            var device = new Device(Guid.NewGuid().ToString());
+
+            // act
+            try
+            {
+                await registryManager.AddDeviceAsync(device).ConfigureAwait(false);
+            }
+            catch (Exception)
+            {
+                // TODO: Test and catch the correct exception.
```
timtay-microsoft, line 79: Don't forget this TODO. I'd expect an IotHubUnauthorizedException to be caught here

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -0,0 +1,225 @@
...
+            using var registryManager = RegistryManager.Create(
+                Configuration.IoTHub.GetIotHubHostName(),
+                sasCredential);
+
+            var device = new Device(Guid.NewGuid().ToString());
+
+            // act
+            try
+            {
+                await registryManager.AddDeviceAsync(device).ConfigureAwait(false);
```
timtay-microsoft, line 75: I'd add a failure case for if this line executes without exception. Maybe throw an AssertionFailedException("Expected previous line to throw an unauthorized exception");

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -0,0 +1,225 @@
...
+            Device createdDevice = await registryManager.AddDeviceAsync(device).ConfigureAwait(false);
+
+            // assert
+            Assert.IsNotNull(createdDevice);
+
+            // cleanup
+            await registryManager.RemoveDeviceAsync(device.Id).ConfigureAwait(false);
+        }
+
+        [Ignore]
```
timtay-microsoft, line 59: Why is this test ignored? Doesn't IoT Hub support this authentication type already?

e2e/test/config/Configuration.IoTHub.cs
```diff
@@ -90,6 +104,39 @@ public static X509Certificate2 GetIntermediate2Certificate()
...
+                TimeSpan secondsFromEpochTime = expiresOn.Subtract(epochTime);
+                long seconds = Convert.ToInt64(secondsFromEpochTime.TotalSeconds, CultureInfo.InvariantCulture);
+                string expiry = Convert.ToString(seconds, CultureInfo.InvariantCulture);
+
+                string stringToSign = WebUtility.UrlEncode(resourceUri) + "\n" + expiry;
+
+                HMACSHA256 hmac = new HMACSHA256(Convert.FromBase64String(sharedAccessKey));
+                string signature = Convert.ToBase64String(hmac.ComputeHash(Encoding.UTF8.GetBytes(stringToSign)));
+
+                // SharedAccessSignature sr=ENCODED(dh://myiothub.azure-devices.net/a/b/c?myvalue1=a)&sig=<Signature>&se=<ExpiresOnValue>[&skn=<KeyName>]
```
bikamani, line 122: Did you mean to keep this here?

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -0,0 +1,225 @@
...
+            await serviceClient.OpenAsync().ConfigureAwait(false);
+            using var message = new Message(Encoding.ASCII.GetBytes("Hello, Cloud!"));
+
+            try
+            {
+                await serviceClient.SendAsync(testDevice.Id, message);
+            }
+            catch (Exception)
+            {
+                // TODO: Test and catch the correct exception.
```
bikamani, line 212: Which exception is expected here?

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -0,0 +1,225 @@
...
+            // cleanup
+            await registryManager.RemoveDeviceAsync(device.Id).ConfigureAwait(false);
+        }
+
+        [Ignore]
+        [LoggedTestMethod]
+        public async Task RegistryManager_Http_SasCredentialAuth_Renewed_Success()
+        {
+            // arrange
+            string signature = Configuration.IoTHub.GetIotHubSharedAccessSignature(TimeSpan.FromHours(-1));
```
timtay-microsoft, line 64: This is probably me being paranoid, but I'd make the SAS token as expired as possible. We aren't trying to test how accurately the service evaluates when our SAS tokens expire. Maybe have them expire at the Unix Epoch rather than 1 hour in the past

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -0,0 +1,225 @@
...
+                        jobId,
+                        query,
+                        twin,
+                        DateTime.UtcNow,
+                        (long)TimeSpan.FromMinutes(2).TotalSeconds)
+                    .ConfigureAwait(false);
+            }
+            catch (ThrottlingException)
+            {
+                // Concurrent jobs can be rejected, but it still means authentication was successful. Ignore the exception.
```
timtay-microsoft, line 121: I wonder about this one, actually. Do you think the service actually validates the auth of a request before throttling? I'd expect it not to since validating the auth seems expensive

e2e/test/iothub/SasCredentialAuthenticationTests.cs
vinagesh: yupp. I have run all these tests and remove the ignore. I am also keeping a note of all the things in a task of mine.

e2e/test/iothub/SasCredentialAuthenticationTests.cs
timtay-microsoft: To add on to this, our test hubs only allow a few concurrent jobs at once, so I'd expect this exception frequently. I'd be wary of assuming this is a success case

e2e/test/iothub/SasCredentialAuthenticationTests.cs
vinagesh: yes that makes sense

e2e/test/iothub/SasCredentialAuthenticationTests.cs
vinagesh: Actually good point! This should be supported unlike TokenCredentials. Let me enable all of them.

e2e/test/iothub/SasCredentialAuthenticationTests.cs
timtay-microsoft, line 212: TODO

e2e/test/config/Configuration.IoTHub.cs
vinagesh: yes. This just shows how a SAS should look like.

e2e/test/iothub/SasCredentialAuthenticationTests.cs
vinagesh: I will know on testing. But expecting IotHubUnauthorized

e2e/test/iothub/SasCredentialAuthenticationTests.cs
vinagesh: I think that would be really bad if the service does not evaluate it correctly. I think it is better we catch it and report to them rather than simply trying to make our tests pass.

e2e/test/iothub/SasCredentialAuthenticationTests.cs
vinagesh: The service will not even receive the request if the auth fails. And the throttling is implemented in the service. So this would most definitely happen.

timtay-microsoft (approved):

PR #1807, opened 2021-02-26
feature(service-client): Adding chaching for aad tokens.

vinagesh:
## Checklist

## Description of the changes
 The HTTP APIs request for a token on every call. As getting an access token is a expensive operation, we should be cache these tokens. This PR adds caching for the AAD tokens. The AMQP protocol does not need this as we get a new token and update the CBS link only when the token is close to expiry.

### DIFF
--- a/iothub/service/src/DigitalTwin/Authentication/DigitalTwinTokenCredential.cs
+++ b/iothub/service/src/DigitalTwin/Authentication/DigitalTwinTokenCredential.cs
@@ -6,6 +6,7 @@ using System.Text;
 using System.Threading;
 using Azure.Core;
 using Microsoft.Azure.Devices.Authentication;
+using Microsoft.Azure.Devices.Common;
 
 namespace Microsoft.Azure.Devices.DigitalTwin.Authentication
 {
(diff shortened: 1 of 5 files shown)


timtay-microsoft (approved):

iothub/service/src/DigitalTwin/Authentication/DigitalTwinTokenCredential.cs
```diff
@@ -25,8 +28,19 @@ public DigitalTwinTokenCredential(TokenCredential credential)
 
         public override string GetAuthorizationHeader()
         {
-            AccessToken token = _credential.GetToken(new TokenRequestContext(), new CancellationToken());
-            return $"Bearer {token.Token}";
+            lock (_tokenLock)
+            {
+                // A new token is generated if it is the first time or the cached token is close to expiry.
+                if (_cachedAccessToken == null
```
drwill-ms, line 34: _cachedToken.HasValue &&

drwill-ms (approved):

azabbasi (approved):

PR #1812, opened 2021-03-04
fix(service-client)- Add IoT hub token scope.

vinagesh:
## Checklist

### DIFF
--- a/common/src/service/CommonConstants.cs
+++ b/common/src/service/CommonConstants.cs
@@ -5,13 +5,14 @@ namespace Microsoft.Azure.Devices.Common
 {
     using System;
 
-    static class CommonConstants
+    internal static class CommonConstants
     {
         // Custom HTTP response contents
         internal const string ErrorCode = "errorCode";
 
         // TODO: move these to ConfigProvider
         public const string DeviceAudienceFormat = "{0}/devices/{1}";
+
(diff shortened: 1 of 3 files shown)


common/src/service/CommonConstants.cs
```diff
@@ -118,8 +127,11 @@ static class CommonConstants
 
         // Custom HTTP headers
         public const string IotHubErrorCode = "IotHubErrorCode";
+
         public const string HttpErrorCodeName = "iothub-errorcode";
 
+        public const string IotHubAadTokenScope = "https://iothubs.azure.net/.default";
```
timtay-microsoft, line 133: Does .NET allow you to declare a const string[] here rather than creating a new string array each time the token is retrieved?

timtay-microsoft (approved):

bikamani (approved):

common/src/service/CommonConstants.cs
vinagesh: It does. I was just torn between that or this, went back and forth and finally chose the current implementation thinking it might be more readable to look at the implementation. But I do see your point and think making a const string[] might be better.

drwill-ms: FWIW, I would not choose the one that required array construction each time. `static readonly` is sufficiently comparable to `const`. I don't see a downside there.

PR #1813, opened 2021-03-04
Add .NET 5.0 support in IoT hub and provisioning sdks

abhipsaMisra:
## Checklist


iothub/device/src/Transport/AmqpIoT/AmqpIoTConnector.cs
```diff
@@ -49,14 +49,9 @@ public async Task<AmqpIoTConnection> OpenConnectionAsync(TimeSpan timeout)
                 MaxFrameSize = AmqpConstants.DefaultMaxFrameSize,
                 ContainerId = CommonResources.GetNewStringGuid(),
                 HostName = _hostName,
+                IdleTimeOut = Convert.ToUInt32(_amqpTransportSettings.IdleTimeout.TotalMilliseconds),
```
drwill-ms, line 52: You lost the null check here. Was that not necessary?

iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -562,12 +562,16 @@ private static StringContent CreateContent<T>(T entity)
         private static async Task<T> ReadAsAsync<T>(HttpContent content, CancellationToken token)
         {
             token.ThrowIfCancellationRequested();
+#if NET5_0
```
drwill-ms, line 565: You could also model this as:

```csharp
using Stream stream = await content.ReadAsStringAsync(
#if NET5_0
    token
#endif
    );
```

I don't think you have to do that, but it is an option.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -33,6 +32,12 @@
 using Microsoft.Azure.Devices.Shared;
 using Newtonsoft.Json;
 
+#if NET5_0
```
drwill-ms, line 35: This is getting quite messy. You might consider making a class somewhere that abstracts these changes away (calling TrySetResult or TryComplete).

iothub/device/src/Transport/AmqpIoT/AmqpIoTConnector.cs
abhipsaMisra: yeah, it wasn't required since a AmqpTransportSettings.IdleTimeout was TimeSpan and not TimeSpan? (nullable). 
These kinds of checks and validations are stronger in NET 5.0. Also, better support for co-orperative cancellation of a lot of async APIs :)

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -122,11 +122,19 @@ internal class ContractApiHttp : IContractApiHttp
                             $"The response message was null when executing operation {httpMethod}.");
                     }
 
+#if NET5_0
```
drwill-ms, line 125: Another example of a lot of DRY violations here, that wouldn't be as bad if you used the #if NET5_0 differently.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: Yeah, I can't say that I like it the way it is now. I am planning on removing dependency on DotNetty's TCS (no params variant) and bring that into our library directly.

provisioning/transport/http/src/Generated/RuntimeRegistration.cs
```diff
@@ -170,7 +170,11 @@ public RuntimeRegistration(DeviceProvisioningServiceRuntimeClient client)
                 var ex = new HttpOperationException(string.Format(CultureInfo.InvariantCulture, "Operation returned an invalid status code '{0}'", _statusCode));
                 if (_httpResponse.Content != null)
                 {
+#if NET5_0
```
drwill-ms, line 173: Is this file really generated? We may not have any other choice than to hand edit it, but I worry what it'll mean when/if we regenerate.

provisioning/service/src/Contract/ContractApiHttp.cs
abhipsaMisra: sure, I wasn't sure how the inline if-else params within a method call would be like, but I'll definitely try it out.

drwill-ms (approved):
Some things to consider

provisioning/transport/http/src/Generated/RuntimeRegistration.cs
abhipsaMisra: Yeah, so about this, it was auto-generated, but I couldn't locate the swagger for it in our repo. There also hasn't been any update on these files for over 2years.

abhipsaMisra: We might have generated it but might not have expected it to change (bad idea). For now, we've only giving it minimum possible updates.

e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
```diff
@@ -94,7 +94,7 @@ public static class FaultInjectionPoolingOverAmqp
                 watch.Start();
 
                 logger.Trace($"{nameof(FaultInjectionPoolingOverAmqp)}: {testDevices[0].Id} Requesting fault injection type={faultType} reason={reason}, delay={delayInSec}s, duration={durationInSec}s");
-                var faultInjectionMessage = FaultInjection.ComposeErrorInjectionProperties(faultType, reason, delayInSec, durationInSec);
+                using Client.Message faultInjectionMessage = FaultInjection.ComposeErrorInjectionProperties(faultType, reason, delayInSec, durationInSec);
```
jamdavi, line 97: Why did you make the type explicit?

e2e/test/iothub/service/IoTHubCertificateValidationE2ETest.cs
```diff
@@ -18,9 +18,9 @@ public class IoTHubCertificateValidationE2ETest : E2EMsTestBase
         [LoggedTestMethod]
         public async Task RegistryManager_QueryDevicesInvalidServiceCertificateHttp_Fails()
         {
-            var rm = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionStringInvalidServiceCertificate);
+            using var rm = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionStringInvalidServiceCertificate);
             IQuery query = rm.CreateQuery("select * from devices");
-            var exception = await Assert.ThrowsExceptionAsync<IotHubCommunicationException>(
+            IotHubCommunicationException exception = await Assert.ThrowsExceptionAsync<IotHubCommunicationException>(
```
jamdavi, line 23: Same, why the explicit change?

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -300,8 +300,8 @@ public async Task ModulesClient_IdentityLifecycle()
         [LoggedTestMethod]
         public async Task ModulesClient_DeviceTwinLifecycle()
         {
-            RegistryManager client = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
-            var module = await TestModule.GetTestModuleAsync(_devicePrefix, _modulePrefix, Logger).ConfigureAwait(false);
+            using RegistryManager client = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+            TestModule module = await TestModule.GetTestModuleAsync(_devicePrefix, _modulePrefix, Logger).ConfigureAwait(false);
```
jamdavi, line 304: Same, why the explicit change?

provisioning/service/src/Contract/ContractApiHttp.cs
jamdavi: I think if the items that are different between frameworks still return the same type a private method would be helpful so you can isolate the NET5_0 compiler statements, or even a local function that does the same thing

provisioning/transport/http/src/TPM/TpmDelegatingHandler.cs
```diff
@@ -37,12 +37,20 @@ public TpmDelegatingHandler(SecurityProviderTpm securityProvider)
...
+#if NET5_0
+                if (request.Options.TryGetValue(new HttpRequestOptionsKey<object>(ProvisioningHeaderName), out object result))
+#else
                 if (request.Properties.TryGetValue(ProvisioningHeaderName, out object result))
+#endif
                 {
                     if (result is Action<string> setSasToken)
                     {
                         string target = GetTarget(request.RequestUri.LocalPath);
+#if NET5_0
```
jamdavi, line 49: Yeah seeing these further down creating a helper that handles the httpResponse ReadAsString with the CTS would go a long way.

jamdavi (approved):
Notes about specific types, and +1 to David's note about the compiler directives.

e2e/test/Helpers/templates/FaultInjectionPoolingOverAmqp.cs
abhipsaMisra: This is because the returned type isn't obvious based on the method call. While we prefer to use `var` only when the type is obvious, eg. when doing a `new <ClassName>()`, we have deviations from that in a lot of places in our code. My approach is to fix it as I go.
This change isn't related to adding NET 5.0 support.

vinagesh:
Is this PR supposed to be a draft? I was curious and I saw it has already been reviewed.

abhipsaMisra:
> Is this PR supposed to be a draft? I was curious and I saw it has already been reviewed.

My intention is to mark it as "Ready to be reviewed" once the test failures have been fixed. I don't want to inundate your inboxes with notifications while I figure out the test failures! :)

abhipsaMisra:
This PR was getting a bit difficult to review, so I've opened a new PR #1836 . Closing this PR.

PR #1819, opened 2021-03-09
build(all): Ignore Edge from Policheck

jamdavi:
Every time we have the word "Edge" shows up in our code we get a violation that we need to suppress. This fix will ignore this term and only this term in both the code and comments.


drwill-ms (approved):

vsts/vsts.yaml
```diff
@@ -233,7 +233,7 @@ jobs:
 
     condition: succeeded()
     pool:
-      name: Hosted VS2017
+      name: vs2017-win2016
```
jamdavi, line 236: Changed this to the correct Azure Pipelines name. After talking to someone from the PoliCheck engineering group it was suggested we use the same hosted pipeline as they do.

vsts/vsts.yaml
```diff
@@ -245,7 +245,14 @@ jobs:
 
       - task: ComponentGovernanceComponentDetection@0
         displayName: "Component Detection"
-
+      
+      - task: securedevelopmentteam.vss-secure-development-tools.build-task-policheck.PoliCheck@1
```
jamdavi, line 249: Moved this to the top to avoid any possible tool interactions that might change the behavior.

Issue #1821, opened 2021-03-10
[Bug Report] Semaphore release mismatch in InternalClient.cs


vipeller:
The implementation of SetMethodHandlerAsync() in InternalClient.cs seems to have an error that may cause releasing the semaphore (used as a lock) acquired by a different (logical) thread:

```
        public async Task SetMethodHandlerAsync(string methodName, MethodCallback methodHandler, object userContext, CancellationToken cancellationToken)
        {
            try
            {
                cancellationToken.ThrowIfCancellationRequested();
   1)           await _methodsDictionarySemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);

                // ... code removed for brevity 
            }
            catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
            {
                cancellationToken.ThrowIfCancellationRequested();
                throw;
            }
            finally
            {
                try
                {
    2)            _methodsDictionarySemaphore.Release();
                }
    3)          catch (SemaphoreFullException)
                {
                    // this semaphore is typically grabbed at the start of the method, but if
                    // this method is canceled while waiting to grab the semaphore, then this semaphore.release
                    // will throw this SemaphoreFullException since it was never grabbed in the first place
                    if (Logging.IsEnabled)
                    {
                        Logging.Info(this, "SemaphoreFullException thrown while releasing methods dictionary semaphore, but will be ignored since that means the semaphore is available for other threads to grab again anyways");
                    }
                }                
            }
        }

```

Let's say, that a call on logical thread A) was made to this method, it acquired the semaphore and now it is working inside 'EnableMethodAsync()' or 'DisableMethodAsync()' (removed from the code above, but you can see it in the original code). Note, that although those calls pass 'cancellationToken', inside their implementation in case of Amqp they switch to an 'operationTimeout'. I mention this to emphasize that the call not necessarily will interrupt when the cancellationToken gets cancelled.

Now let's say that this operation takes really long and in the meantime a new call on thread B) is made. This call blocks on 1) and keeps hanging for a while, but eventually the cancellationToken cancels, so on Thread B) an exception will interrupt this call and lands in the 'finally' block.

Now on thread B) it executes a Release() call on the semaphore. Because earlier Thread A) decreased the counter of the semaphore, there is room to release it once, so this call will not fail (and does not get to 3) ). 

[...]

PR #1823, opened 2021-03-10
fix(iot-dev): Fix semaphore usage within InternalClient

timtay-microsoft:
This PR fixes the InternalClient layer's semaphore usage so that each semaphore will never be released within a method unless it was grabbed. The new pattern that I'm implementing looks like the below:

```csharp
public async Task Foo(CancellationToken cancellationToken)
{
    await _semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);

    try
    {
        // Perform all operations that the semaphore protects like updating local variable states, subscribing over AMQP/MQTT, etc
        // Any number of timeouts, IOExceptions, etc may be thrown here, but the semaphore will be released regardless
    }
    finally
    {
        _methodsDictionarySemaphore.Release();
    }
}
```

With this approach, the finally block only executes if the semaphore is grabbed, and it will be released no matter what exceptions are encountered after grabbing it. Additionally, if the provided cancellation token expires before or during the wait for grabbing the semaphore, an exception will be thrown, and the semaphore won't released since these timeouts don't occur in the try/finally block.

#1821 for additional context.

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -1414,7 +1414,7 @@ namespace Microsoft.Azure.Devices.Client
 
             if (callback != null)
             {
-                await (callback.Invoke(message, callbackContext) ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
+                await callback.Invoke(message, callbackContext).ConfigureAwait(false);
             }
 
             if (Logging.IsEnabled)
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -1405,13 +1405,10 @@ namespace Microsoft.Azure.Devices.Client
                 return;
             }
 
-            Tuple<ReceiveMessageCallback, object> callbackContextTuple = null;
-
             await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
             try
             {
-                callbackContextTuple = _deviceReceiveMessageCallback;
-                await (callbackContextTuple?.Item1?.Invoke(message, callbackContextTuple.Item2)
+                await (_deviceReceiveMessageCallback?.Item1?.Invoke(message, _deviceReceiveMessageCallback.Item2)
                     ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
             }
             finally
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -1406,19 +1406,18 @@ namespace Microsoft.Azure.Devices.Client
             }
 
             await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
-            try
+            ReceiveMessageCallback callback = _deviceReceiveMessageCallback?.Item1;
+            object callbackContext = _deviceReceiveMessageCallback?.Item2;
+            _deviceReceiveMessageSemaphore.Release();
+
+            if (callback != null)
             {
-                await (_deviceReceiveMessageCallback?.Item1?.Invoke(message, _deviceReceiveMessageCallback.Item2)
-                    ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
+                await (callback.Invoke(message, callbackContext) ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
             }
-            finally
+
+            if (Logging.IsEnabled)
             {
-                _deviceReceiveMessageSemaphore.Release();
-             
-                if (Logging.IsEnabled)
-                {
-                    Logging.Exit(this, message, nameof(OnDeviceMessageReceivedAsync));
-                }
+                Logging.Exit(this, message, nameof(OnDeviceMessageReceivedAsync));
             }
         }
 
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -1402,7 +1402,9 @@ namespace Microsoft.Azure.Devices.Client
 
             try
             {
+                await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
                 callbackContextTuple = _deviceReceiveMessageCallback;
+                _deviceReceiveMessageSemaphore.Release();
                 await (callbackContextTuple?.Item1?.Invoke(message, callbackContextTuple.Item2)
                     ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
             }
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -1405,6 +1405,8 @@ namespace Microsoft.Azure.Devices.Client
                 return;
             }
 
+            // Grab this semaphore so that there is no chance that the _deviceReceiveMessageCallback instance is set in between the read of the 
+            // item1 and the read of the item2
             await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
             ReceiveMessageCallback callback = _deviceReceiveMessageCallback?.Item1;
             object callbackContext = _deviceReceiveMessageCallback?.Item2;
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -771,16 +771,16 @@ namespace Microsoft.Azure.Devices.Client
         /// </summary>
         public async Task SetMethodHandlerAsync(string methodName, MethodCallback methodHandler, object userContext, CancellationToken cancellationToken)
         {
+            if (Logging.IsEnabled)
+            {
+                Logging.Enter(this, methodName, methodHandler, userContext, nameof(SetMethodHandlerAsync));
+            }
+
+            cancellationToken.ThrowIfCancellationRequested();
+            await _methodsDictionarySemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+
             try
             {
-                if (Logging.IsEnabled)
-                {
-                    Logging.Enter(this, methodName, methodHandler, userContext, nameof(SetMethodHandlerAsync));
-                }
-
-                cancellationToken.ThrowIfCancellationRequested();
-                await _methodsDictionarySemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
-
                 if (methodHandler != null)
                 {
                     // codes_SRS_DEVICECLIENT_10_005: [ It shall EnableMethodsAsync when called for the first time. ]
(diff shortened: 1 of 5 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -819,20 +819,7 @@ public async Task SetMethodHandlerAsync(string methodName, MethodCallback method
             }
             finally
             {
-                try
-                {
-                    _methodsDictionarySemaphore.Release();
-                }
-                catch (SemaphoreFullException)
```
timtay-microsoft, line 826: This catch was part of an earlier attempt at fixing a similar bug, but it won't be needed anymore since this finally block will only be reached if the semaphore was grabbed

iothub/device/src/InternalClient.cs
```diff
@@ -1835,10 +1753,11 @@ public async Task SetMessageHandlerAsync(MessageHandler messageHandler, object u
                 Logging.Enter(this, messageHandler, userContext, nameof(SetMessageHandlerAsync));
             }
 
+            cancellationToken.ThrowIfCancellationRequested();
+            await _moduleReceiveMessageSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+
             try
             {
-                cancellationToken.ThrowIfCancellationRequested();
-                await _moduleReceiveMessageSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
timtay-microsoft, line 1841: As per the model I described in the description of this PR, grabbing the semaphore and checking the cancellation token must be outside of the try/finally block to ensure that we don't try to release a semaphore that we never grabbed.

iothub/device/src/InternalClient.cs
```diff
@@ -1468,11 +1402,7 @@ private async Task OnDeviceMessageReceivedAsync(Message message)
 
             try
             {
-                // Wait to acquire the _deviceReceiveMessageSemaphore. This ensures that we get a reference to the latest callback set on the device client.
-                await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
                 callbackContextTuple = _deviceReceiveMessageCallback;
-                _deviceReceiveMessageSemaphore.Release();
```
timtay-microsoft, line 1474: This change is a bit orthogonal, but there was no reason to use a semaphore here. It reads from a shared resource, but it doesn't write to the shared resource so there is no concurrency risk

iothub/device/src/InternalClient.cs
```diff
@@ -1468,11 +1402,7 @@ private async Task OnDeviceMessageReceivedAsync(Message message)
 
             try
             {
-                // Wait to acquire the _deviceReceiveMessageSemaphore. This ensures that we get a reference to the latest callback set on the device client.
-                await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
```
abhipsaMisra, line 1472: This lock shouldn't be removed. We want to ensure that the callback is invoked on the latest supplied handler.

iothub/device/src/InternalClient.cs
abhipsaMisra: What about the scenario where a user supplies a different callback right before the client was about to actually invoke it?

iothub/device/src/InternalClient.cs
timtay-microsoft: Even with this lock, the handler could get updated immediately after this value is read. I can remove this for now since it isn't critical for fixing the github issue, but I don't see the value of leaving it.

iothub/device/src/InternalClient.cs
timtay-microsoft: The callback can be updated immediately after this value read even with this lock. Even with the semaphore, there is no guarantee that the read callback will still be valid after a millisecond. Like I mentioned on the other comment, I can remove this change since it isn't critical to fix the github issue in question, but we should still consider removing it at some point

iothub/device/src/InternalClient.cs
abhipsaMisra: I agree, it could, but since we invoke it immediately after, it does reduce the scope of us invoking the handler on an updated callback. Also, you could release the semaphore after invoking the callback, that way a user won't be able to update the callback in the interim.

iothub/device/src/InternalClient.cs
```diff
@@ -1009,22 +984,8 @@ internal async Task OnMethodCalledAsync(MethodRequestInternal methodRequestInter
...
-                    // this method is canceled while waiting to grab the semaphore, then this semaphore.release
-                    // will throw this SemaphoreFullException since it was never grabbed in the first place
-                    if (Logging.IsEnabled)
-                    {
-                        Logging.Info(this, "SemaphoreFullException thrown while releasing methods dictionary semaphore," +
-                            " but will be ignored since that means the semaphore is available for other threads to grab again anyways");
-                    }
-                }
+                methodResponseInternal?.Dispose();
+                _methodsDictionarySemaphore.Release();
```
drwill-ms, line 988: Something to consider is that if the call to Dispose throws, then we won't release the semaphore

drwill-ms (approved):

iothub/device/src/InternalClient.cs
jamdavi: Looking at this more If we're trying to guarantee that the callback is guaranteed we'd need to do the following. If we don't care then Tim is correct the Semaphore is not helping here.

1. Initialize a new tuple so we get the correct references
2. Move the lock to after the callback finished.

```
            try
            {
                // Wait to acquire the _deviceReceiveMessageSemaphore. This ensures that we get a reference to the latest callback set on the device client.
                await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
                Tuple<ReceiveMessageCallback, object> callbackContextTuple = new Tuple<ReceiveMessageCallback, object>(_deviceReceiveMessageCallback.Item1, _deviceReceiveMessageCallback.Item2);
                _deviceReceiveMessageSemaphore.Release();

                await (callbackContextTuple?.Item1?.Invoke(message, callbackContextTuple.Item2)
                    ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
            }
            finally
            {
                if (Logging.IsEnabled)
                {
                    Logging.Exit(this, message, nameof(OnDeviceMessageReceivedAsync));
                }
            }
```

iothub/device/src/InternalClient.cs
timtay-microsoft: I'll move the semaphore here to fully wrap the scope of the read and invoke of the callback then, sure

iothub/device/src/InternalClient.cs
timtay-microsoft: Good point. This will require a small rework to take that into account

bikamani (approved):

bikamani:
> This PR fixes the InternalClient layer's semaphore usage so that each semaphore will never be released within a method unless it was grabbed. The new pattern that I'm implementing looks like the below:
> 
> ```cs
> public async Task Foo(CancellationToken cancellationToken)
> {
>     cancellationToken.ThrowIfCancellationRequested();
>     await _semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
> 
>     try
>     {
>         // Perform all operations that the semaphore protects like updating local variable states, subscribing over AMQP/MQTT, etc
>         // Any number of timeouts, IOExceptions, etc may be thrown here, but the semaphore will be released regardless
>     }
>     finally
>     {
>         _methodsDictionarySemaphore.Release();
>     }
> }
> ```
> 
> With this approach, the finally block only executes if the semaphore is grabbed, and it will be released no matter what exceptions are encountered after grabbing it. Additionally, if the provided cancellation token expires before or during the wait for grabbing the semaphore, an exception will be thrown, and the semaphore won't released since these timeouts don't occur in the try/finally block.
> 
> #1821 for additional context.

Wondering is there a way to add a test for this scenario for future protection?

timtay-microsoft:
> Wondering is there a way to add a test for this scenario for future protection?

I don't know of any way to test that scenario since it relies on the randomness of threading.

iothub/device/src/InternalClient.cs
timtay-microsoft: Yep. Since I'm getting some pushback on removing the semaphore grab here, I did what you suggested above, but without declaring the tuple. Now it is 

```csharp
            await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
            try
            {
                await (_deviceReceiveMessageCallback?.Item1?.Invoke(message, _deviceReceiveMessageCallback.Item2)
                    ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
            }
            finally
            {
                _deviceReceiveMessageSemaphore.Release();
             
                if (Logging.IsEnabled)
                {
                    Logging.Exit(this, message, nameof(OnDeviceMessageReceivedAsync));
                }
            }
```

timtay-microsoft: Now the lock is a bit more useful since it makes sure that the two calls to get values from the callback (.Item1 and .Item2) will come from the same callback instance. Without the lock, there would be a chance that the callback value is set inbetween the retrieval of item1 and item2

abhipsaMisra (approved):
lgtm!

iothub/device/src/InternalClient.cs
```diff
@@ -1464,20 +1405,16 @@ private async Task OnDeviceMessageReceivedAsync(Message message)
...
+            await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
             try
             {
-                // Wait to acquire the _deviceReceiveMessageSemaphore. This ensures that we get a reference to the latest callback set on the device client.
-                await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
-                callbackContextTuple = _deviceReceiveMessageCallback;
-                _deviceReceiveMessageSemaphore.Release();
-
-                await (callbackContextTuple?.Item1?.Invoke(message, callbackContextTuple.Item2)
+                await (_deviceReceiveMessageCallback?.Item1?.Invoke(message, _deviceReceiveMessageCallback.Item2)
```
drwill-ms, line 1411: I think this would be easier to read if you just checked for null

drwill-ms (approved):

iothub/device/src/InternalClient.cs
jamdavi: Not sure I like this since we will block the semaphore until method is complete. Let's get a reference to the two items from the Tuple during the Semaphore operation in the Semaphore context and then invoke the callback after we've released the lock.

jamdavi: (code written not compiled)

```C#
await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
var item1 = _deviceReceiveMessageCallback?.Item1;
var item2 = _deviceReceiveMessageCallback?.Item2;
_deviceReceiveMessageSemaphore.Release();
            try
            {
                await (item1?.Invoke(message, item2)
                    ?? TaskHelpers.CompletedTask).ConfigureAwait(false);
            }
            finally
            {            
                if (Logging.IsEnabled)
                {
                    Logging.Exit(this, message, nameof(OnDeviceMessageReceivedAsync));
                }
            }
```

timtay-microsoft: Good point. Callbacks can take a pretty variable amount of time

abhipsaMisra: Wait, how is this different from the code that we had before?

jamdavi (approved):

iothub/device/src/InternalClient.cs
abhipsaMisra: There isn't a need for us to await the completion of this user supplied callback, we can simply invoke it and then release the semaphore.

timtay-microsoft: As of now, the code behaves the same way as it used to. I just refactored it a bit for readability

PR #1824, opened 2021-03-11
fix(e2e) - Change event logging to opt in to specific events only.

vinagesh:
## Checklist

## Description of the changes
This change ensures that our E2E tests only subscribe to a specific set of event sources. After this change is checked in, I will take the changes into the aad branch as well. Having this change in master directly is beneficial and can help us upgrade the storage sdk.

### DIFF
--- a/e2e/test/Helpers/ConsoleEventListener.cs
+++ b/e2e/test/Helpers/ConsoleEventListener.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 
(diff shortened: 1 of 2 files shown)


e2e/test/Helpers/ConsoleEventListener.cs
```diff
@@ -1,89 +1,50 @@
...
-            }
-        }
-
         protected override void OnEventSourceCreated(EventSource eventSource)
         {
-            base.OnEventSourceCreated(eventSource);
-            EnableEvents(
-                eventSource,
-                EventLevel.LogAlways
+            if (_eventFilters.Any(filter => eventSource.Name.StartsWith(filter, StringComparison.OrdinalIgnoreCase)))
```
vinagesh, line 17: Only subscribe to events in the filter list

e2e/test/Helpers/ConsoleEventListener.cs
```diff
@@ -1,89 +1,50 @@
...
-            {
-                return;
-            }
-
             lock (_lock)
             {
-                if (_eventFilters.Any(ef => eventData.EventSource.Name.StartsWith(ef, StringComparison.Ordinal)))
-                {
-                    string eventIdent;
+                string eventIdent;
```
vinagesh, line 34: We already subscribed to a select set of events so we don't need any filtering here.

e2e/test/Helpers/ConsoleEventListener.cs
```diff
@@ -1,89 +1,50 @@
...
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 
 namespace System.Diagnostics.Tracing
 {
     public sealed class ConsoleEventListener : EventListener
     {
-        private readonly string[] _eventFilters;
+        private static readonly string[] _eventFilters = new string[] { "DotNetty-Default", "Microsoft-Azure-Devices", "Azure-Core", "Azure-Identity" };
```
vinagesh, line 12: Azure-Core and identity will be required later when we merge feature to here. But no harm keeping it in the list. We won't even get those events so should be a no-op.

e2e/test/Helpers/ConsoleEventListener.cs
```diff
@@ -1,89 +1,50 @@
...
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 
 namespace System.Diagnostics.Tracing
 {
     public sealed class ConsoleEventListener : EventListener
     {
-        private readonly string[] _eventFilters;
+        private static readonly string[] _eventFilters = new string[] { "DotNetty-Default", "Microsoft-Azure-Devices" };
```
abhipsaMisra, line 12: One suggestion is to allow the console logger to be initialized with a list of event providers. The reason for this is, we recommend customers use this class to help them debug their test code, and they wouldn't want all of our sdk relevant event sources to be written by default.

abhipsaMisra (approved):

e2e/test/Helpers/ConsoleEventListener.cs
drwill-ms: Doesn't that assume customers log to ETW too?

drwill-ms (approved):

e2e/test/Helpers/ConsoleEventListener.cs
vinagesh: Here is what I found on testing. OnEventSourceCreated get called even before we populate the _eventFilters in the constructor. It was extremely odd but happened. Hence, I removed that. It is possible to miss subscribing to some if we go that route. It becomes misleading. I am open to other suggestions.

PR #1826, opened 2021-03-11
feat(e2e) - Adding azure core and identity event sources.

vinagesh:
## Checklist

### DIFF
--- a/e2e/test/Helpers/ConsoleEventListener.cs
+++ b/e2e/test/Helpers/ConsoleEventListener.cs
@@ -13,13 +13,13 @@ namespace System.Diagnostics.Tracing
         // OnEventSourceCreated is triggered as soon as the EventListener is registered and an event source is created.
         // So trying to configure this value in the ConsoleEventListener constructor does not work.
         // The OnEventSourceCreated can be triggered sooner than the filter is initialized in the ConsoleEventListener constructor.
-        private static string[] _eventFilters = new string[] { "DotNetty-Default", "Microsoft-Azure-Devices" };
+        private static readonly string[] s_eventFilters = new string[] { "DotNetty-Default", "Microsoft-Azure-Devices", "Azure-Core", "Azure-Identity" };
 
         private readonly object _lock = new object();
 
         protected override void OnEventSourceCreated(EventSource eventSource)
         {
-            if (_eventFilters.Any(filter => eventSource.Name.StartsWith(filter, StringComparison.OrdinalIgnoreCase)))
+            if (s_eventFilters.Any(filter => eventSource.Name.StartsWith(filter, StringComparison.OrdinalIgnoreCase)))
             {
                 base.OnEventSourceCreated(eventSource);
                 EnableEvents(


drwill-ms (approved):

e2e/test/Helpers/ConsoleEventListener.cs
```diff
@@ -13,7 +13,7 @@ public sealed class ConsoleEventListener : EventListener
         // OnEventSourceCreated is triggered as soon as the EventListener is registered and an event source is created.
         // So trying to configure this value in the ConsoleEventListener constructor does not work.
         // The OnEventSourceCreated can be triggered sooner than the filter is initialized in the ConsoleEventListener constructor.
-        private static string[] _eventFilters = new string[] { "DotNetty-Default", "Microsoft-Azure-Devices" };
+        private static string[] _eventFilters = new string[] { "DotNetty-Default", "Microsoft-Azure-Devices", "Azure-Core", "Azure-Identity" };
```
drwill-ms, line 16: readonly?

abhipsaMisra, line 16: "readonly", "s_eventFilters"?

abhipsaMisra: We'll also need to update our documentation here: https://github.com/Azure/azure-iot-sdk-csharp/tree/master/tools/CaptureLogs#console-logging

abhipsaMisra: Would the identity team let us know when they have resolved the incompatibility issue with the diagnostics implementation? I'd like for this to be configurable again soon :)

abhipsaMisra (approved):

e2e/test/Helpers/ConsoleEventListener.cs
vinagesh: What do you mean my configurable? That has nothing to do with the Identity team. The only thing it will do is, we can subscribe to every event source instead of just what we care about. But I think it is a best practice to to only subscribe to what we want.

I will update the readme too, it is a task on my list. Once we enable all tests and I am sure of all logs and providers, I will update the logging stuff.

@abhipsaMisra - why s_eventFilters ? What is the significance of s_? Usually I've not seen this pattern of naming in c#. https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions. As far as I know, the underscore is only to signify private.

For readonly, I did not choose that because we really do not want to change this value in the constructor. It is ideally a const but we cannot declare it as a const.

vinagesh: For readonly, I did not choose that because we really do not want to change this value in the constructor. It is ideally a const but we cannot declare it as a const due to the new string{}. If you see my comments above, it explains why we should not be changing this value in the constructor. I felt it might be misleading to use readonly.

drwill-ms: s_ is our prefix for static member fields.

It should be readonly if it will not change after construction. It does not have to be assigned to or changed in the ctor. That is just the last opportunity (besides inline init) to set it before it is effectively const.

abhipsaMisra: Is there a performance degradation if we subscribe to everything? If not, I'd rather we subscribe to everything and print only what we need vs modifying the sourcecode to filter the events.
This is just my preference based on the ease of reusability of this class; i.e. if I have 2 different applications for service and device operations, I'll need to maintain different copies of this file in order to filter only device/ service client related events.

`s_` is for static fields: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/doc/coding-style.md#specific-rules

readonly - I get your point about people wanting to update the value of a readonly field within a constructor, but they can do that even if it isn't marked as readonly. How does this protect us now?

vinagesh: Clarified offline. The comment was to change from 'static' to 'static readonly'. I'll make the change.

vinagesh: yes there is a performance degradation and it is not the best practice to subscribe to everything. Also if you look at track 2, they do it similarly. They subscribe only to what is needed. What is the advantage of filtering while writing? 

How can we avoid copies of the class if the filters are not configurable in the constructor due to the callback subscriptions happening much earlier? It is a corner case but it is very much possible to miss some logs.

vinagesh: I was not not aware of our guidelines on static fields. Will fix that.

Issue #1822, opened 2021-03-10
[Bug Report] Removing/Adding method handlers leads to error


vipeller:
using the following code, it always ends in exception (the second SetMethodDefaultHandlerAsync() call times out)

```
            var client = Microsoft.Azure.Devices.Client.DeviceClient.CreateFromConnectionString(
                            "HostName=[xy].azure-devices.net;DeviceId=dev_a;SharedAccessKey=[somekeys]", Microsoft.Azure.Devices.Client.TransportType.Amqp_Tcp_Only);

            client.OperationTimeoutInMilliseconds = 20000;

            try
            {
                Console.WriteLine("unsubscribing...");
                await client.SetMethodDefaultHandlerAsync(null, null);
                Console.WriteLine("unsubscribed...");

                Console.WriteLine("subscribing...");
                await client.SetMethodDefaultHandlerAsync(MethodHandler, null);
                Console.WriteLine("subscribed...");
            }
            catch (Exception e)
            {
                Console.WriteLine(e);
            }

```

Using: 
Microsoft.Azure.Devices 1.30.0
Microsoft.Azure.Devices.Client 1.35.0
(.NET core 3.1 running on win 10)

The context is that EdgeHub, when a client connects, in some circumstances it clears all subscriptions (this part is simulated by the first call in the example code), then when the client subscribes to methods, that call fails (simulated by the second call)

This is related to the investigation reported in: 
https://github.com/Azure/azure-iot-sdk-csharp/issues/1821
However that bug is not the cause of this one most probably.

I am not sure that it is a regression or it always have failed.

timtay-microsoft:
I can confirm that you are right. For some reason, re-subscribing to methods seems to hang. From my investigation so far, it isn't due to the other issue that you linked.

timtay-microsoft:
Looks like we just don't release a certain semaphore after grabbing it in the AMQP layer's logic for unsubscribing from methods. We'll get the fix for this in shortly

vipeller:
sounds good, thank you

timtay-microsoft:
The fix for this issue has been released as of Microsoft.Azure.Devices.Client 1.36.0, so I'm closing this issue.

PR #1831, opened 2021-03-12
fix(iot-dev): Fix AMQP layer not releasing a semaphore for method links after unsubscribing from methods

timtay-microsoft:
This semaphore was just never released in this method

#1822 for additional context


iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -619,6 +619,7 @@ public async Task DisableMethodsAsync(TimeSpan timeout)
             finally
             {
                 Logging.Exit(this, timeout, nameof(DisableMethodsAsync));
+                _methodLinkSemaphore.Release();
```
timtay-microsoft, line 622: You'll want to expand the view to see the whole method here. At line 593, this semaphore is grabbed. After that, the try block does some logic that may or may not throw an exception, so this finally block is the perfect place to release the semaphore

abhipsaMisra (approved):

PR #1835, opened 2021-03-18
refactor(service-client): Pass the correct amqp audience based on aut…

vinagesh:
…h type

## Checklist

### DIFF
--- a/iothub/service/src/IotHubConnection.cs
+++ b/iothub/service/src/IotHubConnection.cs
@@ -36,8 +36,6 @@ namespace Microsoft.Azure.Devices
         internal static readonly TimeSpan DefaultOperationTimeout = TimeSpan.FromMinutes(1);
         internal static readonly TimeSpan DefaultOpenTimeout = TimeSpan.FromMinutes(1);
 
-        private readonly AccessRights _accessRights;
-
         private readonly bool _useWebSocketOnly;
         private readonly ServiceClientTransportSettings _transportSettings;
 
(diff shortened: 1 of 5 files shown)


iothub/service/src/IotHubConnection.cs
```diff
@@ -553,7 +553,7 @@ private async Task SendCbsTokenAsync(AmqpCbsLink cbsLink, TimeSpan timeout)
                     Credential.AmqpEndpoint,
                     audience,
                     resource,
-                    AccessRightsHelper.AccessRightsToStringArray(_accessRights),
+                    Credential.AmqpAudience,
```
vinagesh, line 556: The IoT hub service ignores this value for AAD auth. But for the sake of correctness, we should not be passing audience related to SAS tokens during AAD auth.

timtay-microsoft (approved):

iothub/service/src/IotHubConnectionProperties.cs
```diff
@@ -35,6 +35,12 @@ protected IotHubConnectionProperties(string hostName)
             IotHubName = GetIotHubName(hostName);
             AmqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, HostName, AmqpConstants.DefaultSecurePort).Uri;
             HttpsEndpoint = new UriBuilder(HttpsEndpointPrefix, HostName).Uri;
+#if NET451
+            AmqpAudience = new string[] { };
+
```
drwill-ms, line 40: extra newline here

drwill-ms (approved):
Are you j

iothub/service/src/IotHubConnectionProperties.cs
vinagesh: I ended up making this code better by using lists :) I'm not a fan of these #if statements.

PR #1836, opened 2021-03-18
Add support for .NET 5 for hub and provisioning clients

abhipsaMisra:
I've grouped related files into separate commits, which I hope will make reviewing this PR easier. It wasn't possible for me to break them into separate PRs, since the build errors and warnings are introduced as soon as we enable .NET 5.0 support.

The significant updates required were:
- update our pipeline build agents to use VS 2019.
- update our test certs to use sha256 signed certificates. This is because openssl dropped support for sha1 signed certificates in ubuntu 20.04.
- .NET 5.0 added additional support for passing cancellation token to a bunch of `System.Net.Http` APIs.
- A couple of test needed to be refactored so that we could dispose `IDisposable` types cleanly.

Edit: Relevant commits are commit 1, 2 and 4. Commit 3 has modifications to our test code, and can be skipped if needed.


common/src/TaskCompletionSource.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    internal sealed class TaskCompletionSource : TaskCompletionSource<int>
```
abhipsaMisra, line 8: TaskCompletionSource is natively supported in NET 5.0. Our sdk used a Dotnetty's implementation for lower versions of the framework, which we have now brought into our sdk directly. The implementation is pretty straight-forward.

e2e/test/Helpers/TestDevice.cs
```diff
@@ -21,11 +22,13 @@ public enum ConnectionStringAuthScope
         Device
     }
 
-    public class TestDevice
+    public class TestDevice : IDisposable
```
abhipsaMisra, line 25: TestDevice has IDisposable types, so I updated it to be disposable as well

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -186,13 +187,15 @@ $rootCACert = New-SelfSignedCertificate `
     -DnsName "$rootCommonName" `
     -KeyUsage CertSign `
     -TextExtension @("2.5.29.19={text}ca=TRUE&pathlength=12") `
+    -HashAlgorithm "$certificateHashAlgorithm" `
```
abhipsaMisra, line 190: ubuntu 20.04 requires all certs to be sha256 signed

e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
```diff
@@ -229,7 +229,7 @@ resource iotHub 'Microsoft.Devices/IotHubs@2020-01-01' = {
     eventHubEndpoints: {
       events: {
         retentionTimeInDays: 1
-        partitionCount: 10
+        partitionCount: 4
```
abhipsaMisra, line 232: I don't recall the exact error, but the build script did not like the event hub partition count being 10

e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
```diff
@@ -260,7 +260,7 @@ resource iotHub 'Microsoft.Devices/IotHubs@2020-01-01' = {
   sku: {
     name: 'S1'
     tier: 'Standard'
-    capacity: 1
+    capacity: 3
```
abhipsaMisra, line 263: our test suite requires the hub instance to have >1 unit, else our tests end up getting throttled or cancelled. We use 3 units for our canary instance so I used the same here.

e2e/test/prerequisites/E2ETestsSetup/test-resources.json
```diff
@@ -498,5 +498,12 @@
       "type": "string",
       "value": "[reference(resourceId('Microsoft.Insights/components', parameters('ApplicationInsightsName')), '2015-05-01').InstrumentationKey]"
     }
+  },
+  "metadata": {
```
abhipsaMisra, line 502: autogenerated via bicep build

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
```diff
@@ -82,14 +82,10 @@ private void SetDeviceId(string deviceId)
         }
 
         /// <summary>
-        /// Dispose the X509 certificate associated with this authentication method.
+        /// The <see cref="Certificate"/> managed resource should be disposed by the user.
         /// </summary>
         public void Dispose()
         {
```
abhipsaMisra, line 88: This was an oversight, we shouldn't really be disposing the supplied certificates, since the calling application might want to reuse it elsewhere.

iothub/device/src/Transport/AmqpIoT/AmqpIoTConnector.cs
```diff
@@ -49,14 +49,9 @@ public async Task<AmqpIoTConnection> OpenConnectionAsync(TimeSpan timeout)
                 MaxFrameSize = AmqpConstants.DefaultMaxFrameSize,
                 ContainerId = CommonResources.GetNewStringGuid(),
                 HostName = _hostName,
+                IdleTimeOut = Convert.ToUInt32(_amqpTransportSettings.IdleTimeout.TotalMilliseconds),
             };
 
-            TimeSpan idleTimeout = _amqpTransportSettings.IdleTimeout;
-            if (idleTimeout != null)
```
abhipsaMisra, line 55: TimeSpan cannot be null

common/src/HttpContentExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    internal static class HttpContentExtensions
```
drwill-ms, line 11: meaningful doc comment for future code readers, please

common/src/HttpContentExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        internal static async Task CopyToStreamAsync(this HttpContent content, Stream stream, CancellationToken cancellationToken)
+        {
+#if NET5_0
```
drwill-ms, line 15: Consider:

```csharp
await content.CopyToAsync(
    stream
#if NET5_0
    , cancellationToken
#endif
    ).ConfigureAwait(false);
```

And then suppress the unused cancellation token.

common/src/HttpContentExtensions.cs
abhipsaMisra: sure, i'll add

common/src/HttpContentExtensions.cs
drwill-ms: Looking over the rest of the stuff, I think what you have is fine.

common/src/HttpContentExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+            await content.CopyToAsync(stream, cancellationToken).ConfigureAwait(false);
+#else
+            _ = cancellationToken;
```
drwill-ms, line 18: I wouldn't skimp on code comments here. Just even saying .NET 5.0 includes a cancellation token parameter, which we'd prefer to make use of. For the other framework versions, we'll just need to ignore the parameter.

common/src/HttpMessageHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+#if NET451
+            requestMessage.Content = new ObjectContent<T>(entity, s_jsonFormatter);
+#else
+            string str = JsonConvert.SerializeObject(entity);
+            requestMessage.Content = new StringContent(str, Encoding.UTF8, ApplicationJson);
+#endif
+        }
+
+        internal static async Task<T> ReadHttpResponseMessageContentAsync<T>(HttpResponseMessage message, CancellationToken token)
```
drwill-ms, line 36: Elsewhere spelled out as `cancellationToken` but here just `token`. Use the full one for consistency?

common/src/TaskCompletionSource.cs
drwill-ms: Sounds like a worthy doc comment.

common/src/TaskCompletionSource.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public TaskCompletionSource()
+        {
+        }
+
+        public bool TrySetResult() => TrySetResult(0);
+
+        public void SetResult() => SetResult(0);
+
+        public override string ToString() => "TaskCompletionSource[status: " + Task.Status.ToString() + "]";
```
drwill-ms, line 25: `$"TaskCompletionsource[status:{Task.Status}]"

common/src/TaskCompletionSource.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        static TaskCompletionSource CreateVoidTcs()
```
drwill-ms, line 27: Do all callers dispose this return value?

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
drwill-ms: How about a code comment then?

e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
drwill-ms: We had it at 1 for devs running tests against their own instance, and not incurring too much cost. As we do not deploy new pipeline resources that often, the idea was that you'd have to go manually change it on the pipeline's instance.

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -530,14 +533,23 @@ public async Task ProvisioningDeviceClient_ValidRegistrationId_MqttWsWithProxy_S
...
             await ConfirmExpectedDeviceCapabilitiesAsync(result, auth, deviceCapabilities).ConfigureAwait(false);
 
             if (attestationType != AttestationMechanismType.X509) //x509 enrollments are hardcoded, should never be deleted
             {
-                await DeleteCreatedEnrollmentAsync(enrollmentType, CreateProvisioningService(proxyServerAddress), security, groupId).ConfigureAwait(false);
+                using ProvisioningServiceClient dpsServiceClient = CreateProvisioningService(proxyServerAddress);
+                await DeleteCreatedEnrollmentAsync(enrollmentType, dpsServiceClient, security, groupId).ConfigureAwait(false);
+            }
+
+            if (auth is IDisposable disposableAuth)
```
drwill-ms, line 550: I prefer this approach over the one several files back that used #if !NET451 around it.

iothub/device/src/Transport/AmqpIoT/AmqpIoTConnector.cs
drwill-ms: There is `TimeSpan.Zero` though.

drwill-ms (approved):

common/src/HttpContentExtensions.cs
abhipsaMisra: sure

common/src/HttpMessageHelper.cs
abhipsaMisra: sure, looks like I missed it here

common/src/TaskCompletionSource.cs
abhipsaMisra: this was an unreferenced private method that added no value to our client, so I've removed it.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
abhipsaMisra: sure

e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
abhipsaMisra: I see, I'll add a flag to make this configurable and add a comment explaining this.

iothub/device/src/Transport/AmqpIoT/AmqpIoTConnector.cs
abhipsaMisra: True, but that would mean that the user deliberately set this value; which even though is weird, we'd allow.

e2e/test/provisioning/ProvisioningE2ETests.cs
abhipsaMisra: I like this approach too, however disposing X509Certificates gets a bit messy since NET451 doesn't support that.

jamdavi (approved):

e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
vinagesh: As discussed offline, you can make this an ARM template parameter with a default of 1.

e2e/test/provisioning/ProvisioningE2ETests.cs
drwill-ms: either it is IDisposable or it isn't, right?

abhipsaMisra: For an IAuthneticationMethod yes, it is or isn't IDisposable. 
Did I leave an IAuthenticationMethod dispose with #ifNET451 directive? If yes, then yeah, that isn't necessary. Do you happen to recall where that was, I couldn't find it within the PR.

drwill-ms: It was some other disposable type. Let me find it.

abhipsaMisra: I think I see what you mean, the X509Certificate wouldn't be disposable to begin with in NET451. Yeah, good point.

common/src/HttpMessageHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+
+#if NET451
+using System.Net.Http.Formatting;
+#endif
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    internal static class HttpMessageHelper
+    {
+#if !NET451
```
vinagesh, line 19: #if NET451
        private static readonly JsonMediaTypeFormatter s_jsonFormatter = new JsonMediaTypeFormatter();
#else
        private const string ApplicationJson = "application/json";
#endif

vinagesh: If we want to keep the order of const and static, we can switch the if-else. It just seemed too much to process when it was #if NET451 and then right after #if !NET451

common/src/service/HttpClientHelper.cs
```diff
@@ -289,12 +269,8 @@ internal sealed class HttpClientHelper : IHttpClientHelper
                 {
                     // TODO: skintali: Use string etag when service side changes are ready
```
bikamani, line 270: Is this still a todo? 
May be not in scope of your PR but wondering since requestMsg is used below

common/src/service/HttpClientHelper.cs
```diff
@@ -364,9 +332,12 @@ private static async Task<T> ReadResponseMessageAsync<T>(HttpResponseMessage mes
 #if NET451
             T entity = await message.Content.ReadAsAsync<T>(token).ConfigureAwait(false);
 #else
-            string str = await message.Content.ReadAsStringAsync().ConfigureAwait(false);
+            string str = await message.Content.ReadHttpContentAsStringAsync(token).ConfigureAwait(false);
             T entity = JsonConvert.DeserializeObject<T>(str);
 #endif
+
+            //T entity = await ReadHttpResponseMessageContentAsync<T>(message, token).ConfigureAwait(false);
```
bikamani, line 339: Is this supposed to be here?

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
vinagesh: Without having this, how can a customer do something like `using new DeviceAuthenticationWithX509Certificate()`
The dispose method is not being called from any of our clients from what I know.

common/src/HttpMessageHelper.cs
abhipsaMisra: sure, I can switch it

abhipsaMisra: the way that you have written reads better to me (positive case then negative case); we can probably switch the order of const vs readonly for the sake of readability for this one instance.
@drwill-ms thought?

common/src/service/HttpClientHelper.cs
abhipsaMisra: I am not sure what this TODO even refers to, maybe something to take up outside the scope of this PR.

common/src/service/HttpClientHelper.cs
abhipsaMisra: you're right, it isn't.

common/src/HttpMessageHelper.cs
drwill-ms: agreed, positive case reads better to me

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
abhipsaMisra: yes, but what about the scenario where an application reuses the x509 certificate instance for creating multiple objects, one of them being a `DeviceAuthenticationWithX509Certificate` instance. Then, if we dispose the supplied certificate when this auth instance is disposed, the other objects that had a reference to that certificate would be rendered unusable.

abhipsaMisra: The way I see it, the user is the owner of the IDisposable type (the x509certificate), so they should decide when it is ready to be disposed.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -858,7 +863,7 @@ private async Task SendMessageAsync(IChannelHandlerContext context, Message mess
             }
 
             PublishPacket packet = await ComposePublishPacketAsync(context, message, qos, topicName).ConfigureAwait(true);
-            var publishCompletion = new TaskCompletionSource();
+            var publishCompletion = new DotNetty.Common.Concurrency.TaskCompletionSource();
```
vinagesh, line 866: Should we use the TaskCompletionSource we have above or DotNetty one? If we need to use DotNetty one here, can you add a code comment for it?

iothub/device/src/DeviceAuthenticationWithX509Certificate.cs
vinagesh: Going by the usage below, it makes sense to not implement the dispose here so that customer can reuse the cert directly in some other place. This also means there is no reason for this class to be IDisposable. However, I guess it is a breaking change so we will just have to stick with a blank implementation. It is worth adding these comments here to the dispose method. At the moment noone is calling this dispose, so it should be ok and unless the customer does using new DeviceAuthenticationWithX509Certificate(), their certificate will not be disposed.

abhipsaMisra: sure thing, I'll add a comment explaining why we are not disposing the certificate in here.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
abhipsaMisra: I'll add a comment for this, sure

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
```diff
@@ -100,10 +114,10 @@ public void GetRetryFromRejectedReturnsNullIfNoError()
         public void GetRetryAfterFromApplicationPropertiesSuccess()
         {
             int expectedRetryAfter = 42;
-            AmqpMessage amqpResponse = AmqpMessage.Create();
+            using var amqpResponse = AmqpMessage.Create();
```
vinagesh, line 117: The type is good to have here. We are not doing a new so Create could be returning any other type as well.

common/src/TaskCompletionSource.cs
```diff
@@ -5,30 +5,22 @@
 
 namespace Microsoft.Azure.Devices.Shared
 {
-    internal sealed class TaskCompletionSource : TaskCompletionSource<int>
+    /// <summary>
+    /// A <see cref="TaskCompletionSource{boolean}"/> implementation that returns a <see cref="Task"/> when completed.
+    /// </summary>
+    /// <remarks>
+    /// This is used for .NET implementations less than .NET 5.0, which lack a native implementation of the non-generic TaskCompletionSource.
```
vinagesh, line 12: It would be great to add the remarks from the TaskCompletionSource here to explain what it really is. The summary added does not really indicate what this class really represents.

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcompletionsource?view=net-5.0

common/src/TaskCompletionSource.cs
```diff
@@ -5,30 +5,22 @@
...
-            : base(state)
-        {
-        }
-
         public TaskCompletionSource()
         {
         }
 
-        public bool TrySetResult() => TrySetResult(0);
+        public bool TrySetResult() => TrySetResult(true);
```
vinagesh, line 20: Should we set this to true or false? Looks like it was 0 (false) previeously.

provisioning/transport/amqp/tests/ProvisioningErrorDetailsAmqpTests.cs
abhipsaMisra: This was intellisense suggested, but I can change it back, sure

common/src/TaskCompletionSource.cs
abhipsaMisra: I thought it was self-explanatory, since all this class does is implement a non-generic version on TCS by setting the type as a bool. I can add the existing description for TCS into the remark, sure.

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -388,7 +406,9 @@ az deployment group create `
     StorageAccountName=$storageAccountName `
     KeyVaultName=$keyVaultName `
     DpsCustomAllocatorRunCsxContent=$dpsCustomAllocatorRunCsxContent `
-    DpsCustomAllocatorProjContent=$dpsCustomAllocatorProjContent
+    DpsCustomAllocatorProjContent=$dpsCustomAllocatorProjContent `
+    HubUnitsCount=$iothubUnitsToBeCreated
```
vinagesh, line 410: Can you update [1N ](https://microsoft.sharepoint.com/teams/IoTDevelopersDevices/_layouts/OneNote.aspx?id=%2Fteams%2FIoTDevelopersDevices%2FShared%20Documents%2FAzure%20IoT%20Managed%20SDK%20Team%2FManaged%20SDK%20team%201N&wd=target%28Testing.one%7CA29E144A-05CF-4713-A987-F36EF7EB324D%2FIotHub%20Tests%7CAC522BC7-5108-4127-8934-3DAE6B34D3CB%2F%29
onenote:https://microsoft.sharepoint.com/teams/IoTDevelopersDevices/Shared%20Documents/Azure%20IoT%20Managed%20SDK%20Team/Managed%20SDK%20team%201N/Testing.one#IotHub%20Tests&section-id={A29E144A-05CF-4713-A987-F36EF7EB324D}&page-id={AC522BC7-5108-4127-8934-3DAE6B34D3CB}&end)to explain this new param?

common/src/TaskCompletionSource.cs
abhipsaMisra: It doesn't really matter, all it does it mark the task as completed. We could have also defined it as an object and set it to null.

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
```diff
@@ -96,9 +97,9 @@ public async Task ProvisioningDeviceClient_RegisterAsyncInvalidServiceCertificat
 
         private static async Task TestInvalidServiceCertificate(ProvisioningTransportHandler transport)
         {
-            using var security =
-                new SecurityProviderX509Certificate(Configuration.Provisioning.GetIndividualEnrollmentCertificate());
-            ProvisioningDeviceClient provisioningDeviceClient = ProvisioningDeviceClient.Create(
+            using X509Certificate2 cert = Configuration.Provisioning.GetIndividualEnrollmentCertificate();
+            using var security = new SecurityProviderX509Certificate(cert);
+            var provisioningDeviceClient = ProvisioningDeviceClient.Create(
```
vinagesh, line 102: Type would make sense anywhere we don't directly use the 'new' keyword.

vinagesh (approved):

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
abhipsaMisra: I think what intellisense infers is:
`objectA.Create()` -> returned type is `objectA` so using `var` is ok
`objectA.CreateSomething()` -> returned type is not obvious, so don't use `var`.

I can update it to be more explicit though, sure.

Issue #1818, opened 2021-03-09
[feature request] Add option to set SAS token expiration for the ModuleClient.CreateFromEnvironmentVariable() flow.


maskimthedog:
In the Azure IoT Hub C SDK, the method below is available to set the default SAS token lifetime.  No such method is present in the Azure IoT Hub C# SDK.  I have to patch the code to set the hardcoded 60 minute expiration to something more tolerable data usage wise for cellular networks.  Please consider adding such a method.

```
size_t sas_token_lifetime = SAS_TOKEN_LIFETIME_SECONDS;
if (IOTHUB_CLIENT_OK != IoTHubClient_LL_SetOption(g_device->iotHubClientHandle, OPTION_SAS_TOKEN_LIFETIME, &sas_token_lifetime))
{​​​​​​​
    msg_error("failure to set option \"sas_token_lifetime\"\r\n");
}​​​​​​​
```

abhipsaMisra:
Rather than configure the sas token's lifespan on an option on the client, we have the following authentication abstract class that you can implement, where you will be able to set custom values for the sas token's time to live and also the buffer percentage after which a token should be renewed: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs
Would this work for you?

maskimthedog:
I totally understand that and I know how to use it.  However, your API method for establishing a connection to IoT Hub uses the constructor with the default TTL and percentage.  There is no way to change this without modifying the SDK unless MS provides a way to override so that the SAS token does not expire after the default 60 minutes and invalidates the connection to where it must be re-established.

abhipsaMisra:
I think I might be misunderstanding your scenario. Could you help me understand it so that I can assist you better.

>  your API method for establishing a connection to IoT Hub uses the constructor with the default TTL and percentage

My understanding is that you would like to initialize a device client that has its sas tokens expire at a different frequency than the one hardcoded within the client (i.e. 1 hour). By default, if you initialize the device client using connection string, it will generate sas tokens that are valid for an hour. To override this behavior, you can implement the abstract class `DeviceAuthenticationWithTokenRefresh` and pass it to the device client `Create` method for initialization.

```csharp
internal class DeviceAuthenticationSasToken : DeviceAuthenticationWithTokenRefresh
{
    private readonly IotHubConnectionStringBuilder _connectionStringBuilder;

[...]

PR #1837, opened 2021-03-18
Add a client option to specify the sas token ttl and renewal buffer

abhipsaMisra:
Currently the .NET sdk requires you to implement `AuthenticationWithTokenRefresh` in order to be able to configure the time to live and renewal buffer value for generated sas tokens. This implementation does require some amount of customer code to be written. It is also currently not supported for edge modules created from environment.

This PR introduces an option for users to set a sas token time to live and renewal buffer value into the client options, which would be honored for the `CreateFromConnectionString` and `CreateFromEnvironment` flows.

Fix for #1818


iothub/device/src/ClientOptions.cs
```diff
@@ -29,5 +30,27 @@ public class ClientOptions
...
         /// </summary>
         public SdkAssignsMessageId SdkAssignsMessageId { get; set; } = SdkAssignsMessageId.Never;
+
+        /// <summary>
+        /// The suggested time to live value for tokens generated for SAS authenticated clients.
+        /// If unset the generated SAS tokens will be valid for 1 hour.
+        /// </summary>
+        /// <remarks>
+        /// This is used only for SAS token authenticated clients through either the
+        /// <see cref="DeviceClient.CreateFromConnectionString(string, ClientOptions)"/>/ <see cref="ModuleClient.CreateFromConnectionString(string, ClientOptions)"/> flow
```
drwill-ms, line 40: looks to be a stray / in there. Is that intended?

iothub/device/src/ClientOptions.cs
```diff
@@ -29,5 +30,27 @@ public class ClientOptions
...
+        /// or the <see cref="ModuleClient.CreateFromEnvironmentAsync(ClientOptions)"/> flow.
+        /// </remarks>
+        public TimeSpan SasTokenTimeToLive { get; set; }
+
+        /// <summary>
+        /// The time buffer before expiry when the token should be renewed, expressed as a percentage of the time to live.
```
drwill-ms, line 46: When I see percentage, I think it could either be 0.15 or 15. I see the type is an int, so I guess it is the latter. Still, it would be good to show an example (e.g. a value of 85 for 85%)

iothub/device/src/ClientOptions.cs
```diff
@@ -29,5 +30,27 @@ public class ClientOptions
...
+        /// If unset the token will be renewed when it has 15% or less of its lifespan left.
+        /// </summary>
+        /// <remarks>
+        /// This is used only for SAS token authenticated clients through either the
+        /// <see cref="DeviceClient.CreateFromConnectionString(string, ClientOptions)"/>/ <see cref="ModuleClient.CreateFromConnectionString(string, ClientOptions)"/> flow
+        /// or the <see cref="ModuleClient.CreateFromEnvironmentAsync(ClientOptions)"/> flow.
+        /// </remarks>
+        public int SasTokenRenewalBuffer { get; set; }
```
drwill-ms, line 54: An int can be way bigger than 100 and way smaller than 1. Are we going to do any value checking?

Same question for TTL.

iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs
```diff
@@ -63,5 +57,19 @@ public override IotHubConnectionStringBuilder Populate(IotHubConnectionStringBui
             iotHubConnectionStringBuilder.DeviceId = DeviceId;
             return iotHubConnectionStringBuilder;
         }
+
+        private static int SetSasTokenSuggestedTimeToLiveSeconds(int suggestedTimeToLiveSeconds)
+        {
+            return suggestedTimeToLiveSeconds == 0
```
drwill-ms, line 63: int can be negative

iothub/device/src/IotHubConnectionStringBuilder.cs
```diff
@@ -187,6 +187,10 @@ public IAuthenticationMethod AuthenticationMethod
         /// Full chain of certificates from the one used to sign the device certificate to the one uploaded to the service.
         internal X509Certificate2Collection ChainCertificates { get; set; }
 
+        internal TimeSpan SasTokenTimeToLive { get; set; }
```
drwill-ms, line 190: need doc comments

iothub/device/src/ClientOptions.cs
abhipsaMisra: Yes, I intend it to be displayed in the below format "option1/ option2", since I was grouping DeviceClient.CreateFromConnectionString and ModuleClient.CreateFromConnectionString together.
Do you think there is a better or clearer way to express this?

iothub/device/src/ClientOptions.cs
abhipsaMisra: Sure thing, I'll add an example.

iothub/device/src/ClientOptions.cs
abhipsaMisra: We do have these validations in place in the base class where these values are used: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/AuthenticationWithTokenRefresh.cs#L36-L44

iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs
abhipsaMisra: The validation is a part of the base class: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/AuthenticationWithTokenRefresh.cs#L36-L44

iothub/device/src/IotHubConnectionStringBuilder.cs
abhipsaMisra: sure thing

iothub/device/src/ClientOptions.cs
drwill-ms: If you wish to separate things with a slash, it should be consistent for spacing before and after.

However, you could just use the word "or" instead which I'd recommend.

iothub/device/src/ClientOptions.cs
drwill-ms: Yeah, but well after it is set. Shouldn't we give the user feedback immediately that the value is invalid?

abhipsaMisra: My idea was to keep the validation in a single location; i.e. you could set the token ttl through multiple locations, but they'd all get set to AuthenticationWithTokenRefresh finally, and that is where the validation would happen. This would help minimize chances of us implementing different validation rules in different places.
Since these values are set as a part of the client initialization process, I thought it was fine to delay it until it reaches AuthenticationWithTokenRefresh

drwill-ms: I can appreciate that, but how does a user get feedback about the validity of their values?

abhipsaMisra: They'll get an ArgumentOutOfRangeException when they create the client instance. I've added some unit tests around this.

e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
```diff
@@ -109,6 +109,63 @@ public async Task DeviceClient_TokenConnectionDoubleRelease_Ok()
             } // Second release
         }
 
+        // The easiest way to test that sas tokens expire with custom expiration time via the CreateFromConnectionString flow is 
+        // by initializing a DeviceClient instance over Mqtt (since sas token expiration over Mqtt is accompanied by a disconnect).
```
abhipsaMisra, line 113: This does have the possibility of giving us false positives, as in, a network issue could also give us a disconnect. Since the token refresh logic is encapsulated entirely within the sdk, we cannot monitor it from outside.
The translation of client options to the token refresher also happens over multiple layers, so this wasn't a good candidate for unit tests either.

Let me know if you have any suggestions for testing this in a different way.

iothub/device/src/ClientOptions.cs
drwill-ms: That's not terrible. Would you at least put acceptable ranges in the doc comments for those properties?

abhipsaMisra: Sure thing.

e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
jamdavi: Is this something we can handle and track in long haul?

abhipsaMisra: We could set a custom token expiration time and monitor if it disconnects as per the schedule over an extended period of time. The core token refresh logic is tested separately within our test suite, what we need to test here is to ensure that the client options are being set into the token refresher class.
This does fit into a unit test better than an E2E test, it is just the way these options are translated in our client that doesn't make it straightforward to test. Nevertheless, that's where I am working on adding a test now.

iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs
```diff
@@ -11,32 +11,26 @@ namespace Microsoft.Azure.Devices.Client
...
-        /// Initializes a new instance of the <see cref="DeviceAuthenticationWithTokenRefresh"/> class using default
-        /// TTL and TTL buffer time settings.
-        /// </summary>
-        /// <param name="deviceId">Device Identifier.</param>
-        public DeviceAuthenticationWithTokenRefresh(string deviceId)
-            : this(deviceId, DefaultTimeToLiveSeconds, DefaultBufferPercentage)
-        {
-        }
+        internal const int DefaultTimeToLiveSeconds = 1 * 60 * 60;
+        internal const int DefaultBufferPercentage = 15;
```
abhipsaMisra, line 15: I updated these variables to be internal since I need them for a unit test and I did not want to copy over the values manually.

iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs
```diff
@@ -28,15 +28,19 @@ public DeviceAuthenticationWithTokenRefresh(string deviceId)
...
+        /// <param name="timeBufferPercentage">
+        /// The time buffer before expiry when the token should be renewed, expressed as a percentage of the time to live.
+        /// The default behavior is that the token will be renewed when it has 15% or less of its lifespan left.
+        ///</param>
         public DeviceAuthenticationWithTokenRefresh(
             string deviceId,
             int suggestedTimeToLiveSeconds,
             int timeBufferPercentage)
-            : base(suggestedTimeToLiveSeconds, timeBufferPercentage)
+            : base(SetSasTokenSuggestedTimeToLiveSeconds(suggestedTimeToLiveSeconds), SetSasTokenRenewalBufferPercentage(timeBufferPercentage))
```
abhipsaMisra, line 43: I had previously merged the two constructors by marking the ttl and renewal buffer as optional, but I reverted that change so as to not break existing customers who might copy over the updated dll and want to run their pre-compiled exe (break binary compatibility). Thanks for informing me about this @drwill-ms !

In order to be binary compatible, it looks like we should we always be adding new APIs with optional params, rather than appending the optional param directly to an existing API.

iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs
```diff
@@ -63,5 +67,19 @@ public override IotHubConnectionStringBuilder Populate(IotHubConnectionStringBui
             iotHubConnectionStringBuilder.DeviceId = DeviceId;
             return iotHubConnectionStringBuilder;
         }
+
+        private static int SetSasTokenSuggestedTimeToLiveSeconds(int suggestedTimeToLiveSeconds)
+        {
+            return suggestedTimeToLiveSeconds == 0
```
drwill-ms, line 73: `<=`?

iothub/device/src/GlobalSuppressions.cs
```diff
@@ -11,3 +11,4 @@
 [assembly: SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "SDK hides non-actionable errors from user", Scope = "module")]
 [assembly: SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "Public API cannot be changed.", Scope = "type", Target = "~T:Microsoft.Azure.Devices.Client.ConnectionStatusChangeReason")]
```
drwill-ms, line 12: We should probably keep code-specific suppressions next to where the code is, and save this file for module-wide suppressions.

iothub/device/src/GlobalSuppressions.cs
```diff
@@ -11,3 +11,4 @@
...
 [assembly: SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "Public API cannot be changed.", Scope = "type", Target = "~T:Microsoft.Azure.Devices.Client.ConnectionStatus")]
+[assembly: SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "The authentication method disposed when the client is disposed.", Scope = "member", Target = "~M:Microsoft.Azure.Devices.Client.Edge.EdgeModuleClientFactory.CreateInternalClientFromEnvironmentAsync~System.Threading.Tasks.Task{Microsoft.Azure.Devices.Client.ModuleClient}")]
```
drwill-ms, line 14: I don't understand the phrase "the authentication method disposed". How is a method disposed?

iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs
drwill-ms: Better not to assume, tho. Simply make it <= to be safe.

iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs
abhipsaMisra: The `<` scenario will get validated in the base class `AuthenticationWithTokenRefresh` and will generate an `ArgumentOutOfRangeException`.

iothub/device/src/GlobalSuppressions.cs
abhipsaMisra: sure

iothub/device/src/GlobalSuppressions.cs
abhipsaMisra: I've updated the text to say: IDisposable ModuleAuthenticationWithHsm is disposed when the client is disposed. (the "authentication method" here was referencing the IAuthenticationMethod type)

drwill-ms (approved):

e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
```diff
@@ -109,6 +109,63 @@ public async Task DeviceClient_TokenConnectionDoubleRelease_Ok()
...
+        [LoggedTestMethod]
+        [TestCategory("LongRunning")]
+        public async Task DeviceClient_CreateFromConnectionString_TokenIsRefreshed_Mqtt()
+        {
+            var sasTokenTimeToLive = TimeSpan.FromSeconds(10);
+            int sasTokenRenewalBuffer = 50;
+            var deviceDisconnected = new SemaphoreSlim(0);
```
drwill-ms, line 120: using?

abhipsaMisra: yes, good catch.

maskimthedog:
Thanks all for expediting this request.

PR #1852, opened 2021-03-23
Update tests to be resilient to CRUD throttling errors

abhipsaMisra:


e2e/test/Helpers/TestDevice.cs
```diff
@@ -82,14 +84,18 @@ private static async Task<TestDevice> CreateDeviceAsync(TestDeviceType type, str
...
-            Device device = await rm.AddDeviceAsync(requestDevice).ConfigureAwait(false);
-
-            _logger.Trace($"{nameof(GetTestDeviceAsync)}: Pausing for {DelayAfterDeviceCreationSeconds}s after device was created.");
-            await Task.Delay(DelayAfterDeviceCreationSeconds * 1000).ConfigureAwait(false);
+            Device device = null;
+            await RetryOnThrottlingAsync(
+                async () =>
+                {
+                    device = await rm.AddDeviceAsync(requestDevice).ConfigureAwait(false);
+                });
```
vinagesh, line 92: .ConfigureAwait(false)

drwill-ms (approved):

e2e/test/Helpers/TestDevice.cs
```diff
@@ -170,5 +176,25 @@ public async Task RemoveDeviceAsync()
             using var rm = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
             await rm.RemoveDeviceAsync(Id).ConfigureAwait(false);
         }
+
+        private static async Task RetryOnThrottlingAsync(Func<Task> asyncOperation)
```
vinagesh, line 180: This could be a helper method available to all tests.. Would be useful whenever we want to retry Throttling exceptions.

e2e/test/Helpers/TestDevice.cs
```diff
@@ -170,5 +176,25 @@ public async Task RemoveDeviceAsync()
...
+        {
+            int counter = 0;
+            do
+            {
+                try
+                {
+                    await asyncOperation().ConfigureAwait(false);
```
vinagesh, line 187: Do you know if there is information in the response on when to retry? Usually all services should send that value back to us in the response. Our tests would be less flaky if we use that. If that information is not there, consider using an exponential back-off. The chances of success is low if we have a linear back-off.

jamdavi (approved):
Looks good, agree with Sindhu about helper method.

e2e/test/Helpers/TestDevice.cs
abhipsaMisra: I am not sure what you mean here

e2e/test/Helpers/TestDevice.cs
abhipsaMisra: I agree, however, the reason I added it in here directly is because this is the only class in our test suite in which devices are created (outside of some specific jobs tests maybe). Provisioning clients can also create device identities but they are designed to handle throttling implicitly.

e2e/test/Helpers/TestDevice.cs
abhipsaMisra: Unfortunately service doesn't give us a retry-after value, but rather indicates in the message directly that we should retry after 10 secs: Microsoft.Azure.Devices.Common.Exceptions.ThrottlingException: {"Message":"ErrorCode:ThrottlingBacklogTimeout;The request has been throttled. Wait 10 seconds and try again. Operation type: CRUD","ExceptionMessage":"Tracking ID:04c922b3aac44715a7569eef9e513463-G:4-TimeStamp:03/23/2021 17:00:30"}

Good point about introducing an exponential backoff, I'll add that in.

e2e/test/Helpers/TestDevice.cs
vinagesh: await RetryOnThrottlingAsync(....)..ConfigureAwait(false)
I meant .ConfigureAwait(false) is missing

vinagesh (approved):

e2e/test/Helpers/TestDevice.cs
vinagesh: Sure but when we need it again for something, we'll rewrite it. This implementation seems generic enough to reuse when needed (in case we find that we need to improve another test)

e2e/test/Helpers/TestDevice.cs
abhipsaMisra: Ah, got it, I thought you were talking about the r..AddDevicesAsync API!
Will add.

e2e/test/Helpers/TestDevice.cs
abhipsaMisra: I can add it to a helper class, no problem!

PR #1857, opened 2021-03-23
Update test resources creation script to generate sha256 certificates

abhipsaMisra:
These changes are a part of `pewviews/net5`, however, since it might be a while before they're merged into `mater`, it is worth getting these changes in specifically.

### DIFF
--- a/e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
+++ b/e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
@@ -1,7 +1,7 @@
 param ApplicationInsightsName string {
   default: '${resourceGroup().name}-ai'
   metadata: {
-      description: 'The name of application insights.'      
+      description: 'The name of application insights.'
   }
 }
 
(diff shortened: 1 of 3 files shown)


e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
```diff
@@ -85,6 +92,13 @@ param OperationalInsightsName string {
   }
 }
 
+param OperationInsightsLocation string {
+  default: 'eastus'
```
drwill-ms, line 96: What other region options do we have? Anything closer to the west coast?

drwill-ms (approved):

e2e/test/prerequisites/E2ETestsSetup/test-resources.bicep
abhipsaMisra: We do have support for west us region, I'll update to that.

azabbasi (approved):

vinagesh (approved):

PR #1858, opened 2021-03-23
feat(doc): Add PnP API surface doc

abhipsaMisra:
Doc: https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/pnpSurface/iothub/device/devdoc/API%20Design.md

Considerations:
- payload sent for pnp-based devices (telemetry/ properties, commands) follow a particular convention and model as defined in the device's DTDL model.
- these conventions define how the data should be serialized and subsequently encoded into bytes.
- the sdk will implement a "default" convention that uses NewtonSoft.Json based serializer to serialize and deserialize the data, and uses utf-8 based encoding.
- users can pass in their custom convention implementation to override this default behavior.
- `Twin` is not being reused for property operations since `Twin` brings with itself the concept of `Desired` and `Reported` properties, whereas these terms don't really exist in pnp-world (even though the concept of "desired" and "reported" properties still exists).
- `TelemetryMessage`, `CommandRequest` and `CommandResponse` are very similar to `Message`, `MethodRequest` and `MethodResponse`, baring some small implementation details:
   - `TelemetryMessage` accepts a convention-based payload, instead of raw bytes
   - `CommandRequest` allows easy access to `CommandName` and `ComponentName`
   - `CommandResponse` allows you to send a convention-based response.


iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
+## Plug and Play convention compatible APIs
+
+### Properties
+
+```csharp
+/// <summary>
+/// Update a single property.
+/// 
+/// Property under the root component:
+/// "reported": {
```
jspaith, line 10: If this turns into user facing documentation, I'd rather not document the protocol at this layer so that app can focus on API shape and not implementation details.  ?

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+/// 
+/// Property under a specified component:
+/// "reported": {
+///     "thermostat1": {
+///         "__t": "c",
+///         "temperature": 21.3
+///     }
+/// }
+/// </summary>
+/// <param name="propertyName">The name of the twin property to update.</param>
```
jspaith, line 22: We should not use the word "twin" anywhere relating to properties, since the app developer shouldn't care.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+///       "ad": "complete"
+///     }
+///   }
+/// }
+/// </summary>
+/// <param name="propertyName">The name of the twin property to update.</param>
+/// <param name="propertyValue">The value of the twin property to update.</param>
+/// <param name="componentName">The name of the component in which the property is defined. Can be null for property defined under the root interface.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task UpdatePropertyAsync(string propertyName, WritableProperty propertyValue, string componentName = default, CancellationToken cancellationToken = default);
```
jspaith, line 81: This seems to be for responding to a writeable property update?  And it's an overload of setting a "read-only" property.  Correct??

If so, I thought we had decided to give this a different name (like RespondToWriteableProperty??) since overload had pros/cons.

If we've changed there maybe good reasons but I'd like to understand more behind the rationale.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+/// "reported": {
+///     "thermostat1": {
+///         "__t": "c",
+///         "temperature": 21.3,
+///         "humudity": 60
+///     }
+/// }
+/// </summary>
+/// <param name="properties">The twin properties and values to update.</param>
+/// <param name="componentName">The name of the component in which the properties are defined. Can be null for properties defined under the root interface.</param>
```
jspaith, line 47: Low prio -- We need to check if this is `root interface` or `root component`.  I think component maybe more correct, but want to call out for later research before I forget.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+///     "targetTemperature": 21.3,
+///     "targetHumidity": 80
+///   }
+/// },
+/// "$version" : 3
+/// 
+/// </summary>
+/// <param name="propertyActionAsTwinCollection">The action to be taken when a writeable property event is received.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public void SubscribeToWritablePropertyEvent(Action<TwinCollection> propertyActionAsTwinCollection, CancellationToken cancellationToken = default);
```
abhipsaMisra, line 107: We can add another API for managing per-component property updates (based on customer feedback)

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+/// Send a single instance of telemetry.
+/// </summary>
+/// <param name="telemetryName">The name of the telemetry, as defined in the DTDL interface.</param>
+/// <param name="telemetryValue">The telemetry payload, in the format defined in the DTDL interface.</param>
+/// <param name="componentName">The name of the component in which the telemetry is defined. Can be null for telemetry defined under the root interface.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SendTelemetryAsync(string telemetryName, dynamic telemetryValue, string componentName = default, CancellationToken cancellationToken = default);
+
+/// <summary>
+/// Send a batched instance of telemetry.
```
jspaith, line 174: Does MQTT support batched telemetry in C#?  At the protocol I know AMQP has 1st class support for it, but MQTT doesn't.

If C#/MQTT doesn't support this somehow and short of a critical need, I'd strongly recommend removing this from API surface.

I worry having an AMQP only construct when we're pushing MQTT as preferred protocol isn't clear.  FWIW C absolutely will not have a batched mode for PnP telemetry.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+```csharp
+/// <summary>
+/// Subscribe to command event.
+/// </summary>
+/// <param name="commandName">The name of the command, as defined in the DTDL interface.</param>
+/// <param name="commandCallback">The action to be taken when a command request is received.</param>
+/// <param name="userContext">Context object that is passed to the event callback.</param>
+/// <param name="componentName">The name of the component in which the command is defined. Can be null for command defined under the root interface.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SubscribeToCommandEvent(string commandName, Func<CommandRequest, object, Task<CommandResponse>> commandCallback, object userContext = default, string componentName = default, CancellationToken cancellationToken = default);
```
jspaith, line 204: Just as subscribing to writeable properties is currently only using a global catchall, I recommend that command handling follow same pattern.  For now at least - if this is needed later we can add certainly.

iothub/device/devdoc/API Design.md
jspaith, line 107: I get that at the DeviceClient layer we're going to be dealing with TwinCollection or in C-world the JSON blob thing.

But how does an application easily deserialize this big blob of JSON into "temperature.desiredTemp"?

Also how does this work in C#?  Does this function call *only* get deltas', or does it get the entire Twin collection (full desired/reported) up front?

If this is deltas only, how would application get ALL properties?  I do think we care about giving the application properties it had previously reported (if it asks for them, like deviceInfo.serialNumber and also temp.desiredTemp) but it's not clear how that would get plumbed here.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+<details>
+<summary>CommandRequest</summary>
+
+```csharp
+public sealed class CommandRequest : MethodRequest
```
jspaith, line 211: Maybe really dumb jspaith vs C# coming up -- will this even work?  MethodRequest is sealed which I thought meant it couldn't be inherited?

iothub/device/devdoc/API Design.md
abhipsaMisra: The examples here are to help us visualize the API and the operation being carried out. I don't expect for this to make it to our official doc comments. We'll probably add a link to the official pnp doc and leave it at that.

iothub/device/devdoc/API Design.md
abhipsaMisra: good point, noted.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
+## Plug and Play convention compatible APIs
+
+### Properties
+
+```csharp
+/// <summary>
+/// Update a single property.
+/// 
+/// Property under the root component:
```
drwill-ms, line 9: Anything beyond a simple summary should go into `<remarks>`

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+public WritablePropertyResponse() { }
+
+/// <summary>
+/// Convenience constructor for specifying the properties.
+/// </summary>
+/// <param name="propertyValue">The unserialized property value.</param>
+/// <param name="ackCode">The acknowledgement code, usually an HTTP Status Code e.g. 200, 400.</param>
+/// <param name="ackVersion">The acknowledgement version, as supplied in the property update request.</param>
+/// <param name="ackDescription">The acknowledgement description, an optional, human-readable message about the result of the property update.</param>
+public WritablePropertyResponse(object propertyValue, int ackCode, long ackVersion, string ackDescription = null)
```
drwill-ms, line 126: Why `object` here for the param, but `dynamic` in the property?

Why this over generic types?

drwill-ms: Have we tested this with various types to ensure it serializes the way we want?

iothub/device/devdoc/API Design.md
abhipsaMisra: Yes, this usually sent as a response to a "writable" property update but doesn't need to be necessarily; eg: if a device starts up and sends an initial value for a "writable property"

https://docs.microsoft.com/en-us/azure/iot-pnp/concepts-convention#writable-properties
"When a device starts up for the first time, it can send an initial value for a reported property if it doesn't receive an initial desired property from the hub. In this case, the device should set av to 1."

It will still always be for a "writable" property, so I don't mind adding that term in here.
I might have missed the name change in my notes, there wasn't a deliberate reason for me calling this `UpdatePropertyAsync`

iothub/device/devdoc/API Design.md
drwill-ms: no, not at the protocol level.

In this case, we're putting more than one property in the same message body.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SendTelemetryAsync(string telemetryName, dynamic telemetryValue, string componentName = default, CancellationToken cancellationToken = default);
+
+/// <summary>
+/// Send a batched instance of telemetry.
+/// </summary>
+/// <param name="telemetryDictionary">The name and value telemetry pairs, as defined in the DTDL interface.</param>
+/// <param name="componentName">The name of the component in which the telemetry is defined. Can be null for telemetry defined under the root interface.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SendTelemetryAsync(IDictionary<string, dynamic> telemetryDictionary, string componentName = default, CancellationToken cancellationToken = default);
```
drwill-ms, line 179: I wouldn't call it a dictionary. Perhaps `telemetryProperties`.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+
+/// <summary>
+/// Send a single instance of telemetry, formatted as per DTDL specifications:
+/// - the payload should be in the format: { "<telemetryName>" : "<telemetry_value>" }
```
drwill-ms, line 183: Why `telemetryName` but also `telemetry_value`? Different naming formats (camelCase vs snake_case).

iothub/device/devdoc/API Design.md
abhipsaMisra: You're right, we cannot inherit sealed classes. 
The idea here is that if we decide to create CommandRequest/Response as a simple wrapper over the existing types (since the differences are minor) we would remove the `sealed` keyword from the existing types.

iothub/device/devdoc/API Design.md
abhipsaMisra: In .NET you'd get only the delta through this API. In order to get all properties you'd call the GetTwin() API. I'll add a corresponding GetProperties() API here.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,255 @@
...
+    {
+
+    }
+
+    internal CommandRequest(string commandName, string componentName, object data) : base (commandName, ConvertToByteArray(data))
+    {
+        ComponentName = componentName;
+    }
+
+    public string ComponentName { get; private set; }
```
drwill-ms, line 227: readonly?

iothub/device/devdoc/API Design.md
abhipsaMisra: This is a copy-paste issue; I copied over the properties and the constructor from different files and they used different types.  I expect this to be an object.

I've not tested this out yet, no. That is the next task.

iothub/device/devdoc/API Design.md
abhipsaMisra: Yes, as David mentioned, this isn't batching multiple messages, but rather creating a single message that contains multiple teletry key-value pairs, put into the same message payload.
.NET doesn't really support message batching over MQTT either, we have a loop-and-send implementation that doesn't really add any value.

iothub/device/devdoc/API Design.md
abhipsaMisra: my bad, I'll update it all to be snake_case (I think it reads better in comments)

iothub/device/devdoc/API Design.md
abhipsaMisra: Sure, but this would deviate from what the .NET client offers right now (we currently support per-method subscription for methods).
By comparison, .NET currently processes all desired property updates through the same callback.

iothub/device/devdoc/API Design.md
ericmitt: In  IoT Explorer it is named: default component

iothub/device/devdoc/API Design.md
ericmitt: Yes I think we should have one  with componentName as parameter

abhipsaMisra: For accessing individual property values from the returned Twin, .NET has an accessor implementation for TwinCollection that'll let you do something like:
TwinCollection receivedPropertyUpdate;
int targetTemperature = receivedPropertyUpdate["thermostat1"]["targetTemperature"];

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,261 @@
...
+public WritablePropertyResponse() { }
+
+/// <summary>
+/// Convenience constructor for specifying the properties.
+/// </summary>
+/// <param name="propertyValue">The unserialized property value.</param>
+/// <param name="ackCode">The acknowledgement code, usually an HTTP Status Code e.g. 200, 400.</param>
+/// <param name="ackVersion">The acknowledgement version, as supplied in the property update request.</param>
+/// <param name="ackDescription">The acknowledgement description, an optional, human-readable message about the result of the property update.</param>
+public WritablePropertyResponse(object propertyValue, int ackCode, long ackVersion, string ackDescription = null)
```
abhipsaMisra, line 132: > Why object here for the param, but dynamic in the property?
Why this over generic types?

object vs dynamic vs T

iothub/device/devdoc/API Design.md
jamdavi: I agree with this. If we ship with the ability per component name we will need to create a dispatcher of sorts.

iothub/device/devdoc/API Design.md
jamdavi: `RespondToWritablePropertyEvent` would be the natural flow here since we have `SubscribeToWritablePropertyEvent`

iothub/device/devdoc/API Design.md
jamdavi: Originally this was a generic type, with a constraint to use some fancy serializer implementation that wasn't great. Moving to dynamic (or object) gave a very simplistic design that meant you could implement one property subscription and get multiple values. 

However, **I agree that generic types would work well for this** and we could have customers use both strongly typed classes and dynamic interchangably. The only caveat here is we'd need to make sure we have a guard around a situation where 

```c#

// Handle allthethings.gif
SubscribeToWritablePropertyEvent<dynamic>("component",  (componentName,  writableProperty) =>
{
    // Fancy stuff here
});

// Use our concrete class to access properties
SubscribeToWritablePropertyEvent<MyCustomObjectClass>("component", "stronglyTypedProperty", (componentName, propertyName, writableProperty) =>
{
    // Fancy stuff here
});

// What happens here? Override the previous, run both?
SubscribeToWritablePropertyEvent<MyCustomObjectClassThatIsNotRelated>("component", "stronglyTypedProperty", (componentName, propertyName, writableProperty) =>
{
    // Fancy stuff here
});
```

iothub/device/devdoc/API Design.md
jamdavi: I will vote for T

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,261 @@
...
+///     "targetTemperature": 21.3,
+///     "targetHumidity": 80
+///   }
+/// },
+/// "$version" : 3
+/// 
+/// </summary>
+/// <param name="propertyActionAsTwinCollection">The action to be taken when a writeable property event is received.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public void SubscribeToWritablePropertyEvent(Action<TwinCollection> propertyActionAsTwinCollection, CancellationToken cancellationToken = default);
```
jamdavi, line 107: Given that we're abstracting the twin away from this, should we also mask this as PropertyCollection?

abhipsaMisra: Get away from Twin, TwinCollection etc "twin"

iothub/device/devdoc/API Design.md
abhipsaMisra: avoid the term "batch" in comments/ API surface

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,261 @@
...
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SendTelemetryAsync(string telemetryName, dynamic telemetryValue, string componentName = default, CancellationToken cancellationToken = default);
+
+/// <summary>
+/// Send a batched instance of telemetry.
+/// </summary>
+/// <param name="telemetryPairs">The name and value telemetry pairs, as defined in the DTDL interface.</param>
+/// <param name="componentName">The name of the component in which the telemetry is defined. Can be null for telemetry defined under the root interface.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SendTelemetryAsync(IDictionary<string, dynamic> telemetryPairs, string componentName = default, CancellationToken cancellationToken = default);
```
abhipsaMisra, line 185: what about when we support other content types (cbor etc)?

iothub/device/devdoc/API Design.md
abhipsaMisra: C is going to have a single global callback

iothub/device/devdoc/API Design.md
drwill-ms: Have y'all gotten feedback from Krzysztof from the Azure SDK team about your API design choices? He's historically given us feedback against using dynamic.

I don't prefer object because it can't be deserialized for user-defined types, and requires casting.

I think you two should try creating and testing these API designs before deciding on a proposed interface.

iothub/device/devdoc/API Design.md
jamdavi: @abhipsaMisra Great point about the RespondTo... naming. Either we use the overload or we create a specific UpdateWritableProperty method.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,302 @@
...
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1010:Generic interface should also be implemented", Justification = "<Pending>")]
+public class PropertyCollection : TwinCollection
+{
+}
+```
+</details>
+
+
+<details>
+<summary>WritableProperty</summary>
```
jspaith, line 136: This isn't referenced anywhere - is it vestigial?

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,302 @@
...
+/// Sends a collection of telemetry.
+/// </summary>
+/// <param name="telemetryDictionary">A dictionary of dynamic objects </param>
+/// <param name="componentName">The component name this telemetry belongs to.</param>
+/// <param name="conventionHandler">A convention handler that defines the content encoding and serializer to use for the telemetry message.</param>
+/// <param name="cancellationToken">A cancellation token.</param>
+/// /// <remarks>
+/// This will either use the <see cref="DefaultConvention"/> to define the encoding and use the default Json serailzier. Refer to the documentation for <see cref="IConventionHandler"/> if you want to use a custom serializer.
+/// </remarks>
+public Task SendTelemetryAsync(IDictionary<string, object> telemetryDictionary, string componentName = default, IConventionHandler conventionHandler = default, CancellationToken cancellationToken = default);
```
jspaith, line 212: So what this would do is to take the IDictinory<string,object> and turn it into some big JSON blob to be sent over, correct?  So like "<"temp",1234>, <"humid",32>" gets turned into {"temp":1234, "humid":32}, correct?

I think we should really think about whether the V1 of API needs this?  It doesn't seem that hard for an application to build up this JSON own their own, even potentially with an Object they own that straight serializes to JSON (so like temp and humid could both be members with [JSON] rules right?).  

So I don't really think this API is needed, it's yet another overload, and especially if there's say need for a CBOR serilaizer at some point we - and not application - end up taking over the responsibility of figuring all this out.

jspaith: Similar though not quite as strong feedback about the `SendTelemetryAsync(string telemetryName, object telemetryValue` overload. 

For dealing with properties, having the SDK have serialization logic adds a bunch of value since there's all that metadata gunk that needs dealing with.  But on telemetry this is just some name:value pair that you can look directly at the DTDL and guess how the underlying body would work.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,302 @@
...
+
+/// <summary>
+/// Update a single property.
+/// </summary>
+/// <param name="propertyName">Property name.</param>
+/// <param name="propertyValue">Property value.</param>
+/// <param name="componentName">The component name this property belongs to.</param>
+/// <param name="conventionHandler">A convention handler that defines the content encoding and serializer to use for the properties.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task UpdatePropertyAsync(string propertyName, object propertyValue, string componentName = default, IConventionHandler conventionHandler = default, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
drwill-ms: `IPropertyFormatHandler`?

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+<details>
+<summary>Properties</summary>
+
+```csharp
+/// <summary>
+/// A container for properties for your device
+/// </summary>
+public class Properties
```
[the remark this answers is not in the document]
drwill-ms: This wouldn't fly for track2 naming in .NET, and we should follow the same standard in track1.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,302 @@
...
+
+<details>
+<summary>PropertyCollection</summary>
+
+```csharp
+/// <summary>
+/// A collection of properties for the device
+/// </summary>
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1010:Generic interface should also be implemented", Justification = "<Pending>")]
+public class PropertyCollection : TwinCollection
```
drwill-ms, line 128: What's the point here? Just to rename from Twin?

Same functionality and existing bugs that come with it?

jspaith: Two reasons - 1 is concept count.  If I'm looking at a DTDL I see something called "properties" all over the place.  Twin is an IoT Hub concept.  A lightweight abstraction makes the story line up more.

Secondly if we even added some sort of additional metadata to Property that wasn't really TwinProperty specific we have a mechanism to do so.

drwill-ms: I'm not saying we like the name `TwinCollection`, but that class has several problems with it. I'm surprised we aren't rewriting it with a name change, rather than just subclassing and giving it a new name.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,302 @@
...
+/// </summary>
+/// <param name="telemetryName">The name of the telemetry to send.</param>
+/// <param name="telemetryValue">The value of the telemetry to send.</param>
+/// <param name="conventionHandler">A convention handler that defines the content encoding and serializer to use for the telemetry message.</param>
+/// <param name="componentName">The component name this telemetry belongs to.</param>
+/// <param name="cancellationToken">A cancellation token.</param>
+/// <remarks>
+/// This will create a single telemetry message and will not combine multiple calls into one message. Use <seealso cref="SendTelemetryAsync(IDictionary{string, dynamic}, string, IConventionHandler, CancellationToken)"/>. Refer to the documentation for <see cref="IConventionHandler"/> if you want to use a custom serializer.
+/// </remarks>
+public Task SendTelemetryAsync(string telemetryName, object telemetryValue, string componentName = default, IConventionHandler conventionHandler = default, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
abhipsaMisra: - I would say that it is safe to ignore the "key-value" and IDict overloads of the telemetry API; I think those are just leftover ones that James probably missed removing.
- We will expose a single API which takes in a `Message`. Now, this `Message` can be constructed with some serialization and encoding customizations (for which we will provide a default implementation). I haven't put the component name into this "serialization and encoding customizations" class, because in my mind component is more of a "modelling" related concept than a telemetry payload related concept. So, in this API, the user would need to set the component name explicitly on the message.
Thoughts?

```csharp
public class TelemetryConvention
{
	private const string ApplicationJsonContentType = "application/json";

	/// <summary>
	///
	/// </summary>
	public static readonly TelemetryConvention Instance = new TelemetryConvention();

	/// <summary>
	///
	/// </summary>
	public Encoding ContentEncoding { get; set; } = Encoding.UTF8;

	/// <summary>
	///
	/// </summary>
	public string ContentType { get; set; } = ApplicationJsonContentType;

	/// <summary>
	///
	/// </summary>
	/// <param name="objectToSerialize"></param>
	/// <returns></returns>
	public virtual string SerializeToString(object objectToSerialize)
	{
		return JsonConvert.SerializeObject(objectToSerialize);
	}

	/// <summary>
	///
	/// </summary>
	/// <param name="contentPayload"></param>
	/// <returns></returns>
	public virtual byte[] EncodeStringToByteArray(string contentPayload)
	{
		return ContentEncoding.GetBytes(contentPayload);
	}

	/// <summary>
	///
	/// </summary>
	/// <param name="objectToSendWithConvention"></param>
	/// <returns></returns>
	public virtual byte[] GetObjectBytes(object objectToSendWithConvention)
	{
		return EncodeStringToByteArray(SerializeToString(objectToSendWithConvention));
	}

	/// <summary>
	///
	/// </summary>
	/// <param name="telemetryName"></param>
	/// <param name="telemetryValue"></param>
	/// <returns></returns>
	public static IDictionary<string, object> FormatTelemetryPayload(string telemetryName, object telemetryValue)
	{
		return new Dictionary<string, object> { { telemetryName, telemetryValue } };
	}
}
```

caller:
```csharp
string telemetryName = "deviceHealth";
var deviceHealth = new DeviceHealth
{
	Status = "Running",
	RunningTimeInSeconds = s_stopwatch.Elapsed.TotalSeconds,
	StopRequested = false,
	StartTime = s_applicationStartTime,
};

IDictionary<string, object> telemetryPayload = TelemetryConvention.FormatTelemetryPayload(telemetryName, deviceHealth);

var messageToSend = new Message(telemetryPayload, TelemetryConvention.Instance)
{
	Properties = { ["property1"] = "myValue" },
	ComponentName = componentName,
};

await _deviceClient.SendTelemetryAsync(messageToSend, cancellationToken);
```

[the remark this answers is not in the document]
abhipsaMisra: the virtual methods in TelemetryConvention can be overridden by the customer to provide their own serilization/ encoding logic.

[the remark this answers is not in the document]
jspaith: OK - I've been staring at this some and I think why we're not totally on same page RE where `component` lives is what we're doing with this "TelemetryConventionThingy" isn't 100% aligned.

In C# if I'm reading correctly, you want this to tie in as a data serializer - in addition to being able to set messageProperties on the TelemetryMessage object as a convenience.  Correct?

C on the other hand is much more limited.  We have the telemetryAttributes thing to help customers "fall into the pit of success" by showing a bunch of parameters that are highly-desirable to be specified.  We are NOT in the serialization business and don't plan to be; the app needs to serialize their body before coming to us.

I'm wondering though if we should be out of this business completely?
So on C, we'd remove telemetryAttributes stuff from proposed API.  Just document the already-existing Message "accessors" (not technically correct but it maps into C# concept) and also add a new "accessor" for ComponentName.  So it becomes

```c
IOTHUB_MESSAGE_HANDLE h = IoTHub_Message_CreateFromString("{some-json-stuff}")
IoTHubMessage_SetEncoding(h,"...); // pre-existing
IoTHubMessage_SetContentType(h, ...);  // pre-existing
IoTHubMessage_SetComponentName(h,...); // would be new.  Right now in summer refresh samples apps need to know that this is "$.sub" and that's only MQTT friendly, so that's lame, but an accesor would "do right"
```

The message is fairly well understood construct already and not hard for a new person to IoT Hub to get head around as this is basic in any messaging framework.  So on C at least I don't see value of the telemtyrAttributes, or if we think it's a good idea we could just add later.

Could C# follow similar pattern?  FWIW (de)serialization of JSON properties is huge value SDK adds, but I think just translating a C# object using JSON serialization the bang-for-bug isn't as big because C# has native patterns around this as-is.  And then if the focus becomes on setting message properties - well add Component as 1st class citizen and document other stuff.

?

iothub/device/devdoc/API Design.md
abhipsaMisra: this was supposed to be "WritablePropertyResponse", renamed.

iothub/device/devdoc/API Design.md
abhipsaMisra: I want to make sure that I understand this correctly - do we want the Message class to only accept the payload bytes and not care about how those bytes were generated?

jspaith: Yes, exactly.  The `Message` class seems pretty straightforward as/is to use, we'd just need to add Component as a top-level component since that is PnP centric.  But otherwise this isn't that hard to serialize at app layer so I think we should leave there.  

C is definitely following this model (we don't have same concept of serializers floating around as common pattern) but on this one I think other languages can keep simple too.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,490 @@
...
+<details>
+<summary>Properties</summary>
+
+```csharp
+/// <summary>
+/// A container for properties.
+/// </summary>
+public class Properties
+{
+    private PropertyCollection _readOnlyProperties;
```
abhipsaMisra, line 68: even though Properties contains all readonly properties as a root level collection, it isn't inherently enumerable right now. This is a TODO

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,490 @@
...
+
+<details>
+<summary>PropertyCollection</summary>
+
+```csharp
+/// <summary>
+///
+/// </summary>
+public class PropertyCollection : IEnumerable<object>
+{
```
abhipsaMisra, line 146: deserialization - todo

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,490 @@
...
+public class WritablePropertyResponse
+{
+    private readonly PropertyConvention _propertyConvention;
+
+    /// <summary>
+    /// Convenience constructor for specifying only the property value.
+    /// </summary>
+    /// <param name="propertyValue">The unserialized property value.</param>
+    /// <param name="propertyConvention"></param>
+    public WritablePropertyResponse(object propertyValue, PropertyConvention propertyConvention)
```
abhipsaMisra, line 220: the writable property response contains user defined payload "Value", so we allow the serialization to be customizable.

iothub/device/devdoc/API Design.md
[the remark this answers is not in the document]
jspaith: Yup, that's exactly what I meant.

FWIW componentName isn't a top-level property on Message in C, so if C# is doing right thing here already you're in good shape.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,490 @@
...
+
+/// <summary>
+/// Update a single property.
+/// </summary>
+/// <param name="propertyName">Property name.</param>
+/// <param name="propertyValue">Property value.</param>
+/// <param name="propertyConvention">A convention handler that defines serializer to use for the properties.</param>
+/// <param name="componentName">The component name this property belongs to.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task UpdatePropertyAsync(string propertyName, object propertyValue, PropertyConvention propertyConvention, string componentName = default, CancellationToken cancellationToken = default);
```
abhipsaMisra, line 21: used PropertyConvention with virtual methods for serialization/ deserialization (vs interfaces), since it is easier to override default behavior (pending internal team discussion though)

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,275 @@
...
+/// <param name="propertyCollection">Reported properties to push.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task UpdatePropertiesAsync(PropertyCollection propertyCollection, CancellationToken cancellationToken = default(CancellationToken));
+
+/// <summary>
+/// Sets the global listener for Writable properties
+/// </summary>
+/// <param name="callback">The global call back to handle all writable property updates.</param>
+/// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
+/// <param name="payloadConvention">A convention handler that defines the content encoding and serializer to use for commands.</param>
```
jamdavi, line 100: Not COMMANDS, but PROPERTIES

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,275 @@
+## Plug and Play convention compatible APIs
+
+#### Common
+
+```csharp
+
+public abstract class PayloadConvention {
```
bikamani, line 7: Do you plan to add headers/summary in this PR? It'll definitely help with PR review

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,275 @@
...
+    protected Serializer();
+    public abstract string ContentType { get; }
+    public abstract bool CheckWritablePropertyResponseType(object typeToCheck);
+    public abstract T ConvertFromObject<T>(object objectToConvert);
+    public abstract T DeserializeToType<T>(string stringToDeserialize);
+    public abstract string SerializeToString(object objectToSerialize);
+}
+
+public sealed class DefaultPayloadConvention : PayloadConvention {
+    public static readonly DefaultPayloadConvention Instance;
```
bikamani, line 30: Why named Instance?

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,275 @@
...
+    public const string ComponentIdentifierValue = "c";
+    public const string ValuePropertyName = "value";
+}
+```
+
+### Properties
+
+```csharp
+/// <summary>
+/// Retrieve the device properties.
```
bikamani, line 81: PnP device.

iothub/device/devdoc/API Design.md
```diff
@@ -0,0 +1,275 @@
...
+/// </summary>
+/// <param name="payloadConvention">A convention handler that defines the content encoding and serializer to use for commands.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+/// <returns>The device properties.</returns>
```
bikamani, line 85: Add PnP here and everywhere for more context.

jamdavi: We'll probably add links to the documentation surrounding the DefaultConvention implementaiton but not be explicit about PnP in our SDK and comments.

Mark and Stefan want to make sure we don't litter the SDK with PnP as that name wasn't favorable.

iothub/device/devdoc/API Design.md
jamdavi: Instead of instantiating the instance it made more sense to create a default constructed instance that way if things change down we can just edit the property in this class and context is preserved

iothub/device/devdoc/API Design.md
jamdavi: I personally didn't want to do that here because of all of the clutter it would add and would spark more conversations about the docs than the API shape. 

I will add these comments when we're ready to get feedback on the docs.

abhipsaMisra:
Since the API surface has now been decided on, closing this PR.
The API surface doc is still available for reference here: https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/pnpApi/iothub/device/devdoc/Convention-based%20operations.md

Note: official feature branch for PnP - https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/pnpApi

PR #1866, opened 2021-03-29
fix(device): Escape the ModelId in the MQTT layer

drwill-ms:
The report from Stefan based on some testing is that when using AMQP this value is escaped, but not for .NET. Other langs (C) have reported always escaping this, and requested that we do too.

The model Id currently can't have characters that would require escaping, so for now this really isn't an issue. However, for self-describing devices there is talk of using a query-string-like approach which would add escapable characters. For MQTT the model Id already sits in the path in the query string so it can't have an embedded query string too.


jamdavi (approved):

abhipsaMisra (approved):

azabbasi (approved):

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -344,7 +344,7 @@ private async Task ConnectAsync(IChannelHandlerContext context)
                 // the SDK will enable plug and play features by appending the model Id to the MQTT CONNECT packet (in the username).
                 if (!(_options?.ModelId).IsNullOrWhiteSpace())
                 {
-                    usernameString += $"&{ModelIdParam}={_options.ModelId}";
+                    usernameString += $"&{ModelIdParam}={Uri.EscapeDataString(_options.ModelId)}";
```
rido-min, line 347: should we escape just the querysting?
I would not expect `dtmi:rido:model;1?hash=C32C4` to be encoded as `dtmi%3Arido%3Amodel%3B1%3Fhash%3DC32C4` rather `dtmi:rido:model;1%3Fhash%3DC32C4`

bikamani (approved):

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
bikamani: What's the value look like for Amqp?

drwill-ms: As long as it is unencoded on the service side, it shouldn't matter how it is encoded over the wire.

drwill-ms: But I'm taking the question seriously and making sure we're doing the right thing. My understanding is that the whole value is being escaped in the C SDK.

drwill-ms: Confirmed whole value is escaped in the C SDK, and arrives at the Hub twin intact (aka unescaped). I'll confirm for this change and then merge.

drwill-ms: Yep, works fine.

Issue #1865, opened 2021-03-26
Singleton HttpClient does not honor DNS changes as in case of failover


rohankhandelwal:
Hi,
The Service SDK seem to be using a singleton httpclient and does not honor dns changes as when they happen in case of Failover of an IoT Hub. 
This blog post explains in detail on how this can be addressed by appropriately setting ConnectionLeaseTimeout
http://byterot.blogspot.com/2016/07/singleton-httpclient-dns.html

- **OS, version, SKU and CPU architecture used:** <VERSION> 
- **Application's .NET Target Framework :** <NETVERSION> 
- **Device:** <DEVICETYPE> 
- **SDK version used:** <VERSION>

PR #1869, opened 2021-03-30
fix(iot-service): Fix bug where registry manager never invalidated DNS cache to force DNS lookup

timtay-microsoft:
The singleton http clients in the service client never update to reflect any DNS changes.

#1865

This new approach simply closes the underlying TCP socket once every 5 minutes and reopens it on the next request. The 5 minutes value here follows track 2 SDK guidance and is a nice balance between not needlessly closing TCP sockets while still being responsive when a failover occurs

Manual testing of this feature looks good. A single registry manager instance stuck in an infinite loop of adding new devices eventually catches on to the failover, and starts successfully registering devices again in a reasonable amount of time.

We don't have any sort of failover tests in our gates, so manual testing is all that I'm relying on here for now.

Edit: After talking a bit with the Azure.Core folks about how they solved this issue, it turns out that they haven't solved this issue for HTTPClients, so they just use a different HTTP primitive (HttpWebRequest) for handling it. With that in mind, this approach is better than trying to migrate our existing code over to the same HTTP primitive just so that we can use the same approach as Azure.Core does.

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -43,13 +43,15 @@ namespace Microsoft.Azure.Devices
 
         private readonly HttpClient _httpClientWithNoTimeout;
         private readonly HttpClientHandler _httpClientHandler;
+        private readonly DelegatingHandler _connectionLeaseTimeoutHandler;
 
         public HttpClientHelper(
             Uri baseAddress,
             IAuthorizationHeaderProvider authenticationHeaderProvider,
             IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping,
             TimeSpan timeout,
-            IWebProxy customHttpProxy)
+            IWebProxy customHttpProxy,
+            int connectionLeaseTimeout)
         {
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
@@ -71,10 +73,12 @@ namespace Microsoft.Azure.Devices
                 _httpClientHandler.Proxy = customHttpProxy;
             }
 
+            _connectionLeaseTimeoutHandler = new ConnectionLeaseTimeoutHandler(connectionLeaseTimeout);
+            _connectionLeaseTimeoutHandler.InnerHandler = _httpClientHandler;
+
             // We need two types of HttpClients, one with our default operation timeout, and one without. The one without will rely on
             // a cancellation token.
-
-            _httpClientWithDefaultTimeout = new HttpClient(_httpClientHandler, false)
+            _httpClientWithDefaultTimeout = new HttpClient(_connectionLeaseTimeoutHandler, false)
             {
                 BaseAddress = _baseAddress,
                 Timeout = _defaultOperationTimeout,
@@ -82,7 +86,7 @@ namespace Microsoft.Azure.Devices
             _httpClientWithDefaultTimeout.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(CommonConstants.MediaTypeForDeviceManagementApis));
             _httpClientWithDefaultTimeout.DefaultRequestHeaders.ExpectContinue = false;
 
-            _httpClientWithNoTimeout = new HttpClient(_httpClientHandler, false)
+            _httpClientWithNoTimeout = new HttpClient(_connectionLeaseTimeoutHandler, false)
             {
                 BaseAddress = _baseAddress,
                 Timeout = Timeout.InfiniteTimeSpan,
@@ -788,6 +792,8 @@ namespace Microsoft.Azure.Devices
         {
             _httpClientWithDefaultTimeout.Dispose();
             _httpClientWithNoTimeout.Dispose();
+
+            _connectionLeaseTimeoutHandler.Dispose();
 
             // Since HttpClientHandler was passed to the 2 HttpClients above, but told them not to dispose it, we want to dispose this after
             _httpClientHandler.Dispose();
--- a/iothub/service/src/ConnectionLeaseTimeoutHandler.cs
+++ b/iothub/service/src/ConnectionLeaseTimeoutHandler.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// HttpClient delegating handler for periodically revoking the connection lease after a specified amount of time. This allows cached HttpClients to periodically do DNS lookups
+    /// which keeps the client communicating with the correct service endpoint in failover scenarios.
+    /// </summary>
+    internal class ConnectionLeaseTimeoutHandler : DelegatingHandler
+    {
+        private readonly Stopwatch _stopwatch;
+        private readonly int _leaseTimeoutMilliseconds;
+        private SemaphoreSlim _timeoutCheckSemaphore;
+
+        internal ConnectionLeaseTimeoutHandler(int leaseTimeoutMilliseconds) : base()
+        {
+            _stopwatch = Stopwatch.StartNew();
+            _timeoutCheckSemaphore = new SemaphoreSlim(1, 1);
+            _leaseTimeoutMilliseconds = leaseTimeoutMilliseconds;
+        }
+
+        protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            // By default, singleton/static HttpClient instances will never do a DNS lookup after the first request, which leads to some
+            // bugs such as https://github.com/Azure/azure-iot-sdk-csharp/issues/1865
+
+            // Special case for lease timeout is when the value is set to a negative value. In those instances, the lease timeout is infinite,
+            // so there is no reason to ever set the ConnectionClose header. As such, there is no need to wait for the semaphore to be available.
+            if (_leaseTimeoutMilliseconds >= 0 && _stopwatch.Elapsed.TotalMilliseconds >= _leaseTimeoutMilliseconds)
+            {
+                // This handler is designed to force the connection to close periodically in order to ensure that httpclients are eventually made aware
+                // of any DNS changes from an IoT Hub failing over.
+                await _timeoutCheckSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+
+                try
+                {
+                    // re-check the stopwatch. This thread may have waited for the semaphore thinking it would need to close the connection and reset
+                    // the timer, but another thread may have already done that since this thread started waiting.
+                    if (_stopwatch.Elapsed.TotalMilliseconds >= _leaseTimeoutMilliseconds)
+                    {
+                        request.Headers.ConnectionClose = true;
+                        _stopwatch.Restart();
+                    }
+                }
+                finally
+                {
+                    _timeoutCheckSemaphore.Release();
+                }
+            }
+
+            return await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            _stopwatch.Reset();
+            _timeoutCheckSemaphore.Dispose();
+            base.Dispose(disposing);
+        }
+    }
+}
--- a/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
+++ b/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Authentication;
 using Microsoft.Azure.Devices.Extensions;
 using Microsoft.Azure.Devices.Generated;
+using Microsoft.Azure.Devices.Shared;
 using Microsoft.Rest;
 using Newtonsoft.Json;
 
@@ -20,6 +21,7 @@ namespace Microsoft.Azure.Devices
     {
         private readonly IotHubGatewayServiceAPIs _client;
         private readonly DigitalTwin _protocolLayer;
+        private readonly DelegatingHandler _connectionLeaseTimeoutHandler;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DigitalTwinClient"/> class.</summary>
@@ -36,7 +38,13 @@ namespace Microsoft.Azure.Devices
 
         private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, params DelegatingHandler[] handlers)
         {
-            _client = new IotHubGatewayServiceAPIs(uri, credentials, handlers);
+            // ConnectionLeaseTimeoutHandler is IDisposable, so save it as a field so it can be disposed later.
+            _connectionLeaseTimeoutHandler = new ConnectionLeaseTimeoutHandler(HttpTransportSettings.DefaultConnectionLeaseTimeoutMilliseconds);
+
+            DelegatingHandler[] handlersWithConnectionLeaseTimeoutHandler = new DelegatingHandler[handlers.Length + 1];
+            handlersWithConnectionLeaseTimeoutHandler[handlers.Length + 1] = _connectionLeaseTimeoutHandler;
+
+            _client = new IotHubGatewayServiceAPIs(uri, credentials, handlersWithConnectionLeaseTimeoutHandler);
             _protocolLayer = new DigitalTwin(_client);
         }
 
@@ -162,6 +170,7 @@ namespace Microsoft.Azure.Devices
         protected virtual void Dispose(bool disposing)
         {
             _client?.Dispose();
+            _connectionLeaseTimeoutHandler?.Dispose();
         }
     }
 }
--- a/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
+++ b/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Azure.Devices
             _connectionLeaseTimeoutHandler = new ConnectionLeaseTimeoutHandler(HttpTransportSettings.DefaultConnectionLeaseTimeoutMilliseconds);
 
             DelegatingHandler[] handlersWithConnectionLeaseTimeoutHandler = new DelegatingHandler[handlers.Length + 1];
-            handlersWithConnectionLeaseTimeoutHandler[handlers.Length + 1] = _connectionLeaseTimeoutHandler;
+            handlersWithConnectionLeaseTimeoutHandler[handlers.Length] = _connectionLeaseTimeoutHandler;
 
             _client = new IotHubGatewayServiceAPIs(uri, credentials, handlersWithConnectionLeaseTimeoutHandler);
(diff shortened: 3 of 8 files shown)


iothub/service/src/HttpTransportSettings.cs
```diff
@@ -27,5 +31,20 @@ public HttpTransportSettings()
...
+        /// </summary>
+        /// <remarks>
+        /// Generally it isn't advised to set this value to either -1 or 0. In the case where this value is set to -1, the singleton HTTP client will indefinitely
+        /// cache the DNS of the host. If the host has any sort of failover, the singleton HTTP client will not be made aware of the DNS change and will 
+        /// continue to send the requests to the now disabled host rather than the backup host. In the case where this value is set to 0, each HTTPS request will close it's socket when finished, rather than allowing for the re-use of sockets for
+        /// subsequent requests. This lack of caching will add a small amount of time to each HTTPS request. For applications that want both the benefits of socket caching and resiliency to DNS changes, this value is advised to be greater than 0. Higher values
+        /// will provide more socket caching performance benefits, while lower values will respond more quickly to DNS changes.
+        /// </remarks>
+        public int ConnectionLeaseTimeoutMilliseconds { get; set; } = DefaultConnectionLeaseTimeoutMilliseconds;
```
timtay-microsoft, line 48: The type of this option matches the type that we set within our SDK. I could see an argument to be made for making this a timespan, but considering how 0 and -1 are the special cases, it would feel weird to ask users construct a timespan. I'm open to suggestions though

iothub/service/src/HttpTransportSettings.cs
```diff
@@ -19,6 +19,10 @@ public HttpTransportSettings()
             Proxy = DefaultWebProxySettings.Instance;
         }
 
+        // -1 is the default value that this SDK had inadvertantly set by not configuring this value in the past. Changing this default may be a
+        // breaking change considering it impacts the performance of the SDK
+        internal const int DefaultConnectionLeaseTimeoutMilliseconds = -1;
```
timtay-microsoft, line 24: This feels worth discussing to me. By defaulting to -1, we avoid any possible breaking change, because this is the default value we already had by not setting this ServicePoint property. However, it also seems reasonable to me to make this value 1 minute or something just so that users don't run into this issue by default when they failover their IoT Hub. The "breaking change" would only be a small performance hit as the http client's check for DNS changes every minute rather than never checking.

iothub/service/src/HttpRegistryManager.cs
```diff
@@ -62,7 +62,8 @@ internal HttpRegistryManager(IotHubConnectionString connectionString, HttpTransp
                 connectionString,
                 ExceptionHandlingHelper.GetDefaultErrorMapping(),
                 s_defaultOperationTimeout,
-                transportSettings.Proxy);
+                transportSettings.Proxy,
```
drwill-ms, line 65: Will we need any changes in other clients?

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// HttpClient delegating handler for periodically revoking the connection lease after a specified amount of time. This allows cached HttpClients to periodically do DNS lookups
+    /// which keeps the client communicating with the correct service endpoint in failover scenarios.
+    /// </summary>
+    internal class ConnectionLeaseTimeoutHandler : DelegatingHandler
```
timtay-microsoft, line 13: Track 2 doesn't have a class defined like this. Instead they use the ServicePointManager class to set a connection lease timeout. For the life of me though, I couldn't get it to work. Since this approach only sets the connection lease timeouts for our own clients rather than for all http clients directed at the same host, I actually prefer this approach anyways though

timtay-microsoft: https://github.com/Azure/azure-sdk-for-net/blob/2c338d26571f4ee84c3c280984af49bcb771fe2a/sdk/core/Azure.Core/src/Pipeline/ServicePointHelpers.cs#L42

common/src/service/HttpClientHelper.cs
```diff
@@ -71,18 +75,19 @@ internal sealed class HttpClientHelper : IHttpClientHelper
                 _httpClientHandler.Proxy = customHttpProxy;
             }
 
+            _connectionLeaseTimeoutHandler = new ConnectionLeaseTimeoutHandler(connectionLeaseTimeout, _httpClientHandler);
```
timtay-microsoft, line 78: The essence of this change is to insert a small delegating handler who periodically inserts the "connectionClose" header on a request to force the http client to close the TCP connection after the request. The subsequent request will open a new TCP connection, but only after doing a DNS lookup first.

timtay-microsoft: This is sort of a breaking change, as users will now see a new TCP connection spawned occasionally, but it is infrequent enough that I don't expect any complaints.

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        private readonly Stopwatch _stopwatch;
+        private readonly TimeSpan _leaseTimeout;
+
+        internal ConnectionLeaseTimeoutHandler(TimeSpan leaseTimeout, HttpClientHandler innerHandler) : base()
+        {
+            _stopwatch = new Stopwatch();
+            _stopwatch.Start();
```
abhipsaMisra, line 21: nit: can be `_stopwatch = Stopwatch.StartNew();`.

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            // By default, singleton/static HttpClient instances will never do a DNS lookup after the first request, which leads to some
+            // bugs such as https://github.com/Azure/azure-iot-sdk-csharp/issues/1865
+
+            // This handler is designed to force the connection to close periodically in order to ensure that httpclients are eventually made aware
+            // of any DNS changes from an IoT Hub failing over.
+            TimeSpan timeSinceConnectionClose = _stopwatch.Elapsed;
+            if (timeSinceConnectionClose >= _leaseTimeout)
+            {
+                request.Headers.ConnectionClose = true;
```
abhipsaMisra, line 36: is this call thread safe? we don't want multiple threads issuing closure one after another.

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+                _stopwatch.Restart();
+            }
+
+            return base.SendAsync(request, cancellationToken);
```
abhipsaMisra, line 40: does `SendAsync()` include implicit open or could we get exceptions saying that the transport layer isn't ready yet?

abhipsaMisra: I think you've answered it here: https://github.com/Azure/azure-iot-sdk-csharp/pull/1869#discussion_r605881735, so it is implicit open.

common/src/service/HttpClientHelper.cs
drwill-ms: As long as we spell out what the impact of the change is in release notes, we should be fine.

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
drwill-ms: Did you try asking Pavel for help with that?

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            _stopwatch.Stop();
```
drwill-ms, line 45: How about `Reset()` instead?

common/src/service/HttpClientHelper.cs
```diff
@@ -43,6 +43,10 @@ internal sealed class HttpClientHelper : IHttpClientHelper
 
         private readonly HttpClient _httpClientWithNoTimeout;
         private readonly HttpClientHandler _httpClientHandler;
+        private readonly DelegatingHandler _connectionLeaseTimeoutHandler;
+
+        // Consistent with track 2 guidance: https://github.com/Azure/azure-sdk-for-net/blob/2c338d26571f4ee84c3c280984af49bcb771fe2a/sdk/core/Azure.Core/src/Pipeline/ServicePointHelpers.cs#L27
+        private readonly TimeSpan connectionLeaseTimeout = TimeSpan.FromMinutes(5);
```
drwill-ms, line 49: `_connectionLeaseTimeout`

common/src/service/HttpClientHelper.cs
drwill-ms: However, it looks like you took out the ability for a client to override our value. Seems like we should make it configurable. Perhaps they want faster than 5 minute detection.

drwill-ms (changes requested):
I'm worried about us going a different route from the track2 SDK, mostly if we haven't checked with them why it isn't working for us.

Also, I want to make sure this fix is for all our service clients.

Finally, I think the value should be configurable and the operation should be threadsafe.

common/src/service/HttpClientHelper.cs
timtay-microsoft: I went ahead and added the ability to configure these timeout again, but with the default still changing to 5 minutes rather than the infinite timeout we had before

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
timtay-microsoft: Good point. I've added a semaphore around the relevant logic so that only one thread should send this header at a time

common/src/service/HttpClientHelper.cs
timtay-microsoft: This field has been deleted, so I'll resolve this comment

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            // By default, singleton/static HttpClient instances will never do a DNS lookup after the first request, which leads to some
+            // bugs such as https://github.com/Azure/azure-iot-sdk-csharp/issues/1865
+
+            // Special case for lease timeout is when the value is set to a negative value. In those instances, the lease timeout is infinite,
+            // so there is no reason to ever set the ConnectionClose header. As such, there is no need to wait for the semaphore to be available.
+            if (_leaseTimeoutMilliseconds >= 0)
```
drwill-ms, line 36: Shouldn't you check to see if the timer elapsed exceeds the threshold before getting the semaphore too? That way you don't get the semaphore on each request.

You'll need to check the value again after the lock, of course, to handle concurrent requests that hit the initial condition and try to get the semaphore.

iothub/service/src/ConnectionLeaseTimeoutHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        protected async override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            // By default, singleton/static HttpClient instances will never do a DNS lookup after the first request, which leads to some
+            // bugs such as https://github.com/Azure/azure-iot-sdk-csharp/issues/1865
+
+            // Special case for lease timeout is when the value is set to a negative value. In those instances, the lease timeout is infinite,
+            // so there is no reason to ever set the ConnectionClose header. As such, there is no need to wait for the semaphore to be available.
+            double millisecondsSinceConnectionClose = _stopwatch.Elapsed.TotalMilliseconds;
+            if (_leaseTimeoutMilliseconds >= 0 && millisecondsSinceConnectionClose >= _leaseTimeoutMilliseconds)
```
drwill-ms, line 37: FWIW, I don't think it makes that code easier to read by making an alias for `_stopwatch.Elapsed.TotalMilliseconds`, so I'd suggest not doing that, but this is just my opinion.

drwill-ms (approved):
I'm happy with the code at this state, but I am still worried we've gone a different direction from the Azure SDK team for track2 SDKs. Please confirm with Pavel why their approach isn't working with us, and that they don't see any trouble with this approach before we merge and ship this.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -36,7 +40,39 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+            {
+                previousDelegatingHandler.InnerHandler = delegatingHandler;
+                previousDelegatingHandler = delegatingHandler;
+            }
+
+            _httpClientHandler = new HttpClientHandler()
+            {
+#if !NET451
+                SslProtocols = TlsVersions.Instance.Preferred,
+                CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck,
```
timtay-microsoft, line 64: These settings weren't present in the HttpClient generated by the generated code layer, but it seemed reasonable to bring them over from our other Http layers for consistency

common/src/service/HttpClientHelper.cs
```diff
@@ -71,18 +73,19 @@ internal sealed class HttpClientHelper : IHttpClientHelper
                 _httpClientHandler.Proxy = customHttpProxy;
             }
 
+            _connectionLeaseTimeoutHandler = new ConnectionLeaseTimeoutHandler(connectionLeaseTimeout, _httpClientHandler);
```
abhipsaMisra, line 76: the `ConnectionLeaseTimeoutHandler` no longer seems to take in an http client handler, missed updating this one I guess

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -36,7 +40,39 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+#endif
+            };
+
+            previousDelegatingHandler.InnerHandler = _httpClientHandler;
```
abhipsaMisra, line 68: This looks good! Some (nit) comments:
- To me, `previousDelegatingHandler` doesn't really share the whole context of the "pipeline" process. Maybe could be renamed to something like `chainedDelegatingHandler`?
- Add a note stating that the `HttpClientHandler` is the one making the network requests, so it needs to be at the end of the pipeline.

common/src/service/HttpClientHelper.cs
timtay-microsoft: Not sure how I missed this, thanks!

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -36,7 +40,40 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+#if !NET451
+                SslProtocols = TlsVersions.Instance.Preferred,
+                CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck,
+#endif
+            };
+
+            // The final chain in the delegating handler chain is the httpClientHandler since it is the one that actually does the network request
+            chainedDelegatingHandler.InnerHandler = _httpClientHandler;
+
+            _httpClient = new HttpClient(_connectionLeaseTimeoutHandler)
```
abhipsaMisra, line 71: shouldn't this be `chainedDelegatingHandler`?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -36,7 +40,40 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+            {
+                BaseAddress = uri
+            };
+
+            _client = new IotHubGatewayServiceAPIs(credentials, _httpClient, false);
```
abhipsaMisra, line 76: I was thinking about this a bit more, are there any other properties that get set into the http client (the one previously created by the protocol layer)? If there is, what do you think about passing the connection lease handler as a delegating handler to the constructor directly, and let the protocol layer create the http client?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -36,7 +40,40 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+            // For instance:
+            // _connectionLeaseTimeoutHandler.InnerHandler = handlers[0],
+            // handlers[0].InnerHandler = handlers[1],
+            // handlers[1].InnerHandler = handlers[2],
+            // ...
+            // handlers[handlers.Length].InnerHandler = _httpClientHandler
+            DelegatingHandler chainedDelegatingHandler = _connectionLeaseTimeoutHandler;
+            foreach (DelegatingHandler delegatingHandler in handlers)
+            {
+                chainedDelegatingHandler.InnerHandler = delegatingHandler;
```
abhipsaMisra, line 56: more questions coming to mind = what if the input is a delegating handler with an inner handler set? we'd then be overwriting it here. maybe some kind of check here could be helpful.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -36,7 +40,40 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+
+            // Assign each custom delegating handler as the inner handler of the previous delegating handler
+            // For instance:
+            // _connectionLeaseTimeoutHandler.InnerHandler = handlers[0],
+            // handlers[0].InnerHandler = handlers[1],
+            // handlers[1].InnerHandler = handlers[2],
+            // ...
+            // handlers[handlers.Length].InnerHandler = _httpClientHandler
+            DelegatingHandler chainedDelegatingHandler = _connectionLeaseTimeoutHandler;
+            foreach (DelegatingHandler delegatingHandler in handlers)
```
abhipsaMisra, line 54: this for-loop is building the handler pipeline from the bottom up, shouldn't we be adding the HttpClientHandler first?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: It should be, good catch

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: Good point. The underlying rest layer seems to generate and cache a single http client, so there isn't really a reason to create it within our hand-written layer.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: I'll be removing this chaining and passing down the delegating handler list + the one new connection lease timeout handler directly to the generated code, so this won't be an issue anymore

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: I'll be removing this chaining and passing down the delegating handler list + the one new connection lease timeout handler directly to the generated code, so this won't be an issue anymore

timtay-microsoft:
Going with the approach in #1874 per track 2 guidance

PR #1870, opened 2021-03-30
fix(service-client): Add default value for transpor type.

vinagesh:
## Checklist

### DIFF
--- a/iothub/service/src/ServiceClient.cs
+++ b/iothub/service/src/ServiceClient.cs
@@ -76,7 +76,7 @@ namespace Microsoft.Azure.Devices
         public static ServiceClient Create(
             string hostName,
             TokenCredential credential,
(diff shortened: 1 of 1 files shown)


iothub/service/src/ServiceClient.cs
```diff
@@ -76,7 +76,7 @@ public static ServiceClient CreateFromConnectionString(string connectionString,
         public static ServiceClient Create(
             string hostName,
             TokenCredential credential,
-            TransportType transportType,
+            TransportType transportType = TransportType.Amqp,
```
vinagesh, line 79: Following the pattern we have for the connection string constructors where the default is AMQP.

drwill-ms (approved):

bikamani (approved):

Issue #1814, opened 2021-03-05
MQTT-WS splits each PUBLISH in to 2 WebSocket data frames


wpbrown:
## Context

- Linux
- net5.0
- VM
- 1.33.1 and master

## Description of the issue

MQTT-WS splits each PUBLISH in to 2 WebSocket data frames (and 2 TCP packets). This doubles the number of packets and could increase latency on highly constrained networks.

### Source

The problem stems from DotNetty [encoding](https://github.com/Azure/DotNetty/blob/47f5ec7303037f1360615d182939e04d8619a2b3/src/DotNetty.Codecs.Mqtt/MqttEncoder.cs#L234) MQTT messages in to 2 buffers.

[ClientWebSocketChannel sends](https://github.com/Azure/azure-iot-sdk-csharp/blob/a0af7b7e03fadfc5f6d9c76ee7380eda4a92b7b2/iothub/device/src/Transport/Mqtt/ClientWebSocketChannel.cs#L267) each of these buffers to the websocket separately resulting multiple frames.

## Code sample exhibiting the issue

SimulatedDevice example set to Mqtt-WS. The header and payload will generate 2 frames.

![image](https://user-images.githubusercontent.com/5326080/110167926-78e0ea00-7dc4-11eb-89c6-142a0b83498d.png)

abhipsaMisra:
Thanks for opening this issue. As you've pointed out, this is a result of the DotNetty implementation that our client sdk uses for Mqtt. This issue would be addressed better in the DotNetty repo: https://github.com/Azure/DotNetty/issues

abhipsaMisra:
@wpbrown I am not able to transfer this issue directly to the DotNetty repo (since I do not have write permissions on that repo). Would you mind closing this issue here and opening it on the DotNetty repo directly?

wpbrown:
@abhipsaMisra it's not clear that this a problem in DotNetty. DotNetty providing downstream handlers separate buffers could be intentional. the closest code to the problem is in this repo in `ClientWebSocketChannel`. that class is taking each buffer and sending it on the websocket separately.

abhipsaMisra:
The splitting of an MQTT publish packets into 2 separate buffers is happening inside the dotnetty library: https://github.com/Azure/DotNetty/blob/dev/src/DotNetty.Codecs.Mqtt/MqttEncoder.cs#L234-L273.

[...]

PR #1873, opened 2021-04-01
Update websocket implementation over MQTT to not send multiple ws frames for a single MQTT packet

abhipsaMisra:
This PR updates our websocket implementation over MQTT to send a single ws frame per MQTT packet being sent.

In short - our websocket implementation sent each message dequeued from the ChannelOutboundBuffer individually, which sometimes resulted in multiple messages being sent for a single MQTT packet (eg. PUBLISH packet encodes the topic and payload separately, so they would show up as different tcp packets on the wire). The fix is to dequeue all enqueued messages and consolidate them before writing to websocket.

I have added the comments in the code explaining the fix, let me know if something is unclear and I can update it.
Fix for #1814 

This github issue also contains some details: https://github.com/Azure/DotNetty/issues/553

### DIFF
--- a/common/src/device/transport/mqtt/ClientWebSocketChannel.cs
+++ b/common/src/device/transport/mqtt/ClientWebSocketChannel.cs
@@ -11,6 +11,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using DotNetty.Buffers;
 using DotNetty.Transport.Channels;
+using Microsoft.Azure.Devices.Shared;
 
 namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 {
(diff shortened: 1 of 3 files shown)


common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -242,38 +243,77 @@ protected override async void DoBeginRead()
 
         protected override async void DoWrite(ChannelOutboundBuffer channelOutboundBuffer)
```
abhipsaMisra, line 244: We have "similar" implementations for both device client mqtt transport and provisioning device client mqtt transport; hence the copy-paste of the same code.

drwill-ms (approved):

barustum (approved):

common/src/device/transport/mqtt/ClientWebSocketChannel.cs
```diff
@@ -242,38 +243,74 @@ protected override async void DoBeginRead()
...
-                        channelOutboundBuffer.Remove();
-                        continue;
+                        // There are two operations carried out while adding a byte buffer component to a CompositeByteBuffer:
+                        // - Increase WriterIndex of the CompositeByteBuffer
+                        //      - increases the count of readable bytes added to the CompositeByteBuffer.
+                        // - Call the method Retain() on the byte buffer being added
+                        //      - The property ReferenceCount of a byte buffer implementation maintains a counter of the no of messages available for dequeueing.
+                        //        A ReferenceCount of 0 indicates that all messages have been flushed and the buffer can be deallocated.
+                        //        By calling the method Retain() on each byte buffer component added to the CompositeByteBuffer,
+                        //        we increase the ReferenceCount by 1 and mark them as ready for dequeuing.
```
barustum, line 294: the spelling of dequeuing here is different than the one on line 291. Not sure which one is right :)

barustum: probably this one?

abhipsaMisra: whoops, "dequeuing" (so many extra alphabets just sitting there needlessly)

PR #1874, opened 2021-04-02
feat(iot-service, shared): Add support for configuring http connection lease timeouts

timtay-microsoft:
(This PR works now) The gist of it is that the servicepointmanager timeout properties work fine for .NET framework applications, but .NET core requires you to construct a SocketsHttpHandler as the basehandler in order to solve this issue. This very closely mimics how Azure.Core solves this same issue.

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -897,6 +897,19 @@ namespace Microsoft.Azure.Devices
 
         internal static HttpClient CreateDefaultClient(IWebProxy webProxy, Uri baseUri, int connectionLeaseTimeoutMilliseconds)
         {
+            // This http client will dispose of this httpMessageHandler when the http client is disposed, so no need to save a local copy
+#pragma warning disable CA2000 // Dispose objects before losing scope (httpClient and messageHandler are disposed outside of this scope)
+            return new HttpClient(CreateDefaultHttpMessageHandler(webProxy, baseUri, connectionLeaseTimeoutMilliseconds))
+#pragma warning restore CA2000 // Dispose objects before losing scope
+            {
+                // Timeouts are handled by the pipeline
+                Timeout = Timeout.InfiniteTimeSpan,
+                BaseAddress = baseUri
+            };
+        }
+
+        internal static HttpMessageHandler CreateDefaultHttpMessageHandler(IWebProxy webProxy, Uri baseUri, int connectionLeaseTimeoutMilliseconds)
+        {
 #pragma warning disable CA2000 // Dispose objects before losing scope (object is returned by this method, so the caller is responsible for disposing it)
 #if NETCOREAPP && !NETCOREAPP2_0 && !NETCOREAPP1_0 && !NETCOREAPP1_1
             // SocketsHttpHandler is only available in netcoreapp2.1 and onwards
@@ -904,13 +917,13 @@ namespace Microsoft.Azure.Devices
             httpMessageHandler.SslOptions.EnabledSslProtocols = TlsVersions.Instance.Preferred;
 #else
             HttpClientHandler httpMessageHandler = new HttpClientHandler();
-#endif
-#pragma warning restore CA2000 // Dispose objects before losing scope
-
-#if !NET451 && !NETCOREAPP
+#if !NET451
             httpMessageHandler.SslProtocols = TlsVersions.Instance.Preferred;
             httpMessageHandler.CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck;
 #endif
+#endif
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
 
             if (webProxy != DefaultWebProxySettings.Instance)
             {
@@ -920,13 +933,7 @@ namespace Microsoft.Azure.Devices
 
             ServicePointHelpers.SetLimits(httpMessageHandler, baseUri, connectionLeaseTimeoutMilliseconds);
 
-            // This http client will dispose of this httpMessageHandler when the http client is disposed, so no need to save a local copy
-            return new HttpClient(httpMessageHandler)
-            {
-                // Timeouts are handled by the pipeline
-                Timeout = Timeout.InfiniteTimeSpan,
-                BaseAddress = baseUri
-            };
+            return httpMessageHandler;
         }
     }
 }
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -40,53 +40,31 @@ namespace Microsoft.Azure.Devices
         // IDisposables
 
         private readonly HttpClient _httpClientWithDefaultTimeout;
-
         private readonly HttpClient _httpClientWithNoTimeout;
-        private readonly HttpClientHandler _httpClientHandler;
 
         public HttpClientHelper(
             Uri baseAddress,
             IAuthorizationHeaderProvider authenticationHeaderProvider,
             IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping,
             TimeSpan timeout,
-            IWebProxy customHttpProxy)
+            IWebProxy customHttpProxy,
+            int connectionLeaseTimeoutMilliseconds)
         {
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
             _defaultOperationTimeout = timeout;
 
-            // HttpClientHandler is IDisposable, so save onto it for disposing.
-            _httpClientHandler = new HttpClientHandler
-            {
-#if !NET451
-                SslProtocols = TlsVersions.Instance.Preferred,
-                CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck,
-#endif
-            };
-
-            if (customHttpProxy != DefaultWebProxySettings.Instance)
-            {
-                _httpClientHandler.UseProxy = customHttpProxy != null;
-                _httpClientHandler.Proxy = customHttpProxy;
-            }
-
             // We need two types of HttpClients, one with our default operation timeout, and one without. The one without will rely on
             // a cancellation token.
 
-            _httpClientWithDefaultTimeout = new HttpClient(_httpClientHandler, false)
-            {
-                BaseAddress = _baseAddress,
-                Timeout = _defaultOperationTimeout,
-            };
+            _httpClientWithDefaultTimeout = CreateDefaultClient(customHttpProxy, _baseAddress, connectionLeaseTimeoutMilliseconds);
+            _httpClientWithDefaultTimeout.Timeout = _defaultOperationTimeout;
             _httpClientWithDefaultTimeout.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(CommonConstants.MediaTypeForDeviceManagementApis));
             _httpClientWithDefaultTimeout.DefaultRequestHeaders.ExpectContinue = false;
 
-            _httpClientWithNoTimeout = new HttpClient(_httpClientHandler, false)
-            {
-                BaseAddress = _baseAddress,
-                Timeout = Timeout.InfiniteTimeSpan,
-            };
+            _httpClientWithNoTimeout = CreateDefaultClient(customHttpProxy, _baseAddress, connectionLeaseTimeoutMilliseconds);
+            _httpClientWithNoTimeout.Timeout = Timeout.InfiniteTimeSpan;
             _httpClientWithNoTimeout.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(CommonConstants.MediaTypeForDeviceManagementApis));
             _httpClientWithNoTimeout.DefaultRequestHeaders.ExpectContinue = false;
 
@@ -788,9 +766,6 @@ namespace Microsoft.Azure.Devices
         {
             _httpClientWithDefaultTimeout.Dispose();
             _httpClientWithNoTimeout.Dispose();
-
-            // Since HttpClientHandler was passed to the 2 HttpClients above, but told them not to dispose it, we want to dispose this after
-            _httpClientHandler.Dispose();
         }
 
         private async Task ExecuteAsync(
@@ -918,6 +893,40 @@ namespace Microsoft.Azure.Devices
             Func<HttpResponseMessage, Task<Exception>> mapToExceptionFunc = errorMapping[response.StatusCode];
             Task<Exception> exception = mapToExceptionFunc(response);
             return await exception.ConfigureAwait(false);
+        }
+
+        internal static HttpClient CreateDefaultClient(IWebProxy webProxy, Uri baseUri, int connectionLeaseTimeoutMilliseconds)
+        {
+#pragma warning disable CA2000 // Dispose objects before losing scope (object is returned by this method, so the caller is responsible for disposing it)
+#if NETCOREAPP && !NETCOREAPP2_0 && !NETCOREAPP1_0 && !NETCOREAPP1_1
+            // SocketsHttpHandler is only available in netcoreapp2.1 and onwards
+            SocketsHttpHandler httpMessageHandler = new SocketsHttpHandler();
+            httpMessageHandler.SslOptions.EnabledSslProtocols = TlsVersions.Instance.Preferred;
+#else
+            HttpClientHandler httpMessageHandler = new HttpClientHandler();
+#endif
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
+#if !NET451 && !NETCOREAPP
+            httpMessageHandler.SslProtocols = TlsVersions.Instance.Preferred;
+            httpMessageHandler.CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck;
+#endif
+
+            if (webProxy != DefaultWebProxySettings.Instance)
+            {
+                httpMessageHandler.UseProxy = webProxy != null;
+                httpMessageHandler.Proxy = webProxy;
+            }
+
+            ServicePointHelpers.SetLimits(httpMessageHandler, baseUri, connectionLeaseTimeoutMilliseconds);
+
+            // This http client will dispose of this httpMessageHandler when the http client is disposed, so no need to save a local copy
+            return new HttpClient(httpMessageHandler)
+            {
+                // Timeouts are handled by the pipeline
+                Timeout = Timeout.InfiniteTimeSpan,
+                BaseAddress = baseUri
+            };
         }
     }
 }
--- a/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
+++ b/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
@@ -37,22 +37,17 @@ namespace Microsoft.Azure.Devices
 
         private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, params DelegatingHandler[] handlers)
         {
-#pragma warning disable CA2000 // Dispose objects before losing scope (objects are disposed of when this client is disposed)
-#if NETCOREAPP && !NETCOREAPP2_0 && !NETCOREAPP1_0 && !NETCOREAPP1_1
-            // SocketsHttpHandler is only available in netcoreapp2.1 and onwards.
-            HttpMessageHandler httpMessageHandler = new SocketsHttpHandler();
-#else
-            HttpMessageHandler httpMessageHandler = new HttpClientHandler();
-#endif
-            // Currently, the connection lease timeout isn't configurable in this client. This may be worth adding later though
-            ServicePointHelpers.SetLimits(httpMessageHandler, uri, ServicePointHelpers.DefaultConnectionLeaseTimeout);
-
+            var httpMessageHandler = HttpClientHelper.CreateDefaultHttpMessageHandler(null, uri, ServicePointHelpers.DefaultConnectionLeaseTimeout);
+#pragma warning disable CA2000 // Dispose objects before losing scope (httpMessageHandlerWithDelegatingHandlers is disposed when the http client owning it is disposed)
             HttpMessageHandler httpMessageHandlerWithDelegatingHandlers = CreateHttpHandlerPipeline(httpMessageHandler, handlers);
+#pragma warning restore CA2000 // Dispose objects before losing scope
 
+#pragma warning disable CA2000 // Dispose objects before losing scope (httpClient is disposed when the protocol layer client owning it is disposed)
             var httpClient = new HttpClient(httpMessageHandlerWithDelegatingHandlers, true)
             {
                 BaseAddress = uri
             };
+#pragma warning restore CA2000 // Dispose objects before losing scope
 
 #pragma warning restore CA2000 // Dispose objects before losing scope
 
--- a/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
+++ b/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
@@ -34,10 +34,64 @@ namespace Microsoft.Azure.Devices
             return new DigitalTwinClient(iotHubConnectionString.HttpsEndpoint, sharedAccessKeyCredential, handlers);
         }
 
+
         private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, params DelegatingHandler[] handlers)
         {
-            _client = new IotHubGatewayServiceAPIs(uri, credentials, handlers);
+#pragma warning disable CA2000 // Dispose objects before losing scope (objects are disposed of when this client is disposed)
+#if NETCOREAPP && !NETCOREAPP2_0 && !NETCOREAPP1_0 && !NETCOREAPP1_1
+            // SocketsHttpHandler is only available in netcoreapp2.1 and onwards.
+            HttpMessageHandler httpMessageHandler = new SocketsHttpHandler();
+#else
+            HttpMessageHandler httpMessageHandler = new HttpClientHandler();
+#endif
+            // Currently, the connection lease timeout isn't configurable in this client. This may be worth adding later though
+            ServicePointHelpers.SetLimits(httpMessageHandler, uri, ServicePointHelpers.DefaultConnectionLeaseTimeout);
+
+            HttpMessageHandler httpMessageHandlerWithDelegatingHandlers = CreateHttpHandlerPipeline(httpMessageHandler, handlers);
+
+            var httpClient = new HttpClient(httpMessageHandlerWithDelegatingHandlers, true)
+            {
+                BaseAddress = uri
+            };
+
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
+            // When this client is disposed, all the http message handlers and delegating handlers will be disposed automatically
+            _client = new IotHubGatewayServiceAPIs(credentials, httpClient, true);
             _protocolLayer = new DigitalTwin(_client);
+        }
+
+        // Creates a single HttpMessageHandler to construct a HttpClient with from a base httpMessageHandler and some number of custom delegating handlers
+        // This is almost a copy of the Microsoft.Rest.ClientRuntime library's implementation, but with the return and parameter type HttpClientHandler replaced
+        // with the more abstract HttpMessageHandler in order for us to set the base handler as either a SocketsHttpHandler for .net core or an HttpClientHandler otherwise
+        // https://github.com/Azure/azure-sdk-for-net/blob/99f4da88ab0aa01c79aa291c6c101ab94c4ac940/sdk/mgmtcommon/ClientRuntime/ClientRuntime/ServiceClient.cs#L376
+        private static HttpMessageHandler CreateHttpHandlerPipeline(HttpMessageHandler httpMessageHandler, params DelegatingHandler[] handlers)
+        {
+            // The RetryAfterDelegatingHandler should be the absoulte outermost handler
+            // because it's extremely lightweight and non-interfering
+            HttpMessageHandler currentHandler =
+#pragma warning disable CA2000 // Dispose objects before losing scope (delegating handler is disposed when the http client that uses it is disposed)
+                new RetryDelegatingHandler(new RetryAfterDelegatingHandler { InnerHandler = httpMessageHandler });
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
+            if (handlers != null)
+            {
+                for (int i = handlers.Length - 1; i >= 0; --i)
+                {
+                    DelegatingHandler handler = handlers[i];
+                    // Non-delegating handlers are ignored since we always
+                    // have RetryDelegatingHandler as the outer-most handler
+                    while (handler.InnerHandler is DelegatingHandler)
+                    {
+                        handler = handler.InnerHandler as DelegatingHandler;
+                    }
+
+                    handler.InnerHandler = currentHandler;
+                    currentHandler = handlers[i];
+                }
+            }
+
+            return currentHandler;
         }
 
         /// <summary>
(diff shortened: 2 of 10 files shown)


common/src/service/HttpClientHelper.cs
```diff
@@ -919,5 +895,37 @@ public void Dispose()
...
+
+            if (webProxy != DefaultWebProxySettings.Instance)
+            {
+                httpMessageHandler.UseProxy = webProxy != null;
+                httpMessageHandler.Proxy = webProxy;
+            }
+
+            ServicePointHelpers.SetLimits(httpMessageHandler, baseUri);
+
+            return new HttpClient(httpMessageHandler)
```
timtay-microsoft, line 923: Double check that this constructor tells the http client to dispose of the httpclienthandler once the httpclient is disposed

iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;net472;net451</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;net472;net451;netcoreapp2.1</TargetFrameworks>
```
timtay-microsoft, line 3: As discussed in standup, in order to handle the use case when a user is on Net Core 2.1, we need to target that framework and add some preprocessor directives to use the SocketsHttpHandler as seen above

shared/src/Microsoft.Azure.Devices.Shared.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">netstandard2.1;netstandard2.0;net472;net451</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">netstandard2.1;netstandard2.0;net472;net451;netcoreapp2.1</TargetFrameworks>
```
timtay-microsoft, line 3: Need to mimic the same target frameworks as the service client library, so added netcoreapp2.1 here as well

common/src/service/HttpClientHelper.cs
timtay-microsoft: It does

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -34,12 +35,33 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
             return new DigitalTwinClient(iotHubConnectionString.HttpsEndpoint, sharedAccessKeyCredential, handlers);
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DigitalTwinClient"/> class.</summary>
+        /// <param name="connectionString">The IoT hub's connection string.</param>
+        /// <param name="transportSettings">The optional settings to apply to the underlying http client.</param>
+        public static DigitalTwinClient CreateFromConnectionString(string connectionString, HttpTransportSettings transportSettings)
```
timtay-microsoft, line 42: Strangely, we didn't have a constructor for our digital twin client that took in the HttpTransportSettings options, so I created one

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -34,12 +35,33 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+
         private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, params DelegatingHandler[] handlers)
         {
             _client = new IotHubGatewayServiceAPIs(uri, credentials, handlers);
             _protocolLayer = new DigitalTwin(_client);
         }
 
+        private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, HttpTransportSettings transportSettings)
+        {
+            _httpClient = HttpClientHelper.CreateDefaultClient(transportSettings.Proxy, uri, transportSettings.ConnectionLeaseTimeoutMilliseconds);
```
timtay-microsoft, line 60: this call to ```CreateDefaultClient``` gives the HttpClient instance all the connectionleasetimeout goodness that the other clients have as well

common/src/service/HttpClientHelper.cs
```diff
@@ -40,53 +40,31 @@ internal sealed class HttpClientHelper : IHttpClientHelper
...
-        private readonly HttpClientHandler _httpClientHandler;
 
         public HttpClientHelper(
             Uri baseAddress,
             IAuthorizationHeaderProvider authenticationHeaderProvider,
             IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping,
             TimeSpan timeout,
-            IWebProxy customHttpProxy)
+            IWebProxy customHttpProxy,
+            int connectionLeaseTimeoutMilliseconds)
```
vinagesh, line 51: Can we use TimeSpan instead of int?

common/src/service/HttpClientHelper.cs
```diff
@@ -919,5 +894,37 @@ public void Dispose()
...
+
+            if (webProxy != DefaultWebProxySettings.Instance)
+            {
+                httpMessageHandler.UseProxy = webProxy != null;
+                httpMessageHandler.Proxy = webProxy;
+            }
+
+            ServicePointHelpers.SetLimits(httpMessageHandler, baseUri, connectionLeaseTimeoutMilliseconds);
+
+            return new HttpClient(httpMessageHandler)
```
abhipsaMisra, line 922: worth adding a note here (and at the warning suppression above) that the http client automatically disposes the handler when it itself is disposed (https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netstandard-2.1#constructors)

common/src/service/HttpClientHelper.cs
```diff
@@ -919,5 +894,37 @@ public void Dispose()
...
+#if NETCOREAPP
+            SocketsHttpHandler httpMessageHandler = new SocketsHttpHandler();
+#else
+            HttpClientHandler httpMessageHandler = new HttpClientHandler();
+#endif
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
+#if !NET451 && !NETCOREAPP
+            httpMessageHandler.SslProtocols = TlsVersions.Instance.Preferred;
+            httpMessageHandler.CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck;
```
abhipsaMisra, line 911: this doesn't work for netcoreapp?

abhipsaMisra: it looks like you can use this for `SocketsHttpHandler`: https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslclientauthenticationoptions?view=netstandard-2.1

common/src/service/HttpClientHelper.cs
```diff
@@ -919,5 +894,37 @@ public void Dispose()
             Task<Exception> exception = mapToExceptionFunc(response);
             return await exception.ConfigureAwait(false);
         }
+
+        internal static HttpClient CreateDefaultClient(IWebProxy webProxy, Uri baseUri, int connectionLeaseTimeoutMilliseconds)
+        {
+
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
vinagesh, line 901: Can you update this comment to why we are disabling this warning here?

common/src/service/HttpClientHelper.cs
```diff
@@ -919,5 +894,37 @@ public void Dispose()
...
+#else
+            HttpClientHandler httpMessageHandler = new HttpClientHandler();
+#endif
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
+#if !NET451 && !NETCOREAPP
+            httpMessageHandler.SslProtocols = TlsVersions.Instance.Preferred;
+            httpMessageHandler.CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck;
+#endif
+
```
vinagesh, line 913: I'm not sure if it is the diff here or if the indentation is really off for the lines below.

vinagesh: Scratch that. It was fine when I did view file. Just looked weird here due to the #if directives.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
abhipsaMisra: I wonder if this will be confusing, since the recommended approach to enable proxy support (a part of HttpTransportSetting) is to do so via a delegating handler; and now, even though we take in an HttpTransportSetting, we can't really set proxy that way.

How does track 2 handle this? They also have concepts of pipeline for the http client, so how do they accept other custom values that are set on the client directly and not via the pipeline.

abhipsaMisra: Also, if we decide to go this way we will need a constructor that takes in both the transport settings and the delegating handlers.

iothub/service/src/ServiceClientTransportSettings.cs
```diff
@@ -29,5 +29,19 @@ public ServiceClientTransportSettings()
...
+        /// each HTTP request and a new TCP connection will be opened upon the next request.
+        /// </summary>
+        /// <remarks>
+        /// By closing cached TCP connections and opening a new one upon the next request, the underlying HTTP client has a chance to do a DNS lookup 
+        /// to validate that it will send the requests to the correct IP address. While it is atypical for a given IoT Hub to change its IP address, it does
+        /// happen when a given IoT Hub fails over into a different region. Because of that, users who expect to failover their IoT Hub at any point
+        /// are advised to set this value to a value of 0 or greater. Larger values will make better use of caching to save network resources over time,
+        /// but smaller values will make the client respond more quickly to failed over IoT Hubs.
+        /// </remarks>
+        public int ConnectionLeaseTimeoutMilliseconds { get; set; } = ServicePointHelpers.DefaultConnectionLeaseTimeout;
```
vinagesh, line 45: TimeSpan would be better than int.

iothub/service/src/ServicePointHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+    //   being the default on NET 4.6.1
+    // The default transport on NET 4.6.1 is WebHttpRequestTransport the default and we are updating both
+    //  limits on the service point
+
+    internal static class ServicePointHelpers
+    {
+#pragma warning disable CA1823 // Unused field
+        internal const int DefaultConnectionLimit = 50;
+#pragma warning restore
+        internal const int DefaultConnectionLeaseTimeout = 300 * 1000; // 5 minutes
```
vinagesh, line 28: TimeSpan here as well

common/src/service/HttpClientHelper.cs
```diff
@@ -919,5 +894,37 @@ public void Dispose()
             Task<Exception> exception = mapToExceptionFunc(response);
             return await exception.ConfigureAwait(false);
         }
+
+        internal static HttpClient CreateDefaultClient(IWebProxy webProxy, Uri baseUri, int connectionLeaseTimeoutMilliseconds)
+        {
+
+#pragma warning disable CA2000 // Dispose objects before losing scope
+#if NETCOREAPP
+            SocketsHttpHandler httpMessageHandler = new SocketsHttpHandler();
```
abhipsaMisra, line 903: it looks like this is available only netcore2.1+, we should make this directive more specific

abhipsaMisra: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler?view=net-5.0#applies-to

iothub/service/src/Microsoft.Azure.Devices.csproj
abhipsaMisra: worth adding a note here stating the reason why we added netcoreapp here (for future reference). Also, did you get a chance to talk to Chris about this?

common/src/service/HttpClientHelper.cs
```diff
@@ -919,5 +894,37 @@ public void Dispose()
...
+            httpMessageHandler.CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck;
+#endif
+
+            if (webProxy != DefaultWebProxySettings.Instance)
+            {
+                httpMessageHandler.UseProxy = webProxy != null;
+                httpMessageHandler.Proxy = webProxy;
+            }
+
+            ServicePointHelpers.SetLimits(httpMessageHandler, baseUri, connectionLeaseTimeoutMilliseconds);
```
abhipsaMisra, line 920: just thinking of the name is, is baseUri really a "limit" that we are setting?

common/src/service/HttpClientHelper.cs
timtay-microsoft: The place where we eventually use this value is as an integer, in milliseconds, so I'd rather not. On top of that, this parameter has special cases when the value is -1 (no connection lease timeout) and 0 (revoke connection lease after every connection)

iothub/service/src/ServicePointHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+    //   we adjust both limits on the client handler
+    // On .NET Standard & NET 4.6+ the HttpClientTransport would default to using the HttpClientHandler
+    //   and there is no easy way to set Keep Alive Connection Timeout but it's mitigated by WebHttpRequestTransport
+    //   being the default on NET 4.6.1
+    // The default transport on NET 4.6.1 is WebHttpRequestTransport the default and we are updating both
+    //  limits on the service point
+
+    internal static class ServicePointHelpers
+    {
+#pragma warning disable CA1823 // Unused field
```
abhipsaMisra, line 25: I'd recommend initializing this field within a pre-processor directive, as opposed to suppressing the warning

iothub/service/src/ServicePointHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+        internal const int DefaultConnectionLimit = 50;
```
abhipsaMisra, line 26: where did we get this value from?

common/src/service/HttpClientHelper.cs
timtay-microsoft: SocketsHttpHandler (which is the class that httpMessageHandler is in the netcoreapp case) doesn't have a SslProtocols field or a CheckCertificateRevocationList field

iothub/service/src/ServicePointHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+#pragma warning restore
+        internal const int DefaultConnectionLeaseTimeout = 300 * 1000; // 5 minutes
+
+#pragma warning disable CA1823 // Unused field
```
vinagesh, line 30: Why do we need to disable this warning?

iothub/service/src/ServicePointHelpers.cs
abhipsaMisra: maybe rename to call out this is "Per Server Connection Limit" and add a comment explaining the reasoning behind the value

common/src/service/HttpClientHelper.cs
timtay-microsoft: I probably goofed the spacing here. I'll fix it

iothub/service/src/ServicePointHelpers.cs
vinagesh, line 26: This can be inside a #if directive instead of disabling the warning.

common/src/service/HttpClientHelper.cs
abhipsaMisra: you can set ssloptions here: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.ssloptions?view=net-5.0#System_Net_Http_SocketsHttpHandler_SslOptions. Does this not work?

common/src/service/HttpClientHelper.cs
vinagesh: Good point. The code will look so ugly :(

vinagesh (approved):

common/src/service/HttpClientHelper.cs
timtay-microsoft: Good catch, I can add that

abhipsaMisra (approved):
A comment about DigitalTwinClient constructors, others are minor.

common/src/service/HttpClientHelper.cs
timtay-microsoft: It already is, haha

common/src/service/HttpClientHelper.cs
timtay-microsoft: In .NET framework, we set the connection lease timeout on a per Uri basis. Admittedly, this reads a bit funny though.

common/src/service/HttpClientHelper.cs
vinagesh: For the first point, it is still recommended to use TimeSpan and in the end to TimeSpan.TotalSeconds. It is less prone to error and easier to read. For the other two, you can use TimeSpan.Zero and TimeSpan.FromSeconds(-1). I recommend to try it out.

iothub/service/src/ServicePointHelpers.cs
timtay-microsoft: I'll add a note that I got these values from Azure.Core and include a link to the values in question

iothub/service/src/ServicePointHelpers.cs
timtay-microsoft: I can remove this now

iothub/service/src/HttpTransportSettings.cs
```diff
@@ -27,5 +27,19 @@ public HttpTransportSettings()
...
+        /// each HTTP request and a new TCP connection will be opened upon the next request.
+        /// </summary>
+        /// <remarks>
+        /// By closing cached TCP connections and opening a new one upon the next request, the underlying HTTP client has a chance to do a DNS lookup 
+        /// to validate that it will send the requests to the correct IP address. While it is atypical for a given IoT Hub to change its IP address, it does
+        /// happen when a given IoT Hub fails over into a different region. Because of that, users who expect to failover their IoT Hub at any point
+        /// are advised to set this value to a value of 0 or greater. Larger values will make better use of caching to save network resources over time,
+        /// but smaller values will make the client respond more quickly to failed over IoT Hubs.
+        /// </remarks>
+        public int ConnectionLeaseTimeoutMilliseconds { get; set; } = ServicePointHelpers.DefaultConnectionLeaseTimeout;
```
vinagesh, line 43: This would also be better as a TimeSpan.

iothub/service/src/ServicePointHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+    // This type manages changing HttpClient defaults to more appropriate values
+    // There are two limits we target:
+    // - Per Server Connection Limit
+    // - Keep Alive Connection Timeout
+    // On .NET Core 2.1 & NET 5.0 the HttpClientTransport would default to using the SocketClientHandler
+    //   we adjust both limits on the client handler
+    // On .NET Standard & NET 4.6+ the HttpClientTransport would default to using the HttpClientHandler
+    //   and there is no easy way to set Keep Alive Connection Timeout but it's mitigated by WebHttpRequestTransport
+    //   being the default on NET 4.6.1
+    // The default transport on NET 4.6.1 is WebHttpRequestTransport the default and we are updating both
```
barustum, line 20: the "...is WebHttpRequestTransport the default..." sounds weird. Missing a word?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: > How does track 2 handle this? They also have concepts of pipeline for the http client, so how do they accept other custom values that are set on the client directly and not via the pipeline.

IIRC they make each library extend the pipelineoptions class to include whatever library specific options there are. In this case, we already don't have that approach and it seems a bit odd to add it now.

The way I see it, the delegating handlers constructor is for users who want complete control over all aspects of the http pipeline that our pnp client will use. This new constructor falls more in line with the other service client constructors we have.

timtay-microsoft: To expand on this a bit, the only constructors in our PL that accept this bug fix are the ones that take in the httpClient rather than taking in any delegating handlers. 

```csharp
// constructors like these allow us to make the httpClient use the SocketsHttpHandler for .netcore cases
protected IotHubGatewayServiceAPIs(HttpClient httpClient, bool disposeHttpClient)

// constructors like these don't accept SocketsHttpHandler since it doesn't extend DelegatingHandler and doesn't extend HttpClientHandler
protected IotHubGatewayServiceAPIs(HttpClientHandler rootHandler, params DelegatingHandler[] handlers)
```

With that in mind, we are forced to own the httpClient in order to fix this issue

timtay-microsoft: > Also, if we decide to go this way we will need a constructor that takes in both the transport settings and the delegating handlers.

This will just lead to confusion since there would be 2 ways to set the proxy, and we don't want that.

abhipsaMisra: > The other reason being that there isn't a way for me to build an httpclient that uses all the delegating handlers.

Isn't this the approach that you had in the other PR? 

Also, even if we do decide to separate the 2 approaches (full control over http client vs we'll set up most of it for you), we need to explicitly specify that even though we have a constructor that takes in HttpTransportSettings, and thus proxy settings, you cannot set proxy up that way.
We will also need to provide some input (doc/ redme/ blog post) on how someone can set up both proxy (and logging, or any other functionality achieved via delegating handler) and also get dns refresh support.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -34,12 +34,35 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+
         private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, params DelegatingHandler[] handlers)
         {
             _client = new IotHubGatewayServiceAPIs(uri, credentials, handlers);
             _protocolLayer = new DigitalTwin(_client);
         }
 
+        private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, HttpTransportSettings transportSettings)
+        {
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
abhipsaMisra, line 59: nit: this also needs explanation -> who is disposing the http client? caller/ digital twin client/ us?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: > we need to explicitly specify that even though we have a constructor that takes in HttpTransportSettings, and thus proxy settings, you cannot set proxy up that way.

This PR does include setting the proxy to the local http client if one is specified in the HttpTransportSettings

timtay-microsoft: > > The other reason being that there isn't a way for me to build an httpclient that uses all the delegating handlers.

> Isn't this the approach that you had in the other PR?

Yes, but I was never fully comfortable with the chaining of delegating handlers that I originally had. As you rightly pointed out, each delegating handler may have some number of inner handlers already. Trying to code around that is a hassle, and I don't see much reason to do so.

abhipsaMisra: > This PR does include setting the proxy to the local http client if one is specified in the HttpTransportSettings

It does? Then mark me satisfied!

common/src/service/HttpClientHelper.cs
```diff
@@ -919,5 +894,39 @@ public void Dispose()
...
+            SocketsHttpHandler httpMessageHandler = new SocketsHttpHandler();
+            httpMessageHandler.SslOptions.EnabledSslProtocols = TlsVersions.Instance.Preferred;
+#else
+            HttpClientHandler httpMessageHandler = new HttpClientHandler();
+#endif
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
+#if !NET451 && !NETCOREAPP
+            httpMessageHandler.SslProtocols = TlsVersions.Instance.Preferred;
+            httpMessageHandler.CheckCertificateRevocationList = TlsVersions.Instance.CertificateRevocationCheck;
```
abhipsaMisra, line 912: nit: I know all these if-def blocks are super messy, but I think I block is missing here:

net451 -> httpclient handler (without ssl protocol and revocation list check)
all other netfm -> httpclient handler (with ssl protocol and revocation list check)
upto netcoreapp 2.0 -> httpclient handler (ssl protocol and revocation list check - available, but missing)
netcoreapp2.1 and above -> sockets http handler (with ssl options)

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -34,12 +34,66 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
-            _client = new IotHubGatewayServiceAPIs(uri, credentials, handlers);
+#pragma warning disable CA2000 // Dispose objects before losing scope (objects are disposed of when this client is disposed)
+#if NETCOREAPP && !NETCOREAPP2_0 && !NETCOREAPP1_0 && !NETCOREAPP1_1
+            // SocketsHttpHandler is only available in netcoreapp2.1 and onwards.
+            HttpMessageHandler httpMessageHandler = new SocketsHttpHandler();
+#else
+            HttpMessageHandler httpMessageHandler = new HttpClientHandler();
+#endif
+            // Currently, the connection lease timeout isn't configurable in this client. This may be worth adding later though
+            ServicePointHelpers.SetLimits(httpMessageHandler, uri, ServicePointHelpers.DefaultConnectionLeaseTimeout);
```
timtay-microsoft, line 48: Since there isn't a great way to add the option to configure this connection lease timeout, I'll just leave it as the default for now. We can always add a constructor later that takes this as a configurable option if a customer asks for it

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -34,12 +34,66 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
             return new DigitalTwinClient(iotHubConnectionString.HttpsEndpoint, sharedAccessKeyCredential, handlers);
         }
 
+
         private DigitalTwinClient(Uri uri, IotServiceClientCredentials credentials, params DelegatingHandler[] handlers)
```
timtay-microsoft, line 38: Unlike in previous iterations, this iteration fixes the stale DNS bug for existing users. They do not need to change constructors because we still only have one, and it has the bug fix

iothub/service/src/ServicePointHelpers.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                case HttpClientHandler httpClientHandler:
+#if !NET451
+                    httpClientHandler.MaxConnectionsPerServer = DefaultMaxConnectionsPerServer;
+#endif
+                    var servicePoint = ServicePointManager.FindServicePoint(baseUri);
+                    servicePoint.ConnectionLeaseTimeout = connectionLeaseTimeoutMilliseconds;
+                    break;
+#if NETCOREAPP
+                case SocketsHttpHandler socketsHttpHandler:
```
abhipsaMisra, line 39: netcoreapp1.0-2.0 won't send the messageHandler as a socket http handler, so this switch block won't be executed, but do you think excluding those versions in the if-def is in any way useful?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -34,12 +34,66 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+        // This is almost a copy of the Microsoft.Rest.ClientRuntime library's implementation, but with the return and parameter type HttpClientHandler replaced
+        // with the more abstract HttpMessageHandler in order for us to set the base handler as either a SocketsHttpHandler for .net core or an HttpClientHandler otherwise
+        // https://github.com/Azure/azure-sdk-for-net/blob/99f4da88ab0aa01c79aa291c6c101ab94c4ac940/sdk/mgmtcommon/ClientRuntime/ClientRuntime/ServiceClient.cs#L376
+        private static HttpMessageHandler CreateHttpHandlerPipeline(HttpMessageHandler httpMessageHandler, params DelegatingHandler[] handlers)
+        {
+            // The RetryAfterDelegatingHandler should be the absoulte outermost handler
+            // because it's extremely lightweight and non-interfering
+            HttpMessageHandler currentHandler =
+#pragma warning disable CA2000 // Dispose objects before losing scope (delegating handler is disposed when the http client that uses it is disposed)
+                new RetryDelegatingHandler(new RetryAfterDelegatingHandler { InnerHandler = httpMessageHandler });
```
timtay-microsoft, line 74: We were getting this retry handler by default already, so this isn't a breaking change.

common/src/service/HttpClientHelper.cs
timtay-microsoft: I gave this a rewrite, can you take another look?

iothub/service/src/ServicePointHelpers.cs
timtay-microsoft: I'll make this if-def match the if-def I have where we decide to make this handler a SocketsHttpHandler vs HttpClientHandler to make this moot

common/src/service/HttpClientHelper.cs
abhipsaMisra: much better!

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -34,12 +34,62 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+#pragma warning restore CA2000 // Dispose objects before losing scope
+
+            if (handlers != null)
+            {
+                for (int i = handlers.Length - 1; i >= 0; --i)
+                {
+                    DelegatingHandler handler = handlers[i];
+                    // Non-delegating handlers are ignored since we always
+                    // have RetryDelegatingHandler as the outer-most handler
+                    while (handler.InnerHandler is DelegatingHandler)
```
drwill-ms, line 80: curious if you could have done: `while (handler.InnerHandler is DelegatingHandler delegatingHandler)`

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -34,12 +34,62 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+                    {
+                        handler = handler.InnerHandler as DelegatingHandler;
```
drwill-ms, line 82: Should only use `as` if you are unsure it is that type, because it will otherwise return null. When you are sure (and you are because of the check in the while loop), you should just cast (assertive).

iothub/service/src/HttpTransportSettings.cs
```diff
@@ -27,5 +27,19 @@ public HttpTransportSettings()
         /// especially when MQTT and AMQP ports are disallowed to the internet.
         /// </remarks>
         public IWebProxy Proxy { get; set; }
+
+        /// <summary>
+        /// How long, in milliseconds, a given cached TCP connection created by this client's HTTP layer will live before being closed.
```
drwill-ms, line 32: Nicely written docs :)

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: Good point. This code was copied from the track 2 folks, so I missed this

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
timtay-microsoft: Possibly. I didn't want to alter this code more than I needed to though since it was copied from track 2 and is designed to mimic the behavior that we already had. I just wanted this method to accept the more generic HttpMessageHandler rather than a HttpClientHandler

PR #1875, opened 2021-04-02
refactor(job-client): Make job client easy to mock

vinagesh:
## Checklist

## Description of the changes
The current JobClient we have is not extendable to mock in a straight forward way.
Given that we do not really have more than one implementation of the client, there is no reason to have a derived class for the concrete implementation.
This PR moves all of the implementation to the base class.
This is not a breaking change as there is no impact to how the customer gets an instance of the JobClient.

I'm submitting this PR for the smallest client we have to see how this pattern will work. I am not sure if we will hit any blockers with the other clients but I will try them out one by one after this.


iothub/service/src/JobClient/JobClient.cs
```diff
@@ -1,32 +1,54 @@
...
-    using Microsoft.Azure.Devices.Shared;
-    using System;
-    using System.Threading;
-    using System.Threading.Tasks;
-
     /// <summary>
     /// Job management
     /// </summary>
-    public abstract class JobClient : IDisposable
+    public class JobClient : IDisposable
```
vinagesh, line 19: No longer abstract and has all the implementation here so all the virtual methods below can be mocked easily.

iothub/service/src/JobClient/JobClient.cs
```diff
@@ -38,67 +60,129 @@ public static JobClient CreateFromConnectionString(string connectionString, Http
...
+                    _httpClientHelper = null;
+                }
+            }
+        }
 
         /// <summary>
         /// Explicitly open the JobClient instance.
         /// </summary>
-        public abstract Task OpenAsync();
+        public virtual Task OpenAsync()
```
vinagesh, line 100: All methods are virtual so that they can be mocked.

drwill-ms (approved):
I like this approach, and think it matches how the code should have originally been written.

As for timing, I know Tim is making some changes to our clients for the DNS caching issue, so we'll want to wait until he has his change in, and rebase on top of it.

We also have the LTS looming, so I figure we should consider waiting until after LTS is declared.

Thoughts?

vinagesh:
> I like this approach, and think it matches how the code should have originally been written.
> 
> As for timing, I know Tim is making some changes to our clients for the DNS caching issue, so we'll want to wait until he has his change in, and rebase on top of it.
> 
> We also have the LTS looming, so I figure we should consider waiting until after LTS is declared.
> 
> Thoughts?

Let me just make a feature branch and check this into that. I can continue this work on a feature branch and merge it when we are ready.

PR #1877, opened 2021-04-07
[PnP] Test impl and samples

abhipsaMisra:
Test implementation by comparing against existing samples

API surface: https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/pnpSurfaceReview/iothub/device/devdoc/API%20Design.md
PR: https://github.com/Azure/azure-iot-sdk-csharp/pull/1858


iothub/device/samples/PnpDeviceSamples/TemperatureController/DeviceHealth.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    internal class DeviceHealth
+    {
+        [JsonPropertyName("status")]
```
abhipsaMisra, line 11: using System.Text.Json here

iothub/device/samples/PnpDeviceSamples/TemperatureController/Models/Thermostat.json
```diff
@@ -11,11 +11,50 @@
...
+      "description": "Temperature in degrees Celsius.",
       "schema": "double",
       "unit": "degreeCelsius"
     },
+    {
+      "@type": [
+        "Telemetry",
+        "DeviceHealth"
+      ],
+      "name": "deviceHealth",
```
abhipsaMisra, line 24: sample telemetry that accepts objects

iothub/device/samples/PnpDeviceSamples/TemperatureController/Models/Thermostat2.json
```diff
@@ -25,17 +25,67 @@
...
+      "@type": [
+        "Property",
+        "InitialValue"
+      ],
+      "name": "initialValue",
+      "displayName": "Initial Value",
+      "description": "Send the initial value for temperature and humidity.",
+      "writable": false,
+      "schema": {
+        "@type": "Object",
```
abhipsaMisra, line 41: sample object based readonly property

iothub/device/samples/PnpDeviceSamples/TemperatureController/Models/Thermostat2.json
```diff
@@ -25,17 +25,67 @@
...
+      "@type": [
+        "Property",
+        "TemperatureRange"
+      ],
+      "name": "temperatureRange",
+      "displayName": "Temperature Range",
+      "description": "Allows to remotely update the thermostat's min and max temperature.",
+      "writable": true,
+      "schema": {
+        "@type": "Object",
```
abhipsaMisra, line 66: sample object based writable property

iothub/device/src/InternalClient.cs
```diff
@@ -1885,6 +1885,77 @@ internal void ValidateModuleTransportHandler(string apiName)
 
         #endregion Module Specific API
 
+        #region Convention driven operations
+
+        internal async Task<Properties> GetPropertiesAsync(CancellationToken cancellationToken)
+        {
+            try
+            {
+                Twin hubTwin = await InnerHandler.SendTwinGetAsync(cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 1894: this is simply converting a twin to properties for now, but this will be reimplemented (since response -> twin -> property brings in deserialization with newtonsoft into picture)

iothub/device/src/WritablePropertyResponse.cs
```diff
@@ -0,0 +1,79 @@
...
+    public class WritablePropertyResponse
+    {
+        private readonly PropertyConvention _propertyConvention;
+
+        /// <summary>
+        /// Convenience constructor for specifying only the property value.
+        /// </summary>
+        /// <param name="propertyValue">The unserialized property value.</param>
+        /// <param name="propertyConvention"></param>
+        public WritablePropertyResponse(object propertyValue, PropertyConvention propertyConvention)
```
abhipsaMisra, line 21: the writable property response contains user defined payload "Value", so we allow the serialization to be customizable.

iothub/device/src/Properties.cs
```diff
@@ -0,0 +1,82 @@
...
+        ///
+        /// </summary>
+        public PropertyCollection Writable { get; private set; }
+
+        /// <summary>
+        /// Get the property from the propeties collection
+        /// </summary>
+        /// <param name="propertyName"></param>
+        /// <returns></returns>
+        public dynamic this[string propertyName]
```
abhipsaMisra, line 47: same logic is repeated within properties and propertycollection (contains, version). Should we place "readonlyprops" under a c# property as well?

iothub/device/devdoc/API Design.md
```diff
@@ -14,72 +15,72 @@ public Task<Properties> GetPropertiesAsync(CancellationToken cancellationToken =
...
 /// <summary>
 /// Sets the global listener for Writable properties
 /// </summary>
 /// <param name="callback">The global call back to handle all writable property updates.</param>
-/// <param name="componentName">The component name this writable property belongs to.</param>
 /// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
-/// <param name="cancellationToken">A cancellation token.</param>
-public Task SubscribeToWritablePropertyEvent(Action<Properties, object> callback, string componentName = default, object userContext = default, CancellationToken cancellationToken = default);
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SubscribeToWritablePropertyEventAsync(Func<PropertyCollection, object, Task> callback, object userContext, CancellationToken cancellationToken = default)
```
abhipsaMisra, line 47: we'll probably want to take something like PropertyConvention in here as well (which will have the deserialization code).
This is TODO, along with getting away from using Twin/TwinCollection internally

iothub/device/devdoc/API Design.md
```diff
@@ -14,72 +15,72 @@ public Task<Properties> GetPropertiesAsync(CancellationToken cancellationToken =
 /// </summary>
 /// <param name="propertyName">Property name.</param>
 /// <param name="propertyValue">Property value.</param>
+/// <param name="propertyConvention">A convention handler that defines serializer to use for the properties.</param>
 /// <param name="componentName">The component name this property belongs to.</param>
-/// <param name="conventionHandler">A convention handler that defines the content encoding and serializer to use for the properties.</param>
 /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
-public Task UpdatePropertyAsync(string propertyName, object propertyValue, string componentName = default, IConventionHandler conventionHandler = default, CancellationToken cancellationToken = default);
+public Task UpdatePropertyAsync(string propertyName, object propertyValue, PropertyConvention propertyConvention, string componentName = default, CancellationToken cancellationToken = default);
```
abhipsaMisra, line 21: Q - do we want to accept reported properties to push as key-value or as PropertyCollection type?

abhipsaMisra: key-value -> we'll serialize (based on what you tell us) vs PropertyCollection will accept a serialized string as input -> user will have to serialize and send

abhipsaMisra: IDictionary<string, object> properties, PropertyConvention propertyConvention,  vs PropoertyCollection - decice internally (managed) since it is essentially serialization detail

iothub/device/devdoc/API Design.md
```diff
@@ -14,72 +15,72 @@ public Task<Properties> GetPropertiesAsync(CancellationToken cancellationToken =
...
 
 <details>
 <summary>Properties</summary>
 
 ```csharp
 /// <summary>
-/// A container for properties for your device
+/// A container for properties.
 /// </summary>
 public class Properties
```
abhipsaMisra, line 57: naming HubProperties?

iothub/device/devdoc/API Design.md
```diff
@@ -14,72 +15,72 @@ public Task<Properties> GetPropertiesAsync(CancellationToken cancellationToken =
 /// </summary>
 /// <param name="propertyName">Property name.</param>
 /// <param name="propertyValue">Property value.</param>
+/// <param name="propertyConvention">A convention handler that defines serializer to use for the properties.</param>
```
abhipsaMisra, line 18: re-naming convention?

iothub/device/devdoc/API Design.md
```diff
@@ -187,41 +395,89 @@ public string AckDescription { get; set; }
...
+    /// <returns></returns>
+    public virtual byte[] GetObjectBytes(object objectToSendWithConvention)
+    {
+        return EncodeStringToByteArray(SerializeToString(objectToSendWithConvention));
+    }
+}
 ```
+</details>
 
 ### Commands
```
abhipsaMisra, line 483: SubscribeForCommands

abhipsaMisra: single global command callback

abhipsaMisra: should we have a convention for commands as well (just for symmetry) - deserialization

iothub/device/src/Properties.cs
```diff
@@ -0,0 +1,82 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.Devices.Shared;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// A container for properties.
+    /// </summary>
+    public class Properties
```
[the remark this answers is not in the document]
abhipsaMisra: done

iothub/device/src/TelemetryConvention.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        /// <summary>
+        ///
+        /// </summary>
+        public static readonly TelemetryConvention Instance = new TelemetryConvention();
+
+        /// <summary>
+        ///
+        /// </summary>
+        public Encoding ContentEncoding { get; set; } = Encoding.UTF8;
```
abhipsaMisra, line 28: setting this via this and message as well - explain why?

iothub/device/devdoc/API Design.md
```diff
@@ -14,72 +15,79 @@ public Task<Properties> GetPropertiesAsync(CancellationToken cancellationToken =
...
+/// <param name="componentName">The component name this property belongs to.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task UpdateWritablePropertyAsync(string propertyName, WritablePropertyResponse writablePropertyResponse, string componentName = default, CancellationToken cancellationToken = default);
+
+/// <summary>
+/// Update properties.
+/// </summary>
+/// <param name="propertyPatch">Reported properties to push.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task UpdatePropertiesAsync(PropertyCollection propertyPatch, CancellationToken cancellationToken = default);
```
abhipsaMisra, line 46: TBD - realted: https://github.com/Azure/azure-iot-sdk-csharp/pull/1877/files#r610183531

iothub/device/devdoc/API Design.md
```diff
@@ -14,72 +15,79 @@ public Task<Properties> GetPropertiesAsync(CancellationToken cancellationToken =
...
 /// <summary>
 /// Sets the global listener for Writable properties
 /// </summary>
 /// <param name="callback">The global call back to handle all writable property updates.</param>
-/// <param name="componentName">The component name this writable property belongs to.</param>
 /// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
-/// <param name="cancellationToken">A cancellation token.</param>
-public Task SubscribeToWritablePropertyEvent(Action<Properties, object> callback, string componentName = default, object userContext = default, CancellationToken cancellationToken = default);
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SubscribeToWritablePropertyEventAsync(Func<PropertyCollection, object, Task> callback, object userContext, CancellationToken cancellationToken = default)
```
jspaith, line 54: I was looking through SubscribeToWritablePropertyEventAsync() and the corresponding PropertyCollection.  It’s not clear to me how the application is going to determine which component a given property is associated with?

If we only had Subscribe where the app had to specify the component name up front and the SDK routed appropriately, we wouldn’t need componentName in this collection.  But since we just have global handler the componentName does need to be part of data deserialized out.  

So not sure if we need some extra handling or i'm missing something?

abhipsaMisra: The component name will be a part of the PropertyCollection returned, it will just not have a unique dedicated accessor. What I mean is, if you iterate through a PropertyCollection returned, they key will always be either a root-level property name or a component name. The application will need to use a combination of switch cases to route their own internal callbacks in their dispatcher.
PropertyCollection:
-> root-property1
-> root-property2
-> component1
    -> component-property1
    -> component-property2
-> root-property3
...

This is how I imagine it to work.
I haven't updated the sample to include a root-level property update yet, but I am planning on adding it in today. I'll share a link once I am done and we can take a look.

iothub/device/devdoc/API Design.md
```diff
@@ -187,95 +402,141 @@ public string AckDescription { get; set; }
...
-public Task SetCommandCallbackHandlerAsync(Func<CommandRequest, object, Task<CommandResponse>> commandCallback, object userContext = default, CancellationToken cancellationToken = default);
+public Task SubscribeToCommandsAsync(Func<CommandRequest, object, Task<CommandResponse>> callback, object userContext, CancellationToken cancellationToken = default)
 ```
 
 <details>
 <summary>CommandRequest</summary>
 
 ```csharp
-public sealed class CommandRequest : MethodRequest
+public sealed class CommandRequest
```
abhipsaMisra, line 505: no longer inherting from MothodRequest, since MethodRequest has other properties defined (response timeout, connection timeout) which aren't relevant for a device-application, and reimplementing it seems better than inheriting and hiding (oop concept-wise)

iothub/device/devdoc/API Design.md
```diff
@@ -187,95 +402,141 @@ public string AckDescription { get; set; }
...
 /// </summary>
-/// <param name="commandCallback">A method implementation that will handle the incoming command.</param>
+/// <param name="callback">A method implementation that will handle the incoming command.</param>
 /// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
 /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
-/// <remarks>
-/// The global command handler will be the fallback handler in the event there is nothing specified in the dispatcher. See the remarks in <see cref="SetCommandCallbackHandlerAsync(Func{CommandRequest, object, Task{CommandResponse}}, object, CancellationToken)"/> for more information.
-/// </remarks>
-public Task SetCommandCallbackHandlerAsync(Func<CommandRequest, object, Task<CommandResponse>> commandCallback, object userContext = default, CancellationToken cancellationToken = default);
+public Task SubscribeToCommandsAsync(Func<CommandRequest, object, Task<CommandResponse>> callback, object userContext, CancellationToken cancellationToken = default)
```
abhipsaMisra, line 498: we'll need to access if providing per-command/ per-component callback is beneficial.
currently users will need to use a combination of command name and component name to implement their own dispatcher

iothub/device/devdoc/API Design.md
jspaith: OK - provided there's a mechanism that's good.  And I like that you don't have an accessor, so you can reuse this same object in `UpdatePropertiesAsync` and let it set the componentName in parameter list of function.   Letting the componentName slip into the properties object gets more complicated.

abhipsaMisra: yeah, we do something like this:

```csharp
PropertyCollection propertyPatch;
if (string.IsNullOrWhiteSpace(componentName))
{
    propertyPatch = new PropertyCollection(properties, propertyConvention);
}
else
{
    properties.Add(PropertyConvention.ComponentIdentifierKey, PropertyConvention.ComponentIdentifierValue);
    var componentProperties = new Dictionary<string, object>
    {
        { componentName, properties }
    };

    propertyPatch = new PropertyCollection(componentProperties, propertyConvention);
}
```

abhipsaMisra: the dispatcher:
https://github.com/Azure/azure-iot-sdk-csharp/blob/e4b3dae81924184093c59a3808d8a78e553cc1ae/iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs#L290-L304

the caller:
https://github.com/Azure/azure-iot-sdk-csharp/blob/e4b3dae81924184093c59a3808d8a78e553cc1ae/iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs#L125-L133

iothub/device/devdoc/API Design.md
```diff
@@ -6,297 +6,160 @@
...
 /// Update a single property.
 /// </summary>
 /// <param name="propertyName">Property name.</param>
 /// <param name="propertyValue">Property value.</param>
 /// <param name="componentName">The component name this property belongs to.</param>
-/// <param name="conventionHandler">A convention handler that defines the content encoding and serializer to use for the properties.</param>
+/// <param name="objectSerializer">A serializer to use for the properties.</param>
 /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
-public Task UpdatePropertyAsync(string propertyName, object propertyValue, string componentName = default, IConventionHandler conventionHandler = default, CancellationToken cancellationToken = default);
+public Task UpdatePropertyAsync(string propertyName, object propertyValue, string componentName = default, ObjectSerializer objectSerializer = default, CancellationToken cancellationToken = default);
```
abhipsaMisra, line 21: updated properties and commands to use the base object serializer class, kept telemetry as "convention" that inherits from object serializer -> to highlight the difference.

- telemetry "convention" also sets some properties on message, which can technically be sone via a static method (helper).
- telemetry "convention" involves both serialization and encoding right now, which properties and commands deal with json/utf-8 by default. can this change in the future?

iothub/device/src/Properties.cs
jamdavi: Agreed, from everything I've read and heard a property is just a property and the only real distinction comes in when it's marked as writable. So making it read only would imply that it could never be writable.

iothub/device/src/Message.cs
```diff
@@ -63,6 +63,22 @@ public Message(byte[] byteArray)
...
 
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="messagePayload"></param>
+        /// <param name="telemetryConvention"></param>
+        public Message(object messagePayload, TelemetryConvention telemetryConvention = default)
+            : this(new MemoryStream(GetTelemetryConventionInstance(telemetryConvention).GetObjectBytes(messagePayload)))
+        {
+            TelemetryConvention convention = GetTelemetryConventionInstance(telemetryConvention);
```
jamdavi, line 74: Do you intend to use GetTelemetryConventionInstance elsewhere in the code? Seems like we could make a private member and set to the constructor instance or the default one.

iothub/device/src/WritablePropertyResponse.cs
```diff
@@ -53,5 +67,11 @@ public WritablePropertyResponse(object propertyValue, int ackCode, long ackVersi
         /// </summary>
         [JsonProperty("ad", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string AckDescription { get; set; }
+
+        /// <summary>
+        /// The serialized property value.
+        /// </summary>
+        [JsonProperty("value")]
+        public JRaw ValueAsJson => new JRaw(_objectSerializer.SerializeToString(Value));
```
jamdavi, line 75: Does this tie the writable property too closely with Newtonsoft? Seeing as how we want to decouple this the serialization shouldn't happen here but when the byte stream for the resulting message is constructed.

iothub/device/devdoc/API Design.md
jamdavi: This sounds correct to me since there can be multiple conventions built on a serializer.

iothub/device/src/Message.cs
abhipsaMisra: GetTelemetryConventionInstance is being used only here, so any kind of refactoring is welcome!

iothub/device/src/WritablePropertyResponse.cs
abhipsaMisra: writable property response could have serialization happening as per two different settings:
- one for the user-supplied value, which could be any object and thus could bring in its own serialization options.
- second for the serialization of the "writablepropertyresponse" type, which does have newtonsoft annotations for its properties (we own it). 

The JRaw ValueAsJson serializes the value as per the serializer provided by the user and stores it as "raw" string. This value isn't serialized again by newtonsoft when serializing the "writablepropertyresponse".

iothub/device/devdoc/API Design.md
abhipsaMisra, line 21: As track 2 about accepting an "ObjectSerializer" in public API.
Should we accept it during device client initialization (clientOptions)? -> this will tie it to the lifetime of the client and can't be changed per API call.

abhipsaMisra: Change property and commands to take in a convention

iothub/device/devdoc/API Design.md
abhipsaMisra: keep PropertyCollection  only

iothub/device/devdoc/API Design.md
```diff
@@ -6,297 +6,142 @@
...
-/// </summary>
-[JsonProperty("ad", DefaultValueHandling = DefaultValueHandling.Ignore)]
-public string AckDescription { get; set; }
+public class PropertyConvention {
+    public static readonly PropertyConvention Instance;
+    public PropertyConvention();
+    public Encoding ContentEncoding { get; }
+    public string ContentType { get; }
+    public ObjectSerializer PayloadSerializer { get; set; }
+    public static PropertyCollection CreatePropertiesPatch(IDictionary<string, object> properties, string componentName=null, PropertyConvention propertyConvention=null);
```
abhipsaMisra, line 73: API takes in IConventionHandler which is composed of ISerializer and IContentEncoder

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -69,22 +100,55 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
+            {
+                var serialNumberReported = properties["serialNumber"];
+            }
+
+            if (properties.Contains(Thermostat2) && properties[Thermostat2].ContainsKey("initialValue"))
+            {
+                var initialStateThermostat2 = properties[Thermostat2]["initialValue"];
+            }
+
+            s_applicationStartTime = DateTimeOffset.Now;
```
drwill-ms, line 125: Should use UtcNow

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -69,22 +100,55 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
+
+            // WritablePropertyEventDispatcherAsync is a dispatcher that we provide to dispatch the individual component/ root-level property callbacks.
+            // Alternatively, you can also have an uber callback that implements a dispatcher internally.
+            // The important thing to note here is that you can only set a single callback for subscribing to property updates.
+            // If you set multiple callbacks to this API, the latest one will be invoked.
+            _logger.LogDebug("Set handler to receive 'targetTemperature' and 'temperatureRange' updates.");
+            _writablePropertyEventCallbacks.Add(Thermostat1, TargetTemperatureUpdateCallbackAsync);
+            _writablePropertyEventCallbacks.Add(Thermostat2, TargetTemperatureUpdateCallbackAsync);
+            _writablePropertyEventCallbacks.Add("temperatureRange", SendTemperatureRangeAsync);
+            await _deviceClient.SubscribeToWritablePropertyEventAsync(WritablePropertyEventDispatcherAsync, userContext: cancellationToken, cancellationToken: cancellationToken);
```
drwill-ms, line 145: `userContent: cancellationToken` ??

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -101,21 +165,75 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
                 await SendTemperatureAsync(Thermostat1, cancellationToken);
                 await SendTemperatureAsync(Thermostat2, cancellationToken);
                 await SendDeviceMemoryAsync(cancellationToken);
+                await SendDeviceHealthTelemetryAsync(Thermostat1, cancellationToken);
 
                 temperatureReset = _temperature[Thermostat1] == 0 && _temperature[Thermostat2] == 0;
-                await Task.Delay(5 * 1000);
+                await Task.Delay(5 * 1000, CancellationToken.None);
```
drwill-ms, line 171: why specify the ct at all?

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -101,21 +165,75 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
+            {
+                Status = "Running",
+                RunningTimeInSeconds = s_stopwatch.Elapsed.TotalSeconds,
+                IsStopRequested = false,
+                StartTime = s_applicationStartTime,
+            };
+
+            IDictionary<string, object> telemetryPayload = TelemetryConvention.FormatTelemetryPayload(deviceHealthName, deviceHealth);
+
+            using var message = new Message(telemetryPayload, s_telemetryConvention)
```
drwill-ms, line 204: What happened to the notion of an overload for Message ctor that takes an object that conforms to an interface to get the json value out of it?

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: yeah, I wanted the cancellation token being observed by the application to be passed to the callback

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: net5 complains if we don't (either supply the API with a cancellation token or explicitly pass in none).

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: this is a more generic version of that, which takes in a dict of telemetry key-value pairs, so you can send multiple telemetry pairs within the same Message body. (we have a "helper" to assist in construction of that payload)

abhipsaMisra:
https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-appendix-overlookedguidelines:
⛔️ DO NOT use abstractions unless the Azure SDK both returns and consumes the abstraction. An abstraction is either an interface or abstract class.

⛔️ DO NOT use interfaces if you can use abstract classes. The only reasons to use an interface are: a) you need to “multiple-inherit”, b) you want structs to implement an abstraction.

iothub/device/devdoc/API Design.md
```diff
@@ -1,302 +1,169 @@
 ## Plug and Play convention compatible APIs
 
+#### Common
+
+```csharp
+public interface ISerializer {
```
abhipsaMisra, line 6: update to abstract class

iothub/device/devdoc/API Design.md
```diff
@@ -1,302 +1,169 @@
...
-/// <remarks>
-/// Use this method when you need to define custom properties for the message.
-/// </remarks>
-/// <param name="telemetryMessage">The custom implemented telemetry message</param>
-/// <param name="componentName">The component name this telemetry belongs to.</param>
-/// <param name="cancellationToken">A cancellation token.</param>
-public Task SendTelemetryAsync(Message telemetryMessage, string componentName = default, CancellationToken cancellationToken = default);
+public class TelemetryCollection : PayloadCollection {
+    public TelemetryCollection(string componentName=null, IPayloadConvention convention=null);
+    public void Add(string telemetryName, object telemetryValue);
```
abhipsaMisra, line 136: throw if repeated key

iothub/device/devdoc/API Design.md
```diff
@@ -1,302 +1,169 @@
...
-/// <param name="componentName">The component name this telemetry belongs to.</param>
-/// <param name="conventionHandler">A convention handler that defines the content encoding and serializer to use for the telemetry message.</param>
-/// <param name="cancellationToken">A cancellation token.</param>
-/// /// <remarks>
-/// This will either use the <see cref="DefaultConvention"/> to define the encoding and use the default Json serailzier. Refer to the documentation for <see cref="IConventionHandler"/> if you want to use a custom serializer.
-/// </remarks>
-public Task SendTelemetryAsync(IDictionary<string, object> telemetryDictionary, string componentName = default, IConventionHandler conventionHandler = default, CancellationToken cancellationToken = default);
+```csharp
+public class TelemetryMessage : Message {
+    public TelemetryMessage(TelemetryCollection telemetryCollection);
```
abhipsaMisra, line 131: what about c2d? is that relevant for pnp?

iothub/device/devdoc/API Design.md
```diff
@@ -1,302 +1,169 @@
...
-/// </summary>
-/// <remarks>
-/// Use this method when you need to define custom properties for the message.
-/// </remarks>
-/// <param name="telemetryMessage">The custom implemented telemetry message</param>
-/// <param name="componentName">The component name this telemetry belongs to.</param>
-/// <param name="cancellationToken">A cancellation token.</param>
-public Task SendTelemetryAsync(Message telemetryMessage, string componentName = default, CancellationToken cancellationToken = default);
+public class TelemetryCollection : PayloadCollection {
+    public TelemetryCollection(string componentName=null, IPayloadConvention convention=null);
```
abhipsaMisra, line 135: make Collection accessible so that you can initialize it beforehand (without having to call .Add multiple times)

abhipsaMisra: telemetry payload => should be updatable after initialization

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -147,93 +283,119 @@ private Task<MethodResponse> HandleMaxMinReportCommand(MethodRequest request, ob
...
-            }
+                AckCode = (int)StatusCode.Completed,
+                AckVersion = desiredProperties.Version,
+                AckDescription = "The operation completed successfully."
+            };
 
-            return Task.CompletedTask;
+            // A property collection with WritablePropertyResponse needs to be always vserialized with NewtonSoft
+            // => we cannot merge it with user defined serialization settings.
+            var propertyPatch = new PropertyCollection();
```
abhipsaMisra, line 357: TODO: WritablePropertyResponse being added to PropertyCollection needs to be serialized using NewtonSoft (has newtonsoft json property names)

iothub/device/devdoc/API Design.md
abhipsaMisra: expose add and add/update

iothub/device/src/Properties.cs
```diff
@@ -0,0 +1,121 @@
...
+        ///
+        /// </summary>
+        public PropertyCollection Writable { get; private set; }
+
+        /// <summary>
+        /// Get the property from the propeties collection
+        /// </summary>
+        /// <param name="propertyName"></param>
+        /// <returns></returns>
+        public dynamic this[string propertyName]
```
abhipsaMisra, line 48: - dynamic vs object vs T
- also make sure that this is in sync with the enumerable return type as well -> however it is implemented

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -101,21 +151,77 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
+            var initialValue = new ThermostatInitialValue()
+            {
+                Temperature = 55,
+                Humidity = 68
+            };
+
+            var propertyPatch = new PropertyCollection(s_payloadConvention);
+            propertyPatch.Add(initialValueName, initialValue, componentName);
+
+            await _deviceClient.UpdatePropertiesAsync(propertyPatch, cancellationToken);
```
abhipsaMisra, line 173: verify that service doesn't actually return anything as a result of this operation

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -125,20 +231,50 @@ private async Task<MethodResponse> HandleRebootCommandAsync(MethodRequest reques
...
+                _logger.LogDebug($"Command: Received - component=\"{request.ComponentName}\"," +
+                    $" updating temperature reading to {updateTemperatureRequest.TargetTemperature}°C after {updateTemperatureRequest.Delay} seconds).");
+                await Task.Delay(TimeSpan.FromSeconds(updateTemperatureRequest.Delay));
+
+                _temperature[request.ComponentName] = updateTemperatureRequest.TargetTemperature;
+
+                var updateTemperatureResponse = new UpdateTemperatureResponse
+                {
+                    TargetTemperature = updateTemperatureRequest.TargetTemperature,
+                    Status = (int)StatusCode.Completed
```
abhipsaMisra, line 255: consider adding a list of common status codes into the library somewhere

iothub/device/src/Properties.cs
```diff
@@ -0,0 +1,121 @@
...
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// A container for properties.
+    /// </summary>
+    public class Properties : IEnumerable<object>
+    {
+        private const string VersionName = "$version";
+        private readonly IDictionary<string, object> _readOnlyProperties = new Dictionary<string, object>();
```
abhipsaMisra, line 17: rename this to "_reportedPropertyCollection"

iothub/device/src/Properties.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            Writable = new PropertyCollection();
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="Properties"/> with the specified collections
+        /// </summary>
+        /// <param name="writablePropertyCollection">A collection of writable properties returned from IoT Hub</param>
+        /// <param name="readOnlyPropertyCollection">A collection of read-only properties returned from IoT Hub</param>
+        internal Properties(PropertyCollection writablePropertyCollection, IDictionary<string, object> readOnlyPropertyCollection)
```
abhipsaMisra, line 32: writablePropertyCollection => requestedPropertyCollection

iothub/device/src/WritablePropertyResponse.cs
```diff
@@ -2,18 +2,29 @@
...
+        private readonly IPayloadConvention _payloadConvention;
+
         /// <summary>
-        /// Empty constructor.
+        /// Convenience constructor for specifying only the property value.
         /// </summary>
-        public WritablePropertyResponse() { }
+        /// <param name="propertyValue">The unserialized property value.</param>
+        /// <param name="payloadConvention"></param>
+        public WritablePropertyResponse(object propertyValue, IPayloadConvention payloadConvention = default)
```
abhipsaMisra, line 21: this constructor isn't required (value, ackCode and ackVersion are required)

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -147,93 +283,119 @@ private Task<MethodResponse> HandleMaxMinReportCommand(MethodRequest request, ob
...
             {
-                string componentName = propertyUpdate.Key;
-                if (_desiredPropertyUpdateCallbacks.ContainsKey(componentName))
+                // The dispatcher key will be either the root-level property name or the component name.
+                string dispatcherKey = propertyUpdate.Key;
+                if (_writablePropertyEventCallbacks.ContainsKey(dispatcherKey))
                 {
-                    _desiredPropertyUpdateCallbacks[componentName]?.Invoke(desiredProperties, componentName);
-                    callbackNotInvoked = false;
+                    return _writablePropertyEventCallbacks[dispatcherKey]?.Invoke(desiredProperties, cancellationToken, dispatcherKey);
```
abhipsaMisra, line 330: don't need to send the entire property collection, just send "KeyValuePair<string, object> propertyUpdate" and "PropertyCollection.Version".

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
jamdavi: Used an abstract base class with virtual overrides. Implemented the Newtonsoft version of this.

Will need to have samples and tutorials on how to extend if using another serializer.

iothub/device/src/Properties.cs
jamdavi: Move everything to object to be inline with the enumerator.

iothub/device/src/Properties.cs
jamdavi: Done

iothub/device/src/Properties.cs
jamdavi: Done

iothub/device/src/WritablePropertyResponse.cs
jamdavi: Done

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
jamdavi: Done

iothub/device/devdoc/API Design.md
jamdavi: DeviceProperties I beleive

iothub/device/src/WritablePropertyResponse.cs
```diff
@@ -22,19 +33,24 @@ public class WritablePropertyResponse
         /// <param name="ackCode">The acknowledgement code, usually an HTTP Status Code e.g. 200, 400.</param>
         /// <param name="ackVersion">The acknowledgement version, as supplied in the property update request.</param>
         /// <param name="ackDescription">The acknowledgement description, an optional, human-readable message about the result of the property update.</param>
-        public WritablePropertyResponse(object propertyValue, int ackCode, long ackVersion, string ackDescription = null)
+        /// <param name="propertyConvention"></param>
+        public WritablePropertyResponse(object propertyValue, int ackCode, long ackVersion, string ackDescription = default, PropertyConvention propertyConvention = default)
         {
-            PropertyValue = propertyValue;
+            // null checks
```
jamdavi: Adding comment for null checks...

iothub/device/devdoc/API Design.md
jamdavi: I think per the conversations last week, it is not beneficial and we should show how to implement the dispatcher.

iothub/device/src/Message.cs
jamdavi: Gone

iothub/device/devdoc/API Design.md
jamdavi: This has been removed and placed inside of the PayloadCollection which removes it from the SDK surface and would need to be discovered to be implemented.

iothub/device/devdoc/API Design.md
abhipsaMisra: ClientProperties (can't put "device" in there)

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -69,22 +86,55 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
+
+            // WritablePropertyEventDispatcherAsync is a dispatcher that we provide to dispatch the individual component/ root-level property callbacks.
+            // Alternatively, you can also have an uber callback that implements a dispatcher internally.
+            // The important thing to note here is that you can only set a single callback for subscribing to property updates.
+            // If you set multiple callbacks to this API, the latest one will be invoked.
+            _logger.LogDebug("Set handler to receive 'targetTemperature' and 'temperatureRange' updates.");
+            _writablePropertyEventCallbacks.Add(Thermostat1, TargetTemperatureUpdateCallbackAsync);
+            _writablePropertyEventCallbacks.Add(Thermostat2, TargetTemperatureUpdateCallbackAsync);
+            _writablePropertyEventCallbacks.Add("temperatureRange", SendTemperatureRangeAsync);
+            await _deviceClient.SubscribeToWritablePropertyEventAsync(WritablePropertyEventDispatcherAsync, userContext: cancellationToken, cancellationToken: cancellationToken);
```
abhipsaMisra, line 131: consider putting the cts in a global var or generate one locally within the callback

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -101,21 +151,80 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
         }
 
         // The callback to handle "reboot" command. This method will send a temperature update (of 0°C) over telemetry for both associated components.
-        private async Task<MethodResponse> HandleRebootCommandAsync(MethodRequest request, object userContext)
+        private async Task<CommandResponse> HandleRebootCommandAsync(CommandRequest request, object userContext)
         {
             try
             {
-                int delay = JsonConvert.DeserializeObject<int>(request.DataAsJson);
+                int delay = request.GetData<int>();
```
abhipsaMisra, line 224: add a sample that works with dynamic/ object as well

iothub/device/src/PropertyConvention.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
jamdavi, line 1: Since we haven't split the conventions up we should get rid of this correct?

We should move the component* items to a constants class.

iothub/device/src/PayloadCollection.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <summary>
+        /// Gets the value of the object from the collection.
+        /// </summary>
+        /// <remarks>
+        /// This class is used for both sending and receiving properties for the device.
+        /// </remarks>
+        /// <typeparam name="T">The type to cast the object to.</typeparam>
+        /// <param name="key">The key of the property to get.</param>
+        /// <returns></returns>
+        public virtual T GetValue<T>(string key)
```
abhipsaMisra, line 68: this is only useful for root level reported properties, we'll have to rethink this

abhipsaMisra: actually this is relevant for root level desired and reported properties.
anything within a component will always be "json-serializer-type"

abhipsaMisra:
Closing in favor of #1924

PR #1878, opened 2021-04-08
refactor(service-client): Make service client easy to mock

vinagesh:
## Checklist

## Description of the changes
Following the same pattern as JobClient, making a change for ServiceClient to make it more mockable.


timtay-microsoft (approved):

iothub/service/src/ServiceClient.cs
```diff
@@ -95,55 +150,184 @@ public static ServiceClient CreateFromConnectionString(string connectionString,
...
+               ExceptionHandlingHelper.GetDefaultErrorMapping(),
+               s_defaultOperationTimeout,
+               transportSettings.HttpProxy);
+            var serviceClient = new ServiceClient(iotHubConnection, iotHubConnectionString.IotHubName, httpClientHelper, options);
             return serviceClient;
         }
 
         /// <summary>
-        /// Open the ServiceClient instance.
+        /// Open the ServiceClient instance. This call is executed over AMQP.
```
drwill-ms, line 165: We should avoid using the word 'executed' for the associated negative connotations. How about 'This call is made over AMQP.'?

iothub/service/tests/ConnectionString/ServiceClientConnectionStringTests.cs
```diff
@@ -26,7 +26,7 @@ public virtual IotHubConnectionStringBuilder Populate(IotHubConnectionStringBuil
         public void ServiceClientConnectionStringDefaultScopeDefaultCredentialTypeTest()
         {
             string connectionString = "HostName=acme.azure-devices.net;SharedAccessKeyName=AllAccessKey;SharedAccessKey=dGVzdFN0cmluZzE=";
-            var serviceClient = (AmqpServiceClient)ServiceClient.CreateFromConnectionString(connectionString);
+            var serviceClient = (ServiceClient)ServiceClient.CreateFromConnectionString(connectionString);
```
drwill-ms, line 29: Does it still require casting now?

drwill-ms (approved):

iothub/service/tests/ConnectionString/ServiceClientConnectionStringTests.cs
vinagesh: Good point. It should not. I will double check and update accordingly.

PR #1879, opened 2021-04-12
fix(iot-dev): Fix null reference exception case within RetryDelegatingHandler

timtay-microsoft:
If Dispose is called while this semaphore is in use, the other thread using the semaphore will hit a null reference exception when it tries to release the now null semaphore. It would be more appropriate to just hit an ObjectDisposedException instead.

The particular customer issue that motivated me to look into this code involved the user disposing a client without closing it first, which opened the possibility that a connection loss event would trigger while the object is being disposed. In general, though, Dispose should only be called when a client is no longer open.

### DIFF
--- a/iothub/device/src/Transport/RetryDelegatingHandler.cs
+++ b/iothub/device/src/Transport/RetryDelegatingHandler.cs
@@ -968,7 +968,6 @@ namespace Microsoft.Azure.Devices.Client.Transport
                 _handleDisconnectCts?.Dispose();
 
                 _handlerSemaphore?.Dispose();
-                _handlerSemaphore = null;
             }
         }
     }


iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -968,7 +968,6 @@ protected override void Dispose(bool disposing)
                 _handleDisconnectCts?.Dispose();
 
                 _handlerSemaphore?.Dispose();
-                _handlerSemaphore = null;
             }
```
timtay-microsoft, line 971: Line 917 is where the customer got the null reference exception. The order of events seemed to be:

1) They hit a connection loss event that was reported from the mqtt layer.
2) This layer acquired the handlerSemaphore in order to start reopening the mqtt layer.
3) The application layer disposed of this whole client, and set this semaphore to null in doing so
4) When the mqtt reconnection logic finished/failed, it attempted to release the semaphore, but it was null, so a null ref exception was thrown

timtay-microsoft: The real customer fix here in my mind is to only dispose of a client that has been closed. That way there is no chance the user would see either the null reference exception (without this change) or the ObjectDisposedException (with this change).

I'd still like to make this change because an ObjectDisposedException makes a lot more sense to throw in the case where a user still decides to dispose a client without closing it first.

timtay-microsoft: It's also worth noting that we don't set any other semaphores to null after disposing them, likely for the same reason. See [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/8df9dd76aa6efa87d9085408d221ca61160e1656/iothub/device/src/InternalClient.cs#L1892) and [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/8df9dd76aa6efa87d9085408d221ca61160e1656/iothub/device/src/Transport/Amqp/AmqpUnit.cs#L854)

drwill-ms (approved):

iothub/device/src/Transport/RetryDelegatingHandler.cs
bikamani: wondering if it makes sense to change behavior (throws ObjectDisposedException) now when you are planning to add real fix (may be soon)?

timtay-microsoft:
There isn't much value in this change after all, so I'm closing this PR

PR #1880, opened 2021-04-14
Update service client API version to 2021-04-12

drwill-ms:
In prep for adding support for parentScope feature

### DIFF
--- a/iothub/service/src/ClientApiVersionHelper.cs
+++ b/iothub/service/src/ClientApiVersionHelper.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-
 namespace Microsoft.Azure.Devices
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


bikamani (approved):

timtay-microsoft (approved):

iothub/service/src/ClientApiVersionHelper.cs
```diff
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
```
vinagesh, line 1: Are we going to release a preview SDK first or just GA SDK?

vinagesh (approved):

iothub/service/src/ClientApiVersionHelper.cs
drwill-ms: Good question. I wasn't planning on doing a preview as the feature add is small (a property on a device class), and otherwise we're just targeting a new service api version.

I'm open to pushback though.

vinagesh: Seems reasonable given the feature is very small(rather not even a new API). We would anyway release this when we release preview next time as we will merge master to preview. Chronologically it might be weird to have a new feature in GA first and preview later but it happens most of the time. I'm not opposed to it but got me thinking. On the contrary, it is a lot of process to add if we always have to release to preview first so this seems good to me.

PR #1885, opened 2021-04-15
Add ParentScope property to Device

drwill-ms:
Also added to Twin, along with ParentScope property. Updated a test to validate edge devices can be put into a parent/child relationship.

### DIFF
--- a/e2e/test/iothub/service/RegistryManagerE2ETests.cs
+++ b/e2e/test/iothub/service/RegistryManagerE2ETests.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
 using FluentAssertions;
-using Microsoft.Azure.Devices.Common.Exceptions;
 using Microsoft.Azure.Devices.E2ETests.Helpers;
 using Microsoft.Azure.Devices.Shared;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -42,24 +41,53 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
         [LoggedTestMethod]
         public async Task RegistryManager_AddAndRemoveDeviceWithScope()
         {
-            var registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+            // arrange
 
+            string edgeId1 = _devicePrefix + Guid.NewGuid();
+            string edgeId2 = _devicePrefix + Guid.NewGuid();
             string deviceId = _devicePrefix + Guid.NewGuid();
 
-            var edgeDevice = new Device(deviceId)
+            using var registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+
+            try
             {
-                Capabilities = new DeviceCapabilities { IotEdge = true }
-            };
-            edgeDevice = await registryManager.AddDeviceAsync(edgeDevice).ConfigureAwait(false);
+                // act
 
-            var leafDevice = new Device(Guid.NewGuid().ToString()) { Scope = edgeDevice.Scope };
-            Device receivedDevice = await registryManager.AddDeviceAsync(leafDevice).ConfigureAwait(false);
+                // Create a top-level edge device.
+                var edgeDevice1 = new Device(edgeId1)
+                {
+                    Capabilities = new DeviceCapabilities { IotEdge = true }
+                };
+                edgeDevice1 = await registryManager.AddDeviceAsync(edgeDevice1).ConfigureAwait(false);
 
-            Assert.IsNotNull(receivedDevice);
-            Assert.AreEqual(leafDevice.Id, receivedDevice.Id, $"Expected Device ID={leafDevice.Id}; Actual Device ID={receivedDevice.Id}");
-            Assert.AreEqual(leafDevice.Scope, receivedDevice.Scope, $"Expected Device Scope={leafDevice.Scope}; Actual Device Scope={receivedDevice.Scope}");
-            await registryManager.RemoveDeviceAsync(leafDevice.Id).ConfigureAwait(false);
-            await registryManager.RemoveDeviceAsync(edgeDevice.Id).ConfigureAwait(false);
+                // Create a second-level edge device with edge 1 as the parent.
+                var edgeDevice2 = new Device(edgeId2)
+                {
+                    Capabilities = new DeviceCapabilities { IotEdge = true },
+                    ParentScopes = { edgeDevice1.Scope },
+                };
+                edgeDevice2 = await registryManager.AddDeviceAsync(edgeDevice2).ConfigureAwait(false);
+
+                // Create a leaf device with edge 2 as the parent.
+                var leafDevice = new Device(deviceId) { Scope = edgeDevice2.Scope };
+                leafDevice = await registryManager.AddDeviceAsync(leafDevice).ConfigureAwait(false);
+
+                // assert
+
+                edgeDevice2.ParentScopes.FirstOrDefault().Should().Be(edgeDevice1.Scope, "The parent scope should be respected as set.");
+
+                leafDevice.Id.Should().Be(deviceId, "The device Id should be respected as set.");
+                leafDevice.Scope.Should().Be(edgeDevice2.Scope, "The device scope should be respected as set.");
+                leafDevice.ParentScopes.FirstOrDefault().Should().Be(edgeDevice2.Scope, "The service should have copied the edge's scope to the leaf device's parent scope array.");
+            }
+            finally
+            {
+                // clean up
+
+                await registryManager.RemoveDeviceAsync(deviceId).ConfigureAwait(false);
+                await registryManager.RemoveDeviceAsync(edgeId1).ConfigureAwait(false);
+                await registryManager.RemoveDeviceAsync(edgeId2).ConfigureAwait(false);
+            }
         }
 
         [LoggedTestMethod]
@@ -243,7 +271,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
             }
             finally
             {
-                await Cleanup(client, testDeviceId).ConfigureAwait(false);
+                await CleanupAsync(client, testDeviceId).ConfigureAwait(false);
             }
         }
 
@@ -290,7 +318,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
             }
             finally
             {
-                await Cleanup(client, testDeviceId).ConfigureAwait(false);
+                await CleanupAsync(client, testDeviceId).ConfigureAwait(false);
             }
         }
 
@@ -325,11 +353,11 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
             }
             finally
             {
-                await Cleanup(client, module.DeviceId).ConfigureAwait(false);
+                await CleanupAsync(client, module.DeviceId).ConfigureAwait(false);
             }
         }
 
-        private async Task Cleanup(RegistryManager client, string deviceId)
+        private async Task CleanupAsync(RegistryManager client, string deviceId)
         {
             // cleanup
             try
--- a/iothub/service/src/Device.cs
+++ b/iothub/service/src/Device.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using Microsoft.Azure.Devices.Shared;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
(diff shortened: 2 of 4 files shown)


iothub/service/src/Device.cs
```diff
@@ -109,9 +110,29 @@ public Device(string id)
...
+        /// relationship.
         /// </summary>
+        /// <remarks>
+        /// For edge devices, the auto-generated device scope consists of a prefix "ms-azure-iot-edge://", the name of the parent edge device, a dash, and the
+        /// generation Id of the parent edge device (e.g. ms-azure-iot-edge://MyEdgeDevice-000000000000000000).
+        ///
+        /// For leaf devices, the value to set a parent edge device can be retrieved from the parent edge device's <see cref="Scope"/> property.
+        /// </remarks>
         [JsonProperty(PropertyName = "deviceScope", NullValueHandling = NullValueHandling.Ignore)]
         public virtual string Scope { get; set; }
```
drwill-ms, line 123: Too late to change this to DeviceScope so it is consistent with the json property name, and what I copied over to Twin. Oh well.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -42,24 +41,42 @@ public async Task RegistryManager_BadProxy_ThrowsException()
         [LoggedTestMethod]
         public async Task RegistryManager_AddAndRemoveDeviceWithScope()
         {
-            var registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
-
+            string edgeId1 = _devicePrefix + Guid.NewGuid();
```
vinagesh, line 44: string interpolation?

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -42,24 +41,42 @@ public async Task RegistryManager_BadProxy_ThrowsException()
...
+                leafDevice = await registryManager.AddDeviceAsync(leafDevice).ConfigureAwait(false);
 
-            Assert.IsNotNull(receivedDevice);
-            Assert.AreEqual(leafDevice.Id, receivedDevice.Id, $"Expected Device ID={leafDevice.Id}; Actual Device ID={receivedDevice.Id}");
-            Assert.AreEqual(leafDevice.Scope, receivedDevice.Scope, $"Expected Device Scope={leafDevice.Scope}; Actual Device Scope={receivedDevice.Scope}");
-            await registryManager.RemoveDeviceAsync(leafDevice.Id).ConfigureAwait(false);
-            await registryManager.RemoveDeviceAsync(edgeDevice.Id).ConfigureAwait(false);
+                edgeDevice2.ParentScopes.FirstOrDefault().Should().Be(edgeDevice1.Scope, "The parent scope should be respected as set.");
+
+                leafDevice.Id.Should().Be(leafDevice.Id, "The device Id shoudl be respected as set.");
```
vinagesh, line 70: leafDevice is replaced by the response in line 65 so this will always be true. You might want to store the initial expected leafDevice values.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -42,24 +41,42 @@ public async Task RegistryManager_BadProxy_ThrowsException()
...
+                leafDevice.Scope.Should().Be(leafDevice.Scope, "The parent scope should be respected as set.");
```
vinagesh, line 71: Same for these

iothub/service/src/Device.cs
```diff
@@ -109,9 +110,29 @@ public Device(string id)
...
+
+        /// <summary>
+        /// The scopes of the upper level edge devices if applicable.
+        /// </summary>
+        /// <remarks>
+        /// For edge devices, the value to set a parent edge device can be retrieved from the parent edge device's <see cref="Scope"/> property.
+        ///
+        /// For leaf devices, this could be set to the same value as <see cref="Scope"/> or left for the service to copy over.
```
vishnureddy17, line 131: Is it true that the service will copy it over? I remember having a discussion on teams talking about how this was not the case.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
drwill-ms: good catch. Thanks.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
drwill-ms: IMO $"{_devicePrefix}{Guid.NewGuild()}" is messier than the current code. Thoughts?

vinagesh: I'm ok with either. I thought the preferred way was using string interpolation.

iothub/service/src/Device.cs
```diff
@@ -109,9 +110,29 @@ public Device(string id)
...
         /// <summary>
-        /// Scope to which this device instance belongs to
+        /// The scope of the device. For edge devices, this is auto-generated and immutable. For leaf devices, set this to create child/parent
+        /// relationship.
         /// </summary>
+        /// <remarks>
+        /// For edge devices, the auto-generated device scope consists of a prefix "ms-azure-iot-edge://", the name of the parent edge device, a dash, and the
+        /// generation Id of the parent edge device (e.g. ms-azure-iot-edge://MyEdgeDevice-000000000000000000).
+        ///
+        /// For leaf devices, the value to set a parent edge device can be retrieved from the parent edge device's <see cref="Scope"/> property.
```
vishnureddy17, line 120: When we say "leaf device" here, does that imply that the device is a non-edge device? 

Because if we are thinking about this in terms of the parent/child relationship tree, an edge device with no children would be a "leaf" of that tree.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -42,24 +41,42 @@ public async Task RegistryManager_BadProxy_ThrowsException()
...
+                leafDevice = await registryManager.AddDeviceAsync(leafDevice).ConfigureAwait(false);
 
-            Assert.IsNotNull(receivedDevice);
-            Assert.AreEqual(leafDevice.Id, receivedDevice.Id, $"Expected Device ID={leafDevice.Id}; Actual Device ID={receivedDevice.Id}");
-            Assert.AreEqual(leafDevice.Scope, receivedDevice.Scope, $"Expected Device Scope={leafDevice.Scope}; Actual Device Scope={receivedDevice.Scope}");
-            await registryManager.RemoveDeviceAsync(leafDevice.Id).ConfigureAwait(false);
-            await registryManager.RemoveDeviceAsync(edgeDevice.Id).ConfigureAwait(false);
+                edgeDevice2.ParentScopes.FirstOrDefault().Should().Be(edgeDevice1.Scope, "The parent scope should be respected as set.");
+
+                leafDevice.Id.Should().Be(deviceId, "The device Id shoudl be respected as set.");
```
vinagesh, line 70: leafDevice value is replaced by the response in line 66. So this will always be true. You might want to store the expected values instead of replacing. Same for the below.

shared/src/Twin.cs
```diff
@@ -151,6 +151,19 @@ public Twin(TwinProperties twinProperties)
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
         public X509Thumbprint X509Thumbprint { get; internal set; }
 
+        /// <summary>
```
vinagesh, line 154: Add same details xml comments here as well? Is this even settable or should one set this value in Twin at all?

shared/src/Twin.cs
```diff
@@ -151,6 +151,19 @@ public Twin(TwinProperties twinProperties)
...
+        /// </summary>
+        [DefaultValue(null)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public string DeviceScope { get; internal set; }
+
+        /// <summary>
+        /// The scopes of the upper level edge devices if applicable. Only available for edge devices.
+        /// </summary>
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public virtual IReadOnlyList<string> ParentScopes { get; internal set; } = new List<string>();
```
vinagesh, line 165: This is a ReadOnly list. Any use of assigning the default value? I guess it will either be de-serialized or not based on weather the service returns it or not right?

vinagesh: Can this be set by users or is it just something the service auto-popylates?

iothub/service/src/Device.cs
drwill-ms: Yes, that is the common term for a non-edge device.

vinagesh (approved):

shared/src/Twin.cs
drwill-ms: As far as I can tell, you can set it and it does nothing. So I made these two properties readonly for customers. I decided not to copy over all the doc comments from Device.cs because most of it is about how you set it. Honestly, I don't understand why it is duplicated in the Twin, and I may end up removing it based on feedback from Anca.

shared/src/Twin.cs
drwill-ms: I think you are right. I'll remove the initialization.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
drwill-ms: I think it depends on what is easiest to read. Also, generally string concatenation is okay if kept at or below 3 concats.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -42,24 +41,42 @@ public async Task RegistryManager_BadProxy_ThrowsException()
...
+                leafDevice.Scope.Should().Be(edgeDevice2.Scope, "The parent scope should be respected as set.");
```
vishnureddy17, line 71: Shouldn't it be "The device scope should be respected as set"?

drwill-ms: totally. Thanks for catching.

barustum (approved):

iothub/service/src/Device.cs
```diff
@@ -109,9 +110,29 @@ public Device(string id)
         public virtual DeviceCapabilities Capabilities { get; set; }
 
         /// <summary>
-        /// Scope to which this device instance belongs to
+        /// The scope of the device. For edge devices, this is auto-generated and immutable. For leaf devices, set this to create child/parent
```
drwill-ms, line 113: I tweaked the comments from swagger here.

iothub/service/src/Device.cs
```diff
@@ -109,9 +110,29 @@ public Device(string id)
...
+        /// relationship.
         /// </summary>
+        /// <remarks>
+        /// For edge devices, the auto-generated device scope consists of a prefix "ms-azure-iot-edge://", the name of the parent edge device, a dash, and the
```
drwill-ms, line 117: I added this extra context to help users understand what this value should look like, and to know how do the right thing.

iothub/service/src/Device.cs
```diff
@@ -109,9 +110,29 @@ public Device(string id)
...
+        /// For leaf devices, the value to set a parent edge device can be retrieved from the parent edge device's <see cref="Scope"/> property.
+        /// </remarks>
         [JsonProperty(PropertyName = "deviceScope", NullValueHandling = NullValueHandling.Ignore)]
         public virtual string Scope { get; set; }
+
+        /// <summary>
+        /// The scopes of the upper level edge devices if applicable.
+        /// </summary>
+        /// <remarks>
+        /// For edge devices, the value to set a parent edge device can be retrieved from the parent edge device's <see cref="Scope"/> property.
```
drwill-ms, line 129: Same here: custom comments based on learnings from Anca, her document, and my experimentation.

shared/src/Twin.cs
```diff
@@ -151,6 +151,19 @@ public Twin(TwinProperties twinProperties)
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
         public X509Thumbprint X509Thumbprint { get; internal set; }
 
+        /// <summary>
+        /// The scope of the device. Auto-generated and immutable for edge devices and modifiable in leaf devices to create child/parent relationship.
+        /// </summary>
+        [DefaultValue(null)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public string DeviceScope { get; internal set; }
```
drwill-ms, line 159: I noticed these values are available in the twin too, so I added them here. However, note this property is read-only and `ParentScopes` I used `IReadOnlyList`. This effectively makes any values deserialized as read-only, as my experimentation showed that changing the values here had no affect.

I chose to make the values visible here so if someone had the twin (since using `RegistryManager` they can call `Query()` to get twins but cannot enumerate devices. If someone just wanted to draw a picture of device hierarchy, they should be able to do that just by enumerating twins, and not having to call `GetDeviceAsync()` on each device Id.

iothub/service/src/Device.cs
```diff
@@ -109,9 +110,26 @@ public Device(string id)
         public virtual DeviceCapabilities Capabilities { get; set; }
 
         /// <summary>
-        /// Scope to which this device instance belongs to
+        /// The scope of the device. For edge devices, this is auto-generated and immutable. For leaf devices, set this to create child/parent
+        /// relationship.
         /// </summary>
+        /// <remarks>
+        /// For leaf devices, the value to set a parent edge device can be retrieved from the parent edge device's <see cref="Scope"/> property.
```
drwill-ms, line 117: I added this extra context to help users understand how do the right thing.

shared/src/Twin.cs
ancaantochi, line 159: maybe use Scope as the property used in Device to make it clear is same property

ancaantochi (approved):

shared/src/Twin.cs
drwill-ms: Thanks for giving it this level of attention, @Ancaantochi. I had thought about that. Here is what I thought about; maybe you can consider that and weigh in again.

I think someone chose to just name it `Device.Scope` because `Device.DeviceScope` sounds repetitive. We also did that for `Device.Id` even though the json property name is `deviceId`.

However, when we get to `Twin.Scope`, I felt it was even more misleading than `Twin.DeviceScope`. I acknowledge it is named differently than in the Device class, but I hoped the comments would help settle that.

I think there is a pro/con to each approach. Given that, what do you think?

ancaantochi: ok, I think you're right, in Device it uses Scope because it's already a device property, using DeviceScope in twin is more clear.

PR #1893, opened 2021-04-19
doc(tpm): Add doc comments about TPM exceptions thrown by the client

abhipsaMisra:


security/tpm/src/SecurityProviderTpmHsm.cs
```diff
@@ -34,12 +34,20 @@ public class SecurityProviderTpmHsm : SecurityProviderTpm
         /// <summary>
         /// Initializes a new instance of the SecurityProviderTpmHsm class using the system TPM.
         /// </summary>
+        /// <remarks>
+        /// Calls to the TPM library can potentially return a <see cref="TssException"/> or a <see cref="TpmException"/>
+        /// if your TPM hardware does not support the relevant API call being carried out.
```
drwill-ms, line 39: you can leave off "being carried out" and it is functionally equivalent

PR #1898, opened 2021-04-20
Bring updates from master to previews/net5

abhipsaMisra:


timtay-microsoft (approved):

security/tpm/src/SecurityProviderTpmHsm.cs
```diff
@@ -34,12 +34,20 @@ public class SecurityProviderTpmHsm : SecurityProviderTpm
         /// <summary>
         /// Initializes a new instance of the SecurityProviderTpmHsm class using the system TPM.
         /// </summary>
+        /// <remarks>
```
jamdavi, line 37: Two things about the code comments here

1. We are repeating the same thing a lot
2. The exceptions should be in their own <exception> block

We can avoid having multiple copies and use <inheritdoc/>

Example:
```c#
/// <summary>
/// Gets the Base64 encoded StorageRootKey.
/// </summary>
/// <remarks>
/// Calls to the TPM library can potentially return a <see cref="TssException"/> or a <see cref="TpmException"/>
/// if your TPM hardware does not support the relevant API call.
/// </remarks>
/// <exception cref="TssException">Thrown when something bad happens with the TPM Software Stack</exception>
/// <exception cref="TpmException">Thrown when something bad happens with the TPM Hardware</exception>
/// <returns>Base64 encoded SRK.</returns>
public override byte[] GetStorageRootKey()
{ ... }

/// <summary>
/// Initializes a new instance of the SecurityProviderTpmHsm class using the specified TPM module.
/// </summary>
/// <param name="registrationId">The Device Provisioning Service Registration Id.</param>
/// <param name="tpm">The TPM device.</param>
/// <inheritdoc cref="GetStorageRootKey" path="/exception"/>
/// <inheritdoc cref="GetStorageRootKey" path="/remarks"/>
public SecurityProviderTpmHsm(string registrationId) : this(registrationId, CreateDefaultTpm2Device()) { }


/// <summary>
/// Gets the Base64 encoded EndorsmentKey.
/// </summary>
/// <returns>Base64 encoded EK.</returns>
/// <inheritdoc cref="GetStorageRootKey" path="/exception['TssException']"/>
/// <inheritdoc cref="GetStorageRootKey" path="/remarks"/>
public override byte[] GetEndorsementKey()
{ ... }
```

PR #1899, opened 2021-04-20
Documentation and changes to the surface

jamdavi:
* Updated documentation
* Refactored the CheckType name to be more descriptive
* Added logic to correctly convert from JObject to whatever is being used by the custom serializer.


iothub/device/samples/PnpDeviceSamples/TemperatureController/CustomJsonSerializer.cs
```diff
@@ -27,9 +27,14 @@ public override T DeserializeToType<T>(string stringToDeserialize)
             return JsonSerializer.Deserialize<T>(stringToDeserialize, s_options);
         }
 
-        public override bool CheckType(object typeToCheck)
+        public override bool CheckWritablePropertyType(object typeToCheck)
```
abhipsaMisra, line 30: CheckWritablePropertyResponseType?

iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -88,7 +88,9 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
 
             var t = await _deviceClient.GetTwinAsync(cancellationToken);
 
-            Properties properties = await _deviceClient.GetPropertiesAsync(cancellationToken);
+            Properties properties = await _deviceClient.GetPropertiesAsync(s_payloadConvention, cancellationToken);
+
+            var sa = properties.Get<string>("");
```
abhipsaMisra, line 93: we can get rid of this line :D

iothub/device/src/DefaultPayloadConvention.cs
```diff
@@ -4,34 +4,32 @@
...
 {
     /// <summary>
-    ///
+    /// The default implementation of the <see cref="IPayloadConvention"/> class.
     /// </summary>
+    /// <remarks>
+    /// This class is the default <see cref="IPayloadConvention"/> that will be used for all <see cref="PayloadCollection"/> implementations. This class makes use of the <see cref="JsonContentSerializer"/> serializer and the <see cref="Utf8ContentEncoder"/> unless another implementation is specified for these.
+    /// </remarks>
+    /// <example>
+    /// You can overrride the default instances of both the <see cref="PayloadSerializer"/> and <see cref="PayloadEncoder"/> classes.
```
abhipsaMisra, line 13: this isn't possible, right? the serializer and encoder only have getters

iothub/device/src/DefaultPayloadConvention.cs
```diff
@@ -4,34 +4,32 @@
...
+    /// <code lang='C#'>DefaultPayloadConvention.Instance.PayloadEncoder = new CustomPayloadEncoder();
+    /// DefaultPayloadConvention.Instance.PayloadSerializer = new CustomPayloadSerializer();
+    /// </code>
+    /// </example>
     public class DefaultPayloadConvention : IPayloadConvention
```
abhipsaMisra, line 18: we should make this sealed, there is no reason for anyone to inherit from this

iothub/device/src/DefaultPayloadConvention.cs
```diff
@@ -4,34 +4,32 @@
...
+        /// The payload encoder.
         /// </summary>
-        public override IContentEncoder PayloadEncoder => Utf8ContentEncoder.Instance;
-
-        /// <summary>
-        ///
-        /// </summary>
-        /// <param name="objectToSendWithConvention"></param>
-        /// <returns></returns>
-        public override byte[] GetObjectBytes(object objectToSendWithConvention)
```
abhipsaMisra, line 31: removed?

iothub/device/src/IPayloadConvention.cs
```diff
@@ -4,25 +4,30 @@
...
     /// <summary>
-    ///
+    /// The payload convention class.
     /// </summary>
+    /// <remarks>The payload convention is used to define a specific serializer as well as a specific content encoding. For example, IoT has a <see href="https://docs.microsoft.com/en-us/azure/iot-pnp/concepts-convention">convention</see> that is designed to make it easier to get started with products that use specific conventions by default.</remarks>
     public abstract class IPayloadConvention
     {
         /// <summary>
-        ///
+        /// The serializer used for the payload of the <see cref="Message"/>.
```
abhipsaMisra, line 13: we shouldn't reference message here, for purposes of pnp Message is now an internal implementation detail

iothub/device/src/DefaultPayloadConvention.cs
abhipsaMisra: default impl in the base abstract class, got it

iothub/device/src/PayloadCollection.cs
```diff
@@ -1,52 +1,106 @@
...
         /// </summary>
-        /// <returns></returns>
+        /// <remarks>
+        /// This will get the fully encoded serialized string using both <see cref="ISerializer.SerializeToString(object)"/> and <see cref="IContentEncoder.EncodeStringToByteArray(string)"/> methods implemented in the <see cref="IPayloadConvention"/>.
+        /// </remarks>
+        /// <returns>A fully encoded serialized string.</returns>
         public virtual byte[] GetPayloadObjectBytes()
         {
-            return Convention.GetObjectBytes(Collection);
+            return Convention.GetObjectBytes(Convention.PayloadSerializer.SerializeToString(Collection));
```
abhipsaMisra, line 56: GetObjectBytes takes in an object, so we can pass in the Collection directly

iothub/device/src/Properties.cs
```diff
@@ -5,16 +5,20 @@
...
 using System.Collections.Generic;
 using Microsoft.Azure.Devices.Shared;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// A container for properties.
+    /// <remarks>
+    /// The Properties class is not meant to be constructed by the customer code is meant to be returned from the <see cref="DeviceClient.GetPropertiesAsync(IPayloadConvention, System.Threading.CancellationToken)"/> method.
```
abhipsaMisra, line 15: punctuation mark seems to be missing in this statement :D

iothub/device/src/PropertyCollection.cs
```diff
@@ -123,81 +137,62 @@ private void Add(IDictionary<string, object> properties, string componentName =
...
+        /// <remarks>This internala class is aware of the implemention of the TwinCollection ad will </remarks>
+        internal static PropertyCollection FromTwinCollection(TwinCollection twinCollection, IPayloadConvention payloadConvention = default)
         {
             if (twinCollection == null)
             {
                 throw new ArgumentNullException(nameof(twinCollection));
             }
-
+            
             var writablePropertyCollection = new PropertyCollection();
```
abhipsaMisra, line 186: payloadConvention to be passed to PropertyCollection constructor

iothub/device/src/Properties.cs
```diff
@@ -82,32 +89,48 @@ public IEnumerator<object> GetEnumerator()
...
+        /// <returns>A new instance of of the class from an existing <see cref="TwinProperties"/> using an optional <see cref="IPayloadConvention"/>.</returns>
+        internal static Properties FromTwinProperties(TwinProperties twinProperties, IPayloadConvention payloadConvention = default)
         {
             if (twinProperties == null)
             {
                 throw new ArgumentNullException(nameof(twinProperties));
             }
 
-            var writablePropertyCollection = new PropertyCollection();
+            var writablePropertyCollection = PropertyCollection.FromTwinCollection(twinProperties.Desired, payloadConvention);
```
abhipsaMisra, line 125: since we are creating the property collection from the .FromTwinCollection static helper, we don't need to iterate over the properties and add them in manually anymore

iothub/device/src/DefaultPayloadConvention.cs
jamdavi: Correct, will change

iothub/device/src/DeviceClient.cs
```diff
@@ -707,10 +707,11 @@ protected virtual void Dispose(bool disposing)
         /// <summary>
         /// Retrieve the device properties.
         /// </summary>
+        /// <param name="payloadConvention">A convention handler that defines the content encoding and serializer to use for commands.</param>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <returns>The device properties.</returns>
-        public Task<Properties> GetPropertiesAsync(CancellationToken cancellationToken)
-            => InternalClient.GetPropertiesAsync(cancellationToken);
+        public Task<Properties> GetPropertiesAsync(IPayloadConvention payloadConvention = default, CancellationToken cancellationToken = default)
```
jamdavi, line 713: Should we reverse the order of the parameters here?

Below we have payload convention coming before the cancellation token, but it might be more natural to write
```c#
// With them reversed
GetPropertiesAsync();
GetPropertiesAsync(cts);
GetPropertiesAsync(cts, convention);

// With them as they are
GetPropertiesAsync(convention, cts);
GetPropertiesAsync(null, cts);
GetPropertiesAsync(cancellationToken: cts);
```

iothub/device/src/Properties.cs
jamdavi: Yeah that is missing a word or two too :)

iothub/device/src/DeviceClient.cs
abhipsaMisra: In my mind I always put cancellation token at the end of any parameter list (baring an actual `params` being passed), so I like it the way we have it here. I'd be comfortable with `GetPropertiesAsync(cancellationToken: cts);`, but this is definitely a personal preference.

iothub/device/src/Common/CommonConstants.cs
```diff
@@ -85,6 +85,10 @@ internal static class CommonConstants
         // Security message constants
         public const string SecurityMessageInterfaceId = "urn:azureiot:Security:SecurityAgent:1";
 
+        // Component constants
```
abhipsaMisra, line 88: nit: this is more impl thing but we should call out that these ae identifiers for property constants used for pnp based device property patches

abhipsaMisra (approved):

PR #1905, opened 2021-04-22
updates to impl

abhipsaMisra:
- decouple twin from properties
- WritablePropertyBase properties being serialized twice


iothub/device/samples/PnpDeviceSamples/TemperatureController/CustomJsonSerializer.cs
```diff
@@ -27,11 +27,6 @@ public override T DeserializeToType<T>(string stringToDeserialize)
             return JsonSerializer.Deserialize<T>(stringToDeserialize, s_options);
         }
 
-        public override bool CheckWritablePropertyResponseType(object typeToCheck)
```
jamdavi, line 30: I think I'm okay with removing the base class and this check method. But we will need to make sure that our documentation is crystal clear around the WritablePropertyResponse.

iothub/device/src/WritablePropertyResponse.cs
```diff
@@ -6,30 +6,72 @@
...
+        public const string AckVersionPropertyName = "av";
+
+        /// <summary>
+        /// Represents the JSON document property name for the Ack Description
+        /// </summary>
+        public const string AckDescriptionPropertyName = "ad";
+
+        /// <summary>
+        /// Convenience constructor for specifying the properties.
+        /// </summary>
```
jamdavi, line 40: Add an <example> tag with a <code> block to show what the System.Text.Json version would look like. This will make sure that if someone comes looking at this they will understand the class is used in the convention and that it will need to be properly serailized.

iothub/device/src/WritablePropertyResponse.cs
```diff
@@ -6,30 +6,72 @@
...
+    /// </remarks>
+    public sealed class WritablePropertyResponse
     {
-        /// <inheritdoc />
-        public WritablePropertyResponse(object propertyValue, int ackCode, long ackVersion, string ackDescription = default)            
-            : base(propertyValue, ackCode, ackVersion, ackDescription)
+        /// <summary>
+        /// Represents the JSON document property name for the value
+        /// </summary>
+        public const string ValuePropertyName = "value";
```
jamdavi, line 21: Move these constants to CommonConstants or some other place. 

We should possibly create a "ConventionConstants" container which will tie this together?

jamdavi (approved):
I like it, I do have some feedback around documentation but it doesn't need to be in this change. I can take that on as well.

iothub/device/src/PropertyCollection.cs
```diff
@@ -85,11 +88,12 @@ private void Add(IDictionary<string, object> properties, string componentName =
             {
                 foreach (KeyValuePair<string, object> entry in properties)
                 {
-                    var checkType = entry.Value is WritablePropertyBase;
-                    if (entry.Value is WritablePropertyBase && !Convention.PayloadSerializer.CheckWritablePropertyResponseType(entry.Value))
+                    bool checkType = entry.Value is IWritablePropertyResponse;
```
jamdavi, line 91: We can remove this line

iothub/device/src/PropertyCollection.cs
```diff
@@ -112,10 +116,10 @@ private void Add(IDictionary<string, object> properties, string componentName =
                 }
                 foreach (KeyValuePair<string, object> entry in properties)
                 {
-                    var checkType = entry.Value is WritablePropertyBase;
-                    if (entry.Value is WritablePropertyBase && !Convention.PayloadSerializer.CheckWritablePropertyResponseType(entry.Value))
+                    bool checkType = entry.Value is IWritablePropertyResponse;
```
jamdavi, line 119: and this one

abhipsaMisra: it is repeated in the statement below, true

PR #1906, opened 2021-04-23
use sha256 thumbprint in e2e setup

wpbrown:
IoT Hub Gateway V2 requires sha256 thumbprint with API versions
2017-06-30 and later. this fixes X509 tests with Gateway V2.

## Checklist

## Description of the changes

Use a SHA256 thumbprint when creating test devices.

OneBranch Bug #9790227

### DIFF
--- a/e2e/test/Helpers/TestDevice.cs
+++ b/e2e/test/Helpers/TestDevice.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client;
@@ -83,7 +84,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
                 {
                     X509Thumbprint = new X509Thumbprint
                     {
-                        PrimaryThumbprint = Configuration.IoTHub.GetCertificateWithPrivateKey().Thumbprint
+                        PrimaryThumbprint = Configuration.IoTHub.GetCertificateWithPrivateKey().GetCertHashString(HashAlgorithmName.SHA256)
                     }
                 };
 
--- a/e2e/test/Helpers/TestDevice.cs
+++ b/e2e/test/Helpers/TestDevice.cs
@@ -80,14 +80,18 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
             var requestDevice = new Device(deviceName);
             if (type == TestDeviceType.X509)
             {
-                requestDevice.Authentication = new AuthenticationMechanism
-                {
-                    X509Thumbprint = new X509Thumbprint
+                using (var hasher = SHA256.Create())
+                {  
+                    var hash = hasher.ComputeHash(Configuration.IoTHub.GetCertificateWithPrivateKey().RawData);
+                    requestDevice.Authentication = new AuthenticationMechanism
                     {
-                        PrimaryThumbprint = Configuration.IoTHub.GetCertificateWithPrivateKey().GetCertHashString(HashAlgorithmName.SHA256)
-                    }
-                };
-
+                        X509Thumbprint = new X509Thumbprint
+                        {
+                            PrimaryThumbprint = BitConverter.ToString(hash).Replace("-", "")
+                        }
+                    }; 
+                }
+                
                 auth = new DeviceAuthenticationWithX509Certificate(deviceName, Configuration.IoTHub.GetCertificateWithPrivateKey());
             }
 


abhipsaMisra (changes requested):
This doesn't really fix the issue referenced, since the certificate referenced at `Configuration.IoTHub.GetCertificateWithPrivateKey()` could still be a sha-1 signed certificate. 
Also, the code is now initializing the thumbprint with the entire hashed value of the certificate.

In order to address the requirement for certificates to be SHA256 signed, our test pre-requisite script handles this during certificate generation: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1#L269

I believe this is what you are looking for. Let me know if I misunderstood what you are trying to achieve.

wpbrown:
> This doesn't really fix the issue referenced, since the certificate referenced at `Configuration.IoTHub.GetCertificateWithPrivateKey()` could still be a sha-1 signed certificate.

This change in our API isn't about the certificate signing algorithm, it's strictly about the thumbprint. 

> Also, the code is now initializing the thumbprint with the entire hashed value of the certificate.

That [is what a thumbprint is](https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate2.cs#L366). `Thumbprint` == `GetCertHashString(HashAlgorithmName.SHA1);`

> In order to address the requirement for certificates to be SHA256 signed, our test pre-requisite script handles this during certificate generation: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1#L269
> 
> I believe this is what you are looking for. Let me know if I misunderstood what you are trying to achieve.

The certificate generation is good. The problem is just the thumbprint of the certificate. You can reproduce the issue by creating a V2 preview hub. Creating devices for the X509 tests will fail without this patch.

abhipsaMisra:
Thank you for the explanation, I was not aware of the internal working of thumbprint.
Just so I understand this correctly, even with the change that you've proposed here, it would need to go hand-in-hand with the actual certificate referenced being sha256 signed, correct?

Two open issues that need to be addressed before this can be merged in:
- `HashAlgorithmName` struct isn't available in `net451`, which is one of the platforms that we need to test against.
- `GetCertHashString(HashAlgorithmName)` is available only for ns2.1, whereas we support ns2.0 and some ns1.4 based tfm as well. 

While we can still make it work with if-directives, the tests will still retrieve SHA1 hash for the above frameworks, and thus fail with gateway v2. Does gateway v2 have a way to handle these scenarios?

wpbrown:
> Just so I understand this correctly, even with the change that you've proposed here, it would need to go hand-in-hand with the actual certificate referenced being sha256 signed, correct?

No the thumbprint is completely independent of cert signing hash algorithm. They do not need to be the same algorithm.

> 
> Two open issues that need to be addressed before this can be merged in:
> 
> * `HashAlgorithmName` struct isn't available in `net451`, which is one of the platforms that we need to test against.
> * `GetCertHashString(HashAlgorithmName)` is available only for ns2.1, whereas we support ns2.0 and some ns1.3 based tfm as well.

Oof. net451/ns1.3 have [SHA256](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha256?view=netframework-4.5.1) so I guess we need a DIY function? 

> While we can still make it work with if-directives, the tests will still retrieve SHA1 hash (current impl) for the above frameworks, and thus fail with gateway v2. Does gateway v2 have a way to handle these scenarios?

Gateway V2 will only accept a SHA1 hash with API version <`2017-06-30`, so probably simpler to do the DIY thumbprint.

wpbrown:
Ok I updated. I believe this is `net451` compatible (sorry I'm on Linux).

abhipsaMisra:
Thanks for adding the implementation @wpbrown !
Could you add some comments to explain - 
- the code that you've added? (the hash computation, string replace in the computed hash, etc.)
- thumbprint returns a SHA1 hash of the certificate by default, and this implementation overrides that behavior.
- thumbprint is a hash of the entire certificate and is independent of the signature algorithm used in certificate generation.

abhipsaMisra:
@wpbrown Were you able to add the comments are requested above?

wpbrown:
After further discussion we are going to revert this requirement in V2. I recommend you apply the patch anyway so that you test with the secure path that customers should be using, but it is not going to be _required_ in the foreseeable future.

PR #1908, opened 2021-04-24
add snippets for no-component and component scenario

abhipsaMisra:


iothub/device/samples/PnpDeviceSamples/TemperatureController/SimpleTemperatureControllerSampleNew.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    public class SimpleTemperatureControllerSampleNew
+    {
+        private static readonly Random s_random = new();
+        private static readonly PayloadConvention s_systemTextJsonPayloadConvention = new SystemTextJsonPayloadConvention();
```
jamdavi, line 16: To make it simple we should probably not show how to use the convention, we have other samples that do that.

abhipsaMisra: good point, I'll remove it for the "simple" sample

jamdavi (approved):

PR #1914, opened 2021-04-26
Update nuget versions for release

drwill-ms:


abhipsaMisra (approved):
I got a bit confused not seeing the `versions.csv` file in here, but turns out that has been updated already!

PR #1915, opened 2021-04-26
feat(registry-manager): Add support to use user defined msi for import export jobs.

vinagesh:
## Checklist

## Description of the changes
This contains the implementation to support using configured user defined managed identity to perform import and export jobs. What this means is the hub itself will use the passed in managed identity (in JobProperties) to perform operations on the storage account used during import and export.

This feature is still under a ff and only works on a specific subscription. I've tested these changes against that and pushing the code to a feature branch. I have not turned on any gates for this branch as the feature is not yet available in canary. Once it is available, I will create the required assets and add the gating.

Details about where the feature is enabled and how to use it:
https://review.docs.microsoft.com/en-us/azure/iot-hub/iot-hub-managed-identity?branch=pr-en-us-154985

Use following subscriptions for creating the hub. 
	1.	DEVICEHUB_DEV1 [91d12660-3dec-467a-be2a-213b5544ddc0] 
	2.	IOT_PLATFORM_UPX_TEST --> portal team test sub

Use this link for your portal access https://ms.portal.azure.com/?feature.canmodifystamps=true&Microsoft_Azure_Iothub=tip1

### DIFF
--- a/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
+++ b/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
@@ -46,9 +46,10 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
         [TestCategory("LongRunning")]
         [Timeout(120000)]
         [DoNotParallelize]
-        [DataRow(StorageAuthenticationType.KeyBased)]
-        [DataRow(StorageAuthenticationType.IdentityBased)]
-        public async Task RegistryManager_ExportDevices(StorageAuthenticationType storageAuthenticationType)
+        [DataRow(StorageAuthenticationType.KeyBased, false)]
+        [DataRow(StorageAuthenticationType.IdentityBased, false)]
+        [DataRow(StorageAuthenticationType.IdentityBased, true)]
+        public async Task RegistryManager_ExportDevices(StorageAuthenticationType storageAuthenticationType, bool isUserAssignedMsi)
         {
             // arrange
 
@@ -86,13 +87,24 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
                 {
                     try
                     {
+                        ManagedIdentity identity = null;
+                        if (isUserAssignedMsi)
+                        {
+                            string userAssignedMsiResourceId = Configuration.IoTHub.UserAssignedMsiResourceId;
+                            identity = new ManagedIdentity
+                            {
+                                userAssignedIdentity = userAssignedMsiResourceId
+                            };
+                        }
+
                         exportJobResponse = await registryManager
                            .ExportDevicesAsync(
                                 JobProperties.CreateForExportJob(
                                     containerUri.ToString(),
                                     true,
                                     null,
-                                    storageAuthenticationType))
+                                    storageAuthenticationType,
+                                    identity))
                            .ConfigureAwait(false);
                         break;
                     }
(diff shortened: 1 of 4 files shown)


iothub/service/src/ManagedIdentity.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// The managed identity used to access the storage account for IoT hub import and export jobs.
+    /// TODO link from service team: For more information, see <see href=""/>
```
timtay-microsoft, line 12: Still TODO, or do you have a link for this?

vinagesh: Still a TODO. This is their current doc, that still needs a review and by the url, seems like it is not the final link - https://review.docs.microsoft.com/en-us/azure/iot-hub/iot-hub-managed-identity?branch=pr-en-us-154985

iothub/service/src/JobProperties.cs
```diff
@@ -118,6 +118,13 @@ public JobProperties()
         [JsonProperty(PropertyName = "storageAuthenticationType", NullValueHandling = NullValueHandling.Ignore)]
         public StorageAuthenticationType? StorageAuthenticationType { get; set; }
 
+        /// <summary>
+        /// The managed identity used to access the storage account for import and export jobs.
```
vinagesh, line 122: @nimengan - Here is the description I added. We should keep this consistent across SDKs and in the swagger. At the moment we don't have any description for this in the swagger.

iothub/service/src/ManagedIdentity.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// The managed identity used to access the storage account for IoT hub import and export jobs.
```
vinagesh, line 11: @nimengan - Here is the description I added. We should keep this consistent across SDKs and in the swagger. At the moment we don't have any description for this in the swagger.

vishnureddy17: I think we should say something like "the resource ID for the managed identity..." just to make it more clear.

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
```diff
@@ -86,13 +87,24 @@ await registryManager
                 {
                     try
                     {
+                        ManagedIdentity identity = null;
+                        if (isUserAssignedMsi)
+                        {
+                            var userAssignedMsiResourceId = Configuration.IoTHub.UserAssignedMsiResourceId;
```
barustum, line 93: use explicit type instead of var.

iothub/service/src/JobProperties.cs
```diff
@@ -118,6 +118,13 @@ public JobProperties()
...
+        /// TODO link from service team: For more information, see <see href=""/>
```
barustum, line 123: TODO still intentional?

iothub/service/src/ManagedIdentity.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
```
barustum, line 3: nit :): new line after copyright header

iothub/service/src/JobProperties.cs
vinagesh: yes same explanation as above.

barustum (approved):

iothub/service/src/ManagedIdentity.cs
vinagesh: I added that below for the actual "userAssignedIdentity" property. This class might contain other things as designed by the service team and hence I kept it a little more generic. Open to other suggestions.

PR #1916, opened 2021-04-26
rename types to specify serializer used

abhipsaMisra:


iothub/device/samples/PnpDeviceSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -87,7 +87,7 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
 
             var t = await _deviceClient.GetTwinAsync(s_cancellationToken);
 
-            Properties properties = await _deviceClient.GetPropertiesAsync(s_payloadConvention, s_cancellationToken);
```
abhipsaMisra, line 90: Re-reading "SubscribeToWritablePropertyEventAsync", I am now reading this a "subscribe to writable-type-of-property event, so property = singular; what do you think?

jamdavi: The way I see it is "Subscribe To WritableProperty Event" 

Where we are subscribing to one WritableProperty Event. But since this is the global handler there will always only be one event for all **properties**

jamdavi (approved):
I think since we already have the SetDesiredPropertyHandler we don't need to rename the WritableProperty Event method. We should start another PR with that so we can get a good discussion.

PR #1918, opened 2021-04-27
remove custom serializer from simple samples

abhipsaMisra:


iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
```diff
@@ -39,19 +38,25 @@ public ComponentTemperatureControllerSampleNew(DeviceClient deviceClient, ILogge
         public async Task PerformOperationsAsync(CancellationToken cancellationToken)
```
abhipsaMisra, line 38: this sample now uses all complex objects

iothub/device/samples/PnpDeviceSamples/TemperatureController/SimpleTemperaturControllerSampleOld.cs
```diff
@@ -62,34 +62,34 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
                 ContentEncoding = "utf-8",
-                ContentType= "application/json",
+                ContentType = "application/json",
             };
 
             await _deviceClient.SendEventAsync(message, cancellationToken);
             _logger.LogDebug($"Telemetry: Sent - {JsonConvert.SerializeObject(telemtry)} in KB.");
 
-            // Subscribe and respond to event for writable property "temperatureRange".
+            // Subscribe and respond to event for writable property "targetHumidity".
```
abhipsaMisra, line 71: added a new "targetHumidity" property on the root interface

jamdavi (approved):

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleOld.cs
```diff
@@ -50,20 +48,23 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
                 Humidity = 20,
                 Temperature = 25
             };
+
             if (!properties.Properties.Reported.Contains(Thermostat2)
-                || ((JObject)properties.Properties.Reported[Thermostat2]).Value<ThermostatInitialValue>("initialValue") != initialValue)
+                || !((JObject)properties.Properties.Reported[Thermostat2]).TryGetValue("initialValue", out JToken initialValueToken)
+                || !initialValue.Equals(System.Text.Json.JsonSerializer.Deserialize<ThermostatInitialValue>(initialValueToken.ToString())))
```
abhipsaMisra, line 54: ThermostatInitialValue is based on System.Text.Json, so updated the serializer

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleOld.cs
```diff
@@ -50,20 +48,23 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
                     ["initialValue"] = initialValue
                 };
-                var componentProperty = new TwinCollection
+                var componentProperty = new Dictionary<string, object>
                 {
                     [Thermostat2] = propertiesToBeUpdated
                 };
-                await _deviceClient.UpdateReportedPropertiesAsync(propertiesToBeUpdated, cancellationToken);
-                _logger.LogDebug($"Property: Update - {propertiesToBeUpdated.ToJson()}.");
+                var twinCollection = new TwinCollection(System.Text.Json.JsonSerializer.Serialize(componentProperty));
```
abhipsaMisra, line 65: twin collection will inherently try to serialize the object based on newtonsoft. Since our type uses System.Text.Json, we will create the TwinCollection using the serialized string directly.

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleOld.cs
```diff
@@ -86,39 +87,40 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
                     return;
                 }
 
-                HumidityRange targetHumidityRange = ((JObject)desired[Thermostat1]).Value<HumidityRange>(propertyName);
+                HumidityRange targetHumidityRange = System.Text.Json.JsonSerializer.Deserialize<HumidityRange>(humidityRangeToken.ToString());
 
-                var propertyPatch = new TwinCollection();
-                var componentPatch = new TwinCollection()
+                var propertyPatch = new Dictionary<string, object>();
+                var componentPatch = new Dictionary<string, object>
```
jamdavi, line 105: Why remove the TwinCollection? Did it fail? This is in our docs.

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleOld.cs
```diff
@@ -86,39 +87,40 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
                 {
                     ["__t"] = "c"
                 };
-                var temperatureUpdateResponse = new TwinCollection
+                var temperatureUpdateResponse = new
```
jamdavi, line 109: Same as here. I opted to not use an anonymous class since we had specified TwinCollection

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleOld.cs
abhipsaMisra: Using TwinCollection as the type to hold the property value will work as expected as long as the "value" is either a primitive type or a newtonsoft.json annotated type. This is because TwinCollection will use newtonsoft to serialize the object.
However, if the "value" is type that uses a different serializer, but we initialize it to a TwinCollection directly, the resultant property patch will no longer have the correct json property keys. For this reason, we need to serialize the object based on custom serializer and then feed the serialized string to TwinCollection. This serialized string can be internally represented as a JObject (or any other newtonsoft type) as required, it won't have any additional negative impact.

abhipsaMisra: This results in:
```json
{{
  "thermostat1": {
    "__t": "c",
    "humidityRange": {
      "value": {
        "MaxTemperature": 50.0,
        "MinTemperature": 20.0
      },
      "ac": 200,
      "av": 8,
      "ad": "The operation completed successfully."
    }
  }
}}
```
instead of
```json
{{
  "thermostat1": {
    "__t": "c",
    "humidityRange": {
      "value": {
        "maxHumidity": 50,
        "minHumidity": 20
      },
      "ac": 200,
      "av": 9,
      "ad": "The operation completed successfully."
    }
  }
}}
```

PR #1922, opened 2021-04-28
Adding support for AAD in our client - Merge previews/RBAC to master

vinagesh:
## Checklist


e2e/test/config/Configuration.IoTHub.cs
```diff
@@ -14,6 +26,34 @@ public static partial class IoTHub
...
+            {
+                ConnectionStringParser connectionString = new ConnectionStringParser(ConnectionString);
+                return connectionString.IotHubHostName;
+            }
+
+#if !NET451
+
+            public static ClientSecretCredential GetClientSecretCredential()
+            {
+                return new ClientSecretCredential(
```
drwill-ms, line 39: Should these environment variables be added to doc comments?

e2e/test/config/Configuration.IoTHub.cs
```diff
@@ -14,6 +26,34 @@ public static partial class IoTHub
...
+                    GetValue("MSFT_TENANT_ID"),
+                    GetValue("IOTHUB_CLIENT_ID"),
+                    GetValue("IOTHUB_CLIENT_SECRET"));
+            }
+
+            public static string GetIotHubSharedAccessSignature(TimeSpan timeToLive)
+            {
+                ConnectionStringParser connectionString = new ConnectionStringParser(ConnectionString);
```
drwill-ms, line 47: var

e2e/test/config/Configuration.IoTHub.cs
```diff
@@ -14,6 +26,34 @@ public static partial class IoTHub
             public static string ConnectionString => GetValue("IOTHUB_CONNECTION_STRING");
             public static string X509ChainDeviceName => GetValue("IOTHUB_X509_CHAIN_DEVICE_NAME");
 
+            public static string GetIotHubHostName()
+            {
+                ConnectionStringParser connectionString = new ConnectionStringParser(ConnectionString);
```
drwill-ms, line 31: var

e2e/test/config/Configuration.IoTHub.cs
```diff
@@ -65,9 +105,42 @@ public static X509Certificate2 GetIntermediate2Certificate()
...
+                {
+                    token += "&skn=" + policyName;
+                }
+
+                return token;
+            }
+
+#endif
+
+            public class ConnectionStringParser
```
drwill-ms, line 141: Is this class def public? Doc comments?

e2e/test/config/Configuration.IoTHub.cs
drwill-ms: I mistook this file for an SDK class, so never mind.

iothub/service/design.md
```diff
@@ -0,0 +1,119 @@
+# RBAC API Design for IoT hub
+
+## Service Client
```
drwill-ms, line 3: Should we remove this file? My impression is that we used them during design, and then removed them after impl. We wouldn't want to have to keep them in sync with impl long term, right?

iothub/service/src/AmqpServiceClient.cs
```diff
@@ -24,11 +24,11 @@ internal sealed class AmqpServiceClient : ServiceClient
         private const string PurgeMessageQueueFormat = "/devices/{0}/commands?" + ClientApiVersionHelper.ApiVersionQueryString;
         private const string DeviceMethodUriFormat = "/twins/{0}/methods?" + ClientApiVersionHelper.ApiVersionQueryString;
         private const string ModuleMethodUriFormat = "/twins/{0}/modules/{1}/methods?" + ClientApiVersionHelper.ApiVersionQueryString;
+        private const string _sendingPath = "/messages/deviceBound";
```
drwill-ms, line 27: Format for a const should be `SendingPath`

iothub/service/src/Common/TokenHelper.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// <summary>
+        /// Determines if the given token expiry date time is close to expiry. The date and time is
+        /// considered close to expiry if it has less than 10 minutes relative to the current time.
+        /// </summary>
+        /// <param name="expiry">The token expiration date and time.</param>
+        /// <returns>True if the token expiry has less than 10 minutes relative to the current time, otherwise false.</returns>
+        public static bool IsCloseToExpiry(DateTimeOffset expiry)
+        {
+            TimeSpan timeToExpiry = expiry - DateTimeOffset.UtcNow;
+            return timeToExpiry.TotalMinutes < 10;
```
drwill-ms, line 18: Interesting that this is hardcoded to 10 minutes.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -10,51 +10,44 @@
 using Microsoft.Azure.Devices.Generated;
 using Microsoft.Rest;
 using Newtonsoft.Json;
+using Microsoft.Azure.Devices.DigitalTwin.Authentication;
+using Azure;
```
drwill-ms, line 14: sort?

iothub/service/src/JobClient/JobClient.cs
```diff
@@ -41,6 +48,62 @@ public static JobClient CreateFromConnectionString(string connectionString, Http
...
+        /// <param name="transportSettings">The HTTP transport settings.</param>
+        /// <returns>An instance of <see cref="JobClient"/>.</returns>
+        public static JobClient Create(
+            string hostName,
+            TokenCredential credential,
+            HttpTransportSettings transportSettings = default)
+        {
+            if (string.IsNullOrEmpty(hostName))
+            {
+                throw new ArgumentNullException($"{nameof(hostName)},  Parameter cannot be null or empty");
```
drwill-ms, line 67: If you want to include a messag, please use the `ArgumentNullException(string, string)` overload. The first parameter is the null variable name, and the second is the message.

Otherwise the `ArgumentNullException(string)` overload just expects a parameter.

So either of these:
```chsarp
throw new ArgumentNullException(nameof(hostName));
// or
throw new ArgumentNullException(nameof(hostName), $"{Parameter cannot be null or empty");
```

I recommend the former if the message doesn't add anything. For instance, for `hostName` it also can't be empty, so the message adds something. For `credential`, I'd stick with the 1 param option.

iothub/service/src/RegistryManager.cs
```diff
@@ -46,6 +54,62 @@ public static RegistryManager CreateFromConnectionString(string connectionString
             return new HttpRegistryManager(iotHubConnectionString, transportSettings);
         }
 
+#if !NET451
+
+        /// <summary>
+        /// Creates an instance of <see cref="RegistryManager"/>.
```
drwill-ms, line 60: This generates a self-reference, meaning the link in the help page will take you to the same page you are on. I don't recommend adding these links.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -10,51 +10,44 @@
...
+using Azure.Core;
+using PnpDigitalTwin = Microsoft.Azure.Devices.Generated.DigitalTwin;
 
 namespace Microsoft.Azure.Devices
 {
     /// <summary>
     /// The Digital Twins Service Client contains methods to retrieve and update digital twin information, and invoke commands on a digital twin device.
+    /// For more information, see <see href="https://github.com/Azure/azure-iot-sdk-csharp#iot-hub-service-sdk"/>
     /// </summary>
     public class DigitalTwinClient : IDisposable
```
drwill-ms, line 24: Did we not need to regenerate to reference the latest API version?

iothub/service/design.md
timtay-microsoft: I'd like to see this removed as well

iothub/service/src/ServiceClient.cs
```diff
@@ -1,11 +1,20 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
timtay-microsoft, line 4: Double copyright header?

timtay-microsoft (approved):

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
vinagesh: This is not in the generated folder. Am I missing something?

iothub/service/src/Common/TokenHelper.cs
vinagesh: This is what we decided collectively across .NET and Java, given the jwt tokens by default expire in an hour.

iothub/service/design.md
vinagesh: yes i'll remove it.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
drwill-ms: I don't know if AAD auth is tied to an API version, and I wondered if it was necessary to increment it to latest.

I did it for the legacy clients for parentScopes, but didn't realize the DT client has it hardcoded from autogeneration. This was the closest file for me to ask this question.

iothub/service/src/JobClient/JobClient.cs
```diff
@@ -41,6 +48,62 @@ public static JobClient CreateFromConnectionString(string connectionString, Http
...
+        /// <param name="transportSettings">The HTTP transport settings.</param>
+        /// <returns>An instance of <see cref="JobClient"/>.</returns>
+        public static JobClient Create(
+            string hostName,
+            TokenCredential credential,
+            HttpTransportSettings transportSettings = default)
+        {
+            if (string.IsNullOrEmpty(hostName))
+            {
+                throw new ArgumentNullException($"{nameof(hostName)}");
```
drwill-ms, line 67: There's no reason to wrap it in a string. `nameof(hostname)` is a string, so just avoid the string interpolation.

drwill-ms (approved):

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
vinagesh: Got your question. Good question but all the tests passed with the current version. I am guessing it was not implemented for a specific API version. The API surface did not really change, only the token we send in the header is a different one so they must have been able to do it without changing the API version.

PR #1923, opened 2021-04-29
Adding PayloadConvention to ClientOptions

jamdavi:
Adds the PayloadConvention to ClientOptions to simplify the command usage around the samples.

### DIFF
--- a/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
+++ b/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
@@ -97,7 +97,6 @@ namespace Microsoft.Azure.Devices.Client.Samples
 
                 var propertyPatch = new ClientPropertyCollection
                 {
-                    { propertyName, "1", Thermostat1 },
                     { propertyName, targetHumidityRange, (int)StatusCode.Completed, writableProperties.Version, "The operation completed successfully.", Thermostat1 },
                 };
 
--- a/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
+++ b/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
         public async Task PerformOperationsAsync(CancellationToken cancellationToken)
         {
             // Retrieve the device's properties.
-            ClientProperties properties = await _deviceClient.GetPropertiesAsync(cancellationToken: cancellationToken);
+            ClientProperties properties = await _deviceClient.GetClientPropertiesAsync(cancellationToken: cancellationToken);
 
             // Verify if the device has previously reported a value for property
             // "initialValue" under component "thermostat2".
@@ -59,7 +59,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 {
                     { "initialValue", initialValue, Thermostat2 }
                 };
-                await _deviceClient.UpdatePropertiesAsync(propertiesToBeUpdated, cancellationToken);
+                await _deviceClient.UpdateClientPropertiesAsync(propertiesToBeUpdated, cancellationToken);
                 _logger.LogDebug($"Property: Update - {propertiesToBeUpdated.GetSerailizedString()}.");
             }
 
@@ -81,7 +81,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
             _logger.LogDebug($"Telemetry: Sent - {message.Telemetry.GetSerailizedString()}.");
 
             // Subscribe and respond to event for writable property "humidityRange" under component "thermostat1".
-            await _deviceClient.SubscribeToWritablePropertyEventAsync(async (writableProperties, userContext) =>
+            await _deviceClient.SubscribeToWritablePropertiesEventAsync(async (writableProperties, userContext) =>
             {
                 string propertyName = "humidityRange";
                 if (!writableProperties.Contains(Thermostat1)
@@ -101,7 +101,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
                     { propertyName, targetHumidityRange, (int)StatusCode.Completed, writableProperties.Version, "The operation completed successfully.", Thermostat1 },
                 };
 
-                await _deviceClient.UpdatePropertiesAsync(propertyPatch, cancellationToken);
+                await _deviceClient.UpdateClientPropertiesAsync(propertyPatch, cancellationToken);
                 _logger.LogDebug($"Property: Update - \"{propertyPatch.GetSerailizedString()}\" is complete.");
             },
             null,
--- a/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
+++ b/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
@@ -95,15 +95,10 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 HumidityRange targetHumidityRange = humidityRangeRequested.ToObject<HumidityRange>();
                 _logger.LogDebug($"Property: Received - component=\"{Thermostat1}\", {{ \"{propertyName}\": {targetHumidityRange} }}.");
 
-                var temperatureUpdateResponse = _deviceClient.CreateWritablePropertyResponse(
-                    targetHumidityRange,
-                    (int)StatusCode.Completed,
-                    writableProperties.Version,
-                    "The operation completed successfully.");
-
                 var propertyPatch = new ClientPropertyCollection
                 {
-                    { propertyName, temperatureUpdateResponse, Thermostat1 }
+                    { propertyName, "1", Thermostat1 },
+                    { propertyName, targetHumidityRange, (int)StatusCode.Completed, writableProperties.Version, "The operation completed successfully.", Thermostat1 },
                 };
 
                 await _deviceClient.UpdatePropertiesAsync(propertyPatch, cancellationToken);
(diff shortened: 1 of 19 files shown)


iothub/device/src/ClientOptions.cs
```diff
@@ -54,5 +54,13 @@ public class ClientOptions
         /// or the <see cref="ModuleClient.CreateFromEnvironmentAsync(ClientOptions)"/> flow.
         /// </remarks>
         public int SasTokenRenewalBuffer { get; set; }
+
+        /// <summary>
+        /// The payload convention to be used to serialize and encode the messages for our convention based methods. The default value is set to <see cref="DefaultPayloadConvention.Instance"/>.
+        /// </summary>
+        /// <remarks>
+        /// The <see cref="PayloadConvention"/> defines both the serializer and encoding to be used for our convetion based messages. You will only need to set this if you have objects that have special serialization rules or require a specific byte encoding.
```
abhipsaMisra, line 62: I think it would help to call out here that our default is newtonsoft json based and encodes with utf-8

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -90,12 +100,11 @@ public void AddOrUpdate(IDictionary<string, object> properties, string component
         /// Adds or updates the value for the collection.
         /// </summary>
         /// <remarks>
-        /// When adding a type of <see cref="IWritablePropertyResponse"/> the <see cref="ObjectSerializer.CheckWritablePropertyResponseType(object)"/>
-        /// checks to make sure it can serialize this type correctly.
-        /// This will only be evaluated at runtime so it will throw an exception if the type does not pass the check.
+        /// When using this as part of the <see cref="DeviceClient.SubscribeToWritablePropertyEventAsync(Func{ClientPropertyCollection, object, System.Threading.Tasks.Task}, object, System.Threading.CancellationToken)"/> flow. You should use the <see cref="PayloadConvention.CreateWritablePropertyResponse(object, int, long, string)"/> method to add the correct <see cref="IWritablePropertyResponse"/> to ensure the correct formatting is applied when the object is serialized.
```
abhipsaMisra, line 103: how about instead of mentioning the subscription flow, we instead mention that the response needs to be of the correct format when responding to a writable property update request?

`When using this as part of the <see cref="DeviceClient.UpdatePropertiesAsync(ClientPropertyCollection, System.Threading.CancellationToken)"/> flow to respond to a writable property update request, you should use the <see cref="PayloadConvention.CreateWritablePropertyResponse(object, int, long, string)"/> method to add the correct <see cref="IWritablePropertyResponse"/> to ensure the correct formatting is applied when the object is serialized.`

iothub/device/src/CommandRequest.cs
```diff
@@ -13,12 +13,12 @@ public sealed class CommandRequest
         private readonly byte[] _data;
         private readonly PayloadConvention _payloadConvention;
 
-        internal CommandRequest(string commandName, string componentName = default, byte[] data = default, PayloadConvention payloadConvention = default)
+        internal CommandRequest(string commandName, PayloadConvention payloadConvention, string componentName = default, byte[] data = default)
```
abhipsaMisra, line 16: I am being really nit picky: I kind of like the command name and component name next to each other; what do you think about moving the payload convention to be beginning?

iothub/device/src/DeviceClient.cs
```diff
@@ -22,6 +22,11 @@ public class DeviceClient : IDisposable
         /// </summary>
         public const uint DefaultOperationTimeoutInMilliseconds = 4 * 60 * 1000;
 
+        /// <summary>
+        /// The object serializer used by the <see cref="PayloadConvention"/> set in <see cref="ClientOptions.PayloadConvention"/>. If no <see cref="PayloadConvention"/> is set this will default to <see cref="NewtonsoftJsonObjectSerializer"/>.
+        /// </summary>
+        public ObjectSerializer ObjectSerializer => InternalClient.ObjectSerializer;
```
abhipsaMisra, line 28: q - for parity, should we also provide a getter for the encoder?

iothub/device/src/CommandRequest.cs
jamdavi: I'm good with that, I just wanted it to be a required param.

iothub/device/src/DeviceClient.cs
jamdavi: I thought about this but I don't think the customer will need to access the encoder for anything. If I were to do that maybe just being able to access the PayloadConvention.

My thought was to keep the overall PayloadConvention methods away from the customer since the convention should just be used to spit out the encoded bytes and deal with the most used part of serialization.

iothub/device/src/ClientOptions.cs
jamdavi: Let me add that to see if it makes it more clear.

iothub/device/src/InternalClient.cs
```diff
@@ -1887,9 +1890,22 @@ internal void ValidateModuleTransportHandler(string apiName)
...
+        /// </summary>
+        /// <param name="value">The value of the property.</param>
+        /// <param name="statusCode">The status code of the write operation.</param>
+        /// <param name="version">The version the property is responding to.</param>
+        /// <param name="description">An optional description of the writable property response.</param>
+        /// <returns></returns>
+        internal IWritablePropertyResponse CreateWritablePropertyResponse(object value, int statusCode, long version, string description = default)
         {
-            payloadConvention ??= DefaultPayloadConvention.Instance;
+            return _clientOptions.PayloadConvention.CreateWritablePropertyResponse(value, statusCode, version, description);
```
abhipsaMisra, line 1903: Is there an advantage of providing this method on the device client/ module client? One that I can think of is for scenarios where a user doesn't set any convention, then they don't even need to know about the existence of a convention but can retrieve the individual properties (serializer, encoder) or call this method from the device client. I like that, but one potential drawback I see is that we'll need to keep adding any new methods/ properties we add to the convention to the clients as well.

What do you think about highlighting the existence of a convention (and the default), so that a user can call these methods/ fetch the properties from the convention object itself; without us having to duplicate these getters?

iothub/device/src/InternalClient.cs
```diff
@@ -1902,11 +1918,11 @@ internal Task<ClientProperties> GetPropertiesAsync(PayloadConvention payloadConv
             }
         }
 
-        internal Task UpdatePropertiesAsync(ClientPropertyCollection propertyPatch, CancellationToken cancellationToken)
+        internal Task UpdatePropertiesAsync(ClientPropertyCollection clientProperties, CancellationToken cancellationToken)
```
abhipsaMisra, line 1921: nit: UpdateClientPropertiesAsync?

iothub/device/src/InternalClient.cs
```diff
@@ -1915,9 +1931,9 @@ internal Task UpdatePropertiesAsync(ClientPropertyCollection propertyPatch, Canc
             }
         }
 
-        internal Task SubscribeToWritablePropertyEventAsync(Func<ClientPropertyCollection, object, Task> callback, object userContext, PayloadConvention payloadConvention, CancellationToken cancellationToken)
+        internal Task SubscribeToWritablePropertyEventAsync(Func<ClientPropertyCollection, object, Task> callback, object userContext, CancellationToken cancellationToken)
```
abhipsaMisra, line 1934: I know that this is not the focus of this PR, but should we rename this as well - SubscribeToWritablePropertiesEventAsync?

abhipsaMisra (approved):
Left some comments, but I like it.

iothub/device/src/PayloadConvention.cs
```diff
@@ -32,5 +32,15 @@ public virtual byte[] GetObjectBytes(object objectToSendWithConvention)
...
         }
+
+        /// <summary>
+        /// Creates the correct <see cref="IWritablePropertyResponse"/> to be used with this serializer
+        /// </summary>
+        /// <param name="value">The value of the property.</param>
+        /// <param name="statusCode">The status code of the write operation.</param>
+        /// <param name="version">The version the property is responding to.</param>
+        /// <param name="description">An optional description of the writable property response.</param>
+        /// <returns></returns>
```
barustum, line 43: fill 'returns'

iothub/device/src/InternalClient.cs
```diff
@@ -149,6 +149,9 @@ internal class InternalClient : IDisposable
 
         internal delegate Task OnModuleEventMessageReceivedDelegate(string input, Message message);
 
+        internal ObjectSerializer ObjectSerializer => _clientOptions.PayloadConvention.PayloadSerializer;
+
```
barustum, line 153: extra new lines

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
```diff
@@ -95,18 +95,13 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
...
-                var temperatureUpdateResponse = new NewtonsoftJsonWritablePropertyResponse(
-                    targetHumidityRange,
-                    (int)StatusCode.Completed,
-                    writableProperties.Version,
-                    "The operation completed successfully.");
-
                 var propertyPatch = new ClientPropertyCollection
                 {
-                    { propertyName, temperatureUpdateResponse, Thermostat1 }
+                    { propertyName, "1", Thermostat1 },
```
abhipsaMisra, line 100: nit: just for this sample we should modify the property name and component name here -> else it'll be overwritten on the final patch

jamdavi: yeah was testing out that accessor, it should be back to normal.

iothub/device/src/DeviceClient.cs
```diff
@@ -22,6 +22,11 @@ public class DeviceClient : IDisposable
         /// </summary>
         public const uint DefaultOperationTimeoutInMilliseconds = 4 * 60 * 1000;
 
+        /// <summary>
+        /// The object serializer used by the <see cref="Client.PayloadConvention"/> set in <see cref="ClientOptions.PayloadConvention"/>. If no <see cref="Client.PayloadConvention"/> is set this will default to <see cref="NewtonsoftJsonObjectSerializer"/>.
```
abhipsaMisra, line 26: nit: needs an update - is now the convention and not the serializer

abhipsaMisra (commented):
Looks good!

PR #1925, opened 2021-04-29
feat(e2e): Adding e2e test for msi support on import jobs.

vinagesh:
## Checklist


e2e/test/iothub/service/RegistryManagerImportDevicesTests.cs
```diff
@@ -84,6 +85,16 @@ public async Task RegistryManager_ImportDevices(StorageAuthenticationType storag
                 {
                     try
                     {
+                        ManagedIdentity identity = null;
```
barustum, line 88: What do we do with identity in this test? I only see it being assigned on line 92.

vinagesh: We just pass it in and the test should succeed. The lines 43-45 dictate if the test should run this job with the identity or  key.

barustum (approved):

PR #1931, opened 2021-05-03
suggestions from review

abhipsaMisra:
- add a convenience getter for retrieving component level properties
- remove Telemetry from TelemetryMessage constructor -> settable through object initializer or property setter
- return `ClientPropertiesUpdateResponse` for `UpdatePropertiesAsync()` operation.


iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -231,6 +231,40 @@ public bool Contains(string propertyName)
...
+        /// <param name="componentName">The component which holds the required property.</param>
+        /// <param name="propertyName">The property to get.</param>
+        /// <param name="propertyValue">The value of the component-level property.</param>
+        /// <returns>true if the property collection contains a component level property with the specified key; otherwise, false.</returns>
+        public virtual bool TryGetValue<T>(string componentName, string propertyName, out T propertyValue)
+        {
+            if (Collection.ContainsKey(componentName))
+            {
+                // Extract the component-level properties into a JObject.
+                // The user specified serializer settings don't really affect this since this is an internal implementation detail.
```
abhipsaMisra, line 250: valid?

jamdavi: Correct point, but based on our conversation earlier about this I like it being part of the Serializer instead of the collection. Getting feedback from the team will help guide this decision.

iothub/device/src/TelemetryMessage.cs
```diff
@@ -51,30 +51,22 @@ public TelemetryCollection Telemetry
...
         /// <summary>
-        /// A conveneince constructor that allows you to set the <see cref="TelemetryCollection"/> and <see cref="Message.ComponentName"/> of this <see cref="TelemetryMessage"/>
+        /// A conveneince constructor that allows you to set the <see cref="Message.ComponentName"/> of this <see cref="TelemetryMessage"/>
         /// </summary>
         /// <param name="componentName">The name of the component.</param>
-        /// <param name="telemetryCollection">A collection of telemetry for this message.</param>
-        public TelemetryMessage(string componentName = default, TelemetryCollection telemetryCollection = default)
-            : base(telemetryCollection)
+        public TelemetryMessage(string componentName = default)
+            : base()
```
jamdavi, line 68: This breaks the content type and encoding feature for the Telemetry message. We'd need to inject the PayloadConvention somehow.

jamdavi (changes requested):

iothub/device/src/TelemetryMessage.cs
abhipsaMisra: actually this is broken even now, since the convention set in client options isn't getting relayed to the telemetry class. I'll fix it

iothub/device/src/Message.cs
```diff
@@ -62,20 +62,6 @@ public Message(byte[] byteArray)
...
 
-        /// <summary>
-        ///
-        /// </summary>
-        /// <param name="payloadCollection"></param>
-        internal Message(PayloadCollection payloadCollection)
-            : this()
-        {
-            ContentEncoding = payloadCollection?.Convention?.PayloadEncoder?.ContentEncoding?.WebName;
-            ContentType = payloadCollection?.Convention?.PayloadSerializer?.ContentType;
```
jamdavi, line 73: This is where the message properties were getting set before.

abhipsaMisra: yeah, the issue with removing the convention from a payload collection context is that the client initialization (where the convention is set) is a completely separate step vs creation of these payload collections.
so what we have now is:
-> create a telemetry message -> no information about convention, so cannot set the ContentEncoding and ContentType.
-> initialize a client instance -> set the convention information
-> call SendTelemetryAsync -> at this point we can set Message.ContentEncoding and Message.ContentType from ClientOptions.PayloadConvention, but that info is still not available in step 1 (cannot be retrieved from a TelemetryMessage object).

This becomes even more problematic with PayloadCollection, since the Add method there needs to convention info while adding a "writable property response", but that info isn't available until (1) a client is initialized, (2) UpdatePropertiesAsync() is called.

abhipsaMisra:
So this impl now does work, but there are 2 things that I'd like to discuss:

- user needs to be guided on how to pass in the "correct" writable property response:
https://github.com/Azure/azure-iot-sdk-csharp/blob/ccbf88805247ef93cad5a0ecaa1f8871d20831c7/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerCustomSerializer.cs#L93-L96

- convention is set somewhere else and the message is created via a different flow -> they are completely unaware of each other until `DeviceClient.SendTelemetryAsync()` is actually called. This means, we cannot fetch the correct `TelemetryMessage.ContentType` and encoding info. One option is to completely hide those fields and highlight that it is set via the convention directly.
https://github.com/Azure/azure-iot-sdk-csharp/blob/ccbf88805247ef93cad5a0ecaa1f8871d20831c7/iothub/device/src/TelemetryMessage.cs#L29-L44

jamdavi (approved):

PR #1932, opened 2021-05-03
Fix potential security issue

nyanzebra:
String equals can be a potentially timed security vulnerability, and could be problematic for authenticating SAS. This was noticed by the iotedge team and this PR is a way to bring notice or fix the issue. 

Please let me know if an issue should be filed instead. :)

## Checklist

## Description of the changes
This handles a potential security issue that could arise where an attacker could try to get information based on the time it takes to compare strings. This is just a quick fix; perhaps a shared location for comparing sensitive strings could be worthwhile?

N/A

### DIFF
--- a/iothub/device/src/Common/Security/SharedAccessSignature.cs
+++ b/iothub/device/src/Common/Security/SharedAccessSignature.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Net;
+using System.Runtime.CompilerServices;
 using System.Security.Cryptography;
 using System.Text;
 
@@ -116,7 +117,7 @@ namespace Microsoft.Azure.Devices.Client
             if (sasAuthorizationRule.PrimaryKey != null)
             {
                 string primareyKeyComputedSignature = ComputeSignature(Convert.FromBase64String(sasAuthorizationRule.PrimaryKey));
-                if (StringComparer.Ordinal.Equals(Signature, primareyKeyComputedSignature))
+                if (FixedTimeEquals(Signature, primareyKeyComputedSignature))
                 {
                     return;
                 }
@@ -125,7 +126,7 @@ namespace Microsoft.Azure.Devices.Client
             if (sasAuthorizationRule.SecondaryKey != null)
             {
                 string secondaryKeyComputedSignature = ComputeSignature(Convert.FromBase64String(sasAuthorizationRule.SecondaryKey));
-                if (StringComparer.Ordinal.Equals(Signature, secondaryKeyComputedSignature))
+                if (FixedTimeEquals(Signature, secondaryKeyComputedSignature))
                 {
                     return;
                 }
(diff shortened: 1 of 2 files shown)


iothub/device/src/Common/Security/SharedAccessSignature.cs
```diff
@@ -116,7 +117,7 @@ public void Authenticate(SharedAccessSignatureAuthorizationRule sasAuthorization
             if (sasAuthorizationRule.PrimaryKey != null)
             {
                 string primareyKeyComputedSignature = ComputeSignature(Convert.FromBase64String(sasAuthorizationRule.PrimaryKey));
-                if (StringComparer.Ordinal.Equals(Signature, primareyKeyComputedSignature))
+                if (FixedTimeEquals(Signature, primareyKeyComputedSignature))
```
abhipsaMisra, line 120: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.cryptographicoperations.fixedtimeequals?view=net-5.0 is available for ns2.1+, so an if-def for lower target frameworks would work here.

iothub/device/src/Common/Security/SharedAccessSignature.cs
```diff
@@ -205,5 +206,26 @@ internal static string Sign(byte[] key, string value)
...
+        private static bool FixedTimeEquals(string left, string right)
+        {
+            if (left.Length != right.Length)
+            {
+                return false;
+            }
+            var result = 0;
+            for (var i = 0; i < left.Length; i++)
+            {
+                result |= left[i] ^ right[i];
```
abhipsaMisra, line 226: I'd recommend we replace the xor with a subtraction. We'd then follow the same implementation that dotnet runtime uses: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/CryptographicOperations.cs#L31

abhipsaMisra: This is something that can definitely live in a common helper, we can propose that change if you'd like.

nyanzebra: Yeah, if there is good location that this should live in, I can move it there or it can be done after this PR. Whichever is fine with me :)

timtay-microsoft:
@nyanzebra After some initial discussion within our team, we don't see this code being used anywhere in our codebase. Rather than fix the functionality, we're inclined to just delete it since there will never be a need for this SDK to authorize requests.

https://github.com/Azure/azure-iot-sdk-csharp/pull/1935

Before we merge though, can you confirm that no one on the Edge team has any sort of dependency on these internal functions within our SDK?

nyanzebra:
> @nyanzebra After some initial discussion within our team, we don't see this code being used anywhere in our codebase. Rather than fix the functionality, we're inclined to just delete it since there will never be a need for this SDK to authorize requests.
> 
> #1935
> 
> Before we merge though, can you confirm that no one on the Edge team has any sort of dependency on these internal functions within our SDK?

It would appear that we might be using our own version/branch of the SDK, double checking this though. I think the plan will be for us to copy this code into our repo and you can go ahead and remove this from yours.

PR #1935, opened 2021-05-04
refactor(iot-dev, iot-service): Remove unused authentication code

timtay-microsoft:
These functions aren't used anywhere in the SDK, and they are internal, so they should be removed

Note that #1932 will be made obsolete if this change is merged since all it did was fix the implementation of these methods.


barustum (approved):

drwill-ms (approved):

abhipsaMisra (approved):

timtay-microsoft:
The Edge team confirmed that they will just bring this code into their repo, and that they don't mind if we remove it.

PR #1936, opened 2021-05-04
Add status codes to pnp samples

jamdavi:
Need feedback on other status codes and the names. 

Refs:
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#:~:text=%20List%20of%20HTTP%20status%20codes%20%201,See%20also.%20%205%20Notes.%20%20More%20

https://docs.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode?view=net-5.0#System_Net_HttpStatusCode_Accepted

### DIFF
--- a/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
+++ b/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Shared;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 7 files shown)


shared/src/StatusCodes.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// A list of common status codes to represent the response from the client.
+    /// </summary>
+    /// <remarks>
+    /// These status codes are based on the HTTP status codes listed here <see href="http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml"/>
+    /// </remarks>
+    public static class StatusCodes
```
abhipsaMisra, line 13: can you extend a static class? If I saw this class I'd want to extend it to add my custom codes in here.

abhipsaMisra: I can't think of a different approach that I prefer, even a partial class isn't the right option since it needs to be in the same assembly and namespace

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
```diff
@@ -125,28 +126,28 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
                                         var updateTemperatureResponse = new UpdateTemperatureResponse
                                         {
                                             TargetTemperature = updateTemperatureRequest.TargetTemperature,
-                                            Status = (int)StatusCode.Completed  // change this
+                                            Status = StatusCodes.OK  // change this
```
barustum, line 129: still need the comment?

shared/src/StatusCodes.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
barustum, line 1: copyrights header.

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerSampleNew.cs
jamdavi: Nope :)

abhipsaMisra (approved):

PR #1937, opened 2021-05-05
Convert to ClientPropertyCollection

jamdavi:
The primary purpose if this PR is to convert the ClientProperties to inherit from ClientPropertyCollection. But there was an opportunity to improve the flow when using this object.

Notes about the `TryGetNestedObjectValue` method highlighted below.

This method will be used by the PayloadCollections to help simplify handling properties under components. The most common use case is to get a single component property value to compare.

Here is an example of it simplifying using Contains on ClientProperties. So instead of using the proper serialization methods in the customer code it's offloaded to the Serializer.

### Questions
* If you were tasked with implementing a serializer would  the new method added be too much to do this? 
* Is there a better place to add this? 
* Is it intuitive to call Contains from the collection?

### Before Update
```c#
if (!properties.Contains(Thermostat2)
                || !((JsonElement)properties[Thermostat2])
                    .TryGetProperty("initialValue", out JsonElement initialValueReported)
                || !initialValue
                    .Equals(_deviceClient
                        .PayloadConvention
                        .PayloadSerializer
                        .DeserializeToType<ThermostatInitialValue>(initialValueReported.GetRawText())))
{

    // work
}
else
{
    (JsonElement)properties[Thermostat2]).TryGetProperty("initialValue", out JsonElement initialValueReported);
    var tValue = _deviceClient
                        .PayloadConvention
                        .PayloadSerializer
                        .DeserializeToType<ThermostatInitialValue>(initialValueReported.GetRawText()))
    // other work
}
```
### After Update
```c#
if (!properties.Contains("initialValue", Thermostat2))
{
    // work
}
else
{
    var tValue = properties.Get<ThermostatInitialValue>("initialValue", Thermostat2);
    // other work
}
```

### DIFF
--- a/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerCustomSerializer.cs
+++ b/iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerCustomSerializer.cs
@@ -49,14 +49,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 Temperature = 25
             };
 
-            if (!properties.Contains(Thermostat2)
-                || !((JsonElement)properties[Thermostat2])
-                    .TryGetProperty("initialValue", out JsonElement initialValueReported)
-                || !initialValue
-                    .Equals(_deviceClient
-                        .PayloadConvention
-                        .PayloadSerializer
-                        .DeserializeToType<ThermostatInitialValue>(initialValueReported.GetRawText())))
+            if (!properties.Contains("initialValue", Thermostat2))
             {
                 var propertiesToBeUpdated = new ClientPropertyCollection
                 {
@@ -64,6 +57,12 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 };
                 await _deviceClient.UpdateClientPropertiesAsync(propertiesToBeUpdated, cancellationToken);
                 _logger.LogDebug($"Property: Update - {propertiesToBeUpdated.GetSerializedString()}.");
+            }
+            else
+            {
+                var tValue = properties.Get<ThermostatInitialValue>("initialValue", Thermostat2);
+                _logger.LogDebug($"Property from tValue: {tValue.Humidity}.");
+                _logger.LogDebug($"Property from tValue: {tValue.Temperature}.");
             }
 
             // Send telemetry "deviceHealth" under component "thermostat1".
--- a/iothub/device/samples/PnpDeviceSamples/TemperatureController/SystemTextJsonSerializer.cs
+++ b/iothub/device/samples/PnpDeviceSamples/TemperatureController/SystemTextJsonSerializer.cs
@@ -31,6 +31,20 @@ namespace Microsoft.Azure.Devices.Client.Samples
         {
             return DeserializeToType<T>(((JsonElement)objectToConvert).ToString());
         }
+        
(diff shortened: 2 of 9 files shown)


jamdavi:
It's not part of this change but what are your thoughts on adding something like this for single level properties for components.

```c#
GetComponentProperty<T>(componentName, propertyName) ?
```

iothub/device/samples/PnpDeviceSamples/TemperatureController/SystemTextJsonSerializer.cs
```diff
@@ -31,6 +31,20 @@ public override T ConvertFromObject<T>(object objectToConvert)
...
+            if (objectToConvert == null || string.IsNullOrEmpty(propertyName))
+            {
+                return false;
+            }
+            if (((JsonElement)objectToConvert).TryGetProperty(propertyName, out var element)) {
+                outValue = DeserializeToType<T>(element.GetRawText());
+                return true;
+            }
+            return false;
+        }
```
jamdavi, line 47: Some questions around adding this method. 

* Does this add too much complexity? 
* Would you decide that this isn't worth the trouble?
* What would you expect to do instead.

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerCustomSerializer.cs
```diff
@@ -65,6 +58,12 @@ public async Task PerformOperationsAsync(CancellationToken cancellationToken)
                 await _deviceClient.UpdateClientPropertiesAsync(propertiesToBeUpdated, cancellationToken);
                 _logger.LogDebug($"Property: Update - {propertiesToBeUpdated.GetSerializedString()}.");
             }
+            else
+            {
+                var tValue = properties.Get<ThermostatInitialValue>("initialValue", Thermostat2);
```
drwill-ms, line 63: TryGet?

drwill-ms: If you had that, they wouldn't need to call `Contains()` first, which is simpler from an end-user POV

iothub/device/samples/PnpDeviceSamples/TemperatureController/SystemTextJsonSerializer.cs
drwill-ms: Add too much complexity for whom? For our impl? For customers who want to implement their own serializer?

I think the resulting user experience is better, and the cost to impl is worth it.

iothub/device/samples/PnpDeviceSamples/TemperatureController/ComponentTemperatureControllerCustomSerializer.cs
abhipsaMisra: yes, this is getting updated to TryGet

iothub/device/samples/PnpDeviceSamples/TemperatureController/SystemTextJsonSerializer.cs
abhipsaMisra: The complexity would be for customers who want to implement their own serializer (since this is a part of the PayloadSerializer abstract class). We would provide newtonsoft and system.text samples, so that should help them out; but it is an additional step that they have to account for.

PR #1940, opened 2021-05-06
fix(device): allow null method payload

drwill-ms:


iothub/device/src/Common/Utils.cs
```diff
@@ -96,17 +96,14 @@ public static string ConvertDeliveryAckTypeToString(DeliveryAcknowledgement valu
 
         public static void ValidateDataIsEmptyOrJson(byte[] data)
         {
-            if (data.Length != 0)
+            if (data != null
```
drwill-ms, line 99: Payload could be null

iothub/device/src/InternalClient.cs
```diff
@@ -106,7 +107,7 @@ public enum MessageResponse
     /// </summary>
     internal class InternalClient : IDisposable
     {
-        private readonly SemaphoreSlim _methodsDictionarySemaphore = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim _methodsSemaphore = new SemaphoreSlim(1, 1);
```
drwill-ms, line 110: This semaphore was used more than just for the dictionary.

iothub/device/src/InternalClient.cs
```diff
@@ -362,16 +356,14 @@ public Task CloseAsync(CancellationToken cancellationToken)
         /// <summary>
         /// Sets a new delegate for the connection status changed callback. If a delegate is already associated,
         /// it will be replaced with the new delegate.
-        /// <param name="statusChangesHandler">The name of the method to associate with the delegate.</param>
         /// </summary>
+        /// <param name="statusChangesHandler">The name of the method to associate with the delegate.</param>
```
drwill-ms, line 360: This misplacement of tags inside the summary was all over the place

iothub/device/src/InternalClient.cs
```diff
@@ -764,52 +753,30 @@ public async Task SetMethodHandlerAsync(string methodName, MethodCallback method
...
+            await _methodsSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
 
             try
             {
                 if (methodHandler != null)
                 {
-                    // codes_SRS_DEVICECLIENT_10_005: [ It shall EnableMethodsAsync when called for the first time. ]
-                    await EnableMethodAsync(cancellationToken).ConfigureAwait(false);
-
-                    // codes_SRS_DEVICECLIENT_10_001: [ It shall lazy-initialize the deviceMethods property. ]
```
drwill-ms, line 789: Look at how much simpler the code here gets when we relegate the logic to (the now renamed) `EnableMethodAsync` and `DisableMethodAsync`.

iothub/device/src/InternalClient.cs
```diff
@@ -893,30 +856,27 @@ public async Task SetMethodDefaultHandlerAsync(MethodCallback methodHandler, obj
...
         /// Sets a new delegate for the named method. If a delegate is already associated with
         /// the named method, it will be replaced with the new delegate.
+        /// </summary>
         /// <param name="methodName">The name of the method to associate with the delegate.</param>
         /// <param name="methodHandler">The delegate to be used when a method with the given name is called by the cloud service.</param>
         /// <param name="userContext">generic parameter to be interpreted by the client code.</param>
-        /// </summary>
 
         [Obsolete("Please use SetMethodHandlerAsync.")]
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
drwill-ms, line 875: This looks like one we should hide from the API surface. Any objection?

iothub/device/src/InternalClient.cs
```diff
@@ -953,32 +909,27 @@ internal async Task OnMethodCalledAsync(MethodRequestInternal methodRequestInter
...
             byte[] requestData = methodRequestInternal.GetBytes();
 
-            await _methodsDictionarySemaphore.WaitAsync().ConfigureAwait(false);
+            await _methodsSemaphore.WaitAsync().ConfigureAwait(false);
             try
             {
                 Utils.ValidateDataIsEmptyOrJson(requestData);
-                _deviceMethods?.TryGetValue(methodRequestInternal.Name, out callbackContextPair);
-                if (callbackContextPair == null)
+                if (!_deviceMethods.TryGetValue(methodRequestInternal.Name, out callbackContextPair))
```
drwill-ms, line 916: Using `TryGet*` the way it was meant to be used.

iothub/device/src/InternalClient.cs
```diff
@@ -1054,11 +1001,16 @@ internal Task SendMethodResponseAsync(MethodResponseInternal methodResponse, Can
...
         }
 
-        private Task EnableMethodAsync(CancellationToken cancellationToken)
+        private async Task HandleMethodEnableAsync(CancellationToken cancellationToken)
         {
-            return _deviceMethods == null && _deviceDefaultMethodCallback == null
-                ? InnerHandler.EnableMethodsAsync(cancellationToken)
-                : TaskHelpers.CompletedTask;
+            // If currently enabled, then skip
+            if (_isDeviceMethodEnabled)
```
drwill-ms, line 1007: Now that `_deviceMethods` is no longer null, it can't be used as a determining factor. So we'll just rely on this boolean to control it.

iothub/device/src/InternalClient.cs
```diff
@@ -1303,28 +1253,33 @@ public Task<Message> ReceiveAsync(CancellationToken cancellationToken)
...
         }
 
-        private Task DisableMethodAsync(CancellationToken cancellationToken)
+        private async Task HandleMethodDisableAsync(CancellationToken cancellationToken)
         {
-            return _deviceMethods == null && _deviceDefaultMethodCallback == null
-                ? InnerHandler.DisableMethodsAsync(cancellationToken)
-                : TaskHelpers.CompletedTask;
+            // Don't disable if it is already disabled or if there are registered device methods
+            if (!_isDeviceMethodEnabled
```
drwill-ms, line 1259: Trying to be a little more careful about when to disable methods, so it still includes the default handler not being null or there being any specific methods.

iothub/device/src/MethodResponseStatusCode.cs
```diff
@@ -8,15 +8,20 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public enum MethodResponseStatusCode
     {
-#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+        /// <summary>
```
drwill-ms, line 11: Why not just fix it, rather than suppress?

iothub/device/src/InternalClient.cs
```diff
@@ -175,30 +176,24 @@ public InternalClient(IotHubConnectionString iotHubConnectionString, ITransportS
...
 
             if (Logging.IsEnabled)
-            {
                 Logging.Associate(this, innerHandler, nameof(InternalClient));
-            }
 
             InnerHandler = innerHandler;
 
             if (Logging.IsEnabled)
-            {
```
timtay-microsoft, line 185: Are we okay with if statements not having braces? This seems a little risky to me

iothub/device/src/InternalClient.cs
timtay-microsoft: Since it is obsolete, I have no objections

iothub/device/src/InternalClient.cs
drwill-ms: This was one area we agreed we'd reduce brace use. We have lots of cases of `if (Logging.IsEnabled) Logging.Trace(...);`

timtay-microsoft: It would certainly be nice if we didn't have to check if logging was enabled before every log statement. If only .NET was more like Java here, haha

drwill-ms: The logging methods don't log if it is not enabled, however  the parameters (namely string concat or interpolation) passed to the logging method are built even if the logging method doesn't log, so the cost to the client is still high (memory allocation and CPU cost) even if the statements are not logged to ETW.

drwill-ms: How does it work for Java?

iothub/device/src/InternalClient.cs
```diff
@@ -861,19 +826,17 @@ public async Task SetMethodDefaultHandlerAsync(MethodCallback methodHandler, obj
...
 
             cancellationToken.ThrowIfCancellationRequested();
-            await _methodsDictionarySemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+            await _methodsSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
 
             try
             {
                 if (methodHandler != null)
                 {
                     // codes_SRS_DEVICECLIENT_10_005: [ It shall EnableMethodsAsync when called for the first time. ]
```
barustum, line 838: the comment still has the old name (same for other lines). Not sure what these codes_*  lines are, but should they be removed?

barustum (approved):

iothub/device/src/InternalClient.cs
```diff
@@ -893,30 +856,27 @@ public async Task SetMethodDefaultHandlerAsync(MethodCallback methodHandler, obj
...
         /// <param name="methodHandler">The delegate to be used when a method with the given name is called by the cloud service.</param>
         /// <param name="userContext">generic parameter to be interpreted by the client code.</param>
-        /// </summary>
 
         [Obsolete("Please use SetMethodHandlerAsync.")]
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public void SetMethodHandler(string methodName, MethodCallback methodHandler, object userContext)
         {
             if (Logging.IsEnabled)
-            {
```
vinagesh, line 917: are the {} brackets gone now?

iothub/device/src/InternalClient.cs
vinagesh: Can _deviceMethods be null at all? The old code seems to think it can be null. We should add a check for null if it can be null to avoid NRE.

iothub/device/src/InternalClient.cs
```diff
@@ -953,32 +909,27 @@ internal async Task OnMethodCalledAsync(MethodRequestInternal methodRequestInter
...
-                if (callbackContextPair == null)
+                if (!_deviceMethods.TryGetValue(methodRequestInternal.Name, out callbackContextPair))
                 {
                     callbackContextPair = _deviceDefaultMethodCallback;
                 }
             }
             catch (Exception ex) when (!ex.IsFatal())
             {
                 if (Logging.IsEnabled)
-                {
```
vinagesh, line 969: Why are the {} gone?

iothub/device/src/InternalClient.cs
vinagesh: I see the explanation for this below. Nevermind.

iothub/device/src/InternalClient.cs
barustum: Yup. Tim made a similar comment above:
https://github.com/Azure/azure-iot-sdk-csharp/pull/1940#discussion_r628430869

iothub/device/src/InternalClient.cs
drwill-ms: I'll remove these specific ones. Yeah, I think we should probably remove (or at least refactor many of these to normal code comments. I'm going to pass on doing that at this time, however.

iothub/device/src/InternalClient.cs
drwill-ms: We agreed we'd be okay with implicit body for logging statements, to cut down on how many lines they take.

iothub/device/src/InternalClient.cs
vinagesh: Hmmm. Is there an advantage of reducing the braces? It somehow looks odd to me. It will also cause of lot of these changes in every PR that touches the test files.

iothub/device/src/InternalClient.cs
```diff
@@ -121,7 +122,9 @@ internal class InternalClient : IDisposable
         private volatile Tuple<MessageHandler, object> _defaultEventCallback;
 
         // Stores methods supported by the client device and their associated delegate.
-        private volatile Dictionary<string, Tuple<MethodCallback, object>> _deviceMethods;
+
+        private volatile bool _isDeviceMethodEnabled;
+        private readonly Dictionary<string, Tuple<MethodCallback, object>> _deviceMethods = new Dictionary<string, Tuple<MethodCallback, object>>();
```
vinagesh, line 127: I'm not very sure if this should be volatile or not. But just an observation that only the bool is volatile now. I have never used volatile in my coding experience but if you have an answer, I would like to learn why it was only used on the bool.

vinagesh (approved):

iothub/device/src/InternalClient.cs
drwill-ms: I've put them back for now. We can talk about it in scrum parking lot on Monday.

iothub/device/src/InternalClient.cs
drwill-ms: It means it could be changed by more than one thread, and the compiler will remove optimizations that might otherwise compromise reading a fresh value.

<https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile>

drwill-ms: Checking again, this boolean is only changed inside a semaphore, so I think we don't need it.

PR #1943, opened 2021-05-06
fix(doc): Updating all documentation to use the updated env variable …

vinagesh:
…names.

## Checklist

## Description of the changes
1) Removed the old setup scripts we no longer use
2) Updated the names of environment variables in readmes to match what we have in our e2e setup script


e2e/test/prerequisites/iot_config.cmd_template
```diff
@@ -1,102 +0,0 @@
-@REM Copyright (c) Microsoft. All rights reserved.
```
abhipsaMisra, line 1: I remember we have this file linked somewhere in our contribution prereq readme/ devbox setup; we'll need to remove that as well. 

So, this file was meant to facilitate users who wish to contribute to our repo, that they could run the e2e tests locally before submitting PRs. 
How do we plan to address that? I do support removing this file (and thus remove this maintenance overhead for us), maybe we could add a link to our resource generation script where we load the env var to keyvault, and users could use that as a guide to discover the env variables that are required for E2E tests?
https://github.com/Azure/azure-iot-sdk-csharp/blob/0ec02be9ac6fa9c97485a77099bdd822907a021d/e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1#L581-L622

Also, on an unrelated note, I just saw that we have some shared access keys exposed on our public repo, are we sure that this isn't a security issue: https://github.com/Azure/azure-iot-sdk-csharp/blob/0ec02be9ac6fa9c97485a77099bdd822907a021d/e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1#L618?

vinagesh: 1) The readme.md was talking about all the env variables we use in the tests and was thinking that would be sufficient. But it would be good to point customers to the scripts like you suggested. Let me see what is the best place in the readme.md to point to that.
2) The invalid certs are not a security issue as they are incorrect and won't work. We don't seem to expose any real shared access keys.

vinagesh: This also brings a good question of anyone who uses linux or mac and not windows. Our script is only for windows powershell. The template itself might not explain how to set it up though.

drwill-ms: Powershell exists on Linux. Are you saying we've written it in a way that won't work for Linux, or something else?

vinagesh: I've never tried it. And I was not aware linux and mac users can use it. Good to know! I'll work on updating the readme to point to that then. Otherwise we have too many things out of sync already in terms of e2e setup readme.

e2e/test/prerequisites/readme.md
```diff
@@ -1,11 +1,34 @@
+
 # Azure IoT C# End-to-end test prerequisites
```
vinagesh, line 2: This file is completely changed. So a good way to look at it is - View file option to read through the contents and add feedback.

e2e/test/prerequisites/iot_config.cmd_template
vinagesh: I have rewritten the readme. Would appreciate feedback.

e2e/test/prerequisites/readme.md
```diff
@@ -1,11 +1,34 @@
...
-Please copy the iot_config.* scripts to a secure location, remove the `_template` suffix and follow the comments to fill in the missing information.
 
-To generate required test certificates you can use the tools available [here](https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/CACertificateOverview.md).
+- Navigate to reporoot\azure-iot-sdk-csharp\e2e\test\prerequisites\E2ETestsSetup
+
+- Open powershell in Administrator mode and run the following command by replacing the variables in brackets with your own preferred values.
+  
+  ```Shell
+  .\e2eTestsSetup -SubscriptionId [yourSubscriptionId] -Region [YourRegion] -ResourceGroup [ResourceGroupName] -GroupCertificatePassword [Password] -InstallDependencies [true/false]
+  ```
```
abhipsaMisra, line 12: Couple of thoughts:

abhipsaMisra: I think we can state something like:
use this script to create all resources that you need to run our entire E2E test suite, or go to this link (link to our resources generation script) to see which resources are required and what env variables they are associated with.

vinagesh: 1) In common errors, it is called out that the script is idempotent meaning, it will not create the resource again and again with a different name unless the user provides a different name. 
2) Given we want uses to run all the E2E tests before submitting any PRs, we are not creating anything unnecessary other than a KeyVault, which is very very negligible cost given how many transactions they will do with it. I don't think we really need to maintain a separate file. Those files are already out of sync.
3) It would fine to mention to look at the script but don't you think that is understood as the instructions say they create all the necessary resources? If they really want to selectively create resources, they are free to do so.

vinagesh: Overall I am seeing this as a convenience we are providing to the user. But if they are an advanced user, they will have the expertise to selectively create what they want and run. I don't see too much value in giving them instructions on running every test. It would be easier if we wrote the tests in a way where each test class has a list of environment variables it needs (similar to samples) so that it is self explanatory. However our tests are not designed like that either so some env variables are just manadatory.

abhipsaMisra: I think this will be easier to discuss over a call :)

abhipsaMisra (approved):

PR #1945, opened 2021-05-07
Hiding the content properties

jamdavi:
We can't actually hide this member unless we were to change the way the Message class is designed.

### DIFF
--- a/iothub/device/src/TelemetryMessage.cs
+++ b/iothub/device/src/TelemetryMessage.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.IO;
 
 namespace Microsoft.Azure.Devices.Client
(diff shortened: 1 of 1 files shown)


iothub/device/src/TelemetryMessage.cs
```diff
@@ -23,23 +24,27 @@ public class TelemetryMessage : Message
...
-        /// The ability to set this property has been hidden by this class to only allow you to set the ContentType via the <see cref="PayloadSerializer"/> class.
+        /// This property has been hidden by this class to only allow you to set the ContentType via the <see cref="PayloadSerializer"/> class.
         /// </remarks>
         /// <value>The base <see cref="Message"/> content type.</value>
+        [Browsable(false)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public new string ContentType
         {
-            get => base.ContentType;        // TODO - this info is available only after deviceClient.SendTelemetryAsync() has been called.
+            get => base.ContentType;
```
jamdavi, line 34: There is no way to make this private without changing the base class.

abhipsaMisra (approved):

Issue #1911, opened 2021-04-26
[Bug Report] NullReferenceException in AuthenticationWithTokenRefresh


its-phil:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64
- **Application's .NET Target Framework :** netcoreapp3.1, net5.0
- **Device:** Laptop
- **SDK version used:** 1.36.0

## Description of the issue
I was testing the DeviceClient connection using TPM for attestation when I stumbled across a `NullReferenceException`.
The desired sequence is:
1. Create a `DeviceClient` instance for direct to the IoT Hub (no gateway).
2. Disconnect the first `DeviceClient` and connect again, this time via an Edge gateway.

If the IoT hub address is valid it's all good. However, if for some reason the IoT Hub address is invalid (e.g. not reachable), i get said `NullReferenceException` in the second step.

## Code sample exhibiting the issue
See this repo for a demo: https://github.com/its-phil/GatewayTest

I debugged into the SDK code and found that the exception occurs in `AuthenticationWithTokenRefresh.cs` in line 78 (`Debug.Assert(_lock != null);`).

Am I using the SDK wrong or might this be an issue in the SDK?

## Console log of the issue
```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Azure.Devices.Client.AuthenticationWithTokenRefresh.GetTokenAsync(String iotHub)
   at Microsoft.Azure.Devices.Client.Transport.AmqpIoT.AmqpIoTCbsTokenProvider.GetTokenAsync(Uri namespaceAddress, String appliesTo, String[] requiredClaims)
   at Microsoft.Azure.Amqp.TaskHelpers.EndAsyncResult(IAsyncResult asyncResult)
   at Microsoft.Azure.Amqp.IteratorAsyncResult`1.<>c.<CallTask>b__24_1(TIteratorAsyncResult thisPtr, IAsyncResult r)
   at Microsoft.Azure.Amqp.IteratorAsyncResult`1.StepCallback(IAsyncResult result)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.Azure.Amqp.ExceptionDispatcher.Throw(Exception exception)
   at Microsoft.Azure.Amqp.AsyncResult.End[TAsyncResult](IAsyncResult result)
   at Microsoft.Azure.Amqp.AsyncResult`1.End(IAsyncResult asyncResult)
[...]

PR #1954, opened 2021-05-11
Update client to not dispose user supplied IDisposable types

abhipsaMisra:
This PR fixes two issues:
- `IotHubConnectionString.TokenRefresher` is an `IDispoable` `IAuthenticationMethod.AuthenticationWithTokenRefresh`, which can be user-supplied. So, we shouldn't dispose it within our client, since they might want to reuse it.
Reference: https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-5.0#implementing-idisposable
"You should implement IDisposable only if your type uses unmanaged resources directly. The consumers of your type can call your IDisposable.Dispose implementation to free resources when the instance is no longer needed."

- `IotHubConnectionString.SharedAccessSignature` should be set only if it is non-null and the authentication method of the device client is not of type `AuthenticationWithTokenRefresh`. Setting the sas value for an `AuthenticationWithTokenRefresh` authentication type will result in tokens not being renewed. This flow can be hit if the same authentication method is always used to initialize the client; as in, on disposal and reinitialization. (the value of the sas token computed is stored within the authentication method, and on reinitialization the client is incorrectly identified as a fixed-sas-token-initialized client, instead of being identified as a sas-token-refresh-enabled-client.)

Fix for #1911 

Dependency code map of `IAuthenticationMethod.AuthenticationWithTokenRefresh`:

image

### DIFF
--- a/doc/devguide.md
+++ b/doc/devguide.md
@@ -9,6 +9,11 @@ Before starting development on the Azure IoT SDK for C# you will need to install
 ### Design
 We are following the [Azure SDK design specification for C#](https://azuresdkspecs.z5.web.core.windows.net/DotNetSpec.html). To preserve backward compatibility, existing code will not change to follow these rules.
 
+#### Using [`IDisposable`](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-5.0#implementing-idisposable) types:
+- The sdk should dispose any `IDisposable` resource that it creates.
+- The sdk should not dispose an `IDisposable` resource that is supplied by the calling application. This is because the caller might want to reuse the resource elsewhere in the application. The responsibility of disposal of caller-supplied `IDisposable` resource is on the caller. An example of such a resource would be an `X509Certificate` instance that is used for authenticating our clients.
+- If the sdk implements `class A` that owns an `IDisposable` resource `X`, then `A` should also be `IDisposable`. In that case, resource `X` can be safely disposed when `class A` is disposed.
+
 ### Code style
 Please read and apply our [coding style](coding-style.md) when proposing PRs against the Azure repository. When changing existing files, please apply changes to the entire file. Otherwise, maintain the same style.
 
--- a/doc/devguide.md
+++ b/doc/devguide.md
@@ -10,9 +10,9 @@ Before starting development on the Azure IoT SDK for C# you will need to install
 We are following the [Azure SDK design specification for C#](https://azuresdkspecs.z5.web.core.windows.net/DotNetSpec.html). To preserve backward compatibility, existing code will not change to follow these rules.
 
 #### Using [`IDisposable`](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-5.0#implementing-idisposable) types:
+- If the sdk implements `class A` that owns an `IDisposable` resource `X`, then `A` should also be `IDisposable`. In that case, resource `X` can be safely disposed when `class A` is disposed.
 - The sdk should dispose any `IDisposable` resource that it creates.
 - The sdk should not dispose an `IDisposable` resource that is supplied by the calling application. This is because the caller might want to reuse the resource elsewhere in the application. The responsibility of disposal of caller-supplied `IDisposable` resource is on the caller. An example of such a resource would be an `X509Certificate` instance that is used for authenticating our clients.
-- If the sdk implements `class A` that owns an `IDisposable` resource `X`, then `A` should also be `IDisposable`. In that case, resource `X` can be safely disposed when `class A` is disposed.
 
 ### Code style
 Please read and apply our [coding style](coding-style.md) when proposing PRs against the Azure repository. When changing existing files, please apply changes to the entire file. Otherwise, maintain the same style.
--- a/e2e/test/Helpers/AmqpConnectionStatusChange.cs
+++ b/e2e/test/Helpers/AmqpConnectionStatusChange.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.Devices.Client;
+
+namespace Microsoft.Azure.Devices.E2ETests.Helpers
+{
+    public class AmqpConnectionStatusChange
+    {
+        private readonly string _deviceId;
+        private readonly MsTestLogger _logger;
+
+        public AmqpConnectionStatusChange(string deviceId, MsTestLogger logger)
+        {
+            LastConnectionStatus = null;
+            LastConnectionStatusChangeReason = null;
+            ConnectionStatusChangeCount = 0;
+            _deviceId = deviceId;
+            _logger = logger;
+        }
+
+        public void ConnectionStatusChangesHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
+        {
+            ConnectionStatusChangeCount++;
+            LastConnectionStatus = status;
+            LastConnectionStatusChangeReason = reason;
+            _logger.Trace($"{nameof(AmqpConnectionStatusChange)}.{nameof(ConnectionStatusChangesHandler)}: {_deviceId}: status={status} statusChangeReason={reason} count={ConnectionStatusChangeCount}");
+        }
+
+        public int ConnectionStatusChangeCount { get; set; }
+
+        public ConnectionStatus? LastConnectionStatus { get; set; }
+
+        public ConnectionStatusChangeReason? LastConnectionStatusChangeReason { get; set; }
+    }
+}
--- a/e2e/test/Helpers/AmqpConnectionStatusChange.cs
+++ b/e2e/test/Helpers/AmqpConnectionStatusChange.cs
@@ -19,6 +19,12 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
             _logger = logger;
         }
 
+        public int ConnectionStatusChangeCount { get; set; }
+
+        public ConnectionStatus? LastConnectionStatus { get; set; }
+
+        public ConnectionStatusChangeReason? LastConnectionStatusChangeReason { get; set; }
+
         public void ConnectionStatusChangesHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
         {
             ConnectionStatusChangeCount++;
@@ -26,11 +32,5 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
             LastConnectionStatusChangeReason = reason;
             _logger.Trace($"{nameof(AmqpConnectionStatusChange)}.{nameof(ConnectionStatusChangesHandler)}: {_deviceId}: status={status} statusChangeReason={reason} count={ConnectionStatusChangeCount}");
         }
-
-        public int ConnectionStatusChangeCount { get; set; }
-
-        public ConnectionStatus? LastConnectionStatus { get; set; }
-
-        public ConnectionStatusChangeReason? LastConnectionStatusChangeReason { get; set; }
     }
 }
--- a/e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
+++ b/e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
@@ -112,7 +112,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                 authenticationMethods.Add(authenticationMethod);
             }
 
-            // Initialize the client instances set the connection status change handler and open the connection.
+            // Initialize the client instances, set the connection status change handler and open the connection.
             for (int i = 0; i < devicesCount; i++)
             {
                 DeviceClient deviceClient = DeviceClient.Create(testDevices[i].IoTHubHostName, authenticationMethods[i], new ITransportSettings[] { amqpTransportSettings });
--- a/e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
+++ b/e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
@@ -1,0 +1,185 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests.Iothub
+{
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    public class AuthenticationWithTokenRefreshDisposalTests : E2EMsTestBase
+    {
+        private readonly string _devicePrefix = $"E2E_{nameof(AuthenticationWithTokenRefreshDisposalTests)}_";
+
+        [LoggedTestMethod]
+        public async Task DeviceSak_ReusableAuthenticationMethod_SingleDevicePerConnection_Amqp()
+        {
+            await ReuseAuthenticationMethod_SingleDevice(Client.TransportType.Amqp_Tcp_Only).ConfigureAwait(false);
+        }
+
+        [LoggedTestMethod]
+        public async Task DeviceSak_ReusableAuthenticationMethod_SingleDevicePerConnection_AmqpWs()
+        {
+            await ReuseAuthenticationMethod_SingleDevice(Client.TransportType.Amqp_WebSocket_Only).ConfigureAwait(false);
+        }
+
+        [LoggedTestMethod]
+        public async Task DeviceSak_ReusableAuthenticationMethod_SingleDevicePerConnection_Mqtt()
+        {
+            await ReuseAuthenticationMethod_SingleDevice(Client.TransportType.Mqtt_Tcp_Only).ConfigureAwait(false);
+        }
+
+        [LoggedTestMethod]
+        public async Task DeviceSak_ReusableAuthenticationMethod_SingleDevicePerConnection_MqttWs()
+        {
+            await ReuseAuthenticationMethod_SingleDevice(Client.TransportType.Mqtt_WebSocket_Only).ConfigureAwait(false);
+        }
+
+        [LoggedTestMethod]
+        public async Task DeviceSak_ReusableAuthenticationMethod_SingleDevicePerConnection_Http()
+        {
+            await ReuseAuthenticationMethod_SingleDevice(Client.TransportType.Http1).ConfigureAwait(false);
+        }
+
+        [LoggedTestMethod]
+        public async Task DeviceSak_ReusableAuthenticationMethod_MuxedDevicesPerConnection_Amqp()
+        {
+            await ReuseAuthenticationMethod_MuxedDevices(Client.TransportType.Amqp_Tcp_Only, 2);
+        }
+
+        [LoggedTestMethod]
+        public async Task DeviceSak_ReusableAuthenticationMethod_MuxedDevicesPerConnection_AmqpWs()
+        {
+            await ReuseAuthenticationMethod_MuxedDevices(Client.TransportType.Amqp_WebSocket_Only, 2);
+        }
+
+        private async Task ReuseAuthenticationMethod_SingleDevice(Client.TransportType transport)
+        {
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix).ConfigureAwait(false);
+            var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString);
+
+            // Create an instance of the device client, send a test message and then close and dispose it.
+            DeviceClient deviceClient = DeviceClient.Create(testDevice.IoTHubHostName, authenticationMethod, transport);
+            await deviceClient.SendEventAsync(new Client.Message()).ConfigureAwait(false);
+            await deviceClient.CloseAsync();
+            deviceClient.Dispose();
+            Logger.Trace("Test with instance 1 completed");
+
+            // Perform the same steps again, reusing the previously created authentication method instance.
+            DeviceClient deviceClient2 = DeviceClient.Create(testDevice.IoTHubHostName, authenticationMethod, transport);
+            await deviceClient2.SendEventAsync(new Client.Message()).ConfigureAwait(false);
+            await deviceClient2.CloseAsync();
+            deviceClient2.Dispose();
+            Logger.Trace("Test with instance 2 completed");
+        }
+
+        private async Task ReuseAuthenticationMethod_MuxedDevices(Client.TransportType transport, int devicesCount)
+        {
+            IList<TestDevice> testDevices = new List<TestDevice>();
+            IList<AuthenticationWithTokenRefresh> authenticationMethods = new List<AuthenticationWithTokenRefresh>();
+
+            // Set up amqp transport settings to multiplex all device sessions over the same amqp connection.
+            var amqpTransportSettings = new AmqpTransportSettings(transport)
+            {
+                AmqpConnectionPoolSettings = new AmqpConnectionPoolSettings
+                {
+                    Pooling = true,
+                    MaxPoolSize = 1,
+                },
+            };
+
+            for (int i = 0; i < devicesCount; i++)
+            {
+                TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix).ConfigureAwait(false);
+                var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString);
+
+                testDevices.Add(testDevice);
+                authenticationMethods.Add(authenticationMethod);
+            }
+
+            // Create an instance of the device client, send a test message and then close and dispose it.
+            for (int i = 0; i < devicesCount; i++)
+            {
+                DeviceClient deviceClient = DeviceClient.Create(testDevices[i].IoTHubHostName, authenticationMethods[i], new ITransportSettings[] { amqpTransportSettings });
+                await deviceClient.SendEventAsync(new Client.Message()).ConfigureAwait(false);
+                await deviceClient.CloseAsync();
+                deviceClient.Dispose();
+                Logger.Trace($"Test with client {i} completed.");
+            }
+            Logger.Trace($"Test run with instance 1 completed.");
+
+            // Perform the same steps again, reusing the previously created authentication method instance.
+            for (int i = 0; i < devicesCount; i++)
+            {
+                DeviceClient deviceClient = DeviceClient.Create(testDevices[i].IoTHubHostName, authenticationMethods[i], new ITransportSettings[] { amqpTransportSettings });
+                await deviceClient.SendEventAsync(new Client.Message()).ConfigureAwait(false);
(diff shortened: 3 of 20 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -1895,7 +1895,6 @@ public void Dispose()
             _fileUploadHttpTransportHandler?.Dispose();
             _deviceReceiveMessageSemaphore?.Dispose();
             _twinDesiredPropertySemaphore?.Dispose();
-            IotHubConnectionString?.TokenRefresher?.Dispose();
```
drwill-ms, line 1898: This is a breaking change. If we don't dispose it, and a customer had been relying on us to do that, now they have a leak.

You need to follow the breaking change process here.

drwill-ms (changes requested):
Breaking change policy

iothub/device/src/InternalClient.cs
abhipsaMisra: The user would need to explicitly dispose it only for scenarios where they supply the sdk with an `AuthenticationWithTokenRefresh` implementation.

drwill-ms: It is a change of behavior.

iothub/device/src/AuthenticationWithTokenRefresh.cs
```diff
@@ -65,6 +65,10 @@ public abstract class AuthenticationWithTokenRefresh : IAuthenticationMethod, ID
         /// </summary>
         public bool IsExpiring => (ExpiresOn - DateTime.UtcNow).TotalSeconds <= _bufferSeconds;
 
+        // This internal property is used by the sdk to determine if the instance was created by the sdk,
+        // and thus, if it should be disposed by the sdk.
+        internal bool ShouldSdkDisposeInstance { get; set; }
```
drwill-ms, line 70: readonly?

iothub/device/src/InternalClient.cs
```diff
@@ -1895,7 +1895,11 @@ public void Dispose()
...
+
+            if ((IotHubConnectionString?.TokenRefresher?.ShouldSdkDisposeInstance).GetValueOrDefault())
+            {
+                IotHubConnectionString?.TokenRefresher?.Dispose();
```
drwill-ms, line 1901: My point yesterday during chat was that, this logic could be centralized inside the `TokenRefresher.Dispose()` call, ensuring that logic doesn't need to be be right anywhere else.

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
```diff
@@ -36,6 +36,18 @@ internal class AmqpConnectionPool : IAmqpUnitManager
                 {
                     AmqpConnectionHolder[] amqpConnectionHolders = ResolveConnectionGroup(deviceIdentity);
                     amqpConnectionHolder = ResolveConnectionByHashing(amqpConnectionHolders, deviceIdentity);
+
+                    if (deviceIdentity.AuthenticationModel == AuthenticationModel.SasGrouped)
+                    {
+                        // For group sas token authenticated devices over a multiplexed connection, the identity
+                        // of the first device connecting will be used for generating the group sas tokens.
+                        // The TokenRefresher of the subsequently connected device identities can be safely disposed.
+                        if (!ReferenceEquals(amqpConnectionHolder.GetDeviceIdentityOfAuthenticationProvider(), deviceIdentity)
```
drwill-ms, line 45: Can you connect this logic to the comments above? I don't see the connection.

iothub/device/src/Transport/AmqpIot/AmqpIotCbsTokenProvider.cs
```diff
@@ -9,9 +9,10 @@
 
 namespace Microsoft.Azure.Devices.Client.Transport.AmqpIot
 {
-    internal class AmqpIotCbsTokenProvider : ICbsTokenProvider
+    internal class AmqpIotCbsTokenProvider : ICbsTokenProvider, IDisposable
     {
         private readonly IotHubConnectionString _connectionString;
+        private bool _disposedValue;
```
drwill-ms, line 15: `_isDisposed`?

iothub/device/src/Transport/AmqpIot/AmqpIotCbsTokenProvider.cs
```diff
@@ -57,5 +58,25 @@ public async Task<CbsToken> GetTokenAsync(Uri namespaceAddress, string appliesTo
...
+                if (disposing)
+                {
+                    _connectionString.TokenRefresher.Dispose();
+                }
+
+                _disposedValue = true;
+            }
+        }
+
+        public void Dispose()
```
drwill-ms, line 75: public methods above protected/internal/private

iothub/device/src/Transport/AmqpIot/AmqpIotCbsTokenProvider.cs
```diff
@@ -57,5 +58,25 @@ public async Task<CbsToken> GetTokenAsync(Uri namespaceAddress, string appliesTo
...
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            Dispose(disposing: true);
```
drwill-ms, line 78: Is the parameter label really needed?

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
drwill-ms: And why doesn't it include the `ShouldSdkDisposeInstance` check?

iothub/device/src/InternalClient.cs
abhipsaMisra: Do you mean that the check for "should this be disposed by sdk" be managed inside the `TokenRefresher.Dispose()` call? How would that work though? 
For scenarios where we don't initialize the instance, "should this be disposed by sdk" would resolve to false and we wouldn't release the managed resources. But then, how would the user make the call to release those resources? They would still call `TokenRefresher.Dispose()`, but the expression there would still evaluate to false.

Since the intended behavior on dispose is the same, wouldn't it make sense for the sdk to just evaluate if it should be calling dispose, and call it if required?

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
abhipsaMisra: If I go with this flow, I'll add some additional comments corelating the logic.

iothub/device/src/InternalClient.cs
drwill-ms: Good call. It would have to know who is making the call, which would require some tricks.

abhipsaMisra:
TODO: E2E tests with fault injection?

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -76,6 +76,8 @@ private void OnConnectionClosed(object o, EventArgs args)
             if (_amqpIotConnection != null && ReferenceEquals(_amqpIotConnection, o))
             {
                 _amqpAuthenticationRefresher?.StopLoop();
+                _amqpAuthenticationRefresher?.Dispose();
```
abhipsaMisra, line 79: is this recoverable? should this be disposed here?

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -100,6 +102,8 @@ public void Shutdown()
             }
 
             _amqpAuthenticationRefresher?.StopLoop();
+            _amqpAuthenticationRefresher?.Dispose();
```
abhipsaMisra, line 105: is this recoverable? should this be disposed here?

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -21,7 +21,7 @@
 
 using ClientOptions = Microsoft.Azure.Devices.Client.ClientOptions;
 
-namespace Microsoft.Azure.Devices.E2ETests.iothub
+namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
```
abhipsaMisra, line 24: minor tangential change - this test was placed under an incorrect namespace

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -173,7 +173,10 @@ private void Dispose(bool disposing)
             if (disposing)
             {
                 StopLoop();
-                _cancellationTokenSource.Dispose();
+                _cancellationTokenSource?.Dispose();
+                _cancellationTokenSource = null;
+
+                _amqpIotCbsTokenProvider?.Dispose();
```
abhipsaMisra, line 179: the token provider is disposed when the token refresher is disposed

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -206,19 +206,17 @@ public async Task<AmqpIotConnection> EnsureConnectionAsync(TimeSpan timeout)
                     // Create AmqpConnection
                     amqpIotConnection = await _amqpIotConnector.OpenConnectionAsync(timeout).ConfigureAwait(false);
 
-                    if (_deviceIdentity.AuthenticationModel != AuthenticationModel.X509)
+                    if (_deviceIdentity.AuthenticationModel == AuthenticationModel.SasGrouped)
```
abhipsaMisra, line 209: _deviceIdentity.AuthenticationModel is an enum, so we can simplify this check

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -850,6 +850,11 @@ private void Dispose(bool disposing)
                     _amqpConnectionHolder?.Dispose();
                 }
 
+                // For device sas authenticated clients the authentication refresher is associated with the AMQP unit itself,
+                // so it needs to be explicitly disposed.
+                _amqpAuthenticationRefresher?.StopLoop();
+                _amqpAuthenticationRefresher?.Dispose();
```
abhipsaMisra, line 856: stop and dispose the authentication refresher on disposal of an amqp unit

iothub/device/src/Transport/AmqpIot/AmqpIotCbsTokenProvider.cs
```diff
@@ -9,9 +9,10 @@
 
 namespace Microsoft.Azure.Devices.Client.Transport.AmqpIot
 {
-    internal class AmqpIotCbsTokenProvider : ICbsTokenProvider
+    internal class AmqpIotCbsTokenProvider : ICbsTokenProvider, IDisposable
```
abhipsaMisra, line 12: AmqpIotCbsTokenProvider contains an IDisposable type, so making it IDisposable as well

iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -538,6 +541,20 @@ public void Dispose()
...
+                // For eg. we create HttpTransportHandler instances for file upload operations even though the client might be
+                // initialized via MQTT/ AMQP. In those scenarios, since the shared TokenRefresher resource would be primarily used by the
+                // corresponding transport layers (MQTT/ AMQP), the diposal should be delegated to them and it should not be disposed here.
+                // The only scenario where the TokenRefresher should be disposed here is when the client has been initialized using HTTP.
+                if (_isClientPrimaryTransportHandler
+                    && _authenticationHeaderProvider is IotHubConnectionString iotHubConnectionString
+                    && iotHubConnectionString.TokenRefresher != null
+                    && iotHubConnectionString.TokenRefresher.InstanceCreatedBySdk)
+                {
+                    iotHubConnectionString.TokenRefresher.Dispose();
```
abhipsaMisra, line 555: if applicable, dispose the token refresher when the http client helper is disposed

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -1017,6 +1017,13 @@ private async void ShutdownAsync(IChannelHandlerContext context)
             }
             finally
             {
+                if (_passwordProvider is IotHubConnectionString iotHubConnectionString
+                    && iotHubConnectionString.TokenRefresher != null
+                    && iotHubConnectionString.TokenRefresher.InstanceCreatedBySdk)
+                {
+                    iotHubConnectionString.TokenRefresher.Dispose();
```
abhipsaMisra, line 1024: if applicable, dispose the token refresher when the mqtt iot hub adapter shuts down.
Note that even though we have a token refresher associated with the mqtt connection, it isn't a "true" refresher in the sense that it doesn't monitor token expiration and doesn't initiate a preemptive renewal. The token here is refreshed only on CONNECT (if expired).

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
```diff
@@ -36,6 +36,21 @@ internal class AmqpConnectionPool : IAmqpUnitManager
...
                     AmqpConnectionHolder[] amqpConnectionHolders = ResolveConnectionGroup(deviceIdentity);
                     amqpConnectionHolder = ResolveConnectionByHashing(amqpConnectionHolders, deviceIdentity);
+
+                    // For group sas token authenticated devices over a multiplexed connection, the TokenRefresher
+                    // of the first client connecting will be used for generating the group sas tokens
+                    // and will be associated with the connection itself.
+                    // For this reason, if the device identity of the client is not the one associated with the
+                    // connection, the associated TokenRefresher can be safely disposed.
+                    // Note - This does not cause any identity related issues since the group sas tokens are generated
+                    // against "{IoT hub name}.azure-devices.net/" as the intended audience (without the "device Id").
```
drwill-ms, line 46: Not if US Gov't cloud, or in the future with vanity domains.

drwill-ms (approved):

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
abhipsaMisra: In that case, I'll edit to to simply say "hub hostname".

doc/devguide.md
```diff
@@ -9,6 +9,11 @@ Before starting development on the Azure IoT SDK for C# you will need to install
 ### Design
 We are following the [Azure SDK design specification for C#](https://azuresdkspecs.z5.web.core.windows.net/DotNetSpec.html). To preserve backward compatibility, existing code will not change to follow these rules.
 
+#### Using [`IDisposable`](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-5.0#implementing-idisposable) types:
+- The sdk should dispose any `IDisposable` resource that it creates.
+- The sdk should not dispose an `IDisposable` resource that is supplied by the calling application. This is because the caller might want to reuse the resource elsewhere in the application. The responsibility of disposal of caller-supplied `IDisposable` resource is on the caller. An example of such a resource would be an `X509Certificate` instance that is used for authenticating our clients.
+- If the sdk implements `class A` that owns an `IDisposable` resource `X`, then `A` should also be `IDisposable`. In that case, resource `X` can be safely disposed when `class A` is disposed.
```
vinagesh, line 15: For the 3rd point, we still have to keep in mind 1st & 2nd point right? Don't dispose anything taken as an input from the user?

e2e/test/Helpers/AmqpConnectionStatusChange.cs
```diff
@@ -0,0 +1,36 @@
...
+        public AmqpConnectionStatusChange(string deviceId, MsTestLogger logger)
+        {
+            LastConnectionStatus = null;
+            LastConnectionStatusChangeReason = null;
+            ConnectionStatusChangeCount = 0;
+            _deviceId = deviceId;
+            _logger = logger;
+        }
+
+        public void ConnectionStatusChangesHandler(ConnectionStatus status, ConnectionStatusChangeReason reason)
```
vinagesh, line 22: Methods below properties.

doc/devguide.md
abhipsaMisra: yes, that is correct.

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
```diff
@@ -0,0 +1,252 @@
...
+            await deviceClient.CloseAsync();
+            deviceClient.Dispose();
+            Logger.Trace("Test with instance 1 completed");
+
+            // Perform the same steps again, reusing the previously created authentication method instance.
+            DeviceClient deviceClient2 = DeviceClient.Create(testDevice.IoTHubHostName, authenticationMethod, transport);
+            await deviceClient2.SendEventAsync(new Client.Message()).ConfigureAwait(false);
+            await deviceClient2.CloseAsync();
+            deviceClient2.Dispose();
+            Logger.Trace("Test with instance 2 completed");
```
vinagesh, line 84: Log statement could be more specific - Reused x509 cert provided for second instance successfully.

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
```diff
@@ -0,0 +1,252 @@
...
+            var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString);
+
+            // Create an instance of the device client, send a test message and then close and dispose it.
+            DeviceClient deviceClient = DeviceClient.Create(testDevice.IoTHubHostName, authenticationMethod, transport);
+            await deviceClient.SendEventAsync(new Client.Message()).ConfigureAwait(false);
+            await deviceClient.CloseAsync();
+            deviceClient.Dispose();
+            Logger.Trace("Test with instance 1 completed");
+
+            // Perform the same steps again, reusing the previously created authentication method instance.
```
vinagesh, line 79: Comment could be more specific - Perform the same steps again to ensure the SDK did not dispose the x509 certificate created by the test.

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
```diff
@@ -0,0 +1,252 @@
...
+            }
+
+            // Close and dispose client instance 1.
+            // The closed client should report a status of "disabled" while the rest of them should be connected.
+
+            await deviceClients[0].CloseAsync().ConfigureAwait(false);
+            deviceClients[0].Dispose();
+
+            amqpConnectionStatuses[0].LastConnectionStatus.Should().Be(ConnectionStatus.Disabled);
+
```
vinagesh, line 135: Could also add amqpConnectionStatuses[1].LastConnectionStatus.Should().Be(ConnectionStatus.Connected);

abhipsaMisra: we actually check the status of the entire list of client instances right below

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
```diff
@@ -0,0 +1,252 @@
...
+                var amqpConnectionStatusChange = new AmqpConnectionStatusChange(testDevices[i].Id, Logger);
+                deviceClient.SetConnectionStatusChangesHandler(amqpConnectionStatusChange.ConnectionStatusChangesHandler);
+                amqpConnectionStatuses.Add(amqpConnectionStatusChange);
+
+                await deviceClient.OpenAsync().ConfigureAwait(false);
+                deviceClients.Add(deviceClient);
+            }
+
+            // Ensure that all clients are connected successfully, and the close and dispose the instances.
```
vinagesh, line 186: Comment could be more specific - Reused x509 cert provided for second instance successfully.

iothub/device/src/IotHubConnectionString.cs
```diff
@@ -75,6 +84,10 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
 
                 Debug.Assert(TokenRefresher != null);
             }
+            else if (!string.IsNullOrWhiteSpace(builder.SharedAccessSignature))
+            {
+                SharedAccessSignature = builder.SharedAccessSignature;
```
vinagesh, line 89: If builder.SharedAccessSignature is null, SharedAccessSignature will remain the default null. Why not just do, SharedAccessSignature = builder.SharedAccessSignature like it was earlier?

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
vinagesh: What about AuthenticationModel.SasIndividual ? Does it not apply to that?

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
```diff
@@ -36,6 +36,21 @@ internal class AmqpConnectionPool : IAmqpUnitManager
...
                     amqpConnectionHolder = ResolveConnectionByHashing(amqpConnectionHolders, deviceIdentity);
+
+                    // For group sas token authenticated devices over a multiplexed connection, the TokenRefresher
+                    // of the first client connecting will be used for generating the group sas tokens
+                    // and will be associated with the connection itself.
+                    // For this reason, if the device identity of the client is not the one associated with the
+                    // connection, the associated TokenRefresher can be safely disposed.
+                    // Note - This does not cause any identity related issues since the group sas tokens are generated
+                    // against the hub host as the intended audience (without the "device Id").
+                    if (deviceIdentity.AuthenticationModel == AuthenticationModel.SasGrouped
```
vinagesh, line 47: Is there no way to avoid creating these for mux scenarios instead of disposing the unnecessary ones? Looks like all we need is a 1SAS token for the connection irrespective of the devices in the mux.

iothub/device/src/Transport/AmqpIot/AmqpIotCbsTokenProvider.cs
```diff
@@ -57,5 +58,29 @@ public async Task<CbsToken> GetTokenAsync(Uri namespaceAddress, string appliesTo
...
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_isDisposed)
+            {
+                if (disposing)
+                {
+                    if (_connectionString?.TokenRefresher != null
+                        && _connectionString.TokenRefresher.InstanceCreatedBySdk)
+                    {
+                        _connectionString.TokenRefresher.Dispose();
```
vinagesh, line 78: _connectionString.TokenRefresher = null after diposing as a best practice?

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
```diff
@@ -0,0 +1,252 @@
...
+            private static string GetDeviceIdFromConnectionString(string connectionString)
+            {
+                if (connectionString == null)
+                {
+                    throw new ArgumentNullException(nameof(connectionString));
+                }
+
+                var builder = IotHubConnectionStringBuilder.Create(connectionString);
+                return builder.DeviceId;
+            }
```
vinagesh, line 249: These tests are doing a lot more than just checking for reusing the disposable created external to the SDK. Is that because we don't have tests that cover these mux scenarios? If they already exist, it would be nice simplify to just testing the reusability.

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
abhipsaMisra: Well, the goal here is to ensure that the client is connected => the tcp connection is established. All the resources that this tcp connection depends on will thus inherently need to be in proper state. I don't want to call out a specific resource here, since that may be misleading in the sense that it is not the only resource whose state is important. Does that make sense?

iothub/device/src/IotHubConnectionString.cs
abhipsaMisra: The reason I moved it here is because the SharedAccessSignature should be set only if the there isn't an AuthenticationWithTokenRefresh authentication method already supplied to the client; i.e. it is placed in the `else` clase.

Reasoning:
"IotHubConnectionString.SharedAccessSignature should be set only if it is non-null and the authentication method of the device client is not of type AuthenticationWithTokenRefresh. Setting the sas value for an AuthenticationWithTokenRefresh authentication type will result in tokens not being renewed. This flow can be hit if the same authentication method is always used to initialize the client; as in, on disposal and reinitialization. (the value of the sas token computed is stored within the authentication method, and on reinitialization the client is incorrectly identified as a fixed-sas-token-initialized client, instead of being identified as a sas-token-refresh-enabled-client.)"

I'll add this to the code comments, just so we don't lose the context in the future.

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
abhipsaMisra: There isn't, no. The client is "set up" in a different layer (in internal client, before the whole concept of transport pipeline comes into picture), and this is where the token refresher is set up. This layer doesn't have any visibility about multiplexing, which is a transport-level construct.

iothub/device/src/Transport/AmqpIot/AmqpIotCbsTokenProvider.cs
abhipsaMisra: I agree, but in this case the _connectionString is readonly and hence, set up only in the constructor. I didn't want to modify it, so as to protect against accidental reinitialization in the future.

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
abhipsaMisra: We have tests that cover basic mux and pooling scenarios (all operations), and that cover fault injection when devices have opted for multiplexing (and pooling).
These new tests are designed to reuse the authentication method, while still testing that disposal on one multiplexed device doesn't cause cascading failures in the rest of the devices on the same tcp connection.

doc/devguide.md
vinagesh: In that case, it might be good to just put this point on top and say - If the sdk implements `class A` that owns an `IDisposable` resource `X`, then `A` should also be `IDisposable`. The other two points seem to cover when to dispose something.

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
abhipsaMisra: for sas individual authenticated devices, the token refresher is associated with the amqp unit: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Amqp/AmqpUnit.cs#L119

sas individual => amqp unit
sas group => amqp connection
x509 => no token refresher (and also, no multiplexing)

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
vinagesh: Sure but also good to call out the point about reusing the certs wherever you are reusing it in the tests.

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
vinagesh: This seems more explanatory. Can you just add this in comments?

vinagesh (approved):

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
abhipsaMisra: sure, I'll add this comment in

PR #1955, opened 2021-05-12
fix(device): Fix the exception type that is returned when a token cancellation happens

vinagesh:
## Checklist

## Description of the changes
Due to changes in 'InternalClient.cs' in this commit - https://github.com/Azure/azure-iot-sdk-csharp/commit/d17e27fe531a4c746a3eb97fa5e6a8fd040ec649, we were returning a different exception type on token cancellation. This is a breaking change to the customer though our current implementation is correct. I'm making changes so that we retain the correct usage or async await but return the same exception as we returned earlier.

Old code example:

If you look at the code below, because of the lack of await, instead of the code hitting the catch block, it returns to the caller with IotHubCommunicationException if there was a token cancellation.

```csharp
// Note: No async here
public Task<Message> ReceiveAsync(CancellationToken cancellationToken)
{
    try
    {
        // Note: No await here
        return InnerHandler.ReceiveAsync(cancellationToken);
    }
    catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
    {
        throw;
    }
}
```

New code:

In you look at the code below, because of the presence of await, we now enter the catch and return OperationCancelledException on token cancellation. 
This is potentially a breaking change and it is failing the MCVP pipeline. To ensure we continue to use async await, instead of returning `cancellationToken.ThrowIfCancellationRequested()`, I've just changed it to `throw` whatever exception was caught.

```csharp
// Note: Added async here
public async Task<Message> ReceiveAsync(CancellationToken cancellationToken)
{
    try
    {
       // Note: added await here
        return await InnerHandler.ReceiveAsync(cancellationToken).ConfigureAwait(false);
    }
    catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
    {
        cancellationToken.ThrowIfCancellationRequested(); // Note: This line has been removed as a fix.
        throw;
    }
}
```

If you think this is a breaking change that is acceptable, please let me know and we can initiate conversations with MCVP. However, I think this will also break any customers who have already built logic around these many APIs catching the incorrect exception. 

I have added a test for just the one API failing on MCVP pipeline. There were a lot of APIs and I will create a task to evaluate and add tests for whatever is necessary.

### DIFF
--- a/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
+++ b/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Client.Exceptions;
 using Microsoft.Azure.Devices.Client.Transport.Mqtt;
 using Microsoft.Azure.Devices.E2ETests.Helpers;
 using Microsoft.Azure.Devices.E2ETests.Helpers.Templates;
@@ -106,6 +107,12 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
         public async Task Message_DeviceReceiveSingleMessageWithCancellationToken_Mqtt()
         {
             await ReceiveSingleMessageWithCancellationTokenAsync(TestDeviceType.Sasl, Client.TransportType.Mqtt_Tcp_Only).ConfigureAwait(false);
+        }
+
+        [LoggedTestMethod]
+        public async Task Message_DeviceReceiveSingleMessageCancelled_Mqtt()
+        {
+            await Mqtt_ReceiveSingleMessageWithCancelledAsync(TestDeviceType.Sasl).ConfigureAwait(false);
         }
 
         [LoggedTestMethod]
@@ -515,6 +522,22 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
             Assert.IsTrue(received, $"No message received for device {deviceId} with payload={payload} in {FaultInjection.RecoveryTime}.");
         }
 
+        public static async Task Mqtt_VerifyReceivedC2dMessageCancelledAsync(DeviceClient dc)
+        {
+            try
+            {
+                using var cts = new CancellationTokenSource(s_fiveSeconds);
+                await dc.ReceiveAsync(cts.Token).ConfigureAwait(false);
+            }
+            catch (IotHubCommunicationException ex)
+                when (ex.IsTransient && ex.InnerException is OperationCanceledException)
+            {
+                return;
+            }
+
+            Assert.Fail();
+        }
+
         public static async Task VerifyReceivedC2dMessageWithCancellationTokenAsync(Client.TransportType transport, DeviceClient dc, string deviceId, string payload, string p1Value, MsTestLogger logger)
         {
             var sw = new Stopwatch();
@@ -625,6 +648,19 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
 
             await deviceClient.CloseAsync().ConfigureAwait(false);
             await serviceClient.CloseAsync().ConfigureAwait(false);
+        }
+
+        private async Task Mqtt_ReceiveSingleMessageWithCancelledAsync(TestDeviceType type)
+        {
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, s_devicePrefix, type).ConfigureAwait(false);
+            using DeviceClient deviceClient = testDevice.CreateDeviceClient(Client.TransportType.Mqtt_Tcp_Only);
+
+            await deviceClient.OpenAsync().ConfigureAwait(false);
+
+            // There is no message being sent so the device client should timeout waiting for the message.
+            await Mqtt_VerifyReceivedC2dMessageCancelledAsync(deviceClient).ConfigureAwait(false);
+
+            await deviceClient.CloseAsync().ConfigureAwait(false);
         }
 
         private async Task ReceiveSingleMessageWithCancellationTokenAsync(TestDeviceType type, Client.TransportType transport)
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -312,11 +312,11 @@ namespace Microsoft.Azure.Devices.Client
         /// Explicitly open the InternalClient instance.
         /// </summary>
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
-        public async Task OpenAsync(CancellationToken cancellationToken)
+        public Task OpenAsync(CancellationToken cancellationToken)
         {
             try
             {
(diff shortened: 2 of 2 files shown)


abhipsaMisra:
Question:
- is there a reason why we need to await the inner handler call? From what I know, every time we add an await statement, it will create a new task internally and the size of the stack will increase. It will also result in an increase in size of the final compiled code. Since this API call simply returns the response of another async API call, `return InnerHandler.ReceiveAsync(cancellationToken);` should have been enough.

abhipsaMisra:
Or are we introducing this behavior so that we can manage the exception handling code better?

iothub/device/src/InternalClient.cs
```diff
@@ -320,7 +320,6 @@ public async Task OpenAsync(CancellationToken cancellationToken)
             }
             catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
```
abhipsaMisra, line 321: slightly tangential note (pending answer to the other question), but this block doesn't really do anything anymore, so might as well remove the `catch` block entirely.

vinagesh: Which other question? Did you post on this PR? I agree, this catch block can be removed. It never did anything earlier as well.

abhipsaMisra: Yeah, I meant this one: https://github.com/Azure/azure-iot-sdk-csharp/pull/1955#issuecomment-839925262

vinagesh: I see it now. It was not showing up till now.

vinagesh:
> Question:
> 
> * is there a reason why we need to await the inner handler call? From what I know, every time we add an await statement, it will create a new task internally and the size of the stack will increase. It will also result in an increase in size of the final compiled code. Since this API call simply returns the response of another async API call, `return InnerHandler.ReceiveAsync(cancellationToken);` should have been enough.

This change was made in an earlier commit, that caused a change in behavior. However my understanding is it is best practice to have async all the way down. There is a high possibility of exception being lost if we await nowhere in the full call stack. Also if some bad practices are followed, we could have deadlocks too. The other thing to note was we had incorrect code even prior to adding the await. It seemed like we always intended to catch this error and do something about it but because we were not awaiting, we were doing something unexpected. I think it is good practice to have async all the way down. It sure increases the compiled code by a bit but it has advantages.

vinagesh:
> Or are we introducing this behavior so that we can manage the exception handling code better?

I believe this change was made in the earlier commit to adapt to the best practices of async await. @drwill made the change so i'll let him comment on it.

abhipsaMisra:
I agree that we should await an async call when we want to either do something with the result or handle exceptions that might arise from the async call, but we aren't really handling the exception here, right? If we were inspecting the inspection to follow different paths based on the exception caught, then awaiting the async call would make sense. But since that breaks the behavior, we decided to not inspect any exception caught and just throw it instead. So, for this scenario it doesn't really make sense for us to await the call, right?
Is there some scenario that I missed?

vinagesh:
> I agree that we should await an async call when we want to either do something with the result or handle exceptions that might arise from the async call, but we aren't really handling the exception here, right? If we were inspecting the inspection to follow different paths based on the exception caught, then awaiting the async call would make sense. But since that breaks the behavior, we decided to not inspect any exception caught and just throw it instead. So, for this scenario it doesn't really make sense for us to await the call, right?
> Is there some scenario that I missed?

If you look the earlier code, you can clearly see that we did not await and still made coding mistakes of trying to catch the exception. It seems to me like a bad coding practice to not await and could lead to bugs. Adding await makes the code safer and easier to extend whenever necessary. Why do you think we should not await? I don't see a disadvantage to it. If there is a different practice followed when coding for SDKs, I would like to know more about it.

iothub/device/src/InternalClient.cs
vinagesh: Fyi: That block never did anything so it is safe to remove at the moment.

abhipsaMisra:
My hesitancy for awaiting these calls is because it isn't necessary for this scenario -> we aren't inspecting the result nor are we handling any possible exceptions that might be thrown. Instead, we are simply increasing the call stack by introducing another Task, but there is no benefit to it.
As for sdk best practices, I couldn't find anything called out specifically in the track2 guidelines doc, but I am sure if we reached out to the track2 team, they'd give us a recommendation. (iirc, even in our ADT client we were not awaiting the calls to the protocol layer, unless we wanted to either transform the result from the protocol layer call, or we wanted to add special exception handling logic).

Again, this is my 2c. I am happy to be convinced otherwise! :)

vinagesh:
> My hesitancy for awaiting these calls is because it isn't necessary for this scenario -> we aren't inspecting the result nor are we handling any possible exceptions that might be thrown. Instead, we are simply increasing the call stack by introducing another Task, but there is no benefit to it.
> As for sdk best practices, I couldn't find anything called out specifically in the track2 guidelines doc, but I am sure if we reached out to the track2 team, they'd give us a recommendation. (iirc, even in our ADT client we were not awaiting the calls to the protocol layer, unless we wanted to either transform the result from the protocol layer call, or we wanted to add special exception handling logic).
> 
> Again, this is my 2c. I am happy to be convinced otherwise! :)

After inspecting the code more, I see that our DeviceClient has no async or await. So basically there is no all the way down asyn at all unless we change the DeviceClient (which is not what I had expected to see). So I guess the naming is weird, ending with async. It seems fine to remove the await and the catch block completely. When the user does await on any of the DeviceClient methods, they will get the exception thrown by the method.

abhipsaMisra:
I agree, I think we might have intended to do what we have now -> await the async call and handle the exception caught; but it was implemented incorrectly.

```csharp
    try
    {
        return await InnerHandler.ReceiveAsync(cancellationToken).ConfigureAwait(false);
    }
    catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
    {
        throw;
    }
```
Since putting in the correct fix would be breaking behavior (even though it is desirable), the safer option is to remove the unnecessary catch block altogether.

e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
```diff
@@ -627,6 +650,34 @@ private async Task ReceiveSingleMessageAsync(TestDeviceType type, Client.Transpo
             await serviceClient.CloseAsync().ConfigureAwait(false);
         }
 
+        private async Task Mqtt_ReceiveSingleMessageWithCancelledAsync(TestDeviceType type)
+        {
+            TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, s_devicePrefix, type).ConfigureAwait(false);
```
abhipsaMisra, line 655: I think this test can be simplified to just have a new device client instance make a deviceClient.ReceiveAsync(cts.Token) call and assert on the exception, right? Meaning, we don't even need the service client side of the flow, right?

abhipsaMisra (approved):

drwill-ms (approved):

e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
vinagesh: yes makes sense. Let me simply it.

PR #1961, opened 2021-05-17
doc(service): Added links to managed identity and AAD docs

vinagesh:
## Checklist

### DIFF
--- a/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
+++ b/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
@@ -103,6 +103,9 @@ namespace Microsoft.Azure.Devices
         /// <param name="credential">Azure Active Directory credentials to authenticate with IoT hub. See <see cref="TokenCredential"/></param>
         /// <param name="handlers">The delegating handlers to add to the http client pipeline. You can add handlers for tracing, implementing a retry strategy, routing requests through a proxy, etc.</param>
         /// <returns>An instance of <see cref="DigitalTwinClient"/>.</returns>
+        /// <remarks>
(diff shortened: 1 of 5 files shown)


iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -98,6 +98,7 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
 
         /// <summary>
         /// Creates an instance of <see cref="DigitalTwinClient"/>.
+        /// For more information on configuring IoT hub with Azure Active Directory, <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-dev-guide-azure-ad-rbac"/>
```
drwill-ms, line 101: This feels like a statement that belongs in a `<remarks>` tag. Thoughts?

drwill-ms, line 101: Remember that the `<see>` tag does not insert a verb. So this would come out as:
> For more information on configuring IoT hub with Azure Active Directory, https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-dev-guide-azure-ad-rbac

You'll want to add your own verb, and give it puncuation.

drwill-ms (approved):

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
vinagesh: Good point. I'll move it to remarks.

timtay-microsoft (approved):

abhipsaMisra (approved):

PR #1962, opened 2021-05-17
Move the proper classes to the shared assembly and namespace.

jamdavi:


iothub/device/src/ClientTwinProperties.cs
```diff
@@ -19,5 +21,14 @@ internal ClientTwinProperties()
 
         [JsonProperty(PropertyName = "reported", DefaultValueHandling = DefaultValueHandling.Ignore)]
         internal IDictionary<string, object> Reported { get; set; }
+
+        internal ClientProperties ToClientProperties(PayloadConvention payloadConvention)
```
jamdavi, line 25: This allows us to pull ClientPropertyCollection out of the Client assembly

shared/src/ClientPropertyCollection.cs
```diff
@@ -23,8 +21,8 @@ public class ClientPropertyCollection : PayloadCollection
         /// <remarks>
         /// If the collection has a key that matches the property name this method will throw an <see cref="ArgumentException"/>.
         /// <para>
-        /// When using this as part of the <see cref="DeviceClient.UpdateClientPropertiesAsync(ClientPropertyCollection, System.Threading.CancellationToken)"/>
-        /// flow to respond to a writable property update you should pass in the value as an instance of <see cref="PayloadSerializer.CreateWritablePropertyResponse(object, int, long, string)"/>
+        /// When using this as part of the writable property flow to respond to a writable property update you should pass in the value
```
jamdavi, line 24: No reference to this since we're in the shared class now

shared/src/PayloadSerializer.cs
```diff
@@ -42,7 +42,7 @@ public abstract class PayloadSerializer
         /// <typeparam name="T">The type to convert to.</typeparam>
         /// <param name="objectToConvert">The object to convert.</param>
         /// <returns>A converted object</returns>
-        /// <remarks>This class is used by the <see cref="ClientPropertyCollection"/> to attempt to convert from the native serailizer type (for example, JObject or JsonElement) to the desired type. When you implement this you need to be aware of what type your serializer will use for anonymous types.</remarks>
+        /// <remarks>This class is used by the <see cref="PayloadCollection"/> class to attempt to convert from the native serailizer type (for example, JObject or JsonElement) to the desired type. When you implement this you need to be aware of what type your serializer will use for anonymous types.</remarks>
```
jamdavi, line 45: This is actually used by the PayloadCollection class and not the specific client collection

shared/src/ClientPropertyCollection.cs
```diff
@@ -2,12 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using Microsoft.Azure.Devices.Shared;
-using Newtonsoft.Json.Linq;
 
-namespace Microsoft.Azure.Devices.Client
+namespace Microsoft.Azure.Devices.Shared
```
jamdavi, line 8: Back to the conversation earlier, does it make more sense to have this inside of the Client class? My thought is no since we can use this as part of the ModuleClient.

shared/src/DebugBuildConfiguration.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Microsoft.Azure.Devices.Client")]
```
abhipsaMisra, line 6: ahh, this will be required!
we'll also need to add the public key, since both of these are strong-name signed: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-5.0

iothub/device/src/DeviceClient.cs
```diff
@@ -23,7 +23,7 @@ public class DeviceClient : IDisposable
         public const uint DefaultOperationTimeoutInMilliseconds = 4 * 60 * 1000;
 
         /// <summary>
-        /// The object serializer used by the <see cref="Client.PayloadConvention"/> set in <see cref="ClientOptions.PayloadConvention"/>. If no <see cref="Client.PayloadConvention"/> is set this will default to <see cref="NewtonsoftJsonPayloadSerializer"/>.
+        /// The object serializer used by the <see cref="Shared.PayloadConvention"/> set in <see cref="ClientOptions.PayloadConvention"/>. If no <see cref="Shared.PayloadConvention"/> is set this will default to <see cref="NewtonsoftJsonPayloadSerializer"/>.
```
abhipsaMisra, line 26: isn't "NewtonsoftJsonPayloadSerializer" a part of shared as well?

abhipsaMisra (approved):
some small comments, else looks good

shared/src/ClientPropertyCollection.cs
abhipsaMisra: to put the client property collection in device client class? why?

iothub/device/src/DeviceClient.cs
jamdavi: Yes, this is because PayloadConvention is ambiguous because the class has a property named that as well.

shared/src/ClientPropertyCollection.cs
jamdavi: No reason, just wanted to make sure we're aligned.

shared/src/DebugBuildConfiguration.cs
abhipsaMisra: another option is to add a compile include for these "common" files to the respective assembly (device client nuget), instead of putting them in the shared library. what do you think?

jamdavi: No, I truly **loathe** this approach and would like to avoid it at all costs. If we planned to do that we need to create a shared project

abhipsaMisra (approved):

PR #1963, opened 2021-05-17
Move the componentName to the front of the method

jamdavi:


shared/src/ClientPropertyCollection.cs
```diff
@@ -14,7 +14,7 @@ public class ClientPropertyCollection : PayloadCollection
     {
         private const string VersionName = "$version";
 
-        /// <inheritdoc path="/exception['ArgumentException']" cref="Add(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
```
jamdavi, line 17: It was getting a bit confusing with all of the Adds

shared/src/ClientPropertyCollection.cs
```diff
@@ -158,7 +196,7 @@ public void AddOrUpdate(string propertyName, object propertyValue, int statusCod
         /// <param name="properties">A collection of properties to add or update.</param>
         /// <param name="componentName">The component with the properties to add or update.</param>
         /// <param name="forceUpdate">Forces the collection to use the Add or Update behavior. Setting to true will simply overwrite the value; setting to false will use <see cref="IDictionary{TKey, TValue}.Add(TKey, TValue)"/></param>
-        private void Add(IDictionary<string, object> properties, string componentName = default, bool forceUpdate = false)
+        private void AddInternal(IDictionary<string, object> properties, string componentName = default, bool forceUpdate = false)
```
jamdavi, line 212: It was getting a bit confusing with all of the Adds

shared/src/ClientPropertyCollection.cs
```diff
@@ -73,44 +97,42 @@ public void Add(IDictionary<string, object> properties, string componentName = d
         /// <param name="version"></param>
         /// <param name="description"></param>
         /// <param name="componentName"></param>
-        public void Add(string propertyName, object propertyValue, int statusCode, long version, string description = default, string componentName = default)
+        public void Add(string componentName, string propertyName, object propertyValue, int statusCode, long version, string description = default)
         {
             if (Convention?.PayloadSerializer == null)
```
abhipsaMisra, line 102: q = would this ever be null, we'd always default this to newtonsoft, right?

abhipsaMisra: actually nvm, this is right -> null when created from scratch and value available if reusing from a returned property.

abhipsaMisra (approved):

PR #1965, opened 2021-05-18
Bring updates from master to preview

abhipsaMisra:


timtay-microsoft (approved):

iothub/service/src/ServiceClient.cs
```diff
@@ -71,23 +78,44 @@ public ServiceClient()
...
+            IotHubConnectionProperties connectionProperties,
+            bool useWebSocketOnly,
+            ServiceClientTransportSettings transportSettings,
+            ServiceClientOptions options)
         {
-            Connection = connection;
-            _iotHubName = iotHubName;
-            _clientOptions = options;
-            _httpClientHelper = httpClientHelper;
+            var iotHubConnection = new IotHubConnection(connectionProperties, useWebSocketOnly, transportSettings);
```
vinagesh, line 87: Combine (Line 87, 88) to one statement : Connection = new IotHubConnection(...)

iothub/service/src/ServiceClient.cs
```diff
@@ -71,23 +78,44 @@ public ServiceClient()
...
+                s_defaultOperationTimeout,
+                transportSettings.HttpProxy,
+                transportSettings.ConnectionLeaseTimeoutMilliseconds);
 
             // Set the trace provider for the AMQP library.
             AmqpTrace.Provider = new AmqpTransportLog();
         }
 
+        // internal test helper
+        internal ServiceClient(IotHubConnection connection, IHttpClientHelper httpClientHelper)
```
vinagesh, line 110: I observed in all clients this is back now. It is ok but the pattern of having internal constructors just for tests is not the best pattern to follow. And the fact that the what is being passed into the other constructor in not mockble, is also not a great pattern. For our codebase it is not really problem so we can leave it as is for now.

vinagesh (approved):

vinagesh:
Looks good!

iothub/service/src/ServiceClient.cs
abhipsaMisra: I agree, the way these clients are designed, the only way to avoid this internal constructor is to have a lot of duplicate code (declare the mockable resources in each call and have a common initializer that accepts the mockable resources); so this is the workaround for now. But definitely something we should keep in mind as we design new clients. :)

PR #1967, opened 2021-05-18
feat(service): Add AAD auth support to Azure US Government cloud.

vinagesh:
## Checklist

## Description of the changes
Adding AAD support for Fairfax

### DIFF
--- a/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
+++ b/iothub/service/src/DigitalTwin/DigitalTwinClient.cs
@@ -105,23 +105,56 @@ namespace Microsoft.Azure.Devices
         /// <returns>An instance of <see cref="DigitalTwinClient"/>.</returns>
         /// <remarks>
         /// For more information on configuring IoT hub with Azure Active Directory, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-dev-guide-azure-ad-rbac"/>
+        /// This constructor internally uses <see cref="IotHubAuthenticationScopes.DefaultAuthenticationScopes"/>, which is used for
+        /// any public or private cloud other than Azure US Government cloud.
+        /// For Azure US Government cloud users, use the constructor with <see cref="DigitalTwinClientOptions"/> and set the
+        /// <see cref="DigitalTwinClientOptions.TokenCredentialAuthenticationScopes"/> to <see cref="IotHubAuthenticationScopes.AzureGovernmentAuthenticationScopes"/>.
         /// </remarks>
         public static DigitalTwinClient Create(
             string hostName,
             TokenCredential credential,
             params DelegatingHandler[] handlers)
         {
+            return Create(hostName, credential, null, handlers);
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="DigitalTwinClient"/>.
+        /// </summary>
+        /// <param name="hostName">IoT hub host name.</param>
+        /// <param name="credential">Azure Active Directory credentials to authenticate with IoT hub. See <see cref="TokenCredential"/></param>
+        /// <param name="options">Options that allow configuration of the DigitalTwinClient instance during initialization.</param>
+        /// <param name="handlers">The delegating handlers to add to the http client pipeline. You can add handlers for tracing, implementing a retry strategy, routing requests through a proxy, etc.</param>
+        /// <returns>An instance of <see cref="DigitalTwinClient"/>.</returns>
+        /// <remarks>
+        /// For more information on configuring IoT hub with Azure Active Directory, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-dev-guide-azure-ad-rbac"/>
+        /// This constructor sets the default for <see cref="DigitalTwinClientOptions.TokenCredentialAuthenticationScopes"/> to
+        /// <see cref="IotHubAuthenticationScopes.DefaultAuthenticationScopes"/>, which is used for any public or private cloud other than Azure US Government cloud.
+        /// For Azure US Government cloud users, set the <see cref="DigitalTwinClientOptions.TokenCredentialAuthenticationScopes"/>
+        /// to <see cref="IotHubAuthenticationScopes.AzureGovernmentAuthenticationScopes"/>.
+        /// </remarks>
+        public static DigitalTwinClient Create(
+            string hostName,
+            TokenCredential credential,
+            DigitalTwinClientOptions options,
+            params DelegatingHandler[] handlers)
+        {
             if (string.IsNullOrEmpty(hostName))
             {
-                throw new ArgumentNullException($"{nameof(hostName)},  Parameter cannot be null or empty");
+                throw new ArgumentNullException(nameof(hostName), "Parameter cannot be null or empty.");
             }
 
             if (credential == null)
             {
-                throw new ArgumentNullException($"{nameof(credential)},  Parameter cannot be null");
+                throw new ArgumentNullException(nameof(credential));
             }
 
-            var tokenCredential = new DigitalTwinTokenCredential(credential);
+            if (options == null)
+            {
+                options = new DigitalTwinClientOptions();
+            }
+
+            var tokenCredential = new DigitalTwinTokenCredential(credential, options.TokenCredentialAuthenticationScopes);
             return new DigitalTwinClient(hostName, tokenCredential, handlers);
         }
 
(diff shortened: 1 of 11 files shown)


iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -121,7 +125,47 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+        /// <param name="options">Options that allow configuration of the DigitalTwinClient instance during initialization.</param>
+        /// <param name="handlers">The delegating handlers to add to the http client pipeline. You can add handlers for tracing, implementing a retry strategy, routing requests through a proxy, etc.</param>
+        /// <returns>An instance of <see cref="DigitalTwinClient"/>.</returns>
+        /// <remarks>
+        /// For more information on configuring IoT hub with Azure Active Directory, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-dev-guide-azure-ad-rbac"/>
+        /// This constructor sets the default for <see cref="DigitalTwinClientOptions.TokenCredentialAuthenticationScopes"/>, to
+        /// <see cref="IotHubAuthenticationScopes.DefaultAuthenticationScopes"/> which is used for any public or private cloud other than Fairfax.
+        /// For Fairfax cloud users, set the <see cref="DigitalTwinClientOptions.TokenCredentialAuthenticationScopes"/>
+        /// to <see cref="IotHubAuthenticationScopes.FairfaxAuthenticationScopes"/>.
```
vinagesh, line 145: We need this override due to usage of params in this client.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -105,6 +105,10 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
         /// <returns>An instance of <see cref="DigitalTwinClient"/>.</returns>
         /// <remarks>
         /// For more information on configuring IoT hub with Azure Active Directory, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-dev-guide-azure-ad-rbac"/>
+        /// This constructor internally uses <see cref="IotHubAuthenticationScopes.DefaultAuthenticationScopes"/>, which is used for
+        /// any public or private cloud other than Fairfax.
+        /// For Fairfax cloud users, use the constructor with <see cref="DigitalTwinClientOptions"/> and set the
```
vinagesh, line 110: I had missed yesterday's conversation on the official name for Fairfax. I will be updating the full PR now by replacing it with AzureGoverment instead.

e2e/test/iothub/TokenCredentialAuthenticationTests.cs
```diff
@@ -103,9 +103,15 @@ public async Task DigitalTwinClient_Http_TokenCredentialAuth_Success()
             // Call openAsync() to open the device's connection, so that the ModelId is sent over Mqtt CONNECT packet.
             await deviceClient.OpenAsync().ConfigureAwait(false);
 
+            var dtClientOptions = new DigitalTwinClientOptions
+            {
+                TokenCredentialAuthenticationScopes = IotHubAuthenticationScopes.DefaultAuthenticationScopes
```
timtay-microsoft, line 108: Is this necessary? I'd expect that the ```DigitalTwinClient.Create()``` call with no options would default to this value anyways

iothub/service/src/RegistryManager.cs
```diff
@@ -62,14 +62,20 @@ public static RegistryManager CreateFromConnectionString(string connectionString
...
+        /// <see cref="IotHubAuthenticationScopes.DefaultAuthenticationScopes"/>, which is used for any public or private cloud other than Azure US Government cloud.
+        /// For Azure US Government cloud users, set the <see cref="RegistryManagerOptions.TokenCredentialAuthenticationScopes"/>
+        /// to <see cref="IotHubAuthenticationScopes.AzureGovernmentAuthenticationScopes"/>.
         /// </remarks>
         public static RegistryManager Create(
             string hostName,
             TokenCredential credential,
-            HttpTransportSettings transportSettings = default)
+            HttpTransportSettings transportSettings = default,
+            RegistryManagerOptions options = default)
```
timtay-microsoft, line 78: Having two options classes here and elsewhere feels a bit odd to me. What about having RegistryManagerOptions contain all the same settings as HttpTransportSettings and then add the new authentication scopes setting.

```csharp
// leave old constructor alone, maybe deprecate it
public static RegistryManager Create(
    string hostName,
    TokenCredential credential,
    HttpTransportSettings transportSettings = default)

// newly added constructor
public static RegistryManager Create(
    string hostName,
    TokenCredential credential,
    RegistryManagerOptions options = default)
```

timtay-microsoft (approved):

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -121,7 +125,47 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+        /// </remarks>
+        public static DigitalTwinClient Create(
+            string hostName,
+            TokenCredential credential,
+            DigitalTwinClientOptions options,
+            params DelegatingHandler[] handlers)
+        {
+            if (string.IsNullOrEmpty(hostName))
+            {
+                throw new ArgumentNullException($"{nameof(hostName)},  Parameter cannot be null or empty");
```
drwill-ms, line 155: Please use a different ctor overload: `ArgumentNullException(nameof(hostName), "Parameter cannot be null or empty.")`

This treats the name of the parameter special instead of formatted into the error message.

e2e/test/iothub/TokenCredentialAuthenticationTests.cs
vinagesh: Not necessary, probably belongs more in the sample. I was thinking in terms of how to demonstrate usage when I added this.

iothub/service/src/RegistryManager.cs
vinagesh: It is an option but it would introduce another overload, which leads to a user trying to think which to use. Everyone is already used to using TransportSettings and moving the same things to another clientOptions class seemed confusing to me. TransportSettings seemed to too specific so I created the other options class instead of adding it there.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
vinagesh: This was probably already there, but I'll update it.

vinagesh: Updated all that I could fine.

e2e/test/iothub/TokenCredentialAuthenticationTests.cs
vinagesh: I had a better idea to test the new overload. I am calling the new overload from the old one so even without this, it should be a good test now.

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
```diff
@@ -105,23 +105,56 @@ public static DigitalTwinClient CreateFromConnectionString(string connectionStri
...
+        public static DigitalTwinClient Create(
+            string hostName,
+            TokenCredential credential,
+            DigitalTwinClientOptions options,
+            params DelegatingHandler[] handlers)
         {
             if (string.IsNullOrEmpty(hostName))
             {
-                throw new ArgumentNullException($"{nameof(hostName)},  Parameter cannot be null or empty");
+                throw new ArgumentNullException(nameof(hostName));
```
drwill-ms, line 144: For this one (because empty check) prefer the 2 param overload: https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=net-5.0#System_ArgumentNullException__ctor_System_String_System_String_

drwill-ms, line 144: For this one (because empty check) prefer the 2 param overload: https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=net-5.0#System_ArgumentNullException__ctor_System_String_System_String_

iothub/service/src/RegistryManager.cs
drwill-ms: Rather than use ctor overloads, should we consider relying on `readonly` property initialization?

iothub/service/src/DigitalTwin/DigitalTwinClient.cs
vinagesh: Talked offline. Will resolve as discussed.

drwill-ms (approved):

iothub/service/src/RegistryManager.cs
vinagesh: The current design does what you are suggesting I suppose. Will talk offline.

Issue #12, opened 2016-12-21
No UploadToBlobAsync Operations Monitoring correlation


martinabbott:
Hi,

When using the UploadToBlobAsync operation the SDK performs three actions that are logged in operational monitoring, e.g.,

{"time":"2016-12-21T07:31:24.1350268Z","operationName":"sasUriRequest","category":"FileUploadOperations","level":"Information","deviceId":"device1"}
{"fileUploadStatusCode":0,"time":"2016-12-21T07:31:24.8878250Z","operationName":"fileUploadToStorage","category":"FileUploadOperations","level":"Information","deviceId":"device1"}
{"time":"2016-12-21T07:31:24.8878250Z","operationName":"completionRequest","category":"FileUploadOperations","level":"Information","deviceId":"device1"}

The problem with this, is that if there are a lot of devices all uploading files, it can become nearly impossible to correlate events that relate to a single invocation of the method. I want to do this to build a dashboard so I can see when things are failing, but would currently have to group by deviceId and some vague bracket of time to determine the number of events I have, or have something like a count of each event type against a deviceId over time.

What would be good would be either the filename to be included in the monitoring, or a correlationId, or even better both.

Thanks,

Martin

nberdy:
Thanks for the feedback! I've added a backlog item to the service to provide these fields.

Best,
Nicole

tameraw:
@martinabbott - Closing this issue as it pertains to the service and not with the SDK. Thanks.

PR #1968, opened 2021-05-18
[PnP] Add common resources for convention-based operations

abhipsaMisra:
Moving the code from the dev branch to preview branch in small chunks, so that it can be easily reviewed.

This PR contains the common resources that will be used for convention-based operations:
- PayloadCollection (this is the base type that telemetry and properties inherit from)
- Convention related abstract classes:
    - PayloadSerializer
    - PayloadEncoder
    - PayloadConvention
    - IWritablePropertyResponse (relevant for properties (twin) operations, but added in this PR since it is a part of PayloadSerializer signature)
- Default implementations provided by the sdk (enabled by default):
    - NewtonsoftJsonPayloadSerializer
    - Utf8PayloadEncoder
    - DefaultPayloadConvention
    - NewtonsoftJsonWritablePropertyResponse
- Sample implementations provided by the sdk:
    - SystemTextJsonPayloadSerializer (not supported on net451)
    - SystemTextJsonWritablePropertyResponse (not supported on net451)
    
The subsequent PRs will deal with telemetry, command and properties APIs specifically.

### DIFF
--- a/iothub/device/devdoc/Convention-based operations.md
+++ b/iothub/device/devdoc/Convention-based operations.md
@@ -161,7 +161,6 @@ public class ClientPropertiesUpdateResponse {
 /// </remarks>
 /// <param name="telemetryMessage">The telemetry message.</param>
 /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
-/// <returns></returns>
 public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationToken cancellationToken = default);
 ```
 #### All related types
@@ -190,7 +189,6 @@ public class TelemetryMessage : Message {
 /// </summary>
 /// <param name="callback">A method implementation that will handle the incoming command.</param>
 /// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
-/// <param name="payloadConvention">A convention handler that defines the content encoding and serializer to use for commands.</param>
 /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
 public Task SubscribeToCommandsAsync(Func<CommandRequest, object, Task<CommandResponse>> callback, object userContext, CancellationToken cancellationToken = default);
 ```
--- a/iothub/device/devdoc/Convention-based operations.md
+++ b/iothub/device/devdoc/Convention-based operations.md
@@ -1,0 +1,213 @@
+## Plug and Play convention compatible APIs
+
+#### Common
+
+```csharp
+
+public abstract class PayloadConvention {
+    protected PayloadConvention();
+    public abstract PayloadEncoder PayloadEncoder { get; }
+    public abstract PayloadSerializer PayloadSerializer { get; }
+    public virtual byte[] GetObjectBytes(object objectToSendWithConvention);
+}
+
+public abstract class PayloadEncoder {
+    protected PayloadEncoder();
+    public abstract Encoding ContentEncoding { get; }
+    public abstract byte[] EncodeStringToByteArray(string contentPayload);
+}
+
+public abstract class PayloadSerializer {
+    protected PayloadSerializer();
+    public abstract string ContentType { get; }
+    public abstract T ConvertFromObject<T>(object objectToConvert);
+    public abstract IWritablePropertyResponse CreateWritablePropertyResponse(object value, int statusCode, long version, string description = null);
+    public abstract T DeserializeToType<T>(string stringToDeserialize);
+    public abstract string SerializeToString(object objectToSerialize);
+    public abstract bool TryGetNestedObjectValue<T>(object objectToConvert, string propertyName, out T outValue);
+}
+
+public sealed class DefaultPayloadConvention : PayloadConvention {
+    public static readonly DefaultPayloadConvention Instance;
+    public DefaultPayloadConvention();
+    public override PayloadEncoder PayloadEncoder { get; }
+    public override PayloadSerializer PayloadSerializer { get; }
+}
+
+public class Utf8PayloadEncoder : PayloadEncoder {
+    public static readonly Utf8PayloadEncoder Instance;
+    public Utf8PayloadEncoder();
+    public override Encoding ContentEncoding { get; }
+    public override byte[] EncodeStringToByteArray(string contentPayload);
+}
+
+public class NewtonsoftJsonPayloadSerializer : PayloadSerializer {
+    public static readonly NewtonsoftJsonPayloadSerializer Instance;
+    public NewtonsoftJsonPayloadSerializer();
+    public override string ContentType { get; }
+    public override T ConvertFromObject<T>(object objectToConvert);
+    public override IWritablePropertyResponse CreateWritablePropertyResponse(object value, int statusCode, long version, string description = null);
+    public override T DeserializeToType<T>(string stringToDeserialize);
+    public override string SerializeToString(object objectToSerialize);
+    public override bool TryGetNestedObjectValue<T>(object objectToConvert, string propertyName, out T outValue);
+}
+
+public abstract class PayloadCollection : IEnumerable, IEnumerable<object> {
+    protected PayloadCollection();
+    public IDictionary<string, object> Collection { get; private set; }
+    public PayloadConvention Convention { get; internal set; }
+    public virtual object this[string key] { get; set; }
+    public virtual void Add(string key, object value);
+    public virtual void AddOrUpdate(string key, object value);
+    public bool Contains(string key);
+    public IEnumerator<object> GetEnumerator();
+    public virtual byte[] GetPayloadObjectBytes();
+    public virtual string GetSerializedString();
+    protected void SetCollection(PayloadCollection payloadCollection);
+    IEnumerator System.Collections.IEnumerable.GetEnumerator();
+    public bool TryGetValue<T>(string key, out T value);
+}
+
+public static class ConventionBasedConstants {
+    public const string AckCodePropertyName = "ac";
+    public const string AckDescriptionPropertyName = "ad";
+    public const string AckVersionPropertyName = "av";
+    public const string ComponentIdentifierKey = "__t";
+    public const string ComponentIdentifierValue = "c";
+    public const string ValuePropertyName = "value";
+}
+```
+
+### Properties
+
+```csharp
+/// <summary>
+/// Retrieve the device properties.
+/// </summary>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+/// <returns>The device properties.</returns>
+public Task<ClientProperties> GetClientPropertiesAsync(CancellationToken cancellationToken = default);
+
+/// <summary>
+/// Update properties.
+/// </summary>
+/// <param name="propertyCollection">Reported properties to push.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+/// <returns>The response containing the operation request Id and updated version no.</returns>
+public Task<ClientPropertiesUpdateResponse> UpdateClientPropertiesAsync(ClientPropertyCollection propertyCollection, CancellationToken cancellationToken = default);
+
+/// <summary>
+/// Sets the global listener for Writable properties
+/// </summary>
+/// <param name="callback">The global call back to handle all writable property updates.</param>
+/// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SubscribeToWritablePropertiesEventAsync(Func<ClientPropertyCollection, object, Task> callback, object userContext, CancellationToken cancellationToken = default);
+```
+
+#### All related types
+
+```csharp
+public class ClientProperties : ClientPropertyCollection {
+    public ClientPropertyCollection Writable { get; private set; }
+}
+
+public class ClientPropertyCollection : PayloadCollection {
+    public ClientPropertyCollection();
+    public long Version { get; protected set; }
+    public void Add(IDictionary<string, object> properties, string componentName = null);
+    public override void Add(string propertyName, object propertyValue);
+    public void Add(string propertyName, object propertyValue, int statusCode, long version, string description = null, string componentName = null);
+    public void Add(string propertyName, object propertyValue, string componentName);
+    public void AddOrUpdate(IDictionary<string, object> properties, string componentNam = null);
+    public override void AddOrUpdate(string propertyName, object propertyValue);
+    public void AddOrUpdate(string propertyName, object propertyValue, int statusCode, long version, string description = null, string componentName = null);
+    public void AddOrUpdate(string propertyName, object propertyValue, string componentName);
+    public bool Contains(string componentName, string propertyName);
+    public virtual bool TryGetValue<T>(string componentName, string propertyName, out T propertyValue);
+}
+
+public interface IWritablePropertyResponse {
+    int AckCode { get; set; }
+    string AckDescription { get; set; }
+    long AckVersion { get; set; }
+    object Value { get; set; }
+}
+
+public sealed class NewtonsoftJsonWritablePropertyResponse : IWritablePropertyResponse {
+    public NewtonsoftJsonWritablePropertyResponse(object propertyValue, int ackCode, long ackVersion, string ackDescription = null);
+    public int AckCode { get; set; }
+    public string AckDescription { get; set; }
+    public long AckVersion { get; set; }
+    public object Value { get; set; }
+}
+
+public class ClientPropertiesUpdateResponse {
+    public ClientPropertiesUpdateResponse();
+    public string RequestId { get; internal set; }
+    public long Version { get; internal set; }
+}
+```
+
+### Telemetry
+
+```csharp
+/// <summary>
+/// Send telemetry using the specified message.
+/// </summary>
+/// <remarks>
+/// Use the <see cref="TelemetryMessage(string, TelemetryCollection)"/> constructor to pass in the optional
+/// <see cref="TelemetryCollection"/> that specifies your payload and serialization and encoding rules.
+/// </remarks>
+/// <param name="telemetryMessage">The telemetry message.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+/// <returns></returns>
+public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationToken cancellationToken = default);
+```
+#### All related types
+
+```csharp
+public class TelemetryCollection : PayloadCollection {
+    public TelemetryCollection();
+    public override void Add(string telemetryName, object telemetryValue);
+    public override void AddOrUpdate(string telemetryName, object telemetryValue);
+}
+
+public class TelemetryMessage : Message {
+    public TelemetryMessage(string componentName = null);
+    public new string ContentEncoding { get; internal set; }
+    public new string ContentType { get; internal set; }
+    public TelemetryCollection Telemetry { get; set; }
+    public override Stream GetBodyStream();
+}
+```
+
+### Commands
+
+```csharp
+/// <summary>
+/// Set the global command callback handler.
+/// </summary>
+/// <param name="callback">A method implementation that will handle the incoming command.</param>
+/// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
+/// <param name="payloadConvention">A convention handler that defines the content encoding and serializer to use for commands.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SubscribeToCommandsAsync(Func<CommandRequest, object, Task<CommandResponse>> callback, object userContext, CancellationToken cancellationToken = default);
+```
+#### All related types
+
+```csharp
+public sealed class CommandRequest {
+    public string CommandName { get; private set; }
+    public string ComponentName { get; private set; }
+    public string DataAsJson { get; }
+    public T GetData<T>();
+}
+
+public sealed class CommandResponse {
+    public CommandResponse(int status);
+    public CommandResponse(object result, int status);
+    public string ResultAsJson { get; }
+    public int Status { get; private set; }
+}
+```
--- a/shared/src/IWritablePropertyResponse.cs
+++ b/shared/src/IWritablePropertyResponse.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// The interface that defines the structure of a writable property response.
+    /// </summary>
+    /// <remarks>
+    /// This interface is used to allow extension to use a different set of attributes for serialization.
+    /// For example our default implementation found in <see cref="NewtonsoftJsonWritablePropertyResponse"/> is based on <see cref="Newtonsoft.Json"/> seriailizer attributes.
+    /// </remarks>
+    public interface IWritablePropertyResponse
+    {
+        /// <summary>
+        /// The unserialized property value.
+        /// </summary>
+        public object Value { get; set; }
+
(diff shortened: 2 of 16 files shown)


shared/src/IWritablePropertyResponse.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// The interface that defines the structure of a writable property response.
```
abhipsaMisra, line 7: this is relevant for scenario where a device responds to a property update request sent by service:
-> service -> update twin patch -> device subscribed to twin property updates -> device receives property update -> device responds to the property update request

In the "device responds to the property update request", for devices following pnp convention, the patch should follow the following format:
https://docs.microsoft.com/en-us/azure/iot-pnp/concepts-convention#writable-properties

shared/src/NewtonsoftJsonPayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            if (objectToConvert == null)
+            {
+                return default;
+            }
+            return ((JObject)objectToConvert).ToObject<T>();
+        }
+
+        /// <inheritdoc/>
+        public override bool TryGetNestedObjectValue<T>(object objectToConvert, string propertyName, out T outValue)
```
abhipsaMisra, line 50: this method is to help with inspection of the json node for property operations. This node traversal will depend on the serializer defined, so it will need to be implemented by the serializer (eg: the node could be JsonObject (newtonsoft) or JsonElement for system.text).
For eg: when a device wants to fetch the value of a property under a component, 
```json
"thermostat1": {
	"__t": "c",
	"targetTemperature": 25
},
```

the sdk only knows that the value under "thermostat1" is an object. The type of the object depends on the serializer used to deserialize the response from the service (the twin received).

shared/src/PayloadCollection.cs
```diff
@@ -0,0 +1,154 @@
...
+    {
+        /// <summary>
+        /// The underlying collection for the payload.
+        /// </summary>
+        public IDictionary<string, object> Collection { get; private set; } = new Dictionary<string, object>();
+
+        /// <summary>
+        /// The convention to use with this payload.
+        /// </summary>
+        public PayloadConvention Convention { get; internal set; }
```
abhipsaMisra, line 23: one drawback of placing these common resources in the shared assembly is that either all APIs need to be public (even though they shouldn't be, design-wise), OR we need to add the client library as a "friend" to the shared library (InternalsVisibleTo).

The other, not-so-elegant approach is to add these common resources to the "common" folder and add a compile-include condition of the client csproj files.

What do you think?
- make all methods public
- add the device client assembly as friend to shared assembly - allow InternalsVisibleTo
- add the common files as compile-include to the project

drwill-ms: Why does it need to be in shared?

abhipsaMisra: The idea is that since these are IoT hub defined conventions, they could be relevant for both device and service side applications, hence, could be relevant for both device clients and service clients.
While we do have very pnp specific APIs on the service client, the "regular" service client and registry manager also can also be used to perform operations on pnp-enabled devices. If we are to ever make those APIs follow these conventions, that would be the motivation for placing these in shared.

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -0,0 +1,213 @@
...
+    public override IWritablePropertyResponse CreateWritablePropertyResponse(object value, int statusCode, long version, string description = null);
+    public override T DeserializeToType<T>(string stringToDeserialize);
+    public override string SerializeToString(object objectToSerialize);
+    public override bool TryGetNestedObjectValue<T>(object objectToConvert, string propertyName, out T outValue);
+}
+
+public abstract class PayloadCollection : IEnumerable, IEnumerable<object> {
+    protected PayloadCollection();
+    public IDictionary<string, object> Collection { get; private set; }
+    public PayloadConvention Convention { get; internal set; }
```
vinagesh, line 58: Is this going to stay here or move to ClientOptions?

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -0,0 +1,213 @@
...
+```csharp
+/// <summary>
+/// Retrieve the device properties.
+/// </summary>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+/// <returns>The device properties.</returns>
+public Task<ClientProperties> GetClientPropertiesAsync(CancellationToken cancellationToken = default);
+
+/// <summary>
+/// Update properties.
```
vinagesh, line 92: Update device properties?

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -0,0 +1,213 @@
...
+/// <returns>The response containing the operation request Id and updated version no.</returns>
+public Task<ClientPropertiesUpdateResponse> UpdateClientPropertiesAsync(ClientPropertyCollection propertyCollection, CancellationToken cancellationToken = default);
+
+/// <summary>
+/// Sets the global listener for Writable properties
+/// </summary>
+/// <param name="callback">The global call back to handle all writable property updates.</param>
+/// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+public Task SubscribeToWritablePropertiesEventAsync(Func<ClientPropertyCollection, object, Task> callback, object userContext, CancellationToken cancellationToken = default);
```
vinagesh, line 105: Are there no sync versions of these?

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -0,0 +1,213 @@
...
+/// <summary>
+/// Send telemetry using the specified message.
+/// </summary>
+/// <remarks>
+/// Use the <see cref="TelemetryMessage(string, TelemetryCollection)"/> constructor to pass in the optional
+/// <see cref="TelemetryCollection"/> that specifies your payload and serialization and encoding rules.
+/// </remarks>
+/// <param name="telemetryMessage">The telemetry message.</param>
+/// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+/// <returns></returns>
```
vinagesh, line 164: No need of returns here

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Commands
+
+```csharp
+/// <summary>
+/// Set the global command callback handler.
+/// </summary>
+/// <param name="callback">A method implementation that will handle the incoming command.</param>
+/// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
+/// <param name="payloadConvention">A convention handler that defines the content encoding and serializer to use for commands.</param>
```
vinagesh, line 193: Is this param not part of the API? - payloadConvention

shared/src/PayloadCollection.cs
drwill-ms: So we're creating a challenge for ourselves by trying to future proof for a problem we don't know we'll have?

shared/src/NewtonsoftJsonPayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// A <see cref="Newtonsoft.Json.JsonConvert"/> <see cref="PayloadSerializer"/> implementation.
+    /// </summary>
+    public class NewtonsoftJsonPayloadSerializer : PayloadSerializer
+    {
+        /// <summary>
+        /// The Content Type string.
```
drwill-ms, line 15: In this case, there is no reason to capitalize "content type".

shared/src/PayloadCollection.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        /// <summary>
+        /// Get the value at the specified key
+        /// </summary>
+        /// <param name="key">Key of value</param>
+        /// <remarks>
```
drwill-ms, line 29: Remarks above params, I believe

shared/src/PayloadCollection.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// This accessor is best used to access simple types. It is recommended to use <see cref="TryGetValue"/> to cast a complex type.
+        /// </remarks>
+        /// <returns>The specified property.</returns>
```
drwill-ms, line 32: ...as an object.

What does a user do with the object after that? Will it cast? Do they need to deserialize? Only for complex types?

shared/src/PayloadCollection.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// <summary>
+        /// Gets the value of the object from the collection.
+        /// </summary>
+        /// <remarks>
+        /// This class is used for both sending and receiving properties for the device.
+        /// </remarks>
+        /// <typeparam name="T">The type to cast the object to.</typeparam>
+        /// <param name="key">The key of the property to get.</param>
+        /// <param name="value">The value of the object from the collection.</param>
+        /// <returns>true if the collection contains an element with the specified key; otherwise, false.</returns>
```
drwill-ms, line 94: True (capitalize first word of sentence)

drwill-ms, line 94: "otherwise, false" does not stand on its own as a sentence. So either make it a sentence (subject, verb) or replace the semicolon with a comma.

shared/src/PayloadCollection.cs
```diff
@@ -0,0 +1,154 @@
...
+            }
+
+            value = default;
+            return false;
+        }
+
+        /// <summary>
+        /// Returns a serailized string of this collection from the <see cref="PayloadSerializer.SerializeToString(object)"/> method.
+        /// </summary>
+        /// <returns></returns>
```
drwill-ms, line 118: Fill in comments

shared/src/PayloadCollection.cs
abhipsaMisra: That is absolutely correct, we are future-proofing ourselves for an issue that we don't know for sure we'll have, but since these operations have both device and service side components, it does seem reasonable that we might want to extend it at some point.

shared/src/PayloadConvention.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// The payload convention class.
+    /// </summary>
+    /// <remarks>The payload convention is used to define a specific serializer as well as a specific content encoding. For example, IoT has a <see href="https://docs.microsoft.com/en-us/azure/iot-pnp/concepts-convention">convention</see> that is designed to make it easier to get started with products that use specific conventions by default.</remarks>
```
drwill-ms, line 9: can we break this line up. it is a bit long

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: These payload collections (TelemetryCollection, PropertiesCollection) will have a component associated with them, so that the client knows what is the type of data these collections hold. This convention within the collection is not settable by the user, but rather is information that the sdk sets and holds internally.
The user will set the Convention via the ClientOptions.

shared/src/PayloadConvention.cs
```diff
@@ -0,0 +1,36 @@
...
+        public abstract PayloadSerializer PayloadSerializer { get; }
+
+        /// <summary>
+        /// Gets the encoder used for the payload to be serialized.
+        /// </summary>
+        /// <value>An encoder that will be used to convert the serialized string to a byte array.</value>
+        public abstract PayloadEncoder PayloadEncoder { get; }
+
+        /// <summary>
+        /// Returns the byte array for the convention based message.
```
drwill-ms, line 25: convention-based

shared/src/PayloadConvention.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// </summary>
+        /// <remarks>This base class will use the <see cref="PayloadSerializer"/> and <see cref="PayloadEncoder"/> to create this byte array.</remarks>
+        /// <param name="objectToSendWithConvention"></param>
```
drwill-ms, line 28: Fill in doc comment

shared/src/PayloadEncoder.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// This class specifies the byte encoding for the the payload.
+    /// </summary>
+    /// <remarks>
+    /// The encoder is responsible for encoding all of your objects into the corrent bytes for the <see cref="PayloadConvention"/> that uses it.
```
drwill-ms, line 12: corrent = current?

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: this could be for either device or module clients though, so we're not specifying anything here

shared/src/PayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// Provides the serialzation for a specified convention
```
drwill-ms, line 7: punctuation

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: true, will remove

shared/src/PayloadSerializer.cs
drwill-ms: There are many other cases too. Remember, this will show up in web documentation, so each sentence should end with punctuation.

shared/src/PayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// By default we have implemented the <see cref="NewtonsoftJsonPayloadSerializer"/> class that uses <see cref="Newtonsoft.Json.JsonConvert"/> to handle the serialization for the <see cref="DefaultPayloadConvention"/> class.
+    /// </para>
+    /// </remarks>
+    public abstract class PayloadSerializer
+    {
+        /// <summary>
+        /// Used to specify what type of content to expect
+        /// </summary>
+        /// <value>A string representing the content type to use when sending a payload</value>
+        /// <remarks>This can be freeform but should adhere to standard MIME types. For example, "application/json" is what we implement by default.</remarks>
```
drwill-ms, line 21: free-form

shared/src/PayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+        public abstract string ContentType { get; }
+
+        /// <summary>
+        /// Serialize the specified object to a string
+        /// </summary>
+        /// <param name="objectToSerialize">Object to serialize.</param>
+        /// <returns>A serilaized string of the object.</returns>
```
drwill-ms, line 28: Misspelled serialized

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: it is not, I'll update this

shared/src/PayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <returns>A fully deserialized type.</returns>
+        public abstract T DeserializeToType<T>(string stringToDeserialize);
+
+        /// <summary>
+        /// Converts the object using the serializer.
+        /// </summary>
+        /// <typeparam name="T">The type to convert to.</typeparam>
+        /// <param name="objectToConvert">The object to convert.</param>
+        /// <returns>A converted object</returns>
+        /// <remarks>This class is used by the <see cref="PayloadCollection"/> class to attempt to convert from the native serailizer type (for example, JObject or JsonElement) to the desired type. When you implement this you need to be aware of what type your serializer will use for anonymous types.</remarks>
```
drwill-ms, line 45: Another case of remarks appearing out of our general order:

1. Summary
2. Remarks
3. Params
4. Returns
5. Exception
6. Example

shared/src/PayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <summary>
+        /// Gets a nested property from the serialized data.
+        /// </summary>
+        /// <remarks>
+        /// This is used internally by our <see cref="PayloadCollection"/> to attempt to get a proprty of the underlying object. An example of this would be a property under the component. 
+        /// </remarks>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="objectToConvert"></param>
+        /// <param name="propertyName"></param>
+        /// <returns></returns>
```
drwill-ms, line 57: Fill in doc comment

shared/src/PayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+        public abstract bool TryGetNestedObjectValue<T>(object objectToConvert, string propertyName, out T outValue);
+
+        /// <summary>
+        /// Creates the correct <see cref="IWritablePropertyResponse"/> to be used with this serializer
+        /// </summary>
+        /// <param name="value">The value of the property.</param>
+        /// <param name="statusCode">The status code of the write operation.</param>
+        /// <param name="version">The version the property is responding to.</param>
+        /// <param name="description">An optional description of the writable property response.</param>
+        /// <returns></returns>
```
drwill-ms, line 68: Fill in doc comment.

shared/src/PayloadCollection.cs
abhipsaMisra: we have this comment above:
"This accessor is best used to access simple types. It is recommended to use <see cref="TryGetValue"/> to cast a complex type."; would it help if we call out explicitly that they should cast simple types?

shared/src/PayloadCollection.cs
drwill-ms: Until we know these specific types, operating the exact same way as they do on the device, are useful to the service, I'd suggest we do not put them in the shared library. Just put them in device. Before we ship GA, if you find a compelling reason to share them, we can address it then.

YAGNI

shared/src/PayloadEncoder.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <para>
```
drwill-ms, line 13: I've not found the `<para>` tag to actually be necessary. Intellisense and web docs both seem to treat whitespace as you'd expect. Have you found otherwise?

drwill-ms (approved):
While I would like the feedback considered/addressed, there's very little here that I'd prioritize before shipping a preview, except maybe the shared code issue.

shared/src/PayloadConvention.cs
abhipsaMisra: would this be better?
For example, IoT has a <see href="https://docs.microsoft.com/en-us/azure/iot-pnp/concepts-convention">convention</see> that is designed
to make it easier to get started with convention-based products.

shared/src/NewtonsoftJsonPayloadSerializer.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// </summary>
+        internal const string ApplicationJson = "application/json";
```
bikamani, line 17: _camelCase

drwill-ms: Negative. Constants take TitleCasing.

See <https://github.com/Azure/azure-iot-sdk-csharp/blob/master/doc/coding-style.md> #12.

shared/src/PayloadEncoder.cs
abhipsaMisra: I believe it was meant to be "correct"

shared/src/PayloadConvention.cs
drwill-ms: I don't mean to reword it, just to line break at a reasonable length

shared/src/PayloadCollection.cs
drwill-ms: Neither comment addresses simple types. I think it is worth mentioning.

Also, TryGetValue doesn't strictly cast, right? It will either cast or deserialize. Right?

shared/src/PayloadSerializer.cs
abhipsaMisra: I wish there was a plugin that we could use for this

drwill-ms: One of the many spell checkers may do it.

shared/src/NewtonsoftJsonPayloadSerializer.cs
bikamani: Where does it take the component information? Or perhaps a sample would help understand how would this method be used.

We can sync on this offline if you think it's easier to explain.

shared/src/IWritablePropertyResponse.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <remarks>
+    /// This interface is used to allow extension to use a different set of attributes for serialization.
+    /// For example our default implementation found in <see cref="NewtonsoftJsonWritablePropertyResponse"/> is based on <see cref="Newtonsoft.Json"/> seriailizer attributes.
+    /// </remarks>
+    public interface IWritablePropertyResponse
+    {
+        /// <summary>
+        /// The unserialized property value.
+        /// </summary>
+        public object Value { get; set; }
```
bikamani, line 18: I see value in calling this propertyValue. Thoughts?

shared/src/PayloadConvention.cs
abhipsaMisra: ah, got it!

shared/src/NewtonsoftJsonPayloadSerializer.cs
abhipsaMisra: This information is something that the sdk would leverage internally when looking for a property nested under a component:
eg: https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/pnpImplDev/iothub/device/src/ClientPropertyCollection.cs#L235
eg: https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/pnpImplDev/iothub/device/src/ClientPropertyCollection.cs#L262

Ping me if this is still unclear and we can discuss it offline.

shared/src/PayloadEncoder.cs
abhipsaMisra: I'll let @jamdavi comment on this, since you've worked on most (if not all :)) of the doc comments.

shared/src/IWritablePropertyResponse.cs
abhipsaMisra: Since the name calls out that this is a "writable property" response, we didn't call it out again in the property (.NET property) names.
Also, the .NET property names are aligned to be as close as possible (while still understandable) to the pnp-json-property names (value, ac, av,ad)

shared/src/PayloadEncoder.cs
jamdavi: Let me test this assertion. IIRC it might treat whitespace as new lines, but it might not create a new <p> tag in the docs.

vinagesh (approved):

shared/src/PayloadEncoder.cs
drwill-ms: You can test it with Intellisense immediately. For web docs, all we have to do is find an existing case where we have newlines and see what the docs look like. I'm pretty sure I did that recently and saw it worked fine.

Check out `Device.ParentScopes` as an example: <https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.device.parentscopes?view=azure-dotnet#Microsoft_Azure_Devices_Device_ParentScopes>

PR #1969, opened 2021-05-19
[PnP] Add support for convention-based telemetry operation

abhipsaMisra:
This PR addresses support for the telemetry APIs.

### DIFF
--- a/iothub/device/src/DeviceClient.ConventionBasedOperations.cs
+++ b/iothub/device/src/DeviceClient.ConventionBasedOperations.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Contains methods that a convention based device can use to send telemetry to the service,
+    /// respond to commands and perform operations on its properties.
+    /// </summary>
+    /// <threadsafety static="true" instance="true" />
+    public partial class DeviceClient : IDisposable
+    {
+        /// <summary>
+        /// Send telemetry using the specified message.
+        /// </summary>
+        /// <remarks>
+        /// Use the <see cref="TelemetryMessage(string)"/> constructor to pass in the optional component name
+        /// that the telemetry message is from.
+        /// </remarks>
+        /// <param name="telemetryMessage">The telemetry message.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <returns></returns>
+        public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationToken cancellationToken = default)
+            => InternalClient.SendTelemetryAsync(telemetryMessage, cancellationToken);
+    }
+}
--- a/iothub/device/src/InternalClient.ConventionBasedOperations.cs
+++ b/iothub/device/src/InternalClient.ConventionBasedOperations.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Shared;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Contains methods that a convention based device/ module can use to send telemetry to the service,
+    /// respond to commands and perform operations on its properties.
+    /// </summary>
+    internal partial class InternalClient
+    {
+        internal PayloadConvention PayloadConvention => _clientOptions.PayloadConvention ?? DefaultPayloadConvention.Instance;
+
+        internal Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationToken cancellationToken)
+        {
+            if (telemetryMessage == null)
+            {
+                throw new ArgumentNullException(nameof(telemetryMessage));
+            }
+
+            if (telemetryMessage.Telemetry != null)
+            {
+                telemetryMessage.Telemetry.Convention = PayloadConvention;
+                telemetryMessage.PayloadContentEncoding = PayloadConvention.PayloadEncoder.ContentEncoding.WebName;
+                telemetryMessage.PayloadContentType = PayloadConvention.PayloadSerializer.ContentType;
+            }
+
+            return SendEventAsync(telemetryMessage, cancellationToken);
+        }
+    }
+}
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Azure.Devices.Client
     /// Contains methods that a device can use to send messages to and receive messages from the service,
     /// respond to direct method invocations from the service, and send and receive twin property updates.
     /// </summary>
-    internal class InternalClient : IDisposable
+    internal partial class InternalClient : IDisposable
     {
         private readonly SemaphoreSlim _methodsSemaphore = new SemaphoreSlim(1, 1);
         private readonly SemaphoreSlim _deviceReceiveMessageSemaphore = new SemaphoreSlim(1, 1);
@@ -643,7 +643,7 @@ namespace Microsoft.Azure.Devices.Client
         /// Sends an event to device hub
         /// </summary>
         /// <returns>The message containing the event</returns>
-        public async Task SendEventAsync(Message message)
+        public async Task SendEventAsync(MessageBase message)
         {
             try
             {
@@ -662,7 +662,7 @@ namespace Microsoft.Azure.Devices.Client
         /// Sends an event to device hub
         /// </summary>
         /// <returns>The message containing the event</returns>
-        public Task SendEventAsync(Message message, CancellationToken cancellationToken)
+        public Task SendEventAsync(MessageBase message, CancellationToken cancellationToken)
         {
             if (message == null)
             {
--- a/iothub/device/src/Message.cs
+++ b/iothub/device/src/Message.cs
@@ -1,12 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.IO;
-using System.Threading;
-using Microsoft.Azure.Devices.Client.Extensions;
-using Microsoft.Azure.Devices.Client.Common.Api;
-using System.Collections.Generic;
 using Microsoft.Azure.Devices.Shared;
 
 namespace Microsoft.Azure.Devices.Client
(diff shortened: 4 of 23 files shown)


iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Contains methods that a convention based device can use to send telemetry to the service,
+    /// respond to commands and perform operations on its properties.
+    /// </summary>
+    /// <threadsafety static="true" instance="true" />
+    public partial class DeviceClient : IDisposable
```
abhipsaMisra, line 15: I updated `DeviceClient` to be a partial class, so that we could separate out the convention-based operations from the non-convention based ones. I preferred to use this over using regions on the existing DeviceClient class.
Let me know if there are any concerns.

iothub/device/src/Message.cs
```diff
@@ -2,19 +2,18 @@
...
-using System.Collections.Generic;
 using Microsoft.Azure.Devices.Shared;
 
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// The data structure represent the message that is used for interacting with IotHub.
     /// </summary>
-    public sealed class Message : IReadOnlyIndicator, IDisposable
+    public class Message : IReadOnlyIndicator, IDisposable
```
abhipsaMisra, line 16: Since TelemetryMessage will inherit from Message, this cannot be sealed anymore.

iothub/device/src/Message.cs
```diff
@@ -156,22 +155,16 @@ public uint DeliveryCount
...
             get => GetSystemProperty<string>(MessageSystemPropertyNames.UserId);
-            set => SystemProperties[MessageSystemPropertyNames.UserId] = value;
+            set => SetSystemProperty(MessageSystemPropertyNames.UserId, value);
         }
 
-        /// <summary>
-        /// For outgoing messages, contains the Mqtt topic that the message is being sent to
-        /// For incoming messages, contains the Mqtt topic that the message arrived on
-        /// </summary>
-        internal string MqttTopicName { get; set; }
```
abhipsaMisra, line 166: moved this to below public methods

iothub/device/src/Message.cs
```diff
@@ -290,7 +283,7 @@ internal bool HasBodyStream()
         /// <exception cref="InvalidOperationException">throws if the method has been called.</exception>
         /// <exception cref="ObjectDisposedException">throws if the event data has already been disposed.</exception>
         /// <remarks>This method can only be called once and afterwards method will throw <see cref="InvalidOperationException"/>.</remarks>
-        public Stream GetBodyStream()
+        public virtual Stream GetBodyStream()
```
abhipsaMisra, line 286: this is virtual so that it can be overridden in TelemetryMessage

iothub/device/src/Message.cs
abhipsaMisra, line 16: One potential issue could be that `SendEventAsync` API accepts `Message, so technically a user could inherit pass in an inherited `Message` instance to that API.
Thoughts?

iothub/device/src/TelemetryMessage.cs
```diff
@@ -0,0 +1,73 @@
...
+        public TelemetryCollection Telemetry { get; set; } = new TelemetryCollection();
+
+        /// <inheritdoc/>
+        /// <remarks>
+        /// This property has been hidden by this class to only allow you to set the ContentType via the <see cref="PayloadSerializer"/> class.
+        /// </remarks>
+        /// <value>The base <see cref="Message"/> content type.</value>
+        [Browsable(false)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public new string ContentType
```
abhipsaMisra, line 33: this is another point worth highlighting:
- since the content type and encoding info is now a part of the client itself (it is set in ClientOptions), these fields have no significance. Since we cannot hide them in the `Message` class directly (programatically), it has been hidden this way.

Thoughts?

abhipsaMisra: " these fields have no significance" => meaning there is no point in setting or getting this information from the `TelemetryMessage` class. It is set properly internally by the sdk though (in the `InternalClient`).

iothub/device/src/Message.cs
jamdavi: In this instance we have a well defined behavior if a customer sends a TelemetryMessage via SendEventAsync. The GetPayloadObjectBytes() method will throw a nullref since the Convention won't be set.

We can either...

1. Make SendEventAsync() TelemetryMessage aware
2. Make GetPayloadObjectBytes() check the convention before serializing.

iothub/device/src/TelemetryMessage.cs
jamdavi: One option would be to strip out everything but the ContentType and ContentEncoding from the Message class and make an abstract MessageBase that has the other properties we want to have. 

So it would be MessageBase -> Message and MessageBase -> TelemetryMessage.

We could then move that to the shared class and start the process of unifying the Message class

iothub/device/src/Message.cs
abhipsaMisra: I think we should add null checks to the convention APIs, to avoid running into unexpected NREs.

I don't think SendEventAsync() needs to be TelemetryMessage aware. My concern is around people inheriting from `Message` and sending their custom implementation to the SendEventAsync() API. I think we can enghance the doc comments to protect against that (add comments to the virtual methods, so that they know what behavior to expect).

iothub/device/src/InternalClient.cs
```diff
@@ -1887,6 +1889,24 @@ internal void ValidateModuleTransportHandler(string apiName)
 
         #endregion Module Specific API
 
+        #region Convention driven operations
```
vinagesh, line 1892: Would a partial class be better here as well?

iothub/device/src/Message.cs
vinagesh: If you think that is a concern, you can also use composition. TelemetryMessage could have a ```Message message``` class property instead of the inheritance.

iothub/device/src/TelemetryMessage.cs
vinagesh: This approach might help keep the Message class sealed as well.

abhipsaMisra: I agree, two birds one stone.

iothub/device/src/InternalClient.cs
abhipsaMisra: yeah, we could do that, sure

iothub/device/src/Message.cs
```diff
@@ -1,39 +1,22 @@
...
-using System.Collections.Generic;
 using Microsoft.Azure.Devices.Shared;
 
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// The data structure represent the message that is used for interacting with IotHub.
     /// </summary>
-    public sealed class Message : IReadOnlyIndicator, IDisposable
+    public sealed class Message : MessageBase
```
abhipsaMisra, line 12: The Message class is split into a `MessageBase`, which has all of the base functionality that is common between both `Message` and `TelemetryMessage`.

iothub/device/src/Message.cs
```diff
@@ -57,374 +36,36 @@ public Message(Stream stream)
...
-        }
-
-        /// <summary>
-        /// Specifies the module Id from which this message was sent, if there is one.
-        /// </summary>
-        public string ConnectionModuleId
-        {
-            get => GetSystemProperty<string>(MessageSystemPropertyNames.ConnectionModuleId);
-            internal set => SystemProperties[MessageSystemPropertyNames.ConnectionModuleId] = value;
+            get => PayloadContentType;
```
abhipsaMisra, line 58: the publicly visible field on `Message` is called `ContentType`, which the internal backing field on `MessageBase` is called `PayloadContentType`.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -152,7 +152,7 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
 
         #region Telemetry
 
-        public override async Task SendEventAsync(Message message, CancellationToken cancellationToken)
+        public override async Task SendEventAsync(MessageBase message, CancellationToken cancellationToken)
```
abhipsaMisra, line 155: The transport layer is updated to operation on `MessageBase`. This could be either `Message` or `TelemetryMessage`, depending on which public API calls it.

iothub/device/src/InternalClient.ConventionBasedOperations.cs
```diff
@@ -0,0 +1,36 @@
...
+        internal PayloadConvention PayloadConvention => _clientOptions.PayloadConvention ?? DefaultPayloadConvention.Instance;
+
+        internal Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationToken cancellationToken)
+        {
+            if (telemetryMessage == null)
+            {
+                throw new ArgumentNullException(nameof(telemetryMessage));
+            }
+
+            if (telemetryMessage.Telemetry != null)
```
abhipsaMisra, line 26: Convention related data is relevant only when `TelemetryMessage` has associate `Telemetry` payload.

jamdavi (approved):
I like the move to message base.

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Send telemetry using the specified message.
+        /// </summary>
+        /// <remarks>
+        /// Use the <see cref="TelemetryMessage(string)"/> constructor to pass in the optional component name
+        /// that the telemetry message is from.
+        /// </remarks>
+        /// <param name="telemetryMessage">The telemetry message.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <returns></returns>
```
abhipsaMisra, line 26: i've removed this tag in a subsequent PR

iothub/device/src/ModuleClient.ConventionBasedOperations.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Send telemetry using the specified message.
+        /// </summary>
+        /// <remarks>
+        /// Use the <see cref="TelemetryMessage(string)"/> constructor to pass in the optional component name
+        /// that the telemetry message is from.
+        /// </remarks>
+        /// <param name="telemetryMessage">The telemetry message.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <returns></returns>
```
abhipsaMisra, line 26: same here, removed in a subsequent PR

vinagesh (approved):

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
bikamani, line 1: wondering reasoning behind the naming DeviceClient.xxx

abhipsaMisra: Since we already have a class called DeviceClient.cs, the other partial class needed to have a different name. The "convention-based-operation" part of the name is to signify that this file holds the APIs that follow a particular convention for their payload.

PR #1970, opened 2021-05-19
[PnP] Add support for convention-based command operations

abhipsaMisra:
This PR addresses support for the command APIs.

### DIFF
--- a/iothub/device/src/CommandRequest.cs
+++ b/iothub/device/src/CommandRequest.cs
@@ -39,7 +39,24 @@ namespace Microsoft.Azure.Devices.Client
         /// <returns>The command request data.</returns>
         public T GetData<T>()
         {
-            return _payloadConvention.PayloadSerializer.DeserializeToType<T>(DataAsJson);
+            string dataAsJson = DataAsJson;
+
+            return dataAsJson == null
+                ? default
+                : _payloadConvention.PayloadSerializer.DeserializeToType<T>(dataAsJson);
+        }
+
+        /// <summary>
+        /// The command request data bytes.
+        /// </summary>
+        /// <returns>
+        /// The command request data bytes.
+        /// </returns>
+        public byte[] GetDataAsBytes()
+        {
+            // Need to return a clone of the array so that consumers
+            // of this library cannot change its contents.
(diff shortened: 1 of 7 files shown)


iothub/device/src/CommandRequest.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// The command name.
+        /// </summary>
+        public string CommandName { get; private set; }
+
+        /// <summary>
+        /// The command request data.
+        /// </summary>
+        /// <typeparam name="T">The type to cast the command request data to.</typeparam>
+        /// <returns>The command request data.</returns>
+        public T GetData<T>()
```
jamdavi, line 40: TryGetData<T> ?

Should we also add TryGetValue<T> to match the payload collection?

iothub/device/src/InternalClient.ConventionBasedOperations.cs
```diff
@@ -32,5 +33,33 @@ internal Task SendTelemetryAsync(TelemetryMessage telemetryMessage, Cancellation
...
             return SendEventAsync(telemetryMessage, cancellationToken);
         }
+
+        internal Task SubscribeToCommandsAsync(Func<CommandRequest, object, Task<CommandResponse>> callback, object userContext, CancellationToken cancellationToken)
+        {
+            // Subscribe to methods default handler internally and use the callback received internally to invoke the user supplied command callback.
+            var methodDefaultCallback = new MethodCallback(async (methodRequest, userContext) =>
+            {
+                CommandRequest commandRequest;
+                if (methodRequest.Name.Contains(ConventionBasedConstants.ComponentLevelCommandSeparator))
```
vinagesh, line 43: any chance methodRequest or methodRequest.Name can be null?

abhipsaMisra: there won't be, a method request will always be accompanied by method name; but no harm in adding a null check here

iothub/device/src/CommandRequest.cs
abhipsaMisra: done

vinagesh (approved):

PR #1974, opened 2021-05-20
[PnP] Add support for convention-based properties operations

abhipsaMisra:
This PR addresses support for the property APIs.

### DIFF
--- a/iothub/device/src/ClientProperties.cs
+++ b/iothub/device/src/ClientProperties.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// A container for properties retrieved from the service.
+    /// <remarks>
+    /// The Properties class is not meant to be constructed by customer code.
+    /// It is intended to be returned fully populated from the client method.
+    /// </remarks>
+    /// </summary>
+    public class ClientProperties : ClientPropertyCollection
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="ClientProperties"/>.
+        /// </summary>
+        internal ClientProperties()
+        {
+            Writable = new ClientPropertyCollection();
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="ClientProperties"/> with the specified collections.
+        /// </summary>
+        /// <param name="requestedPropertyCollection">A collection of writable properties returned from IoT Hub.</param>
+        /// <param name="readOnlyPropertyCollection">A collection of read-only properties returned from IoT Hub.</param>
+        internal ClientProperties(ClientPropertyCollection requestedPropertyCollection, ClientPropertyCollection readOnlyPropertyCollection)
+        {
+            SetCollection(readOnlyPropertyCollection);
+            Version = readOnlyPropertyCollection.Version;
+            Writable = requestedPropertyCollection;
+        }
+
+        /// <summary>
+        /// The collection of writable properties.
+        /// </summary>
+        /// <remarks>
+        /// See the <see href="https://docs.microsoft.com/en-us/azure/iot-pnp/concepts-convention#writable-properties">Writable properties</see> documentation for more information.
+        /// </remarks>
+        public ClientPropertyCollection Writable { get; private set; }
+    }
+}
--- a/iothub/device/src/ClientProperties.cs
+++ b/iothub/device/src/ClientProperties.cs
@@ -5,17 +5,19 @@ namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// A container for properties retrieved from the service.
-    /// <remarks>
-    /// The Properties class is not meant to be constructed by customer code.
-    /// It is intended to be returned fully populated from the client method.
-    /// </remarks>
     /// </summary>
+    /// <remarks>
+    /// The <see cref="ClientProperties"/> class is not meant to be constructed by customer code.
+    /// It is intended to be returned fully populated from the client method <see cref="InternalClient.GetClientPropertiesAsync(System.Threading.CancellationToken)"/>.
+    /// </remarks>
     public class ClientProperties : ClientPropertyCollection
     {
         /// <summary>
         /// Initializes a new instance of <see cref="ClientProperties"/>.
+        /// This is provided for unit testing purposes only.
         /// </summary>
-        internal ClientProperties()
+        /// <inheritdoc path="/remarks" cref="ClientProperties" />
+        public ClientProperties()
         {
             Writable = new ClientPropertyCollection();
         }
--- a/iothub/device/src/ClientPropertiesUpdateResponse.cs
+++ b/iothub/device/src/ClientPropertiesUpdateResponse.cs
@@ -9,10 +9,11 @@ namespace Microsoft.Azure.Devices.Client
     public class ClientPropertiesUpdateResponse
     {
         /// <summary>
-        /// The request Id that is appended to the <see cref="InternalClient.UpdateClientPropertiesAsync(ClientPropertyCollection, System.Threading.CancellationToken)"/> operation.
+        /// The request Id that is associated with the <see cref="InternalClient.UpdateClientPropertiesAsync(ClientPropertyCollection, System.Threading.CancellationToken)"/> operation.
         /// </summary>
         /// <remarks>
-        /// This request Id is relevant only for operations over MQTT, and can be used for debugging the operation from the service side.
+        /// This request Id is relevant only for operations over MQTT, and can be used for tracking the operation on service side logs.
+        /// Note that you would need to contact the support team to track operations on the service side.
         /// </remarks>
         public string RequestId { get; internal set; }
 
--- a/iothub/device/src/ClientPropertiesUpdateResponse.cs
+++ b/iothub/device/src/ClientPropertiesUpdateResponse.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// The response of an <see cref="InternalClient.UpdateClientPropertiesAsync(ClientPropertyCollection, System.Threading.CancellationToken)"/> operation.
+    /// </summary>
+    public class ClientPropertiesUpdateResponse
+    {
+        /// <summary>
+        /// The request Id that is appended to the <see cref="InternalClient.UpdateClientPropertiesAsync(ClientPropertyCollection, System.Threading.CancellationToken)"/> operation.
+        /// </summary>
+        /// <remarks>
+        /// This request Id is relevant only for operations over MQTT, and can be used for debugging the operation from the service side.
+        /// </remarks>
+        public string RequestId { get; internal set; }
+
+        /// <summary>
+        /// The updated version after the property patch has been applied.
+        /// </summary>
+        public long Version { get; internal set; }
+    }
+}
--- a/iothub/device/src/ClientPropertyCollection.cs
+++ b/iothub/device/src/ClientPropertyCollection.cs
@@ -1,0 +1,378 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.Devices.Shared;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// A collection of properties for the client.
+    /// </summary>
+    public class ClientPropertyCollection : PayloadCollection
+    {
+        private const string VersionName = "$version";
+
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <summary>
+        /// Adds the value for the collection.
+        /// </summary>
+        /// <remarks>
+        /// If the collection has a key that matches the property name this method will throw an <see cref="ArgumentException"/>.
+        /// <para>
+        /// When using this as part of the writable property flow to respond to a writable property update you should pass in the value
+        /// as an instance of <see cref="PayloadSerializer.CreateWritablePropertyResponse(object, int, long, string)"/>
+        /// to ensure the correct formatting is applied when the object is serialized.
+        /// </para>
+        /// </remarks>
+        /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <c>null</c> </exception>
+        /// <param name="propertyName">The name of the property to add.</param>
+        /// <param name="propertyValue">The value of the property to add.</param>
+        public override void Add(string propertyName, object propertyValue)
+            => Add(null, propertyName, propertyValue);
+
+        /// <inheritdoc path="/remarks" cref="Add(string, object)" />
+        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <summary>
+        /// Adds the value for the collection.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <c>null</c> </exception>
+        /// <param name="componentName">The component with the property to add.</param>
+        /// <param name="propertyName">The name of the property to add.</param>
+        /// <param name="propertyValue">The value of the property to add.</param>
+        public void Add(string componentName, string propertyName, object propertyValue)
+            => AddInternal(new Dictionary<string, object> { { propertyName, propertyValue } }, componentName, false);
+
+        /// <inheritdoc path="/remarks" cref="Add(string, object)" />
+        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <summary>
+        /// Adds the value for the collection.
+        /// </summary>
+        /// <param name="properties">A collection of properties to add or update.</param>
+        public void Add(IDictionary<string, object> properties)
+            => AddInternal(properties, null, false);
+
+        /// <inheritdoc path="/remarks" cref="Add(string, object)" />
+        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <summary>
+        /// Adds the value for the collection.
+        /// </summary>
+        /// <param name="properties">A collection of properties to add or update.</param>
+        /// <param name="componentName">The component with the properties to add or update.</param>
+        public void Add(string componentName, IDictionary<string, object> properties)
+        => AddInternal(properties, componentName, false);
+
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <summary>
+        /// Adds a writable property to the collection.
+        /// </summary>
+        /// <remarks>
+        /// This method will use the <see cref="PayloadSerializer.CreateWritablePropertyResponse(object, int, long, string)"/> method to create an instance of <see cref="IWritablePropertyResponse"/> that will be properly serialized.
+        /// </remarks>
+        /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <c>null</c> </exception>
+        /// <param name="propertyName">The name of the property to add or update.</param>
+        /// <param name="propertyValue">The value of the property to add or update.</param>
+        /// <param name="statusCode"></param>
+        /// <param name="version"></param>
+        /// <param name="description"></param>
+        public void Add(string propertyName, object propertyValue, int statusCode, long version, string description = default)
+            => Add(null, propertyName, propertyValue, statusCode, version, description);
+
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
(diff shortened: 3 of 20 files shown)


iothub/device/src/ClientProperties.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// A container for properties retrieved from the service.
+    /// <remarks>
+    /// The Properties class is not meant to be constructed by customer code.
+    /// It is intended to be returned fully populated from the client method.
+    /// </remarks>
+    /// </summary>
+    public class ClientProperties : ClientPropertyCollection
```
abhipsaMisra, line 13: The holder for all client properties => similar to `Twin`.

abhipsaMisra: A `ClientPropertyCollection` is similar to what `TwinCollection` is.

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        /// <summary>
+        /// Converts a <see cref="TwinCollection"/> collection to a properties collection.
+        /// </summary>
+        /// <remarks>This internal class is aware of the implementation of the TwinCollection ad will </remarks>
+        /// <param name="twinCollection">The TwinCollection object to convert.</param>
+        /// <param name="payloadConvention">A convention handler that defines the content encoding and serializer to use for the payload.</param>
+        /// <returns>A new instance of the class from an existing <see cref="TwinProperties"/> using an optional <see cref="PayloadConvention"/>.</returns>
+        internal static ClientPropertyCollection FromTwinCollection(TwinCollection twinCollection, PayloadConvention payloadConvention)
```
abhipsaMisra, line 327: the plan is to merge the `Twin` and `Property` flows, post preview and before GA.

iothub/device/src/Common/UrlEncodedDictionarySerializer.cs
```diff
@@ -337,7 +342,11 @@ public IEnumerable<Token> GetTokens()
 
             private Token CreateToken(TokenType tokenType, int readCount)
             {
-                string tokenValue = readCount == 0 ? null : value.Substring(position - readCount, readCount);
+                // '?' is not a valid character for message property names or values, but instead signifies the start of a query string
+                // in the case of an MQTT topic. For this reason, we'll replace the '?' from the property key before adding it into
+                // application properties collection.
+                // https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct
```
abhipsaMisra, line 348: this dict is used when constructing a message from an mqtt packet

iothub/device/src/InternalClient.ConventionBasedOperations.cs
```diff
@@ -62,5 +63,52 @@ internal Task SubscribeToCommandsAsync(Func<CommandRequest, object, Task<Command
 
             return SetMethodDefaultHandlerAsync(methodDefaultCallback, userContext, cancellationToken);
         }
+
+        internal async Task<ClientProperties> GetClientPropertiesAsync(CancellationToken cancellationToken)
+        {
+            try
+            {
+                return await InnerHandler.GetPropertiesAsync(PayloadConvention, cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 71: await operation and throw if cancellation token is cancelled => note that this will exhibit a different behavior than what the existing APIs do, but since this is a new API we could introduce the desired behavior here. 
The question is, do we think that this is the desired behavior => throwing an OperationCancelledException directly if the cancellation token expires. I think that we should, given that it is the inherent behavior on cancellation token expiaryion, and hence, would be the expectation from APIs that use cancellation tokens.

iothub/device/src/MessageBase.cs
```diff
@@ -217,20 +208,11 @@ public string ConnectionModuleId
...
-            get => GetSystemProperty<string>(MessageSystemPropertyNames.ContentEncoding);
-            set => SystemProperties[MessageSystemPropertyNames.ContentEncoding] = value;
-        }
-
         /// <summary>
         /// The DTDL component name from where the telemetry message has originated.
         /// This is relevant only for plug and play certified devices.
         /// </summary>
-        internal string ComponentName
+        public string ComponentName
```
abhipsaMisra, line 215: this was accidentally changed to internal

iothub/device/src/MessageBase.cs
```diff
@@ -181,15 +181,6 @@ public DateTime CreationTimeUtc
         /// </summary>
         public bool IsSecurityMessage => CommonConstants.SecurityMessageInterfaceId.Equals(GetSystemProperty<string>(MessageSystemPropertyNames.InterfaceId), StringComparison.Ordinal);
 
-        /// <summary>
-        /// Used to specify the content type of the message.
-        /// </summary>
-        internal string PayloadContentType
```
abhipsaMisra, line 187: move internal below public properties

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -521,6 +521,20 @@ private async Task DisposeMessageAsync(string lockToken, AmqpIotDisposeActions o
 
         #endregion Accept-Dispose
 
+        #region Convention-based operations
+
+        public override Task<ClientProperties> GetPropertiesAsync(PayloadConvention payloadConvention, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException("Property operations are currently not supported over AMQP. Please use MQTT protocol instead.");
```
abhipsaMisra, line 528: the new APIs aren't supported over amqp yet

iothub/device/src/Transport/HttpTransportHandler.cs
```diff
@@ -397,6 +397,16 @@ public override Task RejectAsync(string lockToken, CancellationToken cancellatio
                 cancellationToken);
         }
 
+        public override Task<ClientProperties> GetPropertiesAsync(PayloadConvention payloadConvention, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException("Property operations are not supported over HTTP. Please use MQTT protocol instead.");
```
abhipsaMisra, line 402: twin property operations aren't supported over http

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -168,7 +168,7 @@ public override async Task WriteAsync(IChannelHandlerContext context, object dat
                     throw new IotHubCommunicationException("MQTT is disconnected.");
                 }
 
-                if (data is Message message)
+                if (data is MessageBase message)
```
abhipsaMisra, line 171: the transport layer should work with the base message class, which has all of the required functionality.

iothub/device/tests/Mqtt/MqttTransportHandlerTests.cs
```diff
@@ -369,7 +370,13 @@ public async Task MqttTransportHandlerSendTwinGetAsyncHappyPath()
...
                     var response = new Message(twinByteStream);
-                    response.MqttTopicName = GetResponseTopic(msg.Arg<Message>().MqttTopicName, statusSuccess);
+                    string mqttTopic = GetResponseTopic(msg.Arg<Message>().MqttTopicName, statusSuccess);
+                    response.MqttTopicName = mqttTopic;
+                    var publishPacket = new PublishPacket(QualityOfService.AtMostOnce, false, false)
+                    {
+                        TopicName = mqttTopic
+                    };
+                    MqttIotHubAdapter.PopulateMessagePropertiesFromPacket(response, publishPacket);
                     transport.OnMessageReceived(response);
```
abhipsaMisra, line 380: the transport layer receives an mqtt packet and then converts it to a Message. The tests were omitting the mqtt packet step, so it has been added back in.

(The reason this was working even when the mqtt packet step was omitted was because we were extracting the same information from the mqtt topic again in the transport layer: https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs#L1101)

iothub/device/src/ClientProperties.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="ClientProperties"/>.
+        /// </summary>
+        internal ClientProperties()
```
vinagesh, line 18: If this is public, you should add a public constructor that does nothing (and say for test purposes only) if there is value in this class being mockable.

abhipsaMisra: no harm in making this mockable anyway, I'll add a public no args constructor

iothub/device/src/ClientProperties.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// It is intended to be returned fully populated from the client method <see cref="InternalClient.GetClientPropertiesAsync(System.Threading.CancellationToken)"/>.
+    /// </remarks>
+    public class ClientProperties : ClientPropertyCollection
+    {
+        /// <summary>
+        /// Initializes a new instance of <see cref="ClientProperties"/>.
+        /// This is provided for unit testing purposes only.
+        /// </summary>
+        /// <inheritdoc path="/remarks" cref="ClientProperties" />
+        public ClientProperties()
```
timtay-microsoft, line 20: Should this be internal if it is for unit testing only?

iothub/device/src/ClientPropertiesUpdateResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// The response of an <see cref="InternalClient.UpdateClientPropertiesAsync(ClientPropertyCollection, System.Threading.CancellationToken)"/> operation.
+    /// </summary>
+    public class ClientPropertiesUpdateResponse
+    {
+        /// <summary>
+        /// The request Id that is appended to the <see cref="InternalClient.UpdateClientPropertiesAsync(ClientPropertyCollection, System.Threading.CancellationToken)"/> operation.
```
timtay-microsoft, line 12: How about "The request Id that is _associated_ to the..."?

iothub/device/src/ClientPropertiesUpdateResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This request Id is relevant only for operations over MQTT, and can be used for debugging the operation from the service side.
```
timtay-microsoft, line 15: Do users really have a use for this field? As far as I know, the SDK picks the value when it sends the request, and I've never seen anything on the service side that exposes this value to the user. I pictured it would only be useful for Kusto logs that only we would have access to

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -0,0 +1,378 @@
...
+        /// </remarks>
+        /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <c>null</c> </exception>
+        /// <param name="propertyName">The name of the property to add.</param>
+        /// <param name="propertyValue">The value of the property to add.</param>
+        public override void Add(string propertyName, object propertyValue)
+            => Add(null, propertyName, propertyValue);
+
+        /// <inheritdoc path="/remarks" cref="Add(string, object)" />
+        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
```
timtay-microsoft, line 37: Nice use of inheriting docs

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -0,0 +1,378 @@
...
+        /// <param name="propertyName">The name of the property to add.</param>
+        /// <param name="propertyValue">The value of the property to add.</param>
+        public void Add(string componentName, string propertyName, object propertyValue)
+            => AddInternal(new Dictionary<string, object> { { propertyName, propertyValue } }, componentName, false);
+
+        /// <inheritdoc path="/remarks" cref="Add(string, object)" />
+        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <summary>
+        /// Adds the value for the collection.
```
timtay-microsoft, line 52: Copy paste error? Wouldn't this be "Adds the value**s**" or "Adds all the value**s**"?

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -0,0 +1,378 @@
...
+        /// </summary>
+        /// <param name="properties">A collection of properties to add or update.</param>
```
timtay-microsoft, line 54: It seems like some of these methods allow for you to "add" a key that already exists in order to update, but some other methods don't. Is there some reason for that? Is that just how IDictionary works?

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -0,0 +1,378 @@
...
+        public void Add(IDictionary<string, object> properties)
+            => AddInternal(properties, null, false);
+
+        /// <inheritdoc path="/remarks" cref="Add(string, object)" />
+        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <inheritdoc path="/exception['ArgumentException']" cref="AddInternal(IDictionary{string, object}, string, bool)" />
+        /// <summary>
+        /// Adds the value for the collection.
```
timtay-microsoft, line 62: "value**s**" and "**to** the collection"

iothub/device/src/ClientPropertyCollection.cs
timtay-microsoft: Looking at ```public void AddOrUpdate(IDictionary<string, object> properties)``` I'm guessing this method is supposed to throw if one or more keys in the added collection already exist? That isn't reflected in this description, if that is the case

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -0,0 +1,378 @@
...
+
+        /// <summary>
+        /// Adds or updates the value for the collection.
+        /// </summary>
+        /// <seealso cref="PayloadConvention"/>
+        /// <seealso cref="PayloadSerializer"/>
+        /// <seealso cref="PayloadEncoder"/>
+        /// <param name="properties">A collection of properties to add or update.</param>
+        /// <param name="componentName">The component with the properties to add or update.</param>
+        /// <param name="forceUpdate">Forces the collection to use the Add or Update behavior. Setting to true will simply overwrite the value; setting to false will use <see cref="IDictionary{TKey, TValue}.Add(TKey, TValue)"/></param>
```
timtay-microsoft, line 212: Replace the semicolon with a period here since those are two separate sentences

iothub/device/src/ClientProperties.cs
abhipsaMisra: it'll also allow external users to be able to mock this, if they want

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -37,5 +36,31 @@ public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationTo
             Func<CommandRequest, object, Task<CommandResponse>> callback, object userContext,
             CancellationToken cancellationToken = default)
             => InternalClient.SubscribeToCommandsAsync(callback, userContext, cancellationToken);
+
+        /// <summary>
+        /// Retrieve the device properties.
```
timtay-microsoft, line 41: Should this be "Retrieve the **client** properties"?

iothub/device/src/ClientPropertiesUpdateResponse.cs
abhipsaMisra: You're correct, this would only be useful if users wanted to track some requests via a support ticket, similar to how they use a tracking Id with an exception. I don't think this will add value for every customer (or even most), but since we are returning a response body anyway (for the version), I thought I'd add it in.

iothub/device/src/ClientPropertyCollection.cs
abhipsaMisra: This is all @jamdavi ! :)

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -37,5 +36,31 @@ public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationTo
...
+        /// Update properties.
+        /// </summary>
+        /// <param name="propertyCollection">Reported properties to push.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <returns>The response of the update operation.</returns>
+        public Task<ClientPropertiesUpdateResponse> UpdateClientPropertiesAsync(ClientPropertyCollection propertyCollection, CancellationToken cancellationToken = default)
+            => InternalClient.UpdateClientPropertiesAsync(propertyCollection, cancellationToken);
+
+        /// <summary>
+        /// Sets the global listener for Writable properties
```
timtay-microsoft, line 58: "Sets the global listener for Writable property update events"?

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -37,5 +36,31 @@ public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationTo
...
+        /// <summary>
+        /// Retrieve the device properties.
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <returns>The device properties.</returns>
+        public Task<ClientProperties> GetClientPropertiesAsync(CancellationToken cancellationToken = default)
+            => InternalClient.GetClientPropertiesAsync(cancellationToken);
+
+        /// <summary>
+        /// Update properties.
```
timtay-microsoft, line 49: Are there any additional details you can provide here about how this API works? Is it a patch or a replace style update?

iothub/device/src/InternalClient.ConventionBasedOperations.cs
timtay-microsoft: I don't mind doing the correct try/catch/throw structure here considering it wouldn't be a breaking change.

iothub/device/src/ClientPropertyCollection.cs
abhipsaMisra: Correct, we have `Add`, which will use IDictionary.Add => will throw ArgumentException if the key already exists. 
`AddOrUpdate` will force-update the dictionary instead => `Collection[key] = value`.

iothub/device/src/ModuleClient.ConventionBasedOperations.cs
```diff
@@ -37,5 +36,31 @@ public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationTo
             Func<CommandRequest, object, Task<CommandResponse>> callback, object userContext,
             CancellationToken cancellationToken = default)
             => InternalClient.SubscribeToCommandsAsync(callback, userContext, cancellationToken);
+
+        /// <summary>
+        /// Retrieve the device properties.
```
timtay-microsoft, line 41: Same comment as for the device client method, but doubly so because this should at least be "module" instead of "device" if we don't want to say "Retrieve the client properties"

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
timtay-microsoft: I'm a little worried that users may misconstrue "Property operations" to include twin operations. Since we do support the legacy twin operations over AMQP, is there a better way to phrase this error?

iothub/device/src/ModuleClient.ConventionBasedOperations.cs
abhipsaMisra: I updated both to say "client", reduces chance of a similar copy-paste error in the future

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
abhipsaMisra: I'll add a note calling out that this is a patch operation and that we don't need to send the entire twin object.

abhipsaMisra: well, not twin, properties

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -88,6 +88,11 @@ internal sealed class MqttTransportHandler : TransportHandler, IMqttIotHubEventH
         private const string ReceiveEventMessagePatternFilter = "devices/{0}/modules/{1}/#";
         private const string ReceiveEventMessagePrefixPattern = "devices/{0}/modules/{1}/";
 
+        // Identifiers for property update operations.
+        public const string VersionKey = "$version";
+
+        public const string RequestIdKey = "$rid";
```
timtay-microsoft, line 94: I'm surprised we didn't have this just for regular twin/method operations already. Were we never sending a requestId for those?

timtay-microsoft: Ah, nevermind. I see below that we just had it locally defined instead

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
abhipsaMisra: Good point, how about:
"This API is currently not supported over AMQP, please use MQTT instead. Note that you can still retrieve the device twin properties using DeviceClient.GetTwinAsync()/ ModuleClient.GetTwinAsync()"

iothub/device/src/ClientPropertiesUpdateResponse.cs
timtay-microsoft: I'd add that additional bit of context to your remarks, then. Without it, it sounds like I'd be able to track requests by Id through the portal or something

iothub/device/src/ClientPropertyCollection.cs
timtay-microsoft: Then nice job, @jamdavi !

timtay-microsoft (approved):

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
abhipsaMisra: updated to:
```csharp
throw new NotImplementedException("This operation is currently not supported over AMQP, please use MQTT protocol instead. " +
    "Note that you can still retrieve a client's properties using the legacy DeviceClient.GetTwinAsync(CancellationToken cancellationToken) or " +
    "ModuleClient.GetTwinAsync(CancellationToken cancellationToken) operations, but the properties will not be formatted " +
    "as per DTDL terminology.");
```

PR #1975, opened 2021-05-20
Adding Telemetry E2E to PnP

jamdavi:


e2e/test/iothub/telemetry/TelemetrySendE2ETests.cs
```diff
@@ -0,0 +1,261 @@
...
+    public partial class TelemetrySendE2ETests : E2EMsTestBase
+    {
+        private const int MessageBatchCount = 5;
+        private const int LargeMessageSizeInBytes = 255 * 1024; // The maximum message size for device to cloud messages is 256 KB. We are allowing 1 KB of buffer for message header information etc.
+        private readonly string DevicePrefix = $"{nameof(TelemetrySendE2ETests)}_";
+        private readonly string ModulePrefix = $"{nameof(TelemetrySendE2ETests)}_";
+        private static string ProxyServerAddress = Configuration.IoTHub.ProxyServerAddress;
+
+        [LoggedTestMethod]
+        public async Task Telemetry_DeviceSendSingleTelemetry_Mqtt()
```
abhipsaMisra, line 25: what do you think about adding tests for amqp? even though we won't advertise support, technically you could still use telemetry over amqp (since we're simply calling SendEvenetAsync underneath.

e2e/test/iothub/telemetry/TelemetrySendE2ETests.cs
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        public static (Client.TelemetryMessage message, string p1Value) ComposeTelemetryMessageWithComponent(MsTestLogger logger)
+        {
+            string messageId = Guid.NewGuid().ToString();
+            string p1Value = Guid.NewGuid().ToString();
+            string userId = Guid.NewGuid().ToString();
+            string componentName = Guid.NewGuid().ToString();
+
+            logger.Trace($"{nameof(ComposeTelemetryMessage)}: messageId='{messageId}' userId='{userId}' p1Value='{p1Value}'");
```
abhipsaMisra, line 169: nit: ComposeTelemetryMessageWithComponent

e2e/test/iothub/telemetry/TelemetrySendE2ETests.cs
```diff
@@ -0,0 +1,261 @@
...
+            var message = new TelemetryMessage
+            {
+                MessageId = messageId,
+                UserId = userId,
+                ComponentName = componentName,
```
abhipsaMisra, line 174: we should maybe add a unit test that checks that setting the component name via TelemetryCollection constructor works as well?

e2e/test/iothub/telemetry/TelemetrySendE2ETests.cs
```diff
@@ -0,0 +1,261 @@
...
+            string p1Value = Guid.NewGuid().ToString();
+
+            logger.Trace($"{nameof(ComposeComplexTelemetryMessage)}: messageId='{messageId}' p1Value='{p1Value}'");
+            var message = new TelemetryMessage
+            {
+                MessageId = messageId,
+            };
+            message.Telemetry.Add("property1", p1Value);
+            message.Telemetry.Add("property2", null);
+            message.Telemetry.Add("property3", new
```
abhipsaMisra, line 196: do you think something like this would be better as a unit test? when we can check that we serialize and encode the message properly?

abhipsaMisra (approved):

e2e/test/iothub/telemetry/TelemetrySendE2ETests.cs
jamdavi: I agree we should. Let me add a work item for that so it doesn't get lost.

Issue #1971, opened 2021-05-19
[Bug Report] TwinCollection.ClearMetadata does not clean metadata for properties


florianbader:
# Context
- **OS, version, SKU and CPU architecture used:** Win 10 Desktop x64 (Windows 10 Desktop x64, Ubuntu 15.04 x86, Windows 10 IoT Core arm32, etc.)
- **Application's .NET Target Framework :** .NET 5 (See https://docs.microsoft.com/en-us/dotnet/standard/frameworks. E.g. netcoreapp2.1, net451, uap10.0, xamarin)
- **Device:** : Laptop (Laptop, Raspberry PI3, Android APIv25 etc.)
- **SDK version used:** Microsoft.Azure.Devices 1.30.0 (Please include the NuGet package version for all involved components)

## Description of the issue
Using the GetTwin method from RegistryManager returns the TwinProperties as TwinCollection. If using [ClearMetadata[(https://github.com/Azure/azure-iot-sdk-csharp/blob/8eefef4b1cacfcca708cab98778c780c30be1358/shared/src/TwinCollection.cs#L290) on the TwinCollection for e.g. the desired properties it will not clear the metadata for the properties.  Because of that if accessing a property afterwards it will return a TwinCollection instead of a JObject in the [TryGetMemberInternal](https://github.com/Azure/azure-iot-sdk-csharp/blob/8eefef4b1cacfcca708cab98778c780c30be1358/shared/src/TwinCollection.cs#L238) method. This is unexpected because after clearing the metadata I would expect a JObject which causes a RuntimeBindingException.

## Code sample exhibiting the issue
```csharp
var twin = await registryManager.GetTwinAsync("device");
twin.Properties.Desired.ClearMetadata();
twin.Properties.Desired["property"].GetType() // expected JObject, got TwinCollection
```

Current workaround which is really ugly is this:
```csharp
var twinPatch = new Twin(new TwinProperties { Desired = new TwinCollection(twin.Properties.Desired.ToJson()) });
```

jamdavi:
Hey @florianbader,

I was able to reproduce the same behavior. I see a way around this by separating the metadata JObject from the payload JObject. Let me reach out to the team internally to see if there is any negative implications to doing this.

jamdavi:
@florianbader We have added a new method called `ClearAllMetadata` this is a more extensive method that destroys the underlying metadata object and will have the desired effect you're looking for. I'll leave this issue open and update when we make a new release.

[...]

PR #1976, opened 2021-05-20
fix(shared): Add ClearAllMetadata() to clear all metadata

jamdavi:
Per issue #1971 the customer expects that when using RegistryManager.GetTwin() they should be able to clear the metadata from the TwinCollection and retrieve the JObject back from the collection instead of a new TwinCollection.

This seems reasonable as the name `ClearMeradata()` suggests that it will remove all Metadata from the TwinCollection

### DIFF
--- a/shared/src/TwinCollection.cs
+++ b/shared/src/TwinCollection.cs
@@ -96,7 +96,7 @@ namespace Microsoft.Azure.Devices.Shared
         }
 
         /// <summary>
-        /// Gets the count of properties in the Collection
+        /// Gets the count of properties in the Collection.
         /// </summary>
         public int Count
         {
@@ -165,7 +165,7 @@ namespace Microsoft.Azure.Devices.Shared
         }
 
         /// <summary>
-        /// Gets the Metadata for this property
+        /// Gets the Metadata for this property.
         /// </summary>
         /// <returns>Metadata instance representing the metadata for this property</returns>
         public Metadata GetMetadata()
@@ -185,7 +185,7 @@ namespace Microsoft.Azure.Devices.Shared
         }
 
         /// <summary>
-        /// Gets the LastUpdatedVersion for this property
+        /// Gets the LastUpdatedVersion for this property.
         /// </summary>
         /// <returns>LastUpdatdVersion if present, null otherwise</returns>
         public long? GetLastUpdatedVersion()
@@ -194,7 +194,7 @@ namespace Microsoft.Azure.Devices.Shared
         }
 
         /// <summary>
-        /// Gets the TwinProperties as a JSON string
+        /// Gets the TwinProperties as a JSON string.
         /// </summary>
         /// <param name="formatting">Optional. Formatting for the output JSON string.</param>
         /// <returns>JSON string</returns>
@@ -204,7 +204,7 @@ namespace Microsoft.Azure.Devices.Shared
         }
 
         /// <summary>
-        /// Determines whether the specified property is present
+        /// Determines whether the specified property is present.
         /// </summary>
         /// <param name="propertyName">The property to locate</param>
         /// <returns>true if the specified property is present; otherwise, false</returns>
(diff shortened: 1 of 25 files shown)


shared/src/TwinCollection.cs
```diff
@@ -289,10 +292,10 @@ private void TryClearMetadata(string propertyName)
...
         public void ClearMetadata()
         {
-            TryClearMetadata(MetadataName);
-            TryClearMetadata(LastUpdatedName);
-            TryClearMetadata(LastUpdatedVersionName);
-            TryClearMetadata(VersionName);
+            // GitHub Issue: https://github.com/Azure/azure-iot-sdk-csharp/issues/1971
+            // When we clear the metadata from the underlying collection we need to also clear
+            // the _metadata object so the TryGetMemberInternal will return a JObject instead of a new TwinCollection
+            _metadata.RemoveAll();
```
drwill-ms, line 298: What is the effective change in behavior between `TryClearMetadata(VersionName)` and this full clear? Might we ever expect a customer's app to be broken by this change?

jamdavi: > What is the effective change in behavior between `TryClearMetadata(VersionName)` and this full clear? 

TryClearMetadata just removes the property from the underlying payload dictionary but doesn't touch the _metadata object. Unsure of why it does this but it's been this way for a while.

Currently the ClearMetadata() method only clears the top level metadata and does not clear the property level metadata. 

> Might we ever expect a customer's app to be broken by this change?

The behavior of this method **WILL** change. An example would be if someone who is used to the behavior clears the metadata and then attempts to cast the property to TwinCollection.

```c#
var twin = await registryManager.GetTwinAsync("device");
twin.Properties.Desired.ClearMetadata(); // New implementation
var actualProp = completeTwin.Properties.Reported[propName]; // Returns dynamic, but underlying type is JObject
actualProp.ToJson(); // Fails beacause the member doesn't exist.
```

drwill-ms: Perhaps instead we should add a new customer-facing method that does a full clear?

jamdavi: Sounds reasonable.

shared/src/TwinCollection.cs
```diff
@@ -285,14 +296,33 @@ private void TryClearMetadata(string propertyName)
...
         }
+
+        /// <summary>
+        /// Clears all metadata out of the twin collection as well as all of the metadata from the properties collection.
+        /// </summary>
+        /// <remarks>
+        /// This will remove all metadata from the <see cref="TwinCollection"/> metadata object as well as the metadata for all of the properties collection. The difference from the <see cref="ClearMetadata"/> method is it will also clear the underlying metadata container which will affect methods such as <see cref="GetMetadata"/> and <see cref="GetLastUpdatedVersion"/>.
+        /// This method would be useful if you are performing any operations that require <see cref="TryGetMemberInternal(string, out object)"/> to return a <see cref="JToken"/> regardless of the client you are using.
+        /// </remarks>
+        public void ClearAllMetadata()
```
jamdavi, line 319: Similar named to the other method as it does somehting similar, however it's more destructive.

DestroyMetadata instead?

drwill-ms: I think the name, as you have it, is fine.

drwill-ms (approved):

shared/src/TwinCollection.cs
```diff
@@ -285,14 +296,33 @@ private void TryClearMetadata(string propertyName)
...
+        {
+            ClearMetadata();
+            // GitHub Issue: https://github.com/Azure/azure-iot-sdk-csharp/issues/1971
```
bikamani, line 322: I do like the comments on line 323 and 324 but does adding a link to github issue here help?

shared/src/TwinCollection.cs
```diff
@@ -285,14 +296,33 @@ private void TryClearMetadata(string propertyName)
         }
 
         /// <summary>
-        /// Clear metadata out of the collection
+        /// Clears metadata out of the properties collection.
```
bikamani, line 299: By properties collection do you mean twin collection?

shared/src/TwinCollection.cs
bikamani: I like this one.

bikamani (approved):

vinagesh (approved):

shared/src/TwinCollection.cs
jamdavi: I do think it helps since we might need to go back to it at some point in time "Why does this clear both" type of question.

shared/src/TwinCollection.cs
jamdavi: I do, and that is more clear

PR #1977, opened 2021-05-20
feat(device): add support for DisposeAsync

drwill-ms:
I'm not sure how to go about unit or E2E testing this. Ideas?

### DIFF
--- a/iothub/device/src/DeviceClient.cs
+++ b/iothub/device/src/DeviceClient.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
@@ -16,6 +17,9 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     /// <threadsafety static="true" instance="true" />
     public class DeviceClient : IDisposable
+#if !NET451 && !NET472 && !NETSTANDARD2_0
+        , IAsyncDisposable
+#endif
     {
         /// <summary>
         /// Default operation timeout.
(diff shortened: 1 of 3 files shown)


iothub/device/src/DeviceClient.cs
```diff
@@ -613,12 +617,45 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
         /// Releases the unmanaged resources used by the DeviceClient and optionally disposes of the managed resources.
         /// </summary>
+        /// <remarks>
+        /// The method <see cref="CloseAsync()"/> should be called before disposing.
+        /// </remarks>
         public void Dispose()
         {
             Dispose(true);
             GC.SuppressFinalize(this);
         }
```
vinagesh, line 628: Should this block of code not be under #if !NET451 || NET472 || NETSTANDARD2_0 ?

vinagesh (approved):

iothub/device/src/DeviceClient.cs
```diff
@@ -613,12 +617,45 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
 
+        /// <summary>
+        /// Disposes the client in an async way. See <see cref="IAsyncDisposable"/> for more information.
+        /// </summary>
+        /// <remarks>
+        /// Includes a call to <see cref="CloseAsync()"/>.
+        /// </remarks>
+        /// <example>
+        /// <code>
+        /// await using var client = DeviceClient.CreateFromConnectionString(...);
```
vinagesh, line 637: Does this mean if you use the `await` here, when time comes to dispose, the `DisposeAsync` will be called automatically?

iothub/device/src/DeviceClient.cs
drwill-ms: It doesn't hurt to have it there. It is just a method someone could call, even if they don't have .NET 5.0. The fact it comes from IAsyncDisposable interface, which is only supported in .NET 5.0 is what matters most.

That said, I don't have a problem wrapping it if people prefer that.

iothub/device/src/DeviceClient.cs
drwill-ms: you have to use `await` otherwise `DisposeAsync()` won't be awaited.

drwill-ms: Correction, if you don't use `await` it calls `Dispose()` instead of `DisposeAsync()`.

iothub/device/src/DeviceClient.cs
drwill-ms: I've added it, mainly because the doc comments fail to resolve the type.

abhipsaMisra (approved):
ModuleClient as well?

PR #1979, opened 2021-05-20
[PnP] Add Thermostat and TemperatureController samples

abhipsaMisra:
This PR adds the pnp samples to the preview branch: https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/master/iot-hub/Samples/device/PnpDeviceSamples
These samples have been updated to use the new proposed API.

Once we GA (merge code to master), we'll move these samples to the samples repository.


iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,375 @@
...
+            }
+        }
+
+        // Report the property updates on "deviceInformation" component.
+        private async Task UpdateDeviceInformationAsync(CancellationToken cancellationToken)
+        {
+            const string componentName = "deviceInformation";
+            var deviceInformationProperties = new Dictionary<string, object>
+            {
+                { "manufacturer", "element15" },
```
jamdavi, line 273: nit: tab

iothub/device/samples/convention-based-samples/TemperatureController/SystemTextJsonPayloadConvention.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.Devices.Shared;
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    /// <summary>
+    /// A <see cref="PayloadConvention"/> that uses <see cref="SystemTextJsonPayloadSerializer"/>.
+    /// </summary>
+    public class SystemTextJsonPayloadConvention : PayloadConvention
```
drwill-ms, line 11: Why make this class?

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,375 @@
...
+        }
+
+        private async Task HandleTargetTemperatureUpdateRequestAsync(string componentName, double targetTemperature, long version, object userContext)
+        {
+            const string targetTemperatureProperty = "targetTemperature";
+            _logger.LogDebug($"Property: Received - component=\"{componentName}\", {{ \"{targetTemperatureProperty}\": {targetTemperature}°C }}.");
+
+            // Update Temperature in 2 steps
+            double step = (targetTemperature - _temperature[componentName]) / 2d;
+            for (int i = 1; i <= 2; i++)
```
drwill-ms, line 123: If you are executing twice because `_temperature` array is of size 2, why not base it off the size of that array? Safer, no?

iothub/device/samples/convention-based-samples/TemperatureController/SystemTextJsonPayloadConvention.cs
abhipsaMisra: TemperatureControllerSample uses System.Text.Json as its object serializer while Thermostat uses NewtonSoft.Json. I thought we'd demonstrate both options this way. While we do provide a SystemTextJsonSerializer in our assembly, we won't be providing the corresponding Convention implementation. We'll only provide a NewtonSoft based convention in the assembly, by default.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,375 @@
...
+                }
+            }
+        }
+
+        private async Task HandleTargetTemperatureUpdateRequestAsync(string componentName, double targetTemperature, long version, object userContext)
+        {
+            const string targetTemperatureProperty = "targetTemperature";
+            _logger.LogDebug($"Property: Received - component=\"{componentName}\", {{ \"{targetTemperatureProperty}\": {targetTemperature}°C }}.");
+
+            // Update Temperature in 2 steps
```
drwill-ms, line 121: I think a more accurate comment is to update the temperature on both components

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,375 @@
...
+        {
+            const string propertyName = "serialNumber";
+            const string serialNumber = "SR-123456";
+
+            var reportedProperties = new ClientPropertyCollection();
+            reportedProperties.Add(propertyName, serialNumber);
+
+            ClientPropertiesUpdateResponse updateResponse = await _deviceClient.UpdateClientPropertiesAsync(reportedProperties, cancellationToken);
+
+            _logger.LogDebug($"Property: Update - {{ {reportedProperties.GetSerializedString()} }} is complete " +
```
drwill-ms, line 317: What is the reason for the double curly braces? It looks like you are escaping them, but it makes the formatable string hard to read. Can we just go with a different text format?

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: The fact that we're incrementing the value in "2" steps is a randomly chosen value; just to simulate an actual temperature change event (instead of it going from 15 (current) to 25 (desired) directly). It actually doesn't have any relation to the size of the dict (no. of components).

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,375 @@
...
+            // -> Periodically send "temperature" over telemetry - on "Thermostat" components.
+            // -> Send "maxTempSinceLastReboot" over property update, when a new max temperature is set - on "Thermostat" components.
+
+            _logger.LogDebug("Subscribe to writable property updates.");
+            await _deviceClient.SubscribeToWritablePropertiesEventAsync(HandlePropertyUpdatesAsync, null, cancellationToken);
+
+            _logger.LogDebug($"Subscribe to commands.");
+            await _deviceClient.SubscribeToCommandsAsync(HandleCommandsAsync, null, cancellationToken);
+
+            await UpdateDeviceInformationAsync(cancellationToken);
```
drwill-ms, line 61: Seems on start up, the client should get twin to see what the latest writable setting values are. Or does that happen automatically with a callback?

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,375 @@
...
+            }
+        }
+
+        private async Task HandleTargetTemperatureUpdateRequestAsync(string componentName, double targetTemperature, long version, object userContext)
+        {
+            const string targetTemperatureProperty = "targetTemperature";
+            _logger.LogDebug($"Property: Received - component=\"{componentName}\", {{ \"{targetTemperatureProperty}\": {targetTemperature}°C }}.");
+
+            // Update Temperature in 2 steps
+            double step = (targetTemperature - _temperature[componentName]) / 2d;
```
drwill-ms, line 122: Should have more protection around supported component names, otherwise a bad service app could crash the device.

iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
```diff
@@ -0,0 +1,205 @@
...
+                await Task.Delay(5 * 1000);
+            }
+        }
+
+        // The callback to handle property update requests.
+        private async Task HandlePropertyUpdatesAsync(ClientPropertyCollection writableProperties, object userContext)
+        {
+            foreach (KeyValuePair<string, object> writableProperty in writableProperties)
+            {
+                switch (writableProperty.Key)
```
drwill-ms, line 72: Could use a comment about what the key is expected to be.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: What I mean to convey is: this will simulate the temperature update in two steps; i.e. if the current temp is 10 and the desired is 30, it'll go 10 (current) => 20 => 30 (desired).  It'll still happen on the single component on which it was requested though.

abhipsaMisra: I'll add an example to describe what I mean

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
drwill-ms: What is the purpose of updating it in two steps? To show it could be long-running?

iothub/device/samples/convention-based-samples/TemperatureController/SystemTextJsonPayloadConvention.cs
drwill-ms: Why not provide a convention impl?

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: Yes, the intension is to escape them, is there a different format that can be used for braces when using string interpolation?

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: No, the sample does not have that logic implemented right now. If you think it is worth adding here, we could add it into the serial number update flow, just to demonstrate how to get the current property value and update only if required.
This will be a non-writable (from service perspective) property update though. Writable property updated are currently only executed through the callback (and are always executed). We could add a similar check there as well, if we want.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: sure, I'll add some checks here

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: yeah, it'll simulate scenarios where you perform some action within your callback, have delays etc; so a bit closer to how a real device might behave.

iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
abhipsaMisra: sure, I'll add

iothub/device/samples/convention-based-samples/TemperatureController/SystemTextJsonPayloadConvention.cs
abhipsaMisra: No reason, we can add one if it'll be helpful. Since building the serializer is the most involved process, we added a sample for only that step.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: I've added the check for the serial number flow, the target temperature (writable) flow will be similar, so I didn't add it there again.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
drwill-ms: I'm a fan of square brackets. You could use quotation marks. You could choose to surround it with nothing.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
drwill-ms: When we don't follow best practices in our samples, and then wonder why our customers do things we don't suggest, we shouldn't be surprised. Let's stick to best practices.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,391 @@
...
+            // -> Send "maxTempSinceLastReboot" over property update, when a new max temperature is set - on "Thermostat" components.
+
+            _logger.LogDebug("Subscribe to writable property updates.");
+            await _deviceClient.SubscribeToWritablePropertiesEventAsync(HandlePropertyUpdatesAsync, null, cancellationToken);
+
+            _logger.LogDebug($"Subscribe to commands.");
+            await _deviceClient.SubscribeToCommandsAsync(HandleCommandsAsync, null, cancellationToken);
+
+            await UpdateDeviceInformationAsync(cancellationToken);
+            await SendDeviceSerialNumberAsync(cancellationToken);
```
vinagesh, line 62: Maybe call this SendTelemetry as what we are showing in this is how to send telemetry?

iothub/device/samples/convention-based-samples/TemperatureController/SystemTextJsonPayloadConvention.cs
abhipsaMisra: There are a bunch of customizations that are required for system.text.json, so we're planning to ship preview v1 without system.text.json in our assembly, but rather to add it to samples. Subsequently we'll promote system.text.json based serializer and convention to the assembly itself. What do you say, @drwill-ms ?

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
vinagesh: I was sort of guessing the reason for this 2 step update till I saw this. I'm wondering if this is really needed in a sample? Why not keep it simple and just update it at once? I feel the aim is to show best practices to use the APIs but not guess user scenarios. To me it feels like that logic should be determined by the user based on the use case.

abhipsaMisra: sure, I don't mind a one-step update. I just felt a 2-step better simulates a real scenario

vinagesh:
Few suggestions:

1) It would be good to have separate PRs for the two samples so we could review and approve faster.
2) For the Temperature Controller sample

Most of the methods are Send xyz. To me they all seemed like sending telemetry but when looking in detail, some were UpdateClientProperties. Ex: SendDeviceSerialNumberAsync.

The other thing that was not clear to me was why are there so many examples for sending telemetry until I realized one is for a component and other is not. It might be good to have this sample explicitly call it out.
Ex: SendRootLevelTelemetry, SendComponentTelemetry, HandleRoolLevelCommand, HandleComponentCommand.
Something like this might help understand the reason for the TemperatureController sample with nested components vs the simple Thermostat sample.

abhipsaMisra:
> 1. It would be good to have separate PRs for the two samples so we could review and approve faster.

Sure, I'll slipt them.

> 2. For the Temperature Controller sample
> 
> Most of the methods are Send xyz. To me they all seemed like sending telemetry but when looking in detail, some were UpdateClientProperties. Ex: SendDeviceSerialNumberAsync.

I can call out the operation in those method names as well, sure.

> The other thing that was not clear to me was why are there so many examples for sending telemetry until I realized one is for a component and other is not. It might be good to have this sample explicitly call it out.
> Ex: SendRootLevelTelemetry, SendComponentTelemetry, HandleRoolLevelCommand, HandleComponentCommand.
> Something like this might help understand the reason for the TemperatureController sample with nested components vs the simple Thermostat sample.

I wouldn't want the method name to say "root level" and "component-level", because I don't think this reads well "SendRootLevelDeviceSerialNumberAsync"; but how would you feel if I added comments to the method calls that explained if this were a root-level or component-level call. Would that help?

iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
```diff
@@ -0,0 +1,191 @@
...
+                default:
+                    _logger.LogWarning($"Received a command request that isn't" +
+                        $" implemented - command name = {commandRequest.CommandName}");
+
+                    return Task.FromResult(new CommandResponse(StatusCodes.NotFound));
+            }
+        }
+
+        // Send temperature updates over telemetry. The sample also sends the value of max temperature since last reboot over reported property update.
+        private async Task SendTemperatureAsync()
```
vinagesh, line 151: Why not make the main function call SendTemperatureTelemetryAsync and ReportMaxTemperatureSinceLastRebootAsync separately? The reason I felt that would be more readable was because they are both showing different APIs in the SDK and felt like they would be explicitly called out that way.

vinagesh (commented):
> > 1. It would be good to have separate PRs for the two samples so we could review and approve faster.
> 
> Sure, I'll split them.
> 
> > 1. For the Temperature Controller sample
> > 
> > Most of the methods are Send xyz. To me they all seemed like sending telemetry but when looking in detail, some were UpdateClientProperties. Ex: SendDeviceSerialNumberAsync.
> 
> I can call out the operation in those method names as well, sure.
> 
> > The other thing that was not clear to me was why are there so many examples for sending telemetry until I realized one is for a component and other is not. It might be good to have this sample explicitly call it out.
> > Ex: SendRootLevelTelemetry, SendComponentTelemetry, HandleRoolLevelCommand, HandleComponentCommand.
> > Something like this might help understand the reason for the TemperatureController sample with nested components vs the simple Thermostat sample.
> 
> I wouldn't want the method name to say "root level" and "component-level", because I don't think this reads well "SendRootLevelDeviceSerialNumberAsync"; but how would you feel if I added comments to the method calls that explained if this were a root-level or component-level call. Would that help?

Don't split them now. Just something for the future :)
Sure comments would work as well. But add them both at the main program level (PerformOperationsAsync) and the function itself to make it obvious.

iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
abhipsaMisra: the reason the main function doesn't call them directly is because UpdateMaxTemperatureSinceLastRebootPropertyAsync isn't always called, it is called only based on some calculation, and it seemed cleaner to abstract away the calculation logic to a different method. I've added the comment to the main func, which should help highlight the 2 calls happening in here.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,372 @@
...
+            ClientPropertiesUpdateResponse updateResponse = await _deviceClient.UpdateClientPropertiesAsync(reportedProperty);
+
+            _logger.LogDebug($"Property: Update - component=\"{componentName}\", {reportedProperty.GetSerializedString()} is {nameof(StatusCodes.OK)} " +
+                $"with a version of {updateResponse.Version}.");
+        }
+
+        // The callback to handle command invocation requests.
+        private Task<CommandResponse> HandleCommandsAsync(CommandRequest commandRequest, object userContext)
+        {
+            switch (commandRequest.ComponentName)
```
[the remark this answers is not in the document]
abhipsaMisra: so this will always need to be nested switches => switch on component name, and per-component switch on command name. Additionally, for the default case (component name = null) switch on the command name.

Adding an if-else for the component name will also work the same way, right -> if (component-name == expected value1) else if (component-name == expected value) else if (component name == null); ans each of the if blocks again having either if-else/ switch cases.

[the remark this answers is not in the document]
abhipsaMisra: even though this is a "command" callback, the only way to highlight the command name before the component name is to have the componentname == null case first.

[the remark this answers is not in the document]
vinagesh: I agree there needs to be some kind of pivot on the component name. But the switch case within default of one switch case for root level felt out of place. There could be if - else for the component name and separate switch cases for each. If you had switch on the command name, it would still work but you'll have to check component name and throw if it was not present. I was suggesting more of this format:

// Handle Commands defined on root level
If(commandRequest.ComponentName == null)
{
swtich(commandName)
}
// Handle commands defined on component level
else
{
switch (componentName):
.....
}

Ofcourse you can also have  if (component-name == expected value1) else if (component-name == expected value) else if (component name == null);  as you suggested. But at least separating out root level and component level seems important to me in this example.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -0,0 +1,396 @@
...
+        {
+            _deviceClient = deviceClient ?? throw new ArgumentNullException($"{nameof(deviceClient)} cannot be null.");
+            _logger = logger ?? LoggerFactory.Create(builer => builer.AddConsole()).CreateLogger<TemperatureControllerSample>();
+        }
+
+        public async Task PerformOperationsAsync(CancellationToken cancellationToken)
+        {
+            // Set handler to receive and respond to writable property update requests.
+            _logger.LogDebug("Subscribe to writable property updates.");
+            await _deviceClient.SubscribeToWritablePropertiesEventAsync(HandlePropertyUpdatesAsync, null, cancellationToken);
```
vinagesh, line 50: Can we not have writable properties on the root level? Just wondering if the example shows that we are only handling writable properties on the components.

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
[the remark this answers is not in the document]
vinagesh: We can keep going back and forth on this as it is an implementation style. But I would add more comments to what you already have to add clarity if that is the direction you take.

vinagesh (approved):

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
abhipsaMisra: We can, for sure, but the DTDL model that this sample uses does not define any root-level writable property for TemperatureController. We have a task for enhancing these DTDL models to include more scenarios, but we'll take it up post preview.

drwill-ms (approved):

iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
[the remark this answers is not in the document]
abhipsaMisra: sure, I'll add some comments to describe the approach; leaving it as-is does make it a bit ambiguous.

iothub/device/samples/convention-based-samples/TemperatureController/Parameter.cs
```diff
@@ -0,0 +1,86 @@
...
+            Required = false,
+            HelpText = "The running time for this console application. Leave it unassigned to run the application until it is explicitly canceled using Control+C.")]
+        public double? ApplicationRunningTime { get; set; }
+
+        public bool Validate(ILogger logger)
+        {
+            if (string.IsNullOrWhiteSpace(DeviceSecurityType))
+            {
+                logger.LogWarning("Device provisioning type not set, please set the environment variable \"IOTHUB_DEVICE_SECURITY_TYPE\"" +
+                    "or pass in \"-s | --DeviceSecurityType\" through command line. \nWill default to using \"dps\" flow.");
```
bikamani, line 69: Do we still default to dps?

Trying to recall what is the reasoning or was it the change only for bug bash?

abhipsaMisra: I'm not changing this part of the logic, since this is what the readme refers to. I am only focusing on updating the API calls.
If we want to update this flow, we can add a task and take it up later.

PR #1983, opened 2021-05-21
[PnP] Adding Command Test to E2E

jamdavi:
## Checklist

### DIFF
--- a/e2e/test/iothub/command/CommandE2ETests.cs
+++ b/e2e/test/iothub/command/CommandE2ETests.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-#if !NET451
 using System;
 using System.Diagnostics.Tracing;
 using System.Net;
@@ -69,12 +68,41 @@ namespace Microsoft.Azure.Devices.E2ETests.Commands
 
         public static async Task DigitalTwinsSendCommandAndVerifyResponseAsync(string deviceId, string componentName, string commandName, MsTestLogger logger)
         {
-            DigitalTwinClient digitalTwinClient = null;
-            digitalTwinClient = DigitalTwinClient.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
-
             string payloadToSend = JsonConvert.SerializeObject(new ServiceCommandRequestObject { A = 123 });
+            string responseExpected = JsonConvert.SerializeObject(new DeviceCommandResponse());
+            string payloadReceived = string.Empty;
+            int statusCode = 0;
+#if NET451
+
+            ServiceClient serviceClient = ServiceClient.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
 
             logger.Trace($"{nameof(DigitalTwinsSendCommandAndVerifyResponseAsync)}: Invoke command {commandName}.");
+            
+            CloudToDeviceMethodResult serviceClientResponse = null;
+            if (string.IsNullOrEmpty(componentName))
+            {
+                var serviceCommand = new CloudToDeviceMethod(commandName).SetPayloadJson(payloadToSend);
+                serviceClientResponse =
+                await serviceClient.InvokeDeviceMethodAsync(
+                    deviceId, serviceCommand).ConfigureAwait(false);
+            }
+            else
+            {
+                var serviceCommand = new CloudToDeviceMethod($"{componentName}*{commandName}").SetPayloadJson(payloadToSend);
+                serviceClientResponse =
+                await serviceClient.InvokeDeviceMethodAsync(
+                    deviceId, serviceCommand).ConfigureAwait(false);
+            }
+
+            statusCode = serviceClientResponse.Status;
+            payloadReceived = serviceClientResponse.GetPayloadAsJson();
+
+            serviceClient.Dispose();
+#else
+            DigitalTwinClient digitalTwinClient = DigitalTwinClient.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+
+            logger.Trace($"{nameof(DigitalTwinsSendCommandAndVerifyResponseAsync)}: Invoke command {commandName}.");
+
             Rest.HttpOperationResponse<DigitalTwinCommandResponse, DigitalTwinInvokeCommandHeaders> response = null;
             if (string.IsNullOrEmpty(componentName))
             {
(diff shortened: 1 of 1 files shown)


e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if !NET451
```
abhipsaMisra, line 4: why !net451? commands are supposed to be supported on net451

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
...
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Common.Exceptions;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests.Commands
+{
+    public static class ServiceCommandRequestAssertion
+    {
+        public static int a => 123;
```
abhipsaMisra, line 19: uppercase, since it is a property

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    public class ServiceCommandRequestObject
+    {
+        public int a { get; set; }
```
abhipsaMisra, line 24: same, here too

abhipsaMisra: and below

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
...
+                        var valueToTest = request.GetData<ServiceCommandRequestObject>();
+                        if (!string.IsNullOrEmpty(componentName))
+                        {
+                            Assert.AreEqual(componentName, request.ComponentName, $"The expected component name should be {componentName} but was {request.ComponentName}");
+                        } else
+                        {
+                            Assert.AreEqual(null, request.ComponentName, $"The expected component name should be null but was {request.ComponentName}");
+                        }
+                        
+                        Assert.AreEqual(ServiceRequestJson, request.DataAsJson, $"The expected respose payload should be {ServiceRequestJson} but was {request.DataAsJson}");
```
abhipsaMisra, line 180: nit: typo - response

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
...
+                null)
+                .ConfigureAwait(false);
+
+            await deviceClient.SubscribeToCommandsAsync(null, null).ConfigureAwait(false);
+
+            // Return the task that tells us we have received the callback.
+            return methodCallReceived.Task;
+        }
+
+        public static async Task<Task> SetDeviceReceiveCommandAsync(DeviceClient deviceClient, string componentName, string methodName, MsTestLogger logger)
```
abhipsaMisra, line 160: nit: methodName -> commandName

abhipsaMisra: nit: and other "method" usages

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            await deviceClient.SubscribeToCommandsAsync(
+                (request, context) =>
+                {
+                    logger.Trace($"{nameof(SetDeviceReceiveCommandAsync)}: DeviceClient method: {request.CommandName}.");
+
+                    try
+                    {
+                        var valueToTest = request.GetData<ServiceCommandRequestObject>();
+                        if (!string.IsNullOrEmpty(componentName))
```
abhipsaMisra, line 172: extreme nit: what do think about reversing the checks, since is it usually easier to evaluate the positive case than having to evaluate the expression and then reverse it

e2e/test/iothub/command/CommandE2ETests.cs
abhipsaMisra: https://marketplace.visualstudio.com/items?itemName=EWoodruff.VisualStudioSpellCheckerVS2017andLater :)

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
...
+    {
+        public string name => "e2e_test";
+    }
+
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    public class CommandE2ETests : E2EMsTestBase
+    {
+        public const string DeviceResponseJson = "{\"name\":\"e2e_test\"}";
```
abhipsaMisra, line 37: Q - instead of using json string here, could we deal with objects on both sides -> service and device flow?

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            string componentName = null;
+            if (withComponent)
+            {
+                componentName = ComponentName;
+            }
+
+            await subscribeAndUnsubscribeMethod(deviceClient, componentName, CommandName, Logger).ConfigureAwait(false);
+
+            await SubscribeAndUnsubscribeCommandAsync(deviceClient, componentName, CommandName, Logger);
```
abhipsaMisra, line 210: L208 and L210 call the same function, what did we want to do here?

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -0,0 +1,238 @@
...
+                    else
+                    {
+                        logger.Trace($"{nameof(SubscribeAndUnsubscribeCommandAsync)}: DeviceClient method: {request.ComponentName} {request.CommandName}.");
+                    }
+                    return Task.FromResult(new CommandResponse(new DeviceCommandResponse(), Shared.StatusCodes.OK));
+                },
+                null)
+                .ConfigureAwait(false);
+
+            await deviceClient.SubscribeToCommandsAsync(null, null).ConfigureAwait(false);
```
abhipsaMisra, line 154: are we simply testing that we can call `SubscribeToCommandsAsync` multiple times/ call it wilh null callback? This seems like something that should be unit tested instead, right?

abhipsaMisra: Or if we want to test the whole subscribe -> unsubscribe flow, then we'd have to test that we can subscribe -> receive the callback -> unsubscribe -> should not receive the callback.

e2e/test/iothub/command/CommandE2ETests.cs
jamdavi: Actually that's fine. I had my others this way

e2e/test/iothub/command/CommandE2ETests.cs
jamdavi: Actually, yeah, this was some odd refactoring thing. I removed this test as it makes no sense.

e2e/test/iothub/command/CommandE2ETests.cs
jamdavi: DigitalTwinClient is not supported on 451...

image

abhipsaMisra: we could use service client for net451 and digital twin for the others, or go back to service client for all (as you did have previously); your call. I thought using digital twin client would simplify the tests and make them more pnp-fied, but we do still want to test against all supported frameworks.

abhipsaMisra (approved):
only comment is to include net451, everything else looks good!

jamdavi:
image

PR #1984, opened 2021-05-21
[PnP] Adding Properties E2E Tests

jamdavi:
This PR adds end-to-end tests for properties operations.

### DIFF
--- a/e2e/test/Helpers/TestDeviceCallbackHandler.cs
+++ b/e2e/test/Helpers/TestDeviceCallbackHandler.cs
@@ -29,6 +29,10 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         private ExceptionDispatchInfo _receiveMessageExceptionDispatch;
         private Message _expectedMessageSentByService = null;
 
+        private readonly SemaphoreSlim _clientPropertyCallbackSemaphore = new SemaphoreSlim(0, 1);
+        private ExceptionDispatchInfo _clientPropertyExceptionDispatch;
+        private object _expectedClientPropertyValue = null;
+
         public TestDeviceCallbackHandler(DeviceClient deviceClient, TestDevice testDevice, MsTestLogger logger)
         {
             _deviceClient = deviceClient;
@@ -46,6 +50,12 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         {
             get => Volatile.Read(ref _expectedMessageSentByService);
             set => Volatile.Write(ref _expectedMessageSentByService, value);
+        }
+
+        public object ExpectedClientPropertyValue
+        {
+            get => Volatile.Read(ref _expectedClientPropertyValue);
+            set => Volatile.Write(ref _expectedClientPropertyValue, value);
         }
 
         public async Task SetDeviceReceiveMethodAsync(string methodName, string deviceResponseJson, string expectedServiceRequestJson)
@@ -156,6 +166,42 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         {
             await _receivedMessageCallbackSemaphore.WaitAsync(ct).ConfigureAwait(false);
             _receiveMessageExceptionDispatch?.Throw();
+        }
+
+        public async Task SetClientPropertyUpdateCallbackHandlerAsync<T>(string expectedPropName)
+        {
+            string userContext = "myContext";
+
+            await _deviceClient.SubscribeToWritablePropertiesEventAsync(
+                (patch, context) =>
+                {
+                    _logger.Trace($"{nameof(SetClientPropertyUpdateCallbackHandlerAsync)}: DeviceClient {_testDevice.Id} callback property: WritableProperty: {patch}, {context}");
+
+                    try
+                    {
+                        bool isPropertyPresent = patch.TryGetValue(expectedPropName, out T propertyFromCollection);
+                        isPropertyPresent.Should().BeTrue();
+                        propertyFromCollection.Should().BeEquivalentTo((T)ExpectedClientPropertyValue);
+                        context.Should().Be(userContext);
+                    }
+                    catch (Exception ex)
+                    {
+                        _clientPropertyExceptionDispatch = ExceptionDispatchInfo.Capture(ex);
+                    }
+                    finally
+                    {
+                        // Always notify that we got the callback.
+                        _clientPropertyCallbackSemaphore.Release();
+                    }
+
+                    return Task.FromResult(true);
+                }, userContext).ConfigureAwait(false);
+        }
+
+        public async Task WaitForClientPropertyUpdateCallbcakAsync(CancellationToken ct)
+        {
+            await _clientPropertyCallbackSemaphore.WaitAsync(ct).ConfigureAwait(false);
+            _clientPropertyExceptionDispatch?.Throw();
         }
 
         public void Dispose()
--- a/e2e/test/Helpers/TestDeviceCallbackHandler.cs
+++ b/e2e/test/Helpers/TestDeviceCallbackHandler.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Linq;
 using System.Runtime.ExceptionServices;
 using System.Text;
 using System.Threading;
@@ -60,31 +59,34 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
 
         public async Task SetDeviceReceiveMethodAsync(string methodName, string deviceResponseJson, string expectedServiceRequestJson)
         {
-            await _deviceClient.SetMethodHandlerAsync(methodName,
-                (request, context) =>
-                {
-                    try
+            await _deviceClient
+                .SetMethodHandlerAsync(
+                    methodName,
+                    (request, context) =>
                     {
-                        _logger.Trace($"{nameof(SetDeviceReceiveMethodAsync)}: DeviceClient {_testDevice.Id} callback method: {request.Name} {request.ResponseTimeout}.");
-                        request.Name.Should().Be(methodName, "The expected method name should match what was sent from service");
-                        request.DataAsJson.Should().Be(expectedServiceRequestJson, "The expected method data should match what was sent from service");
+                        try
+                        {
+                            _logger.Trace($"{nameof(SetDeviceReceiveMethodAsync)}: DeviceClient {_testDevice.Id} callback method: {request.Name} {request.ResponseTimeout}.");
+                            request.Name.Should().Be(methodName, "The expected method name should match what was sent from service");
+                            request.DataAsJson.Should().Be(expectedServiceRequestJson, "The expected method data should match what was sent from service");
 
-                        return Task.FromResult(new MethodResponse(Encoding.UTF8.GetBytes(deviceResponseJson), 200));
-                    }
-                    catch (Exception ex)
-                    {
-                        _logger.Trace($"{nameof(SetDeviceReceiveMethodAsync)}: Error during DeviceClient callback method: {ex}.");
+                            return Task.FromResult(new MethodResponse(Encoding.UTF8.GetBytes(deviceResponseJson), 200));
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.Trace($"{nameof(SetDeviceReceiveMethodAsync)}: Error during DeviceClient callback method: {ex}.");
 
-                        _methodExceptionDispatch = ExceptionDispatchInfo.Capture(ex);
-                        return Task.FromResult(new MethodResponse(500));
-                    }
-                    finally
-                    {
-                        // Always notify that we got the callback.
-                        _methodCallbackSemaphore.Release();
-                    }
-                },
-                null).ConfigureAwait(false);
+                            _methodExceptionDispatch = ExceptionDispatchInfo.Capture(ex);
+                            return Task.FromResult(new MethodResponse(500));
+                        }
+                        finally
+                        {
+                            // Always notify that we got the callback.
+                            _methodCallbackSemaphore.Release();
+                        }
+                    },
+                    null)
+                .ConfigureAwait(false);
         }
 
         public async Task WaitForMethodCallbackAsync(CancellationToken ct)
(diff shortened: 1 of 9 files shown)


e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -0,0 +1,492 @@
...
+            var propName = Guid.NewGuid().ToString();
+
+            logger.Trace($"{nameof(Properties_DeviceSetsPropertyAndGetsItBackAsync)}: name={propName}, value={propValue}");
+
+            var props = new ClientPropertyCollection();
+            props[propName] = propValue;
+            await deviceClient.UpdateClientPropertiesAsync(props).ConfigureAwait(false);
+
+            // Validate the updated twin from the device-client
+            Twin deviceTwin = await deviceClient.GetTwinAsync().ConfigureAwait(false);
```
abhipsaMisra, line 219: GetClientPropertiesAsync instead of GetTwinAsync?

e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -0,0 +1,492 @@
...
+                        [propName1] = new Dictionary<string, object>
+                        {
+                            [propName2] = null
+                        }
+                    })
+                .ConfigureAwait(false);
+            serviceTwin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+            Assert.IsTrue(serviceTwin.Properties.Reported.Contains(propName1));
+            string value2 = serviceTwin.Properties.Reported[propName1].ToString();
+            Assert.AreEqual(value2, propEmptyValue);
```
abhipsaMisra, line 449: repeated?

e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -0,0 +1,483 @@
...
+            await deviceClient.UpdateClientPropertiesAsync(patch).ConfigureAwait(false);
+            await deviceClient.CloseAsync().ConfigureAwait(false);
+
+            Twin serviceTwin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+            Assert.AreEqual<string>(serviceTwin.Properties.Reported[propName].ToString(), propValue);
+
+            Logger.Trace("verified " + serviceTwin.Properties.Reported[propName].ToString() + "=" + propValue);
+        }
+
+        private async Task Properties_ServiceDoesNotCreateNullPropertyInCollectionAsync(Client.TransportType transport)
```
abhipsaMisra, line 406: how about we also we add a test that verifies that updating an existing property patch with null value removes the property from the twin

e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -0,0 +1,483 @@
...
+            Assert.IsFalse(serviceTwin.Properties.Reported.Contains(propName1));
+
+            await deviceClient
+                .UpdateClientPropertiesAsync(
+                    new ClientPropertyCollection
+                    {
+                        [propName1] = new Dictionary<string, object>
+                        {
+                            [propName2] = null
+                        }
```
abhipsaMisra, line 433: interesting, I wasn't aware that propName2 wouldn't make it to the final twin

abhipsaMisra: I thought the removal of "null" value properties referred to only root level properties, and not nested ones.

e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -0,0 +1,483 @@
...
+            }
+            catch (IotHubException)
+            {
+                exceptionThrown = true;
+            }
+
+            Assert.IsTrue(exceptionThrown, "IotHubException was expected for updating reported property with an invalid property name, but was not thrown.");
+
+            Twin serviceTwin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+            Assert.IsFalse(serviceTwin.Properties.Reported.Contains(propName1));
```
abhipsaMisra, line 472: does propName2 make it through? is it an all-or-none operation? we should add explicit assertions for propeName2 as well, whatever be the case.

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
```diff
@@ -0,0 +1,502 @@
...
+            {
+                Assert.AreEqual(JsonConvert.SerializeObject(propFromCollection), JsonConvert.SerializeObject(propValue));
+            }
+            else
+            {
+                Assert.Fail($"The property {propName} was not found in the  collection");
+            }
+            // Validate the updated twin from the service-client
+            Twin completeTwin = await _registryManager.GetTwinAsync(deviceId).ConfigureAwait(false);
+            var actualProp = completeTwin.Properties.Reported[ComponentName][propName];
```
abhipsaMisra, line 234: nit: dynamic instead of var?

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
```diff
@@ -0,0 +1,502 @@
...
+            using var registryManager = RegistryManager.CreateFromConnectionString(Configuration.IoTHub.ConnectionString);
+
+            var twinPatch = new Twin();
+            twinPatch.Properties.Desired[componentName] = new
+            {
+                __t = "c",
+            };
+            if (propValue is List<object>)
+            {
+                twinPatch.Properties.Desired[componentName][propName] = (Newtonsoft.Json.Linq.JToken)(JsonConvert.DeserializeObject(JsonConvert.SerializeObject(propValue)));
```
abhipsaMisra, line 290: why do we need to serialize and deserialize it a list value? isn't Twin able to accept the object as-is?

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
```diff
@@ -0,0 +1,502 @@
...
+
+            await deviceClient
+                .UpdateClientPropertiesAsync(
+                    new ClientPropertyCollection
+                    {
+                        { ComponentName, propName1, null }
+                    })
+                .ConfigureAwait(false);
+            Twin serviceTwin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+            Assert.IsTrue(serviceTwin.Properties.Reported.Contains(ComponentName));
```
abhipsaMisra, line 434: assertion for component name containing an empty value? we should probably check with service if that is the expected behavior, i.e. to remove an updated property we pass in an empty value, how to remove a component?

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
```diff
@@ -0,0 +1,502 @@
...
+            }
+            catch (IotHubException)
+            {
+                exceptionThrown = true;
+            }
+
+            Assert.IsTrue(exceptionThrown, "IotHubException was expected for updating reported property with an invalid property name, but was not thrown.");
+
+            Twin serviceTwin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
+            Assert.IsFalse(serviceTwin.Properties.Reported.Contains(propName1));
```
abhipsaMisra, line 491: shouldn't this assertion be `serviceTwin.Properties.Reported[ComponentName].Contains(propName1)`?

abhipsaMisra (approved):
A couple of questions

abhipsaMisra:
I wonder if it is adding fault injection tests for the two new APIs (GetClientProperties, UpdateClientProperties), since the transport layer is new for those.

e2e/test/iothub/properties/PropertiesE2ETests.cs
abhipsaMisra: added

e2e/test/iothub/properties/PropertiesE2ETests.cs
abhipsaMisra: it is all-or-non, none of the properties make it through

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
abhipsaMisra: don't need this

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
abhipsaMisra: done

e2e/test/Helpers/TestDeviceCallbackHandler.cs
```diff
@@ -48,33 +51,42 @@ public Message ExpectedMessageSentByService
...
-                        _logger.Trace($"{nameof(SetDeviceReceiveMethodAsync)}: Error during DeviceClient callback method: {ex}.");
+            get => Volatile.Read(ref _expectedClientPropertyValue);
+            set => Volatile.Write(ref _expectedClientPropertyValue, value);
+        }
 
-                        _methodExceptionDispatch = ExceptionDispatchInfo.Capture(ex);
-                        return Task.FromResult(new MethodResponse(500));
-                    }
-                    finally
+        public async Task SetDeviceReceiveMethodAsync(string methodName, string deviceResponseJson, string expectedServiceRequestJson)
```
abhipsaMisra, line 60: this change is only fixing the indentation

e2e/test/Helpers/TestDeviceCallbackHandler.cs
```diff
@@ -87,29 +99,32 @@ public async Task SetTwinPropertyUpdateCallbackHandlerAsync(string expectedPropN
...
-                    _logger.Trace($"{nameof(SetTwinPropertyUpdateCallbackHandlerAsync)}: DeviceClient {_testDevice.Id} callback twin: DesiredProperty: {patch}, {context}");
-
-                    try
-                    {
-                        string propertyValue = patch[expectedPropName];
-                        propertyValue.Should().Be(ExpectedTwinPropertyValue, "The property value should match what was set by service");
-                        context.Should().Be(userContext, "The context should match what was set by service");
-                    }
-                    catch (Exception ex)
+            await _deviceClient
```
abhipsaMisra, line 102: this change is only fixing the indentation

e2e/test/Helpers/TestDeviceCallbackHandler.cs
```diff
@@ -120,31 +135,33 @@ public async Task WaitForTwinCallbackAsync(CancellationToken ct)
 
         public async Task SetMessageReceiveCallbackHandlerAsync()
         {
-            await _deviceClient.SetReceiveMessageHandlerAsync(
-                async (receivedMessage, context) =>
-                {
-                    _logger.Trace($"{nameof(SetMessageReceiveCallbackHandlerAsync)}: DeviceClient {_testDevice.Id} received message with Id: {receivedMessage.MessageId}.");
-
-                    try
+            await _deviceClient
```
abhipsaMisra, line 138: this change is only fixing the indentation

e2e/test/Helpers/TestDeviceCallbackHandler.cs
```diff
@@ -158,6 +175,48 @@ public async Task WaitForReceiveMessageCallbackAsync(CancellationToken ct)
             _receiveMessageExceptionDispatch?.Throw();
         }
 
+        public async Task SetClientPropertyUpdateCallbackHandlerAsync<T>(string expectedPropName, string componentName = default)
```
abhipsaMisra, line 178: new callback handler for property update notifications

e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -0,0 +1,448 @@
...
+
+        private static readonly Dictionary<string, object> s_mapOfPropertyValues = new Dictionary<string, object>
+        {
+            { "key1", 123 },
+            { "key2", "someString" },
+            { "key3", true }
+        };
+
+        [LoggedTestMethod]
+        public async Task Properties_DeviceSetsPropertyAndGetsItBack_Mqtt()
```
vinagesh, line 36: I know we follow this style of tests in our code. But for the future something to consider for better readability of e2e tests is to add sections for `arrange` `act` `assert` and use [Theory] with multiple protocol inputs. It is very hard to understand what is being setup, what is being called and checked for like this. There might be some amount of repeated code with the arrange, act, assert format but each test will be very readable.

iothub/device/src/PayloadCollection.cs
```diff
@@ -33,6 +33,14 @@ public abstract class PayloadCollection : IEnumerable<object>
         /// <remarks>
         /// This accessor is best used to access and cast to simple types.
         /// It is recommended to use <see cref="TryGetValue"/> to deserialize to a complex type.
+        /// <para>
+        /// <remarks>
+        /// For setting component-level property values see <see cref="ClientPropertyCollection.AddComponentProperty(string, string, object)"/>
+        /// and <see cref="ClientPropertyCollection.AddComponentProperties(string, IDictionary{string, object})"/> instead.
+        /// These convenience methods ensure that component-level properties include the component identifier markers { "__t": "c" }.
+        /// For more information see <see href="https://docs.microsoft.com/en-us/azure/iot-pnp/concepts-convention#sample-multiple-components-read-only-property"/>.
```
barustum, line 41: en-us should not be included in uri

e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -0,0 +1,448 @@
...
+
+            logger.Trace($"{nameof(Properties_DeviceSetsPropertyAndGetsItBackAsync)}: name={propName}, value={propValue}");
+
+            var props = new ClientPropertyCollection();
+            props.AddRootProperty(propName, propValue);
+            await deviceClient.UpdateClientPropertiesAsync(props).ConfigureAwait(false);
+
+            // Validate the updated properties from the device-client
+            ClientProperties clientProperties = await deviceClient.GetClientPropertiesAsync().ConfigureAwait(false);
+            bool isPropertyPresent = clientProperties.TryGetValue<T>(propName, out T propFromCollection);
```
vinagesh, line 213: propValueFromCollection?

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -297,18 +296,20 @@ internal static ClientPropertyCollection FromClientTwinDictionary(IDictionary<st
         /// <param name="componentName">The component with the properties to add or update.</param>
         /// <param name="forceUpdate">Forces the collection to use the Add or Update behavior.
         /// Setting to true will simply overwrite the value. Setting to false will use <see cref="IDictionary{TKey, TValue}.Add(TKey, TValue)"/></param>
-        /// <exception cref="ArgumentNullException"><paramref name="properties"/> is <c>null</c>. </exception>
+        /// <exception cref="ArgumentNullException"><paramref name="properties"/> is <c>null</c> for a root-level property operation. </exception>
```
barustum, line 299: space after '.'

e2e/test/iothub/properties/PropertiesE2ETests.cs
abhipsaMisra: I agree. I've found unit tests to be very easy to use the arrange-act-assert format, but since end-to-end tests usually have these actions spread out within the test, I've found those comments unnecessarily cluttering up the test. We can try it out in a couple of tests and see how we feel.

abhipsaMisra: I haven't used `[Theory]` before, is that like parameterized tests?

iothub/device/src/PayloadCollection.cs
abhipsaMisra: good catch!

e2e/test/iothub/properties/PropertiesE2ETests.cs
abhipsaMisra: sure, I don't mind renaming this variable

e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -0,0 +1,448 @@
...
+                await deviceClient
+                    .UpdateClientPropertiesAsync(
+                        new ClientPropertyCollection
+                        {
+                            [propName1] = 123,
+                            [propName2] = "abcd"
+                        })
+                    .ConfigureAwait(false);
+            };
+            await func.Should().ThrowAsync<IotHubException>();
```
vinagesh, line 433: Does this throw a specific code for the exception that we can validate here based on protocol?

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
```diff
@@ -0,0 +1,480 @@
...
+                                ComponentName,
+                                new Dictionary<string, object> {
+                                    { propName1, 123 },
+                                    { propName2, "abcd" }
+                                }
+                            }
+                        })
+                    .ConfigureAwait(false);
+            };
+            await func.Should().ThrowAsync<IotHubException>();
```
vinagesh, line 466: same, any code that can be validated?

vinagesh (approved):

e2e/test/iothub/properties/PropertiesE2ETests.cs
abhipsaMisra: It does not, no. Since this is mqtt, we don't get the advantage of receiving descriptive error codes from the service.

e2e/test/iothub/properties/PropertiesE2ETests.cs
vinagesh: yupp those are parameterized tests. Theory is in xunit but it is called datarow in ms test Ex of something I added recently - https://github.com/Azure/azure-iot-sdk-csharp/blob/54e8280f775f7c6db00a29b98169588532180fc2/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs#L45

I feel writing tests using 'arrange act assert' makes it very readable and is a commonly followed format.

abhipsaMisra: Ah, I see. Parameterized MS tests don't support parallelization yet, so we have refrained from using those in our E2E tests suite since it'll end up increasing the test run time.

PR #1987, opened 2021-05-24
Updating system.text.json serializer

jamdavi:


shared/src/SystemTextJsonPayloadSerializer.cs
```diff
@@ -12,6 +14,31 @@ namespace Microsoft.Azure.Devices.Shared
     /// </summary>
     public class SystemTextJsonPayloadSerializer : PayloadSerializer
     {
+        /// <summary>
+        /// Class taken from https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#deserialize-inferred-types-to-object-properties
+        /// </summary>
+        internal class ObjectToInferredTypesConverter : JsonConverter<object>
```
jamdavi, line 20: If we need to make this visible or outside of this class I'm okay with it. However since it's purely for this implementation I thought it'd be best to just make it part of this.

shared/src/SystemTextJsonWritablePropertyResponse.cs
```diff
@@ -19,13 +19,13 @@ public sealed class SystemTextJsonWritablePropertyResponse : IWritablePropertyRe
         /// <summary>
         /// Convenience constructor for specifying the properties.
         /// </summary>
-        /// <param name="propertyValue">The unserialized property value.</param>
+        /// <param name="value">The unserialized property value.</param>
         /// <param name="ackCode">The acknowledgment code, usually an HTTP Status Code e.g. 200, 400.</param>
         /// <param name="ackVersion">The acknowledgment version, as supplied in the property update request.</param>
         /// <param name="ackDescription">The acknowledgment description, an optional, human-readable message about the result of the property update.</param>
-        public SystemTextJsonWritablePropertyResponse(object propertyValue, int ackCode, long ackVersion, string ackDescription = default)
+        public SystemTextJsonWritablePropertyResponse(object value, int ackCode, long ackVersion, string ackDescription = default)
```
jamdavi, line 26: In order for us to be able to deserialize a Writable property the constructor parameters have to match the property names.

shared/src/SystemTextJsonPayloadSerializer.cs
abhipsaMisra: I'd vote for moving this to its own class since there is quite a bit of logic in here. If we keep it separate we can add proper comments and extend it in the future, without having to worry about it crowding our PayloadSerializer implementation.

abhipsaMisra:
This will be addressed post preview.

PR #1988, opened 2021-05-25
[PnP] Add ClientPropertyCollection unit tests.

jamdavi:
Unit tests added for testing both flows with `ClientProprtyCollection`:
1. Create a `ClientPropertyCollection` and retrieve the values stored.
2. Receive a `ClientPropertyCollection` from the sdk (from service) and retrieve the values.

For testing (2) we are utilizing `Twin` to simulate service response. Once we remove `Twin` dependency, we'll rewrite these tests to use json strings instead.

### DIFF
--- a/iothub/device/tests/ClientPropertyCollectionTests.cs
+++ b/iothub/device/tests/ClientPropertyCollectionTests.cs
@@ -1,9 +1,11 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
-using Microsoft.Azure.Devices.Client;
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
 using System;
 using System.Collections.Generic;
-using System.Text;
+using FluentAssertions;
 using Microsoft.Azure.Devices.Shared;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Azure.Devices.Client.Tests
 {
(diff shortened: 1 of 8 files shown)


iothub/device/tests/ClientPropertyCollectionTests.cs
```diff
@@ -0,0 +1,140 @@
...
+                { propertyName, simplePropertyValue }
+            };
+            Assert.ThrowsException<ArgumentException>(() =>
+            {
+                clientProperties.Add(propertyName, simplePropertyValue);
+            });
+        }
+
+        [TestMethod]
+        public void ClientPropertyCollection_AddSimpleObjectWithComponentAgainThrowsException()
```
abhipsaMisra, line 86: add 2 different properties and both get added?

iothub/device/tests/ClientPropertyCollectionTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        [TestMethod]
+        public void ClientPropertyCollection_AddingComponentAddsComponentIdentifier()
+        {
+            var clientProperties = new ClientPropertyCollection
+            {
+                { componentName, propertyName, simplePropertyValue }
+            };
+            clientProperties.TryGetValue<string>(componentName, propertyName, out var outValue);
+            clientProperties.TryGetValue<string>(componentName, "__t", out var componentOut);
```
abhipsaMisra, line 134: "__t" => ConventionBasedConstants.ComponentIdentifierKey

iothub/device/tests/ClientPropertyCollectionTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            Assert.AreEqual(simplePropertyValue, outValue, $"Property values do not match, expected {simplePropertyValue} but got {outValue}");
+            Assert.AreEqual("c", componentOut, $"Property values do not match, expected c but got {componentOut}");
```
abhipsaMisra, line 137: "c" => ConventionBasedConstants.ComponentIdentifierValue

iothub/device/tests/ClientPropertyCollectionTests.cs
```diff
@@ -0,0 +1,140 @@
...
+    public class ClientPropertyCollectionTests
+    {
+        private const string propertyName = "propertyOne";
+        private const string simplePropertyValue = "propertyValue";
+        private const string componentName = "testableComponent";
+        private const string writablePropertyDescription = "testableWritablePropertyDescription";
+        private const string updatedPropertyValue = "updatedPropertyValue";
+
+        [TestMethod]
+        public void ClientPropertyCollection_CanAddSimpleObjectAndGetBackWithoutDeviceClient()
```
abhipsaMisra, line 21: similar tests for other primitive types, complex objects, arrays, nested objects etc.
We could use parameterized tests here, unit tests don't take long, so running them serially won't have much of an impact.

iothub/device/tests/ClientPropertyCollectionTestsNewtonsoft.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            var clientProperties = ClientPropertyCollection.FromTwinCollection(collectionWithComponentToRoundTrip, DefaultPayloadConvention.Instance);
+            clientProperties.TryGetValue<string>(componentName, propertyName, out var outValue);
+            Assert.AreEqual(simplePropertyValue, outValue, $"Property values do not match, expected {simplePropertyValue} but got {outValue}");
+        }
+
+        [TestMethod]
+        public void ClientPropertyCollectionNewtonsoft_CanAddSimpleWritablePropertyAndGetBackWithoutDeviceClient()
+        {
+            var clientProperties = ClientPropertyCollection.FromTwinCollection(collectionWritablePropertyToRoundTrip, DefaultPayloadConvention.Instance);
```
abhipsaMisra, line 51: this shouldn't be using twin collection, right? this is user created collection, isn't it?

iothub/device/tests/ClientPropertyCollectionTestsNewtonsoft.cs
```diff
@@ -0,0 +1,84 @@
...
+        private const string twinJsonWithComponent = "{ \"testableComponent\" : { \"__t\" : \"c\",  \"propertyOne\" : \"propertyValue\" } }";
+        private const string twinJsonWritablePropertyWithComponent = "{ \"testableComponent\" : { \"__t\" : \"c\", \"propertyOne\" : { \"value\" : \"propertyValue\",  \"ac\" : 200, \"av\" : 2, \"ad\" : \"testableWritablePropertyDescription\" } } }";
+
+        private static TwinCollection collectionToRoundTrip = new TwinCollection(twinJson);
+        private static TwinCollection collectionWritablePropertyToRoundTrip = new TwinCollection(twinJsonWritableProperty);
+        private static TwinCollection collectionWithComponentToRoundTrip = new TwinCollection(twinJsonWithComponent);
+        private static TwinCollection collectionWritablePropertyWithComponentToRoundTrip = new TwinCollection(twinJsonWritablePropertyWithComponent);
+
+        [TestMethod]
+        public void ClientPropertyCollectionNewtonsoft_CanGetSimpleValue()
```
abhipsaMisra, line 31: this should also cover the range of supported data types

jamdavi: got it so all DTDL primitives

iothub/device/tests/ClientPropertyCollectionTestsNewtonsoft.cs
jamdavi: The idea here is we're testing the flow coming off of the desired property callback for SubscribeToWritablePropertiesEventAsync. Tthis will test the deserialization of the JSON object and our ability to get the values.

We also need to test ToClientProperties to cover the MqttTransport (and others),

abhipsaMisra: I see, the test method name needs a renaming then, right? (right now it says "...WithoutDeviceClient"), whereas if we're testing the writable property callback flow then both device client and the convention would already be set up.

iothub/device/tests/ClientPropertyCollectionTests.cs
jamdavi: This is handled up a test case. We might need to check to see if we need to change how Add works for component...

iothub/device/tests/ClientPropertyCollectionTestsNewtonsoft.cs
```diff
@@ -0,0 +1,114 @@
...
+            
+            var clientProperties = ClientPropertyCollection.FromTwinCollection(collectionToRoundTrip, DefaultPayloadConvention.Instance);
+            clientProperties.Convention = DefaultPayloadConvention.Instance;
+            
+            clientProperties.TryGetValue(stringPropertyName, out string stringOutValue);
+            Assert.AreEqual(stringPropertyValue, stringOutValue, $"Property values do not match, expected {stringPropertyValue} but got {stringOutValue}");
+            clientProperties.TryGetValue(boolPropertyName, out bool boolOutValue);
+            Assert.AreEqual(boolPropertyValue, boolOutValue, $"Property values do not match, expected {boolPropertyValue} but got {boolOutValue}");
+            clientProperties.TryGetValue(doublePropertyName, out double doubleOutValue);
+            Assert.AreEqual(doublePropertyValue, doubleOutValue, $"Property values do not match, expected {doublePropertyValue} but got {doubleOutValue}");
```
jamdavi, line 51: @abhipsa, this fails when trying to convert some of the primitive types.

What happens is the float value type is deserialized from Newtonsoft as a double so the casting doesn't work as expected.

This also happens with int as the base type is Int64.

We will need to force the cast for the specified type or document the default behavior of non typed deserializaion.

jamdavi: I added the casts for some specific numeric primitives (float, int32) and the left the default ones (long, string, double) as is. This will allow us to generically handle these. DTDL will dictate that values have a certain format, such as date time, but not for others like long. It's possible we should choose to create very specific methods for **ALL** of the primitives.

| Digital twin primitive schema | Description |
| --- | --- |
|~`boolean`~| ~A boolean value~|
| `date` | A full-date as defined in [section 5.6 of RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) |
| `dateTime` | A date-time as defined in [RFC 3339](https://tools.ietf.org/html/rfc3339) |
| ~`double`~ | ~An IEEE 8-byte floating point~ |
| `duration` | A duration in ISO 8601 format |
| ~`float`~ | ~An IEEE 4-byte floating point~ |
| ~`integer`~ | ~A signed 4-byte integer~ |
| ~`long`~ | ~A signed 8-byte integer~ |
| ~`string`~ | ~A UTF8 string~ |
| `time` | A full-time as defined in [section 5.6 of RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) |


### NOTE

Consider that a customer could create their own JSON payload with decimal values stored like this `{ "myProp" : "3,14579" }`. When Newtonsoft gets the value it will treat it as a string because of the quotes. If the customer wanted to treat this as a `long` we'd eventually call `Convert.ToLong()` and it would fail with a NumberFormat failure. Or, possibly someone tries to convert `"awesomeString` to a `float`.

In both of these situations we don't handle this conversion so we'll throw an exception from the `Convert` class which is acceptable. But, we need to explore these and document the exceptions and the reasons why.

shared/src/NewtonsoftJsonPayloadSerializer.cs
```diff
@@ -39,11 +39,11 @@ public override T DeserializeToType<T>(string stringToDeserialize)
         /// <inheritdoc/>
         public override T ConvertFromObject<T>(object objectToConvert)
         {
-            if (objectToConvert == null)
-            {
-                return default;
-            }
-            return ((JToken)objectToConvert).ToObject<T>();
+            var token = JToken.FromObject(objectToConvert);
```
abhipsaMisra, line 42: Instead of casting to a `JToken`, create one instead.

shared/src/TwinCollection.cs
```diff
@@ -55,28 +55,28 @@ public TwinCollection(string twinJson, string metadataJson)
         }
 
         /// <summary>
-        /// Creates a <see cref="TwinCollection"/> using a JSON fragment as the body.
+        /// Creates a <see cref="TwinCollection"/> using the given JSON fragments for the body and metadata.
         /// </summary>
         /// <param name="twinJson">JSON fragment containing the twin data.</param>
-        internal TwinCollection(JObject twinJson)
```
abhipsaMisra, line 61: Moved internal constructor below public ones

shared/src/TwinCollection.cs
```diff
@@ -227,6 +225,19 @@ public IEnumerator GetEnumerator()
             }
         }
 
+        /// <summary>
+        /// Clear metadata out of the collection
+        /// </summary>
+        public void ClearMetadata()
```
abhipsaMisra, line 231: Moved public method above internal ones

iothub/device/tests/NumericHelpersTests.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Microsoft.Azure.Devices.Client.Tests
+{
+    [TestClass]
+    [TestCategory("Unit")]
+    public class NumericHelpersTests
+    {
+        [TestMethod]
+        public void CanConvertNumericTypes()
+        {
+            TestNumericConversion<float>(1.001d, true, 1.001f);
```
abhipsaMisra, line 19: MS test doesn't support parameterized tests with generic types, so implementing it this way

iothub/device/tests/ClientPropertyCollectionTests.cs
```diff
@@ -0,0 +1,320 @@
...
+using Microsoft.Azure.Devices.Shared;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.Client.Tests
+{
+    [TestClass]
+    [TestCategory("Unit")]
+    public class ClientPropertyCollectionTests
+    {
+        private const string BoolPropertyName = "boolProperty";
```
barustum, line 16: I see some of these string values end 'Name', but some do not.

barustum (approved):

iothub/device/tests/ClientPropertyCollectionTests.cs
abhipsaMisra: Good catch, I'll fix it in a subsequent PR

PR #1989, opened 2021-05-25
Logic with writable property request flow

abhipsaMisra:
Is this something we feel is useful for the sample?

Pro:
Demonstrate how to retrieve a previously reported writable property update value, and use that to determine if the device should respond to the update request received.

Con:
Complicated, and might mislead people into thinking that they always have to implement this kind of guard code (user scenario will usually differ case-by-case basis).

### DIFF
--- a/iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
+++ b/iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
@@ -74,13 +74,39 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 switch (writableProperty.Key)
                 {
                     case "targetTemperature":
-                        const string tagetTemperatureProperty = "targetTemperature";
+                        const string targetTemperatureProperty = "targetTemperature";
                         double targetTemperatureRequested = Convert.ToDouble(writableProperty.Value);
-                        _logger.LogDebug($"Property: Received - [ \"{tagetTemperatureProperty}\": {targetTemperatureRequested}°C ].");
+
+                        _logger.LogDebug($"Property: Received - [ \"{targetTemperatureProperty}\": {targetTemperatureRequested}°C ].");
+
+                        // Evaluate if this is a request that needs to be responded to, i.e.:
+                        // -> The twin's last reported "targetTemperature" value is not the same as the requested value.
+                        // -> The twin's last reported "targetTemperature" version is less than the requested property version.
+
+                        // Retrieve the device's properties.
+                        ClientProperties properties = await _deviceClient.GetClientPropertiesAsync();
+
+                        bool wasTargetTemperatureReported = properties.TryGetValue(targetTemperatureProperty, out NewtonsoftJsonWritablePropertyResponse targetTemperatureWritableResponse);
+
+                        if (wasTargetTemperatureReported)
+                        {
+                            double targetTemperatureReported = Convert.ToDouble(targetTemperatureWritableResponse.Value);
+
+                            // If the last reported "targetTemperature" value is already at the requested value
+                            // and with a version greater than or equal to the requested version, then the device has already responded to this request.
+                            if (targetTemperatureReported == targetTemperatureRequested
+                                && targetTemperatureWritableResponse.AckVersion >= writableProperties.Version)
+                            {
+                                _logger.LogDebug($"Property: Update - {targetTemperatureProperty} is already at {targetTemperatureRequested} " +
+                                $"with a version of {targetTemperatureWritableResponse.AckVersion}. The update request with a version of {writableProperties.Version} " +
+                                $"has been discarded.");
+                            }
+                            break;
+                        }
 
                         _temperature = targetTemperatureRequested;
                         var reportedProperty = new ClientPropertyCollection();
-                        reportedProperty.Add(tagetTemperatureProperty, _temperature, StatusCodes.OK, writableProperties.Version, "Successfully updated target temperature");
+                        reportedProperty.Add(targetTemperatureProperty, _temperature, StatusCodes.OK, writableProperties.Version, "Successfully updated target temperature");
 
                         ClientPropertiesUpdateResponse updateResponse = await _deviceClient.UpdateClientPropertiesAsync(reportedProperty);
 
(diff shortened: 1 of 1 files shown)


vinagesh:
I feel like the Thermostat sample is better off being simple and bare bones. I would prefer adding a new sample for advanced scenarios and name it appropriately. However, would be good to hear more opinions :)

abhipsaMisra:
> I feel like the Thermostat sample is better off being simple and bare bones. I would prefer adding a new sample for advanced scenarios and name it appropriately. However, would be good to hear more opinions :)

Even for an advanced scenario specific sample, do you think something like this would be useful, or would it confuse the customers.

For me, the biggest value that this can lend is guide customers how to retrieve a previously reported writable property value, simply because we don't mandate they send in a strongly typed object (we can format the response for them)..
i.e. `bool wasTargetTemperatureReported = properties.TryGetValue(targetTemperatureProperty, out NewtonsoftJsonWritablePropertyResponse targetTemperatureWritableResponse);`

It might not be intuitive for them to figure out that the `out` param needs to be `NewtonsoftJsonWritablePropertyResponse` (or a similar `IWritablePropertyResponse` implementation that they have provided).

iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
```diff
@@ -74,13 +74,40 @@ private async Task HandlePropertyUpdatesAsync(ClientPropertyCollection writableP
...
-                        _logger.LogDebug($"Property: Received - [ \"{tagetTemperatureProperty}\": {targetTemperatureRequested}°C ].");
+
+                        _logger.LogDebug($"Property: Received - [ \"{targetTemperatureProperty}\": {targetTemperatureRequested}°C ].");
+
+                        // Evaluate if this is a request that needs to be responded to, i.e.:
+                        // -> The twin's last reported "targetTemperature" value is not the same as the requested value.
+                        // -> The twin's last reported "targetTemperature" version is less than the requested property version.
+
+                        // Retrieve the device's properties.
+                        ClientProperties properties = await _deviceClient.GetClientPropertiesAsync();
```
drwill-ms, line 87: I'd rather see the device get properties when it connects, and hold them in memory. Then it can compare its latest value (and version) to requests that come in.

vinagesh:
> > I feel like the Thermostat sample is better off being simple and bare bones. I would prefer adding a new sample for advanced scenarios and name it appropriately. However, would be good to hear more opinions :)
> 
> Even for an advanced scenario specific sample, do you think something like this would be useful, or would it confuse the customers.
> 
> For me, the biggest value that this can lend is guide customers how to retrieve a previously reported writable property value, simply because we don't mandate they send in a strongly typed object when sending the property update (we can format the response for them).
> i.e. `bool wasTargetTemperatureReported = properties.TryGetValue(targetTemperatureProperty, out NewtonsoftJsonWritablePropertyResponse targetTemperatureWritableResponse);`
> 
> It might not be intuitive for them to figure out that the `out` param needs to be `NewtonsoftJsonWritablePropertyResponse` (or a similar `IWritablePropertyResponse` implementation that they have provided).

Good point! As I've not really tried using it, I had to go back and look at all of that. There seems value in showing the out param type. However, it brought me to another question. 

Why not have ?
public void Add<T>(string propertyName, T writableProperty) where T : IWritablePropertyResponse

This sort of tells user they should use NewtonsoftJsonWritablePropertyResponse or their own implementation of IWritablePropertyResponse for writable properties.

iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
abhipsaMisra: Interesting; since the twin value will keep on changing, are you suggesting that we prepend each twin operation with a call to GetClientProperties?

abhipsaMisra:
>Why not have ?
>public void Add(string propertyName, T writableProperty) where T : IWritablePropertyResponse
>
>This sort of tells user they should use NewtonsoftJsonWritablePropertyResponse or their own implementation of IWritablePropertyResponse for writable properties.

In one of the earlier designs we did have this API, but the feedback was that it wasn't that intuitive since customers using the default convention wouldn't necessarily be knowing about the existence of `NewtonsoftJsonWritablePropertyResponse`. 
We have a task to support `TryGetValue(string componentName, string propertyName, IWritablePropertyResponse value)` as well, so that in the return flow they don't need to necessarily know about the writable response type either; but we don't have this in place yet. (The plan is to add this post the preview release - as a patch update).

So this code in the sample could be helpful until then.

iothub/device/samples/convention-based-samples/Thermostat/ThermostatSample.cs
abhipsaMisra: Fetching the twin on initial startup is something that docs talk about as well, so it would be good for us to add it into our sample.  I wonder if there is any value in fetching it throughout the lifetime of the sample, since I worry that it might misguide customers into thinking that this is something that they have to do always (extra API call = more bandwidth consumed + more expensive)

drwill-ms: > since the twin value will keep on changing, are you suggesting that we prepend each twin operation with a call to GetClientProperties?

No. The desired value will change and the client (while connected) will be notified. That means it only needs to know what the latest request was when it connects. When it first connects it should find out what the latest reported and desired values are. It'll initialize it's local, in-memory copy to the latest reported value (and store the version for later version checking), evaluate any new desired values, and then wait for future desired values.

drwill-ms: I don't see any value in getting the latest reported value during a property update callback.

vinagesh:
> > Why not have ?
> > public void Add(string propertyName, T writableProperty) where T : IWritablePropertyResponse
> > This sort of tells user they should use NewtonsoftJsonWritablePropertyResponse or their own implementation of IWritablePropertyResponse for writable properties.
> 
> In one of the earlier designs we did have this API, but the feedback was that it wasn't that intuitive since customers using the default convention wouldn't necessarily be knowing about the existence of `NewtonsoftJsonWritablePropertyResponse`.
> We have a task to support `TryGetValue(string componentName, string propertyName, IWritablePropertyResponse value)` as well, so that in the return flow they don't need to necessarily know about the writable response type either; but we don't have this in place yet. (The plan is to add this post the preview release - as a patch update).
> 
> So this code in the sample could be helpful until then.
TryGetValue(string componentName, string propertyName, IWritablePropertyResponse value) will still mean they have to know which implementation to use right? If you have that, a counter part in the Add will also be good.  Add(string propertyName, IWritablePropertyResponse  writableProperty)

Being really new to this, this would have been more clear to me than anything else. Wondering why there was pushback on that design. It is confusing to get the property back in a different format than the Add.

abhipsaMisra:
Closing this PR, I'll open a separate PR for sample updates.

PR #1992, opened 2021-05-25
Update API design doc and move SystemTextJson helper to samples

abhipsaMisra:


iothub/device/devdoc/Convention-based operations.md
```diff
@@ -109,20 +124,25 @@ public Task SubscribeToWritablePropertiesEventAsync(Func<ClientPropertyCollectio
...
 public class ClientProperties : ClientPropertyCollection {
+    public ClientProperties();
     public ClientPropertyCollection Writable { get; private set; }
 }
 
 public class ClientPropertyCollection : PayloadCollection {
     public ClientPropertyCollection();
     public long Version { get; protected set; }
-    public void Add(IDictionary<string, object> properties, string componentName = null);
+    public void Add(IDictionary<string, object> properties);
```
vinagesh, line 134: Is this new change to make it similar to other languages? I thought in C# we try not to have too many overloads.

vinagesh (approved):

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: No, the idea here was to keep the component name as the first param, for all APIs that do accept a component name. 
This will ensure that we have the same experience throughout; eg. if we do `TryGetValue(string componentName, string propertyName, out T value)` , then we should be able to do `Add(string componentName, string propertyName, object value)`.

The other alternative was to put componentName as the last param for all APIs (since it is optional), but calling this did not feel intuitive:
```csharp
var value = properties.TryGetValue("property1", out T properytValue, "component1");
```

abhipsaMisra: This does bring in additional overloads, which we do try to avoid; however, in this case adding optional params gave rise to a non-intuitive API signature.

drwill-ms: It might be that we have competing priorities, in which case we should discuss which way we'd rather go. Either component name is always first, or it is optional and at the end.

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -2,6 +2,12 @@
 
 #### Common
 
+```diff
+public class ClientOptions {
++    public PayloadConvention PayloadConvention { get; set; }
```
drwill-ms, line 7: Shouldn't this be readonly?

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -69,13 +75,22 @@ public abstract class PayloadCollection : IEnumerable, IEnumerable<object> {
...
+    public const char ComponentLevelCommandSeparator = '*';
     public const string AckCodePropertyName = "ac";
     public const string AckDescriptionPropertyName = "ad";
     public const string AckVersionPropertyName = "av";
     public const string ComponentIdentifierKey = "__t";
     public const string ComponentIdentifierValue = "c";
     public const string ValuePropertyName = "value";
 }
+
+public class StatusCodes {
```
drwill-ms, line 87: You'll probably want a more unique name for this. How about `IotStatusCodes`? Or maybe `DeviceStatusCodes`?

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: Good point, yeah, should be. I'll add this to a constructor in a follow-up PR.

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: I like `IotStatusCodes`, only thing is, would people think that this is a status code that IoT Hub sends?
Since these status codes are sent by the client, how about `ClientStatusCodes`?

iothub/device/devdoc/Convention-based operations.md
drwill-ms: Readonly doesn't mean it has to be a a ctor. It simply must be set on initialization, and can't be changed after.

`public PayloadConvention PayloadConvention { get; }`

iothub/device/devdoc/Convention-based operations.md
drwill-ms: `CommonClientStatusCodes` ?

iothub/device/devdoc/Convention-based operations.md
drwill-ms: Any thoughts about this, Abhipsa?

abhipsaMisra: Good point, let us discuss this during PL tomorrow.

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: But the user needs to be able to set it on initialization =>so it needs to be initialized via a constructor. Is there a different way to do this?

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: I wonder if `CommonClientStatusCodes` would make people think that these are common status codes that represent the status of a client? Should we call it `CommonClientResponseCodes`?

PR #1993, opened 2021-05-26
Updates tests to implement/ supress warnings for IDisposable resource disposal

abhipsaMisra:
These warnings were highlighted in the `master` -> `preview` PR. 
This is because `preview` has .NET 5.0 support, which enforces disposal of `IDisposable` types very strictly.

### DIFF
--- a/e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
+++ b/e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
@@ -71,7 +71,8 @@ namespace Microsoft.Azure.Devices.E2ETests
 
             // Create an instance of the device client, send a test message and then close and dispose it.
             DeviceClient deviceClient = DeviceClient.Create(testDevice.IoTHubHostName, authenticationMethod, transport);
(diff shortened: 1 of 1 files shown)


timtay-microsoft (approved):

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
```diff
@@ -71,15 +71,17 @@ private async Task ReuseAuthenticationMethod_SingleDevice(Client.TransportType t
 
             // Create an instance of the device client, send a test message and then close and dispose it.
             DeviceClient deviceClient = DeviceClient.Create(testDevice.IoTHubHostName, authenticationMethod, transport);
-            await deviceClient.SendEventAsync(new Client.Message()).ConfigureAwait(false);
+            using var message1 = new Client.Message();
```
drwill-ms, line 74: I wonder if `SendEventAsync` fails to dispose on exception. I'd hope so, because then the user will want to try to resend it. That means it should be conditionally disposed by the creator.

PR #2000, opened 2021-06-03
[PnP] Separate out root-level and component-level property addition operations

abhipsaMisra:
This PR separates out the root-level and component-level property operations.
This is to resolve the ambiguity between adding a root-level property that is of map schema vs adding a collection of component-level properties.

Eg:

Model:
```json
{
	"@id": "dtmi:com:example:NestedInterface;1",
	"@type": "Interface",
	"displayName": "Sample Model",
	"contents": [
		{
			"@type": "Component",
			"displayName": {
				"en": "myComponent"
			},
			"name": "myComponent",
			"schema": "dtmi:com:example:RootInterface;1"
		}
	],
	"@context": "dtmi:dtdl:context;2"
}
```

```json
{
	"@id": "dtmi:com:example:RootInterface;1",
	"@type": "Interface",
	"displayName": "Sample Model",
	"contents": [
		{
			"@type": "Property",
			"name": "modules",
			"writable": true,
			"schema": {
				"@type": "Map",
				"mapKey": {
					"name": "moduleName",
					"schema": "string"
				},
				"mapValue": {
					"name": "moduleState",
					"schema": "string"
				}
			}
		},
		{
			"@type": "Property",
			"name": "manufacturer",
			"displayName": "Manufacturer",
			"schema": "string",
			"description": "Company name of the device manufacturer. This could be the same as the name of the original equipment manufacturer (OEM). Ex. Contoso."
		},
		{
			"@type": "Property",
			"name": "model",
			"displayName": "Device model",
			"schema": "string",
			"description": "Device model name or ID. Ex. Surface Book 2."
		}
	],
	"@context": "dtmi:dtdl:context;2"
}
```

(1) Device registers with no-component model `"dtmi:com:example:RootInterface;1"` and creates a `ClientPropertyCollection` to update a root-level property `modules` that is of map schema:
```csharp
var modulesMap = new Dictionary<string, string>
{
    { "myModuleA", "running" },
    { "myModuleB", "stopped" }
};

var rootProperty = new ClientPropertyCollection();
rootProperty.Add("modules", modulesMap);
```

(2) Device registers with nested component model `"dtmi:com:example:NestedInterface;1"` and creates a `ClientPropertyCollection` to update a collection of component-level properties:
```csharp
var allProperties = new Dictionary<string, object>
{
    { "manufacturer", "Intel" },
    { "model", "DFC5646" },
    { "modules", modulesMap }
};

var componentProperties = new ClientPropertyCollection();
componentProperties.Add("myComponent", allProperties);
```

At this point, the sdk cannot differentiate between (1) and (2), since both have the same method signature; and as a result, cannot correctly identify the component-level property addition in order to add the component identifiers: `{ "__t" : "c" }`.

For this reason, we are splitting the `Add` APIs into separate APIs for root-level operations and component-level operations.

### DIFF
--- a/iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
+++ b/iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
@@ -129,8 +129,13 @@ namespace Microsoft.Azure.Devices.Client.Samples
             _logger.LogDebug($"Property: Received - component=\"{componentName}\", [ \"{targetTemperatureProperty}\": {targetTemperature}°C ].");
 
             _temperature[componentName] = targetTemperature;
+            IWritablePropertyResponse writableResponse = _deviceClient
+                .PayloadConvention
+                .PayloadSerializer
+                .CreateWritablePropertyResponse(_temperature[componentName], StatusCodes.OK, version, "Successfully updated target temperature.");
+
             var reportedProperty = new ClientPropertyCollection();
-            reportedProperty.Add(componentName, targetTemperatureProperty, _temperature[componentName], StatusCodes.Accepted, version, "Successfully updated target temperature.");
+            reportedProperty.AddComponentProperty(componentName, targetTemperatureProperty, writableResponse);
 
             ClientPropertiesUpdateResponse updateResponse = await _deviceClient.UpdateClientPropertiesAsync(reportedProperty);
 
@@ -288,7 +293,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 { "totalMemory", 1024 },
             };
             var deviceInformation = new ClientPropertyCollection();
-            deviceInformation.Add(componentName, deviceInformationProperties);
+            deviceInformation.AddComponentProperties(componentName, deviceInformationProperties);
 
             ClientPropertiesUpdateResponse updateResponse = await _deviceClient.UpdateClientPropertiesAsync(deviceInformation, cancellationToken);
 
@@ -330,7 +335,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
                 || serialNumberReported != currentSerialNumber)
             {
                 var reportedProperties = new ClientPropertyCollection();
-                reportedProperties.Add(serialNumber, currentSerialNumber);
+                reportedProperties.AddRootProperty(serialNumber, currentSerialNumber);
 
                 ClientPropertiesUpdateResponse updateResponse = await _deviceClient.UpdateClientPropertiesAsync(reportedProperties, cancellationToken);
 
@@ -391,7 +396,7 @@ namespace Microsoft.Azure.Devices.Client.Samples
             const string propertyName = "maxTempSinceLastReboot";
             double maxTemp = _maxTemp[componentName];
             var reportedProperties = new ClientPropertyCollection();
-            reportedProperties.Add(componentName, propertyName, maxTemp);
+            reportedProperties.AddComponentProperty(componentName, propertyName, maxTemp);
 
             ClientPropertiesUpdateResponse updateResponse = await _deviceClient.UpdateClientPropertiesAsync(reportedProperties, cancellationToken);
 
--- a/iothub/device/src/ClientPropertyCollection.cs
+++ b/iothub/device/src/ClientPropertyCollection.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Azure.Devices.Shared;
 
 namespace Microsoft.Azure.Devices.Client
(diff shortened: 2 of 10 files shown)


iothub/device/samples/convention-based-samples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -129,8 +129,13 @@ private async Task HandleTargetTemperatureUpdateRequestAsync(string componentNam
...
 
             _temperature[componentName] = targetTemperature;
+            IWritablePropertyResponse writableResponse = _deviceClient
+                .PayloadConvention
+                .PayloadSerializer
+                .CreateWritablePropertyResponse(_temperature[componentName], StatusCodes.OK, version, "Successfully updated target temperature.");
+
             var reportedProperty = new ClientPropertyCollection();
-            reportedProperty.Add(componentName, targetTemperatureProperty, _temperature[componentName], StatusCodes.Accepted, version, "Successfully updated target temperature.");
+            reportedProperty.AddComponentProperty(componentName, targetTemperatureProperty, writableResponse);
```
abhipsaMisra, line 138: The user is recommended to create an instance of `IWritableProperty` response (we've added this to doc comments). This will also tie well with how they retrieve this value using the `TryGetValue` APIs, as this will give then the entire reponse object directly, instead of giving them the property value, ack code, ack version and ack description as separate values.

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -14,126 +15,104 @@ public class ClientPropertyCollection : PayloadCollection
...
         /// </para>
         /// </remarks>
-        /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <c>null</c> </exception>
         /// <param name="propertyName">The name of the property to add.</param>
         /// <param name="propertyValue">The value of the property to add.</param>
-        public override void Add(string propertyName, object propertyValue)
-            => Add(null, propertyName, propertyValue);
+        /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <c>null</c>. </exception>
+        /// <exception cref="ArgumentException"><paramref name="propertyName"/> already exists in the collection.</exception>
+        public void AddRootProperty(string propertyName, object propertyValue)
```
abhipsaMisra, line 33: This diff gets a little messy, I'd recommend reviewing the updated file by itself and verifying that the APIs make sense.

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -14,126 +15,104 @@ public class ClientPropertyCollection : PayloadCollection
...
-        /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <c>null</c> </exception>
-        /// <param name="propertyName">The name of the property to add or update.</param>
-        /// <param name="propertyValue">The value of the property to add or update.</param>
-        /// <param name="statusCode"></param>
-        /// <param name="version"></param>
-        /// <param name="description"></param>
-        /// <param name="componentName"></param>
-        public void Add(string componentName, string propertyName, object propertyValue, int statusCode, long version, string description = default)
+        /// <param name="properties">A collection of properties to add.</param>
+        public void Add(IDictionary<string, object> properties)
```
abhipsaMisra, line 82: I want to highlight this API that allows the user to add a dictionary directly to the collection.
For component-level properties, we recommend using the `AddComponentProperties(string, IDictionary{string, object})` convenience method instead, so that they don't need to ensure that the component-identifiers "__t": "c" are a part of the collection.

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -14,6 +15,11 @@ namespace Microsoft.Azure.Devices.Client
     /// <threadsafety static="true" instance="true" />
     public partial class DeviceClient : IDisposable
     {
+        /// <summary>
+        /// The <see cref="PayloadConvention"/> that the client uses for convention-based operations.
+        /// </summary>
+        public PayloadConvention PayloadConvention => InternalClient.PayloadConvention;
```
abhipsaMisra, line 21: This is exposed on the device client to help users build their writable property response.

iothub/device/src/NumericHelpers.cs
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Globalization;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    internal class NumericHelpers
+    {
+        internal static bool TryCastNumericTo<T>(object input, out T result)
```
abhipsaMisra, line 11: NewtonSoft will deserialize a float into a double (long to int, etc), at which point the `TryGetValue` API with the higher range value type will fail. This helper is added to handle those numeric conversions.

iothub/device/src/PayloadCollection.cs
```diff
@@ -133,6 +149,14 @@ public virtual string GetSerializedString()
             return Convention.PayloadSerializer.SerializeToString(Collection);
         }
 
+        /// <summary>
+        /// Remove all items from the collection.
+        /// </summary>
+        public void ClearCollection()
```
abhipsaMisra, line 155: Once a value is added to a `PayloadCollection`, it can only be updated but not removed. This is because property collection needs to be able to support `null` values for properties, so that service knows that those properties need to be removed: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins#device-operations
For this reason we're adding a `Clear` method, that will allow the users to clear an entire collection and start afresh.

abhipsaMisra: If customers ask for "remove" functionality, we can add `RemoveRootProperty(string propertyName)`, `RemoveComponentProperty(string componentName, string propertyName)` and `RemoveComponent(string componentName)` methods.

iothub/device/src/NumericHelpers.cs
```diff
@@ -0,0 +1,43 @@
...
+                }
+            }
+
+            result = default;
+            return false;
+        }
+
+        private static bool TryGetNumeric(object expression)
+        {
+            if (expression == null)
```
drwill-ms, line 31: I know we agreed to bypass braces for logging statements, but not for any other cases.

drwill-ms (approved):

PR #2001, opened 2021-06-04
[PnP] Add fault injection tests for properties operations

abhipsaMisra:
The properties APIs are the only APIs that have new transport layer counterparts added, so I am adding fault injection tests to ensure that subscriptions are successfully recovered on reconnection.

By contrast, the telemetry and command APIs reuse the existing "events" and "direct method" APIs at the transport level.

### DIFF
--- a/e2e/test/iothub/properties/PropertiesFaultInjectionTests.cs
+++ b/e2e/test/iothub/properties/PropertiesFaultInjectionTests.cs
@@ -1,0 +1,223 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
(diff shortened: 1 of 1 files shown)


timtay-microsoft:
> The properties APIs are the only APIs that have new transport layer counterparts added

Can you expand on this a bit? I was under the impression that the SDK was still using the same basic reported properties/desired properties flows just with payloads enriched with component/digital twin details. If we are using the same twin links in the same way, I'm not sure I follow why we'd need to add these tests

abhipsaMisra:
> > The properties APIs are the only APIs that have new transport layer counterparts added
> 
> Can you expand on this a bit? I was under the impression that the SDK was still using the same basic reported properties/desired properties flows just with payloads enriched with component/digital twin details. If we are using the same twin links in the same way, I'm not sure I follow why we'd need to add these tests

For properties we aren't actually reusing the same desired/reported property flow, we've added new methods to operate on the "twin" topics: https://github.com/Azure/azure-iot-sdk-csharp/blob/3c14bcdfb8009bac029a90cbf14b970638754bc7/iothub/device/src/IDelegatingHandler.cs#L70-L74

(Technically we could have done without the "desired" property fault injection tests, but I still added them. It'll benefit us when we remove the twin dependency entirely).

timtay-microsoft (approved):

vinagesh (approved):

PR #2003, opened 2021-06-04
Update AuthenticationWithTokenRefresh disposal fix to be non-breaking

abhipsaMisra:
Update `AuthenticationWithTokenRefresh` to be disposed by the sdk, by default, but make the setting configurable.

Linked: #1954, #1911

### DIFF
--- a/e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
+++ b/e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Azure.Devices.E2ETests
         private async Task AuthenticationMethodDisposesTokenRefresher(Client.TransportType transport)
         {
             TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix).ConfigureAwait(false);
-            var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString, disposalBySdk: true);
+            var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString, disposeWithClient: true);
 
             // Create an instance of the device client, send a test message and then close and dispose it.
             DeviceClient deviceClient = DeviceClient.Create(testDevice.IoTHubHostName, authenticationMethod, transport);
@@ -125,7 +125,7 @@ namespace Microsoft.Azure.Devices.E2ETests
         private async Task ReuseAuthenticationMethod_SingleDevice(Client.TransportType transport)
         {
             TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix).ConfigureAwait(false);
-            var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString, disposalBySdk: false);
+            var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString, disposeWithClient: false);
 
             // Create an instance of the device client, send a test message and then close and dispose it.
             DeviceClient deviceClient = DeviceClient.Create(testDevice.IoTHubHostName, authenticationMethod, transport);
@@ -168,7 +168,7 @@ namespace Microsoft.Azure.Devices.E2ETests
             {
                 TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix).ConfigureAwait(false);
 #pragma warning disable CA2000 // Dispose objects before losing scope - the authentication method is disposed at the end of the test.
-                var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString, disposalBySdk: false);
+                var authenticationMethod = new DeviceAuthenticationSasToken(testDevice.ConnectionString, disposeWithClient: false);
 #pragma warning restore CA2000 // Dispose objects before losing scope
 
                 testDevices.Add(testDevice);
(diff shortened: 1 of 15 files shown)


e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
```diff
@@ -55,19 +55,77 @@ public async Task DeviceSak_ReusableAuthenticationMethod_SingleDevicePerConnecti
...
+        public async Task DeviceClient_AuthenticationMethodDisposesTokenRefresher_AmqpWs()
+        {
+            await AuthenticationMethodDisposesTokenRefresher(Client.TransportType.Amqp_WebSocket_Only).ConfigureAwait(false);
+        }
+
+        // Even on encountering an exception, the MQTT layer keeps on reattempting CONNECT when communicating via DotNetty's TCP stack.
+        // As a result, instead of throwing the actual exception encountered an IotHubCommunicationException is thrown (on operation timeout).
+        // This is not an issue when the communication is over websockets (where we use the sdk's websocket implementation).
+        // This test has been ignored until we root-cause the issue on DotNetty's TCP stack.
+        [Ignore]
```
abhipsaMisra, line 89: I just want to call out that mqtt going into endless reconnects isn't a behavior change, this has always been the case.
Previously the sdk would keep on hitting an NRE but still keep retrying. Now it'll hit an ObjectDisposedException instead, but still keep retrying. Based on my investigation this seems to be stemming from the tcp stack implementation used in DotNetty, but I don't have the exact root cause yet.
I am ignoring this test so that we can get this change checked in, while continuing investigation on this issue.

abhipsaMisra, line 89: I want to call out that mqtt going into endless reconnects isn't a behavior change, this has always been the case.
Previously the sdk would keep on hitting an NRE but still keep retrying. Now it'll hit an ObjectDisposedException instead, but still keep retrying. Based on my investigation this seems to be stemming from the tcp stack implementation used in DotNetty, but I don't have the exact root cause yet.
I am ignoring this test so that we can get this change checked in, while continuing investigation on this issue.

iothub/device/src/AuthenticationWithTokenRefresh.cs
```diff
@@ -65,21 +88,26 @@ public abstract class AuthenticationWithTokenRefresh : IAuthenticationMethod, ID
...
-        internal bool InstanceCreatedBySdk { get; set; }
+        // This internal property is used by the sdk to determine if the disposal should be handled by the sdk.
+        internal bool DisposalBySdk { get; }
 
         /// <summary>
         /// Gets a snapshot of the security token associated with the device. This call is thread-safe.
         /// </summary>
         public async Task<string> GetTokenAsync(string iotHub)
         {
+            if (_isDisposed)
```
abhipsaMisra, line 99: I made this fail-fast, since using a disposed authentication method might introduce some other issues down the line.

iothub/device/src/HsmAuthentication/ModuleAuthenticationWithHsm.cs
```diff
@@ -35,7 +35,9 @@ internal ModuleAuthenticationWithHsm(ISignatureProvider signatureProvider, strin
             string moduleId,
             string generationId,
             TimeSpan sasTokenTimeToLive,
-            int sasTokenRenewalBuffer) : base(deviceId, moduleId, (int)sasTokenTimeToLive.TotalSeconds, sasTokenRenewalBuffer)
+            int sasTokenRenewalBuffer,
+            bool disposalBySdk)
+            : base(deviceId, moduleId, (int)sasTokenTimeToLive.TotalSeconds, sasTokenRenewalBuffer, disposalBySdk)
```
abhipsaMisra, line 40: This is an internal class, so I directly added the new param to the existing constructor.

iothub/device/src/AuthenticationWithTokenRefresh.cs
```diff
@@ -25,13 +25,34 @@ public abstract class AuthenticationWithTokenRefresh : IAuthenticationMethod, ID
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AuthenticationWithTokenRefresh"/> class.
+        /// </summary>
+        /// <param name="suggestedTimeToLiveSeconds">Token time to live suggested value. The implementations of this abstract
+        /// may choose to ignore this value.</param>
+        /// <param name="timeBufferPercentage">Time buffer before expiry when the token should be renewed expressed as
+        /// a percentage of the time to live.</param>
+        /// <param name="disposalBySdk">True if the authentication method should be disposed of by sdk; false if you intend to reuse the authentication method.</param>
```
timtay-microsoft, line 51: "... should be disposed of by the client when the client using this instance is itself disposed"?

iothub/device/src/AuthenticationWithTokenRefresh.cs
```diff
@@ -25,13 +25,34 @@ public abstract class AuthenticationWithTokenRefresh : IAuthenticationMethod, ID
...
+        public AuthenticationWithTokenRefresh(
+            int suggestedTimeToLiveSeconds,
+            int timeBufferPercentage,
+            bool disposalBySdk)
```
timtay-microsoft, line 55: Maybe ```disposeWithClient``` to highlight that this object is disposed when the client using it is disposed? "SDK" feels like the wrong term to use here

timtay-microsoft (approved):

PR #2004, opened 2021-06-04
[PnP] Renaming StatusCodes and add a comment to highlight ClientOptions behavior

abhipsaMisra:
- Addresses comments from #1992 :
  - `ClientOptions.PayloadConvention` readonly => making it readonly would require the user to initialize it via a constructor. Technically all properties inside `ClientOptions` can be updated as many times as the caller wants, but only until it is actually passed to a DeviceClient/ ModuleClient instance during initialization. For this reason, instead of making the properties read-only, I added a code comment to the class that highlights the actual behavior.
  - Rename `StatusCodes` => since these are response codes that a client would send to service, either through a command response or a writable property response, I felt `CommonClientResponseCodes` represented the purpose of this class.

### DIFF
--- a/iothub/device/devdoc/Convention-based operations.md
+++ b/iothub/device/devdoc/Convention-based operations.md
@@ -6,6 +6,14 @@
 public class ClientOptions {
 +    public PayloadConvention PayloadConvention { get; set; }
 }
+
+public class DeviceClient : IDisposable {
++    public PayloadConvention PayloadConvention { get; }
+}
+
+public class ModuleClient : IDisposable {
++    public PayloadConvention PayloadConvention { get; }
+}
 ```
 
 ```csharp
@@ -85,12 +93,12 @@ public static class ConventionBasedConstants {
     public const string ValuePropertyName = "value";
 }
 
-public class StatusCodes {
-    public StatusCodes();
-    public static int Accepted { get; }
-    public static int BadRequest { get; }
-    public static int NotFound { get; }
-    public static int OK { get; }
+public class CommonClientResponseCodes {
+    public const int Accepted = 202;
+    public const int BadRequest = 400;
+    public const int NotFound = 404;
+    public const int OK = 200;
+    public CommonClientResponseCodes();
 }
 ```
 
--- a/iothub/device/src/ClientOptions.cs
+++ b/iothub/device/src/ClientOptions.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// Options that allow configuration of the device or module client instance during initialization.
-    /// These options cannot be updated once the client has been initialized.
+    /// Updating these options after the client has been initialized will not change the behavior of the client.
     /// </summary>
     public class ClientOptions
     {
--- a/iothub/device/src/ClientOptions.cs
+++ b/iothub/device/src/ClientOptions.cs
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// Options that allow configuration of the device or module client instance during initialization.
+    /// These options cannot be updated once the client has been initialized.
     /// </summary>
     public class ClientOptions
     {
--- a/shared/src/CommonClientResponseCodes.cs
+++ b/shared/src/CommonClientResponseCodes.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics.CodeAnalysis;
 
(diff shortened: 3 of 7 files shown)


iothub/device/src/ClientOptions.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// Options that allow configuration of the device or module client instance during initialization.
+    /// These options cannot be updated once the client has been initialized.
```
vinagesh, line 11: cannot be or should not be updated? Do you intend to say "Updating these options after the client has been initialized will not change the behavior of the client?"

abhipsaMisra: I like this - Updating these options after the client has been initialized will not change the behavior of the client

shared/src/CommonClientResponseCodes.cs
```diff
@@ -10,21 +11,25 @@ namespace Microsoft.Azure.Devices.Shared
...
         /// </summary>
         public static int OK => 200;
+
         /// <summary>
         /// Status code 202.
         /// </summary>
         public static int Accepted => 202;
+
         /// <summary>
         /// Status code 400.
```
vinagesh, line 29: Maybe add more details to the doc comments? Like 400 means client error, 202 means request has been accepted for processing but the processing has not been completed, 200 means request has succeeded?

abhipsaMisra: It doesn't need to necessarily need to mean that though. These code aren't being referenced directly within the sdk to represent a certain state, rather they are meant to be used by the users to represent the response of a command or a property update request. Their application logic would dictate what these codes mean for them => 202 for them could mean "this request completed successfully". It would not be 100% as per API specs, but they are free to make any choice they want.

How about we say: "As per IANA (Internet Assigned Numbers Authority) this code represents ..." instead, to help nudge them in the right direction?

shared/src/CommonClientResponseCodes.cs
```diff
@@ -10,21 +11,25 @@ namespace Microsoft.Azure.Devices.Shared
...
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1052:Static holder types should be Static or NotInheritable", Justification = "To allow customers to extend this class we need to not mark it static.")]
-    public class StatusCodes
+    [SuppressMessage("Design", "CA1052:Static holder types should be Static or NotInheritable",
+        Justification = "To allow customers to extend this class we need to not mark it static.")]
+    public class CommonClientResponseCodes
     {
         /// <summary>
         /// Status code 200.
         /// </summary>
         public static int OK => 200;
```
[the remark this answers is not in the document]
abhipsaMisra: could be a const field, instead of a static property, sure.

shared/src/CommonClientResponseCodes.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Microsoft.Azure.Devices.Shared
+{
+    /// <summary>
+    /// A list of common status codes to represent the response from the client.
+    /// </summary>
+    /// <remarks>
+    /// These status codes are based on the HTTP status codes listed here <see href="http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml"/>.
+    /// </remarks>
+    [SuppressMessage("Design", "CA1052:Static holder types should be Static or NotInheritable",
```
drwill-ms, line 14: Seems low value to me. Did you consult with Azure SDK team?

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -6,6 +6,14 @@
 public class ClientOptions {
 +    public PayloadConvention PayloadConvention { get; set; }
```
drwill-ms, line 7: probably should be readonly too

shared/src/CommonClientResponseCodes.cs
abhipsaMisra: I haven't, I'll ask if they have a recommendation.

iothub/device/devdoc/Convention-based operations.md
abhipsaMisra: So, in reality, you can modify these options as many times as you want until you actually pass it to the client, so I don't think a setter is entirely misleading here. 
I added the following text to the doc comments instead, to highlight this behavior:
"Updating these options after the client has been initialized will not change the behavior of the client."

What do you think?

drwill-ms: It also means they could pass it to the client, and change it after. We're both looking at the same instance.

abhipsaMisra: Synced up offline, I'll make the change.

PR #2040, opened 2021-06-18
feat(service): add support for DeviceScope to import/export job

drwill-ms:
It looks like this was missed back when DeviceScope was added to the SDK.

### DIFF
--- a/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
+++ b/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
@@ -54,6 +54,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
             // arrange
 
             StorageContainer storageContainer = null;
+            string edgeId = $"{nameof(RegistryManager_ExportDevices)}-Edge-{StorageContainer.GetRandomSuffix(4)}";
             string deviceId = $"{nameof(RegistryManager_ExportDevices)}-{StorageContainer.GetRandomSuffix(4)}";
             var registryManager = RegistryManager.CreateFromConnectionString(TestConfiguration.IoTHub.ConnectionString);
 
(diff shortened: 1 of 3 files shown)


e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
```diff
@@ -54,6 +54,7 @@ public async Task RegistryManager_ExportDevices(StorageAuthenticationType storag
             // arrange
 
             StorageContainer storageContainer = null;
+            string edgeId = $"{nameof(RegistryManager_ExportDevices)}-Edge-{StorageContainer.GetRandomSuffix(4)}";
```
drwill-ms, line 57: To verify that DeviceScope is exported, I need a parent edge device.

iothub/service/src/Device.cs
```diff
@@ -114,7 +114,7 @@ public Device(string id)
         /// relationship.
         /// </summary>
         /// <remarks>
-        /// For leaf devices, the value to set a parent edge device can be retrieved from the parent edge device's <see cref="Scope"/> property.
```
drwill-ms, line 117: removing self-reference

iothub/service/src/ExportImportDevice.cs
```diff
@@ -128,17 +129,29 @@ public string TwinETag
...
+        /// The scope of the device. For edge devices, this is auto-generated and immutable. For leaf devices, set this to create child/parent
+        /// relationship.
+        /// </summary>
+        /// <remarks>
+        /// For leaf devices, the value to set a parent edge device can be retrieved from the parent edge device's device scope property.
+        ///
+        /// For more information, see <see href="https://docs.microsoft.com/azure/iot-edge/iot-edge-as-gateway?view=iotedge-2020-11#parent-and-child-relationships"/>.
+        /// </remarks>
+        [JsonProperty(PropertyName = "deviceScope", NullValueHandling = NullValueHandling.Include)]
+        public string DeviceScope { get; set; }
```
drwill-ms, line 153: The name in the `Device` object was `Scope` even though the on-the-wire-property is `deviceScope`.

When I added device scope to Twin when working on parent scopes, I named it `DeviceScope`, because in context of `Twin` it required the full name. I feel the same is true if `ExportImportDevice`.

I wish we hadn't named it `Device.Scope`, but it is too late to reconsider that.

barustum (approved):

azabbasi (approved):

Issue #2036, opened 2021-06-17
[Bug Report] ProvisioningDeviceClient.RegisterAsync: CancellationToken not effective when offline


its-phil:
# Context

- **OS, version, SKU and CPU architecture used:** Ubuntu 18.04 x86
- **Application's .NET Target Framework :** net5.0
- **SDK version used:** Microsoft.Azure.Devices.Provisioning.Transport.Amqp 1.14.1

## Description of the issue
Calling `ProvisioningDeviceClient.RegisterAsync` with a cancellation token ignores the token timeout if the device is offline (tested on an Azure VM where all outbound connections are blocked). The `ProvisioningDeviceClient` instance is created with a `ProvisioningTransportHandlerAmqp` instance.

I would expect `RegisterAsync` to cancel if I set the cancellation token source timeout to e.g. 15 seconds. Instead, the timeout is always 1 minute. 

This seems related to [this issue](https://github.com/Azure/azure-iot-sdk-csharp/issues/1203) and [this issue](https://github.com/Azure/azure-iot-sdk-csharp/issues/1921) but the workaround mentioned [here](https://github.com/Azure/azure-iot-sdk-csharp/issues/1203#issuecomment-610667859) doesn't seem work with the `ProvisioningDeviceClient`.

Is there another way to work around this issue? I'd like the `RegisterAsync` call to fail faster when the device is offline.

## Code sample exhibiting the issue
```
using var transport = new ProvisioningTransportHandlerAmqp(TransportFallbackType.TcpOnly);
var provisioningDeviceClient = ProvisioningDeviceClient.Create(dpsUrl, scopeId, securityProvider, transport);

const int timeoutMs = 15000;
using var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(timeoutMs));
var sw = new Stopwatch();
try
{
    sw.Start();
    var result = await provisioningDeviceClient.RegisterAsync(cts.Token).ConfigureAwait(false);
    Console.WriteLine(result.Status);
}
catch (Exception ex)
{
    Console.WriteLine(ex.Message);
    Console.WriteLine(ex.InnerException?.Message);
}
finally
{
    sw.Stop();
    Console.WriteLine($"Registration finished or cancelled after {sw.ElapsedMilliseconds} ms.");
    // Here I'd expect ~15 seconds. Instead I always get ~60 seconds.
}
```

timtay-microsoft:
[...]

PR #2041, opened 2021-06-18
feat(provisioning-device, prov-amqp, prov-mqtt, prov-https): Add support for timespan timeouts to provisioning device client

timtay-microsoft:
As discussed in #2036, AMQP provisioning device clients have no way to configure the timeout given to the AMQP library for operations like opening links. This adds overloads to the existing provisioning device client's registerAsync methods that allow for users to configure this timespan timeout.

Old API surface for provisioning device client (which didn't use optional params for some reason?)
```csharp
public Task<DeviceRegistrationResult> RegisterAsync();

public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data);

public Task<DeviceRegistrationResult> RegisterAsync(CancellationToken cancellationToken);

public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data, CancellationToken cancellationToken);
```

Proposed new API surface
```csharp
public Task<DeviceRegistrationResult> RegisterAsync(CancellationToken cancellationToken = default);

public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data, CancellationToken cancellationToken = default);

public Task<DeviceRegistrationResult> RegisterAsync(TimeSpan timeout);

public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data, TimeSpan timeout);
```

### DIFF
--- a/provisioning/device/src/ProvisioningDeviceClient.cs
+++ b/provisioning/device/src/ProvisioningDeviceClient.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Azure.Devices.Provisioning.Client.Transport;
 using Microsoft.Azure.Devices.Shared;
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -63,45 +64,32 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
         /// <summary>
         /// Registers the current device using the Device Provisioning Service and assigns it to an IoT Hub.
         /// </summary>
+        /// <param name="timeout">The maximum amount of time to allow this operation to run for before timing out.</param>
+        /// <remarks>
+        /// Due to the AMQP library used by this library uses not accepting cancellation tokens, this overload and <see cref="RegisterAsync(ProvisioningRegistrationAdditionalData, TimeSpan)"/>
+        /// are the only overloads for this method that allow for a specified timeout to be respected in the middle of an AMQP operation such as opening
+        /// the AMQP connection. MQTT and HTTPS connections do not share that same limitation, though.
+        /// </remarks>
         /// <returns>The registration result.</returns>
-        public Task<DeviceRegistrationResult> RegisterAsync()
+        public Task<DeviceRegistrationResult> RegisterAsync(TimeSpan timeout)
         {
-            return RegisterAsync(CancellationToken.None);
+            Logging.RegisterAsync(this, _globalDeviceEndpoint, _idScope, _transport, _security);
+
+            return RegisterAsync(null, timeout);
         }
 
         /// <summary>
         /// Registers the current device using the Device Provisioning Service and assigns it to an IoT Hub.
         /// </summary>
         /// <param name="data">The optional additional data.</param>
+        /// <param name="timeout">The maximum amount of time to allow this operation to run for before timing out.</param>
+        /// <remarks>
+        /// Due to the AMQP library used by this library uses not accepting cancellation tokens, this overload and <see cref="RegisterAsync(TimeSpan)"/>
+        /// are the only overloads for this method that allow for a specified timeout to be respected in the middle of an AMQP operation such as opening
+        /// the AMQP connection. MQTT and HTTPS connections do not share that same limitation, though.
+        /// </remarks>
         /// <returns>The registration result.</returns>
-        public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data)
-        {
-            return RegisterAsync(data, CancellationToken.None);
-        }
-
-        /// <summary>
-        /// Registers the current device using the Device Provisioning Service and assigns it to an IoT Hub.
-        /// </summary>
-        /// <param name="cancellationToken">The cancellation token.</param>
-        /// <returns>The registration result.</returns>
-        public Task<DeviceRegistrationResult> RegisterAsync(CancellationToken cancellationToken)
-        {
-            Logging.RegisterAsync(this, _globalDeviceEndpoint, _idScope, _transport, _security);
-
-            var request = new ProvisioningTransportRegisterMessage(_globalDeviceEndpoint, _idScope, _security)
-            {
-                ProductInfo = ProductInfo
-            };
-            return _transport.RegisterAsync(request, cancellationToken);
-        }
-
-        /// <summary>
-        /// Registers the current device using the Device Provisioning Service and assigns it to an IoT Hub.
-        /// </summary>
-        /// <param name="data">The custom content.</param>
-        /// <param name="cancellationToken">The cancellation token.</param>
-        /// <returns>The registration result.</returns>
-        public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data, CancellationToken cancellationToken)
+        public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data, TimeSpan timeout)
         {
             Logging.RegisterAsync(this, _globalDeviceEndpoint, _idScope, _transport, _security);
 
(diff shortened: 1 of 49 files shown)


provisioning/device/src/ProvisioningDeviceClient.cs
```diff
@@ -63,52 +64,83 @@ public class ProvisioningDeviceClient
...
         /// </summary>
         /// <param name="cancellationToken">The cancellation token.</param>
+        /// <remarks>
+        /// Due to the AMQP library used by this library uses not accepting cancellation tokens, the provided cancellation token will only be checked 
+        /// for cancellation in between AMQP operations, and not during. In order to have a timeout for this operation that is checked during AMQP operations
+        /// (such as opening the connection), you must use <see cref="RegisterAsync(TimeSpan)"/> instead. MQTT and HTTPS connections do not have the same 
+        /// behavior as AMQP connections in this regard. MQTT and HTTPS connections will check this cancellation token for cancellation during their protocol level operations.
+        /// </remarks>
         /// <returns>The registration result.</returns>
-        public Task<DeviceRegistrationResult> RegisterAsync(CancellationToken cancellationToken)
```
timtay-microsoft, line 87: I'm not sure why we didn't use optional parameters here before, but I'm assuming

```csharp
public Task<DeviceRegistrationResult> RegisterAsync(CancellationToken cancellationToken = default)
```

is equivalent to

```csharp
public Task<DeviceRegistrationResult> RegisterAsync();
public Task<DeviceRegistrationResult> RegisterAsync(CancellationToken cancellationToken);
```

When CancellationToken.None is passed as the token from RegisterAsync() to RegisterAsync(CancellationToken)

provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
     /// </summary>
     public class ProvisioningTransportHandlerAmqp : ProvisioningTransportHandler
     {
-        private static readonly TimeSpan s_defaultOperationPoolingInterval = TimeSpan.FromSeconds(2);
+        private static readonly TimeSpan s_defaultOperationPollingInterval = TimeSpan.FromSeconds(2);
```
timtay-microsoft, line 25: A bit of a tangential change, but this is a **polling** interval not a **pooling** interval

provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -106,19 +135,21 @@ public class ProvisioningTransportHandlerAmqp : ProvisioningTransportHandler
                 string linkEndpoint = $"{message.IdScope}/registrations/{registrationId}";
 
                 using AmqpClientConnection connection = authStrategy.CreateConnection(builder.Uri, message.IdScope);
-                await authStrategy.OpenConnectionAsync(connection, s_timeoutConstant, useWebSocket, Proxy, RemoteCertificateValidationCallback).ConfigureAwait(false);
```
timtay-microsoft, line 109: ```s_timeoutConstant``` was the hardcoded 1 minute timespan that we passed into the AMQP library to use in the place of a cancellation token. This PR allows users to control what that timespan is configured to instead of this hardcoded value

provisioning/transport/http/src/ProvisioningTransportHandlerHttp.cs
```diff
@@ -32,6 +32,20 @@ public ProvisioningTransportHandlerHttp()
...
+        /// </summary>
+        /// <param name="message">The provisioning message.</param>
+        /// <param name="timeout">The maximum amount of time to allow this operation to run for before timing out.</param>
+        /// <returns>The registration result.</returns>
+        public override async Task<DeviceRegistrationResult> RegisterAsync(
+            ProvisioningTransportRegisterMessage message,
+            TimeSpan timeout)
+        {
+            using var cts = new CancellationTokenSource(timeout);
+            return await RegisterAsync(message, cts.Token).ConfigureAwait(false);
```
timtay-microsoft, line 46: Since HTTP and MQTT do support using cancellation tokens all the way through, we can just create a cancellation token out of the timespan here.

provisioning/device/src/ProvisioningDeviceClient.cs
```diff
@@ -63,52 +64,89 @@ public class ProvisioningDeviceClient
...
+        /// </param>
+        /// <param name="timeout">The maximum amount of time to allow this operation to run for before timing out.</param>
+        /// <remarks>
+        /// Due to the AMQP library used by this library uses not accepting cancellation tokens, this overload and <see cref="RegisterAsync(TimeSpan)"/>
+        /// are the only overloads for this method that allow for a specified timeout to be respected in the middle of an AMQP operation such as opening
+        /// the AMQP connection. MQTT and HTTPS connections do not share that same limitation, though.
+        /// </remarks>
         /// <returns>The registration result.</returns>
-        public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data)
+        public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data, TimeSpan timeout)
```
timtay-microsoft, line 95: I was tempted to make the TimeSpan an optional parameter here like the cancellation token is, but we can't have two overloads with the same first parameter and different optional parameters

```csharp
public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data, TimeSpan timeout = default);
public Task<DeviceRegistrationResult> RegisterAsync(ProvisioningRegistrationAdditionalData data, CancellationToken cancellationToken = default);
```

```csharp
public void myMethod()
{
    provisioningDeviceClient.RegisterAsync(data); // Which of the two above options would the compiler choose?
}
```

drwill-ms: Hmm

Could always add TimeSpan as another optional parameter to the one that has a CancellationToken. Would that be less weird?

provisioning/device/src/ProvisioningDeviceClient.cs
drwill-ms: It is, but not binary compatible. Could be seen as a breaking change. I'd be okay with it, though. Simply requires recompile.

drwill-ms (approved):

azabbasi (approved):

provisioning/device/src/ProvisioningDeviceClient.cs
timtay-microsoft: I could go either way on this suggestion. If we ever get cancellation token support for our AMQP library, then we can just deprecate all the APIs that take TimeSpans, but if we start mixing TimeSpans and Cancellation tokens then it gets messy.

abhipsaMisra: I'd like to keep the timespan and cancellation token APIs separate. Since both are ways to signal an API that it should stop the in-progress operation gracefully, I don't think we should mix them together.

provisioning/device/src/ProvisioningDeviceClient.cs
```diff
@@ -63,52 +64,89 @@ public class ProvisioningDeviceClient
...
+        /// Due to the AMQP library used by this library uses not accepting cancellation tokens, this overload and <see cref="RegisterAsync(ProvisioningRegistrationAdditionalData, TimeSpan)"/>
+        /// are the only overloads for this method that allow for a specified timeout to be respected in the middle of an AMQP operation such as opening
+        /// the AMQP connection. MQTT and HTTPS connections do not share that same limitation, though.
+        /// </remarks>
         /// <returns>The registration result.</returns>
-        public Task<DeviceRegistrationResult> RegisterAsync()
+        public Task<DeviceRegistrationResult> RegisterAsync(TimeSpan timeout)
         {
-            return RegisterAsync(CancellationToken.None);
+            Logging.RegisterAsync(this, _globalDeviceEndpoint, _idScope, _transport, _security);
```
abhipsaMisra, line 76: nit: we don't need this logging statement, this API is simply calling the overload which has the relevant logging statements.

provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
abhipsaMisra: amqp just loves pooling everything

provisioning/transport/amqp/src/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
     /// </summary>
     public class ProvisioningTransportHandlerAmqp : ProvisioningTransportHandler
     {
-        private static readonly TimeSpan s_defaultOperationPoolingInterval = TimeSpan.FromSeconds(2);
+        private static readonly TimeSpan s_defaultOperationPollingInterval = TimeSpan.FromSeconds(2);
         private static readonly TimeSpan s_timeoutConstant = TimeSpan.FromMinutes(1);
```
abhipsaMisra, line 26: Since we're already updating this file, it would be useful to rename this to signify which operation this timeout is for.

abhipsaMisra (approved):

abhipsaMisra:
Unit tests/ E2E tests?

timtay-microsoft:
> Unit tests/ E2E tests?

Working on a few E2E tests now, yeah

PR #2046, opened 2021-06-21
Add documentation to repo about which platforms this SDK supports

### LINKED ISSUE #2032: Add specific documentation around what platforms this library supports
It'd be nice to have a formal document where we list which .NET versions and operating systems that we test this SDK against, and lists our stances on other platforms like Xamarin.

Once this document is added, the "new issue" templates should be updated to point users to them so that issues aren't filed for unsupported frameworks/operating systems.

timtay-microsoft:
fixes #2032

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -65,6 +65,8 @@ If you would like to build or change the SDK source code, please follow the [dev
 
 For an official list of all the operating systems and .NET platforms that we support, please see [this document](./supported_platforms.md)
 
+Note that you can configure your TLS protocol version and ciphers by following [this document](./configure_tls_protocol_version_and_ciphers.md)
+
 ## Key features and roadmap
 
 ### IoT Hub Device SDK
--- a/readme.md
+++ b/readme.md
@@ -61,30 +61,6 @@ It shows how to connect a device, handle disconnect events, cases to handle when
 
 If you would like to build or change the SDK source code, please follow the [devguide](doc/devguide.md).
 
-## OS platforms and hardware compatibility
-
-> .NET Standard 1.3 (IoT Hub SDKs only) is last supported in the [2020-02-27](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2020-2-27) and in the [2020-1-31 LTS](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-1-31) releases.
-
-The IoT Hub device SDK for .NET can be used with a broad range of device platforms and is officially supported on the following Operating Systems:
-
-*  Windows versions officially supported by Microsoft.
-*  [Linux distributions](https://docs.microsoft.com/en-us/dotnet/core/install/linux) supported by .NET core.
-
-> Note: For Linux, we test our clients against Ubuntu 16.04.7 LTS.
-
-The NuGet packages provide support for the following .NET flavors:
-- .NET Standard 2.1
-- .NET Standard 2.0
-- .NET Framework 4.7.2 (IoT Hub SDKs only)
-- .NET Framework 4.5.1 (IoT Hub SDKs only)
-
-For details on .NET support see the [.NET Standard documentation](https://docs.microsoft.com/dotnet/standard/net-standard).
-For details on OS support see the following resources:
-
-- [.NET Core Runtime ID Catalog](https://docs.microsoft.com/dotnet/core/rid-catalog)
-- [.NET Framework System Requirements](https://docs.microsoft.com/dotnet/framework/get-started/system-requirements)
-- [Configure TLS Protocol Version and Ciphers](./configure_tls_protocol_version_and_ciphers.md)
-
 ## Key features and roadmap
 
 ### IoT Hub Device SDK
--- a/readme.md
+++ b/readme.md
@@ -61,6 +61,10 @@ It shows how to connect a device, handle disconnect events, cases to handle when
 
 If you would like to build or change the SDK source code, please follow the [devguide](doc/devguide.md).
 
+## OS platforms and hardware compatibility
+
+For an official list of all the operating systems and .NET platforms that we support, please see [this document](./supported_platforms.md)
+
 ## Key features and roadmap
 
 ### IoT Hub Device SDK
--- a/supported_platforms.md
+++ b/supported_platforms.md
@@ -14,7 +14,7 @@ The NuGet packages provide support for the following .NET versions:
 This SDK _may_ work with other versions of .NET, but there are no guarantees that they will _always_ work for those 
 versions nor are there guarantees that we will fix bugs that are only present on those versions.
 
-Note that applications targeting multiple .NET frameworks have some rules for target framework precedence, so applications using this library will pick the net451/net472 dll as per framework precedence.
+Note that applications will resolve the dll that is being referenced based on the framework precedence rules. This means that .NET Framework targeting applications will look for the closet .NET Framework dll. In the absence of that, it will pick up the closest .NET Standard dll. Similarly for netcoreapp applications will look for the closest netcoreapp dll and in the absence of one will pick the closest .NET Standard dll. Since we publish the above list of dlls, you should target the appropriate net target to ensure you get the desired .net API coverage
 
 ## Windows 10
 
--- a/supported_platforms.md
+++ b/supported_platforms.md
@@ -20,7 +20,11 @@ Note that, while we only directly test on Windows 10, we do support other Window
 
 Nightly test platform details:
 
-.NET versions: .NET Framework 4.5.1, .NET Framework 4.7.2, .NET Core 2.1.18, .NET Core 3.1
+.NET versions tested on
+  -.NET Framework 4.5.1
+  -.NET Framework 4.7.2
+  -.NET Core 2.1.18
+  -.NET Core 3.1
 
 Default locale: en_US, platform encoding: Cp1252
 
@@ -32,7 +36,9 @@ Note that, while we only directly test on Ubuntu 1604, we do generally support o
 
 Nightly test platform details:
 
-.NET versions: .NET Core 2.1.18, .NET Core 3.1
+.NET versions tested on:
+  -.NET Core 2.1.18
+  -.NET Core 3.1
 
 Default locale: en_US, platform encoding: UTF-8
 
--- a/supported_platforms.md
+++ b/supported_platforms.md
@@ -47,6 +47,7 @@ OS name: "linux", version: "4.15.0-1113-azure", arch: "amd64", family: "unix"
 
 ## Miscellaneous support notes
 
+- This library has a [preview version](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/preview_2021-6-8) that supports .NET 5.0, but we don't officially support it in our main releases yet.
 - This library does not officially support being run on MacOS.
 - This library does not officially support being run in Xamarin applications.
 - .NET Standard 1.3 (IoT Hub SDKs only) is last supported in the [2020-02-27](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2020-2-27) and in the [2020-1-31 LTS](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-1-31) releases.
--- a/supported_platforms.md
+++ b/supported_platforms.md
@@ -16,15 +16,16 @@ versions nor are there guarantees that we will fix bugs that are only present on
 
 ## Windows 10
 
-Note that, while we only directly test on Windows 10, we do support other Windows versions as well.
+Note that, while we only directly test on Windows 10, we do support other Windows versions officially supported by Microsoft.
 
 Nightly test platform details:
 
 .NET versions tested on
-- .NET Framework 4.5.1
-- .NET Framework 4.7.2
-- .NET Core 2.1.18
 - .NET Core 3.1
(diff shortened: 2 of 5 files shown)


readme.md
```diff
@@ -61,30 +61,6 @@ It shows how to connect a device, handle disconnect events, cases to handle when
...
-- .NET Standard 2.0
-- .NET Framework 4.7.2 (IoT Hub SDKs only)
-- .NET Framework 4.5.1 (IoT Hub SDKs only)
-
-For details on .NET support see the [.NET Standard documentation](https://docs.microsoft.com/dotnet/standard/net-standard).
-For details on OS support see the following resources:
-
-- [.NET Core Runtime ID Catalog](https://docs.microsoft.com/dotnet/core/rid-catalog)
-- [.NET Framework System Requirements](https://docs.microsoft.com/dotnet/framework/get-started/system-requirements)
-- [Configure TLS Protocol Version and Ciphers](./configure_tls_protocol_version_and_ciphers.md)
```
timtay-microsoft, line 86: Most of this would be duplicate information, so it was moved to the new file

readme.md
```diff
@@ -61,30 +61,6 @@ It shows how to connect a device, handle disconnect events, cases to handle when
 
 If you would like to build or change the SDK source code, please follow the [devguide](doc/devguide.md).
 
-## OS platforms and hardware compatibility
```
abhipsaMisra, line 64: the supported versions doc needs to be linked in highlighted in our main readme

readme.md
drwill-ms: shouldn't we link to the other file tho?

supported_platforms.md
```diff
@@ -0,0 +1,51 @@
...
+- .NET Standard 2.0
+- .NET Framework 4.7.2 (IoT Hub SDKs only)
+- .NET Framework 4.5.1 (IoT Hub SDKs only)
+
+This SDK _may_ work with other versions of .NET, but there are no guarantees that they will _always_ work for those 
+versions nor are there guarantees that we will fix bugs that are only present on those versions.
+
+## Windows 10
+
+Note that, while we only directly test on Windows 10, we do support other Windows versions as well.
```
abhipsaMisra, line 19: nit: "....Windows versions officially supported by Microsoft."

supported_platforms.md
```diff
@@ -0,0 +1,51 @@
...
+- .NET Core 2.1.18
+- .NET Core 3.1
+
+Default locale: en_US, platform encoding: Cp1252
+
+OS name: "windows server 2016", version: "10.0", arch: "amd64", family: "windows"
+
+## Ubuntu 1604
+
+Note that, while we only directly test on Ubuntu 1604, we do generally support other popular linux distributions.
```
abhipsaMisra, line 35: same here as well: "[Linux distributions](https://docs.microsoft.com/en-us/dotnet/core/install/linux) supported by .NET core."

readme.md
```diff
@@ -61,30 +61,6 @@ It shows how to connect a device, handle disconnect events, cases to handle when
...
-The NuGet packages provide support for the following .NET flavors:
-- .NET Standard 2.1
-- .NET Standard 2.0
-- .NET Framework 4.7.2 (IoT Hub SDKs only)
-- .NET Framework 4.5.1 (IoT Hub SDKs only)
-
-For details on .NET support see the [.NET Standard documentation](https://docs.microsoft.com/dotnet/standard/net-standard).
-For details on OS support see the following resources:
-
-- [.NET Core Runtime ID Catalog](https://docs.microsoft.com/dotnet/core/rid-catalog)
```
abhipsaMisra, line 84: looks like we lost this and the below links in the new doc

supported_platforms.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Supported .NET versions
+
+The NuGet packages provide support for the following .NET versions:
+- .NET Standard 2.1
+- .NET Standard 2.0
+- .NET Framework 4.7.2 (IoT Hub SDKs only)
+- .NET Framework 4.5.1 (IoT Hub SDKs only)
+
+This SDK _may_ work with other versions of .NET, but there are no guarantees that they will _always_ work for those
```
drwill-ms, line 14: This isn't the right way to think about this.

.NET Standard 2.0 and 2.1 are "standards" not frameworks. A user wouldn't target them. Rather they target a version of .NET that is compatible with these standards and, based on framework loading preferences, gets one of these.

Secondly, if a user us running 4.8, the [framework loading precedence](https://nugettools.azurewebsites.net/5.5.0-preview.2.6382/framework-precedence?framework=net4.8&exludedIdentifiers=) will choose 4.7.2. That doesn't mean we won't support them. My take is that is fully supported by .NET and us.

What are your thoughts?

supported_platforms.md
```diff
@@ -0,0 +1,51 @@
...
+versions nor are there guarantees that we will fix bugs that are only present on those versions.
+
+## Windows 10
+
+Note that, while we only directly test on Windows 10, we do support other Windows versions as well.
+
+Nightly test platform details:
+
+.NET versions tested on
+- .NET Framework 4.5.1
```
abhipsaMisra, line 24: might be worth calling out here as well that these are for hub sdks only

abhipsaMisra: extremely nit: should we order these latest to oldest?

readme.md
timtay-microsoft: Sure, that's reasonable

supported_platforms.md
```diff
@@ -0,0 +1,51 @@
...
+- .NET Framework 4.7.2
+- .NET Core 2.1.18
+- .NET Core 3.1
+
+Default locale: en_US, platform encoding: Cp1252
+
+OS name: "windows server 2016", version: "10.0", arch: "amd64", family: "windows"
```
abhipsaMisra, line 31: How about we add a link to this in our yaml file as well? That way we can remember to update this doc whenever we update our build agents

readme.md
timtay-microsoft: I don't follow why we had these other links within our README in the first place. Do you know why we need them around still?

supported_platforms.md
```diff
@@ -0,0 +1,51 @@
+# Microsoft Azure IoT SDKs for .NET
+
+This SDK is tested nightly on a mix of .NET frameworks on both Windows 10 and on Ubuntu 1604. For additional details
```
abhipsaMisra, line 3: I'd call this ".NET implementations", rather than .NET frameworks, since .netcoreapp isn't .NET framework (the whole naming thin

supported_platforms.md
timtay-microsoft: > Secondly, if a user us running 4.8, the framework loading precedence will choose 4.7.2. That doesn't mean we won't support them. My take is that is fully supported by .NET and us.

This SDK isn't really responsible for defining how framework loading precedence works, so I'm not convinced we need to outline that here.

timtay-microsoft: > .NET Standard 2.0 and 2.1 are "standards" not frameworks. A user wouldn't target them. Rather they target a version of .NET that is compatible with these standards and, based on framework loading preferences, gets one of these.

Would you rather I list the .NET Core and .NET framework versions we support since those are the .NET versions users will actually target when building their applications?

readme.md
abhipsaMisra: - afaik rid are used to specify specific platforms that applications resolve to, i.e. I can compile a comsole app a different way for an x64 system vs an x86 system. I don't remember seeing such references in our csproj file though, so it might just be legacy code from a time where we defined projects differently.
- the system requirements would be useful to highlight here, since that is the min req for using our sdks
- the TLS specifications also need to be highlighted on our landing page (esp with all certs related changes that seem to be coming in the future)

supported_platforms.md
drwill-ms: No. I think listing our targets are fine. I just think we shouldn't say 'there are no guarantees that they will always work".

The SDK isn't responsible for framework loading precedence, but the .NET framework for the app they are running is. MS supports having a newer app load an older library. We shouldn't make it sound like they have to run an app with the target we have, or it might not work or we might not support them.

That is my point.

readme.md
timtay-microsoft: I can add the TLS link and the system requirements link then, sure. I don't see any references to the `<RuntimeIdentifier>` tag that the rid document outlines, so I think we are safe to remove that link though

abhipsaMisra: That sounds good

supported_platforms.md
abhipsaMisra: The iffy versions here are the .NET framework versions that we support. Technically, since our lib is ns2.0 compatible, all implementations of ns2.0+ should work fine. If it doesn't, that's usually a bug in our code => the readonly dict implementation that we had handwritten, and was broken for .NET 5.0.

How about we call out the .NET framework versions saying that they'll pick the net451/ net472 dll as per framework precedence. Anything ns2.0+ should always work.

supported_platforms.md
```diff
@@ -0,0 +1,55 @@
...
+- .NET Core 3.1
+- .NET Core 2.1.18
+
+Default locale: en_US, platform encoding: UTF-8
+
+OS name: "linux", version: "4.15.0-1113-azure", arch: "amd64", family: "unix"
+
+## Miscellaneous support notes
+
+- This library has a [preview version](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/preview_2021-6-8) that supports .NET 5.0, but we don't officially support it in our main releases yet.
```
abhipsaMisra, line 50: thanks for adding this!

timtay-microsoft: Someone needed to celebrate your hard work on that feature!

supported_platforms.md
drwill-ms: sure, try that.

vsts/vsts.yaml
```diff
@@ -32,6 +32,7 @@ jobs:
 
     condition: succeeded()
     pool:
+      # If this is changed, don't forget to update supported_platforms.md in the root directory. That document outlines what OS we test on and should stay up to date.
```
drwill-ms, line 35: nice

drwill-ms (approved):

supported_platforms.md
```diff
@@ -0,0 +1,57 @@
...
+The NuGet packages provide support for the following .NET versions:
+- .NET Standard 2.1
+- .NET Standard 2.0
+- .NET Framework 4.7.2 (IoT Hub SDKs only)
+- .NET Framework 4.5.1 (IoT Hub SDKs only)
+
+This SDK _may_ work with other versions of .NET, but there are no guarantees that they will _always_ work for those 
+versions nor are there guarantees that we will fix bugs that are only present on those versions.
+
+Note that applications targeting multiple .NET frameworks have some rules for target framework precedence, so applications using this library will pick the net451/net472 dll as per framework precedence.
```
timtay-microsoft, line 17:
```suggestion
Note that applications will resolve the dll that is being referenced based on the framework precedence rules. This means that .NET Framework targeting applications will look for the closet .NET Framework dll. In the absence of that, it will pick up the closest .NET Standard dll. Similarly for netcoreapp applications will look for the closest netcoreapp dll and in the absence of one will pick the closest .NET Standard dll. Since we publish the above list of dlls, you should target the appropriate net target to ensure you get the desired .net API coverage
```

supported_platforms.md
```diff
@@ -0,0 +1,57 @@
...
+The NuGet packages provide support for the following .NET versions:
+- .NET Standard 2.1
+- .NET Standard 2.0
+- .NET Framework 4.7.2 (IoT Hub SDKs only)
+- .NET Framework 4.5.1 (IoT Hub SDKs only)
+
+This SDK _may_ work with other versions of .NET, but there are no guarantees that they will _always_ work for those 
+versions nor are there guarantees that we will fix bugs that are only present on those versions.
+
+Note that applications will resolve the dll that is being referenced based on the framework precedence rules. This means that .NET Framework targeting applications will look for the closet .NET Framework dll. In the absence of that, it will pick up the closest .NET Standard dll. Similarly for netcoreapp applications will look for the closest netcoreapp dll and in the absence of one will pick the closest .NET Standard dll. Since we publish the above list of dlls, you should target the appropriate net target to ensure you get the desired .net API coverage
```
abhipsaMisra, line 17: nit: uppercase .NET (in the last statement)

supported_platforms.md
abhipsaMisra: Now that we have the below explanation, we can remove this statement. Based on the .NET dll targeted, their app should always work if the API support is available

abhipsaMisra (approved):

supported_platforms.md
timtay-microsoft:
```suggestion
Note that applications will resolve the dll that is being referenced based on the framework precedence rules. This means that .NET Framework targeting applications will look for the closet .NET Framework dll. In the absence of that, it will pick up the closest .NET Standard dll. Similarly for netcoreapp applications will look for the closest netcoreapp dll and in the absence of one will pick the closest .NET Standard dll. Since we publish the above list of dlls, you should target the appropriate net target to ensure you get the desired .NET API coverage
```

supported_platforms.md
timtay-microsoft:
```suggestion
This SDK _may_ work with newer versions of .NET, but there are no guarantees that they will _always_ work for those until we officially add support for them.
```

PR #2060, opened 2021-07-01
Fix for AMQP ReceiveMessage callback

bikamani:
Do not await user callbacks in the SDK as a good practice.

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -1411,12 +1411,12 @@ namespace Microsoft.Azure.Devices.Client
             await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
             ReceiveMessageCallback callback = _deviceReceiveMessageCallback?.Item1;
             object callbackContext = _deviceReceiveMessageCallback?.Item2;
-            _deviceReceiveMessageSemaphore.Release();
 
             if (callback != null)
             {
-                await callback.Invoke(message, callbackContext).ConfigureAwait(false);
+                _ = callback.Invoke(message, callbackContext);
             }
+            _deviceReceiveMessageSemaphore.Release();
 
             if (Logging.IsEnabled)
             {


vinagesh (approved):

iothub/device/src/InternalClient.cs
```diff
@@ -1411,12 +1411,12 @@ private async Task OnDeviceMessageReceivedAsync(Message message)
...
             ReceiveMessageCallback callback = _deviceReceiveMessageCallback?.Item1;
             object callbackContext = _deviceReceiveMessageCallback?.Item2;
-            _deviceReceiveMessageSemaphore.Release();
 
             if (callback != null)
             {
-                await callback.Invoke(message, callbackContext).ConfigureAwait(false);
+                _ = callback.Invoke(message, callbackContext);
             }
+            _deviceReceiveMessageSemaphore.Release();
```
drwill-ms, line 1419: What is the impact of releasing the semaphore after this callback?

jamdavi: I thought we actually didn't want to do this because we'd be waiting on some customer's code to complete (correctly at that) before we released the lock

jamdavi: If we did want something like this we'd need to add both a cancellation token and a try/finally

jamdavi: Since we're not awaiting the callback probably not much.

drwill-ms: I'd rather not change it unless we have a specific reason for it.

drwill-ms: I thought I saw a comment in email from James about putting this in a try/finally, and I'd agree with that.

jamdavi: I made two comments on the basis of the await still being there (which its not). But I agree the try/finally should do it.

vinagesh: 1) This is required for thread safety. The reason for the semaphore is that the value of callback does not change. _deviceReceiveMessageCallback can be changed, which means callback can start pointing to the new value prematurely.
2) We are no longer awaiting the user callback so moving the release down will not change user behavior
3) We removed the try finally a few commits ago. (I'm not sure why). But it seems useful to add.

vinagesh: Added the try-catch.

drwill-ms (approved):

PR #2069, opened 2021-07-09
[Error Codes] Updated ServerBusyException

jamdavi:


drwill-ms (approved):

iothub/device/src/Common/Exceptions/ServerBusyException.cs
```diff
@@ -9,6 +9,9 @@ namespace Microsoft.Azure.Devices.Client.Exceptions
     /// <summary>
     /// The exception that is thrown when the IoT Hub is busy.
     /// </summary>
+    /// <remarks>
+    /// This exception typically means the service is unavilable due to high load or an unexpected error and is usually transient. The best course of action is to retry your operation after some time. By default, the SDK will utilize the <see cref="ExponentialBackoff"/> retry strategy.
```
vinagesh, line 13: Can point to this doc instead - https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md

PR #2070, opened 2021-07-09
[Error Codes] Updated QuotaExceededException

jamdavi:


iothub/device/src/Common/Exceptions/QuotaExceededException.cs
```diff
@@ -7,8 +7,12 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to add a device fails because the maximum number of registered devices has been reached.
+    /// The exception that is thrown by the device client when the daily message quota for the IoT hub is exceeded.
+    /// <seealso href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-troubleshoot-error-403002-iothubquotaexceeded">Troubleshoot Quota Exceeded</seealso>
```
drwill-ms, line 11: Do we want both SeeAlso and the same link in remarks section? Seems redundant to me.

drwill-ms: Remove the `en-us` path from the URL.

drwill-ms (approved):

iothub/device/src/Common/Exceptions/QuotaExceededException.cs
jamdavi: Yeah thats fair probably don't need both.

iothub/device/src/Common/Exceptions/QuotaExceededException.cs
```diff
@@ -7,8 +7,11 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to add a device fails because the maximum number of registered devices has been reached.
+    /// The exception that is thrown by the device client when the daily message quota for the IoT hub is exceeded.
```
vinagesh, line 10: This is only thrown as an internal exception because we did not want to make breaking changes. https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs#L240

PR #2073, opened 2021-07-09
[Error Codes] Updated MessageTooLarge

jamdavi:


iothub/device/src/Common/Exceptions/MessageTooLargeException.cs
```diff
@@ -10,6 +10,9 @@ namespace Microsoft.Azure.Devices.Client.Exceptions
     /// <summary>
     /// The exception that is thrown when an attempt to send a message fails because the length of the message exceeds the maximum size allowed.
     /// </summary>
+    /// <remarks>
+    /// When the message is too large for IoT Hub you will receive this exception. You should attempt to reduce your message size and send again. For more information on message sizes, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling#other-limits">IoT Hub quotas and throttling | Other limits</see>
```
drwill-ms, line 14: We recently learned that we can get 2 different errors depending on how large the message is. Meaning, if > 1MB, we get a different one, and if too large but < 1MB we get this one.

It might be worth noting that. @abhipsaMisra has more info, IIRC.

abhipsaMisra: You're right, we did have this behavior which has now been fixed in `master`: https://github.com/Azure/azure-iot-sdk-csharp/pull/2053.  We'll now see `MessageTooLargeException` for all message sizes over 256kb limit then sent via AMQP/HTTP.
MQTT , however, will perform a client-level validation and throw an `InvalidOperationException` instead. This would be worth calling out: https://github.com/Azure/azure-iot-sdk-csharp/blob/d46e0f07fe8d80e21e07b41c2e75b0bd1fcb8f80/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs#L1175

jamdavi: So, just to keep things clean we should probably not mention this exception because they are only related by the concept of message size. A customer who is getting the MessageTooLargeException wouldn't need to know about the other limit.

We probably should have repurposed this exception for the >1MB issue.

drwill-ms (approved):

PR #2075, opened 2021-07-09
[Error Codes] Update PartitionNotFound error code

jamdavi:


iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -66,7 +66,11 @@ public enum ErrorCode
         DeviceNotFound = 404001,
 
         JobNotFound = 404002,
-        PartitionNotFound = 404003,
```
jamdavi, line 69: Take note we had the wrong error code in the first place...

drwill-ms: The new name for 404003 is CannotModifyImmutableConfigurationContent, in the service file. I don't see it being returned anywhere also, so not adding that code here seems right.

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -66,7 +66,11 @@ public enum ErrorCode
...
+
+        /// <summary>
+        /// The error is internal to IoT Hub and is likely transient. For more details, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-troubleshoot-error-503003-partitionnotfound">503003 PartitionNotFound</see>.
+        /// </summary>
+        PartitionNotFound = 503003, // We do not handle this error code in our SDK
```
drwill-ms, line 73: I don't see that the service returns this either. Should we add these attributes too then?

```csharp
        [Obsolete("This error does not appear to be thrown by the service.")]
        [EditorBrowsable(EditorBrowsableState.Never)]
```

drwill-ms (approved):

PR #2081, opened 2021-07-09
More updates to exceptions thrown

vinagesh:
## Checklist


iothub/service/src/Common/Data/AmqpErrorMapper.cs
```diff
@@ -10,187 +10,6 @@ namespace Microsoft.Azure.Devices.Common.Exceptions
 {
     internal static class AmqpErrorMapper
     {
-        private const int MaxSizeInInfoMap = 32 * 1024;
```
vinagesh, line 13: This is an internal class and none of these methods were being used.

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
```diff
@@ -7,10 +7,9 @@
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to communicate with a device fails
-    /// because the lock token was lost (if the connection is lost and regained for example).
-    /// This timeout has the same effect as if the message was abandoned.
+    /// This exception does not appear to be returned by the service for ServiceClient operations.
     /// </summary>
+    /// [Obsolete]
```
vinagesh, line 12: Marking as obsolete as there seems to be no reason to get this exception in the ServiceClient.

iothub/device/src/Common/Exceptions/DeviceNotFoundException.cs
```diff
@@ -8,7 +8,19 @@
...
+    /// connection status handler. This exception also corresponds to the following error codes on operation responses:
+    /// <list>
+    /// <item>
+    /// AmqpErrorCode.NotFound
+    /// </item>
+    /// <item>
+    /// HttpStatusCode.NotFound
+    /// </item>
+    /// <item>
+    /// HttpStatusCode.NoContent
```
timtay-microsoft, line 21: In case you raise an eyebrow at this, @vinagesh is just documenting what we currently do. I haven't the foggiest idea why we'd treat a 204 (NoContent) as a DeviceNotFoundException, though

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
timtay-microsoft: Why is the [obsolete] commented out like this?

iothub/device/src/Common/Exceptions/DeviceNotFoundException.cs
```diff
@@ -8,7 +8,19 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to communicate with a device fails because the given device identifier cannot be found.
+    /// The exception is thrown when the device is disabled and will be used to set the status to device disabled in the
+    /// connection status handler. This exception also corresponds to the following error codes on operation responses:
+    /// <list>
+    /// <item>
```
drwill-ms, line 14: For brevity?

```csharp
/// <item>AmqpErrorCode.NotFound</item>
```

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
drwill-ms:
```suggestion
    [Obsolete("This exception does not appear to be returned by the service.")]
```

iothub/device/src/Common/Exceptions/DeviceNotFoundException.cs
vinagesh: yes unfortunately I don't know why we do this but too late to change anyway.

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
vinagesh: This is what I meant to do but looks like I just did it in my head :D

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
```diff
@@ -7,10 +7,9 @@
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to communicate with a device fails
-    /// because the lock token was lost (if the connection is lost and regained for example).
-    /// This timeout has the same effect as if the message was abandoned.
+    /// This exception does not appear to be returned by the service for ServiceClient operations.
     /// </summary>
+    [Obsolete("This exception does not appear to be returned by the service.")]
```
drwill-ms, line 12: Add `[EditorBrowsable(EditorBrowsableState.Never)]`?

drwill-ms (approved):

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
drwill-ms: On second thought, the SDK still throws this exception (to avoid breaking change), so I don't think we should mark it as obsolete.

drwill-ms (changes requested):
The type is still necessary as the SDK still throws it. I don't think we should mark it as deprecated.

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -59,12 +59,14 @@ internal class ExceptionHandlingHelper
                         message: await GetExceptionMessageAsync(response).ConfigureAwait(false))
                 },
 
+#pragma warning disable CS0618 // This exception does not appear to be returned by the service for ServiceClient operations
```
abhipsaMisra, line 62: Is this being marked as obsolete because service code doesn't generate `HttpStatusCode.PreconditionFailed`? How sure are we of that? 
Does our SDK translate some error code to `HttpStatusCode.PreconditionFailed`?

iothub/device/src/Common/Exceptions/DeviceMaximumQueueDepthExceededException.cs
```diff
@@ -8,7 +8,8 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to enqueue a message fails because the message queue for the device is already full.
+    /// This exception actually corresponds to IoTHubQuotaExceeded. For more information, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-troubleshoot-error-403002-iothubquotaexceeded"/>.
```
abhipsaMisra, line 11: How about instead od simply saying "For more information, see...", we say "For more information on what causes this error and the steps to fix it, see ..."?

iothub/device/src/Common/Exceptions/DeviceMessageLockLostException.cs
```diff
@@ -8,10 +8,14 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to communicate with a device fails because the lock token was lost (if the connection is lost and regained for example). This timeout has the same effect as if the message was abandonned.
+    /// The exception that is thrown when an attempt to communicate with a device fails because the lock token was lost
+    /// (if the connection is lost and regained for example). This timeout has the same effect as if the message was abandonned.
```
abhipsaMisra, line 12: "This timeout has the same effect as if the message was abandoned." - what timeout? (also, typo with "abandonned").

How about we call out - This exception is thrown when attempting to reject/abandon/complete a cloud-to-device message with a lock token that has already expired. This can happen if you reject/abandon/complete a message after 1 minute of receiving it, or your client connection was lost and regained receiving the message but before you could reject/abandon/complete it.

iothub/device/src/Common/Exceptions/DeviceMessageLockLostException.cs
```diff
@@ -8,10 +8,14 @@
...
     /// </summary>
     /// <remarks>
-    /// An abandoned message will be re-enqueued in the per-device queue, and the <see cref="DeviceClient"/> instance will receive it again. A rejected message will be deleted from the queue and not received again by the device.
+    /// An abandoned message will be re-enqueued in the per-device queue, and the <see cref="DeviceClient"/> instance will receive it again.
```
abhipsaMisra, line 15: device/module

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
vinagesh: Based on my analysis, this exception should not be thrown by the ServiceClient. However, there is no good way to confirm. There is no explanation I can add for when this exception would be thrown. Thoughts on how to proceed other than marking it as obsolete?

iothub/device/src/Common/Exceptions/DeviceMaximumQueueDepthExceededException.cs
vinagesh: Sure

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
vinagesh: Actually on further analysis, I think we are converting HttpStatus PreconditionFailed to this error for some reason. I will revert this change and just document that for this exception.

common/src/service/ExceptionHandlingHelper.cs
vinagesh: Actually this was a result of marking the DeviceMessageLockLostException as obsolete. But this is converting a specific Http Status to this exception so I'll be reverting the change and just documenting this in the exception.

iothub/device/src/Common/Exceptions/DeviceMessageLockLostException.cs
vinagesh: "This timeout has the same effect as if the message was abandoned." - This was already present. I don't have full context so left it as is. If you have better knowledge on this, I can update.

vinagesh: As of now,  I'm adding what you suggested above instead.

vinagesh: FYI - It does not timeout in 1 minute so I don't think we should specify that value.

abhipsaMisra: I think the timeout statement was probably copied over from somewhere. My best is that it meant that even if you abandon a message, if the lock duration has already timed out then this exception will be thrown.

iothub/device/src/Common/Exceptions/DeviceMessageLockLostException.cs
```diff
@@ -8,10 +8,16 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to communicate with a device fails because the lock token was lost (if the connection is lost and regained for example). This timeout has the same effect as if the message was abandonned.
+    /// The exception that is thrown when an attempt to communicate with a device fails because the lock token was lost
```
abhipsaMisra, line 11: nit: I don't particularly like the wording "...an attempt to communicate with a device fails..." that we had previously. The communication with the device is fine, it is rather telling you over that communication what the operation that you wanted to execute failed. 
"Communication with the device failed" reads to me as the communication act itself failed.

vinagesh: Sure we can change the wording. What do you have in mind? How about just removing the first sentence altogether? We can start with "This exception is thrown when attempting to reject/abandon/complete ...." ?

iothub/device/src/Common/Exceptions/DeviceMessageLockLostException.cs
```diff
@@ -8,10 +8,16 @@
...
+    /// token that has already expired. The lock token expires after the lock timeout set by the service, or if your
+    /// client connection was lost and regained while receiving the message but before you could reject/abandon/complete it.
     /// </summary>
     /// <remarks>
-    /// An abandoned message will be re-enqueued in the per-device queue, and the <see cref="DeviceClient"/> instance will receive it again. A rejected message will be deleted from the queue and not received again by the device.
+    /// An abandoned message will be re-enqueued in the per-device/module queue, and the <see cref="DeviceClient"/> instance will receive it again.
+    /// A rejected message will be deleted from the queue and not received again by the device.
+    /// For more information on the cause for this error and how to resolve, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-troubleshoot-error-412002-devicemessagelocklost"/>.
+    /// For more information on cloud-to-device message life cycle, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-c2d#the-cloud-to-device-message-life-cycle"/>.
```
drwill-ms, line 20: nit: Although `life cycle` and `lifecycle` are both valid ways to spell this, in tech we usually choose the compound word.

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
```diff
@@ -7,9 +7,11 @@
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to communicate with a device fails
-    /// because the lock token was lost (if the connection is lost and regained for example).
-    /// This timeout has the same effect as if the message was abandoned.
+    /// This exception is not directly returned by the service for ServiceClient operations. However, the following
+    /// status codes are converted to this exception:
+    /// <list>
+    /// <item>HttpStatusCode.PreconditionFailed</item>
```
drwill-ms, line 13: If it is a list of 1, shall we just mention the one error code? As in, no list?

iothub/service/src/Common/Exceptions/ServerBusyException.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Azure.Devices.Common.Exceptions
     /// The exception that is thrown when the IoT Hub is busy.
     /// </summary>
     /// <remarks>
-    /// This exception typically means the service is unavilable due to high load or an unexpected error and is usually transient. The best course of action is to retry your operation after some time.
+    /// This exception typically means the service is unavailable due to high load or an unexpected error and is usually transient. The best course of action is to retry your operation after some time. By default, the SDK will utilize the <see cref="ExponentialBackoff"/> retry strategy.
```
drwill-ms, line 13: The service doesn't use a retry strategy. This type is in the device client, and my local build fails due to this reference.

iothub/service/src/Common/Exceptions/DeviceMessageLockLostException.cs
vinagesh: Yes we can do that. I thought there would be more, but it ended up being just one.

iothub/service/src/Common/Exceptions/ServerBusyException.cs
vinagesh: Oh this was the resolution of a rebase. I think someone added it previously and it got to my branch.

iothub/device/src/Common/Exceptions/DeviceMessageLockLostException.cs
vinagesh: Let me know if you have other thoughts. I've resolved it by removing the first line.

iothub/service/src/Common/Exceptions/ServerErrorException.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Azure.Devices.Common.Exceptions
     /// The exception that is thrown when the IoT Hub returned an internal service error.
     /// </summary>
     /// <remarks>
-    /// This exception typically means the IoT Hub service has encountered an unexpected error and is usually transient. Please review the <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-troubleshoot-error-500xxx-internal-errors">500xxx Internal errors</see> guide for more information. The best course of action is to retry your operation after some time. By default, the SDK will utilize the <see cref="ExponentialBackoff"/> retry strategy.
+    /// This exception typically means the IoT Hub service has encountered an unexpected error and is usually transient. Please review the <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-troubleshoot-error-500xxx-internal-errors">500xxx Internal errors</see> guide for more information. The best course of action is to retry your operation after some time.
```
vinagesh, line 13: This was causing a build error as the retry is not present in the ServiceClient.

drwill-ms (approved):

PR #2082, opened 2021-07-12
More codes doc'd and code cleanup

drwill-ms:


common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -17,97 +17,88 @@ namespace Microsoft.Azure.Devices
 {
     internal class ExceptionHandlingHelper
     {
-        public static IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> GetDefaultErrorMapping()
+        private static readonly IReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> s_mappings =
```
drwill-ms, line 20: Instead of creating this dictionary for every call, make it static/readonly/readonlydictionary and return the same instance.

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -121,22 +112,16 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
             // to 'error code' enum mapping, the SDK will check if both values are a match. If so, the SDK will populate the exception with the proper Code. In the case where
             // there is a mismatch between the error code and the description, the SDK returns ErrorCode.InvalidErrorCode and log a warning.
 
-            int errorCode;
+            int errorCodeValue = (int)ErrorCode.InvalidErrorCode;
```
drwill-ms, line 115: Initialize it and no longer need the `else` condition.

common/src/service/HttpClientHelper.cs
```diff
@@ -45,14 +45,14 @@ internal sealed class HttpClientHelper : IHttpClientHelper
...
             IAuthorizationHeaderProvider authenticationHeaderProvider,
-            IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping,
+            IReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping,
             TimeSpan timeout,
             IWebProxy customHttpProxy,
             int connectionLeaseTimeoutMilliseconds)
         {
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
-            _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
```
drwill-ms, line 55: It was worse than I first thought. No only was a new dictionary being created on each call, we were copying it into another new dictionary.

iothub/device/src/TransientFaultHandling/RetryStrategy.cs
```diff
@@ -63,73 +66,57 @@ internal abstract class RetryStrategy
         /// </summary>
         public const bool DefaultFirstFastRetry = true;
 
-        private static readonly RetryStrategy s_noRetry = new FixedInterval(0, DefaultRetryInterval);
```
drwill-ms, line 66: Turned into autoproperties, therefore moved below the ctor.

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -7,14 +7,24 @@
...
     /// <summary>
-    /// Error Codes for common IoT hub exceptions.
+    /// Error codes for common IoT hub exceptions.
     /// </summary>
     public enum ErrorCode
     {
 #pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+
+        /// <summary>
+        /// Used when the error code returned by the hub is unrecognized. If encountered, please report the issue so it can be added here.
```
drwill-ms, line 17: I hate that we return `InvalidErrorCode` when we don't know. Why don't we blindly pass the error code we receive back to the user? Seems we enforce a known filter, and that seems problematic to me. Ideally we'd have the code that was returned, and class with known string constants for users and other code.

Too late to change, I think. Open to ideas.

iothub/device/src/TransientFaultHandling/ExponentialBackoffRetryStrategy.cs
```diff
@@ -1,31 +1,34 @@
+// Copyright (c) Microsoft. All rights reserved.
```
vinagesh, line 1: Duplicate license info

iothub/device/src/TransientFaultHandling/ExponentialBackoffRetryStrategy.cs
```diff
@@ -1,31 +1,34 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //Copyright(c) Microsoft.All rights reserved.
```
azabbasi, line 3: I think these two can be consolidated into one header no?

iothub/service/src/Common/Exceptions/ServerErrorException.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Azure.Devices.Common.Exceptions
     /// The exception that is thrown when the IoT Hub returned an internal service error.
     /// </summary>
     /// <remarks>
-    /// This exception typically means the IoT Hub service has encountered an unexpected error and is usually transient. Please review the <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-troubleshoot-error-500xxx-internal-errors">500xxx Internal errors</see> guide for more information. The best course of action is to retry your operation after some time. By default, the SDK will utilize the <see cref="ExponentialBackoff"/> retry strategy.
```
drwill-ms, line 13: We have service SDK exception referring to a device SDK type, which fails my local build. Also, the service SDK doesn't have a retry strategy, right?

iothub/device/src/TransientFaultHandling/ExponentialBackoffRetryStrategy.cs
```diff
@@ -97,7 +104,7 @@ public override ShouldRetry GetShouldRetry()
             {
                 if (currentRetryCount < _retryCount)
                 {
-                    Random random = new Random();
+                    var random = new Random();
```
vinagesh, line 107: Make this static in the class?

iothub/device/src/TransientFaultHandling/RetryStrategy.cs
```diff
@@ -1,24 +1,27 @@
-//Copyright(c) Microsoft.All rights reserved.
-//Microsoft would like to thank its contributors, a list
-//of whom are at http://aka.ms/entlib-contributors
+// Copyright (c) Microsoft. All rights reserved.
```
vinagesh, line 1: Line 1 and 3 are duplicate

iothub/device/src/TransientFaultHandling/ExponentialBackoffRetryStrategy.cs
```diff
@@ -1,31 +1,34 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied. See the License for the specific language governing permissions
+// and limitations under the License.
 
 // THIS FILE HAS BEEN MODIFIED FROM ITS ORIGINAL FORM.
 // Change Log:
 // 9/1/2017 jasminel Renamed namespace to Microsoft.Azure.Devices.Client.TransientFaultHandling and modified access modifier to internal.
+// 7/12/2021 drwill Renamed class from ExponentialBackoff to ExponentialBackoffRetryStrategy to avoid naming internal conflict.
```
timtay-microsoft, line 24: It's a bit unusual to have a changelog within a particular file. Given that we likely made changes between 2017 and now, I'd be fine with scrapping these comments

iothub/service/src/Common/Exceptions/ErrorCode.cs
vinagesh: Probably something we can add to the list of meaningful breaking changes we want to propose as part of a release like we did in Java?

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -7,14 +7,24 @@
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
     /// <summary>
-    /// Error Codes for common IoT hub exceptions.
+    /// Error codes for common IoT hub exceptions.
     /// </summary>
     public enum ErrorCode
     {
 #pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
vinagesh, line 14: Do we need to suppress this after adding the xml comments?

iothub/service/src/Common/Exceptions/ServerErrorException.cs
vinagesh: No retry strategy that I know of.

timtay-microsoft: Correct. None of our service clients (ServiceClient, RegistryManager, JobClient, etc) have any retry logic. Good catch here

vinagesh (approved):

timtay-microsoft (approved):

iothub/device/src/TransientFaultHandling/ExponentialBackoffRetryStrategy.cs
drwill-ms: I suspect it is because we copied this code from someone else, and as a condition of that license agreement, we list the changes we've made. I don't know if it matters, but it cost me little to append the change log.

iothub/service/src/Common/Exceptions/ErrorCode.cs
drwill-ms: I'll start a list

iothub/service/src/Common/Exceptions/ErrorCode.cs
drwill-ms: When we finish, I'll remove the suppression.

drwill-ms: Note this is for the whole file, as the end pragma is at the end of the class.

azabbasi (approved):

PR #2083, opened 2021-07-12
refactor(iot-service): Add comments for service returned error codes

abhipsaMisra:


iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -79,8 +79,8 @@ public enum ErrorCode
         /// request or user input.
         /// </para>
         /// </summary>
-        //[Obsolete("This error does not appear to be returned by the service.")]
-        //[EditorBrowsable(EditorBrowsableState.Never)]
```
abhipsaMisra, line 83: Was this mistakenly commented out?

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -128,9 +128,7 @@ public enum ErrorCode
         JobQuotaExceeded = 403003,
 
         /// <summary>
-        /// The underlying cause is that the number of messages enqueued for the device exceeds the queue limit (50).
-        /// The most likely reason that you're running into this limit is because you're using HTTPS to receive the message,
-        /// which leads to continuous polling using ReceiveAsync, resulting in IoT hub throttling the request.
```
abhipsaMisra, line 133: The statement about HTTP polling for receive operations causing requests to be throttled is irrelevant to device queue depth exceeded error code. Let me know if you had something else in mind while adding this.

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -156,7 +154,8 @@ public enum ErrorCode
         JobNotFound = 404002,
 
         /// <summary>
-        /// The error is internal to IoT Hub and is likely transient. For more details, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-troubleshoot-error-503003-partitionnotfound">503003 PartitionNotFound</see>.
+        /// The error is internal to IoT Hub and is likely transient.
+        /// For more details, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-troubleshoot-error-503003-partitionnotfound">503003 PartitionNotFound</see>.
```
abhipsaMisra, line 158: new line

iothub/service/src/Common/Exceptions/ErrorCode.cs
drwill-ms: No, I was still checking how we can deprecate this. The problem is we have some code that refers to it, so that could would need to be updated (aka removed) if we choose to obsolete it.

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -172,7 +171,8 @@ public enum ErrorCode
         ModuleAlreadyExistsOnDevice = 409301,
 
         /// <summary>
-        /// The etag in the request does not match the etag of the existing resource, as per <see href="https://datatracker.ietf.org/doc/html/rfc7232">RFC7232</see>. The etag is controlled by the service and is based on the device identity it should not be updated in normal operations.
+        /// The etag in the request does not match the etag of the existing resource, as per <see href="https://datatracker.ietf.org/doc/html/rfc7232">RFC7232</see>.
+        /// The etag is controlled by the service and is based on the device identity it should not be updated in normal operations.
```
abhipsaMisra, line 175: new line

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -128,9 +128,7 @@ public enum ErrorCode
         JobQuotaExceeded = 403003,
 
         /// <summary>
-        /// The underlying cause is that the number of messages enqueued for the device exceeds the queue limit (50).
-        /// The most likely reason that you're running into this limit is because you're using HTTPS to receive the message,
-        /// which leads to continuous polling using ReceiveAsync, resulting in IoT hub throttling the request.
+        /// The underlying cause is that the number of cloud-to-device messages enqueued for the device exceeds the queue limit (50).
```
drwill-ms, line 131: So what does a user do if they see this? What is our guidance?

iothub/service/src/Common/Exceptions/ErrorCode.cs
abhipsaMisra: "The etag is controlled by the service and is based on the device identity it should not be updated in normal operations." - was this meant to mean "...is based on the device identity _and_ it should not be ~updated~ used in normal operations."?

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -185,7 +185,8 @@ public enum ErrorCode
 
         // RequestEntityTooLarge - 413
         /// <summary>
-        /// When the message is too large for IoT Hub you will receive this error. You should attempt to reduce your message size and send again. For more information on message sizes, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-quotas-throttling#other-limits">IoT Hub quotas and throttling | Other limits</see>
+        /// When the message is too large for IoT Hub you will receive this error. You should attempt to reduce your message size and send again.
+        /// For more information on message sizes, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-quotas-throttling#other-limits">IoT Hub quotas and throttling | Other limits</see>
```
abhipsaMisra, line 189: new line

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -219,8 +220,8 @@ public enum ErrorCode
         /// <summary>
         /// IoT hub ran into a server side issue.
         /// There can be a number of causes for a 500xxx error response. In all cases, the issue is most likely transient.
-        /// IoT hub nodes can occasionally experience transient faults. When your device tries to connect to a node that is
-        /// having issues, you receive this error. To mitigate 500xxx errors, issue a retry from the device.
+        /// IoT hub nodes can occasionally experience transient faults. When your application tries to connect to a node that is
+        /// having issues, you receive this error. To mitigate 500xxx errors, issue a retry from your application.
```
abhipsaMisra, line 224: this error is returned to the service-side application, which isn't associated with a particular device identity.

iothub/service/src/Common/Exceptions/ErrorCode.cs
abhipsaMisra: I'll add the recovery steps, sure

iothub/service/src/Common/Exceptions/ErrorCode.cs
abhipsaMisra: I see, so I am going to leave it commented out. Are you tracking this error code or have we added a note in the excel doc to track this?

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -128,9 +128,10 @@ public enum ErrorCode
         JobQuotaExceeded = 403003,
 
         /// <summary>
-        /// The underlying cause is that the number of messages enqueued for the device exceeds the queue limit (50).
-        /// The most likely reason that you're running into this limit is because you're using HTTPS to receive the message,
-        /// which leads to continuous polling using ReceiveAsync, resulting in IoT hub throttling the request.
+        /// The underlying cause is that the number of cloud-to-device messages enqueued for the device exceeds the queue limit (50).
+        /// You will need to receive and complete/reject the messages from the device-side before you can enqueue any additional messages.
+        /// If you want to discard the currently enqueued messages,
+        /// you can <see href="https://github.com/Azure/azure-iot-sdk-csharp/blob/c76a64de272da986b6840251f482249e094a725c/iothub/service/src/ServiceClient.cs#L211">purge your device message queue</see>.
```
timtay-microsoft, line 134: Could you link the method within our codebase via cref rather than through github?

abhipsaMisra: sure

iothub/service/src/Common/Exceptions/ErrorCode.cs
timtay-microsoft: I'd add something to the effect of "In order to get the up-to-date ETag for a twin use <link to registryManager.getTwin(deviceId)>"

iothub/service/src/Common/Exceptions/ErrorCode.cs
timtay-microsoft: I don't follow what this error was getting at either. I'm in favor of deleting it as well

iothub/service/src/Common/Exceptions/ErrorCode.cs
timtay-microsoft: And maybe mention that this is a mechanism for protecting against the race conditions of multiple clients updating the same resource and overwriting each other. Linking the RFC is nice, but I don't expect most people to want to follow the link in order to learn why we are doing this

timtay-microsoft (approved):
Just a few small comments

vinagesh (approved):

iothub/service/src/Common/Exceptions/ErrorCode.cs
abhipsaMisra: @drwill-ms I've re-commented out the code. Let me know if you have any other comments.

iothub/service/src/Common/Exceptions/ErrorCode.cs
jamdavi: Do we want to include a real number here? What if this changes? 

In all cases where we have a condition where you are meeting or exceeding some hard limit defined by the service we should refer to the documentation for this number.

The only exception would be when we add a limitation in code and then we should link to the property/setting that controls it.

jamdavi (approved):
Approved with suggestions

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -128,9 +128,10 @@ public enum ErrorCode
         JobQuotaExceeded = 403003,
 
         /// <summary>
-        /// The underlying cause is that the number of messages enqueued for the device exceeds the queue limit (50).
-        /// The most likely reason that you're running into this limit is because you're using HTTPS to receive the message,
-        /// which leads to continuous polling using ReceiveAsync, resulting in IoT hub throttling the request.
+        /// The underlying cause is that the number of cloud-to-device messages enqueued for the device exceeds the queue limit (50).
```
jamdavi, line 131: Do we want to include a real number here? What if this changes?

In all cases where we have a condition where you are meeting or exceeding some hard limit defined by the service we should refer to the documentation for this number.

The only exception would be when we add a limitation in code and then we should link to the property/setting that controls it.

drwill-ms: Best refer to service documentation

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -156,7 +157,8 @@ public enum ErrorCode
         JobNotFound = 404002,
 
         /// <summary>
-        /// The error is internal to IoT Hub and is likely transient. For more details, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-troubleshoot-error-503003-partitionnotfound">503003 PartitionNotFound</see>.
+        /// The error is internal to IoT Hub and is likely transient.
+        /// For more details, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-troubleshoot-error-503003-partitionnotfound">503003 PartitionNotFound</see>.
         /// </summary>
         PartitionNotFound = 503003, // We do not handle this error code in our SDK
```
drwill-ms, line 163: obsolete and not browsable?

iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs
```diff
@@ -26,7 +26,6 @@ internal static class AmqpIotErrorAdapter
         public static readonly AmqpSymbol ArgumentError = AmqpIotConstants.Vendor + ":argument-error";
         public static readonly AmqpSymbol ArgumentOutOfRangeError = AmqpIotConstants.Vendor + ":argument-out-of-range";
         public static readonly AmqpSymbol DeviceContainerThrottled = AmqpIotConstants.Vendor + ":device-container-throttled";
-        public static readonly AmqpSymbol PartitionNotFound = AmqpIotConstants.Vendor + ":partition-not-found";
```
abhipsaMisra, line 29: the service team confirmed that this error code should never be returned to clients.

drwill-ms (approved):

PR #2085, opened 2021-07-12
Updating exception details for IotHubCommunicationException

vinagesh:
## Checklist

### DIFF
--- a/iothub/device/src/Common/Exceptions/IotHubCommunicationException.cs
+++ b/iothub/device/src/Common/Exceptions/IotHubCommunicationException.cs
@@ -7,8 +7,16 @@ using System.Runtime.Serialization;
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to communicate with the IoT Hub service fails.
+    /// This exception is thrown when an attempt to communicate with the IoT hub service fails due to transient
+    /// network errors after exhausting all the retries based on the retry policy set on the client or
+    /// due to operation timeouts.
     /// </summary>
+    /// <remarks>
+    /// By default, the SDK indefinitely retries dropped connections, unless the retry policy is overridden.
+    /// For more information on the SDK's retry policy and how to override it, see <see href="https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md"/>.
+    /// When the exception is thrown due to operation timeouts, the inner exception will have OperationCanceledException.
+    /// Retrying operations failed due to timeouts could resolve the error.
+    /// </remarks>
     [Serializable]
     public sealed class IotHubCommunicationException : IotHubException
     {
--- a/iothub/service/src/Common/Exceptions/IotHubCommunicationException.cs
+++ b/iothub/service/src/Common/Exceptions/IotHubCommunicationException.cs
@@ -7,7 +7,8 @@ using System.Runtime.Serialization;
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
     /// <summary>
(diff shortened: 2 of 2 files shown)


iothub/service/src/Common/Exceptions/IotHubCommunicationException.cs
```diff
@@ -7,7 +7,8 @@
 namespace Microsoft.Azure.Devices.Common.Exceptions
 {
     /// <summary>
```
vinagesh, line 9: https://github.com/Azure/azure-iot-sdk-csharp/blob/cf5bb3e1c5bd865ee36345772dfa862b71a21b1a/iothub/service/src/AmqpClientHelper.cs#L22

iothub/device/src/Common/Exceptions/IotHubCommunicationException.cs
```diff
@@ -7,8 +7,14 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when an attempt to communicate with the IoT Hub service fails.
+    /// This exception is thrown when an attempt to communicate with the IoT Hub service fails due to transient
```
drwill-ms, line 10: After retries have exhausted?

vinagesh: We have indefinite retries by default in the DeviceClient unless the customers changes the retry policy. In that case, they will have to recreate the client. Do you want me to add those details?

drwill-ms: I think it was unclear if I'd get the exception and a retry would be happening automatically for me, or if I'd only get the exception after retries are exhausted.

I assume the latter. In that case, yes I also think it would be worth mentioning that, unless overridden, the SDK will retry indefinitely.

vinagesh: @drwill-ms - Updated the messaging.

drwill-ms (approved):

iothub/device/src/Common/Exceptions/IotHubCommunicationException.cs
```diff
@@ -7,8 +7,16 @@
...
-    /// The exception that is thrown when an attempt to communicate with the IoT Hub service fails.
+    /// This exception is thrown when an attempt to communicate with the IoT hub service fails due to transient
+    /// network errors after exhausting all the retries based on the retry policy set on the client or
+    /// due to operation timeouts.
     /// </summary>
+    /// <remarks>
+    /// By default, the SDK indefinitely retries dropped connections, unless the retry policy is overridden.
+    /// For more information on the SDK's retry policy and how to override it, see <see href="https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md"/>.
+    /// When the exception is thrown due to operation timeouts, the inner exception will have OperationCanceledException.
+    /// Retrying operations failed due to timeouts could resolve the error.
```
vinagesh, line 18: Note: I changed this to say "could resolve the error" because we cannot guarantee success on retries.

iothub/service/src/Common/Exceptions/IotHubCommunicationException.cs
```diff
@@ -7,7 +7,8 @@
...
-    /// The exception that is thrown when an attempt to communicate with the IoT Hub fails.
+    /// This exception is thrown when an attempt to communicate with the IoT hub service fails due to transient
+    /// network issues or operation timeouts. Retrying failed operations could resolve the error.
```
vinagesh, line 11: Note: I changed this to say "could resolve the error" because we cannot guarantee success on retries.

PR #2086, opened 2021-07-12
Updating exception details for IotHubSuspendedException

vinagesh:
## Checklist

### DIFF
--- a/iothub/device/src/Common/Exceptions/IotHubSuspendedException.cs
+++ b/iothub/device/src/Common/Exceptions/IotHubSuspendedException.cs
@@ -8,7 +8,8 @@ using Microsoft.Azure.Devices.Client.Extensions;
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when the IoT Hub has been suspended.
+    /// This exception is thrown when the IoT hub has been suspended. This is likely due to exceeding Azure
+    /// spending limits. To resolve the error, check the Azure bill and ensure there are enough credits.
     /// </summary>
     [Serializable]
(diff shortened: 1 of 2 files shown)


iothub/device/src/Common/Exceptions/IotHubSuspendedException.cs
```diff
@@ -8,7 +8,8 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when the IoT Hub has been suspended.
+    /// This exception is thrown when the IoT Hub has been suspended. This is likely due to exceeding your
```
drwill-ms, line 11: Casing should be "IoT hub" with a lowercase 'h' per style guidelines.

drwill-ms: FWIW, I prefer not using 2nd person in errors. Feels too personal. I think it sounds less accusatory to phrase it like "This is likely due to exceeding Azure spending limits. To resolve the error, check the Azure bill and ensure there are enough credits."

drwill-ms (approved):

iothub/device/src/Common/Exceptions/IotHubSuspendedException.cs
vinagesh: Any suggestions on wording here? "This error is likely due to exceeding Azure spending limits. This error will be resolved when where is enough credits in the Azure account." ?

vinagesh: Cool. I'll update.

jamdavi (approved):

PR #2088, opened 2021-07-12
Updating exception details for IotHubThrottledException

vinagesh:
## Checklist

### DIFF
--- a/iothub/device/src/Common/Exceptions/IotHubThrottledException.cs
+++ b/iothub/device/src/Common/Exceptions/IotHubThrottledException.cs
@@ -8,8 +8,12 @@ using Microsoft.Azure.Devices.Client.Extensions;
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when the service requires exponential back-off because it has exceeded the maximum number of allowed active requests.
+    /// This exception is thrown when the requests to the IoT hub exceed the limits based on the tier of the hub.
+    /// Retrying with exponential back-off could resolve this error.
     /// </summary>
+    /// <remarks>
+    /// For information on the IoT hub quotas and throttling, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-quotas-throttling"/>.
+    /// </remarks>
     [Serializable]
     public sealed class IotHubThrottledException : IotHubException
     {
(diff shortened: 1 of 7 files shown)


timtay-microsoft (approved):

drwill-ms (approved):

iothub/device/src/Common/Exceptions/IotHubThrottledException.cs
```diff
@@ -8,8 +8,12 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when the service requires exponential back-off because it has exceeded the maximum number of allowed active requests.
+    /// This exception is thrown when the requests to the IoT hub exceed the limits based on the tier of the hub.
+    /// Retrying with exponential back-off could resolve this error.
     /// </summary>
+    /// <remarks>
+    /// For information on the IoT hub quotas and throttling, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-quotas-throttling"/>.
```
drwill-ms, line 15: need to remove `/en-us` from path

iothub/device/src/Common/Exceptions/IotHubThrottledException.cs
```diff
@@ -8,8 +8,12 @@
 namespace Microsoft.Azure.Devices.Client.Exceptions
 {
     /// <summary>
-    /// The exception that is thrown when the service requires exponential back-off because it has exceeded the maximum number of allowed active requests.
+    /// This exception is thrown when the requests to the IoT hub exceed the limits based on the tier of the hub.
+    /// Retrying with exponential back-off could resolve this error.
```
abhipsaMisra, line 12: nit: do you think it would be useful to inform users why retrying with exponential backoff with jitter is more useful than retrying with constant time?

abhipsaMisra (approved):

iothub/device/src/Common/Exceptions/IotHubThrottledException.cs
vinagesh: It could be but it might also be information overload. But this is all very dependent on the use-case of the customer and what IoT hub service does (which we ourselves don't know). I'm inclined to leave it this way and see if customers really ask for more information.

PR #2089, opened 2021-07-12
Document InvalidOperation and ArgumentInvalid

drwill-ms:


iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -35,8 +35,21 @@ public enum ErrorCode
...
+        /// </summary>
+        // Note: although infrequent, this does appear in logs for "Amqp Message.Properties.To must contain the device identifier".
+        // and perhaps other cases.
         InvalidOperation = 400003,
+
+        /// <summary>
+        /// Something in the request payload is invalid. Check the error message for more information about what
+        /// is invalid.
+        /// </summary>
+        // Note: one example found in logs is for invalid characters in a twin property name.
```
vinagesh, line 50: Should we mention that we found it in logs? It is probably not relevant to the customer where we found it.

vinagesh (approved):

iothub/service/src/Common/Exceptions/ErrorCode.cs
drwill-ms: Yeah, not sure how to word this. It is a code comment for us. I don't specify what logs, just logs, so I thought maybe being ambiguous was sufficient.

Thoughts?

vinagesh: I missed that it was just a code comment and not in the summary. Seems reasonable to keep it as a code comment for our reference.

PR #2090, opened 2021-07-12
400004 and 400005

timtay-microsoft:
As per our recent guidance from the service team

### DIFF
--- a/iothub/service/src/Common/Exceptions/ErrorCode.cs
+++ b/iothub/service/src/Common/Exceptions/ErrorCode.cs
@@ -36,7 +36,21 @@ namespace Microsoft.Azure.Devices.Common.Exceptions
         DeviceInvalidResultCount = 400002,
 
         InvalidOperation = 400003,
+
+        /// <summary>
+        /// Returned by the service if a given method invocation request has an invalid argument in it.
+        /// <para>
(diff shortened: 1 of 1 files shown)


timtay-microsoft:
Looks like someone got this already, so closing this PR

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -36,7 +36,21 @@ public enum ErrorCode
         DeviceInvalidResultCount = 400002,
 
         InvalidOperation = 400003,
+
```
drwill-ms, line 39: Yes, but it looks like you have some info to improve upon it. I'd suggest reactivating.

Also, be sure to remove `en-us` path from URLs.

drwill-ms: Also, as a note I've been claiming items in the tracking spreadsheet. You might check there to avoid duplicate work.

iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -36,7 +36,21 @@ public enum ErrorCode
...
+        /// <summary>
+        /// Returned by the service if a given method invocation request has an invalid argument in it.
```
drwill-ms, line 41: I'm not sure this is the exclusive reason a client will see this, and even if it is right now, it may not be in the future. I suggest making it generic, and then noting a case why it might be hit. Most importantly, we should be telling people what to do if they see it, like in your para, but I'd like to start generic and then mention specific cases so it doesn't sound like it'll only happen in the specific case.

PR #2091, opened 2021-07-12
Document ArgumentNull

drwill-ms:
I couldn't find any real cases in the logs from the last 10 days, but it may be a very rare event.

### DIFF
--- a/iothub/service/src/Common/Exceptions/ErrorCode.cs
+++ b/iothub/service/src/Common/Exceptions/ErrorCode.cs
@@ -50,7 +50,12 @@ namespace Microsoft.Azure.Devices.Common.Exceptions
         // Note: one example found in logs is for invalid characters in a twin property name.
         ArgumentInvalid = 400004,
 
+        /// <summary>
+        /// Something in the payload is unexpectedly null. Check the error message for more information about what is invalid.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


iothub/service/src/Common/Exceptions/ErrorCode.cs
```diff
@@ -50,7 +50,12 @@ public enum ErrorCode
         // Note: one example found in logs is for invalid characters in a twin property name.
         ArgumentInvalid = 400004,
 
+        /// <summary>
+        /// Something in the payload is unexpectedly null. Check the error message for more information about what is invalid.
+        /// </summary>
+        // Note: an example suggest is null method payloads, but our client converts null to a JSON null, which is allowed.
```
vinagesh, line 56: suggested?

vinagesh (approved):

iothub/service/src/Common/Exceptions/ErrorCode.cs
timtay-microsoft: To use the cool trick James taught us:

```suggestion
        // Note: an example suggested is null method payloads, but our client converts null to a JSON null, which is allowed.
```

timtay-microsoft (approved):

PR #2094, opened 2021-07-12
fix(iot-service): Hide unreferenced error code

abhipsaMisra:


iothub/service/src/Common/TrackingHelper.cs
```diff
@@ -220,10 +220,6 @@ public static ErrorCode GetErrorCodeFromAmqpError(Error ex)
             {
                 return ErrorCode.DeviceNotFound;
             }
-            if (ex.Condition.Equals(IotHubAmqpErrorCode.MessageLockLostError))
-            {
-                return ErrorCode.DeviceMessageLockLost;
-            }
```
abhipsaMisra, line 226: this is dead code for a service application

timtay-microsoft (approved):

iothub/service/src/Common/TrackingHelper.cs
vinagesh: Agreed. However, it is public so just calling it out as a breaking change though in reality it should not be.

vinagesh (approved):

drwill-ms (approved):

PR #2096, opened 2021-07-12
Adding readme to explain authentication, retry and timeouts in DeviceClient

vinagesh:
## Checklist

## Description of the changes
Over the past few weeks, as I've been working through some GitHub issues and understanding the the DeviceClient more, I started adding a readme to close this request as part of a [GitHub issue](https://github.com/Azure/azure-iot-sdk-csharp/issues/1185). This is a very old request but should be very useful.

Please take a look and let me know if there is something that is missed, can be added or improved to make this document more useful to our customers.

### DIFF
--- a/DeviceConnectionAndReliabilityReadme.md
+++ b/DeviceConnectionAndReliabilityReadme.md
@@ -1,0 +1,85 @@
+# Azure IoT Device Client .NET SDK
+
+## Device connection and messaging reliability
+
+### Overview
+
+In this document you will find information about:
+
+- The connection authentication and renewal methods.
+- The reconnection logic and retry policies.
+- The timeout controls.
+
+### Connection authentication
+
+Authentication can be done using one of the following:
+
+- [SAS tokens for the device](https://docs.microsoft.com/azure/iot-hub/iot-hub-dev-guide-sas?tabs=node#use-sas-tokens-as-a-device) - Using IoT hub [device shared access key](https://docs.microsoft.com/azure/iot-hub/iot-hub-dev-guide-sas?tabs=node#use-a-shared-access-policy-to-access-on-behalf-of-a-device) or [symmetric key](https://docs.microsoft.com/azure/iot-hub/iot-hub-dev-guide-sas?tabs=node#use-a-symmetric-key-in-the-identity-registry) from DPS identity registry
+- [x509 certificate](https://docs.microsoft.com/azure/iot-hub/iot-hub-dev-guide-sas#supported-x509-certificates)  - Self signed or [CA-signed](https://docs.microsoft.com/azure/iot-hub/iot-hub-x509ca-overview)
+- [TPM based authentication](https://azure.microsoft.com/blog/device-provisioning-identity-attestation-with-tpm/)
+
+Samples:
+- IoT hub device shared access key based authentication sample - [DeviceReconnectionSample](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/DeviceReconnectionSample/DeviceReconnectionSample.cs#L102)
+- Device provisioning service symmetric key based authentication sample - [ProvisioningDeviceClientSample](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/provisioning/Samples/device/SymmetricKeySample/ProvisioningDeviceClientSample.cs#L62)
+- x509 based authentication sample using CA-signed certificates - [X509DeviceCertWithChainSample](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/X509DeviceCertWithChainSample/Program.cs#L43)
+- TPM based authentication sample - [ProvisioningDeviceClientSample](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/provisioning/Samples/device/TpmSample/ProvisioningDeviceClientSample.cs#L49)
+
+When using SAS tokens, authentication can be done by:
+
+- Providing the shared access key of the IoT hub and letting the SDK create the SAS tokens by using one of the `CreateFromConnectionString` methods on the [DeviceClient](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceclient).
+
+    If you choose this option, the SDK will create the SAS tokens and renew them before expiry. The default values for time-to-live and renewal buffer can be changed using the `ClientOptions` properties.
+
+  - [SasTokenTimeToLive](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.clientoptions.sastokentimetolive): The suggested time-to-live value for tokens generated for SAS authenticated clients. Default value is 60 minutes.
+  - [SasTokenRenewalBuffer](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.clientoptions.sastokenrenewalbuffer): The time buffer before expiry when the token should be renewed, expressed as a percentage of the time-to-live. Acceptable values lie between 0 and 100. Default value is 15%.  
+  
+  > Note: If the shared access policy name is not specified in the connection string, the audience for the token generation will be set by default to - `<iotHubHostName>/devices/<deviceId>`
+
+- Providing only the shared access signature
+
+    If you only provide the shared access signature, there will never be any renewal handled by the SDK.  
+  
+- Providing your own SAS token using [DeviceAuthenticationWithTokenRefresh](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithtokenrefresh)
+
+    If you choose to use `DeviceAuthenticationWithTokenRefresh` to provide your own implementation of token generation, you can provide the time-to-live and time buffer before expiry through the `DeviceAuthenticationWithTokenRefresh` constructor. The `ClientOptions` only apply to other `IAunthenticationMethod` implementations.
+
+When using x509 certificates, [DeviceAuthenticationWithX509Certificate](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithx509certificate) can be used. The client authentication will be valid until the certificate is valid. Any renewal will have to be done manually and the client needs to be recreated.
+
+When using TPM based authentication, the [DeviceAuthenticationWithTpm](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithtpm) can be used. TPM based authentication will eventually generate a SAS token but is more secure than using the shared access key of the IoT hub to generate the token.
+
+### Authentication methods implemented by the SDK
+
+The different `IAuthenticationMethod` implementations provided by the SDK are:
+
+- [DeviceAuthenticationWithRegistrySymmetricKey](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithregistrysymmetrickey) - Authentication method that uses the symmetric key associated with the device in the device registry.
+- [DeviceAuthenticationWithSharedAccessPolicyKey](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithsharedaccesspolicykey) - Authentication method that uses a shared access policy key.
+- [DeviceAuthenticationWithToken](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithtoken) - Authentication method that uses a shared access signature token.
+- [DeviceAuthenticationWithTokenRefresh](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithtokenrefresh) - Abstract class that can be implemented to generate a shared access signature token and allows for token refresh.
+- [DeviceAuthenticationWithTpm](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithtpm) - Authentication method that uses a shared access signature token generated using TPM and allows for token refresh.
+- [DeviceAuthenticationWithX509Certificate](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.deviceauthenticationwithx509certificate) - Authentication method that uses a X.509 certificates.
+
+### Connection retry logic
+
+For both AMQP and MQTT, the SDK will try to reconnect anytime there is any network related disruption. The default retry policy does not have a time limit and will follow exponential back-off.
+
+> Note: The default retry policy has support for jitter, which ensures that if you have N devices that disconnected at the same time, all of them won't start reconnecting with the same delay.
+
+For more details on the default retry policy and how to override it, see [retry policy documentation](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md).
+
+HTTP is a stateless protocol and will work whenever there is network connectivity.
+
+### Timeout controls
+
+There are different timeout values that can be configured for the `DeviceClient`/`ModuleClient` based on the protocol. These values are configuarable through the following transport settings that are passed while creating the client. Once the client is created, the settings cannot be changed. The client will need to be recreated with new settings to make changes.
+
+AMQP timeout settings:
+
+- [IdleTimeout](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.amqptransportsettings.idletimeout) - The interval that the client establishes with the service, for sending keep-alive pings. The default value is 2 minutes.
+- [OperationTimeout](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.amqptransportsettings.operationtimeout) - The time to wait for any operation to complete. The default is 1 minute.
+- [OpenTimeout](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.amqptransportsettings.opentimeout) - This value is not used (TODO: Confirm and update)
+
+MQTT timeout settings:
+
+- [ConnectArrivalTimeout](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.transport.mqtt.mqtttransportsettings.connectarrivaltimeout) - The time to wait for receiving an acknowledgment for a CONNECT packet. The default is 1 minute.
+- [KeepAliveInSeconds](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.transport.mqtt.mqtttransportsettings.keepaliveinseconds) - The interval, in seconds, that the client establishes with the service, for sending keep-alive pings. The default value is 5 minutes. The client will send a ping request 4 times per keep-alive duration set. It will wait for 30 seconds for the ping response, else mark the connection as disconnected.
+- [DeviceReceiveAckTimeout](https://docs.microsoft.com/dotnet/api/microsoft.azure.devices.client.transport.mqtt.mqtttransportsettings.devicereceiveacktimeout) -  The time a device will wait for an acknowledgment from service. The default is 5 minutes.
--- a/iothub/device/src/AmqpTransportSettings.cs
+++ b/iothub/device/src/AmqpTransportSettings.cs
@@ -115,12 +115,18 @@ namespace Microsoft.Azure.Devices.Client
 
         /// <summary>
         /// Specify client-side heartbeat interval.
+        /// The interval, that the client establishes with the service, for sending keep alive pings.
         /// The default value is 2 minutes.
         /// </summary>
+        /// <remarks>
+        /// The client will consider the connection as disconnected if the keep alive ping fails.
+        /// Setting a very low idle timeout value can cause aggressive reconnects, and might not give the
+        /// client enough time to establish a connection before disconnecting and reconnecting.
+        /// </remarks>
         public TimeSpan IdleTimeout { get; set; }
 
         /// <summary>
-        /// The operation timeout
+        /// The time to wait for any operation to complete. The default is 1 minute.
         /// </summary>
         public TimeSpan OperationTimeout
         {
@@ -129,8 +135,11 @@ namespace Microsoft.Azure.Devices.Client
         }
 
         /// <summary>
-        /// The open timeout
+        /// The open timeout. The default is 1 minute.
         /// </summary>
+        /// <remarks>
+        /// This property is currently unused.
+        /// </remarks>
         public TimeSpan OpenTimeout
         {
             get => _openTimeout;
@@ -172,7 +181,7 @@ namespace Microsoft.Azure.Devices.Client
         }
 
         /// <summary>
-        /// Returns the default current receive timeout
+        /// The time to wait for a receive operation. The default value is 1 minute.
         /// </summary>
         public TimeSpan DefaultReceiveTimeout => DefaultOperationTimeout;
 
--- a/iothub/device/src/Http1TransportSettings.cs
+++ b/iothub/device/src/Http1TransportSettings.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public sealed class Http1TransportSettings : ITransportSettings
     {
-        private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromSeconds(60);
+        private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromMinutes(1);
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Http1TransportSettings"/> class.
@@ -40,6 +40,9 @@ namespace Microsoft.Azure.Devices.Client
         /// <summary>
         /// The time to wait for a receive operation. The default value is 1 minute.
         /// </summary>
+        /// <remarks>
+        /// This property is currently unused.
+        /// </remarks>
         public TimeSpan DefaultReceiveTimeout => s_defaultOperationTimeout;
 
         /// <summary>
--- a/iothub/device/src/ITransportSettings.cs
+++ b/iothub/device/src/ITransportSettings.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Azure.Devices.Client
         TransportType GetTransportType();
 
         /// <summary>
-        /// The default receive timeout.
+        /// The time to wait for a receive operation.
         /// </summary>
         TimeSpan DefaultReceiveTimeout { get; }
     }
--- a/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         public bool DeviceReceiveAckCanTimeout { get; set; }
 
         /// <summary>
-        /// The time a device will wait, for an acknowledgment from service.
+        /// The time a device will wait for an acknowledgment from service.
         /// The default is 5 minutes.
         /// </summary>
         /// <remarks>
@@ -182,12 +182,14 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         public bool CleanSession { get; set; }
 
         /// <summary>
-        /// The interval, in seconds, that the client establishes with the service, for sending keep alive pings.
+        /// The interval, in seconds, that the client establishes with the service, for sending keep-alive pings.
         /// The default is 300 seconds.
         /// </summary>
         /// <remarks>
         /// The client will send a ping request 4 times per keep-alive duration set.
         /// It will wait for 30 seconds for the ping response, else mark the connection as disconnected.
+        /// Setting a very low keep-alive value can cause aggressive reconnects, and might not give the
+        /// client enough time to establish a connection before disconnecting and reconnecting.
         /// </remarks>
         public int KeepAliveInSeconds { get; set; }
 
--- a/readme.md
+++ b/readme.md
@@ -147,6 +147,7 @@ This repository contains [provisioning service client SDK](./provisioning/servic
 - [Set up your development environment](./doc/devbox_setup.md) to prepare your development environment as well as how to run the samples on Linux, Windows or other platforms.
 - [API reference documentation for .NET](https://docs.microsoft.com/dotnet/api/overview/azure/devices?view=azure-dotnet)
 - [Get Started with IoT Hub using .NET](https://docs.microsoft.com/azure/iot-hub/iot-hub-csharp-csharp-getstarted)
+- [Device connection and messaging reliability](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/DeviceConnectionAndReliabilityReadme.md)
 
 > Device Explorer is no longer supported. A replacement tool can be found [here](https://github.com/Azure/azure-iot-explorer).
 


iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -138,7 +144,7 @@ public TimeSpan OpenTimeout
         }
 
         /// <summary>
-        /// The pre-fetch count
+        /// The number of messages that the message receiver can simultaneously request. The default is 50.
```
vinagesh, line 147: This is what the official meaning is but I was not able to use it or figure out what this does in reality. Anyone has any info?

iothub/device/src/Http1TransportSettings.cs
```diff
@@ -40,6 +40,9 @@ public TransportType GetTransportType()
         /// <summary>
         /// The time to wait for a receive operation. The default value is 1 minute.
         /// </summary>
+        /// <remarks>
+        /// This property is currently unused.
+        /// </remarks>
         public TimeSpan DefaultReceiveTimeout => s_defaultOperationTimeout;
```
vinagesh, line 46: I could not find usages of this. Can someone else also confirm?

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Timeout controls
+
+There are different timeout values that can be configured for the DeviceClient based on the protocol. These values are configuarable through the following transport settings that are passed while creating the client. There will be no changes to any of the settings once the DeviceClient is created. The client needs to be recreated with new settings to make changes.
+
+AMQP transport settings - [AmqpTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/AmqpTransportSettings.cs)
+
+* IdleTimeout - The interval, in seconds, that the client establishes with the service, for sending keep alive pings. The default value is 2 minutes.
+* OperationTimeout - The time to wait for any operation to complete. The default is 1 minute.
+* OpenTimeout - This value is not used (TODO: Confirm and update)
```
vinagesh, line 57: When I looked at the implementation, it seemed like we use OperationTimeout for Open call as well. Let me know if this is not correct. Otherwise it would be worth updating that this value is unused by our SDK.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
+# Azure IoT Device Client C# SDK
```
drwill-ms, line 1: Prefer `.NET SDK`, as it can be used in applications that aren't C#.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Connection authentication
+
+Authentication can be done using either
+
+- [SAS tokens](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#security-tokens) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/TwinSample/Program.cs#L37)
+- [x509 certificates](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#supported-x509-certificates) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/X509DeviceCertWithChainSample/Program.cs#L43)
+- [Device Provisioning Service](https://docs.microsoft.com/en-us/azure/iot-dps) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/PnpDeviceSamples/Thermostat/Program.cs#L90)
+
+When using SAS tokens, authentication can be done by:
```
drwill-ms, line 21: Colon suffix used here, but not above when starting a list. I think the colon is nice, and would suggest it in all places.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+- Providing the shared access key and letting the SDK create the SAS tokens by using one of the [CreateFromConnectionString](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceClient.cs#L121) options on the DeviceClient or an authenticationMethod other than DeviceAuthenticationWithTokenRefresh.
+
+    If you choose this option, the SDK will create the SAS tokens and renew them before expiry. The default values for time to live and renewal buffer can be changed using the [ClientOptions](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/ClientOptions.cs) properties.
+    - SasTokenTimeToLive : The suggested time to live value for tokens generated for SAS authenticated clients. Default value is 60 minutes.
```
drwill-ms, line 26: When using `Label: value` notation, no space before colon

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+    - SasTokenRenewalBuffer : The time buffer before expiry when the token should be renewed, expressed as a percentage of the time to live. Acceptable values lie between 0 and 100.Default value is 15%.
```
drwill-ms, line 27: Space between `.` and `Default`

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+    Note: If the shared access policy name is not specified in the connection string, it will be set to - `<iotHubHostName>/devices/<deviceId>`
+>
```
drwill-ms, line 30: Why `>`?

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+    Note: If the shared access policy name is not specified in the connection string, it will be set to - `<iotHubHostName>/devices/<deviceId>`
+>  
+- Providing only the shared access signature
+
+    If you only provide the shared access signature and authenticationMethod is not of type AuthenticationWithTokenRefresh, there will be no automatic renewal.  
+>  
+- Providing your own SAS token using [DeviceAuthenticationWithTokenRefresh](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs) authenticationMethod
+
+    If you choose to use your own implementation to generate tokens, you can provide the time to live and time buffer before exipry through the DeviceAuthenticationWithTokenRefresh constructor. The ClientOptions apply only to other authentication methods.
```
drwill-ms, line 37: time-to-live

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+When using x509 certificates, [DeviceAuthenticationWithX509Certificate](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceAuthenticationWithX509Certificate.cs) authenticationMethod can be used. The client authentication will be valid until the certificate is valid. Any renewal will have to be done manually and the client needs to be recreated.
+
+### Connection retry logic
+
+For both AMQP and MQTT, the SDK will try to re-connect everytime there is any disruption. The default retry policy does not have a time limit and will follow exponential back-off.
```
drwill-ms, line 43: reconnect requires no hyphen

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+For more details, see [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md).
```
drwill-ms, line 45: Instead of [here], perhaps [retry policy documentation]?

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+HTTP is a stateless protocol and will work whenever there is network connectivity.
+
+### Timeout controls
+
+There are different timeout values that can be configured for the DeviceClient based on the protocol. These values are configuarable through the following transport settings that are passed while creating the client. There will be no changes to any of the settings once the DeviceClient is created. The client needs to be recreated with new settings to make changes.
+
+AMQP transport settings - [AmqpTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/AmqpTransportSettings.cs)
+
+* IdleTimeout - The interval, in seconds, that the client establishes with the service, for sending keep alive pings. The default value is 2 minutes.
```
drwill-ms, line 55: Elsewhere you use `-` for lists. Let's be consistent.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+* OperationTimeout - The time to wait for any operation to complete. The default is 1 minute.
```
drwill-ms, line 56: Also in seconds?

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+* OpenTimeout - This value is not used (TODO: Confirm and update)
+
+MQTT transport settings - [MqttTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs)
```
drwill-ms, line 59: Markdown linter suggests empty lines before and empty lists (which you've done above, but missed here).

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+There are different timeout values that can be configured for the DeviceClient based on the protocol. These values are configuarable through the following transport settings that are passed while creating the client. There will be no changes to any of the settings once the DeviceClient is created. The client needs to be recreated with new settings to make changes.
+
+AMQP transport settings - [AmqpTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/AmqpTransportSettings.cs)
+
+* IdleTimeout - The interval, in seconds, that the client establishes with the service, for sending keep alive pings. The default value is 2 minutes.
+* OperationTimeout - The time to wait for any operation to complete. The default is 1 minute.
+* OpenTimeout - This value is not used (TODO: Confirm and update)
+
+MQTT transport settings - [MqttTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs)  
+* ConnectArrivalTimeout - The time to wait for receiving an acknowledgment for a CONNECT packet. The default is 1 minute.
```
drwill-ms, line 60: Unit?

DeviceConnectionAndReliabilityReadme.md
drwill-ms: Oh, none of these are really in seconds, as they are `TimeSpan`s. No need to specify the unit unless the data type is integer, right?

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -115,12 +115,18 @@ public AmqpTransportSettings(TransportType transportType, uint prefetchCount, Am
 
         /// <summary>
         /// Specify client-side heartbeat interval.
+        /// The interval, in seconds, that the client establishes with the service, for sending keep alive pings.
```
drwill-ms, line 118: Last comma not needed
```suggestion
                /// The interval, in seconds, that the client establishes with the service for sending keep alive pings.
```

iothub/device/src/AmqpTransportSettings.cs
abhipsaMisra: iirc this was being corelated to the no of c2d messages a client queue could hold and receive at once. Since the c2d message queue cannot be greater than 50, I am surprised that this is user-configurable (maybe can be lowered but not increased beyond 50).

iothub/device/src/Http1TransportSettings.cs
abhipsaMisra: This is quite possibly code left over from previous design, when we used to only support amqp and thus did not have the other transport protocol specific settings.

iothub/device/src/ITransportSettings.cs
```diff
@@ -17,7 +17,7 @@ public interface ITransportSettings
         TransportType GetTransportType();
 
         /// <summary>
-        /// The default receive timeout.
+        /// The time to wait for a receive operation. The default value is 1 minute.
```
abhipsaMisra, line 20: I wouldn't add the comment about the default value here. The default isn't relevant for the interface but rather only for the individual implementations.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+- The connection authentication and renewal methods.
+- The reconnection logic and retry policies.
+- The timeout controls.
+
+### Connection authentication
+
+Authentication can be done using either
+
+- [SAS tokens](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#security-tokens) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/TwinSample/Program.cs#L37)
```
abhipsaMisra, line 17: would rather link to the reconnection sample here instead, since that is the one we would like to encourage customers to look at

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+- [x509 certificates](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#supported-x509-certificates) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/X509DeviceCertWithChainSample/Program.cs#L43)
+- [Device Provisioning Service](https://docs.microsoft.com/en-us/azure/iot-dps) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/PnpDeviceSamples/Thermostat/Program.cs#L90)
```
abhipsaMisra, line 19: I think the dps symmetric key sample would work better here: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/provisioning/Samples/device/SymmetricKeySample/ProvisioningDeviceClientSample.cs

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+In this document you will find information about
+
+- The connection authentication and renewal methods.
+- The reconnection logic and retry policies.
+- The timeout controls.
+
+### Connection authentication
+
+Authentication can be done using either
```
abhipsaMisra, line 15: even though TPM based devices eventually produce sas token, I think it is worth calling out those as a separate category  of authentication mechanism (TPM based auth is more secure and sas key based, and those are the ones customers are recommended to use in prod scenarios)

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+- [SAS tokens](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#security-tokens) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/TwinSample/Program.cs#L37)
+- [x509 certificates](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#supported-x509-certificates) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/X509DeviceCertWithChainSample/Program.cs#L43)
+- [Device Provisioning Service](https://docs.microsoft.com/en-us/azure/iot-dps) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/PnpDeviceSamples/Thermostat/Program.cs#L90)
+
+When using SAS tokens, authentication can be done by:
+
+- Providing the shared access key and letting the SDK create the SAS tokens by using one of the [CreateFromConnectionString](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceClient.cs#L121) options on the DeviceClient or an authenticationMethod other than DeviceAuthenticationWithTokenRefresh.
```
abhipsaMisra, line 23: how about we list the auth methods that they can use, instead of saying ..."or an authenticationMethod other than DeviceAuthenticationWithTokenRefresh.". That way we don't leave this ambiguous.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+    If you choose this option, the SDK will create the SAS tokens and renew them before expiry. The default values for time to live and renewal buffer can be changed using the [ClientOptions](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/ClientOptions.cs) properties.
+    - SasTokenTimeToLive : The suggested time to live value for tokens generated for SAS authenticated clients. Default value is 60 minutes.
+    - SasTokenRenewalBuffer : The time buffer before expiry when the token should be renewed, expressed as a percentage of the time to live. Acceptable values lie between 0 and 100.Default value is 15%.
+
+    Note: If the shared access policy name is not specified in the connection string, it will be set to - `<iotHubHostName>/devices/<deviceId>`
+>  
+- Providing only the shared access signature
+
+    If you only provide the shared access signature and authenticationMethod is not of type AuthenticationWithTokenRefresh, there will be no automatic renewal.
```
abhipsaMisra, line 33: If they provide the sas token directly, there will never be any renewal handled by the sdk.

DeviceConnectionAndReliabilityReadme.md
abhipsaMisra: any time there is a networking related disconnection

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+For both AMQP and MQTT, the SDK will try to re-connect everytime there is any disruption. The default retry policy does not have a time limit and will follow exponential back-off.
+
+For more details, see [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md).
+
+HTTP is a stateless protocol and will work whenever there is network connectivity.
+
+### Timeout controls
+
+There are different timeout values that can be configured for the DeviceClient based on the protocol. These values are configuarable through the following transport settings that are passed while creating the client. There will be no changes to any of the settings once the DeviceClient is created. The client needs to be recreated with new settings to make changes.
```
abhipsaMisra, line 51: DeviceClient/ModuleClient

DeviceConnectionAndReliabilityReadme.md
drwill-ms: Remove `en-us` path

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,62 @@
...
+
+AMQP transport settings - [AmqpTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/AmqpTransportSettings.cs)
+
+* IdleTimeout - The interval, in seconds, that the client establishes with the service, for sending keep alive pings. The default value is 2 minutes.
+* OperationTimeout - The time to wait for any operation to complete. The default is 1 minute.
+* OpenTimeout - This value is not used (TODO: Confirm and update)
+
+MQTT transport settings - [MqttTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs)  
+* ConnectArrivalTimeout - The time to wait for receiving an acknowledgment for a CONNECT packet. The default is 1 minute.
+* KeepAliveInSeconds - The interval, in seconds, that the client establishes with the service, for sending keep alive pings. The default value is 5 minutes.
```
abhipsaMisra, line 61: The pinf request is actually sent once every (300/4) seconds = 75 seconds: https://github.com/Azure/azure-iot-sdk-csharp/blob/c76a64de272da986b6840251f482249e094a725c/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs#L189
Whatever the ping request duration is, it will wait an additional 30 seconds for the ping response. The ping response timeout is not user-configurable.

iothub/device/src/AmqpTransportSettings.cs
vinagesh: Ok in that case let me check with the service team before updating the comment.

DeviceConnectionAndReliabilityReadme.md
vinagesh: Yes that is correct. They are all TimeSpan.

DeviceConnectionAndReliabilityReadme.md
vinagesh: This is also a TimeSpan so I did not specify unit.

iothub/device/src/AmqpTransportSettings.cs
vinagesh: yes this is a TimeSpan so your suggestion is good.

DeviceConnectionAndReliabilityReadme.md
vinagesh: I looked at the above sample. It is not the DeviceClient though. It is initializing the Provisioning Client. However, we want to be able to use the DeviceClient using DPS symmetric key right?

DeviceConnectionAndReliabilityReadme.md
vinagesh: Will add to the list

DeviceConnectionAndReliabilityReadme.md
vinagesh: I can add a section on the auth methods we provide implementation for. However, customers can have their own implementation as well so I did not say, these are the only ones you can use. What do you think?

DeviceConnectionAndReliabilityReadme.md
vinagesh: What about the IdleTimeout in AMQP? Does that work similarly too?

DeviceConnectionAndReliabilityReadme.md
abhipsaMisra: You do use the symmetric key based security provider to create the client instance: https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/d275140a8116d496e923ee3202ea639dff2a9ee4/provisioning/Samples/device/SymmetricKeySample/ProvisioningDeviceClientSample.cs#L62

DeviceConnectionAndReliabilityReadme.md
abhipsaMisra: The AMQP idle timeout is a bit more encapsulated within the amqp library itself. When we establish a timeout of 2mins, the service needs to send a ping request once every 2 mins. This value is monitored by the amqp library itself. If it doesn't receive the ping request, it initiates a disconnect.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+- [Device Provisioning Service](https://docs.microsoft.com/azure/iot-dps) - [Sample usage](https://github.com/Azure-Samples/azure-iot-samples-csharp/blob/master/iot-hub/Samples/device/PnpDeviceSamples/Thermostat/Program.cs#L90)
+
+When using SAS tokens, authentication can be done by:
+
+- Providing the shared access key and letting the SDK create the SAS tokens by using one of the [CreateFromConnectionString](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceClient.cs#L121) options on the DeviceClient or an authenticationMethod other than DeviceAuthenticationWithTokenRefresh.
+
+    If you choose this option, the SDK will create the SAS tokens and renew them before expiry. The default values for time-to-live and renewal buffer can be changed using the [ClientOptions](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/ClientOptions.cs) properties.
+    - SasTokenTimeToLive: The suggested time-to-live value for tokens generated for SAS authenticated clients. Default value is 60 minutes.
```
jamdavi, line 26: Some thoughts I have here are to augment the readme with links to the doc comments along side--or in some cases instead of--the code implementation in GitHub. The implementation is great for us and other folks trying to improve the client but others will see this doc as a getting started guide and might be more inclined to explore the API this way.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+    - SasTokenRenewalBuffer: The time buffer before expiry when the token should be renewed, expressed as a percentage of the time-to-live. Acceptable values lie between 0 and 100. Default value is 15%.
+
+    Note: If the shared access policy name is not specified in the connection string, it will be set to - `<iotHubHostName>/devices/<deviceId>`
+>  
+- Providing only the shared access signature
+
+    If you only provide the shared access signature and authenticationMethod is not of type AuthenticationWithTokenRefresh, there will never be any renewal handled by the sdk.
```
jamdavi, line 33:
```suggestion
    If you only provide the shared access signature and authenticationMethod is not of type AuthenticationWithTokenRefresh, the renewal will not be handled by the sdk.  
```

jamdavi: Note, I see **sdk** and **SDK** in some places in the doc, we should pick one.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+
+For both AMQP and MQTT, the SDK will try to reconnect anytime there is any network related disruption. The default retry policy does not have a time limit and will follow exponential back-off.
+
+For more details, see [retry policy documentation](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md).
+
+HTTP is a stateless protocol and will work whenever there is network connectivity.
+
+### Timeout controls
+
+There are different timeout values that can be configured for the DeviceClient/ModuleClient based on the protocol. These values are configuarable through the following transport settings that are passed while creating the client. There will be no changes to any of the settings once the DeviceClient is created. The client needs to be recreated with new settings to make changes.
```
jamdavi, line 51:
```suggestion
There are different timeout values that can be configured for the DeviceClient/ModuleClient based on the protocol. These values are configurable through the following transport settings that are passed while creating the client. Once the DeviceClient is created the settings cannot be changed. The client will need to be recreated with new settings to make changes.
```

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -115,12 +115,18 @@ public AmqpTransportSettings(TransportType transportType, uint prefetchCount, Am
 
         /// <summary>
         /// Specify client-side heartbeat interval.
+        /// The interval, that the client establishes with the service, for sending keep alive pings.
```
jamdavi, line 118:
```suggestion
        /// The interval that the client establishes with the service for sending keep alive pings.
```

DeviceConnectionAndReliabilityReadme.md
vinagesh: @abhipsaMisra - wanted to check on this

DeviceConnectionAndReliabilityReadme.md
vinagesh: Sure sounds like a good idea. I like the idea of pointing to MSDN docs better than pointing to our code. Is that what you mean?

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+
+In this document you will find information about:
+
+- The connection authentication and renewal methods.
+- The reconnection logic and retry policies.
+- The timeout controls.
+
+### Connection authentication
+
+Authentication can be done using either:
```
drwill-ms, line 15: The word "either" implies 2 options. A better choice would be "one of the following"

```suggestion
Authentication can be done using one of the following options:
```

DeviceConnectionAndReliabilityReadme.md
drwill-ms: Language ambiguity about whether the user or the SDK using `CreateFromConnectionString` or `DeviceAuthenticationWithTokenRefresh`.

Also, I don't know if `authenticationMethod` is meant to be a parameter. I changed it to plain text. If it is, then use backticks to render it as code.

```suggestion
- Providing the shared access key and letting the SDK create the SAS tokens. To do this use one of the [CreateFromConnectionString](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceClient.cs#L121) options on the `DeviceClient` or an authentication method other than DeviceAuthenticationWithTokenRefresh.
```

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+
+When using SAS tokens, authentication can be done by:
+
+- Providing the shared access key and letting the SDK create the SAS tokens by using one of the [CreateFromConnectionString](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceClient.cs#L121) options on the DeviceClient or an authenticationMethod other than DeviceAuthenticationWithTokenRefresh.
+
+    If you choose this option, the SDK will create the SAS tokens and renew them before expiry. The default values for time-to-live and renewal buffer can be changed using the [ClientOptions](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/ClientOptions.cs) properties.
+    - SasTokenTimeToLive: The suggested time-to-live value for tokens generated for SAS authenticated clients. Default value is 60 minutes.
+    - SasTokenRenewalBuffer: The time buffer before expiry when the token should be renewed, expressed as a percentage of the time-to-live. Acceptable values lie between 0 and 100. Default value is 15%.
+
+    Note: If the shared access policy name is not specified in the connection string, it will be set to - `<iotHubHostName>/devices/<deviceId>`
```
drwill-ms, line 29:
```suggestion
    > Note: If the shared access policy name is not specified in the connection string, it will be set to - `<iotHubHostName>/devices/<deviceId>`
```

DeviceConnectionAndReliabilityReadme.md
drwill-ms: Yep, acronyms in plain speech should generally be all caps.

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+>  
+- Providing only the shared access signature
+
+    If you only provide the shared access signature and authenticationMethod is not of type AuthenticationWithTokenRefresh, there will never be any renewal handled by the sdk.  
+>  
+- Providing your own SAS token using [DeviceAuthenticationWithTokenRefresh](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/DeviceAuthenticationWithTokenRefresh.cs) authenticationMethod
+
+    If you choose to use your own implementation to generate tokens, you can provide the time-to-live and time buffer before exipry through the DeviceAuthenticationWithTokenRefresh constructor. The ClientOptions apply only to other authentication methods.
```
drwill-ms, line 37:
```suggestion
    If you choose to use your own implementation to generate tokens, you can provide the time-to-live and time buffer before expiry through the `DeviceAuthenticationWithTokenRefresh` constructor. The `ClientOptions` apply only to other authentication methods.
```

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+
+HTTP is a stateless protocol and will work whenever there is network connectivity.
+
+### Timeout controls
+
+There are different timeout values that can be configured for the DeviceClient/ModuleClient based on the protocol. These values are configuarable through the following transport settings that are passed while creating the client. There will be no changes to any of the settings once the DeviceClient is created. The client needs to be recreated with new settings to make changes.
+
+AMQP transport settings - [AmqpTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/AmqpTransportSettings.cs):
+
+- IdleTimeout - The interval, in seconds, that the client establishes with the service, for sending keep alive pings. The default value is 2 minutes.
```
drwill-ms, line 55:
```suggestion
- IdleTimeout - The interval, in seconds, that the client establishes with the service, for sending keep-alive pings. The default value is 2 minutes.
```

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+- OperationTimeout - The time to wait for any operation to complete. The default is 1 minute.
+- OpenTimeout - This value is not used (TODO: Confirm and update)
+
+MQTT transport settings - [MqttTransportSettings](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs):
+
+- ConnectArrivalTimeout - The time to wait for receiving an acknowledgment for a CONNECT packet. The default is 1 minute.
+- KeepAliveInSeconds - The interval, in seconds, that the client establishes with the service, for sending keep alive pings. The default value is 5 minutes. The client will send a ping request 4 times per keep-alive duration set. It will wait for 30 seconds for the ping response, else mark the connection as disconnected.
```
drwill-ms, line 62:
```suggestion
- KeepAliveInSeconds - The interval, in seconds, that the client establishes with the service, for sending keep-alive pings. The default value is 5 minutes. The client will send a ping request 4 times per keep-alive duration set. It will wait for 30 seconds for the ping response, else mark the connection as disconnected.
```

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,63 @@
...
+- DeviceReceiveAckTimeout -  The time a device will wait, for an acknowledgment from service. The default is 5 minutes.
```
drwill-ms, line 63:
```suggestion
- DeviceReceiveAckTimeout -  The time a device will wait for an acknowledgment from service. The default is 5 minutes.
```

DeviceConnectionAndReliabilityReadme.md
vinagesh: That is basically the default audience to generated the shared access signature. It will determine the access level. I will update the statement to make it clear.

DeviceConnectionAndReliabilityReadme.md
jamdavi: Correct.

DeviceConnectionAndReliabilityReadme.md
vinagesh: Fixed all markdown linter suggestions

DeviceConnectionAndReliabilityReadme.md
vinagesh: Sure I can get onboard with that. I think it should matter if it is DeviceClient or some other client but we want to show the use of DeviceAuthenticationWithRegistrySymmetricKey.

DeviceConnectionAndReliabilityReadme.md
jamdavi: I think David means why do you have `>` on a line by itself?

DeviceConnectionAndReliabilityReadme.md
vinagesh: Made updates to the language and used back ticks to clarify.

DeviceConnectionAndReliabilityReadme.md
vinagesh: Chatting offline to improve this.

DeviceConnectionAndReliabilityReadme.md
drwill-ms: Yep, a greater-than-sign is for making a note. Why does one exist on a line by itself?

DeviceConnectionAndReliabilityReadme.md
```diff
@@ -0,0 +1,78 @@
...
+
+For both AMQP and MQTT, the SDK will try to reconnect anytime there is any network related disruption. The default retry policy does not have a time limit and will follow exponential back-off.
+
+For more details on the default retry policy and how to override it, see [retry policy documentation](https://github.com/Azure/azure-iot-sdk-csharp/blob/master/iothub/device/devdoc/retrypolicy.md).
+
+HTTP is a stateless protocol and will work whenever there is network connectivity.
+
+### Timeout controls
+
+There are different timeout values that can be configured for the `DeviceClient/ModuleClient` based on the protocol. These values are configuarable through the following transport settings that are passed while creating the client. Once the client is created, the settings cannot be changed. The client will need to be recreated with new settings to make changes.
```
drwill-ms, line 66:
```suggestion
There are different timeout values that can be configured for the `DeviceClient`/`ModuleClient` based on the protocol. These values are configurable through the following transport settings that are passed while creating the client. Once the client is created, the settings cannot be changed. The client will need to be recreated with new settings to make changes.
```

DeviceConnectionAndReliabilityReadme.md
vinagesh: Ahh it does not need to be there. It should been for the note but I messed up.

iothub/device/src/AmqpTransportSettings.cs
vinagesh: I have not heard back from the service team yet. So I'll leave the statement as it is and update in a subsequent PR.

drwill-ms (approved):

PR #2101, opened 2021-07-13
refactor(iot-service): Add API example for DeviceAlreadyExists error

abhipsaMisra:


common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -116,12 +117,56 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
...
+                    }
+                }
+                catch (JsonReaderException ex)
+                {
+                    if (Logging.IsEnabled)
+                        Logging.Error(null, $"Failed to deserialize error message into a dictionary: {ex}. Message body: '{responseContentStr}.'");
+
+                    // In some scenarios, the error response string is a ';' delimited string with the service-returned error code.
+                    const char errorFieldsDelimiter = ';';
+                    string[] messageFields = responseContent.Message.Split(errorFieldsDelimiter);
```
vinagesh, line 138: guard against null -> responseContent.Message?.Split

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -116,12 +117,56 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
...
+#if NET451 || NET472 || NETSTANDARD2_0
+                                if (field.IndexOf(CommonConstants.ErrorCode, StringComparison.OrdinalIgnoreCase) >= 0)
+#else
+                                if (field.Contains(CommonConstants.ErrorCode, StringComparison.OrdinalIgnoreCase))
+#endif
+                                {
+                                    const char errorCodeDelimiter = ':';
+                                    string[] errorCodeFields = field.Split(errorCodeDelimiter);
+
+                                    if (Enum.TryParse(errorCodeFields[1], out ErrorCode errorCode))
```
vinagesh, line 154: If there is no : in the string, it could lead to IndexOutOfRangeException. Check errorCodeFields.Length > 1 before accessing errorCodeFields[1]

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -116,12 +117,56 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
...
+
+                    if (messageFields != null)
+                    {
+                        _ = messageFields
+                            .Any(field =>
+                            {
+#if NET451 || NET472 || NETSTANDARD2_0
+                                if (field.IndexOf(CommonConstants.ErrorCode, StringComparison.OrdinalIgnoreCase) >= 0)
+#else
+                                if (field.Contains(CommonConstants.ErrorCode, StringComparison.OrdinalIgnoreCase))
```
vinagesh, line 148: Is contains not available in the other targets? Can we just not use IndexOf for all targets in that case?

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -116,12 +117,56 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
...
+                                {
+                                    const char errorCodeDelimiter = ':';
+                                    string[] errorCodeFields = field.Split(errorCodeDelimiter);
+
+                                    if (Enum.TryParse(errorCodeFields[1], out ErrorCode errorCode))
+                                    {
+                                        errorCodeValue = (int)errorCode;
+                                    }
+                                }
+                                return true;
```
jamdavi, line 159: Should this return a status code and not true??

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -116,12 +117,56 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
...
+                        Logging.Error(null, $"Failed to deserialize error message into a dictionary: {ex}. Message body: '{responseContentStr}.'");
+
+                    // In some scenarios, the error response string is a ';' delimited string with the service-returned error code.
+                    const char errorFieldsDelimiter = ';';
+                    string[] messageFields = responseContent.Message.Split(errorFieldsDelimiter);
+
+                    if (messageFields != null)
+                    {
+                        _ = messageFields
+                            .Any(field =>
```
vinagesh, line 143: Any reason for using Linq instead of foreach here? The result of .Any is not being so true/false does not matter but we always return it as true, which is a little confusing.

vinagesh (approved):

common/src/service/ExceptionHandlingHelper.cs
abhipsaMisra: This code will be executed only if there is an entry for "errorCode" in the message response, which is always a key-value pair. So the string will always have a ":". I can still add a a check though, not a problem.

common/src/service/ExceptionHandlingHelper.cs
abhipsaMisra: yeah, contains is available only on ns2.1+, and using indexOf for these newer frameworks generates https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2249, which I was surprised to see

common/src/service/ExceptionHandlingHelper.cs
abhipsaMisra: you're right, I wanted to change the return value based on if error code is present in inner response message or not, so I had chosen "Any". However, I changed my approach seeing that the default value for error code is already initialized to InvalidErrorCode. So using linq doesn't really lend any value here.

common/src/service/ExceptionHandlingHelper.cs
abhipsaMisra: No, this is the return value for the linq "Any" function, which is bool. We don't need that return value for our existing logic, so I've now removed this.

common/src/service/ExceptionHandlingHelper.cs
vinagesh: That's sad :( These compiler statements looks so messy.

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -116,12 +117,62 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
...
-                    && messageFields.TryGetValue(CommonConstants.ErrorCode, out string errorCodeObj))
+                try
                 {
-                    errorCodeValue = Convert.ToInt32(errorCodeObj, CultureInfo.InvariantCulture);
+                    Dictionary<string, string> messageFields = JsonConvert.DeserializeObject<Dictionary<string, string>>(responseContent.Message);
+
+                    if (messageFields != null
+                        && messageFields.TryGetValue(CommonConstants.ErrorCode, out string errorCodeObj))
+                    {
+                        errorCodeValue = Convert.ToInt32(errorCodeObj, CultureInfo.InvariantCulture);
```
drwill-ms, line 128: Curious if you considered `int.TryParse()` and if so, why you went with `Convert.ToInt32()`.

drwill-ms (approved):

common/src/service/ExceptionHandlingHelper.cs
abhipsaMisra: I didn't really inspect this code, this was already in place. I'm seeing that int.TryPrase is supposed to be faster, and will also not throw an exception; so that seems like a better choice here.

Issue #2067, opened 2021-07-07
[Bug Report] PnP APIs: TryGetValue should check for the component flag


rido-min:
Given the twin

```json
{
  "desired" : {
    "AComplexObject" : {
      "AProperty" : 23
    } 
  }
}
```

Trying to get the value as a component should fail if the marker `__t : 'c'` is not present
```csharp
class Program
{
    class AComplexObject
    {
        public double AProperty { get; set; }
    }

    static async Task Main(string[] args)
    {
        var mid = "dtmi:rido:simulator;1";
        var CS = Environment.GetEnvironmentVariable("CS");
        var dc = DeviceClient.CreateFromConnectionString(CS, TransportType.Mqtt, new ClientOptions { ModelId = mid });

        var desired = (await dc.GetClientPropertiesAsync()).Writable;
        if (desired.TryGetValue<AComplexObject>("AComplexObject", "AProperty", out AComplexObject aComplexObject))
        {
            Console.WriteLine("This should have fail, since AComplexObject is not a component");
        }
    }
}
```

abhipsaMisra:
This bug has been fixed and released in https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/preview_2021-10-8.

Issue #2066, opened 2021-07-07
[Bug Report] PnP APIs `TryGetValue` should not throw expected exceptions


rido-min:
The `Try..` pattern in .NET, like `int.TryParse`, is intended to do not throw exceptions. 

from the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=net-5.0):

> The TryParse method is like the Parse method, except the TryParse method does not throw an exception if the conversion fails. It eliminates the need to use exception handling to test for a FormatException in the event that s is invalid and cannot be successfully parsed.

reproes with #2065

abhipsaMisra:
This bug has been fixed and released in https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/preview_2021-10-8.

PR #2111, opened 2021-07-16
TryGetValue methods should check for the component identifier when applicable and should not throw any exceptions

abhipsaMisra:
Fix for #2066 , #2067

### DIFF
--- a/iothub/device/src/ClientPropertyCollection.cs
+++ b/iothub/device/src/ClientPropertyCollection.cs
@@ -201,7 +201,7 @@ namespace Microsoft.Azure.Devices.Client
                     // First verify that the retrieved dictionary contains the component identifier { "__t": "c" }.
                     // If not, then the retrieved nested dictionary is actually a root-level property of type map.
                     if (nestedDictionary.TryGetValue(ConventionBasedConstants.ComponentIdentifierKey, out object componentIdentifierValue)
-                        && componentIdentifierValue == ConventionBasedConstants.ComponentIdentifierValue)
+                        && componentIdentifierValue.ToString() == ConventionBasedConstants.ComponentIdentifierValue)
                     {
                         if (nestedDictionary.TryGetValue(propertyName, out object dictionaryElement))
                         {
--- a/iothub/device/src/ClientPropertyCollection.cs
+++ b/iothub/device/src/ClientPropertyCollection.cs
@@ -182,7 +182,7 @@ namespace Microsoft.Azure.Devices.Client
             }
 
             // If either the component name or the property name is null, empty or whitespace,
-            // then return unsuccessfully (false) with the default value of the type <T> passed in.
+            // then return false with the default value of the type <T> passed in.
             if (string.IsNullOrWhiteSpace(componentName) || string.IsNullOrWhiteSpace(propertyName))
             {
                 propertyValue = default;
@@ -201,11 +201,11 @@ namespace Microsoft.Azure.Devices.Client
                     // First verify that the retrieved dictionary contains the component identifier { "__t": "c" }.
                     // If not, then the retrieved nested dictionary is actually a root-level property of type map.
                     if (nestedDictionary.TryGetValue(ConventionBasedConstants.ComponentIdentifierKey, out object componentIdentifierValue)
-                        && componentIdentifierValue.Equals(ConventionBasedConstants.ComponentIdentifierValue))
+                        && componentIdentifierValue == ConventionBasedConstants.ComponentIdentifierValue)
                     {
                         if (nestedDictionary.TryGetValue(propertyName, out object dictionaryElement))
                         {
-                            // If the value associated with the key is null, then return successfully (true) with the default value of the type <T> passed in.
+                            // If the value associated with the key is null, then return true with the default value of the type <T> passed in.
                             if (dictionaryElement == null)
                             {
                                 propertyValue = default;
@@ -234,7 +234,7 @@ namespace Microsoft.Azure.Devices.Client
                         if (Convention
                             .PayloadSerializer
                             .TryGetNestedObjectValue(componentProperties, ConventionBasedConstants.ComponentIdentifierKey, out string componentIdentifierValue)
-                            && componentIdentifierValue.Equals(ConventionBasedConstants.ComponentIdentifierValue))
+                            && componentIdentifierValue == ConventionBasedConstants.ComponentIdentifierValue)
                         {
                             // Since the value cannot be cast to <T> directly, we need to try to convert it using the serializer.
                             // If it can be successfully converted, go ahead and return it.
@@ -245,7 +245,7 @@ namespace Microsoft.Azure.Devices.Client
                     catch
                     {
                         // In case the value cannot be converted using the serializer,
-                        // then return unsuccessfully (false) with the default value of the type <T> passed in.
+                        // then return false with the default value of the type <T> passed in.
                     }
                 }
             }
--- a/iothub/device/src/ClientPropertyCollection.cs
+++ b/iothub/device/src/ClientPropertyCollection.cs
@@ -206,9 +206,16 @@ namespace Microsoft.Azure.Devices.Client
                 }
                 else
                 {
-                    // If it's not, we need to try to convert it using the serializer.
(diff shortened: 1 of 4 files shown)


iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -170,8 +170,9 @@ public bool Contains(string componentName, string propertyName)
         /// <typeparam name="T">The type to cast the object to.</typeparam>
         /// <param name="componentName">The component which holds the required property.</param>
         /// <param name="propertyName">The property to get.</param>
-        /// <param name="propertyValue">The value of the component-level property.</param>
-        /// <returns>true if the property collection contains a component level property with the specified key; otherwise, false.</returns>
+        /// <param name="propertyValue">When this method returns successfully, this contains the value of the component-level property.
```
vinagesh, line 173: When this method returns ~~successfully/unsuccessfully~~ true/false ?

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -180,35 +181,72 @@ public virtual bool TryGetValue<T>(string componentName, string propertyName, ou
                     $"TryGetValue will attempt to get the property value but may not behave as expected.", nameof(TryGetValue));
             }
 
+            // If either the component name or the property name is null, empty or whitespace,
```
vinagesh, line 184: This is self explanatory in the code. Might want to say  - If the component name is not provided, return false with default value of type <T> passed in.

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -180,35 +181,72 @@ public virtual bool TryGetValue<T>(string componentName, string propertyName, ou
...
-                        }
-
-                        // If the object is of type T or can be cast to type T, go ahead and return it.
-                        if (dictionaryElement is T valueRef
-                            || NumericHelpers.TryCastNumericTo(dictionaryElement, out valueRef))
+                        if (nestedDictionary.TryGetValue(propertyName, out object dictionaryElement))
                         {
-                            propertyValue = valueRef;
-                            return true;
+                            // If the value associated with the key is null, then return successfully (true) with the default value of the type <T> passed in.
```
vinagesh, line 208: I'm not sure if you need to say successfully. Just saying true/false makes it less confusing as the return value is a bool.

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -180,35 +181,72 @@ public virtual bool TryGetValue<T>(string componentName, string propertyName, ou
...
+                // If the ClientPropertyCollection was constructed by the user application (eg. for updating the client properties)
+                // then the componentProperties are retrieved as a dictionary.
+                // The required property value can be fetched from the dictionary directly.
                 if (componentProperties is IDictionary<string, object> nestedDictionary)
                 {
-                    if (nestedDictionary.TryGetValue(propertyName, out object dictionaryElement))
+                    // First verify that the retrieved dictionary contains the component identifier { "__t": "c" }.
+                    // If not, then the retrieved nested dictionary is actually a root-level property of type map.
+                    if (nestedDictionary.TryGetValue(ConventionBasedConstants.ComponentIdentifierKey, out object componentIdentifierValue)
+                        && componentIdentifierValue.Equals(ConventionBasedConstants.ComponentIdentifierValue))
```
vinagesh, line 204: This seems like a string comparison, `componentIdentifierValue == ConventionBasedConstants.ComponentIdentifierValue` is preferred. And even if you choose to use .Equals, switch the order to avoid nullref exception if `componentIdentifierValue` is `null`. `ConventionBasedConstants.ComponentIdentifierValue.Equals(componentIdentifierValue)`

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -180,35 +181,72 @@ public virtual bool TryGetValue<T>(string componentName, string propertyName, ou
...
+                    // then the componentProperties are retrieved as the json object that is defined in the PayloadConvention.
+                    // The required property value then needs to be deserialized accordingly.
+                    try
+                    {
+                        // First verify that the retrieved dictionary contains the component identifier { "__t": "c" }.
+                        // If not, then the retrieved nested dictionary is actually a root-level property of type map.
+                        if (Convention
+                            .PayloadSerializer
+                            .TryGetNestedObjectValue(componentProperties, ConventionBasedConstants.ComponentIdentifierKey, out string componentIdentifierValue)
+                            && componentIdentifierValue.Equals(ConventionBasedConstants.ComponentIdentifierValue))
```
vinagesh, line 237: Preferably use == for string comparisons. Switch order if you use .Equals to avoid nullref exception.

iothub/device/tests/ClientPropertyCollectionTests.cs
```diff
@@ -160,24 +160,46 @@ public void ClientPropertyCollection_CanAddMultiplePropertyAndGetBackWithoutDevi
             outIntValue.Should().Be(IntPropertyValue);
         }
 
+        [TestMethod]
+        public void ClientPropertyCollection_TryGetValueShouldReturnFalseIfValueNotFound()
+        {
+            var clientProperties = new ClientPropertyCollection();
```
vinagesh, line 166: It is preferred to add //arrange, //act, //assert as comments appropriately in test code.

iothub/device/src/PayloadCollection.cs
```diff
@@ -107,13 +107,11 @@ public bool Contains(string key)
...
         /// Gets the value of the object from the collection.
         /// </summary>
-        /// <remarks>
-        /// This class is used for both sending and receiving properties for the device.
-        /// </remarks>
         /// <typeparam name="T">The type to cast the object to.</typeparam>
         /// <param name="key">The key of the property to get.</param>
-        /// <param name="value">The value of the object from the collection.</param>
-        /// <returns>True if the collection contains an element with the specified key; otherwise, it returns false.</returns>
+        /// <param name="value">When this method returns successfully, this contains the value of the object from the collection.
```
vinagesh, line 112: true/false instead of successfully/unsuccessfully?

iothub/device/src/ClientPropertyCollection.cs
abhipsaMisra: I've used the same language as used in Dictionary.TryGetValue: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=net-5.0

iothub/device/src/ClientPropertyCollection.cs
abhipsaMisra: Ah, I always keep on forgetting that string `==` works like value comparison in c#. I'll change this to `==`, sure.

iothub/device/src/PayloadCollection.cs
abhipsaMisra: I've used the same language as used in Dictionary.TryGetValue: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=net-5.0

iothub/device/tests/ClientPropertyCollectionTests.cs
```diff
@@ -255,10 +315,13 @@ public void ClientPropertyCollection_CanAddNullPropertyWithComponentAndGetBackWi
         [TestMethod]
         public void ClientPropertyCollection_CanAddMultiplePropertyWithComponentAndGetBackWithoutDeviceClient()
         {
+            // arrange
             var clientProperties = new ClientPropertyCollection();
             clientProperties.AddComponentProperty(ComponentName, StringPropertyName, StringPropertyValue);
             clientProperties.AddComponentProperty(ComponentName, IntPropertyName, IntPropertyValue);
 
+            // act, assert
+
```
jamdavi, line 324: nit: all of these act, assert comments have an extra line? Intentional?

jamdavi (approved):

iothub/device/tests/ClientPropertyCollectionTests.cs
abhipsaMisra: Yes, wherever we have a group of act-assert statements in a single test method, I've added the "act, assert" comment at the top and then left a newline to indicate that the comment is for the next set of statements, instead of being for the immediate following statement only.

vinagesh (approved):

PR #2121, opened 2021-07-22
refactor(e2e): fix one timeout test and disable another

timtay-microsoft:


drwill-ms (approved):

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -459,10 +459,17 @@ public async Task ProvisioningDeviceClient_ValidRegistrationId_TimeSpanTimeoutRe
             {
                 return; // expected exception was thrown, so exit the test
             }
+            catch (ProvisioningTransportException ex) when (ex.InnerException is OperationCanceledException)
+            {
+                // The provisioning transport layer will typically throw an OperationCanceledException, but some clock issues
+                // may cause this test to throw the OperationCanceledException nested in a ProvisioningTransportException instead.
+                return; // expected exception was thrown, so exit the test
+            }
```
timtay-microsoft, line 467: We have a check on the cancellation token that we create from a timespan as the first line in our provisioning transport handler for HTTP, but for some reason, it can sometimes pass that check even with Timespan.Zero. The next check on the timespan will throw an OperationCanceledException, but nested in a ProvisioningTransportException.

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -459,10 +459,17 @@ public async Task ProvisioningDeviceClient_ValidRegistrationId_TimeSpanTimeoutRe
...
 
             throw new AssertFailedException("Expected an OperationCanceledException to be thrown since the timeout was set to TimeSpan.Zero");
         }
 
+        [Ignore] // AMQP library itself is unreliable at returning the expected error code, and sometimes throws an ObjectDisposedException
```
timtay-microsoft, line 472: There is some threading at play since the AMQP library sets a timer based on the timespan we provide. Because the library does not behave consistently, I'm proposing we disable the test for now

e2e/test/provisioning/ProvisioningE2ETests.cs
abhipsaMisra: this is interesting, do you have a link to the code that you're referring to?

e2e/test/provisioning/ProvisioningE2ETests.cs
abhipsaMisra: is the plan to disable this test altogether or do we want to investigate this issue further?

abhipsaMisra: meaning, this being temporary so that our gates pass :)

e2e/test/provisioning/ProvisioningE2ETests.cs
timtay-microsoft: https://github.com/Azure/azure-amqp/blob/0c4cf8a15b7df2cbe155e59c26ea23cfb6d3f652/src/Transport/TcpTransportInitiator.cs#L38

e2e/test/provisioning/ProvisioningE2ETests.cs
timtay-microsoft: We can take this up as an issue with the AMQP library we use, but for now it should be disabled since it is flakey

e2e/test/provisioning/ProvisioningE2ETests.cs
abhipsaMisra: Can we add a check in our SDK to ensure TimeSpan.Zero always throws an OperationCancelledException. The reason I say this is, while these layered checks might be ok for our tests, they'll make an uncomfortable usage experience for our customers.

e2e/test/provisioning/ProvisioningE2ETests.cs
abhipsaMisra: Do you think marking it as flaky would be better than disabling it? (Flaky tests won't run on PR gates.)

e2e/test/provisioning/ProvisioningE2ETests.cs
timtay-microsoft: I can, but do you expect users to call these APIs with a timespan of 0?

e2e/test/provisioning/ProvisioningE2ETests.cs
timtay-microsoft: I'll defer to you on this one. I see this test as safe to disable since we likely won't mess up the timespan pass-down logic, but if you don't mind seeing the test fail occasionally on the nightly build then I'll mark it as flakey instead

e2e/test/provisioning/ProvisioningE2ETests.cs
timtay-microsoft: I get that a user wouldn't want to make the same catch statements that I made for this test, but adding a check for Timespan.Zero seems like the wrong fix for that issue. I'd rather the AMQP library itself be fixed to be more consistent with handling it

PR #2123, opened 2021-07-26
fix(tools): Fix readme for ConsoleEventListener usage

abhipsaMisra:
We no longer accept the event provider filter through `ConsoleEventListener` constructor. The readme is updated to reflect that.
I'm also renaming the test folder `Helpers` to `helpers` -> to follow the C# convention of using lower case folder names.


e2e/test/helpers/ConsoleEventListener.cs
```diff
@@ -10,16 +9,20 @@ namespace System.Diagnostics.Tracing
     public sealed class ConsoleEventListener : EventListener
     {
         // Configure this value to filter all the necessary events when OnEventSourceCreated is called.
-        // OnEventSourceCreated is triggered as soon as the EventListener is registered and an event source is created.
-        // So trying to configure this value in the ConsoleEventListener constructor does not work.
+        // The EventListener base class constructor creates an event listener in which all events start off turned off.
```
barustum, line 12: "start off turned off" sounds wrong.

abhipsaMisra: I thought so too, but it is the exact wording from the doc: https://github.com/dotnet/runtime/blob/6696065ab0f517f5a9e5f55c559df0010a816dbe/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L4009-L4018.
How about: "The EventListener base class constructor creates an event listener in which all events ~start off turned off~ are disabled by default."

barustum (approved):

abhipsaMisra:
I've removed the readme updates from this PR since it needs to be linked to some changes that are a part of this PR. Will submit a new PR to add those.

PR #2129, opened 2021-08-04
Target new DotNetty version when not net451

drwill-ms:


abhipsaMisra (approved):

provisioning/service/src/Config/IndividualEnrollment.cs
```diff
@@ -201,46 +198,13 @@ public override string ToString()
...
-        public string DeviceId
-        {
-            get
-            {
-                return _deviceId;
-            }
-
-            set
-            {
-                if (value == null)
```
drwill-ms, line 232: Not sure what this logic is. If it is null, then assign null, otherwise assign the value? Isn't that just the same as assigning the value?

With this custom logic removed, we can use an autoproperty.

Issue #2100, opened 2021-07-13
[feature request] PnP APIs: ClientProperties can provide better abstraction of Plug and Play convention from device twin implementation details


wickste:
### Issue
Some IoT Plug and Play developer tasks are still more complicated than they should be, and still require the developer to understand the inner workings of the device twin with respect to the IoT Plug and Play convention:

1. On device launch I need to know which properties have been updated while I was offline. The SDK is not helping with this at all. I need to go through all properties and compare version numbers on reported vs desired myself manually.
2. To ack property updates I need to create a whole new ClientPropertyCollection, create new properties, copy the value, set the status code and version. All things that are really device twin implementation details that the SDK could do for me. I would like to just receive the updated writable property and be able to call ack() on it – optionally provide a description and modify the value/status if needed.
3. Components are exposed as properties in the ClientPropertyCollection (device twin implementation detail!). To walk the collection and figure out the state of my properties I still need to know about device twin magic, like ‘__t’: ‘c’.

### Proposed Solution
ClientProperties should provide an abstraction over the device twin implementation details of the IoT Plug and Play convention. I would expect that ClientProperties understands the concept of components, and writeable properties (incl. the concept of ack'ing property updates).

abhipsaMisra:
> 1. On device launch I need to know which properties have been updated while I was offline. The SDK is not helping with this at all. I need to go through all properties and compare version numbers on reported vs desired myself manually.

This is design. Since IoT hub does not preserve twin update notifications, it becomes the client's responsibility to fetch the twin and ensure that update notifications weren't lost. Since the SDK does not hold state in any form, this responsibility now shifts to the calling application. I'll updating our [ReconnectionSample](https://github.com/Azure-Samples/azure-iot-samples-csharp/tree/master/iot-hub/Samples/device/DeviceReconnectionSample) to highlight this.
Adding this twin-sync logic to the SDK will also increase the bandwidth usage of the SDK since every reconnection attempt will then need to be followed by a call to retrieve the full twin. Since network traffic usage requirements and limitations vary per customer, we're leaving this logic to be implemented by the customer themselves, and are publishing a sample showing them how to do so.

> 2. To ack property updates I need to create a whole new ClientPropertyCollection, create new properties, copy the value, set the status code and version. All things that are really device twin implementation details that the SDK could do for me. I would like to just receive the updated writable property and be able to call ack() on it – optionally provide a description and modify the value/status if needed.

[...]

Issue #2097, opened 2021-07-13
[feature request] PnP APIs: Expose CreateWritablePropertyResponse directly on DeviceClient


wickste:
### Issue
Current API exposes the developer to the advanced concept of PayloadConvention and PayloadSerializer even when they don't use it explicitly. It also makes the CreateWritablePropertyResponse very hard to discover. There is no obvious reason why we are hiding this method two extra levels down. The DeviceClient should be able to hide that complexity and do the straight call-through

### With current API
```
IWritablePropertyResponse writablePropertyResponse = 
  deviceClient.PayloadConvention.PayloadSerializer.CreateWritablePropertyResponse(...);
```

### With proposed change
```
IWritablePropertyResponse writablePropertyResponse = 
  deviceClient.CreateWritablePropertyResponse(...);

```

abhipsaMisra:
We've introduced a [`WritableClientProperty`](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/pnpApi/iothub/device/src/WritableClientProperty.cs) that can be used to acknowledge the writable property update request. This will remove the need to having to discover/identify the appropriate `IWritablePropertyResponse` to format the property update response.

abhipsaMisra:
This has been released in https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/preview_2021-10-8.

Issue #2065, opened 2021-07-07
[Bug Report] PnP APIs. TryGetValue fails with ACK properties


rido-min:
Reported properties can use the ACK format (ac, av, ad, value). 

I expected `CloudProperty.TryGetValue` knows about that format and tries to deserialize the `value` if present.

Repro code

```csharp
using Microsoft.Azure.Devices.Client;
using Microsoft.Azure.Devices.Shared;
using System;
using System.Threading.Tasks;

namespace Rido.Samples
{
    
    class Program
    {
        static async Task Main(string[] args)
        {
            var mid = "dtmi:rido:simulator;1";
            var CS = Environment.GetEnvironmentVariable("CS");
            var dc = DeviceClient.CreateFromConnectionString(CS, TransportType.Mqtt, new ClientOptions { ModelId = mid });

            bool SendDataEnabled = true;

            //device reports the default value using the ack format
            IWritablePropertyResponse ack = dc.PayloadConvention.PayloadSerializer
                         .CreateWritablePropertyResponse(
                             SendDataEnabled, CommonClientResponseCodes.OK,1);

            var propertiesToBeUpdated = new ClientPropertyCollection();
            propertiesToBeUpdated.AddRootProperty("SendDataEnabled", ack);
            await dc.UpdateClientPropertiesAsync(propertiesToBeUpdated);

            var props = await dc.GetClientPropertiesAsync();
            // TryGetValue fails because the property is not bool
            if (props.TryGetValue<bool>("SendDataEnabled", out SendDataEnabled))
            {
                Console.WriteLine("Found reported SendDataEnabled: " + SendDataEnabled);
            } 
            else
            {
                Console.WriteLine("reported SendDataEnabled not found");
            }

            Console.WriteLine("Enter to Exit");
            Console.ReadLine();
        }
    }
}

```

abhipsaMisra:
@rido-min I think there is a mismatch in the expectations of calling `TryGetValue` on a writable property.
(1) To retrieve the value of service requested writable property update - `props.TryGetValue<bool>("SendDataEnabled", out SendDataEnabled)`
(2) To retrieve the value of device reported writable property value - `props.TryGetValue<NewtonsoftJsonWritablePropertyResponse>("SendDataEnabled", out SendDataEnabled)`

For (2), the expectation is that you would get the complete writable property response that is associated with that property (value, ac, av, ad (if present)). You can call `.Value` on it to fetch the actual value.

[...]

PR #2157, opened 2021-08-25
Add convenience method for acknowledging writable property update requests

abhipsaMisra:
Fix for #2065 , #2097 , #2100 

As highlighted in the above feature requests, constructing the _correct_ `IWritablePropertyResponse` for writable property update requests was not easily discoverable.
This PR aims to make the flow more convenient.

- On subscribing to `SubscribeToWritablePropertyUpdateRequestsAsync`, a client application will receive the writable property update requests as a collection of `WritableClientProperty` values.
- `WritableClientProperty` has a convenience method to "ack" these property update requests. 
- The "ack'ed" response is formatted in the format as required by IoT plug and play.

You can now do:

```csharp
await deviceClient
    .SubscribeToWritablePropertyUpdateRequestsAsync(
        async (writableProperties, userContext) =>
        {
            var writablePropertyAcks = new ClientPropertyCollection();
            foreach (KeyValuePair<string, object> writableProperty in writableProperties)
            {
                if (writableProperty.Value is WritableClientProperty rootLevelWritableClientProperty)
                {
                    string rootLevelPropertyName = writableProperty.Key;
                    writablePropertyAcks.Add(rootLevelPropertyName, rootLevelWritableClientProperty.AcknowledgeWith(CommonClientResponseCodes.OK));
                }
                else if (writableProperty.Value is IDictionary<string, object> componentProperties)
                {
                    string componentName = writableProperty.Key;
                    foreach (KeyValuePair<string, object> componentProperty in componentProperties)
                    {
                        if (componentProperty.Value is WritableClientProperty componentLevelwritableClientProperty)
                        {
                            string componentLevelPropertyName = componentProperty.Key;
                            writablePropertyAcks.AddComponentProperty(componentName, componentLevelPropertyName, componentLevelwritableClientProperty.AcknowledgeWith(CommonClientResponseCodes.OK));
                        }
                    }
                }
            }

            await deviceClient.UpdateClientPropertiesAsync(writablePropertyAcks).ConfigureAwait(false);
        },
        null,
        cts.Token)
    .ConfigureAwait(false);
```

Note: In earlier discussions we had contemplated updating:
```csharp
public Task SubscribeToWritablePropertyUpdateRequestsAsync(Func<ClientPropertyCollection, object, Task> callback, object userContext, CancellationToken cancellationToken = default);
```
to 
```csharp
public Task RespondToWritablePropertyUpdateRequestsAsync(Func<ClientPropertyCollection, object, Task<ClientPropertyCollection>> callback, object userContext, CancellationToken cancellationToken = default);
```

However, we decided against doing this because this would associate each writable property request with a response (reported property update) API call. Instead, we are formatting the response that the user can then add to a `ClientPropertyCollection` and issue the update call themselves.

### DIFF
--- a/iothub/device/src/WritableClientProperty.cs
+++ b/iothub/device/src/WritableClientProperty.cs
@@ -13,12 +13,18 @@ namespace Microsoft.Azure.Devices.Client
     /// This type contains a convenience method to format the reported property as per IoT Plug and Play convention.
     /// For more details see <see href="https://docs.microsoft.com/azure/iot-develop/concepts-convention#writable-properties"/>.
     /// </remarks>
-    public class WritableClientProperty
+    public class WritableClientProperty<T>
     {
+        private T _value;
+
         /// <summary>
         /// The value of the writable property update request.
         /// </summary>
-        public object Value { get; internal set; }
+        public T Value
+        {
+            get => Convention.PayloadSerializer.ConvertFromObject<T>(_value);
+            internal set => _value = value;
+        }
 
         /// <summary>
         /// The version number associated with the writable property update request.
--- a/iothub/device/src/WritableClientProperty.cs
+++ b/iothub/device/src/WritableClientProperty.cs
@@ -17,7 +17,6 @@ namespace Microsoft.Azure.Devices.Client
     {
         internal WritableClientProperty()
         {
-
         }
 
         /// <summary>
--- a/iothub/device/src/WritableClientProperty.cs
+++ b/iothub/device/src/WritableClientProperty.cs
@@ -13,18 +13,12 @@ namespace Microsoft.Azure.Devices.Client
     /// This type contains a convenience method to format the reported property as per IoT Plug and Play convention.
     /// For more details see <see href="https://docs.microsoft.com/azure/iot-develop/concepts-convention#writable-properties"/>.
     /// </remarks>
-    public class WritableClientProperty<T>
+    public class WritableClientProperty
     {
-        private T _value;
-
         /// <summary>
         /// The value of the writable property update request.
         /// </summary>
-        public T Value
-        {
-            get => Convention.PayloadSerializer.ConvertFromObject<T>(_value);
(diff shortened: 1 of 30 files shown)


iothub/device/src/WritableClientProperty.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// If you would like to construct your own writable property update response with custom value and version number, you can
+        /// create an instance of <see cref="PayloadSerializer.CreateWritablePropertyResponse(object, int, long, string)"/>.
+        /// See <see href="https://docs.microsoft.com/azure/iot-develop/concepts-convention#writable-properties"/> for more details.
+        /// </remarks>
+        /// <param name="statusCode">An acknowledgment code that uses an HTTP status code.</param>
+        /// <param name="description">An optional acknowledgment description.</param>
+        /// <returns>A writable property update response that can be reported back to the service.</returns>
+        public IWritablePropertyResponse AcknowledgeWith(int statusCode, string description = default)
+        {
+            return Convention.PayloadSerializer.CreateWritablePropertyResponse(Value, statusCode, Version, description);
```
BertKleewein, line 48: >Value [](http://example.com/codeflow?start=79&length=5)

This assumes that you're acknowledging the exact value that was sent by the service.  What if you want to respond with a value that is slightly off?  For exmple, the service sends a writable property request to set the temperature to 72.3 degrees, and the device app returns a WritablePropertyResponse that says "we set it to 72.2 degrees" because the hardware had to round down.  (I assume this is a valid scenario, but I'm assuming and you know what that means :))

abhipsaMisra: That's fair. We've included the following doc comment in the ack function which calls out how to achieve this:
https://github.com/Azure/azure-iot-sdk-csharp/blob/7cc63d2ac338c3527d21329d5fe5add50e80653b/iothub/device/src/WritableClientProperty.cs#L38-L41

PR #2161, opened 2021-09-02
fix(iot-device): Update MQTT layer to subscribe to topics with QoS of 1

abhipsaMisra:
Our MS docs call out that device SDKs should subscribe to topics with a QoS of 1 (at least once), but we had a couple of topics being subscribed to with QoS of 0 (at most once): https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-device-sdks

QoS of 0 = at most once = message delivery is attempted at most once. The publisher does not wait for an ack, meaning that the message delivery is not ack'ed or guaranteed. This is also called "fire and forget".
QoS of 1 = at least once = message delivery is attempted and the publisher waits for an ack from the subscriber. This QoS guarantees delivery of message at least once. 

Nice write-up about QoS: https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels/

The impact of this change will be that our messaging should be more reliable (prevent loss of messages), albeit a _bit_ slower (since calls will now wait for an ack rather than completing immediately).

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -823,7 +823,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_001:  `EnableMethodsAsync` shall subscribe using the '$iothub/methods/POST/' topic filter.
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_002:  `EnableMethodsAsync` shall wait for a SUBACK for the subscription request.
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_003:  `EnableMethodsAsync` shall return failure if the subscription request fails.
-            await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(MethodPostTopicFilter, QualityOfService.AtMostOnce))).ConfigureAwait(true);
+            await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(MethodPostTopicFilter, QualityOfService.AtLeastOnce))).ConfigureAwait(true);
         }
 
         public override async Task DisableMethodsAsync(CancellationToken cancellationToken)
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -823,7 +823,7 @@ public override async Task EnableMethodsAsync(CancellationToken cancellationToke
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_001:  `EnableMethodsAsync` shall subscribe using the '$iothub/methods/POST/' topic filter.
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_002:  `EnableMethodsAsync` shall wait for a SUBACK for the subscription request.
             // Codes_SRS_CSHARP_MQTT_TRANSPORT_18_003:  `EnableMethodsAsync` shall return failure if the subscription request fails.
-            await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(MethodPostTopicFilter, QualityOfService.AtMostOnce))).ConfigureAwait(true);
+            await _channel.WriteAsync(new SubscribePacket(0, new SubscriptionRequest(MethodPostTopicFilter, QualityOfService.AtLeastOnce))).ConfigureAwait(true);
```
drwill-ms, line 826: Should we never use `QualityOfService.AtMostOnce` in our SDK? Should we put a comment on that option to discourage it (or is that type from the MQTT library)?

drwill-ms (approved):

drwill-ms:
> Our MS docs call out that device SDKs should subscribe to topics with a QoS of 1 (at least once), but we had a couple of topics being subscribed to with QoS of 0 (at most once): https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-device-sdks
> 
> QoS of 0 = at most once = message delivery is attempted at most once. The publisher does not wait for an ack, meaning that the message delivery is not ack'ed or guaranteed. This is also called "fire and forget".
> QoS of 1 = at least once = message delivery is attempted and the publisher waits for an ack from the subscriber. This QoS guarantees delivery of message at least once.
> 
> Nice write-up about QoS: https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels/
> 
> The impact of this change will be that our messaging should be more reliable (prevent loss of messages), albeit a _bit_ slower (since calls will now wait for an ack rather than completing immediately).

In terms of possibly change of behavior, does this mean messages could now be delivered more than once where that wouldn't have happened before? You already mentioned a possible perf impact. Do you think this deserves a conversation at office hours with Lorenzo and Stefan?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: I was thinking along the same lines that we should always be avoiding QoS of 0, but turns out we do still publish messages with QoS of 0. I'm digging into it deeper to see if this should be changed as well (us publishing messages with QoS 1 will require that service ack those messages).

abhipsaMisra:
> In terms of possibly change of behavior, does this mean messages could now be delivered more than once where that wouldn't have happened before? You already mentioned a possible perf impact. Do you think this deserves a conversation at office hours with Lorenzo and Stefan?

Good point.
Yes, you could definitely have some messages that are delivered more than once to the SDK. Now, if these messages make it to the calling application multiple times or not depends on how we've implemented the processing logic; i.e. if there is a sequence to processing these packets, i.e. are they first ack'ed and then processed by the SDK or if they are independent tasks that are executed as soon as we receive the packet. I'll confirm this from our code and we can then assess if client application de-duplication logic needs to be implemented (we can then discuss this with Lorenzo and Stefan as well).

abhipsaMisra:
Closing this PR for now.
Twin updates, even if subscribed to with a QoS of 1, are always published with a QoS of 0. Hence updating our SDK's subscription QoS won't change the behavior.

Will follow-up with a task to get the documentation updated instead.

PR #2164, opened 2021-09-07
fix(client): Adding graceful shutdown to event loop

### LINKED ISSUE #2163: [Bug Report] selecting MQTT as DeviceClient transport type causes WPF app to hangs on close?
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64
- **Application's .NET Target Framework :** net5 + WPF
- **Device:** Laptop
- **SDK version used:** 1.38.0

## Description of the issue
I've stumbled upon a strange issue when creating and using `DeviceClient.Create` in WPF app with `Transport.Mqtt`.
I'm sending a message to IoT Hub and when closing the main window, the application just hangs and never terminates.
When I select `TransportType.Amqp` this hang does not occur and the application terminates correctly.

Create a simple WPF application, add two buttons in XAML:

``` xaml
<Window x:Class="AzIoTDev.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:AzIoTDev"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800">
    <StackPanel Orientation="Horizontal" VerticalAlignment="Top">
        <Button Content="Send data MQTT" Click="Button_Click"/>
        <Button Content="Send data AMQP" Click="Button_Click_1" />
    </StackPanel>
</Window>
```

then in codebehind set this code (set your own credentials for device):

``` csharp
public partial class MainWindow : Window
	{
		public MainWindow()
		{
			InitializeComponent();
		}

		private const string DeviceId = "IoT hub device ID";
		private const string AuthKey = "IoT hub auth key for device";
		private const string ServerUrl = "IoT hub URL";
		private const int OperationTimeoutMs = 1000;

		private DeviceClient _azureClient;

		private async Task sendHello(TransportType tt)
		{
			var auth = new DeviceAuthenticationWithRegistrySymmetricKey(DeviceId, AuthKey);
			_azureClient = DeviceClient.Create(ServerUrl, auth, tt);
			_azureClient.OperationTimeoutInMilliseconds = OperationTimeoutMs;
			
			byte[] bytes = Encoding.UTF8.GetBytes("HelloPayload");
			Message eventMessage = new Message(bytes);
			eventMessage.Properties.Add("MessageType", "Hello");
			await _azureClient.SendEventAsync(eventMessage).ConfigureAwait(false);

			await _azureClient.CloseAsync().ConfigureAwait(false);
			_azureClient.Dispose();
		}

		private async void Button_Click(object sender, RoutedEventArgs e)
		{
			await sendHello(TransportType.Mqtt);
		}

		private async void Button_Click_1(object sender, RoutedEventArgs e)
		{
			await sendHello(TransportType.Amqp);
		}
	}
```
You will see that clicking "Send with MQTT" will hang the app when closing and "Send with AQMP" will not hang when closing.
Is there an issue or am I doing something wrong?

all the best
Krzysztof

jamdavi:
When we create an Mqtt transport handler we set the EventLoopGroup but we do not gracefully shut it down. This leaves a number of TaskSchedulers open. 

This is specific to DotNetty (well Netty really) and does not affect Amqp.

In most cases this will not cause a problem if the application closes after client shutdown, but in WPF applications it will cause the main UI thread to hang and never close.

By default this shutdown will take 15 seconds to complete. 

This is documented in the [Java Netty](https://netty.io/wiki/user-guide-for-4.x.html#shutting-down-your-application) wiki.

This will address #2163 and #2194


timtay-microsoft (approved):

abhipsaMisra (approved):

jamdavi:
Also relevant to #2194

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -250,5 +250,13 @@ public TransportType GetTransportType()
...
         /// </summary>
         internal string AuthenticationChain { get; set; }
+
+        /// <summary>
+        /// The DotNetty event loop GracefulShutdown timeout.
+        /// </summary>
+        /// <remarks>
+        /// The DotNetty Mqtt transport has a graceful shutdown of the event loop that can block the <see cref="DeviceClient.CloseAsync(System.Threading.CancellationToken)"/> method. This timeout allows you to configure how long the shutdown event will wait for before forcefully shutting down the event loop. Change this if you need your application to respond to the CloseAsync command faster.
+        /// </remarks>
+        public TimeSpan GracefulEventLoopShutdownTimeout { get; set; } = TimeSpan.FromMilliseconds(1);
```
drwill-ms, line 260: Why 1 millisecond default?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -493,6 +493,8 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
                     OnTransportClosedGracefully();
 
                     await _closeRetryPolicy.ExecuteAsync(CleanUpImplAsync, cancellationToken).ConfigureAwait(true);
+                    var timeout = ((MqttTransportSettings)_transportSettings).GracefulEventLoopShutdownTimeout;
```
drwill-ms, line 496: Any unit tests you can add to validate?

jamdavi: No unit tests, but I can add a E2E test that checks to see if the timeout is honored when disposing the object.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -94,7 +94,7 @@ internal sealed class MqttTransportHandler : TransportHandler, IMqttIotHubEventH
         private const string ReceiveEventMessagePrefixPattern = "devices/{0}/modules/{1}/";
 
         private static readonly int s_generationPrefixLength = Guid.NewGuid().ToString().Length;
-        private static readonly Lazy<IEventLoopGroup> s_eventLoopGroup = new Lazy<IEventLoopGroup>(GetEventLoopGroup);
+        private readonly IEventLoopGroup _eventLoopGroup = GetEventLoopGroup();
```
jamdavi, line 97: @drwill-ms, @timtay-microsoft, @azabbasi, @abhipsaMisra Does anyone know why this was originally static? Was it to set an upper bound of threads in the event of multiple clients/transports?

e2e/test/iothub/MqttTransportHandlerShutdownTests.cs
```diff
@@ -0,0 +1,128 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
jamdavi, line 1: Pay attention to the timeout and the "close to" amounts. The idea was to give enough wiggle room but take into account of a busy thread scheduler.

These values produced the most consistent results with my testing.

e2e/test/iothub/MqttTransportHandlerShutdownTests.cs
```diff
@@ -0,0 +1,128 @@
...
+            Logger.Trace($"{nameof(MqttTransport_Tcp_DefaultTimeout_OnCloseAsync)}: deviceId={testDevice.Id}");
+            Logger.Trace($"{nameof(MqttTransport_Tcp_DefaultTimeout_OnCloseAsync)}: calling OpenAsync...");
+            await deviceClient.OpenAsync().ConfigureAwait(false);
+            Logger.Trace($"{nameof(MqttTransport_Tcp_DefaultTimeout_OnCloseAsync)}: calling CloseAsync...");
+            var sw = Stopwatch.StartNew();
+            await deviceClient.CloseAsync();
+            sw.Stop();
+            sw.Elapsed.Should().BeCloseTo(TimeSpan.FromSeconds(1), 200);
+        }
+
```
jamdavi, line 126:
```suggestion
```

e2e/test/iothub/MqttTransportHandlerShutdownTests.cs
```diff
@@ -0,0 +1,127 @@
...
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    public class MqttTransportHandlerShutdownTests : E2EMsTestBase
+    {
+        private static readonly string _devicePrefix = $"E2E_{nameof(MqttTransportHandlerShutdownTests)}_";
+
+        [LoggedTestMethod]
+        public async Task MqttTransport_Tcp_LongTimeout_OnCloseAsync()
+        {
+            using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix, TestDeviceType.Sasl).ConfigureAwait(false);
```
drwill-ms, line 25: Would you mind adding these template comment?

```csharp
// arrange

// act

// assert
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: Sorry, no idea. Frankly, I find the notion of a static thing in our SDK slightly frightening.

drwill-ms: Why not lazy anymore?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -493,6 +515,17 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
                     OnTransportClosedGracefully();
 
                     await _closeRetryPolicy.ExecuteAsync(CleanUpImplAsync, cancellationToken).ConfigureAwait(true);
+                    // If the customer chooses multiple groups we need to honor that here.
+                    if (_useSingleEventLoopGroup)
+                    {
+                        // The value here is likely one that won't change often. And in the most common case there is usually only a single device that
+                        // is created per process so we shouldn't have conflicting expectations.
```
jamdavi, line 522:
```suggestion
```

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -250,5 +250,13 @@ public TransportType GetTransportType()
...
         /// </summary>
         internal string AuthenticationChain { get; set; }
+
+        /// <summary>
+        /// The DotNetty event loop GracefulShutdown timeout.
+        /// </summary>
+        /// <remarks>
+        /// The DotNetty Mqtt transport has a graceful shutdown of the event loop that can block the <see cref="DeviceClient.CloseAsync(System.Threading.CancellationToken)"/> method. This timeout allows you to configure how long the shutdown event will wait for before forcefully shutting down the event loop. Change this if you need your application to respond to the CloseAsync command faster.
+        /// </remarks>
+        public TimeSpan GracefulEventLoopShutdownTimeout { get; set; } = TimeSpan.FromSeconds(1);
```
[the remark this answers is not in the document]
jamdavi: So by default the timeout value is 14 seconds... I can make this a TimeSpan.Zero and then let the customer decide if they want to lower it.

shared/src/ReferenceCounter.cs
```diff
@@ -0,0 +1,174 @@
...
+                return _obejctToCount;
+            }
+        }
+
+        /// <summary>
+        /// Sets the internal object to null, clears the reference counter, and if <see cref="CreateWithRemoveAction"/> was used it will execute the removal function defined.
+        /// </summary>
+        /// <remarks>
+        /// This will only execute if there is atleast one reference count.
+        /// Take care to properly tear down your object. It is best to use the <see cref="CreateWithRemoveAction(Func{T}, Action{T})"/> method to create your reference counted object. The action will be called />
```
jamdavi, line 123:
```suggestion
        /// Take care to properly tear down your object. It is best to use the <see cref="CreateWithRemoveAction(Func{T}, Action{T})"/> method to create your reference counted object.
```

shared/src/ReferenceCounter.cs
```diff
@@ -0,0 +1,174 @@
...
+        ///     
+        ///     public ContainerUsage()
+        ///     {
+        ///         _refObject.Create(() => new FileStream(), (teardown) => teardown.Dispose());
+        ///         _refObject.Clear();
+        ///     }
+        /// }
+        /// </code>
+        /// </example>
+        /// </remarks>
```
jamdavi, line 137:
```suggestion
```

shared/src/ReferenceCounter.cs
```diff
@@ -0,0 +1,174 @@
...
+                    if (_executeOnDispose != null)
+                    {
+                        _executeOnDispose(_obejctToCount);
+                    }
+                    _obejctToCount = null;
+                    _referenceCount = 0;
+                }
+            }
+        }
+
```
jamdavi, line 158:
```suggestion
```

shared/src/ReferenceCounter.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                CreateWithoutLocking(objectToCreate);
+                Interlocked.Increment(ref _referenceCount);
+                return _obejctToCount;
+            }
+        }
+
+        private void CreateWithoutLocking(Func<T> objectToCreate)
+        {
+            if (_obejctToCount == null && _referenceCount <= 0)
```
azabbasi, line 71: under what circumstances does `_referenceCount` go below zero?

azabbasi (approved):

shared/src/ReferenceCounter.cs
jamdavi: None, now that think of it.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -94,9 +94,14 @@ internal sealed class MqttTransportHandler : TransportHandler, IMqttIotHubEventH
...
         private static readonly int s_generationPrefixLength = Guid.NewGuid().ToString().Length;
-        private static readonly Lazy<IEventLoopGroup> s_eventLoopGroup = new Lazy<IEventLoopGroup>(GetEventLoopGroup);
+        private static readonly ReferenceCounter<IEventLoopGroup> _eventLoopGroupRefCounted = new ReferenceCounter<IEventLoopGroup>();
         private static readonly TimeSpan s_regexTimeoutMilliseconds = TimeSpan.FromMilliseconds(500);
         private static readonly TimeSpan s_defaultTwinTimeout = TimeSpan.FromSeconds(60);
+        private static TimeSpan s_timeoutValueForEventLoop = TimeSpan.Zero;
+
+        private readonly bool _useSingleEventLoopGroup;
+        private readonly IEventLoopGroup _eventLoopGroup;
+        private readonly TimeSpan _timeoutValueForEventLoop = TimeSpan.Zero;
```
abhipsaMisra, line 104: Do we need both `s_timeoutValueForEventLoop` and `_timeoutValueForEventLoop`, or is one of them a duplicate? If we need both then maybe we should rename one of them.

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -250,5 +250,21 @@ public TransportType GetTransportType()
...
+
+        /// <summary>
+        /// Sets the MqttTransport to use a single Event Loop Group.
+        /// </summary>
+        /// <remarks>
+        /// This setting is to retain legacy behavior that will use a single Event Loop Group which is the thread pool model for DotNetty. If this is set to false a new group will be created per instance of the Device Client.
```
abhipsaMisra, line 266: We have 1 instance of mqtt adapter per device client, and we can have 1 instance of device client for a particular device identity. Considering that, what does this mean:
"If this is set to false a new group will be created per instance of the Device Client.", i.e. does this refer to the scenario where we dispose a client instance and reinitialize?

shared/src/ReferenceCounter.cs
```diff
@@ -0,0 +1,217 @@
...
+    ///     {
+    ///         _refObject.Create(() => new object());
+    ///     }
+    /// }
+    /// </code>
+    /// </example>
+    /// <typeparam name="T">Any class based object type.</typeparam>
+    public class ReferenceCounter<T> where T : class
+    {
+        private T _obejctToCount;
```
abhipsaMisra, line 28: nit: typo on object

shared/src/ReferenceCounter.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        /// <summary>
+        /// Create the object or increment the counter.
+        /// </summary>
+        /// <remarks>
+        /// If the object is already set this will not execute the code in <paramref name="objectToCreate"/>. Instead it will return the object that was previously created.
+        /// 
+        /// 
+        /// 
+        /// This will return a mutable object. It is possible to alter the object state which could cause unexpected behavior.
```
abhipsaMisra, line 54: q - What kind of unexpected behavior are we referencing here?

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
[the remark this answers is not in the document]
abhipsaMisra: Same questions about he timeout values selected. It would be helpful to add a reference to the netty specs which define these timeout values (default/recommended etc).

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
jamdavi: No, I simplified the execution path.

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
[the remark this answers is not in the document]
jamdavi: I can do that.

shared/src/ReferenceCounter.cs
jamdavi: In general if you have an object that is reference counted there is no way to determine if the object has been changed. For example, an object could be disposed of while there is still a reference count of 1 but we'd have no way to tell and we'd still return the reference to the disposed object.

jamdavi:
Closing this as the following fix will be released by DotNetty

https://github.com/Azure/DotNetty/pull/570

PR #2171, opened 2021-09-17
[PnP] Update PayloadCollection to accept dictionary of values

abhipsaMisra:
This update allows users to pass in a collection of values for TelemetryPayload and PropertyPayload.

While we currently don't have an implementation to accept raw classes as payload values, it can easily be achieved by serializing the raw class and deserializing it into a dictionary.

We'll can add support for it if customers ask for it.

### DIFF
--- a/iothub/device/src/ClientPropertyCollection.cs
+++ b/iothub/device/src/ClientPropertyCollection.cs
@@ -44,8 +44,16 @@ namespace Microsoft.Azure.Devices.Client
         /// <param name="componentName">The component with the property to add.</param>
         /// <param name="propertyName">The name of the property to add.</param>
         /// <param name="propertyValue">The value of the property to add.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="componentName"/> or <paramref name="propertyName"/> is <c>null</c>.</exception>
         public void AddComponentProperty(string componentName, string propertyName, object propertyValue)
-            => AddInternal(new Dictionary<string, object> { { propertyName, propertyValue } }, componentName, false);
+        {
+            if (componentName == null)
+            {
+                throw new ArgumentNullException(nameof(componentName));
+            }
+
+            AddInternal(new Dictionary<string, object> { { propertyName, propertyValue } }, componentName, false);
+        }
 
         /// <inheritdoc path="/remarks" cref="AddRootProperty(string, object)" />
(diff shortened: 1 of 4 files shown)


iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -55,8 +63,16 @@ public void AddComponentProperty(string componentName, string propertyName, obje
...
+        /// <exception cref="ArgumentNullException"><paramref name="componentName"/> or a property name in <paramref name="properties"/> is <c>null</c>.</exception>
         public void AddComponentProperties(string componentName, IDictionary<string, object> properties)
-            => AddInternal(properties, componentName, true);
+        {
+            if (componentName == null)
+            {
+                throw new ArgumentNullException(nameof(componentName));
+            }
+
+            AddInternal(properties, componentName, false);
```
timtay-microsoft, line 74: Should the final parameter here be ```true```? If not, why is it being changed from its previous behavior here?

timtay-microsoft (approved):

iothub/device/src/ClientPropertyCollection.cs
abhipsaMisra: The final parameter here determines if the operation is a "force" update. This is set to false for the "Add" APIs and is set to true for the "AddOrUpdate" APIs.
This was incorrectly set to true previously.

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -442,6 +480,12 @@ private void AddInternal(IDictionary<string, object> properties, string componen
 
                 foreach (KeyValuePair<string, object> entry in properties)
                 {
+                    // A null property key is not allowed. Throw an ArgumentNullException.
+                    if (entry.Key == null)
```
jamdavi, line 484: As an aside, a dictionary will not allow you to have a null value as a key. Might be overkill to check for this unless we're thinking that someone will violate the IDictionary<> Add method.

jamdavi (approved):
One note about the entry.Key null check

iothub/device/src/ClientPropertyCollection.cs
jamdavi: It will likely save us from some headache since we don't want to be in a position where we would try to serialize a null key

PR #2172, opened 2021-09-21
fix(dps-service): Remove client side validation of x509 CA references format

timtay-microsoft:


provisioning/service/src/Config/X509CAReferences.cs
```diff
@@ -42,11 +42,6 @@ public class X509CAReferences
         [JsonConstructor]
         internal X509CAReferences(string primary, string secondary = null)
         {
-            /* SRS_X509_CAREFERENCE_21_001: [The constructor shall throw ArgumentException if the primary CA reference is null or empty.] */
-            if(string.IsNullOrWhiteSpace(primary))
```
abhipsaMisra, line 46: In which scenario would we expect this to happen, i.e. all CA refs to be empty?

provisioning/service/src/Config/X509CAReferences.cs
```diff
@@ -42,11 +42,6 @@ public class X509CAReferences
         [JsonConstructor]
```
drwill-ms, line 42: You'd need to clean up the param note about it not being allowed to be null or empty.

provisioning/service/src/Config/X509CAReferences.cs
timtay-microsoft: Had a conversation with the DPS service folks and they confirmed both CA references can be null if the enrollment group is using symmetric keys, client certificates or signing certificates.

abhipsaMisra: Primary and secondary are the only properties on this class. If both are null, then is there a valid case for this type even being initialized, i.e. if the enrollment group is using symmetric keys, client certificates or signing certificates => these refs are null but we'd still expect X509CAReferences to be initialized?

timtay-microsoft: It may be better for the service not to return the CA references object if all its contents are null, sure. However, the SDK appears to be receiving them from the service in cases like that anyways.

abhipsaMisra (approved):

PR #2187, opened 2021-10-04
fix(iot-svc): Cleanup and deprecation warning of code in CryptoKeyGenerator

drwill-ms:
A service team member pointed out they like to use this method we have `GeneratePassword` for testing, but they are on net472 and it doesn't work for them.

Consequently, the method we call to generate the password, off the `System.Web.Security.Membership` is not available after net472, so even if we do expand support for this method to net472, we can't offer it for newer .net targets unless we find another way to do it.

The service team member was happy to copy the code and use it locally, so this isn't urgent.

> Edit: team decided to not expand support, and mark existing methods as obsolete.

### DIFF
--- a/iothub/service/src/Common/Security/CryptoKeyGenerator.cs
+++ b/iothub/service/src/Common/Security/CryptoKeyGenerator.cs
@@ -1,17 +1,20 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Linq;
 using System;
 using System.Text;
+using System.Security.Cryptography;
+
+#if NET451 || NET472
+
+using System.Web.Security;
+
+#endif
 
 #if !NET451
 
-using System.Security.Cryptography;
+using System.Linq;
 
-#else
-    using System.Web.Security;
-    using System.Security.Cryptography;
 #endif
 
 namespace Microsoft.Azure.Devices.Common
@@ -21,9 +24,9 @@ namespace Microsoft.Azure.Devices.Common
     /// </summary>
     static public class CryptoKeyGenerator
     {
-#if NET451
-        const int DefaultPasswordLength = 16;
-        const int GuidLength = 16;
+#if NET451 || NET472
+        private const int DefaultPasswordLength = 16;
+        private const int GuidLength = 16;
 #endif
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


iothub/service/src/Common/Security/CryptoKeyGenerator.cs
```diff
@@ -67,19 +66,18 @@ public static string GenerateKey(int keySize)
             return Convert.ToBase64String(GenerateKeyBytes(keySize));
         }
 
-#if NET451
+#if NET451 || NET472
```
drwill-ms, line 69: This is the key to enabling it for net472.

iothub/service/src/Common/Security/CryptoKeyGenerator.cs
```diff
@@ -38,20 +41,16 @@ static public class CryptoKeyGenerator
         /// <returns>Byte array representing the key.</returns>
         public static byte[] GenerateKeyBytes(int keySize)
         {
-#if !NET451
```
drwill-ms, line 41: Flipped the logic so it reads easier.

iothub/service/src/Common/Security/CryptoKeyGenerator.cs
```diff
@@ -38,20 +41,16 @@ static public class CryptoKeyGenerator
         /// <returns>Byte array representing the key.</returns>
         public static byte[] GenerateKeyBytes(int keySize)
```
jamdavi, line 42: Did you want to just go ahead and mark this depreciated?

iothub/service/src/Common/Security/CryptoKeyGenerator.cs
```diff
@@ -22,8 +25,8 @@ namespace Microsoft.Azure.Devices.Common
     static public class CryptoKeyGenerator
```
timtay-microsoft, line 25: The order of the modifiers here is throwing me off. Can you change this to "public static class" instead?

timtay-microsoft (approved):

azabbasi (approved):
It appears that CS0618 is being thrown by the build agent, indicating the obsolete method has references.

drwill-ms:
> It appears that CS0618 is being thrown by the build agent, indicating the obsolete method has references.

I'll make a test helper that isn't marked as obsolete so our test code has a local copy.

jamdavi (approved):

PR #2192, opened 2021-10-05
feat(device): add support for DisposeAsync

drwill-ms:
To ensure clients call `CloseAsync` before disposing, we can make use of this new dispose pattern that comes with .NET 3.0 and up. It has been in the latest preview for a bit (original idea came from a customer) but we haven't had any feedback for it. Do we like having this as part of our client?

<https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable?view=net-5.0>

### DIFF
--- a/iothub/device/src/DeviceClient.cs
+++ b/iothub/device/src/DeviceClient.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
@@ -16,6 +17,9 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     /// <threadsafety static="true" instance="true" />
     public class DeviceClient : IDisposable
+#if !NET451 && !NET472 && !NETSTANDARD2_0
+        , IAsyncDisposable
+#endif
     {
         /// <summary>
         /// Default operation timeout.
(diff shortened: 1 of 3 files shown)


iothub/device/src/DeviceClient.cs
```diff
@@ -613,12 +617,50 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+
+        /// <summary>
+        /// Disposes the client in an async way. See <see cref="IAsyncDisposable"/> for more information.
+        /// </summary>
+        /// <remarks>
+        /// Includes a call to <see cref="CloseAsync()"/>.
+        /// </remarks>
+        /// <example>
+        /// <code>
+        /// await using var client = DeviceClient.CreateFromConnectionString(...);
```
timtay-microsoft, line 640: Great code examples here. Thanks for this

timtay-microsoft (approved):

timtay-microsoft:
Is ```Dispose``` being phased out from .NET in general, or is the new expectation to implement both ```IDisposable``` and ```IAsyncDisposable``` moving forward?

drwill-ms:
> Is `Dispose` being phased out from .NET in general, or is the new expectation to implement both `IDisposable` and `IAsyncDisposable` moving forward?

`IAsyncDisposable` is interesting. Here's what I found that I can best demonstrate with a code snippet:

```csharp
    class Program
    {
        static async Task Main(string[] args)
        {
            // only sync dispose called
            using var mdc2 = new MyDisposableClass("unawaited");

            // calls the async dispose, but because the async impl calls the sync impl they both run
            await using var mdc = new MyDisposableClass("awaited");
        }
    }

    internal class MyDisposableClass : IDisposable, IAsyncDisposable
    {
        private readonly string _note;
        public MyDisposableClass(string note)
        {
            _note = note;
        }

        public void Dispose()
        {
            Console.WriteLine($"Dispose for {_note} called");
        }

        public ValueTask DisposeAsync()
        {
            Console.WriteLine($"DisposeAsync for {_note} called");
            Dispose();

            return ValueTask.CompletedTask;
        }
    }
```

This means that `Dispose` still exists and can be called explicitly or with a `using` statement. If one chooses to use the phrase `using await` instead, then they get `DisposeAsync` called instead. The only difference is the extra work the SDK does for you if you choose the awaitable one.

abhipsaMisra (approved):

Issue #2185, opened 2021-10-04
[Bug Report] Can not read/write ModulesContent due to inconsistent property naming


SaraOlsson:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64
- **Application's .NET Target Framework :** netcoreapp3.1
- **Device:** Laptop
- **SDK version used:** 1.35.0

## Description of the issue
When using the RegistryManager to create IoT Edge module Configurations, the SDK expects the ConfigurationContent to be of type **ModulesContent**, while on the server side it is named **ModuleContent**, which could be confirmed by looking at the network traffic:

```
"content":{"moduleContent":{"properties.desired.configurations":{"key":"value" ..
```
note that while module**s**Content (with an 's') shows up on configurations that are IoT Edge deployments:

```
"content":{"modulesContent":{"$edgeAgent":{"properties.desired.runtime.settings ..
```

However, in code, I can only create module configuration content with:
```
configuration.Content = new ConfigurationContent
{
    ModulesContent = new Dictionary<string, IDictionary<string, object>>
    {
        {
            "properties.desired.configurations",
            new Dictionary<string, object>
            {
                {"name", "Sara"}
            }
        }
    }
};
````

.. as the ConfigurationContent does not contain a definition for ModuleContent. Using ModulesContent causes *The deployment configuration is malformed or invalid*. Due to this issue, we can not list our module configurations nor create new ones. Can this be fixed soon, or what may be a workaround?

timtay-microsoft:
Taking a look at the service documentation around this [ConfigurationContent](https://docs.microsoft.com/en-us/rest/api/iothub/service/configuration/apply-on-edge-device) object shows that there is supposed to be both a ```ModuleContent``` and a ```ModulesContent``` object in this type. This strikes me as a bit odd though, so I'll reach out to a service engineer to confirm this.

There is no workaround for this bug as of now, but I'll try to get the fix checked in as soon as I can confirm what the service expects here.

drwill-ms:
A [release](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2021-11-01) with this fix is out.

PR #2208, opened 2021-10-14
feat(iot): Add missing moduleContent field to ConfigurationContent class

timtay-microsoft:
#2185

### DIFF
--- a/iothub/service/src/Configurations/ConfigurationContent.cs
+++ b/iothub/service/src/Configurations/ConfigurationContent.cs
@@ -12,8 +12,11 @@ namespace Microsoft.Azure.Devices
     public class ConfigurationContent
     {
         /// <summary>
-        /// Gets or sets Module Configurations
+        /// Gets or sets the configurations to be applied to the Edge agent.
         /// </summary>
+        /// <remarks>
+        /// For more information on this field, see <see href="https://docs.microsoft.com/en-us/azure/iot-edge/module-composition?view=iotedge-2020-11#create-a-deployment-manifest">this documentation</see>.
+        /// </remarks>
         [JsonProperty(PropertyName = "modulesContent")]
 #pragma warning disable CA2227 // Collection properties should be read only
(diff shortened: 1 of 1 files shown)


iothub/service/src/Configurations/ConfigurationContent.cs
```diff
@@ -12,16 +12,28 @@ namespace Microsoft.Azure.Devices
...
         public IDictionary<string, IDictionary<string, object>> ModulesContent { get; set; }
 
 #pragma warning restore CA2227 // Collection properties should be read only
 
         /// <summary>
-        /// Gets or sets Module Configurations
+        /// Gets or sets the configurations to be applied on device modules
+        /// </summary>
+        [JsonProperty(PropertyName = "moduleContent")]
+#pragma warning disable CA2227 // Collection properties should be read only
```
azabbasi, line 30: Do these have any benefits over the attribute version of warning suppression? As far as I know using the preprocessor directives and the using the suppression attribute are alternatives. Attributes read and look much cleaner specially for property warning suppressions.

iothub/service/src/Configurations/ConfigurationContent.cs
```diff
@@ -12,16 +12,28 @@ namespace Microsoft.Azure.Devices
...
+        public IDictionary<string, object> ModuleContent { get; set; }
```
azabbasi, line 31: Interesting, so we never even had it? do we know if this was added at some point and we just never got the message or we originally missed it? Not that it really matters here, I am just curious.

azabbasi (approved):
LGTM

iothub/service/src/Configurations/ConfigurationContent.cs
timtay-microsoft: I'm not sure. I imagine the SDK team wasn't given a full spec for this feature, and this field was lost in the shuffle somehow

Issue #2202, opened 2021-10-13
[Bug Report] Device client retry policy not working as expected


Werner-Prbk:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64 (also on linux arm32) 
- **Application's .NET Target Framework :** .netcoreapp 3.1
- **Device:** Laptop (also on arm32 device with custom linux)
- **SDK version used:** 1.38.0

## Description of the issue
Setting a ExponentialBackoff retry policy with retry count N leads to N*N retry attempts before it completely stops retrying. 
I assume the expected behavior is, that the retry policy starts when the device client enters connection state (Disconnected_Retrying, Communication_Error), retries N times and finally changes to (Disconnected, Retry_Expired). Once it is in (Disconnected, Retry_Expired) no more retries shall be carried out and the connection state shall no more change.

I am not completely sure, but I guess the problem is in `RetryDelegationHandler.cs` because of a nested retry block:

* [SendEventAsync()](https://github.com/Azure/azure-iot-sdk-csharp/blob/main/iothub/device/src/Transport/RetryDelegatingHandler.cs#L70) calls [EnsureOpenedAsync()](https://github.com/Azure/azure-iot-sdk-csharp/blob/4f888245d253692b6c18509fc68c2412ddd4b540/iothub/device/src/Transport/RetryDelegatingHandler.cs#L641) in a retry block
* EnsureOpenedAsync() calls [OpenInternalAsync()](https://github.com/Azure/azure-iot-sdk-csharp/blob/4f888245d253692b6c18509fc68c2412ddd4b540/iothub/device/src/Transport/RetryDelegatingHandler.cs#L750) in another retry block

## Code sample exhibiting the issue
```
static async Task Main()
{
      deviceClient = DeviceClient.CreateFromConnectionString(connectionString, TransportType.Mqtt);

       // This should result in disconnected/retry_expired after a few seconds
      deviceClient.SetRetryPolicy(
          new ExponentialBackoff(
              6,   // retry counter
              TimeSpan.FromSeconds(1),
              TimeSpan.FromSeconds(10),
              TimeSpan.FromSeconds(1)));

      deviceClient.SetConnectionStatusChangesHandler((status, reason) =>
          Print($"Connection status changed: {status}, {reason}"));

      await deviceClient.OpenAsync();
      Print("Connected");

      try
      {
          while (true)
          {
              await Task.Delay(TimeSpan.FromSeconds(5));
[...]

PR #2209, opened 2021-10-18
fix(iot-dev): Fix bug where client's retry policy applied n^2 times rather than n times

timtay-microsoft:
#2202

previously:
retry(retry(open) + send/receive/etc)

now:
retry(open + send/receive/etc)

This is as opposed to doing something like:
retry(open) + retry(send/receive/etc)

because the connection may be lost after the retry(open) block has finished, and we don't want to keep trying to send/receive/etc without checking that the connection is open first


iothub/device/src/Transport/RetryDelegatingHandler.cs
```diff
@@ -747,31 +747,54 @@ private async Task EnsureOpenedAsync(TimeoutHelper timeoutHelper)
             }
         }
 
-        private Task OpenInternalAsync(CancellationToken cancellationToken)
+        private async Task OpenInternalAsync(bool withRetry, CancellationToken cancellationToken)
```
timtay-microsoft, line 750: Previously this method returned a Task, but that task was immediately awaited wherever it was called, so making this async should be safe

PR #2226, opened 2021-11-02
Migrate from FxCop analyziers to .NET analyzers

drwill-ms:
https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/
https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2019

The old packages are deprecated.
The new one has an option for .NET 5.0+, but we aren't homogenous for .NET target, so we'll go with the NuGet reference option.

With the default options, we don't have any new warnings. However, as spelled out in the instructions, if we add `<AnalysisMode>AllEnabledByDefault</AnalysisMode>` to the Hub Device SDK (as an example), it adds 28 new code analysis warnings, so we should have a follow-up effort to up the bar for analysis and fix or suppress these additional rules.

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -3,7 +3,6 @@
     <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net5.0;netstandard2.1;netstandard2.0;net472;net451</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net5.0;netstandard2.1;netstandard2.0</TargetFrameworks>
     <LangVersion>8.0</LangVersion>
-    <!-- TODO #176: Enable warnings as errors. -->
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <RootNamespace>Microsoft.Azure.Devices.Client</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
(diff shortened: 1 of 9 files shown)


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -3,7 +3,6 @@
     <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net5.0;netstandard2.1;netstandard2.0;net472;net451</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net5.0;netstandard2.1;netstandard2.0</TargetFrameworks>
     <LangVersion>8.0</LangVersion>
-    <!-- TODO #176: Enable warnings as errors. -->
```
drwill-ms, line 6: This should have been removed back in the day when we turned warnings as errors on.

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -121,12 +124,4 @@
     <PackageReference Include="Microsoft.AspNet.WebApi.Client" Version="5.2.3" />
     <PackageReference Include="Microsoft.Owin" Version="4.0.0" />
   </ItemGroup>
-
-  <ItemGroup>
-    <!-- FXCop -->
-    <PackageReference Condition=" '$(Configuration)' == 'Debug' " Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.1">
```
drwill-ms, line 127: We used to have this on only in Debug. I don't quite think that is what we'd have done if we started fresh. Any opinions?

azabbasi (approved):

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
azabbasi: I personally think we should add it for both debug and release unless there is a specific drawback to having it enabled.

timtay-microsoft (approved):

PR #2229, opened 2021-11-03
Replace xUnit in Shared.Tests with MSTest

drwill-ms:
I originally wrote these tests in xUnit as an evaluation on whether we should move to a generally considered richer test framework.

However, we soon learned that xUnit doesn't fully support 4.5.1 (which we have to) and test discovery doesn't work. The work around at the time was to reference a beta version of xUnit with hope they'd eventually fix it for real. Well, the GitHub issue we saved shows they chose instead to close it and not worry about it given net451 is such an old framework.

So let's just get rid of this xUnit usage and consolidate on the MS unit test framework.

### DIFF
--- a/e2e/test/E2ETests.csproj
+++ b/e2e/test/E2ETests.csproj
@@ -37,6 +37,7 @@
     <PackageReference Include="MSTest.TestAdapter" Version="2.0.0" />
     <PackageReference Include="MSTest.TestFramework" Version="2.0.0" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7" />
+    <PackageReference Include="System.Diagnostics.Tracing" Version="4.3.0" />
   </ItemGroup>
 
   <!-- Legacy .NET 4.5.1 -->
--- a/iothub/device/tests/Edge/EdgeModuleClientFactoryTest.cs
+++ b/iothub/device/tests/Edge/EdgeModuleClientFactoryTest.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Azure.Devices.Client.Test.Edge
         }
 
         [TestMethod]
-        public async void TestCreate_FromConnectionStringEnvironment_ShouldCreateClient()
+        public async Task TestCreate_FromConnectionStringEnvironment_ShouldCreateClient()
         {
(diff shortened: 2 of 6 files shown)


iothub/device/tests/Edge/EdgeModuleClientFactoryTest.cs
```diff
@@ -36,7 +36,7 @@ public EdgeModuleClientFactoryTest()
         }
 
         [TestMethod]
-        public async void TestCreate_FromConnectionStringEnvironment_ShouldCreateClient()
+        public async Task TestCreate_FromConnectionStringEnvironment_ShouldCreateClient()
```
drwill-ms, line 39: Looking at test discovery logs, I found errors for these tests, and fixed them.

provisioning/service/tests/Microsoft.Azure.Devices.Provisioning.Service.Tests.csproj
```diff
@@ -5,7 +5,6 @@
     <!-- By setting this as not a test project for other frameworks, we can run "dotnet test" for a specific framework from the root directory without dotnet throwing if any test project in the solution doesn't support that framework -->
     <IsTestProject Condition="'$(TargetFramework)' != 'netcoreapp3.1' And '$(TargetFramework)' != 'netcoreapp2.1.18'">False</IsTestProject>
     <LangVersion>8.0</LangVersion>
-    <!-- FXCop TODO: #176 re-enable warnings as errors. -->
```
drwill-ms, line 8: This should have been removed when we turned on warnings as errors.

e2e/test/E2ETests.csproj
```diff
@@ -37,6 +37,7 @@
     <PackageReference Include="MSTest.TestAdapter" Version="2.0.0" />
     <PackageReference Include="MSTest.TestFramework" Version="2.0.0" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7" />
+    <PackageReference Include="System.Diagnostics.Tracing" Version="4.3.0" />
```
drwill-ms, line 40: I added this in response to an error printed in test discovery.

timtay-microsoft (approved):

azabbasi (approved):

PR #2230, opened 2021-11-04
Test for export/import device needed updating; service bug still exis…

drwill-ms:
Welp, we accidentally introduced the ParentScopes property in ExportImportDevice but the service is still not exporting that value. I've updated the test in anticipation of it being fixed.

### DIFF
--- a/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
+++ b/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
 
         private const string ExportFileNameDefault = "devices.txt";
         private const int MaxIterationWait = 30;
-        private static readonly TimeSpan _waitDuration = TimeSpan.FromSeconds(5);
+        private static readonly TimeSpan s_waitDuration = TimeSpan.FromSeconds(3);
 
         private static readonly char[] s_newlines = new char[]
         {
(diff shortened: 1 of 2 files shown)


iothub/service/src/ExportImportDevice.cs
```diff
@@ -21,19 +21,18 @@ public sealed class ExportImportDevice
         private string _twinETag;
 
         /// <summary>
-        /// Property container
+        /// Tyep definition for the <see cref="Properties"/> property.
```
timtay-microsoft, line 24:
```suggestion
        /// Type definition for the <see cref="Properties"/> property.
```

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
```diff
@@ -47,14 +47,15 @@ public class RegistryManagerExportDevicesTests : E2EMsTestBase
         [Timeout(120000)]
         [DoNotParallelize]
         [DataRow(StorageAuthenticationType.KeyBased, false)]
-        [DataRow(StorageAuthenticationType.IdentityBased, false)]
-        [DataRow(StorageAuthenticationType.IdentityBased, true)]
+        //[DataRow(StorageAuthenticationType.IdentityBased, false)]
```
azabbasi, line 50: Can you elaborate why these are commented out?

timtay-microsoft (approved):

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
drwill-ms: Accidentally included in the PR. Thanks for catching. They were making it awkward to debug so I (meant to) temporarily disable them.

azabbasi (approved):

PR #2231, opened 2021-11-04
Add a test for RegistryManager operations for configurations CRUD

drwill-ms:


e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -18,8 +18,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
     [TestCategory("IoTHub")]
     public class RegistryManagerE2ETests : E2EMsTestBase
     {
-        private readonly string _devicePrefix = $"E2E_{nameof(RegistryManagerE2ETests)}_";
```
drwill-ms, line 21: As they are the same, just generalize it. I wanted to use it for configurations, too.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -189,12 +188,80 @@ public async Task RegistryManager_AddDeviceWithProxy()
...
+            try
+            {
+                var expected = new Configuration(configurationId)
+                {
+                    Priority = 2,
+                    Labels = { { "labelName", "labelValue" } },
+                    TargetCondition = "*",
+                    Content =
+                    {
+                        DeviceContent = { { "properties.desired", "\"setting1\": 4" } },
```
drwill-ms, line 209: I'm unsure on what this is supposed to be. As a payload to test, it works fine because GIGO is fine for testing the SDK, but I'd rather this be a logical payload that a customer might use.

Anyone have an idea here?

iothub/service/src/Configurations/Configuration.cs
```diff
@@ -47,31 +50,31 @@ internal Configuration()
...
-        public string SchemaVersion { get; internal set; }
+        public string SchemaVersion { get; }
 
         /// <summary>
         /// Gets or sets labels for the configuration
         /// </summary>
         [JsonProperty(PropertyName = "labels", NullValueHandling = NullValueHandling.Ignore)]
 #pragma warning disable CA2227 // Collection properties should be read only
-        public IDictionary<string, string> Labels { get; set; }
+        public IDictionary<string, string> Labels { get; set; } = new Dictionary<string, string>();
```
drwill-ms, line 60: Added these initializers to avoid NREs in customer code.

iothub/service/src/Configurations/Configuration.cs
```diff
@@ -47,31 +50,31 @@ internal Configuration()
...
         [JsonProperty(PropertyName = "content", NullValueHandling = NullValueHandling.Ignore)]
-        public ConfigurationContent Content { get; set; }
+        public ConfigurationContent Content { get; set; } = new ConfigurationContent();
 
         /// <summary>
         /// Gets the content type for configuration
         /// </summary>
         [JsonProperty(PropertyName = "contentType")]
-        public string ContentType { get; internal set; }
+        public string ContentType { get; }
```
drwill-ms, line 74: We seem to treat this as read-only for now, so removing the internal set.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
azabbasi: I do not know either.

drwill-ms: Rick Kessler may be able to fill in the blanks here. I reached out to him on Teams and waiting for his availability today.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -189,12 +188,80 @@ public async Task RegistryManager_AddDeviceWithProxy()
             await registryManager.AddDeviceAsync(device).ConfigureAwait(false);
         }
 
+        [LoggedTestMethod]
+        public async Task RegistryManager_ConfigurationOperations_Work()
+        {
+            // arrange
+
+            bool configCreated = false;
+            string configurationId = (_idPrefix + Guid.NewGuid()).ToLower(); // Configuration Id characters must be all lower-case.
```
azabbasi, line 197: is this requirement enforced by the service? should we document that in the `Configuration` constructor and the Id property?

azabbasi (approved):

e2e/test/iothub/service/RegistryManagerE2ETests.cs
drwill-ms: Yes. Added documentation to the ctor. As it can't be set on the Id property, I didn't carry the documentation over to the property.

e2e/test/iothub/service/RegistryManagerE2ETests.cs
drwill-ms: Rick helped me. I feel ramped up now. :)

PR #2234, opened 2021-11-08
(Fix: DeviceClient): Fix the concurrency issue in MQTT stack

### LINKED ISSUE #2232: DeviceClient-SendEventAsync throwing InvalidOperationException
# Context

- **OS:** Mac BigSur 11.6
- **Application's .NET Target Framework :** netcoreapp3.1 
- **Device:** Laptop
- **SDK version used:** Microsoft.Azure.Devices.Client 1.39.0

## Description of the issue
Wrote a small console app trying to test the robustness of the iothub when sending 125000 messages(5000 messages every 5 min triggered by 25 timers). It all works fine but suddenly I see lot of error messages thrown especially related to InvalidOperationException - Collection was modified after the enumerator was instantiated.

This happens on the deviceClient.SendEventAsync API call.

Note sure if this is a bug or something we are not doing right, but it happens when ran for a long time. Please let us know if there is something on the application side we need to change.

```cs
       private static DeviceClient _iotHubClient = DeviceClient.CreateFromConnectionString(ConnectionString, TransportType.Mqtt_WebSocket_Only);
       private static string Message = "{}"; // Some valid json message.
        static void Main(string[] args)
        {
            Console.WriteLine("Number of instances to span: ");
            var instanceInput = Convert.ToInt32(Console.ReadLine());  // 25
            Console.WriteLine("Interval to send messages: ");
            var interval = Convert.ToInt32(Console.ReadLine());   // 5
            Console.WriteLine("Number of messages to send: Default is 5000");
            string messageCount = Console.ReadLine();   // 5000
            if (!String.IsNullOrEmpty(messageCount))
                _messageCount = Convert.ToInt32(messageCount);

            for (int instanceCount = 0; instanceCount < instanceInput; instanceCount++)
            {
                System.Timers.Timer timer = new System.Timers.Timer(TimeSpan.FromMinutes(interval).TotalMilliseconds);
                timer.AutoReset = true;
                timer.Elapsed += async(sender, e) => await SendMessagesToIOTHub(sender, e);
                timer.Start();
                Console.WriteLine($"{DateTime.Now}: Timer started to execute every {interval} minutes");
            }

            while (true)
            {
                Thread.Sleep(1000);
            }
        }

        private static async Task SendMessagesToIOTHub(object sender, ElapsedEventArgs e)
        {
            Console.WriteLine($"{DateTime.Now}: Sending {_messageCount} messages");
            for (int i = 1; i <= _messageCount; i++)
            {
                try
                {
                    var message = new Message(Encoding.ASCII.GetBytes(Message));
                    message.ContentType = "application/json";
                    message.ContentEncoding = "utf-8";

                    await _iotHubClient.SendEventAsync(message);
                }
                catch (IotHubCommunicationException ice)
                {
                    Console.WriteLine($"{DateTime.Now}: IotHubCommunicationException - {ice.ToString()}");
                }
                catch (NullReferenceException nre)
                {
                    Console.WriteLine($"{DateTime.Now}: NullReferenceException - {nre.ToString()}");
                }
                catch (AggregateException agE)
                {
                    Console.WriteLine($"{DateTime.Now}: AggregateException - {agE.ToString()}");
                }
                catch (Exception ex)
                {
                    Console.WriteLine($"{DateTime.Now}: General Exception sending message {ex.ToString()}");
                }
            }
            Console.WriteLine($"{DateTime.Now}: {_messageCount} messages sent");
        }
```

## Console log of the issue
Here is the call stack of the error noticed...

11/4/2021 6:55:15 PM: General Exception sending message System.InvalidOperationException: Collection was modified after the enumerator was instantiated.
   at System.Collections.Generic.Queue`1.Enumerator.MoveNext()
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.OrderedTwoPhaseWorkQueue`2.CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.ProcessMessage(IChannelHandlerContext context, Packet packet)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.SendMessageAsync(IChannelHandlerContext context, Message message)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.WriteAsync(IChannelHandlerContext context, Object data)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttTransportHandler.SendEventAsync(Message message, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.SendMessageAsync(IChannelHandlerContext context, Message message)
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.<>c__DisplayClass27_0.<<ExecuteWithErrorHandlingAsync>b__0>d.MoveNext()
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.WriteAsync(IChannelHandlerContext context, Object data)
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.ExecuteWithErrorHandlingAsync[T](Func`1 asyncOperation)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.SendMessageAsync(IChannelHandlerContext context, Message message)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttTransportHandler.SendEventAsync(Message message, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.WriteAsync(IChannelHandlerContext context, Object data)
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.<>c__DisplayClass27_0.<<ExecuteWithErrorHandlingAsync>b__0>d.MoveNext()
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.<>c__DisplayClass15_0.<<SendEventAsync>b__0>d.MoveNext()
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.ExecuteWithErrorHandlingAsync[T](Func`1 asyncOperation)
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.<>c__DisplayClass27_0.<<ExecuteWithErrorHandlingAsync>b__0>d.MoveNext()
   at Microsoft.Azure.Devices.Client.Transport.Mqtt.MqttIotHubAdapter.WriteAsync(IChannelHandlerContext context, Object data)
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.SendEventAsync(Message message, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.Transport.RetryDelegatingHandler.SendEventAsync(Message message, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.Transport.ErrorDelegatingHandler.ExecuteWithErrorHandlingAsync[T](Func`1 asyncOperation)
   at Microsoft.Azure.Devices.Client.InternalClient.SendEventAsync(Message message)
   at Microsoft.Azure.Devices.Client.InternalClient.SendEventAsync(Message message)

azabbasi:
This PR fixes #2232

In a multithreaded solution, there is a chance of Enqueuing and Dequeuing happening simultaneously and we need to guard against those cases. 
We have two approaches available to us to take advantage of, 
1- Use concurrent collections
2- Use regular collections with a semaphore lock.

Based on an interesting article [here](https://docs.microsoft.com/en-us/dotnet/standard/collections/thread-safe/when-to-use-a-thread-safe-collection#concurrentqueuet-vs-queuet), It's recommended to use a regular collection with an external lock instead of a concurrent collection.

I quote: 
```
Mixed producer-consumer scenario: 
Any given thread is both adding and removing elements.

In mixed producer-consumer scenarios, when the processing time is very small, a Queue<T> that has an external lock scales better than ConcurrentQueue<T> does. However, when processing time is around 500 FLOPS or more, then the ConcurrentQueue<T> scales better.
```
After discussing this more, I went back on my decision to use external locks to protect the queue.

Even though scalability may play a role here, it seems like it won't be worth the hassle and complexity of making multiple classes disposable since we have had many issues with disposing in the past.
Concurrent collections come with default thread-safety behaviors and seem to be the way to go.

Feedback and insights are more than welcome :)


timtay-microsoft (approved):

timtay-microsoft:
There's something to be said about the simplicity of just using concurrency safe data structures, but considering the downsides the article brings up and the fact that you've already done the work of building out these locks, I'm fine with this approach.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapterFactory.cs
```diff
@@ -7,11 +7,11 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 {
     internal class MqttIotHubAdapterFactory
     {
-        private readonly MqttTransportSettings settings;
+        private readonly MqttTransportSettings _settings;
```
andyk-ms, line 10: Good standard for underscore variable name for readonly.

andyk-ms (approved):

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -77,18 +77,19 @@ private enum StateFlags
...
         private readonly IAuthorizationProvider _passwordProvider;
-        private readonly SimpleWorkQueue<PublishWorkItem> _serviceBoundOneWayProcessor;
-        private readonly OrderedTwoPhaseWorkQueue<int, PublishWorkItem> _serviceBoundTwoWayProcessor;
+        private SimpleWorkQueue<PublishWorkItem> _serviceBoundOneWayProcessor;
+        private OrderedTwoPhaseWorkQueue<int, PublishWorkItem> _serviceBoundTwoWayProcessor;
         private readonly IWillMessage _willMessage;
 
         private DateTime _lastChannelActivityTime;
         private StateFlags _stateFlags;
+        private bool _disposed;
```
drwill-ms, line 92: _isDisposed: bools should start with a verb that makes the type clearer

iothub/device/src/Transport/Mqtt/MqttIotHubAdapterFactory.cs
drwill-ms: We use underscore for private members whether they are readonly or not.

If they are static, then we use s_.

If they are constant we use TitleCase.

drwill-ms (changes requested):

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -46,11 +46,13 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
                 throw new IotHubException("Nothing to complete.", isTransient: false);
             }
 
-            IncompleteWorkItem incompleteWorkItem = _incompleteQueue.Peek();
-            if (incompleteWorkItem.Id.Equals(workId))
+            if (_incompleteQueue.TryPeek(out IncompleteWorkItem incompleteWorkItem)
```
drwill-ms, line 49: I don't think you are supposed to TryPeek and then assume when you call TryDequeue that you get the same item. In the intervening time another thread could have dequeued it.

drwill-ms (changes requested):
I'm concerned about the TryPeek/TryDequeue logic.

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
jamdavi: Some questions I have here are, is this really a queue model since we're not dealing with FIFO? Are the workid's guaranteed to be unique? If so couldn't we do this with a dictionary?

azabbasi: Well, messages have to be processed in the order they came in, so it is in fact FIFO in this MQTT messaging model.

azabbasi: I have removed the TryPeek approach and dequeuing it instead, if TryPeek returned the wrong workitem, we would have thrown a non-transient exception so it doesn't seem to matter whether or not we dequeue the message or not at that point, we can have the same functionality using dequeue and that should resolve the issue.

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -46,16 +46,22 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
...
             {
-                _incompleteQueue.Dequeue();
-                return _completeWorkAsync(context, incompleteWorkItem.WorkItem);
-            }
+                if (incompleteWorkItem.Id.Equals(workId))
+                {
+                    return _completeWorkAsync(context, incompleteWorkItem.WorkItem);
+                }
 
-            throw new IotHubException(
```
abhipsaMisra, line 56: Q - is this exception caught higher in the stack and then the operation retried? If yes, then the new implementation will show a different behavior since previously the queue was unchanged but now the queue will have been modified (peek vs dequeue).

azabbasi: No retry happens here, in case this exception is thrown, everything shuts down and the queue will be emptied out.

WillooWisp:
@azabbasi will this solve this "deadlock" issue, https://github.com/Azure/azure-iot-sdk-csharp/issues/2237?

WillooWisp:
@azabbasi just verified that this branch (PR) does not solve the deadlock referenced here, https://github.com/Azure/azure-iot-sdk-csharp/issues/2237.

In fact disconnecting from internet not even triggers the ConnectionStatusChangesHandler sometimes using this branch.

azabbasi:
> @azabbasi just verified that this branch (PR) does not solve the deadlock referenced here, #2237.
> 
> In fact disconnecting from internet not even triggers the ConnectionStatusChangesHandler sometimes using this branch.

This PR doesn't address any deadlock issues, this helps fix concurrency issue with the internal queue we have for MQTT

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -46,16 +46,22 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
...
-                $"Work must be complete in the same order as it was started. Expected work id: '{incompleteWorkItem.Id}', actual work id: '{workId}'",
-                isTransient: false);
+                throw new IotHubException(
+                    $"Work must be complete in the same order as it was started. Expected work id: '{incompleteWorkItem.Id}', actual work id: '{workId}'",
+                    isTransient: false);
+            }
+#if NET451
```
drwill-ms, line 60: Consider using `TaskHelpers.CompletedTask`

iothub/device/src/Transport/Mqtt/SimpleWorkQueue.cs
```diff
@@ -111,9 +111,8 @@ public virtual void Abort(Exception exception)
                 case States.FinalProcessing:
                     State = States.Aborted;
 
-                    while (_backlogQueue.Any())
+                    while (_backlogQueue.Any() && _backlogQueue.TryDequeue(out TWork workItem))
```
drwill-ms, line 114: I'm wondering if `_backlogQueue.Any()` is needed here.

azabbasi: You are correct, not needed.

drwill-ms: There is one a little further down in the PR too.

drwill-ms (approved):

PR #2249, opened 2021-12-03
Upgrade AMQP library to 2.5.8

azabbasi:


iothub/service/src/ServiceClient.cs
```diff
@@ -272,7 +272,9 @@ public virtual async Task OpenAsync()
         {
             Logging.Enter(this, $"Opening AmqpServiceClient", nameof(OpenAsync));
 
-            await _faultTolerantSendingLink.OpenAsync(_openTimeout).ConfigureAwait(false);
+            using var ctx = new CancellationTokenSource(_openTimeout);
+
+            await _faultTolerantSendingLink.OpenAsync(ctx.Token).ConfigureAwait(false);
```
azabbasi, line 277: This method no longer supports timespans.

drwill-ms (approved):

iothub/service/src/ServiceClient.cs
abhipsaMisra: oh, breaking change for the amqp library then!

abhipsaMisra (approved):

iothub/service/src/ServiceClient.cs
azabbasi: :D

PR #2250, opened 2021-12-06
feat(hub-svc): add support and tests for configurations export

drwill-ms:


e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
```diff
@@ -101,123 +103,222 @@ await registryManager
                         })
                     .ConfigureAwait(false);
 
+                Configuration configuration = await registryManager
```
drwill-ms, line 106: To test exporting configs, we have to have created some first.

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
```diff
@@ -101,123 +103,222 @@ await registryManager
...
+                        configuration)
+                    .ConfigureAwait(false);
+            }
+            finally
+            {
+                await CleanUpDevicesAsync(edgeId1, edgeId2, deviceId, configurationId, registryManager).ConfigureAwait(false);
+            }
+        }
+
+        private async Task<JobProperties> RunExportJobAsync(StorageAuthenticationType storageAuthenticationType, bool isUserAssignedMsi, string devicesFileName, string configsFileName, RegistryManager registryManager, Uri containerUri)
```
drwill-ms, line 157: The method above was getting pretty long, so I broke out running the job to a separate method.

e2e/test/helpers/ImportExportHelpers.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.E2ETests.Helpers
+{
+    internal static class ImportExportHelpers
```
drwill-ms, line 10: Renamed because it is no longer used just for devices import

iothub/service/src/Configurations/ConfigurationImportMode.cs
```diff
@@ -0,0 +1,21 @@
...
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum ConfigurationImportMode
+    {
+        /// <summary>
+        /// If a configuration does not exist with the specified Id, it is newly registered.
+        /// If the configuration already exists, existing information is overwritten with the provided input data only if there is an ETag match.
+        /// If there is an ETag mismatch, an error is written to the log file.
+        /// </summary>
+        [EnumMember(Value = "createOrUpdateIfMatchETag")]
+        CreateOrUpdateIfMatchETag = 4,
```
drwill-ms, line 19: Isao says they only support this mode for now, so we're not using the same `ImportMode` class as is used for ExportImportDevice, but I matched the same enum value as the other enum to allow room for the other items to be inserted above (numerically).

iothub/service/src/JobProperties.cs
```diff
@@ -124,13 +124,31 @@ public JobProperties()
...
         public ManagedIdentity Identity { get; set; }
 
+        /// <summary>
+        /// Whether or not to include configurations in the import or export job.
+        /// </summary>
+        /// <remarks>
+        /// The service assumes this is false, if not specified. If true, then configurations are included in the data export/import.
+        /// </remarks>
+        [JsonProperty(PropertyName = "includeConfigurations", NullValueHandling = NullValueHandling.Ignore)]
+        public bool? IncludeConfigurations { get; set; }
```
drwill-ms, line 134: One has to remember to set this to true, even if they've set the ConfigurationsBlobName property. I accidentally did that and couldn't figure out why my configs weren't being imported.

I've considered setting this to true if/when ConfigurationsBlobName is set. Thoughts?

iothub/service/src/ManagedIdentity.cs
```diff
@@ -15,10 +13,21 @@ namespace Microsoft.Azure.Devices
     /// </summary>
     public class ManagedIdentity
     {
+        /// <summary>
+        /// The user identity resource Id used to access the storage account for import and export jobs.
+        /// </summary>
+        [Obsolete("Use UserAssignedIdentity instead")]
+        [JsonIgnore]
+        public string userAssignedIdentity
```
drwill-ms, line 21: Migrating users to the new, properly-cased property.

e2e/test/helpers/ImportExportHelpers.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
jamdavi, line 2: Whitespace.

```suggestion
﻿// Copyright (c) Microsoft. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.

```

iothub/service/src/Configurations/Configuration.cs
```diff
@@ -41,19 +41,19 @@ internal Configuration()
         }
 
         /// <summary>
-        /// Gets the identifier for the configuration.
+        /// The unique identifier of the configuration.
```
jamdavi, line 44: For all of the property changes I recall a while back that we had an analysis rule that flags properties as needing "Gets ..." "Sets ...", or "Gets or sets ..." depending on the property access modifiers. Did we remove that?

Maybe we need to utilize the value feild?

```csharp
/// <summary>
/// Gets the identifier for the configuration.
/// </summary>
/// <value>
/// The full identifier of the configuration as a string.
/// </value>
```

iothub/service/src/ExportImportDevice.cs
```diff
@@ -91,32 +106,43 @@ public string ETag
...
         /// <summary>
-        /// Import mode of the device.
+        /// The type of registry operation and ETag preferences.
         /// </summary>
         [JsonProperty(PropertyName = "importMode", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
         public ImportMode ImportMode { get; set; }
 
         /// <summary>
-        /// Status of the device.
+        /// The status of the device or module. If disabled, it cannot connect to the service.
```
jamdavi, line 115: > If disabled, it cannot connect to the service.

Better as a remark.

iothub/service/src/JobProperties.cs
jamdavi: No, while it's an extra step if the customer decides to run this export job again but doesn't want the configurations included it'd potentially overwrite the configs.

iothub/service/src/ManagedIdentity.cs
jamdavi: Ugh....

jamdavi:
```suggestion
        [EditorBrowsable(EditorBrowsableState.Never)]
        public string userAssignedIdentity
```

iothub/service/src/Configurations/Configuration.cs
drwill-ms: I changed these to match the (what I thought were better) swagger comments.

I don't recall ever seeing a C# standard that properties need to say "Gets or sets...". Maybe that is a Java thing?

jamdavi (approved):

e2e/test/helpers/ImportExportHelpers.cs
```diff
@@ -0,0 +1,30 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.E2ETests.Helpers
+{
+    internal static class ImportExportHelpers
+    {
+        /// <summary>
+        /// Makes a stream compatible for writing to a storage blob of serialized, newline-delimited rows of the specified objects
+        /// </summary>
+        /// <param name="items">The objects to serialize</param>
```
azabbasi, line 16: missing a period.

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
```diff
@@ -101,123 +93,234 @@ await registryManager
...
+
+                if (StringComparer.Ordinal.Equals(exportedDevice.Id, edge2.Id) && exportedDevice.Capabilities.IotEdge)
+                {
+                    Logger.Trace($"Found edge2 in export as [{serializedDevice}]");
+                    foundEdge2InExport = true;
+                    exportedDevice.DeviceScope.Should().Be(edge2.Scope, "Edges retain their own scope");
+
+                    // This is broken. The export doesn't include the ParentScopes property.
+                    // Disable this assert until it is fixed in the service.
+                    //exportedDevice.ParentScopes.First().Should().Be(edge1.Scope);
```
azabbasi, line 266: Should it be commented out?

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
```diff
@@ -101,123 +93,234 @@ await registryManager
...
 
         private static async Task<string> DownloadFileAsync(StorageContainer storageContainer, string fileName)
         {
             CloudBlockBlob exportFile = storageContainer.CloudBlobContainer.GetBlockBlobReference(fileName);
-            string fileContents = await exportFile.DownloadTextAsync().ConfigureAwait(false);
+            return await exportFile.DownloadTextAsync().ConfigureAwait(false);
+        }
 
-            return fileContents;
+        private async Task CleanUpDevicesAsync(string edgeId1, string edgeId2, string deviceId, string configurationId, RegistryManager registryManager)
```
azabbasi, line 311: consider breaking it down to multiple lines.

e2e/test/iothub/service/RegistryManagerImportDevicesTests.cs
```diff
@@ -130,38 +122,45 @@ public async Task RegistryManager_ImportDevices(StorageAuthenticationType storag
...
+                }
             }
             finally
             {
                 try
                 {
                     await registryManager.RemoveDeviceAsync(deviceId).ConfigureAwait(false);
+                    await registryManager.RemoveConfigurationAsync(configId).ConfigureAwait(false);
                 }
                 catch { }
```
azabbasi, line 156: maybe log the exception here?

iothub/service/src/Configurations/Configuration.cs
```diff
@@ -9,22 +9,22 @@
 namespace Microsoft.Azure.Devices
 {
     /// <summary>
-    /// Device configurations provide the ability to perform IoT device configuration at scale.
-    /// You can define configurations and summarize compliance as the configuration is applied.
+    /// The configuration for Iot Hub device and module twins.
```
azabbasi, line 12: IoT?

azabbasi (approved):

PR #2260, opened 2021-12-16
fix(deviceClient): Fix issue with AMQP connection pool and TokenReferesher disposal.

### LINKED ISSUE #2235: [Bug Report] GetFileUploadSasUriAsync throws ObjectDisposedException
# Context

- **OS, version, SKU and CPU architecture used:** Both Windows & Linux
- **Application's .NET Target Framework :** .net core 3.1
- **Device:** Desktop, Web App
- **SDK version used:**

The bug is in 1.38, 1.39
Everything worked as it should in earlier versions, tested: 1.37.2, 1.37, 1.36, 1.35

## Description of the issue
The SDK version 1.38 introduced issues when uploading files using DeviceClient, with Pooling and Amqp_Tcp_Only

To reproduce (100% reproduceable, on both Linux and Windows):
1. Create a DeviceClient for Device A, do a file upload (call GetFileUploadSasUriAsync, then upload to blob, then call CompleteFileUploadAsync)
2. Dispose DeviceClient for Device A.
3. Create a DeviceClient for Device B, do a fileupload
4. Dispose DeviceClient for Device B
5. Create a DeviceClient for Device A, do a file upload, (will throw on GetFileUploadSasUriAsync)

you will need a textfile called "example.txt" in the same folder.

This will throw when using SDK version 1.38 & 1.39 but NOT on earlier versions.

```
class Program
    {
        static ExponentialBackoff retryPolicy
            = new ExponentialBackoff(int.MaxValue, TimeSpan.FromMilliseconds(100), TimeSpan.FromSeconds(10), TimeSpan.FromMilliseconds(100));

        static async Task Main(string[] args)
        {
            var tasks = new List<Task>();
            for (int i = 1; i < 10; i++)
            {
                int clientIndex = i % 2 == 0 ? 1 : 2;
                var task = Task.Run(async () => await RunClient(clientIndex));
                await task;
                tasks.Add(task);
            }

            await Task.WhenAll(tasks);

            Console.WriteLine("All Done");
        }

        private static async Task RunClient(int clientIndex)
        {
            using (
            var client = DeviceClient
            .CreateFromConnectionString("insertconnectionstringhere",
                $"testissues_{clientIndex}",
                new ITransportSettings[] {
                    new AmqpTransportSettings(TransportType.Amqp_Tcp_Only)
                    {
                        AmqpConnectionPoolSettings = new AmqpConnectionPoolSettings()
                        {
                            Pooling = true
                        }
                    }})
            )
            {
                client.SetRetryPolicy(retryPolicy);
                Twin twin = await client.GetTwinAsync();

                Console.WriteLine($"Connected client : {clientIndex}");

                for (int i = 0; i < 2; i++)
                {
                    string filename = $"dontdispose/test{i}.txt";
                    Console.WriteLine(filename);
                    var filelocation = await client.GetFileUploadSasUriAsync(new FileUploadSasUriRequest { BlobName = filename },
                        new CancellationTokenSource(1000 * 30).Token);
                    var uri = filelocation.GetBlobUri();

                    Console.WriteLine(JsonConvert.SerializeObject(filelocation));
                    Console.WriteLine(uri);

                    using (var httpClient = new HttpClient())
                    {
                        httpClient.DefaultRequestHeaders.Add("x-ms-blob-type", "BlockBlob");
                        await httpClient.PutAsync(uri, new ByteArrayContent(await File.ReadAllBytesAsync("example.txt")));
                    }

                    Console.WriteLine("File uploaded");

                    await client.CompleteFileUploadAsync(new FileUploadCompletionNotification
                    {
                        CorrelationId = filelocation.CorrelationId,
                        IsSuccess = true,
                        StatusCode = 201,
                        StatusDescription = "successfully uploaded file"
                    });

                    await Task.Delay(100);
                }
            }
        }
    }
```

## Console log of the issue

```
Unhandled exception. System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'The authentication method instance has already been disposed, so this client is no longer usable. Please close and dispose your current client instance. To continue carrying out operations from your device/ module, create a new authentication method instance and use it for reinitializing your client.'.
   at Microsoft.Azure.Devices.Client.AuthenticationWithTokenRefresh.GetTokenAsync(String iotHub)
   at Microsoft.Azure.Devices.Client.Transport.HttpClientHelper.ExecuteAsync(HttpMethod httpMethod, Uri requestUri, Func`3 modifyRequestMessageAsync, Func`2 isSuccessful, Func`3 processResponseMessageAsync, IDictionary`2 errorMappingOverrides, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.Transport.HttpClientHelper.PostAsync[T1,T2](Uri requestUri, T1 entity, IDictionary`2 errorMappingOverrides, IDictionary`2 customHeaders, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.Transport.HttpTransportHandler.GetFileUploadSasUriAsync(FileUploadSasUriRequest request, CancellationToken cancellationToken)
   at IoTHubUploadFile.Program.RunClient(Int32 clientIndex) in C:\Work\Elux\IoTHubUploadFile\IoTHubUploadFile\Program.cs:line 60
   at IoTHubUploadFile.Program.<>c__DisplayClass1_0.<<Main>b__0>d.MoveNext() in C:\Work\Elux\IoTHubUploadFile\IoTHubUploadFile\Program.cs:line 25
   at IoTHubUploadFile.Program.Main(String[] args) in C:\Work\Elux\IoTHubUploadFile\IoTHubUploadFile\Program.cs:line 26
   at IoTHubUploadFile.Program.<Main>(String[] args)
```

azabbasi:
There are two separate issues this PR is tackling.

1- When the last connection in a connection pool is disposed, we need to remove it from the connection pool map.
2- We should not dispose TokenRefereshers in connection pooling scenarios not because they are used by the AMQP connections, because it will be used in the HTTP transport handler for the file upload notification.

fixes #2235


iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
```diff
@@ -99,8 +84,16 @@ public void RemoveAmqpUnit(AmqpUnit amqpUnit)
...
                     AmqpConnectionHolder[] amqpConnectionHolders = ResolveConnectionGroup(deviceIdentity);
                     amqpConnectionHolder = ResolveConnectionByHashing(amqpConnectionHolders, deviceIdentity);
+
+                    amqpConnectionHolder.RemoveAmqpUnit(amqpUnit);
+
+                    // If the connection holder does not have any more units, the entry needs to be nullified.
+                    if (amqpConnectionHolder.IsEmpty())
+                    {
+                        int index = GetDeviceIdentityIndex(deviceIdentity, amqpConnectionHolders.Length);
+                        amqpConnectionHolders[index] = null;
```
abhipsaMisra, line 94: why do we need to do this?

azabbasi: So when the last amqp unit gets removed, the connection pool has a blank slate.

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
```diff
@@ -154,5 +148,12 @@ private AmqpConnectionHolder ResolveConnectionByHashing(AmqpConnectionHolder[] p
 
             return pool[index];
         }
+
+        private static int GetDeviceIdentityIndex(DeviceIdentity deviceIdentity, int poolLength)
+        {
+            return deviceIdentity == null
```
abhipsaMisra, line 154: This is a good null check to have, but by design we would not be getting the index of a device identity that is null, right?

iothub/device/src/Transport/Amqp/AmqpUnitManager.cs
```diff
@@ -47,9 +47,9 @@ internal static AmqpUnitManager GetInstance()
 
         public void RemoveAmqpUnit(AmqpUnit amqpUnit)
         {
-            amqpUnit.Dispose();
             IAmqpUnitManager amqpConnectionPool = ResolveConnectionPool(amqpUnit.GetDeviceIdentity().IotHubConnectionString.HostName);
             amqpConnectionPool.RemoveAmqpUnit(amqpUnit);
+            amqpUnit.Dispose();
```
abhipsaMisra, line 52: 👍🏽

abhipsaMisra: remove and then dispose 👍🏽

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
```diff
@@ -36,21 +36,6 @@ internal class AmqpConnectionPool : IAmqpUnitManager
...
-                    // For this reason, if the device identity of the client is not the one associated with the
-                    // connection, the associated TokenRefresher can be safely disposed.
-                    // Note - This does not cause any identity related issues since the group sas tokens are generated
-                    // against the hub host as the intended audience (without the "device Id").
-                    if (deviceIdentity.AuthenticationModel == AuthenticationModel.SasGrouped
-                        && !ReferenceEquals(amqpConnectionHolder.GetDeviceIdentityOfAuthenticationProvider(), deviceIdentity)
-                        && deviceIdentity.IotHubConnectionString?.TokenRefresher != null
-                        && deviceIdentity.IotHubConnectionString.TokenRefresher.DisposalWithClient)
-                    {
-                        deviceIdentity.IotHubConnectionString.TokenRefresher.Dispose();
```
abhipsaMisra, line 52: Did we make sure that multiplexed clients are always getting their token refresher disposed at their EOL?

azabbasi: Yes, they all get disposed when the device client gets disposed.

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
azabbasi: that is correct, that's why it will throw an exception indicating something went wrong

jamdavi (approved):
Looks good, as discussed will be looking forward to the unit test

iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
```diff
@@ -99,8 +84,17 @@ public void RemoveAmqpUnit(AmqpUnit amqpUnit)
...
                     amqpConnectionHolder = ResolveConnectionByHashing(amqpConnectionHolders, deviceIdentity);
+
+                    amqpConnectionHolder.RemoveAmqpUnit(amqpUnit);
+
+                    // If the connection holder does not have any more units, the entry needs to be nullified.
+                    if (amqpConnectionHolder.IsEmpty())
+                    {
+                        int index = GetDeviceIdentityIndex(deviceIdentity, amqpConnectionHolders.Length);
+                        amqpConnectionHolders[index] = null;
+                        amqpConnectionHolder?.Dispose();
```
jamdavi, line 95: Ok, yeah this is what I was wondering this morning on the call. So we do have to dispose this.

drwill-ms:
> We should not dispose TokenRefereshers in connection pooling scenarios not because they are used by the AMQP connections, because it will be used in the HTTP transport handler for the file upload notification.

When will they be disposed then?

iothub/device/src/IotHubConnectionString.cs
```diff
@@ -92,6 +92,33 @@ public IotHubConnectionString(IotHubConnectionStringBuilder builder)
             }
         }
 
+        // This constructor is only used for unit testing.
+        internal IotHubConnectionString(
+            string ioTHubName = null,
```
drwill-ms, line 97:
```suggestion
            string iotHubName = null,
```

azabbasi:
> > We should not dispose TokenRefereshers in connection pooling scenarios not because they are used by the AMQP connections, because it will be used in the HTTP transport handler for the file upload notification.
> 
> When will they be disposed then?

they will be disposed when the client is disposed.

drwill-ms (approved):

PR #2261, opened 2021-12-17
sdl(all): Create SBOM for net packages 

jamdavi:
We need to create an SBOM to allow customers to see the dependencies of our software. This adds the task to do so.


vsts/build-release-artifacts.yml
```diff
@@ -1,9 +1,10 @@
 # update .csproj file versions and create a release branch/commit
 name: BumpVersion_$(BuildID)_$(Date:yyyyMMdd)$(Rev:.r)
+pool:
```
jamdavi, line 3: The changes that are outside of the added task were validation errors exposed by ADO when testing this new build.

andyk-ms (approved):

Issue #2252, opened 2021-12-07
Debug Assertion with MQTT and User-Supplied SAS Token


timali:
- Windows 10 Desktop x64, v1909, 18363.1500
- UWP app using netstandard2.0
- Windows Desktop PC
- Release 2021-11-01 (Microsoft.Azure.Devices.Client 1.39.0)

A debug assertion is incorrectly raised in iothub\device\src\IotHubConnectionString.Core.cs, line 22, when connecting to the IoT hub under the following conditions:

- The transport is MQTT
- Authentication is performed via a user-supplied SAS token.
- It is a debug build (hence the debug assertion)

Here is how to reproduce it:

1. Create a device client like so:
`client = DeviceClient.Create("IoT hub URI",                        AuthenticationMethodFactory.CreateAuthenticationWithToken("Device ID", "SAS token"),                              Microsoft.Azure.Devices.Client.TransportType.Mqtt);`

2. `await client.OpenAsync();`.

The failing debug assertion is "Debug.Assert(TokenRefresher != null);". When using a user-supplied SAS token, the SDK cannot refresh the token because it does not have the private key, so TokenRefresher should indeed be null in this case.

Should the assertion be something like this instead? `Debug.Assert(!string.IsNullOrWhiteSpace(SharedAccessSignature) || (TokenRefresher != null));`

## Console log of the issue
2021-12-07T11:00:02.8315042 [Microsoft-Azure-Devices-Device-Client-Enter] (InternalClient#1707556, InternalClient_ctor, (Microsoft.Azure.Devices.Client.ITransportSettings[2], DeviceClientPipelineBuilder#15368010)).
2021-12-07T11:00:02.8402831 [Microsoft-Azure-Devices-Device-Client-Info] (PipelineContext#32001227, Set, ITransportSettings[] = Microsoft.Azure.Devices.Client.ITransportSettings[2]).
2021-12-07T11:00:02.8402831 [Microsoft-Azure-Devices-Device-Client-Info] (PipelineContext#32001227, Set, IotHubConnectionString = IotHubConnectionString#19575591).
2021-12-07T11:00:02.8402831 [Microsoft-Azure-Devices-Device-Client-Info] (PipelineContext#32001227, Set, OnMethodCalledDelegate = OnMethodCalledDelegate#41962596).
[...]

PR #2264, opened 2022-01-03
fix(iot-svc): update debug assert

drwill-ms:
For #2252

### DIFF
--- a/iothub/device/src/IotHubConnectionString.Core.cs
+++ b/iothub/device/src/IotHubConnectionString.Core.cs
@@ -19,11 +19,14 @@ namespace Microsoft.Azure.Devices.Client
                 if (Logging.IsEnabled)
                     Logging.Enter(this, $"{nameof(IotHubConnectionString)}.{nameof(IAuthorizationProvider.GetPasswordAsync)}");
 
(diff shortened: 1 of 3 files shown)


iothub/device/src/IotHubConnectionString.Core.cs
```diff
@@ -19,11 +19,11 @@ Task<string> IAuthorizationProvider.GetPasswordAsync()
                 if (Logging.IsEnabled)
                     Logging.Enter(this, $"{nameof(IotHubConnectionString)}.{nameof(IAuthorizationProvider.GetPasswordAsync)}");
 
-                Debug.Assert(TokenRefresher != null);
+                Debug.Assert(string.IsNullOrWhiteSpace(SharedAccessSignature) && TokenRefresher == null, "Both can't be null");
```
timtay-microsoft, line 22:
```suggestion
                Debug.Assert(string.IsNullOrWhiteSpace(SharedAccessSignature) && TokenRefresher == null, "The token refresher and the shared access signature can't both be null");
```

timtay-microsoft (approved):

abhipsaMisra (approved):

PR #2272, opened 2022-01-10
Adding Common 0.7.1

jamdavi:
Adding a reference to DotNetty.Common 0.7.1 to fix the threading issue introduced in 0.7.0.

Ref: #2163 #2194

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -79,6 +79,7 @@
     <Compile Include="ModernDotNet\HsmAuthentication\Transport\HttpRequestResponseSerializer.cs" />
     <Compile Include="ModernDotNet\HsmAuthentication\Transport\UnixDomainSocketEndPoint.cs" />
     <PackageReference Include="DotNetty.Codecs.Mqtt" Version="0.7.0" />
+    <!-- We will need to keep this reference to DotNetty.Common until the next release of DotNetty includes this dependency -->
     <PackageReference Include="DotNetty.Common" Version="0.7.1" />
     <PackageReference Include="DotNetty.Handlers" Version="0.7.0" />
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -79,6 +79,7 @@
     <Compile Include="ModernDotNet\HsmAuthentication\Transport\HttpRequestResponseSerializer.cs" />
     <Compile Include="ModernDotNet\HsmAuthentication\Transport\UnixDomainSocketEndPoint.cs" />
     <PackageReference Include="DotNetty.Codecs.Mqtt" Version="0.7.0" />
+    <PackageReference Include="DotNetty.Common" Version="0.7.1" />
```
drwill-ms, line 82: Do we want to consider adding a comment that this is temporarily needed until DotNetty.Codecs.Mqtt updates its reference?

drwill-ms (approved):

iothub/device/tests/Microsoft.Azure.Devices.Client.Tests.csproj
```diff
@@ -26,6 +26,7 @@
     <Compile Remove="ClientWebSocketTransportTests.cs" />
     <Compile Remove="Mqtt\ClientWebSocketChannelTests.cs" />
     <PackageReference Include="DotNetty.Codecs.Mqtt" Version="0.7.0" />
+    <PackageReference Include="DotNetty.Common" Version="0.7.1" />
```
abhipsaMisra, line 29: q - with the test project having a reference to the source code project, don't we get all these dependencies automatically?

abhipsaMisra (approved):

azabbasi (approved):

iothub/device/tests/Microsoft.Azure.Devices.Client.Tests.csproj
jamdavi: I did this blindly by adding references to any assemblies that referenced DotNetty. I'm not 100% sure if the transitive dependencies overwrite the specific ones from the referenced package.

PR #2274, opened 2022-01-10
Streamline an RBAC test

drwill-ms:
It creates a device to send a message to it but doesn't actually receive that message. Regardless, is that required to ensure the client is authenticated?

Using this PR to sanity check this idea. There are other tests that could be streamlined.

### DIFF
--- a/e2e/test/iothub/TokenCredentialAuthenticationTests.cs
+++ b/e2e/test/iothub/TokenCredentialAuthenticationTests.cs
@@ -124,22 +124,19 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
         public async Task Service_Amqp_TokenCredentialAuth_Success()
         {
             // arrange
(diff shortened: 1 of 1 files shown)


e2e/test/iothub/TokenCredentialAuthenticationTests.cs
```diff
@@ -124,22 +124,19 @@ public async Task DigitalTwinClient_Http_TokenCredentialAuth_Success()
...
             using var message = new Message(Encoding.ASCII.GetBytes("Hello, Cloud!"));
-            await serviceClient.SendAsync(testDevice.Id, message);
 
-            // cleanup
-            await testDevice.RemoveDeviceAsync().ConfigureAwait(false);
+            // act
+            Func<Task> act = async () => await serviceClient.SendAsync(ghostDevice, message).ConfigureAwait(false);
+
+            // assert
+            await act.Should().ThrowAsync<DeviceNotFoundException>();
```
azabbasi, line 139: awesome !

azabbasi (approved):

PR #2275, opened 2022-01-10
Fix doc comment list bullets

drwill-ms:
Per email from Jimaco, these aren't showing up right at https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.provisioning.service.query?view=azure-dotnet#remarks, directed me to spec at https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/examples#math-class-example.

I don't see that we have any description to include; hopefully this works fine.

### DIFF
--- a/provisioning/service/src/Query.cs
+++ b/provisioning/service/src/Query.cs
@@ -3,14 +3,13 @@
 
 using System;
 using System.Collections.Generic;
-using System.Net;
+using System.Globalization;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Common;
-using Newtonsoft.Json;
 using Microsoft.Azure.Devices.Common.Service.Auth;
-using System.Globalization;
+using Newtonsoft.Json;
 
 namespace Microsoft.Azure.Devices.Provisioning.Service
 {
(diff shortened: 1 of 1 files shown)


azabbasi (approved):

JimacoMS4:
Hey David -- Thanks! A few things here:

1. QueryResult.cs and X509Attestation.cs also have the same issues with bullets.

2. If only one child (**term** or **description**) is needed, I'm not sure which one it should be. For example, FeedbackRecord.cs has a bulleted list for the **StatusCode** property, which uses **description** only, and that renders just fine: https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.feedbackrecord.statuscode?view=azure-dotnet#Microsoft_Azure_Devices_FeedbackRecord_StatusCode. 

3. Also, MessageSystemPropertyNames.cs has a bulleted list for the **Ack** property, which only uses **description**. Based on how it renders: https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.messagesystempropertynames.ack?view=azure-dotnet, my guess is that it probably should have used both a **term** and a **description**.

(Note: the only 5 files that I found in the c# repo were the 4 listed above and Query.cs.)

PR #2276, opened 2022-01-11
Update API surface for properties and commands callback APIs

abhipsaMisra:
Changes:
- Remove `userContext` from property update callback API.
- Remove `userContext` from command invocation callback API.

Reasoning:
Since the same callback APIs are invoked for all property/command invocation requests, the user context passed in would be the same for all scenarios. This user context can be set at a class level instead, if required.

### DIFF
--- a/e2e/test/iothub/command/CommandE2ETests.cs
+++ b/e2e/test/iothub/command/CommandE2ETests.cs
@@ -128,36 +128,37 @@ namespace Microsoft.Azure.Devices.E2ETests.Commands
         {
             var commandCallReceived = new TaskCompletionSource<bool>();
 
-            await deviceClient.SubscribeToCommandsAsync(
-                (request, context) =>
-                {
-                    logger.Trace($"{nameof(SetDeviceReceiveCommandAsync)}: DeviceClient command: {request.CommandName}.");
-
-                    try
+            await deviceClient
+                .SubscribeToCommandsAsync(
+                    request =>
                     {
-                        var valueToTest = request.GetData<ServiceCommandRequestObject>();
-                        if (string.IsNullOrEmpty(componentName))
-                        {
-                            Assert.AreEqual(null, request.ComponentName, $"The expected component name should be null but was {request.ComponentName}");
-                        }
-                        else
-                        {
-                            Assert.AreEqual(componentName, request.ComponentName, $"The expected component name should be {componentName} but was {request.ComponentName}");
-                        }
-                        var assertionObject = new ServiceCommandRequestAssertion();
-                        string responseExpected = JsonConvert.SerializeObject(assertionObject);
-                        Assert.AreEqual(responseExpected, request.DataAsJson, $"The expected response payload should be {responseExpected} but was {request.DataAsJson}");
-                        Assert.AreEqual(assertionObject.A, valueToTest.A, $"The expected response object did not decode properly. Value a should be {assertionObject.A} but was {valueToTest?.A ?? int.MinValue}");
-                        commandCallReceived.SetResult(true);
-                    }
-                    catch (Exception ex)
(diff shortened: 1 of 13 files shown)


e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -128,36 +128,37 @@ public static async Task<Task> SetDeviceReceiveCommandAsync(DeviceClient deviceC
...
-                        if (string.IsNullOrEmpty(componentName))
+                        logger.Trace($"{nameof(SetDeviceReceiveCommandAsync)}: DeviceClient command: {request.CommandName}.");
+
+                        try
                         {
-                            Assert.AreEqual(null, request.ComponentName, $"The expected component name should be null but was {request.ComponentName}");
+                            var valueToTest = request.GetPayload<ServiceCommandRequestObject>();
+                            if (string.IsNullOrEmpty(componentName))
+                            {
+                                Assert.AreEqual(null, request.ComponentName, $"The expected component name should be null but was {request.ComponentName}");
```
drwill-ms, line 142: Can we use FluentAssertions instead?

iothub/device/src/CommandRequest.cs
```diff
@@ -52,16 +60,21 @@ public T GetData<T>()
...
+        public ReadOnlyCollection<byte> GetPayloadAsBytes()
         {
             // Need to return a clone of the array so that consumers
             // of this library cannot change its contents.
-            return (byte[])_data.Clone();
+            return _payload;
         }
 
         /// <summary>
         /// The command data in Json format.
```
drwill-ms, line 71: Comment needs updating

iothub/device/src/PayloadCollection.cs
```diff
@@ -20,7 +20,7 @@ public abstract class PayloadCollection : IEnumerable<KeyValuePair<string, objec
         /// <summary>
         /// The underlying collection for the payload.
         /// </summary>
-        public IDictionary<string, object> Collection { get; private set; } = new Dictionary<string, object>();
+        public IDictionary<string, object> Collection { get; } = new Dictionary<string, object>();
```
drwill-ms, line 23: Why do we expose the collection when the containing class `is` an IEnumerable<KeyValuePair>string, object>>?

abhipsaMisra: I should be able to make this collection internal. I don't see a need for users to be accessing this explicitly.

azabbasi (approved):

e2e/test/iothub/command/CommandE2ETests.cs
```diff
@@ -120,44 +121,45 @@ public static async Task DigitalTwinsSendCommandAndVerifyResponseAsync(string de
...
-                        if (string.IsNullOrEmpty(componentName))
+                        logger.Trace($"{nameof(SetDeviceReceiveCommandAsync)}: DeviceClient command: {request.CommandName}.");
+
+                        try
                         {
-                            Assert.AreEqual(null, request.ComponentName, $"The expected component name should be null but was {request.ComponentName}");
+                            var valueToTest = request.GetPayload<ServiceCommandRequestObject>();
+                            if (string.IsNullOrEmpty(componentName))
+                            {
+                                request.ComponentName.Should().BeNull();
```
drwill-ms, line 143: So `componentName` could be whitespace (legally) but we'd still expect `request.ComponentName` to be null?

drwill-ms (commented):
LGTM

e2e/test/iothub/command/CommandE2ETests.cs
abhipsaMisra: This can be simplified, for sure.

PR #2277, opened 2022-01-11
Exclude low pri .net targets in PR builds

drwill-ms:
Using info from [this guide](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=example%2Cparameter-schema)

image


provisioning/service/src/Query.cs
```diff
@@ -22,8 +22,8 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
     ///     <item>
     ///         <description>
     ///         <see cref="ProvisioningServiceClient.CreateIndividualEnrollmentQuery(QuerySpecification, int)">IndividualEnrollment</see>
-        ///     </description>
```
drwill-ms, line 25: I noticed I left this with bad tabbing, somehow.

abhipsaMisra (approved):

azabbasi (approved):

PR #2278, opened 2022-01-11
PnP doc comment updates

drwill-ms:


iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.Devices.Client
     public partial class DeviceClient : IDisposable
     {
         /// <summary>
-        /// The <see cref="PayloadConvention"/> that the client uses for convention-based operations.
```
drwill-ms, line 19: No need to reference the type we are, as the web docs will automatically link to types.

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -47,7 +47,7 @@ public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationTo
         /// Retrieve the client properties.
         /// </summary>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
-        /// <returns>The device properties.</returns>
```
drwill-ms, line 50: Device or module => client, right?

iothub/device/src/PayloadCollection.cs
```diff
@@ -28,18 +28,16 @@ public abstract class PayloadCollection : IEnumerable<KeyValuePair<string, objec
         public PayloadConvention Convention { get; internal set; }
 
         /// <summary>
-        /// Get the value at the specified key.
+        /// Gets the value with the specified key.
```
drwill-ms, line 31: I checked c# dictionary web docs to conform to their verbiage.

iothub/device/src/PayloadCollection.cs
```diff
@@ -114,6 +112,8 @@ public bool Contains(string key)
         /// <returns>True if a value of type <c>T</c> with the specified key was found; otherwise, it returns false.</returns>
         public bool TryGetValue<T>(string key, out T value)
         {
+            value = default;
```
drwill-ms, line 115: Best practice is to set an out parameter right away, so you are sure it is set no matter the code path. Then we don't need DRY violations below in the method either.

shared/src/DefaultPayloadConvention.cs
```diff
@@ -14,7 +14,7 @@ public sealed class DefaultPayloadConvention : PayloadConvention
         /// <summary>
         /// A static instance of this class.
         /// </summary>
-        public static readonly DefaultPayloadConvention Instance = new DefaultPayloadConvention();
+        public static DefaultPayloadConvention Instance { get; } = new DefaultPayloadConvention();
```
drwill-ms, line 17: Properties not fields in public APIs

abhipsaMisra (approved):

azabbasi (approved):

PR #2282, opened 2022-01-12
Update IDeviceIdentity interface to add doc comments.

azabbasi:


timtay-microsoft (approved):

iothub/device/src/Transport/IDeviceIdentity.cs
```diff
@@ -3,15 +3,44 @@
 
 namespace Microsoft.Azure.Devices.Client.Transport
 {
+    /// <summary>
+    /// Interface for device configurations and common attributes.
+    /// </summary>
     internal interface IDeviceIdentity
```
abhipsaMisra, line 9: since this looks like it is specific to amqp only; if we haven't released yet we could move this to the amqp namespace

abhipsaMisra, line 9: q - is this for device clients only and not for module clients?

azabbasi: Sure thing, I will inspect other places it has been used and if they are all in AMQP I will move it

azabbasi: It appears so. Only in AMQP and Device connections only.

azabbasi: Done

iothub/device/src/Transport/Amqp/IDeviceIdentity.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        AmqpTransportSettings AmqpTransportSettings { get; }
+
+        /// <summary>
+        /// Device connection string details.
+        /// </summary>
+        IotHubConnectionString IotHubConnectionString { get; }
+
+        /// <summary>
+        /// Device details and information.
```
drwill-ms, line 27: This is environment (OS), .NET version info, and client library version info, right? Is it really info about the device?

iothub/device/src/Transport/Amqp/IDeviceIdentity.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        ClientOptions Options { get; }
+
+        /// <summary>
+        /// Device authentication audience.
+        /// </summary>
+        string Audience { get; }
+
+        /// <summary>
+        /// Whether or not Device is part of a connection pooling.
```
drwill-ms, line 42:
```suggestion
        /// Whether or not the device is part of a connection pooling.
```

drwill-ms (approved):

iothub/device/src/Transport/Amqp/IDeviceIdentity.cs
azabbasi: updating.

abhipsaMisra (approved):

PR #2290, opened 2022-01-26
Update service clients instantiation doc comments

drwill-ms:


iothub/service/src/RegistryManager.cs
```diff
@@ -70,48 +70,30 @@ public class RegistryManager : IDisposable
         private IHttpClientHelper _httpClientHelper;
 
         /// <summary>
-        /// Creates an instance of <see cref="RegistryManager"/>, provided for unit testing purposes only.
```
drwill-ms, line 73: No need to self-ref to the same topic we're already on.

iothub/service/src/RegistryManager.cs
```diff
@@ -70,48 +70,30 @@ public class RegistryManager : IDisposable
...
+        /// Creates an instance of RegistryManager, provided for unit testing purposes only.
         /// Use the CreateFromConnectionString method to create an instance to use the client.
         /// </summary>
         public RegistryManager()
         {
         }
 
-        internal RegistryManager(IotHubConnectionProperties connectionProperties, HttpTransportSettings transportSettings)
```
drwill-ms, line 80: Moved internals below publics.

iothub/service/src/RegistryManager.cs
azabbasi: The only benefit of self-ref would be to get consistency if we ever rename this class. hint hint

azabbasi (approved):

iothub/service/src/RegistryManager.cs
drwill-ms: Meh, cost of user confusion is worse IMO.

timtay-microsoft (approved):

azabbasi (approved):

PR #2292, opened 2022-01-27
Fix class field constants casing

drwill-ms:


andyk-ms (approved):

azabbasi (approved):

iothub/service/src/ServiceClient.cs
```diff
@@ -215,6 +213,8 @@ public static ServiceClient CreateFromConnectionString(string connectionString,
 
 #endif
 
+        internal IotHubConnection Connection { get; }
```
abhipsaMisra, line 216: properties before methods?

abhipsaMisra (approved):

iothub/service/src/ServiceClient.cs
drwill-ms: Yes, but I'd say not before static factory methods, which are more like ctors. Would you agree?

PR #2296, opened 2022-02-01
Update iteration path in vsts.yaml

drwill-ms:
It looks like someone inserted a path part called "Custom" in our iteration paths.

### DIFF
--- a/vsts/vsts.yaml
+++ b/vsts/vsts.yaml
@@ -538,7 +538,7 @@ jobs:
           instanceUrlForTsaV2: 'MSAZURE'
           projectNameMSAZURE: 'One'
           areaPath: 'One\IoT\Platform and Devices\IoT Devices\SDKs\Managed'
-          iterationPath: 'One\IoT\Backlog'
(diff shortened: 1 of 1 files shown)


vsts/vsts.yaml
```diff
@@ -538,7 +538,7 @@ jobs:
           instanceUrlForTsaV2: 'MSAZURE'
           projectNameMSAZURE: 'One'
           areaPath: 'One\IoT\Platform and Devices\IoT Devices\SDKs\Managed'
-          iterationPath: 'One\IoT\Backlog'
+          iterationPath: 'One\Custom\IoT\Backlog'
```
azabbasi, line 541: Interesting, I thought they wanted it to be set to the root of the project iteration path aka `One`

azabbasi (approved):

vsts/vsts.yaml
drwill-ms: That wouldn't be under IoT path, though. I just moved it to the renamed path.

PR #2297, opened 2022-02-01
Feature/dps rbac support

dylanbulfinMS:
## Checklist


e2e/test/config/TestConfiguration.Provisioning.cs
```diff
@@ -38,6 +77,35 @@ public static X509Certificate2Collection GetGroupEnrollmentChain()
...
+                string token = string.Format(
+                    CultureInfo.InvariantCulture,
+                    "SharedAccessSignature sr={0}&sig={1}&se={2}",
+                    WebUtility.UrlEncode(resourceUri),
+                    WebUtility.UrlEncode(signature),
+                    expiry);
+
+                if (!string.IsNullOrWhiteSpace(policyName))
+                {
+                    token += "&skn=" + policyName;
```
drwill-ms, line 104: Might policy name need to be URL encoded?

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -610,13 +610,14 @@ if ($certExists)
     az iot dps certificate delete -g $ResourceGroup --dps-name $dpsName --name $uploadCertificateName --etag $etag
 }
 Write-Host "`nUploading new certificate to DPS."
-az iot dps certificate create -g $ResourceGroup --path $rootCertPath --dps-name $dpsName --certificate-name $uploadCertificateName --output none
```
drwill-ms, line 613: Why remove `--output none`?

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -610,13 +610,14 @@ if ($certExists)
...
+az iot dps certificate create -g $ResourceGroup --path $rootCertPath --dps-name $dpsName --certificate-name $uploadCertificateName
 
 $isVerified = az iot dps certificate show -g $ResourceGroup --dps-name $dpsName --certificate-name $uploadCertificateName --query 'properties.isVerified' --output tsv
 if ($isVerified -eq 'false')
 {
     Write-Host "`nVerifying certificate uploaded to DPS."
     $etag = az iot dps certificate show -g $ResourceGroup --dps-name $dpsName --certificate-name $uploadCertificateName --query 'etag'
+    Write-Host "`n az iot dps certificate generate-verification-code -g $ResourceGroup --dps-name $dpsName --certificate-name $uploadCertificateName -e $etag --query 'properties.verificationCode'"
```
drwill-ms, line 620: Why print this?

provisioning/service/src/Auth/ProvisioningSasCredential.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.Azure.Devices.Common.Service.Auth;
+
+namespace Microsoft.Azure.Devices.Provisioning.Service.Auth
+{
+    /// <summary>
+    /// Allows authentication to the API using a Shared Access Key provided by custom implementation.
+    /// The PnP client is auto generated from swagger and needs to implement a specific class to pass to the protocol layer
+    /// unlike the rest of the clients which are hand-written. So, this implementation for authentication is specific to digital twin (PnP).
+    /// </summary>
+    internal class ProvisioningSasCredential: IAuthorizationHeaderProvider
```
drwill-ms, line 13:
```suggestion
    internal class ProvisioningSasCredential : IAuthorizationHeaderProvider
```

provisioning/service/src/Auth/ProvisioningSasCredential.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Azure;
+using Microsoft.Azure.Devices.Common.Service.Auth;
+
+namespace Microsoft.Azure.Devices.Provisioning.Service.Auth
+{
+    /// <summary>
+    /// Allows authentication to the API using a Shared Access Key provided by custom implementation.
+    /// The PnP client is auto generated from swagger and needs to implement a specific class to pass to the protocol layer
```
drwill-ms, line 10: I don't think this "PnP client" comment is accurate here, is it?

provisioning/service/src/Auth/ProvisioningTokenCredential.cs
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Threading;
+using Azure.Core;
+using Microsoft.Azure.Devices.Common.Service.Auth;
+
+namespace Microsoft.Azure.Devices.Provisioning.Service.Auth
+{
+    /// <summary>
+    /// Allows authentication to the API using a JWT token generated for Azure active directory.
+    /// The PnP client is auto generated from swagger and needs to implement a specific class to pass to the protocol layer
```
drwill-ms, line 11: Same here, copy pasta error?

provisioning/service/src/Config/IndividualEnrollment.cs
```diff
@@ -212,6 +212,12 @@ public override string ToString()
         [JsonProperty(PropertyName = "registrationState", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public DeviceRegistrationState RegistrationState { get; private set; }
 
+        /// <summary>
+        /// Optional Device Information.
```
drwill-ms, line 216: What kind of information would a user put or read from here? How will it be processed? How should it be interpreted?

provisioning/service/src/ProvisioningServiceClient.cs
```diff
@@ -104,6 +111,75 @@ public static ProvisioningServiceClient CreateFromConnectionString(string connec
             return new ProvisioningServiceClient(connectionString, httpTransportSettings);
         }
 
+        /// <summary>
+        /// Create a new instance of the <code>ProvisioningServiceClient</code> from the provided hostName and <code>TokenCredential</code>
```
drwill-ms, line 115: You can use this alias for code that will reduce line length by a reasonable amount, especially since we're using it several times.

```suggestion
        /// Create a new instance of the <c>ProvisioningServiceClient</c> from the provided hostName and <c>TokenCredential</c>
```

provisioning/service/src/ProvisioningServiceClient.cs
```diff
@@ -121,12 +197,43 @@ private ProvisioningServiceClient(string connectionString, HttpTransportSettings
...
                 _provisioningConnectionString.HttpsEndpoint,
                 _provisioningConnectionString,
                 httpTransportSettings);
         }
 
+        private ProvisioningServiceClient(string hostName, TokenCredential credential, HttpTransportSettings httpTransportSettings)
+        {
+            if (string.IsNullOrWhiteSpace(hostName ?? throw new ArgumentNullException(nameof(hostName))))
+            {
+                throw new ArgumentException($"{nameof(hostName)} cannot be empty string");
```
drwill-ms, line 211: `ArgumentException` has a ctor combo that takes the name of the parameter independent of the message. It is preferred to use.

provisioning/service/src/Query.cs
```diff
@@ -80,50 +80,86 @@ public class Query : IDisposable
...
+            CancellationToken cancellationToken)
+        {
+            if (hostName == null)
+            {
+                throw new ArgumentNullException(nameof(hostName));
+            }
+
+            if (string.IsNullOrWhiteSpace(serviceName ?? throw new ArgumentNullException(nameof(serviceName))))
+            {
+                throw new ArgumentException($"{nameof(serviceName)} cannot be an empty string");
```
drwill-ms, line 136: same here, perhaps later on too

drwill-ms (changes requested):
Inline comments

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
dylanbulfinMS: That was for local debugging of the script, wasn't supposed to be committed, fixed now

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
dylanbulfinMS: Same as above

provisioning/service/src/ProvisioningServiceClient.cs
dylanbulfinMS: Updated

provisioning/service/src/Config/IndividualEnrollment.cs
dylanbulfinMS: It's basically just a Map, it's for arbitrary information a customer might want to store so there's no real specific use case. It was added to reconcile the SDK with the current API version since I updated it. I can move it into a different PR if that would be better since it's not strictly related to RBAC

dylanbulfinMS: moved to https://github.com/Azure/azure-iot-sdk-csharp/pull/2299

drwill-ms: The reason I ask is for our customers. The doc comments should give them more info. How should I (as an SDK user) use this? With this SDK, do I set these key value pairs? Are they interpreted by the DPS service? Will the device have access to them? It would be good to have helpful doc comments to guide the customer.

e2e/test/config/TestConfiguration.Provisioning.cs
drwill-ms:
```suggestion
                    token += "&skn=" + WebUtility.UrlEncode(policyName);
```

drwill-ms: Policy names can be supplied by the user, right? What if they include a character that needs to be URL encoded?

dylanbulfinMS: Oh, since this is only used in e2e testing I didn't think about that, changed now

PR #2302, opened 2022-02-09
Update QoS settings to AtLeastOnce for mqtt layer subscriptions

abhipsaMisra:
The IoT hub - MQTT spec, for reference: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -33,7 +33,9 @@ using Microsoft.Azure.Devices.Shared;
 using Newtonsoft.Json;
 
 #if NET5_0
+
 using TaskCompletionSource = System.Threading.Tasks.TaskCompletionSource;
+
 #else
 using TaskCompletionSource = Microsoft.Azure.Devices.Shared.TaskCompletionSource;
 #endif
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -168,7 +171,8 @@ internal sealed class MqttTransportHandler : TransportHandler, IMqttIotHubEventH
             _deviceboundMessageFilter = string.Format(CultureInfo.InvariantCulture, DeviceBoundMessagesTopicFilter, iotHubConnectionString.DeviceId);
             _deviceboundMessagePrefix = string.Format(CultureInfo.InvariantCulture, DeviceBoundMessagesTopicPrefix, iotHubConnectionString.DeviceId);
 
-            _qos = settings.PublishToServerQoS;
+            _qosSendPacketToService = settings.PublishToServerQoS;
+            _qosReceivePacketFromService = settings.ReceivingQoS;
```
drwill-ms, line 175: What change in behavior might this have for customers?

abhipsaMisra: Not much, really. 
For customers who don't set this value in settings, the default for both is 1 - they wouldn't see any difference. 
For the ones that do set it explicitly to something else, they still wouldn't see much of a difference either since IoT hub mqtt implementation isn't a true broker. For us, it means that receive C2D messaging is the only call that really looks at the QoS. For the other calls, the service doesn't call out how to configure it. What that means is, even if we subscribe to twin property updates with a QoS of 1, service might still be publishing the packets with QoS of 0. They don't state the contract as a part of the hub-mqtt implementation.
https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support

drwill-ms (approved):

PR #2319, opened 2022-03-10
Refactor AsyncExecution

drwill-ms:


iothub/device/src/TransientFaultHandling/AsyncExecution.cs
```diff
@@ -88,7 +88,7 @@ private static Task<bool> StartAsGenericTask(Func<Task> taskAction)
                     nameof(taskAction));
             }
 
-            var tcs = new TaskCompletionSource<bool>();
+            TaskCompletionSource<bool> tcs = CreateTaskCompletionSource();
```
drwill-ms, line 91: Centralize `TaskCompletionSource` initialization to specify the preferred value (https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously) upon initialization.

iothub/device/src/TransientFaultHandling/AsyncExecution[T].cs
```diff
@@ -37,108 +37,128 @@ internal class AsyncExecution<TResult>
...
-        private readonly CancellationToken cancellationToken;
-
-        private readonly Stopwatch stopwatch;
-
-        private Task<TResult> previousTask;
-
-        private int retryCount;
-
-        public AsyncExecution(Func<Task<TResult>> taskFunc, ShouldRetry shouldRetry, Func<Exception, bool> isTransient, Action<int, Exception, TimeSpan> onRetrying, bool fastFirstRetry, CancellationToken cancellationToken)
+        protected static readonly TaskContinuationOptions s_taskContinuationOption =
```
Comment (drwill-ms) [inline iothub/device/src/TransientFaultHandling/AsyncExecution[T].cs:40]: `RunContinuationAsynchronously` does not exist in NET451. It will retain existing behavior.

Also note that the guidance (https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously) specifies a similar change (not this one) with the constructor, but doesn't say anything about this parameter to `Task.ContinueWith`. I assume we should change them both, but I'm not sure. Leaving the old value seems wrong, though.

iothub/device/src/TransientFaultHandling/AsyncExecution[T].cs
```diff
@@ -149,30 +169,48 @@ private Task<TResult> ExecuteAsyncContinueWith(Task<TResult> runningTask)
...
             }
-            return this.ExecuteAsyncImpl(null);
+
+            return ExecuteAsyncImpl(null);
+        }
+
+        protected static TaskCompletionSource<TResult> CreateTaskCompletionSource()
+        {
+            return new TaskCompletionSource<TResult>(
+#if !NET451
```
Comment (drwill-ms) [inline iothub/device/src/TransientFaultHandling/AsyncExecution[T].cs:210]: Only pass this parameter if not NET451.

PR #2326, opened 2022-03-28
Update readme.md on LTS info

andyk-ms:
Clarification to the LTS versions:
1) Updated Readme to keep only major version listed. Customer should be able to see patches on the LTS branch and select the patch level they needed.
2) Changed the maintenance end date to "current" so in case the last good LTS is the current best version that will continue to receive patches.
3) Removed LTS End Date that is not actually enforceable from the customer perspective.

## Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -204,10 +204,10 @@ LTS releases may include additional extended support for security bug fixes as l
 This table shows previous LTS releases and end dates.
 
 | Release                                                                                                                        | LTS Start Date | Maintenance End Date |
-| :----------------------------------------------------------------------------------------------------------------------------: | :------------: | :---------------------: |
-| [2021-3-18](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2021-3-18)                                          | 2021-03-18     | current                  |
-| [2020-8-19](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-8-19)                                          | 2020-08-19     | 2021-08-19              |
-| [2020-1-31](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-1-31)                                          | 2020-01-31     | 2021-01-30              |
+| :----------------------------------------------------------------------------------------------------------------------------: | :------------: | :------------------: |
+| [2021-3-18](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2021-3-18)                                          | 2021-03-18     | current               |
+| [2020-8-19](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-8-19)                                          | 2020-08-19     | 2021-08-19            |
+| [2020-1-31](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-1-31)                                          | 2020-01-31     | 2021-01-30            |
 
 > Note some releases have extended end dates.
 
--- a/readme.md
+++ b/readme.md
@@ -205,7 +205,7 @@ This table shows previous LTS releases and end dates.
 
 | Release                                                                                                                        | LTS Start Date | Maintenance End Date |
 | :----------------------------------------------------------------------------------------------------------------------------: | :------------: | :------------------: |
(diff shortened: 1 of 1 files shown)


azabbasi:
The table format is busted I think
image

andyk-ms:
> The table format is busted I think image

Thanks, missed the spacing and extra column. Fixed.

readme.md
```diff
@@ -203,18 +203,11 @@ LTS releases may include additional extended support for security bug fixes as l
...
-| [2020-8-19](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-8-19)                                          | 2020-08-19     | 2021-08-19           | 2023-08-19   |
-| [2020-4-3](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-1-31_patch1) <sub>patch 1 of 2020-01-31</sub>   | 2020-04-03     | 2021-01-30           | 2023-01-30   |
-| [2020-1-31](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-1-31)                                          | 2020-01-31     | 2021-01-30           | 2023-01-30   |
+| Release                                                                                                                        | LTS Start Date | Maintenance End Date |
+| :----------------------------------------------------------------------------------------------------------------------------: | :------------: | :------------------: |
+| [2021-3-18](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2021-3-18)                                          | 2021-03-18     | current              |
```
azabbasi, line 208: This is the initial LTS release, we have patched it 4 times so the latest LTS release is
https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2021_03_18-patch4

azabbasi: We can potentially also add a link to the current LTS branch itself. Something to think about. For example, the latest LTS branch is :
https://github.com/Azure/azure-iot-sdk-csharp/tree/lts_2021_03

abhipsaMisra: Instead of linking the tags it would help if we linked the branches directly. There isn't an easy way to navigate from one tag to another (for the LTS patches) without going through the release page or finding a common link (through the tag prefix, etc.).

drwill-ms: I thought our plan was to not keep/maintain branches, rather to use tags.

abhipsaMisra: Oh, I thought we still had branches for LTS releases but all other non-LTS releases were off of tags.

abhipsaMisra: Yeah, we have branches (and tags) for LTS releases and only tags for all other releases.
image

drwill-ms: k

readme.md
```diff
@@ -203,18 +203,11 @@ LTS releases may include additional extended support for security bug fixes as l
...
-| [2020-8-19](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-8-19)                                          | 2020-08-19     | 2021-08-19           | 2023-08-19   |
-| [2020-4-3](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-1-31_patch1) <sub>patch 1 of 2020-01-31</sub>   | 2020-04-03     | 2021-01-30           | 2023-01-30   |
-| [2020-1-31](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2020-1-31)                                          | 2020-01-31     | 2021-01-30           | 2023-01-30   |
+| Release                                                                                                                        | LTS Start Date | Maintenance End Date |
+| :----------------------------------------------------------------------------------------------------------------------------: | :------------: | :------------------: |
+| [2022-01-18](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/lts_2021-3-18-patch4) <sub>patch 4 of 2021-03-18</sub> | 2021-03-18     | current              |
```
azabbasi, line 208:
```suggestion
| [2022-01-18](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2021_03_18-patch4) <sub>patch 4 of 2021-03-18</sub> | 2021-03-18     | current              |
```

azabbasi (approved):

readme.md
andyk-ms: created a new tag that is consistent with the previous tags and updated the URL

readme.md
andyk-ms: will use tag with consistent tag name

PR #2328, opened 2022-03-31
DPS client certificate issuance over MQTT

abhipsaMisra:
This PR covers the implementation idea for DPS client certificate issuance over MQTT protocol for individual enrollments.

The DPS client certificate feature is built on the following idea:	
- you have a CA that can accept certificate signing requests from you and issue you with a signed public certificate.
    - test CA provided by DPS team for now. We might need to create a separate test account for our test pipeline
- you link your DPS instance to your CA
   - service API (not implemented here). The tests are written based on this link already being in place.
- you link your DPS enrollments to your CA 
    - service API (implemented via service API).
- generate your public-private key pair and corresponding certificate signing request 
    - implemented va `openssl` commands for the tests
- register your enrollment passing in the certificate request. This certificate request contains your public key and "your" identifying information (subject name, etc). DPS receives this request, forwards it to your linked CA, receives a signed public certificate in return. DPS will provision your enrollment to IoT hub with this signed public certificate and also return it to you.
    - implemented over MQTT for nowby the SDK. Service support is available over all three protocols.
- use the issued certificate and previously generate key pair and generate your pfx file. This pfx file contains your device's identifying information and the private key necessary to negotiate the TLS handshake 
    - implemented va `openssl` commands for the tests
- use the pfx file to create an `X509Certificate2` which can be used to authenticate with IoT Hub using `DeviceAuthenticationWithX509Certificate` authentication method.
    - support already available on the SDK

API diff

```diff
{
     namespace Microsoft.Azure.Devices.Provisioning.Client {
         public class DeviceRegistrationResult {
+           public string IssuedClientCertificate { get; set; }
         }
         public class ProvisioningRegistrationAdditionalData {
+          public string OperationalCertificateRequest { get; set; }
         }
     }
     namespace Microsoft.Azure.Devices.Provisioning.Client.Transport {
         public class ProvisioningTransportRegisterMessage {
+          public string OperationalCertificateRequest { get; private set; }
         }
     }
 }
```

```diff
 {
     namespace Microsoft.Azure.Devices.Provisioning.Service {
+       public class ClientCertificateIssuancePolicy {
+          public ClientCertificateIssuancePolicy();
+          public string CertificateAuthorityName { get; set; }
+       }
         public class IndividualEnrollment : IETagHolder {
+          public ClientCertificateIssuancePolicy ClientCertificateIssuancePolicy { get; set; }
         }
     }
 }
```

### DIFF
--- a/common/src/device/provisioning/transport/DeviceRegistrationResult.cs
+++ b/common/src/device/provisioning/transport/DeviceRegistrationResult.cs
@@ -14,128 +14,65 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Models
     [SuppressMessage("Microsoft.Performance", "CA1812", Justification = "Used by the JSon parser.")]
     internal partial class DeviceRegistrationResult
     {
-        public DeviceRegistrationResult()
-        {
-            CustomInit();
-        }
-
-        internal DeviceRegistrationResult(
-            TpmRegistrationResult tpm = default(TpmRegistrationResult),
-            X509RegistrationResult x509 = default(X509RegistrationResult),
-            SymmetricKeyRegistrationResult symmetricKey = default(SymmetricKeyRegistrationResult),
-            string registrationId = default(string),
-            DateTime? createdDateTimeUtc = default(DateTime?),
-            string assignedHub = default(string),
-            string deviceId = default(string),
-            string status = default(string),
-            string substatus = default(string),
-            string generationId = default(string),
-            DateTime? lastUpdatedDateTimeUtc = default(DateTime?),
-            int? errorCode = default(int?),
-            string errorMessage = default(string),
-            string etag = default(string),
-            JRaw payload = default(JRaw))
-        {
-            Tpm = tpm;
-            X509 = x509;
-            SymmetricKey = symmetricKey;
-            RegistrationId = registrationId;
-            CreatedDateTimeUtc = createdDateTimeUtc;
-            AssignedHub = assignedHub;
-            DeviceId = deviceId;
-            Status = status;
-            Substatus = substatus;
-            GenerationId = generationId;
-            LastUpdatedDateTimeUtc = lastUpdatedDateTimeUtc;
-            ErrorCode = errorCode;
-            ErrorMessage = errorMessage;
-            Etag = etag;
-            Payload = payload;
-            CustomInit();
-        }
-
-        /// <summary>
-        /// An initialization method that performs custom operations like setting defaults
-        /// </summary>
-        partial void CustomInit();
-
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "tpm")]
         public TpmRegistrationResult Tpm { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "x509")]
         public X509RegistrationResult X509 { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "symmetricKey")]
         public SymmetricKeyRegistrationResult SymmetricKey { get; set; }
 
-
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "registrationId")]
         public string RegistrationId { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "createdDateTimeUtc")]
         public DateTime? CreatedDateTimeUtc { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "assignedHub")]
         public string AssignedHub { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "deviceId")]
         public string DeviceId { get; set; }
 
         /// <summary>
-        /// Gets or sets possible values include: 'unassigned', 'assigning',
-        /// 'assigned', 'failed', 'disabled'
+        /// The status of the operation.
+        /// Possible values include: 'unassigned', 'assigning', 'assigned', 'failed', 'disabled'
         /// </summary>
         [JsonProperty(PropertyName = "status")]
         public string Status { get; set; }
 
         /// <summary>
+        /// The substatus of the operation.
         /// Possible values include: 'initialAssignment', 'deviceDataMigrated', 'deviceDataReset'
         /// </summary>
         [JsonProperty(PropertyName = "substatus")]
         public string Substatus { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "generationId")]
         public string GenerationId { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "lastUpdatedDateTimeUtc")]
         public DateTime? LastUpdatedDateTimeUtc { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "errorCode")]
         public int? ErrorCode { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "errorMessage")]
         public string ErrorMessage { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "etag")]
         public string Etag { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "payload")]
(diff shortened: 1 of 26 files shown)


e2e/test/provisioning/ReprovisioningE2ETests.cs
```diff
@@ -420,7 +420,7 @@ private async Task<SecurityProvider> CreateSecurityProviderFromName(AttestationM
                             return new SecurityProviderSymmetricKey(registrationIdSymmetricKey, primaryKeyIndividual, secondaryKeyIndividual);
 
                         case EnrollmentType.Individual:
-                            IndividualEnrollment symmetricKeyEnrollment = await CreateIndividualEnrollment(provisioningServiceClient, AttestationMechanismType.SymmetricKey, reprovisionPolicy, allocationPolicy, customAllocationDefinition, iothubs, capabilities).ConfigureAwait(false);
+                            IndividualEnrollment symmetricKeyEnrollment = await CreateIndividualEnrollmentAsync(provisioningServiceClient, AttestationMechanismType.SymmetricKey, reprovisionPolicy, allocationPolicy, customAllocationDefinition, iothubs, capabilities).ConfigureAwait(false);
```
azabbasi, line 423: probably should break this one into multiple lines as well.

common/src/device/provisioning/transport/DeviceRegistrationResult.cs
```diff
@@ -14,128 +14,52 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Models
...
 
-        /// <summary>
-        /// Gets or sets possible values include: 'unassigned', 'assigning',
-        /// 'assigned', 'failed', 'disabled'
-        /// </summary>
         [JsonProperty(PropertyName = "status")]
         public string Status { get; set; }
 
-        /// <summary>
-        /// Possible values include: 'initialAssignment', 'deviceDataMigrated', 'deviceDataReset'
```
azabbasi, line 106: I realize this is an internal class but some of these xml comments might be useful, why remove them?

abhipsaMisra: I'll add the status and substatus comments back in and put in a note for the certificate. Every other property seems self-explanatory, right?

timtay-microsoft:
> ```diff
>  {
>      namespace Microsoft.Azure.Devices.Provisioning.Service {
> +       public class ClientCertificateIssuancePolicy {
> +          public ClientCertificateIssuancePolicy();
> +          public string CertificateAuthorityName { get; set; }
> +       }
>          public class IndividualEnrollment : IETagHolder {
> +          public ClientCertificateIssuancePolicy ClientCertificateIssuancePolicy { get; set; }
>          }
>      }
>  }
> ```

Is this feature limited to Individual enrollments? What about enrollment groups?

abhipsaMisra:
> Is this feature limited to Individual enrollments? What about enrollment groups?

This isn't limited to individual enrollments. This feature is available for enrollment groups as well, but I haven't tested them yet. Enrollment groups will follow in the next PR.

common/src/device/provisioning/transport/DeviceRegistrationResult.cs
```diff
@@ -14,128 +14,64 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Models
...
         [JsonProperty(PropertyName = "etag")]
         public string Etag { get; set; }
 
-        /// <summary>
-        /// </summary>
         [JsonProperty(PropertyName = "payload")]
         public JRaw Payload { get; set; }
+
+        /// <summary>
+        /// The PEM encoded operational client certificate that was returned by the CA. This operational certificate was used by DPS to register the enrollment with IoT Hub.
```
timtay-microsoft, line 71:
```suggestion
        /// The PEM encoded operational client certificate that was returned by the certificate authority. This operational certificate was used by DPS to register the enrollment with IoT Hub.
```

I'd avoid using acronyms when possible since users may not be familiar with them.

common/src/device/provisioning/transport/DeviceRegistrationResult.cs
```diff
@@ -14,128 +14,64 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Models
...
+        /// The IoT device can use this operational certificate to authenticate with IoT Hub.
+        /// </summary>
+        [JsonProperty(PropertyName = "issuedClientCertificate")]
+        public string IssuedClientCertificate { get; set; }
```
timtay-microsoft, line 75: Is there any reason why this needs to have a public setter? It sounds like this is more of a read-only field

abhipsaMisra: You're right, this is a read-only field and should not require a public setter. However, newtonsoft json requires the properties to be public for it to be able to deserialize them.

provisioning/device/src/ProvisioningRegistrationAdditionalData.cs
```diff
@@ -11,8 +10,15 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
...
+        /// Additional (optional) Json Data to be sent to the service
         /// </summary>
         public string JsonData { get; set; }
+
+        /// <summary>
+        /// The PEM encoded operational certificate request that DPS will send to its linked CA and the CA will sign and return an X509 device identity operational client certificate to the device.
+        /// DPS will register the device and operational client certificate thumbprint in IoT Hub and return the certificate to the IoT device.
+        /// The IoT device can then use the operational certificate to authenticate with IoT Hub.
+        /// </summary>
+        public string OperationalCertificateRequest { get; set; }
```
timtay-microsoft, line 22: This feels like an odd spot to put this new field. This class, as I understand it, is for user defined data only and for users who are using custom webhooks for provisioning. Maybe this class could be renamed in v2 ("RegistrationOptions" or something), but right now this feels misplaced

provisioning/transport/generated/src/Models/DeviceRegistration.cs
```diff
@@ -17,28 +18,20 @@ internal partial class DeviceRegistration
...
-        /// An initialization method that performs custom operations like setting defaults
+        /// Gets or set the custom content payload.
         /// </summary>
-        partial void CustomInit();
+        [JsonProperty(PropertyName = "payload", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public JRaw Payload { get; set; }
 
         /// <summary>
-        /// Gets or set the custom content payload.
+        /// The PEM encoded operational certificate request that DPS will send to its linked CA and the CA will sign and return an X509 device identity operational client certificate to the device.
```
timtay-microsoft, line 30:
```suggestion
        /// The PEM encoded operational certificate request that DPS will send to its linked certificate authority which will sign and return an X509 device identity operational client certificate to the device.
```

provisioning/device/src/ProvisioningRegistrationAdditionalData.cs
abhipsaMisra: > This class, as I understand it, is for user defined data only and for users who are using custom webhooks for provisioning.

Is that so? From the name I inferred that this class was a holder for any additional "data" that accompanies the registration request. That could be a payload containing custom webhook data or the pnp model data, and/or in this case, the certificate signing request.

I agree that calling this `RegistrationOptions` would have been a better option that we can definitely include in v2.

abhipsaMisra: Technically, even the certificate request is a part of the "json data", since it is all json at the end. So, I can see how the current structure can get confusing. 
But other than adding another overload and a new type, this seemed like the better of the two options.
What do you think?

timtay-microsoft: Your current solution is fine for v1, yeah. I'd definitely like to see this class renamed in v2 though if we go this direction

timtay-microsoft: > Is that so? From the name I inferred that this class was a holder for any additional "data" that accompanies the registration request. That could be a payload containing custom webhook data or the pnp model data, and/or in this case, the certificate signing request.

The user defined json object here does nothing unless a user is using a custom allocation policy to decide which hub a given device is assigned to. By contrast, your new field will be used regardless of what allocation policy is attached to an enrollment. At the very least, I think it's worthwhile to call this out in the header docs

timtay-microsoft (approved):

azabbasi (approved):

PR #2329, opened 2022-03-31
refactor(e2e-tests): Rename E2E tests for provisioning client

abhipsaMisra:
The only code changes in this PR are renaming the provisioning tests to the format `<protocol>_<enrollment_type>_<feature>_<result>`.
The existing naming was all over the place.

Some irrelevant tests are also removed, eg: testing proxy with both tcp with ws fallback and again with only ws specified. Both tests would go through the same code path.


azabbasi:
> The only code changes in this PR are renaming the provisioning tests to the format `<protocol>_<enrollment_type>_<feature>_<result>`. The existing naming was all over the place.
> 
> Some irrelevant tests are also removed, eg: testing proxy with both tcp with ws fallback and again with only ws specified. Both tests would go through the same code path.

I like bringing consistency to test names, but once you look at the test names in isolation, it's difficult to tell that it's related to DPS client, it may be helpful to prefix the test names with appropriate area path.

azabbasi (approved):

PR #2343, opened 2022-04-11
Add support for .NET 6

azabbasi:


vsts/test-release-nuget.yaml
```diff
@@ -13,6 +13,8 @@ jobs:
       matrix:
         .Net 5.0:
           FRAMEWORK: net5.0
+        .Net 6.0:
```
abhipsaMisra, line 16: do we plan to release a .net 6.0 dll as well?

azabbasi: good question, I don't think so, there is nothing new there, net5.0 dll should suffice. I shouldn't add it to the test-release matrix then right?

timtay-microsoft (approved):

Issue #2323, opened 2022-03-24
[Bug Report] deviceClient.GetFileUploadSasUriAsync throws null ref exception 


Skaanning:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 11 and raspberry pi os
![image](https://user-images.githubusercontent.com/4522165/159920942-416aa75d-dcc2-4ad9-a2f7-ca598feac9e7.png)

- **Application's .NET Target Framework :** dotnet6
- **Device:** Laptop & raspberry pi 4
- **SDK version used:** 
      Microsoft.Azure.Devices 1.37.0
      Microsoft.Azure.Devices.Client 1.40.0
      Microsoft.Azure.Devices.Provisioning.Client 1.19.1
      Microsoft.Azure.Devices.Provisioning.Transport.Mqtt 1.17.1

## Description of the issue
Using the `ProvisioningDeviceClient` and a X509Certificate I register my device. I then use the registration result, to create a new `DeviceClient` including the certificate via the DeviceAuthenticationWithX509Certificate. I then try to get a `FileUploadSasUri` from the device client. This however fails deep inside a `HttpClientHelper`. As far as I can see this happens because a flag _usingX509ClientCertificate is false.

fails in the highlighted area
![image](https://user-images.githubusercontent.com/4522165/159928038-87e84fe4-2dee-4cda-946a-4c10e08efb86.png) 

Exception
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Azure.Devices.Client.Transport.HttpClientHelper.ExecuteAsync(HttpMethod httpMethod, Uri requestUri, Func`3 modifyRequestMessageAsync, Func`2 isSuccessful, Func`3 processResponseMessageAsync, IDictionary`2 errorMappingOverrides, CancellationToken cancellationToken)
```

## Code sample exhibiting the issue

```c#
 public async Task SetupIotDeviceClient(X509Certificate2 certificate)
    {
        var registrationResult = await RegisterDeviceWithProvisioningService(certificate);
        var auth = new DeviceAuthenticationWithX509Certificate(registrationResult.DeviceId, certificate);
        var clientOptions = new ClientOptions{ SdkAssignsMessageId = SdkAssignsMessageId.WhenUnset };
        
        var deviceClient = DeviceClient.Create(registrationResult.AssignedHub, auth, TransportType.Mqtt, clientOptions);
        //this throws the error
[...]

PR #2350, opened 2022-04-15
File upload SAS URI operation requires certificate to be copied

drwill-ms:
Former code only worked if no ClientOptions was specified. It needs to apply the cert no matter which instances exist.

Other code clean up.

#2323

### DIFF
--- a/e2e/test/iothub/FileUploadE2ETests.cs
+++ b/e2e/test/iothub/FileUploadE2ETests.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Net;
 using System.Security.Cryptography.X509Certificates;
@@ -50,6 +49,14 @@ namespace Microsoft.Azure.Devices.E2ETests
         {
             string smallFileBlobName = await GetTestFileNameAsync(FileSizeSmall).ConfigureAwait(false);
             await GetSasUriAsync(Client.TransportType.Http1, smallFileBlobName, true).ConfigureAwait(false);
+        }
+
+        [LoggedTestMethod]
+        [TestCategory("LongRunning")]
+        public async Task FileUpload_GetFileUploadSasUri_Mqtt_x509_NoFileTransportSettingSpecified()
+        {
+            string smallFileBlobName = await GetTestFileNameAsync(FileSizeSmall).ConfigureAwait(false);
+            await GetSasUriAsync(Client.TransportType.Mqtt, smallFileBlobName, true).ConfigureAwait(false);
         }
 
         [LoggedTestMethod]
@@ -200,10 +207,14 @@ namespace Microsoft.Azure.Devices.E2ETests
 
         private async Task GetSasUriAsync(Client.TransportType transport, string blobName, bool useX509auth = false)
         {
-            using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(
-                Logger,
-                _devicePrefix,
-                useX509auth ? TestDeviceType.X509 : TestDeviceType.Sasl).ConfigureAwait(false);
+            using TestDevice testDevice = await TestDevice
+                .GetTestDeviceAsync(
+                    Logger,
+                    _devicePrefix,
+                    useX509auth
+                        ? TestDeviceType.X509
+                        : TestDeviceType.Sasl)
+                .ConfigureAwait(false);
 
             DeviceClient deviceClient;
             X509Certificate2 cert = null;
--- a/iothub/device/src/ClientFactory.cs
+++ b/iothub/device/src/ClientFactory.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Azure.Devices.Client
             var connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
 
             // Make sure client options is initialized with the correct transport setting.
-            EnsureOptionsIsSetup(connectionStringBuilder.Certificate, null, ref options);
+            EnsureOptionsIsSetup(connectionStringBuilder.Certificate, ref options);
 
             if (authenticationMethod is DeviceAuthenticationWithX509Certificate)
             {
@@ -123,12 +123,14 @@ namespace Microsoft.Azure.Devices.Client
                     throw new ArgumentException("No certificate was found. To use certificate authentication certificate must be present.");
                 }
 
-                InternalClient dc = CreateFromConnectionString(
+                InternalClient internalClient = CreateFromConnectionString(
                     connectionStringBuilder.ToString(),
+                    authenticationMethod,
                     PopulateCertificateInTransportSettings(connectionStringBuilder, transportType),
+                    null,
                     options);
 
-                dc.Certificate = connectionStringBuilder.Certificate;
+                internalClient.Certificate = connectionStringBuilder.Certificate;
 
                 // Install all the intermediate certificates in the chain if specified.
                 if (connectionStringBuilder.ChainCertificates != null)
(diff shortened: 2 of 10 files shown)


e2e/test/iothub/FileUploadE2ETests.cs
```diff
@@ -52,6 +51,14 @@ public async Task FileUpload_GetFileUploadSasUri_Http_x509_NoFileTransportSettin
             await GetSasUriAsync(Client.TransportType.Http1, smallFileBlobName, true).ConfigureAwait(false);
         }
 
+        [LoggedTestMethod]
+        [TestCategory("LongRunning")]
+        public async Task FileUpload_GetFileUploadSasUri_Mqtt_x509_NoFileTransportSettingSpecified()
```
drwill-ms, line 56: Add a condition where the client is initialized with a non-HTTP transport.

iothub/device/src/ClientFactory.cs
```diff
@@ -113,7 +114,7 @@ internal static InternalClient Create(string hostname, IAuthenticationMethod aut
             var connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
 
             // Make sure client options is initialized with the correct transport setting.
-            EnsureOptionsIsSetup(connectionStringBuilder.Certificate, ref options);
+            EnsureOptionsIsSetup(connectionStringBuilder.Certificate, null, ref options);
```
drwill-ms, line 117: Adds support for copying the proxy too, but this factory method doesn't have a ClientOptions.

iothub/device/src/ClientFactory.cs
```diff
@@ -430,8 +439,9 @@ internal static ITransportSettings[] GetTransportSettings(TransportType transpor
 
             // Clients that derive their authentication method from AuthenticationWithTokenRefresh will need to specify
             // the token time to live and renewal buffer values through the corresponding AuthenticationWithTokenRefresh
-            // implementation constructors instead.
-            if (!(builder.AuthenticationMethod is AuthenticationWithTokenRefresh))
+            // implementation constructors instead, and these values are irrelevant for cert-based auth.
+            if (!(builder.AuthenticationMethod is AuthenticationWithTokenRefresh)
+                && !(builder.AuthenticationMethod is DeviceAuthenticationWithX509Certificate))
```
drwill-ms, line 444: Don't need to assign SAS token TTL and renewal buffer when using cert auth.

iothub/device/src/ClientFactory.cs
```diff
@@ -499,20 +509,30 @@ internal static ITransportSettings[] GetTransportSettings(TransportType transpor
         /// Ensures that the ClientOptions is configured and initialized.
         /// If a certificate is provided, the fileUploadTransportSettings will use it during initialization.
         /// </summary>
-        private static void EnsureOptionsIsSetup(X509Certificate2 cert, ref ClientOptions options)
+        private static void EnsureOptionsIsSetup(X509Certificate2 cert, IWebProxy proxy, ref ClientOptions options)
         {
-            if (options?.FileUploadTransportSettings == null)
-            {
-                var fileUploadTransportSettings = new Http1TransportSettings { ClientCertificate = cert };
                 if (options == null)
```
drwill-ms, line 514: This is the main fix. If no options, create one. If options has no FileUploadTransportSettings, add one. If cert is provided and ClientCertificate is not null, assign it. Same for proxy.

iothub/device/src/IotHubConnectionString.Core.cs
```diff
@@ -24,9 +24,17 @@ Task<string> IAuthorizationProvider.GetPasswordAsync()
                         || TokenRefresher != null,
                     "The token refresher and the shared access signature can't both be null");
 
-                return !string.IsNullOrWhiteSpace(SharedAccessSignature)
```
drwill-ms, line 27: This logic was faulty, and produced the NRE. By making the method `async`, we return the value, not the Task. This way the caller will get a Task with the result of `null` instead of an NRE when evaluating the Task.

It shouldn't happen, but it did, so better have code that handles it.

iothub/device/src/Transport/HttpClientHelper.cs
```diff
@@ -434,7 +432,10 @@ private static void InsertEtag(HttpRequestMessage requestMessage, IETagHolder en
             if (!_usingX509ClientCert)
             {
                 string authHeader = await _authenticationHeaderProvider.GetPasswordAsync().ConfigureAwait(false);
-                msg.Headers.Add(HttpRequestHeader.Authorization.ToString(), authHeader);
+                if (!string.IsNullOrWhiteSpace(authHeader))
```
drwill-ms, line 435: In the unlikely event that we do get null back, don't try to add it as a header, because it would be invalid.

iothub/device/src/ClientFactory.cs
```diff
@@ -201,7 +204,7 @@ internal static InternalClient Create(string hostname, IAuthenticationMethod aut
             var connectionStringBuilder = IotHubConnectionStringBuilder.Create(hostname, gatewayHostname, authenticationMethod);
 
             // Make sure client options is initialized with the correct transport setting.
-            EnsureOptionsIsSetup(connectionStringBuilder.Certificate, ref options);
+            EnsureOptionsIsSetup(connectionStringBuilder.Certificate, transportSettings.FirstOrDefault()?.Proxy, ref options);
```
abhipsaMisra, line 207: The `FirstOrDefault()` approach might not work for proxy as well as it does for the client certificates; eg. if my transport options are [mqtt_tcp, mqtt_ws], then firstOrDefault would look at the tcp settings for the proxy whereas the actual proxy setting would only be under the ws settings.
This could be a bit complicated to set up since I believe we set up the file upload transport handler before we've finalized on a transport protocol to use.
We set up file upload transport handler as a part of the client set up whereas the transport protocol will be chosen only once an attempt is made to open up the connection.

iothub/device/src/InternalClient.cs
```diff
@@ -1841,12 +1752,9 @@ internal async Task OnModuleEventMessageReceivedAsync(string input, Message mess
                     _moduleReceiveMessageSemaphore.Release();
                 }
 
-                // codes_SRS_DEVICECLIENT_33_002: [ The OnReceiveEventMessageCalled shall invoke the specified delegate. ]
```
abhipsaMisra, line 1844: Thanks for removing these comments. It's been far too long since they were maintained, and they are unnecessarily crowding up our files.

iothub/device/src/ClientFactory.cs
drwill-ms: How about this?

```csharp
transportSettings.Where(x => x.Proxy != null).FirstOrDefault()?.Proxy
```

drwill-ms: Should I also add an exclusion for `DefaultWebProxySettings.Instance`?

iothub/device/src/ClientOptions.cs
```diff
@@ -19,9 +19,9 @@ public class ClientOptions
 
         /// <summary>
         /// The transport settings to use for all file upload operations, regardless of what protocol the device
-        /// client is configured with. All file upload operations take place over https. 
-        /// If FileUploadTransportSettings is not provided, then file upload operations will use the client certificates configured
-        /// in the transport settings set for the non-file upload operations.
+        /// client is configured with. All file upload operations take place over https.
+        /// If FileUploadTransportSettings is not provided, then file upload operations will use the same client certificates
+        /// configured in the transport settings set for client connect.
         /// </summary>
```
abhipsaMisra, line 25: should we add a comment to call out how users should set proxy?

abhipsaMisra (approved):

iothub/device/src/ClientOptions.cs
drwill-ms: I mean, all we call out is client cert here, so I don't think we need to specify all the other things you'd need to set. Do you?

PR #2352, opened 2022-04-19
Allow users to configure websocket keep-alive

drwill-ms:


iothub/device/src/ClientOptions.cs
```diff
@@ -54,5 +54,11 @@ public class ClientOptions
         /// or the <see cref="ModuleClient.CreateFromEnvironmentAsync(ClientOptions)"/> flow.
         /// </remarks>
         public int SasTokenRenewalBuffer { get; set; }
+
+        /// <summary>
+        /// A keep-alive for when using web sockets.
+        /// </summary>
+        /// <seealso href="https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocketoptions.keepaliveinterval"/>
+        public TimeSpan WebSocketKeepAlive { get; set; }
```
drwill-ms, line 62: James suggested adding it to `ClientOptions` as it applies to MQTT and AMQP.

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -27,11 +27,9 @@ internal class AmqpConnectionHolder : IAmqpConnectionHolder, IAmqpUnitManager
         public AmqpConnectionHolder(IDeviceIdentity deviceIdentity)
         {
             _deviceIdentity = deviceIdentity;
-            _amqpIotConnector = new AmqpIotConnector(deviceIdentity.AmqpTransportSettings, deviceIdentity.IotHubConnectionString.HostName);
+            _amqpIotConnector = new AmqpIotConnector(deviceIdentity.AmqpTransportSettings, deviceIdentity.Options, deviceIdentity.IotHubConnectionString.HostName);
```
drwill-ms, line 30: But now we have to pipe client options through several classes.

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -177,13 +185,20 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
...
                 }
                 catch (PlatformNotSupportedException)
                 {
                     // .NET Core 2.0 doesn't support proxy. Ignore this setting.
-                    Logging.Error(this, $"{nameof(CreateClientWebSocketAsync)} PlatformNotSupportedException thrown as .NET Core 2.0 doesn't support proxy");
+                    if (Logging.IsEnabled)
+                        Logging.Error(this, $"{nameof(CreateClientWebSocketAsync)} PlatformNotSupportedException thrown as .NET Core 2.0 doesn't support proxy");
+                }
+
+                if (_clientOptions?.WebSocketKeepAlive != null)
```
drwill-ms, line 199: AMQP: assign it, if not null.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1292,13 +1296,18 @@ private async Task<Message> SendTwinRequestAsync(Message request, string rid, Ca
                         Logging.Error(this, $"{nameof(CreateWebSocketChannelFactory)} PlatformNotSupportedException thrown as .NET Core 2.0 doesn't support proxy");
                 }
 
+                if (options?.WebSocketKeepAlive != null)
```
drwill-ms, line 1299: MQTT: assign it, if not null.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1292,13 +1296,18 @@ private async Task<Message> SendTwinRequestAsync(Message request, string rid, Ca
...
+                }
+
                 if (settings.ClientCertificate != null)
                 {
                     websocket.Options.ClientCertificates.Add(settings.ClientCertificate);
                 }
 
                 // Support for RemoteCertificateValidationCallback for ClientWebSocket is introduced in .NET Standard 2.1
-#if NETSTANDARD2_1
+#if NETSTANDARD2_1_OR_GREATER
```
drwill-ms, line 1310: We failed to future proof this, which perhaps wasn't available at the time.

iothub/device/src/ClientOptions.cs
abhipsaMisra: I like consolidating similar settings in a single holder like ClientOptions, but historically the .NET SDK has had transport level settings configured in the corresponding transport settings file. Maybe we could add it to our ITransportSettings interface to avoid duplicating code and still preserve the existing flow of setting all transport level settings in a single place?

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
abhipsaMisra: q - just a thought - do you think it'll be easier if we pipe the entire device identity instance to our amqp classes?

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
abhipsaMisra: q - does `TimeSpan` default to null or do we need to define it as `TimeSpan?` explicitly?

iothub/device/src/ClientOptions.cs
abhipsaMisra: We should also call out explicitly that this keep-alive is different from mqtt keep-alive or amqp idle timeout.

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -182,7 +182,7 @@ public bool CertificateRevocationCheck
         public bool CleanSession { get; set; }
 
         /// <summary>
-        /// The interval, in seconds, that the client establishes with the service, for sending keep-alive pings.
+        /// The interval, in seconds, that the client establishes with the service, for sending keep-alive pings over TCP.
         /// The default is 300 seconds.
```
abhipsaMisra, line 186: I know this was already there, but we should probably add an explanation here that even though this property value defaults to 300 seconds, that means that the default keep-alive is 300/4 = 75 seconds.

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
drwill-ms: I wondered about that, but chose not to change it that drastically.

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
drwill-ms: Good point. It must be declared to be nullable.

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
drwill-ms: This is so messed up, I'm not convinced we can make it much better than it is with documentation. I'd rather defer to v2.

iothub/device/src/ClientOptions.cs
drwill-ms: What would that mean for HTTP1 transport, I wonder?

abhipsaMisra: That's true, yes. Didn't we recently add a keep-alive for http clients so that we could trigger faster DNS refreshes? Do you remember if we made it configurable and if yes, where that property is set by users?

abhipsaMisra: I would still vote for us to duplicate the property and mqtt and amqp transport settings rather than putting it in client options. Given the way the sdk is set up today I don't think it would be intuitive for users to look at client options for configuring transport level settings.

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
timtay-microsoft: As a rule of thumb, I'd avoid passing references of "higher level" objects down to "lower level" objects. I think just passing the clientOptions bundle here should be fine

iothub/device/src/ClientOptions.cs
```diff
@@ -54,5 +54,11 @@ public class ClientOptions
         /// or the <see cref="ModuleClient.CreateFromEnvironmentAsync(ClientOptions)"/> flow.
         /// </remarks>
         public int SasTokenRenewalBuffer { get; set; }
+
+        /// <summary>
+        /// A keep-alive for when using web sockets.
+        /// </summary>
+        /// <seealso href="https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocketoptions.keepaliveinterval"/>
+        public TimeSpan? WebSocketKeepAlive { get; set; }
```
timtay-microsoft, line 62: I like this as a home for this new option. I don't know if this has been discussed yet, but I'd like to see us remove the transport specific settings classes when we move to v2. With that in mind, I like adding to ClientOptions like this.

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
abhipsaMisra: I'd like to understand why we should avoid passing "higher level" objects down to "lower level" objects? Are you worried about lower level object inadvertently changing the state of the "higher level" instance and other objects getting a stale reference?

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -177,13 +182,20 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
...
                 }
                 catch (PlatformNotSupportedException)
                 {
                     // .NET Core 2.0 doesn't support proxy. Ignore this setting.
-                    Logging.Error(this, $"{nameof(CreateClientWebSocketAsync)} PlatformNotSupportedException thrown as .NET Core 2.0 doesn't support proxy");
+                    if (Logging.IsEnabled)
+                        Logging.Error(this, $"{nameof(CreateClientWebSocketAsync)} PlatformNotSupportedException thrown as .NET Core 2.0 doesn't support proxy");
+                }
+
+                if (_amqpTransportSettings.WebSocketKeepAlive.HasValue)
```
abhipsaMisra, line 196: Interesting!

iothub/device/src/ClientOptions.cs
jamdavi: I too like it here since it impacts both transports.

iothub/device/src/ClientOptions.cs
jamdavi: While I do like them to be consolidated, I have to agree that if we've done it a specific way before we shouldn't break that convention.

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -146,6 +146,12 @@ public TimeSpan OpenTimeout
             set => SetOpenTimeout(value);
         }
 
+        /// <summary>
+        /// A keep-alive for when using web sockets.
```
abhipsaMisra, line 150: I think it is worth calling out here that this is different from mqtt spec keep-alive. Else we might have people setting this value when using mqtt_ws and then wondering why they aren't receiving a ping response from the service.

abhipsaMisra (commented):
looks good!

iothub/device/src/ClientOptions.cs
drwill-ms: I'm probably not a fan of consolidating everything, or at least having transport-specific sections in `ClientOptions`. We have enough of a problem today with this strategy pattern where not everything applies to every transport. I think we're better off keeping things separate to avoid confusion.

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
drwill-ms: how so?

iothub/device/src/AmqpTransportSettings.cs
drwill-ms: Please suggest a wording change.

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
abhipsaMisra: I had not tried the Nullable.HasValue before. Seems like a useful property to have so that we longer have to do null and default value checks.

iothub/device/src/AmqpTransportSettings.cs
abhipsaMisra: A keep-alive for sending ping/pong control frames when using web sockets. This value is different from the protocol level keep-alive packets that are sent over the overlaying MQTT transport protocol.

timtay-microsoft (approved):

brycewang-microsoft (approved):

abhipsaMisra (approved):

iothub/device/src/AmqpTransportSettings.cs
```diff
@@ -146,6 +146,12 @@ public TimeSpan OpenTimeout
             set => SetOpenTimeout(value);
         }
 
+        /// <summary>
+        /// A keep-alive for the transport layer in sending ping/pong control frames when using web sockets.
+        /// </summary>
+        /// <seealso href="https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocketoptions.keepaliveinterval"/>
```
jamdavi, line 152: For consistency with the other protocol docs. 

```suggestion
        /// <remarks>
        /// This value is different from the protocol-level keep-alive packets that are sent over the overlaying AMQP transport protocol.
        /// </remarks>
        /// <seealso href="https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocketoptions.keepaliveinterval"/>
```

jamdavi (approved):

Issue #2344, opened 2022-04-12
Why the default timeout for the Microsoft.Azure.Devices.ServiceClient cannot be set below 100 seconds?


ZolotaikoRoman:
I faced a problem that doesn't allow setting the default operation timeout below 100 seconds. There is a logic that says that the custom timeout for the device cloud method can be only if it is more than 100 sec. The case is that I want to invoke a direct method on the device and I want the timeout to be 10-20 seconds, after it, I want to handle the timeout exception using my custom code. Currently, it is impossible due to the internal rules of the ServiceClient, and there is no possibility to override these rules.
It would be good if you explain why it is developed this way and suggest/advise me on what can I do.
Thanks.

![image](https://user-images.githubusercontent.com/37477984/162925736-e2c0e76a-7730-4d25-85a7-c8b537951e20.png)

drwill-ms:
I don't see a good reason for it, and according to our [service documentation](https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-direct-methods#method-invocation) the limits are between 5 and 300 seconds.

The change was made back in 2018 and there aren't good comments in the commit to specify why, but it was changed from 60 seconds to 100 minimum.

@andyk-ms We should hand this off to the service team (who now owns this SDK) to let them decide how they want to change it. We've generally moved to the stance that the SDK should not to error checking like this, rather rely on the service to return an error.

ZolotaikoRoman:
@drwill-ms Thanks for the quick response!
Please let me know when you have any updates from the service team.

drwill-ms:
The fix for this has shipped with version 1.37.1.

<https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2022-5-6>

PR #2355, opened 2022-04-21
Remove default timeout

vighatke:
Bug: https://github.com/Azure/azure-iot-sdk-csharp/issues/2344

### DIFF
--- a/iothub/service/src/ServiceClient.cs
+++ b/iothub/service/src/ServiceClient.cs
@@ -636,7 +636,7 @@ namespace Microsoft.Azure.Devices
             var timeout = TimeSpan.FromSeconds(15); // For wire time
             timeout += TimeSpan.FromSeconds(cloudToDeviceMethod.ConnectionTimeoutInSeconds ?? 0);
             timeout += TimeSpan.FromSeconds(cloudToDeviceMethod.ResponseTimeoutInSeconds ?? 0);
-            return timeout < TimeSpan.FromSeconds(5) ? s_defaultOperationTimeout : timeout;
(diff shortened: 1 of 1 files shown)


firstaye-zz (approved):
Looks good to me, but in what scenario the timeout would be negative, just curious,

drwill-ms (approved):

iothub/service/src/ServiceClient.cs
```diff
@@ -634,9 +635,24 @@ private static TimeSpan GetInvokeDeviceMethodOperationTimeout(CloudToDeviceMetho
...
+            timeout += responseTimeOut;
+
+            return timeout;
+        }
+
+        private static void ValidateTimeOut(TimeSpan connectionTimeOut)
+        {
+            if (connectionTimeOut.Seconds < 0)
+            {
+                throw new ArgumentException("Negative timeout");
```
drwill-ms, line 654: This technically introduces a breaking change.

drwill-ms:
> Looks good to me, but in what scenario the timeout would be negative, just curious

@firstaye-zz an end user could specify any `int` value they want, including negative values, so we mustn't allow them.

iothub/service/src/ServiceClient.cs
vighatke: true. Should I remove this and let the service deal with negative values?

drwill-ms: The user can already get an exception for invalid values (when too large, right)? So that seems like it would be fine, that is not a breaking change.

PR #2358, opened 2022-04-27
Updates from usability study

drwill-ms:
- Simplify ClientPropertiesCollection interface by removing plural methods and `Add` methods that throw if duplicate.
- Improve doc comments
- Make `CommandResponse` easier to initialize/use.

### DIFF
--- a/e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
+++ b/e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
@@ -472,7 +472,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Properties
 
             // For the property patch sent here will result in the component being removed.
             var propertyPatch4 = new ClientPropertyCollection();
-            propertyPatch4.AddComponentProperties(ComponentName, null);
+            propertyPatch4.AddRootProperty(ComponentName, null);
             await deviceClient.UpdateClientPropertiesAsync(propertyPatch4).ConfigureAwait(false);
 
             serviceTwin = await registryManager.GetTwinAsync(testDevice.Id).ConfigureAwait(false);
--- a/iothub/device/devdoc/Convention-based operations.md
+++ b/iothub/device/devdoc/Convention-based operations.md
@@ -1,6 +1,6 @@
-## Plug and Play convention compatible APIs
+# Plug and Play convention compatible APIs
 
-#### Common
+## Common
 
 ```diff
 public class ClientOptions {
@@ -18,7 +18,6 @@ public class ModuleClient : IDisposable {
 ```
 
 ```csharp
-
 public abstract class PayloadConvention {
     protected PayloadConvention();
     public abstract PayloadEncoder PayloadEncoder { get; }
@@ -72,7 +71,6 @@ public abstract class PayloadCollection : IEnumerable, IEnumerable<KeyValuePair<
     public PayloadConvention Convention { get; internal set; }
     public virtual object this[string key] { get; set; }
     public virtual void Add(string key, object value);
-    public virtual void AddOrUpdate(string key, object value);
     public void ClearCollection();
     public bool Contains(string key);
     public IEnumerator<KeyValuePair<string, object>> GetEnumerator();
@@ -101,7 +99,7 @@ public class CommonClientResponseCodes {
 }
 ```
 
-### Properties
+## Properties
 
 ```csharp
 /// <summary>
@@ -127,7 +125,7 @@ public Task<ClientPropertiesUpdateResponse> UpdateClientPropertiesAsync(ClientPr
 public Task SubscribeToWritablePropertyUpdateRequestsAsync(Func<ClientPropertyCollection, Task> callback, CancellationToken cancellationToken = default);
 ```
 
-#### All related types
+### All related types
 
 ```csharp
 public class ClientProperties {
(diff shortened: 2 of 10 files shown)


e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
```diff
@@ -472,7 +472,7 @@ private async Task Properties_DeviceSendsNullValueForPropertyResultsServiceRemov
 
             // For the property patch sent here will result in the component being removed.
             var propertyPatch4 = new ClientPropertyCollection();
-            propertyPatch4.AddComponentProperties(ComponentName, null);
+            propertyPatch4.AddRootProperty(ComponentName, null);
```
drwill-ms, line 475: This is an oddity. Since we don't have the plural overload anymore, we can't null out the entire component collection. Instead we need to treat it like a root property.

@abhipsaMisra, thoughts?

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -17,162 +16,35 @@ public class ClientPropertyCollection : PayloadCollection
...
-        /// When using this as part of the writable property flow to respond to a writable property update you should pass in the value
-        /// as an instance of <see cref="PayloadSerializer.CreateWritablePropertyResponse(object, int, long, string)"/>
-        /// to ensure the correct formatting is applied when the object is serialized.
-        /// </para>
-        /// </remarks>
-        /// <param name="propertyName">The name of the property to add.</param>
-        /// <param name="propertyValue">The value of the property to add.</param>
-        /// <exception cref="ArgumentNullException"><paramref name="propertyName"/> is <c>null</c>.</exception>
-        /// <exception cref="ArgumentException"><paramref name="propertyName"/> already exists in the collection.</exception>
-        public void AddRootProperty(string propertyName, object propertyValue)
```
drwill-ms, line 35: Remove this overload that throws if the value already exists. Note the `AddOrUpdate` overload is renamed to just `AddRootProperty` and effectively replaces this.

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -17,162 +16,35 @@ public class ClientPropertyCollection : PayloadCollection
...
-        /// <inheritdoc path="/remarks" cref="AddRootProperty(string, object)" />
-        /// <inheritdoc path="/seealso" cref="AddInternal(IDictionary{string, object}, string, bool)" />
-        /// <summary>
-        /// Adds the value to the collection.
-        /// </summary>
-        /// <param name="componentName">The component with the properties to add.</param>
-        /// <param name="properties">A collection of properties to add.</param>
-        /// <exception cref="ArgumentException">A property name in <paramref name="properties"/> already exists in the collection.</exception>
-        /// <exception cref="ArgumentNullException"><paramref name="componentName"/> or a property name in <paramref name="properties"/> is <c>null</c>.</exception>
-        public void AddComponentProperties(string componentName, IDictionary<string, object> properties)
```
drwill-ms, line 67: Remove the plural versions of these APIs.

iothub/device/src/ClientPropertyCollection.cs
```diff
@@ -204,7 +76,11 @@ public bool Contains(string componentName, string propertyName)
         /// Gets the value of a component-level property.
         /// </summary>
         /// <remarks>
+        /// If this instance is in an update request for a writable property, <paramref name="propertyValue"/> should be of type
```
drwill-ms, line 79: Help someone discover the type of the out value they should use.

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
abhipsaMisra: Implementation wise, this should be fine. Any property with a null value will end up getting removed. This does feel a bit weird naming wise, though.

iothub/device/tests/Convention/ClientPropertyCollectionTests.cs
```diff
@@ -442,19 +446,6 @@ public void ClientPropertyCollection_AddNullPropertyNameThrows()
             testAction.Should().Throw<ArgumentNullException>();
         }
 
-        [TestMethod]
-        public void ClientPropertyCollection_AddOrUpdateNullPropertyNameThrows()
```
abhipsaMisra, line 446: why did we remove this?

abhipsaMisra: Ah, I see that there is an "add" variation tested already. Nevermind.

abhipsaMisra (approved):

e2e/test/iothub/properties/PropertiesWithComponentsE2ETests.cs
drwill-ms: I feel kinda weird about `AddRootProperty` instead of `SetRootProperty`. Thoughts on the verb?

iothub/device/devdoc/Convention-based operations.md
```diff
@@ -1,8 +1,8 @@
-## Plug and Play convention compatible APIs
+# Plug and Play convention compatible APIs
 
-#### Common
+## Common
 
-```diff
+```csharp
```
abhipsaMisra, line 5: we're going to want to keep this block as `diff` so that it highlights the 2 APIs added to ClientOptions

drwill-ms: what does it do for markdown?

drwill-ms: Oh, fascinating. I didn't know that. Thanks!

PR #2363, opened 2022-04-28
Move convention classes from shared to device client

drwill-ms:


iothub/device/src/Convention/WritableClientProperty.cs
```diff
@@ -32,13 +30,15 @@ internal WritableClientProperty()
         internal PayloadConvention Convention { get; set; }
 
         /// <summary>
-        /// Creates a writable property update response that can be reported back to the service.
+        /// Creates a writable property update response that contains the requested property value and version that can be reported back to the service.
```
drwill-ms, line 33: Also tweaked this doc comment to make this fact more prominent.

abhipsaMisra (approved):

brycewang-microsoft (approved):

PR #2372, opened 2022-05-04
Add .net6.0 testing target

drwill-ms:


vsts/vsts.yaml
```diff
@@ -426,8 +438,16 @@ jobs:
           performMultiLevelLookup: true
           installationPath: $(Agent.ToolsDirectory)/dotnet
 
+      - task: UseDotNet@2
+        displayName: 'Use .NET SDK 6.0'
+        inputs:
+          packageType: sdk
+          version: 6.x
```
abhipsaMisra, line 445: Any reason why the pri 1 windows job doesn't require us to manually install the .net 6 SDK while this one does?

abhipsaMisra (approved):

vsts/vsts.yaml
drwill-ms: I think the pri1 job not stating .net version requirements was a bug I introduced, covered up by the fact that the image we use has all the .net framework verisons we need, except 6.0.

PR #2373, opened 2022-05-04
IoT hub device client DotNetty -> MQTTNet

timtay-microsoft:
There are still 2 TODOs here, but they can only be addressed after this code is checked in.

- Make Patch twin operations return the new twin version
- Remove ReceiveMessageAsync call

Both require changes outside of the MQTT stack that I feel would fit better in a separate PR.

The bulk of the code changes here are in MqttTransportHandler so it may be easier to review this file in [source](https://github.com/Azure/azure-iot-sdk-csharp/blob/timtay/mqttnet/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs) rather than in diff.


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -126,4 +120,28 @@
     <PackageReference Include="Microsoft.AspNet.WebApi.Client" Version="5.2.3" />
     <PackageReference Include="Microsoft.Owin" Version="4.0.0" />
   </ItemGroup>
+
```
drwill-ms, line 123: Pretty sure the plan is to only target netstandard2.0 in v2. If it makes your life easier, I'd suggest removing all other targets.

iothub/device/src/Transport/Mqtt/MqttTransportSettings.cs
```diff
@@ -107,18 +101,6 @@ public bool CertificateRevocationCheck
         /// </remarks>
         public TimeSpan DeviceReceiveAckTimeout { get; set; }
 
-        /// <summary>
-        /// The QoS to be used when sending packets to service.
-        /// The default value is <see cref="QualityOfService.AtLeastOnce"/>.
-        /// </summary>
-        public QualityOfService PublishToServerQoS { get; set; }
```
drwill-ms, line 114: Does MQTTnet not offer these settings?

timtay-microsoft: It does, but I haven't worked it in yet

iothub/device/src/Transport/Mqtt/IWillMessage.cs
```diff
@@ -1,23 +1,21 @@
...
-    /// An MQTT "will" message, used when a client disconnects
+    /// An MQTT "will" message to be sent by this client when the client disconnects.
     /// </summary>
     public interface IWillMessage
     {
         /// <summary>
-        /// The message to be sent
+        /// The payload to be sent in the will message
         /// </summary>
-        Message Message { get; }
```
abhipsaMisra, line 16: I'm happy that we're finally moving away from using IoT `Message` to represent an MQTT message.

iothub/device/src/Transport/Mqtt/MqttLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Diagnostics.Tracing;
+using MQTTnet.Diagnostics;
+
+namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
+{
+    internal class MqttLogger : IMqttNetLogger
+    {
+        public bool IsEnabled => Logging.IsEnabled;
+
+        public void Publish(MqttNetLogLevel logLevel, string source, string message, object[] parameters, Exception exception)
```
abhipsaMisra, line 14: When you have some time, I'm interested to understand how you came to this implementation.

iothub/device/src/Transport/Mqtt/QualityOfService.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        AtMostOnce = 0,
+
+        /// <summary>
+        /// The message will be sent once, but will be resent if the service fails to acknowledge the message.
+        /// </summary>
+        AtLeastOnce = 1,
+
+        /// <summary>
+        /// TODO does hub even support this?
```
abhipsaMisra, line 25: It doesn't! 😄 
Publishing with QoS 2 will result in Hub closing the connection.
Subscribing with QoS 2 will result in Hub granting QoS 1.

iothub/device/src/ClientSettings/MqttTransportSettings.cs
```diff
@@ -17,7 +17,7 @@ public class MqttTransportSettings : ITransportSettings
 
         private const bool DefaultCleanSession = false;
         private const bool DefaultHasWill = false;
-        private const int DefaultKeepAliveInSeconds = 300;
+        private readonly TimeSpan DefaultKeepAlive = TimeSpan.FromSeconds(300);
```
drwill-ms, line 20: Convert to minutes for easier comprehension, given the scale of the value for seconds.
```suggestion
        private readonly TimeSpan DefaultKeepAlive = TimeSpan.FromMinutes(5);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -94,12 +95,12 @@ internal sealed class MqttTransportHandler : TransportHandler
         private readonly Func<string, Message, Task> _moduleMessageReceivedListener;
         private readonly Func<Message, Task> _deviceMessageReceivedListener;
 
-        private readonly Dictionary<string, MqttApplicationMessageReceivedEventArgs> messagesToAcknowledge = new Dictionary<string, MqttApplicationMessageReceivedEventArgs>();
+        private readonly ConcurrentDictionary<string, MqttApplicationMessageReceivedEventArgs> messagesToAcknowledge = new ConcurrentDictionary<string, MqttApplicationMessageReceivedEventArgs>();
```
drwill-ms, line 98: Shorter lines.

```suggestion
        private readonly ConcurrentDictionary<string, MqttApplicationMessageReceivedEventArgs> messagesToAcknowledge = new();
```

iothub/device/src/IotHubDeviceClient.cs
```diff
@@ -161,12 +162,6 @@ public async Task<Message> ReceiveMessageAsync(CancellationToken cancellationTok
             {
                 _deviceReceiveMessageSemaphore.Release();
 
-                if (_deviceReceiveMessageCallback != null)
-                {
-                    // Any previously received C2D messages will also need to be delivered.
-                    await InnerHandler.EnsurePendingMessagesAreDeliveredAsync(cancellationToken).ConfigureAwait(false);
```
timtay-microsoft, line 167: This method does nothing now that we aren't using DotNetty

iothub/device/src/Transport/Mqtt/GetTwinResponse.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>
+        /// Constructor with the retrieved twin. Used when the get twin request succeeds and the twin was retrieved.
+        /// </summary>
+        /// <param name="status">The status code the service responded with. Should be 200 in this case.</param>
+        /// <param name="twin">The twin returned by the service. Should not be null in this case.</param>
+        internal GetTwinResponse(int status, Twin twin)
+        {
+            Status = status;
+            Twin = twin;
+        }
```
drwill-ms, line 33: Why have any ctors if we can just set the properties?

iothub/device/src/Transport/Mqtt/GetTwinResponse.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// The status code that the service responded to the get twin request with. 200 indicates a successful request.
```
drwill-ms, line 36: Second sentence to remarks? And below.

iothub/device/src/Transport/Mqtt/IWillMessage.cs
```diff
@@ -1,23 +1,21 @@
...
 {
     /// <summary>
-    /// An MQTT "will" message, used when a client disconnects
+    /// An MQTT "will" message to be sent by this client when the client disconnects.
     /// </summary>
     public interface IWillMessage
     {
         /// <summary>
-        /// The message to be sent
+        /// The payload to be sent in the will message
```
drwill-ms, line 12:
```suggestion
        /// The payload to be sent in the will message.
```

iothub/device/src/Transport/Mqtt/MqttLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            var options = new EventSourceOptions()
```
drwill-ms, line 16:
```suggestion
            var options = new EventSourceOptions
```

iothub/device/src/Transport/Mqtt/MqttLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                formattedMessage = string.Format(message, parameters);
+            }
+
+            Logging.Log.Write(formattedMessage, options);
+        }
+
+        // MQTTNet has an enum to represent log level and our logging library has a separate enum. They have the
+        // same values, though
+        internal static EventLevel toEventLevel(MqttNetLogLevel mqttNetLevel) => mqttNetLevel switch
```
drwill-ms, line 32: Not Java
```suggestion
        internal static EventLevel ToEventLevel(MqttNetLogLevel mqttNetLevel) => mqttNetLevel switch
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
+        public override async Task SendEventAsync(Message message, CancellationToken cancellationToken)
         {
-            if (Logging.IsEnabled)
-                Logging.Enter(this, $"Completing a message with lockToken: {lockToken}", nameof(CompleteMessageAsync));
-
             cancellationToken.ThrowIfCancellationRequested();
-            EnsureValidState();
 
-            if (_qosReceivePacketFromService == QualityOfService.AtMostOnce)
+            string TopicName;
```
drwill-ms, line 387: string topicName = PopulateMessagePropertiesFromMessage(
    _moduleId == null
        ? _deviceToCloudMessagesTopic
        : _moduleToCloudMessagesTopic,
    message);

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
-                        $"Expected lock token to end with: '{actualLockToken}'; actual lock token: '{lockToken}'.",
-                        isTransient: false);
-                }
+            MqttClientPublishResult result = await _mqttClient.PublishAsync(mqttMessage, cancellationToken);
 
-                _completionQueue.Dequeue();
-                completeOperationCompletion = _channel.WriteAndFlushAsync(actualLockToken);
+            if (result.ReasonCode != MqttClientPublishReasonCode.Success)
+            {
+                throw new IotHubClientException($"Failed to publish the mqtt packet for message with correlation id {message.CorrelationId} with reason code {result.ReasonCode}", true);
```
drwill-ms, line 414:
```suggestion
                throw new IotHubClientException($"Failed to publish the MQTT packet for message with correlation Id {message.CorrelationId} with reason code {result.ReasonCode}", true);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: That's more concise, but I don't love sticking ternary operators within parameter lists

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
-        public override Task AbandonMessageAsync(string lockToken, CancellationToken cancellationToken)
+        public override async Task SendEventAsync(IEnumerable<Message> messages, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
 
-            throw new NotSupportedException("MQTT protocol does not support this operation");
+            // Note that this sends all messages at once and then waits for all the acknowledgements. This
+            // is the recommended pattern for sending large numbers of messages over an asynchronous
+            // protocol like MQTT
+            List<Task> sendTasks = new List<Task>();
```
drwill-ms, line 425:
```suggestion
            var sendTasks = new List<Task>();
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: I can move the ternary operator above the method call though

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
+            foreach (Message message in messages)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                sendTasks.Add(SendEventAsync(message, cancellationToken));
+            }
+
+            await Task.WhenAll(sendTasks);
```
drwill-ms, line 432:
```suggestion
            await Task.WhenAll(sendTasks).ConfigureAwait(false);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
             cancellationToken.ThrowIfCancellationRequested();
 
-            throw new NotSupportedException("MQTT protocol does not support this operation");
+            // Note that this sends all messages at once and then waits for all the acknowledgements. This
+            // is the recommended pattern for sending large numbers of messages over an asynchronous
+            // protocol like MQTT
+            List<Task> sendTasks = new List<Task>();
+            foreach (Message message in messages)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
```
drwill-ms, line 428: Why check on each message? Feels like that will just slow this operation down.

I'd check before, and then let Task.WhenAll below up (note it may throw an AggregateException or something and that will need to be managed.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
-                        _deviceReceiveMessageSemaphore?.Dispose();
-                        _deviceReceiveMessageSemaphore = null;
-
-                        if (_channel is IDisposable disposableChannel)
-                        {
-                            disposableChannel.Dispose();
-                            _channel = null;
-                        }
-                    }
+            await SubscribeAsync(DirectMethodsRequestTopic, cancellationToken);
```
drwill-ms, line 444: ConfigureAwait

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: good idea

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
+        public override async Task SendMethodResponseAsync(DirectMethodResponse methodResponse, CancellationToken cancellationToken)
         {
-            try
-            {
-                if (Logging.IsEnabled)
-                    Logging.Enter(this, "", $"{nameof(MqttTransportHandler)}.{nameof(CloseAsync)}");
+            var topic = DirectMethodsResponseTopicFormat.FormatInvariant(methodResponse.Status, methodResponse.RequestId);
 
-                cancellationToken.ThrowIfCancellationRequested();
+            MqttApplicationMessage mqttMessage = new MqttApplicationMessageBuilder()
```
drwill-ms, line 456: Can we cache the MqttApplicationMessageBuilder so we don't call new on it so much?

drwill-ms: Might be able to include some of the calls below to that cached instance, like topic and QoS.

drwill-ms: Oh, maybe not because then each additional call to it would modify the builder. Hmm.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
+                .WithTopic(topic)
+                .WithPayload(methodResponse.Payload)
+                .WithQualityOfServiceLevel(publishingQualityOfService)
+                .Build();
 
-                if (TryStop())
-                {
-                    OnTransportClosedGracefully();
+            MqttClientPublishResult result = await _mqttClient.PublishAsync(mqttMessage, cancellationToken);
```
drwill-ms, line 462: Missing ConfigureAwaits in a lot of calls.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
-                {
-                    _fatalException.Throw();
-                }
-            }
-            finally
+            if (result.ReasonCode != MqttClientPublishReasonCode.Success)
             {
-                if (Logging.IsEnabled)
-                    Logging.Exit(this, "", $"{nameof(MqttTransportHandler)}.{nameof(CloseAsync)}");
+                throw new IotHubClientException($"Failed to publish the mqtt packet for direct method response with reason code {result.ReasonCode}", true);
```
drwill-ms, line 466:
```suggestion
                throw new IotHubClientException($"Failed to publish the MQTT packet for direct method response with reason code {result.ReasonCode}", true);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
-            finally
+
+            Logging.Info(this, $"Sent twin get request with request id {requestId}. Now waiting for the service response.");
+            while (!getTwinResponses.ContainsKey(requestId))
             {
-                if (Logging.IsEnabled)
-                    Logging.Exit(this, exception, nameof(OnError));
+                // May need to wait multiple times. This semaphore is released each time a get twin
+                // request gets a response, but it may not be in response to this particular get twin request.
+                _getTwinSemaphore.Wait(cancellationToken);
```
drwill-ms, line 558: This makes me slightly nervous. Are there multiple listeners? Are other twin responses getting lost?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
-        private int _state = (int)TransportState.NotInitialized;
-        private Action<Message> _twinResponseEvent;
-        private AdditionalClientInformation _additionalClientInformation;
+        private readonly ConcurrentDictionary<string, MqttApplicationMessageReceivedEventArgs> messagesToAcknowledge = new();
 
-        internal MqttTransportHandler(
-            PipelineContext context,
-            IotHubClientMqttSettings settings)
-            : this(context, settings, null)
+        private readonly ConcurrentDictionary<string, GetTwinResponse> getTwinResponses = new();
```
drwill-ms, line 100:
```suggestion
        private readonly ConcurrentDictionary<string, GetTwinResponse> _getTwinResponses = new();
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: There could be multiple listeners, yes. A user could call getTwin from multiple threads on the same device client instance

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
 
-        private bool _isDeviceReceiveMessageCallbackSet;
-        private Func<Task> _cleanupFunc;
-        private IChannel _channel;
-        private ExceptionDispatchInfo _fatalException;
-        private IPAddress[] _serverAddresses;
-        private int _state = (int)TransportState.NotInitialized;
-        private Action<Message> _twinResponseEvent;
-        private AdditionalClientInformation _additionalClientInformation;
+        private readonly ConcurrentDictionary<string, MqttApplicationMessageReceivedEventArgs> messagesToAcknowledge = new();
```
drwill-ms, line 98:
```suggestion
        private readonly ConcurrentDictionary<string, MqttApplicationMessageReceivedEventArgs> _messagesToAcknowledge = new();
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: I get that involving semaphores gets scary quickly, but I can't think of another way to make this scenario work. We send a message that gets a basic ack, but we need to then wait for the service to send a message to the client that contains the twin. We could make this a callback, but that seems like an implementation detail that would frustrate users.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
+        protected override void Dispose(bool disposing)
         {
-            cancellationToken.ThrowIfCancellationRequested();
-            EnsureValidState();
-
-            await _channel.WriteAsync(new UnsubscribePacket(0, MethodPostTopicFilter)).ConfigureAwait(true);
+            base.Dispose(disposing);
+            _mqttClient?.Dispose();
+
+            // TODO notify the user for these that they failed? Clear these when closing instead?
```
drwill-ms, line 638: Log them out, maybe.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1043 @@
...
 
-                await responseReceived.WaitAsync(TwinTimeout, cancellationToken).ConfigureAwait(false);
-
-                if (responseException != null)
+                    _getTwinSemaphore.Release();
+                }
+                else if (_inProgressUpdateReportedPropertiesRequests.Contains(receivedRequestId))
                 {
-                    responseException.Throw();
+                    Logging.Info(this, $"Received response to patch twin request with request id {receivedRequestId}.");
```
drwill-ms, line 879: If Logging.IsEnabled. Elsewhere too

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: You could make the call twice in the ternary, or just use that for the parameter and make the call once.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: The concurrent dictionary has its own semaphores. It is thread safe.

Also, this loop could be very tight. Aren't you worried about over taxing the process checking so fast?

iothub/device/src/Transport/Mqtt/PatchTwinResponse.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Construct a new instance of this class.
+        /// </summary>
+        /// <param name="status">The status the service responded with. 204 indicates a successful patch twin request.</param>
+        /// <param name="version">The new version of the twin after the patch.</param>
+        internal PatchTwinResponse(int status, int version)
+        {
+            Status = status;
+            Version = version;
+        }
```
drwill-ms, line 20: Down to 1, but still, why have any?

iothub/device/src/Transport/Mqtt/WillMessage.cs
```diff
@@ -11,20 +9,20 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
     public class WillMessage : IWillMessage
     {
         /// <inheritdoc />
-        public Message Message { get; private set; }
+        public byte[] Payload { get; private set; }
```
drwill-ms, line 12: Properties under ctors

iothub/device/src/TransportSettings/QualityOfService.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// The policy for which a particular message will be sent. Note that IoT hub does not support QoS 2.
+    /// </summary>
+    public enum QualityOfService
```
drwill-ms, line 9: We expose this to users?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: The only point of commonality between every message is the QoS, so I don't see a lot of value in trying to cache this builder

iothub/device/src/TransportSettings/QualityOfService.cs
timtay-microsoft: Users can choose the QoS of the MQTT messages they send, yes

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: There is no blocking call to get an item from a concurrent dictionary, though. The semaphore here isn't for protecting the resource saving these responses. It is for waiting for the service's response to the request this client's invocation of getTwin sent.

timtay-microsoft: If you know of a ```concurrentDictionary.BlockingGet("myKey");``` then that would be perfect here. I haven't found such an implementation in .NET though

timtay-microsoft: I'll take a look at how we handled this in AMQP since we should have the same problem over there

drwill-ms: No way to be notified by the concurrent dictionary for an element.

What if the call to wait never returns?

timtay-microsoft: Then we time out with the cancellation token and throw operation cancelled exception up to the user. That would only happen if the service never sent the twin response message, though

timtay-microsoft: We use TaskCompletionSource in our AMQP layer as the way to wait on another thread to receive a message, so I'll adopt that pattern here as well.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Got rid of these collections, so this is now moot

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+
+            _hostName = context.IotHubConnectionCredentials.IotHubHostName;
+            _connectionCredentials = context.IotHubConnectionCredentials;
+
+            if (_mqttTransportSettings.Protocol == IotHubClientTransportProtocol.WebSocket)
             {
-                ProductInfo = context.ProductInfo,
-                ModelId = context.ModelId,
-            };
+                var uri = $"wss://{_hostName}/$iothub/websocket";
```
patilsnr, line 182: Would this be a place to use `string` instead of `var`? I guess the type is pretty explicit already, so it might be a subjective point.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
 
-            _serverAddresses = null; // this will be resolved asynchronously in OpenAsync
-            _hostName = connectionCredentials.HostName;
-            _receiveEventMessageFilter = string.Format(CultureInfo.InvariantCulture, ReceiveEventMessagePatternFilter, connectionCredentials.DeviceId, connectionCredentials.ModuleId);
-            _receiveEventMessagePrefix = string.Format(CultureInfo.InvariantCulture, ReceiveEventMessagePrefixPattern, connectionCredentials.DeviceId, connectionCredentials.ModuleId);
+                IWebProxy proxy = _transportSettings.Proxy;
+                if (proxy != null)
+                {
+                    var serviceUri = new Uri(uri);
+                    var proxyUri = _transportSettings.Proxy.GetProxy(serviceUri);
```
patilsnr, line 189: Another possible place to use an explicit type if there is one.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                    else
+                    {
+                        _mqttClientOptionsBuilder.WithProxy(proxyUri.AbsoluteUri);
+                    }
+                }
+            }
+            else
+            {
+                // "ssl://" prefix is not needed here
+                var uri = _hostName;
```
patilsnr, line 207: another possible place to use string

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
 
         public override async Task SendEventAsync(IEnumerable<Message> messages, CancellationToken cancellationToken)
         {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // Note that this sends all messages at once and then waits for all the acknowledgements. This
+            // is the recommended pattern for sending large numbers of messages over an asynchronous
+            // protocol like MQTT
+            int index = 0;
+            Task[] sendTasks = new Task[messages.Count()];
```
patilsnr, line 378: Possible use for `var` instead of `Task[]`

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+            base.Dispose(disposing);
+            _mqttClient?.Dispose();
+        }
 
-        public void OnConnected()
+        public override async Task CloseAsync(CancellationToken cancellationToken)
         {
-            if (TryStateTransition(TransportState.Opening, TransportState.Open))
+            OnTransportClosedGracefully();
+            MqttClientDisconnectOptions disconnectOptions = new MqttClientDisconnectOptions();
```
patilsnr, line 665: Consider using `var` instead of `MqttClientDisconnectOptions`

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+            // multi-level wildcard, it receives all messages of a topic that begins with the pattern
+            // before the wildcard character, no matter how long or deep the topic is.
+            //
+            // This is important to add here because topic strings will contain key-value pairs for metadata about
+            // each message. For instance, a cloud to device message sent by the service with a correlation
+            // id of "1" would have a topic like:
+            // "devices/myDevice/messages/devicebound/?$cid=1"
+            string fullTopic = topic + "#";
+
+            MqttClientSubscribeOptions subscribeOptions = new MqttClientSubscribeOptionsBuilder()
```
patilsnr, line 696: `var` instead of `MqttClientSubscribeOptions`

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-            Justification = "The created message is handed to the user and the user application is in charge of disposing the message.")]
-        private async Task HandleIncomingMessagesAsync()
+        private async Task UnsubscribeAsync(string topic, CancellationToken cancellationToken)
         {
-            if (Logging.IsEnabled)
-                Logging.Enter(this, "Process C2D message via callback", nameof(HandleIncomingMessagesAsync));
+            string fullTopic = topic + "#";
 
-            Message message = ProcessC2dMessage();
+            MqttClientUnsubscribeOptions unsubscribeOptions = new MqttClientUnsubscribeOptionsBuilder()
```
patilsnr, line 719: Consider using `var`

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-            finally
-            {
-                if (Logging.IsEnabled)
-                    Logging.Exit(this, message, nameof(OnMessageReceived));
-            }
+            return Task.CompletedTask;
         }
 
-        private async Task HandleIncomingEventMessageAsync(Message message)
+        private async Task HandleReceivedMessage(MqttApplicationMessageReceivedEventArgs receivedEventArgs)
```
patilsnr, line 755: If the method is Async, should it then be titled `HandleRecievedMessageAsync()`?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-                _deviceReceiveMessageSemaphore.Release();
-
                 if (Logging.IsEnabled)
-                    Logging.Exit(this, cancellationToken, nameof(EnableReceiveMessageAsync));
+                    Logging.Error(this, $"Received an MQTT message on unexpected topic {topic}. Ignoring message.");
             }
         }
 
-        public override async Task EnsurePendingMessagesAreDeliveredAsync(CancellationToken cancellationToken)
+        private async Task HandleReceivedCloudToDeviceMessage(MqttApplicationMessageReceivedEventArgs receivedEventArgs)
```
patilsnr, line 790: Another place where `Async` could be part of the method name

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-                case TransportState.Error:
-                    return false;
-
-                case TransportState.NotInitialized:
-                case TransportState.Opening:
-                    _connectCompletion.TrySetCanceled();
-                    break;
+                        try
+                        {
+                            Twin twin = new Twin
```
patilsnr, line 887: Consider using `var`

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-                    return;
+                    if (Logging.IsEnabled)
+                        Logging.Info(this, $"Received response to an unknown twin request with request id {receivedRequestId}. Discarding it.");
                 }
             }
-            await _subscribeCompletionSource.Task.ConfigureAwait(false);
         }
 
-        private Task SubscribeTwinResponsesAsync()
+        private async Task HandleIncomingEventMessage(MqttApplicationMessageReceivedEventArgs receivedEventArgs)
```
patilsnr, line 932: Consider adding `Async` to method name

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-            {
-                await cleanupTask().ConfigureAwait(true);
+            string properties = UrlEncodedDictionarySerializer.Serialize(mergedProperties);
 
-                if (currentCleanupFunc != null)
-                {
-                    await currentCleanupFunc().ConfigureAwait(true);
-                }
-            };
-        }
```
patilsnr, line 1234: This conditional is personally pretty hard to read. Quite subjective opinion though that others are free to disagree with.

patilsnr: Just want to clarify I mean line `1034`

patilsnr (approved):
Left some pretty minor comments, but otherwise looks good!

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -104,14 +104,14 @@ public IotHubClientException(string message, Exception innerException)
         /// <param name="message">The message that describes the error.</param>
         /// <param name="innerException">The exception that is the cause of the current exception.</param>
         /// <param name="isTransient">Indicates if the error is transient and should be retried.</param>
-        protected IotHubClientException(string message, Exception innerException, bool isTransient)
+        public IotHubClientException(string message, Exception innerException, bool isTransient)
```
drwill-ms, line 107: Why public?

iothub/device/src/Transport/Mqtt/GetTwinResponse.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <remarks>
+        /// 200 indicates a successful request.
+        /// </remarks>
+        internal int Status { get; set; }
+
+        /// <summary>
+        /// The twin that the service responded to the get twin request with.
+        /// </summary>
+        /// <remarks>
+        /// This value is null if the get twin request failed.
```
drwill-ms, line 26: Why not throw an exception if the call fails?

drwill-ms: Oh, wait, this is how we parse the payload. Never mind. :)

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+
+        private readonly string _deviceId;
+        private readonly string _moduleId;
+        private readonly string _hostName;
+        private readonly string _modelId;
+        private readonly ProductInfo _productInfo;
+        private readonly IotHubClientMqttSettings _mqttTransportSettings;
+        private readonly IotHubConnectionCredentials _connectionCredentials;
+
+        private static readonly Dictionary<string, string> s_toSystemPropertiesMap = new Dictionary<string, string>
```
drwill-ms, line 111:
```suggestion
        private static readonly Dictionary<string, string> s_toSystemPropertiesMap = new()
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+            {MessageSystemPropertyNames.Operation, MessageSystemPropertyNames.Operation},
+            {MessageSystemPropertyNames.Ack, MessageSystemPropertyNames.Ack},
+            {IotHubWirePropertyNames.ConnectionDeviceId, MessageSystemPropertyNames.ConnectionDeviceId },
+            {IotHubWirePropertyNames.ConnectionModuleId, MessageSystemPropertyNames.ConnectionModuleId },
+            {IotHubWirePropertyNames.MqttDiagIdKey, MessageSystemPropertyNames.DiagId},
+            {IotHubWirePropertyNames.MqttDiagCorrelationContextKey, MessageSystemPropertyNames.DiagCorrelationContext},
+            {IotHubWirePropertyNames.InterfaceId, MessageSystemPropertyNames.InterfaceId}
+        };
+
+        private static readonly Dictionary<string, string> s_fromSystemPropertiesMap = new Dictionary<string, string>
```
drwill-ms, line 131:
```suggestion
        private static readonly Dictionary<string, string> s_fromSystemPropertiesMap = new()
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: It is pretty obviously a string on the right side, so I'd say no.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+
+            tlsParameters.UseTls = true;
+            tlsParameters.SslProtocol = _mqttTransportSettings.SslProtocols;
+            _mqttClientOptionsBuilder.WithTls(tlsParameters);
+
+            _mqttClientOptionsBuilder.WithProtocolVersion(MQTTnet.Formatter.MqttProtocolVersion.V311); // 3.1.1
+
+            _mqttClientOptionsBuilder.WithCleanSession(_mqttTransportSettings.CleanSession);
+
+            _mqttClientOptionsBuilder.WithKeepAlivePeriod(_mqttTransportSettings.KeepAlive);
```
drwill-ms, line 232: You can chain these, IIRC.

```suggestion
            _mqttClientOptionsBuilder
                .WithTls(tlsParameters)
                .WithProtocolVersion(MQTTnet.Formatter.MqttProtocolVersion.V311) // 3.1.1
                .WithCleanSession(_mqttTransportSettings.CleanSession)
                .WithKeepAlivePeriod(_mqttTransportSettings.KeepAlive);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                }
+                else if (_mqttTransportSettings.WillMessage.QualityOfService == QualityOfService.AtLeastOnce)
+                {
+                    _mqttClientOptionsBuilder.WithWillQualityOfServiceLevel(MqttQualityOfServiceLevel.AtLeastOnce);
+                }
             }
 
-            _closeRetryPolicy = new RetryPolicy(new TransientErrorIgnoreStrategy(), 5, TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(1));
+            _mqttClient.ApplicationMessageReceivedAsync += HandleReceivedMessage;
+            _mqttClient.DisconnectedAsync += HandleDisconnection;
```
drwill-ms, line 250: `+=` for events can cause memory leaks if they are never removed. Are you ensuring they are removed when necessary?

I haven't checked the code yet, but at least I need a reminder for myself to check.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                args.Chain,
+                args.SslPolicyErrors);
+        }
 
         #region Client operations
 
         public override async Task OpenAsync(CancellationToken cancellationToken)
         {
-            try
+            string clientId = _moduleId == null ? _deviceId : $"{_deviceId}/{_moduleId}";
```
drwill-ms, line 272: Here you only check for null but on 277 you check for whitespace too. We should be consistent.

Perhaps we should throw an invalid operation exception if it is whitespace?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                        // https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html
+                        // MQTT 5 supports a larger set of connect codes. See the MQTT 5.0 specification section "3.2.2.2 Connect Reason Code"
+                        // https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901074
+                        throw new IotHubClientException("Failed to open the MQTT connection", cfe);
+                }
             }
         }
 
         public override async Task SendEventAsync(Message message, CancellationToken cancellationToken)
         {
```
drwill-ms, line 337: Check message != null?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                        // should never happen. For more details, see the MQTT 3.1.1 specification section "3.2.2.3 Connect Return code"
+                        // https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html
+                        // MQTT 5 supports a larger set of connect codes. See the MQTT 5.0 specification section "3.2.2.2 Connect Reason Code"
+                        // https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901074
+                        throw new IotHubClientException("Failed to open the MQTT connection", cfe);
+                }
             }
         }
 
         public override async Task SendEventAsync(Message message, CancellationToken cancellationToken)
```
drwill-ms, line 336: Consider making publics internals.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                MqttClientPublishResult result = await _mqttClient.PublishAsync(mqttMessage, cancellationToken).ConfigureAwait(false);
 
-                await _channel.WriteAndFlushAsync(message).ConfigureAwait(true);
+                if (result.ReasonCode != MqttClientPublishReasonCode.Success)
+                {
+                    throw new IotHubClientException($"Failed to publish the MQTT packet for message with correlation Id {message.CorrelationId} with reason code {result.ReasonCode}", true);
+                }
             }
-            finally
+            catch (Exception e) when (e is not IotHubClientException && e is not InvalidOperationException && e is not OperationCanceledException)
```
drwill-ms, line 364: We seem to mostly often use `ex` as the variable name.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+            int index = 0;
+            Task[] sendTasks = new Task[messages.Count()];
             foreach (Message message in messages)
             {
-                cancellationToken.ThrowIfCancellationRequested();
-                await SendEventAsync(message, cancellationToken).ConfigureAwait(false);
+                sendTasks[index] = SendEventAsync(message, cancellationToken);
+                index++;
             }
+
```
drwill-ms, line 384: `await Task.WhenAll(messages.Select(x => SendEventAsync(x)).ConfigureAwait(false);`

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-                        {
-                            CleanUpAsync().GetAwaiter().GetResult();
-                        }
+                MqttClientPublishResult result = await _mqttClient.PublishAsync(mqttMessage, cancellationToken).ConfigureAwait(false);
 
-                        _disconnectAwaitersCancellationSource?.Dispose();
-                        _disconnectAwaitersCancellationSource = null;
+                if (result.ReasonCode != MqttClientPublishReasonCode.Success)
+                {
+                    throw new IotHubClientException($"Failed to publish the mqtt packet for getting this client's twin with reason code {result.ReasonCode}", true);
```
drwill-ms, line 558:
```suggestion
                    throw new IotHubClientException($"Failed to publish the MQTT packet for getting this client's twin with reason code {result.ReasonCode}", true);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                if (Logging.IsEnabled)
+                    Logging.Info($"Sent get twin request. Waiting on service response with request id {requestId}");
 
-                        _deviceReceiveMessageSemaphore?.Dispose();
-                        _deviceReceiveMessageSemaphore = null;
+                // Wait until IoT hub sends a message to this client with the response to this patch twin request.
+                GetTwinResponse getTwinResponse =
+                    await GetTaskCompletionSourceResultAsync(
+                        taskCompletionSource,
+                        cancellationToken).ConfigureAwait(false);
```
drwill-ms, line 568:
```suggestion
                GetTwinResponse getTwinResponse = await GetTaskCompletionSourceResultAsync(
                            taskCompletionSource,
                            cancellationToken)
                        .ConfigureAwait(false);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: Also, inline it with the method call.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-            if (_onDesiredStatePatchListener != null)
+            // "#" postfix is a multi-level wildcard in MQTT. When a client subscribes to a topic with a
+            // multi-level wildcard, it receives all messages of a topic that begins with the pattern
+            // before the wildcard character, no matter how long or deep the topic is.
+            //
+            // This is important to add here because topic strings will contain key-value pairs for metadata about
+            // each message. For instance, a cloud to device message sent by the service with a correlation
+            // id of "1" would have a topic like:
+            // "devices/myDevice/messages/devicebound/?$cid=1"
+            string fullTopic = topic + "#";
```
drwill-ms, line 694: Should this be a centralized constant?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
 
-        private async Task SubscribeCloudToDeviceMessagesAsync()
-        {
-            if (TryStateTransition(TransportState.Open, TransportState.Subscribing))
-            {
-                await _channel
-                    .WriteAsync(new SubscribePacket(0, new SubscriptionRequest(_deviceboundMessageFilter, _qosReceivePacketFromService)))
-                    .ConfigureAwait(true);
+                    // This received message is in response to an update reported properties request.
+                    var patchTwinResponse = new PatchTwinResponse()
```
drwill-ms, line 915:
```suggestion
                    var patchTwinResponse = new PatchTwinResponse
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                    {
+                        Status = status,
+                        Version = version,
+                        Message = payloadString,
+                    };
 
-                if (TryStateTransition(TransportState.Subscribing, TransportState.Receiving)
-                    && _subscribeCompletionSource.TrySetResult())
+                    patchTwinCompletion.SetResult(patchTwinResponse);
```
drwill-ms, line 922:
```suggestion
                    patchTwinCompletion.TrySetResult(patchTwinResponse);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
 
-                var clientTlsSettings = new ClientTlsSettings(
-                     settings.SslProtocols,
-                     settings.CertificateRevocationCheck,
-                     certs,
-                     connectionCredentials.HostName);
+            if (property.Key == IotHubWirePropertyNames.AbsoluteExpiryTime ||
+                property.Key == IotHubWirePropertyNames.CreationTimeUtc)
+            {
+                return DateTime.ParseExact(property.Value, "o", CultureInfo.InvariantCulture);
```
drwill-ms, line 999: Can this be DateTimeOffset?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
         }
 
-        private Func<IPAddress[], int, Task<IChannel>> CreateWebSocketChannelFactory(
-            IConnectionCredentials connectionCredentials,
-            AdditionalClientInformation additionalClientInformation,
-            IotHubClientMqttSettings settings)
+        internal static string PopulateMessagePropertiesFromMessage(string topicName, Message message)
         {
-            return async (address, port) =>
+            var systemProperties = new Dictionary<string, string>();
```
drwill-ms, line 1019:
```suggestion
            var systemProperties = new Dictionary<string, string>(message.SystemProperties.Length);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-                    .Pipeline.AddLast(
-                        MqttEncoder.Instance,
-                        new MqttDecoder(false, MaxMessageSize),
-                        new LoggingHandler(LogLevel.DEBUG),
-                        _mqttIotHubAdapterFactory.Create(this, connectionCredentials, additionalClientInformation, settings));
-
-                await s_eventLoopGroup.Value.RegisterAsync(clientWebSocketChannel).ConfigureAwait(true);
+            Dictionary<string, string> mergedProperties = systemProperties.Concat(message.Properties)
+                .ToLookup(x => x.Key, x => x.Value)
+                .ToDictionary(x => x.Key, g => g.First());
```
drwill-ms, line 1030: Confusing alias `g`. Is it a valid? If so `v`?

drwill-ms: FirstOrDefault()?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-            }
+                status = Convert.ToInt32(match.Groups[1].Value, CultureInfo.InvariantCulture);
 
-            if (State == TransportState.Closed)
-            {
-                Debug.Fail($"{nameof(MqttTransportHandler)}.{nameof(EnsureValidState)}: Attempting to reuse transport after it was closed.");
-                throw new InvalidOperationException($"Invalid transport state: {State}");
-            }
+                NameValueCollection queryStringKeyValuePairs = HttpUtility.ParseQueryString(match.Groups[2].Value);
+                rid = queryStringKeyValuePairs.Get("$rid");
```
drwill-ms, line 1060: Constant?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
 
-        private void EnsureValidState(bool throwIfNotOpen = true)
-        {
-            if (State == TransportState.Error)
+            Match match = _twinResponseTopicRegex.Match(topicName);
+            if (match.Success)
             {
-                _fatalException.Throw();
-            }
+                status = Convert.ToInt32(match.Groups[1].Value, CultureInfo.InvariantCulture);
```
drwill-ms, line 1057: Magic numbers at least require some explanation, if not an alias variable to describe what it is.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
+                rid = queryStringKeyValuePairs.Get("$rid");
 
-            if (throwIfNotOpen && (State & TransportState.Open) == 0)
-            {
-                throw new IotHubClientException("MQTT connection is not established. Please retry later.", null, true, IotHubStatusCode.NetworkErrors);
+                if (status == 204)
+                {
+                    // This query string key-value pair is only expected in a successful patch twin response message.
+                    // Get twin requests will contain the twin version in the payload instead.
+                    version = int.Parse(queryStringKeyValuePairs.Get("$version"));
```
drwill-ms, line 1066: Constant

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-                return new MultithreadEventLoopGroup();
+                return await ((Task<T>)finishedTask).ConfigureAwait(false);
             }
 
-            if (Logging.IsEnabled)
-                Logging.Info(null, "EventLoopGroup threads count was not set.");
-
-            return new MultithreadEventLoopGroup();
+            // Otherwise throw operation cancelled exception since the cancellation token was cancelled before the task finished.
+            throw new OperationCanceledException();
```
drwill-ms, line 1097: Message?

iothub/device/src/Transport/Mqtt/WillMessage.cs
```diff
@@ -1,30 +1,17 @@
...
 {
     /// <summary>
-    /// An MQTT "will" message, used when a client disconnects
+    /// An MQTT "will" message to be sent by this client before the client disconnects.
     /// </summary>
     public class WillMessage : IWillMessage
     {
         /// <inheritdoc />
-        public Message Message { get; private set; }
+        public byte[] Payload { get; private set; }
```
drwill-ms, line 12: private set with no ctor?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1097 @@
...
-                await cleanupTask().ConfigureAwait(true);
+            string properties = UrlEncodedDictionarySerializer.Serialize(mergedProperties);
 
-                if (currentCleanupFunc != null)
-                {
-                    await currentCleanupFunc().ConfigureAwait(true);
-                }
-            };
-        }
+            string msg = properties.Length != 0
```
drwill-ms, line 1034: Nick indicates this side. I'd agree, a ternary inside a ternary is confusing.

iothub/device/src/Transport/Mqtt/GetTwinResponse.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
+{
+    /// <summary>
+    /// The service's response to a get twin request.
+    /// </summary>
+    /// <remarks>
+    /// May not contain a twin object if the service did not respond with a twin.
+    /// </remarks>
+    internal class GetTwinResponse
```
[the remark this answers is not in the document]
drwill-ms: Never mind, as discussed may make this suitable for AMQP too.

drwill-ms (approved):
Signing off optimistically.

iothub/device/src/IotHubClientOptions.cs
```diff
@@ -69,11 +69,11 @@ public IotHubClientOptions(IotHubClientTransportSettings transportSettings)
...
         /// </summary>
         public string CustomProductInfo
-        { 
+        {
             get => ProductInfo.Extra;
             set => ProductInfo.Extra = value;
         }
 
-        internal ProductInfo ProductInfo { get; } = new();
+        internal virtual ProductInfo ProductInfo { get; } = new();
```
abhipsaMisra, line 77: Q - why is this virtual?

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
```diff
@@ -58,25 +45,6 @@ public IotHubClientMqttSettings(IotHubClientTransportProtocol transportProtocol
         /// </summary>
         public QualityOfService ReceivingQoS { get; set; }
 
-        /// <summary>
-        /// The maximum no. of inbound messages that are read from the channel.
-        /// The default value is 50.
-        /// </summary>
-        public int MaxPendingInboundMessages { get; set; }
```
abhipsaMisra, line 65: iirc this was used to set the client queue size for c2d messages. Is this no longer configurable?

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
```diff
@@ -58,25 +45,6 @@ public IotHubClientMqttSettings(IotHubClientTransportProtocol transportProtocol
...
-        /// The default is 60 seconds.
-        /// </summary>
-        /// <remarks>
-        /// In the event that IoT hub receives burst traffic, it will implement traffic shaping in order to process the incoming requests.
-        /// In such cases, during client connection the CONNECT requests can have a delay in being acknowledged and processed by IoT hub.
-        /// The <c>ConnectArrivalTimeout</c> governs the duration the client will wait for a CONNACK packet before disconnecting and reopening the connection.
-        /// To know more about IoT hub's throttling limits and traffic shaping feature, see
-        /// <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-quotas-throttling#operation-throttles"/>.
-        /// </remarks>
-        public TimeSpan ConnectArrivalTimeout { get; set; }
```
abhipsaMisra, line 78: Is this also no longer configurable? Do we wait indefinitely as long as the cancellation token is valid?

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
```diff
@@ -97,7 +65,7 @@ public IotHubClientMqttSettings(IotHubClientTransportProtocol transportProtocol
         /// Setting a very low keep-alive value can cause aggressive reconnects, and might not give the
         /// client enough time to establish a connection before disconnecting and reconnecting.
         /// </remarks>
-        public int KeepAliveInSeconds { get; set; }
+        public TimeSpan IdleTimeout { get; set; }
```
abhipsaMisra, line 68: We'll need to update the docs to no longer reference the "divide by 4" logic.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1110 @@
...
+                _mqttClientOptionsBuilder.WithCredentials(username);
+            }
 
-                EnsureValidState(throwIfNotOpen: false);
+            _mqttClientOptions = _mqttClientOptionsBuilder.Build();
 
-                await OpenInternalAsync(cancellationToken).ConfigureAwait(false);
+            try
+            {
+                _mqttClient.DisconnectedAsync += HandleDisconnectionAsync;
```
abhipsaMisra, line 313: since we've already set the disconnection handler and incoming message handler in the constructor (MqttTransportHandler), do we need to set them again here?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1110 @@
...
+                    .WithWillTopic(_deviceToCloudMessagesTopic)
+                    .WithWillPayload(_mqttTransportSettings.WillMessage.Payload);
+
+                if (_mqttTransportSettings.WillMessage.QualityOfService == QualityOfService.AtMostOnce)
                 {
-                    IotHubClientTransportProtocol.Tcp => CreateChannelFactory(connectionCredentials, _additionalClientInformation, settings),
-                    IotHubClientTransportProtocol.WebSocket => CreateWebSocketChannelFactory(connectionCredentials, _additionalClientInformation, settings),
-                    _ => throw new InvalidOperationException($"Unsupported transport setting: {settings.Protocol}"),
-                };
+                    _mqttClientOptionsBuilder.WithWillQualityOfServiceLevel(MqttQualityOfServiceLevel.AtMostOnce);
```
abhipsaMisra, line 245: I was about to suggest that we use MqttQualityOfServiceLevel directlty in our settings, but I think it is a good choice that we're exposing our own type and not mqtt library's. That way we can save our sdk from intentional breaks if mqtt library makes any breaking changes.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1110 @@
...
             {
-                if (Logging.IsEnabled)
-                    Logging.Enter(this, cancellationToken, nameof(SendEventAsync));
+                // Topic depends on if the client is a module client or a device client
+                string baseTopicName = _moduleId == null ? _deviceToCloudMessagesTopic : _moduleToCloudMessagesTopic;
+                string TopicName = PopulateMessagePropertiesFromMessage(baseTopicName, message);
+
+                if (message.HasPayload && message.Payload.Length > MaxMessageSize)
+                {
+                    throw new InvalidOperationException($"Message size ({message.Payload.Length} bytes) is too big to process. Maximum allowed payload size is {MaxMessageSize}");
```
abhipsaMisra, line 354: The max message size here is what IoT hub defines. However, they also include other metadata in that calculation, which means there's a small chance we're allowing some messages that hub would eventually refuse. Should we remove any validation from our client and delegate all that to the service directly?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1110 @@
...
-                cancellationToken.ThrowIfCancellationRequested();
+                var mqttMessage = new MqttApplicationMessageBuilder()
+                    .WithTopic(TopicName)
+                    .WithPayload(message.Payload)
+                    .WithQualityOfServiceLevel(publishingQualityOfService)
+                    .Build();
 
-                EnsureValidState();
-                Debug.Assert(_channel != null);
+                MqttClientPublishResult result = await _mqttClient.PublishAsync(mqttMessage, cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 363: This is so much more straightforward that dotnetty, it's great!

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1110 @@
...
 
         public override async Task SendEventAsync(IEnumerable<Message> messages, CancellationToken cancellationToken)
         {
-            foreach (Message message in messages)
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // Note that this sends all messages at once and then waits for all the acknowledgements. This
+            // is the recommended pattern for sending large numbers of messages over an asynchronous
+            // protocol like MQTT
+            await Task.WhenAll(messages.Select(x => SendEventAsync(x, cancellationToken))).ConfigureAwait(false);
```
abhipsaMisra, line 383: How is error handling managed here, i.e. if one send fails.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -4,1310 +4,1110 @@
...
+                {
+                    throw new IotHubClientException($"Failed to publish the MQTT packet for message with correlation Id {message.CorrelationId} with reason code {result.ReasonCode}", true);
+                }
             }
-            finally
+            catch (Exception ex) when (ex is not IotHubClientException && ex is not InvalidOperationException && ex is not OperationCanceledException)
             {
-                if (Logging.IsEnabled)
-                    Logging.Exit(this, cancellationToken, nameof(SendEventAsync));
+                throw new IotHubClientException("Failed to send message.", true, ex);
```
abhipsaMisra, line 372: Is there a reason we're deciding to default to retryable? Wouldn't it be better to default to not-retry and retry only for known cases? I say this because of data/bandwidth concerns.
We had a case where the sdk was not returning the correct exception, which should have been non-retryable. I think it was a message-size related error. Instead, the sdk returned the generic communication exception, which was retryable and the customer eventually ran up a huge bill because they were continuously retrying needlessly.

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
timtay-microsoft: > Is this also no longer configurable? Do we wait indefinitely as long as the cancellation token is valid?

Yes, that is the idea now. It does appear that MQTTNet's client has a configurable "operation timeout" of 100 seconds though.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: I'd like to do that, sure.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: I know, right!?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Picking a default value here is hard, yeah. I can make this false if we are concerned about cases like that, sure.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: I'll make it so we += after open and -= before closing. That way we don't have += at constructor time and += at open time

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
timtay-microsoft: MQTTNet does appear to allow us to configure this, but I don't know of any use cases for configuring this value. We could add it back in later if a customer asks for it

iothub/device/src/IotHubClientOptions.cs
timtay-microsoft: For mocking purposes within the SDK. While writing an MqttTransportHandler test, I needed to be able to mock this type

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: An AggregateException would be thrown. Is there a better way to handle something like this?

abhipsaMisra: I think a comment in the public API doc guiding users to look at the inner exception for the actual error reason will be helpful.

PR #2381, opened 2022-05-09
Use windows 2022 for pipelines and add .net 6.0 testing

drwill-ms:


e2e/test/E2ETests.csproj
```diff
@@ -1,7 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1.18;net472;net451</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net5.0;netcoreapp3.1;netcoreapp2.1.18</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net6.0;net5.0;netcoreapp3.1;netcoreapp2.1.18;net472;net451</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net6.0;net5.0;netcoreapp3.1;netcoreapp2.1.18</TargetFrameworks>
+    <!-- By setting this as not a test project for other frameworks, we can run "dotnet test" for a specific framework from the root directory without dotnet throwing if any test project in the solution doesn't support that framework -->
+    <IsTestProject Condition="'$(OS)' != 'Windows_NT' And '$(TargetFramework)' != 'net6.0' And '$(TargetFramework)' != 'net5.0' And '$(TargetFramework)' != 'netcoreapp3.1' And '$(TargetFramework)' != 'netcoreapp2.1.18'">False</IsTestProject>
```
abhipsaMisra, line 6: nit: instead of a double negative `if (OS != Window) then evaluate expression to "false"`, maybe we could reverse the check?

abhipsaMisra (approved):

e2e/test/E2ETests.csproj
drwill-ms: In order to turn IsTestProject to false, it needs to be the conditions that make it not Linux, meaning not windows and any framework not listed above for Linux.

brycewang-microsoft (approved):

PR #2384, opened 2022-05-11
Pre-v2 branching code cleanup

drwill-ms:
Making non-breaking changes that apply to both v1 and v2 to cut down on rebasing hell when starting an active v2 feature branch. This is just a subset of all projects.

### DIFF
--- a/common/src/device/provisioning/transport/X509CertificateInfo.cs
+++ b/common/src/device/provisioning/transport/X509CertificateInfo.cs
@@ -18,21 +18,21 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport.Models
         /// </summary>
         public X509CertificateInfo()
         {
-          CustomInit();
+            CustomInit();
         }
 
         /// <summary>
         /// Initializes a new instance of the X509CertificateInfo class.
         /// </summary>
         public X509CertificateInfo(
-            string subjectName = default(string), 
-            string sha1Thumbprint = default(string), 
-            string sha256Thumbprint = default(string), 
-            string issuerName = default(string), 
-            DateTime? notBeforeUtc = default(DateTime?), 
-            DateTime? notAfterUtc = default(DateTime?), 
-            string serialNumber = default(string), 
-            int? version = default(int?))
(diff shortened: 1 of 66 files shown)


common/src/device/provisioning/transport/X509CertificateInfo.cs
```diff
@@ -51,41 +51,49 @@ public X509CertificateInfo()
...
         public string Sha1Thumbprint { get; set; }
 
         /// <summary>
+        /// The certificate SHA256 thumbprint.
         /// </summary>
         [JsonProperty(PropertyName = "sha256Thumbprint")]
         public string Sha256Thumbprint { get; set; }
 
         /// <summary>
+        /// The certfificate issuer name.
```
timtay-microsoft, line 72:
```suggestion
        /// The certificate issuer name.
```

common/src/device/provisioning/transport/X509CertificateInfo.cs
```diff
@@ -51,41 +51,49 @@ public X509CertificateInfo()
...
         /// </summary>
         [JsonProperty(PropertyName = "issuerName")]
         public string IssuerName { get; set; }
 
         /// <summary>
+        /// The certfificate invalidity before date in UTC.
```
timtay-microsoft, line 78:
```suggestion
        /// The certificate invalidity before date in UTC.
```

provisioning/transport/amqp/src/SaslTpmHandler.cs
```diff
@@ -267,7 +264,7 @@ private enum SaslChallengeAction
         {
             First,
             Interim,
-            Final
+            Final,
```
timtay-microsoft, line 267: I don't expect this enum to grow any further, for what it's worth.

timtay-microsoft (approved):

provisioning/transport/amqp/src/SaslTpmHandler.cs
drwill-ms: Yeah, I just style them this way as a rule, for consistency.

PR #2386, opened 2022-05-11
Code cleanup for DPS clients

drwill-ms:


provisioning/service/src/Config/AttestationMechanism.cs
```diff
@@ -129,111 +82,47 @@ private AttestationMechanism(AttestationMechanismType type, TpmAttestation tpm,
         /// <returns>The attestation of this object.</returns>
         public Attestation GetAttestation()
         {
-            switch (Type)
+            return Type switch
```
timtay-microsoft, line 85: This is pretty neat

timtay-microsoft (approved):

provisioning/service/src/Config/AttestationMechanism.cs
drwill-ms: yeah, a newer C# feature. I'm not all the way used to it, but getting more used to seeing it.

brycewang-microsoft (approved):

PR #2389, opened 2022-05-17
Hub service client code cleanup

drwill-ms:


iothub/service/src/Common/Data/SharedAccessSignatureAuthorizationRule.cs
```diff
@@ -60,12 +60,12 @@ public bool Equals(SharedAccessSignatureAuthorizationRule other)
                 return false;
             }
 
-            bool equals = string.Equals(KeyName, other.KeyName, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(PrimaryKey, other.PrimaryKey, StringComparison.Ordinal) &&
-                string.Equals(SecondaryKey, other.SecondaryKey, StringComparison.Ordinal) &&
-                Equals(Rights, other.Rights);
+            bool areEqual = string.Equals(KeyName, other.KeyName, StringComparison.OrdinalIgnoreCase)
```
timtay-microsoft, line 63: Just return this value instead of assigning a local variable?

common/src/service/CommonConstants.cs
```diff
@@ -1,14 +1,14 @@
...
+using System;
+
 namespace Microsoft.Azure.Devices.Common
 {
-    using System;
-
     internal static class CommonConstants
     {
         // Custom HTTP response contents
-        internal const string ErrorCode = "errorCode";
```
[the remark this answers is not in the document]
drwill-ms: The class is internal, which makes it impossible for any members to be public, and any "public" members are effectively internal. Why make this one public? All the others were and I changed it to be consistent.

iothub/service/src/Common/Amqp/LegacyClientWebSocketTransport.cs
```diff
@@ -1,63 +1,52 @@
...
-            this.operationTimeout = operationTimeout;
-            this.localEndPoint = localEndpoint;
-            this.remoteEndPoint = remoteEndpoint;
-            this.asyncReadBufferSize = MaxReadBufferSize; // TODO: read from Config Settings
-            this.asyncReadBuffer = new byte[this.asyncReadBufferSize];
+            _webSocket = webSocket;
+            _operationTimeout = operationTimeout;
+            _localEndPoint = localEndpoint;
+            _remoteEndPoint = remoteEndpoint;
+            _asyncReadBufferSize = MaxReadBufferSize; // TODO: read from Config Settings
```
andyk-ms, line 39: Should we go ahead with reading from config settings?

timtay-microsoft (approved):

iothub/service/src/Common/Amqp/LegacyClientWebSocketTransport.cs
drwill-ms: I don't know what is entailed in that TODO, and I wouldn't want to do it as part of simple code cleanup. @timtay-microsoft, do you have any context on this work item?

andyk-ms (approved):

PR #2391, opened 2022-05-17
Code cleanup for hub device common folder

drwill-ms:


iothub/device/src/Common/ExceptionTrace.cs
```diff
@@ -103,50 +103,19 @@ public TException TraceException<TException>(TException exception, TraceEventTyp
...
 
+#if NET451
                 switch (level)
                 {
                     case TraceEventType.Critical:
                     case TraceEventType.Error:
-#if NET451
-                        Trace.TraceError("An Exception is being thrown: {0}", GetDetailsForThrownException(exception));
-#endif
-                        ////if (MessagingClientEtwProvider.Provider.IsEnabled(
```
drwill-ms, line 113: I presume this is dead code, so I just removed it. @timtay-microsoft, any context to sway this change?

timtay-microsoft: No objections here. This class as a whole seems out of place so I'm not sure what the commented out code was for.

iothub/device/src/Common/Exceptions/IotHubException.cs
```diff
@@ -19,14 +19,14 @@ public class IotHubException : Exception
         private const string TrackingIdValueSerializationStoreName = "IotHubException-TrackingId";
 
         /// <summary>
-        /// Creates an instance of <see cref="IotHubException"/> an empty error message.
+        /// Creates an instance of this class an empty error message.
```
timtay-microsoft, line 22:
```suggestion
        /// Creates an instance of this class with an empty error message.
```

timtay-microsoft (approved):

PR #2394, opened 2022-05-19
Generate X509 certificates on-demand for DPS E2E tests

abhipsaMisra:


e2e/test/provisioning/ProvisioningServiceClientE2ETests.cs
```diff
@@ -115,30 +129,44 @@ public async Task ProvisioningServiceClient_GetEnrollmentGroupAttestation_Symmet
...
+            IndividualEnrollment individualEnrollment = await CreateIndividualEnrollmentAsync(
+                provisioningServiceClient,
+                registrationId,
+                attestationType,
+                null,
+                null,
+                AllocationPolicy.Static,
+                null,
+                null,
+                null);
```
drwill-ms, line 143:
```suggestion
            IndividualEnrollment individualEnrollment = await CreateIndividualEnrollmentAsync(
                    provisioningServiceClient,
                    registrationId,
                    attestationType,
                    null,
                    null,
                    AllocationPolicy.Static,
                    null,
                    null,
                    null)
                .ConfigureAwait(false);
```

timtay-microsoft:
I'd consider moving the cert generation logic to a helper class so that we can leverage it in non-DPS test scenarios. That way, we could dynamically generate certificates for basic IoT Hub send message tests as well.

drwill-ms (approved):

PR #2397, opened 2022-05-19
feat(prov-device): Add new device registration substatus

abhipsaMisra:
This change was introduced in API version "2019-04-15".

The ‘ReprovisionedToInitialAssignment’ value is simply an extension to the original “InitialAssignment” value.

The ‘InitialAssigment’ value was being returned only when a device was being provisioned into Hub for the first time (from a DPS standpoint).

Starting with API version "2019-04-15”, the substatus returned would be “ReprovisionedtoInitialAssignment” if the device previously existed in Hub, but DPS was unaware and assumed that it was provisioning it for the first time.

Reprovisioning a device (which means that DPS is aware that it previously provisioned this device) would return different substatus values like ‘deviceDataMigrated’ or ‘ deviceDataReset’.

### DIFF
--- a/provisioning/device/src/ProvisioningRegistrationSubstatus.cs
+++ b/provisioning/device/src/ProvisioningRegistrationSubstatus.cs
@@ -23,6 +23,15 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
         /// Device has been assigned to a different IoT hub and its device data was populated from the initial state stored in the enrollment.
         /// Device data was removed from the previously assigned IoT hub.
         /// </summary>
-        DeviceDataReset = 3
+        DeviceDataReset = 3,
+
+        /// <summary>
+        /// Device has been re-provisioned to a previously assigned IoT hub.
+        /// </summary>
+        /// <remarks>
+        /// For API versions prior to "2019-04-15" a substatus of <see cref="InitialAssignment"/> was returned if the device previously existed in hub.
+        /// Starting API version "2019-04-15" a substatus of ReprovisionedToInitialAssignment is returned if the device previously existed in hub.
+        /// </remarks>
+        ReprovisionedToInitialAssignment = 4,
     }
 }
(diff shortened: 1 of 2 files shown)


provisioning/device/src/ProvisioningRegistrationSubstatus.cs
```diff
@@ -23,6 +23,15 @@ public enum ProvisioningRegistrationSubstatusType
...
+        DeviceDataReset = 3,
+
+        /// <summary>
+        /// Device has been re-provisioned to a previously assigned IoT hub.
+        /// </summary>
+        /// <remarks>
+        /// For API versions prior to "2019-04-15" a substatus of <see cref="InitialAssignment"/> was returned if the device previously existed in hub.
+        /// Starting API version "2019-04-15" a substatus of ReprovisionedToInitialAssignment is returned if the device previously existed in hub.
+        /// </remarks>
+        ReprovisionedToInitialAssignment = 4,
```
drwill-ms, line 35: This is a breaking change from the service, no?

drwill-ms: Also, customers won't know what API version we use, so we should probably indicate relevant SDK version, if we were to simply take this change.

abhipsaMisra: Yes, it is. They have signed off on it and have kept it behind an API version change.

abhipsaMisra: > Also, customers won't know what API version we use, so we should probably indicate relevant SDK version, if we were to simply take this change.

That's a good point. I'll add a note to update this once we are ready to put the release out (so that we can reference the latest nuget version numbers).

drwill-ms: > Yes, it is. They have signed off on it and have kept it behind an API version change.

Who signed off on it? API version changes don't protect customers who upgrade from 1 SDK version to the next. It is still a breaking change.

abhipsaMisra:
The DPS API version update brings this breaking behavior change and should ideally be GA'd in our v2 clients (which will have other breaking changes). This PR is being merged into preview.

brycewang-microsoft (approved):

PR #2401, opened 2022-05-20
Reorganize device client folder structure

timtay-microsoft:
No file has had its contents altered. It will probably be easier to review this PR by viewing the folder structure in your IDE or on github [here](https://github.com/Azure/azure-iot-sdk-csharp/tree/timtay/reorganize/iothub/device/src)

New folders include "Authentication", "ClientSettings", "DirectMethod", "FileUpload", and "Pipeline"


brycewang-microsoft (approved):

abhipsaMisra:
Since we have transport specific folders, what do you think about moving the transport settings files to the protocol specific folder instead? That way, anyone looking for mqtt files will find all of them together in one place.

timtay-microsoft:
> Since we have transport specific folders, what do you think about moving the transport settings files to the protocol specific folder instead? That way, anyone looking for mqtt files will find all of them together in one place.

I considered that, but since these options are user facing, I didn't want to bury them quite so deep

PR #2403, opened 2022-05-20
refactor PipelineContext class for better understandability

timtay-microsoft:
Previously, this class contained all the relevant settings, callbacks, and more in it, but was defined in such a way that didn't make that obvious. Instead of PipelineContext just containing a map of these objects, it will now have properties for each one.


iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -49,7 +49,7 @@ static AmqpTransportHandler()
         {
             _operationTimeout = transportSettings.OperationTimeout;
             _onDesiredStatePatchListener = onDesiredStatePatchReceivedCallback;
-            IDeviceIdentity deviceIdentity = new DeviceIdentity(connectionString, transportSettings, context.Get<ProductInfo>(), context.Get<ClientOptions>());
```
timtay-microsoft, line 52: Here is an example of this class's previous usage. You would call ```.Get<SomeClass>``` to get the instance of ```SomeClass``` saved within it. Someone who wants to use this class but isn't familiar with what fields it actually contains would have no way of knowing what all could be in there though. Hence, changing these expected fields into properties makes it a lot easier to understand how to use this pipeline context class

abhipsaMisra (approved):
I like it.

PR #2409, opened 2022-05-23
Reorganize IoT hub service client code

timtay-microsoft:
There are some classes without a proper home for now, but that will be resolved when we refactor these clients for v2.

No file has had its contents altered. It will probably be easier to review this PR by viewing the folder structure in your IDE or on github [here](https://github.com/Azure/azure-iot-sdk-csharp/tree/timtay/reorganizeServiceClient/iothub/service/src)


brycewang-microsoft (approved):
It looks pretty clean! Btw, the github link above needs to correct into [this](https://github.com/Azure/azure-iot-sdk-csharp/tree/timtay/reorganizeServiceClient/iothub/service/src).

PR #2411, opened 2022-05-24
Update provisioning tests to retry on throttling

abhipsaMisra:
One of the biggest pain points while working on the feature for DPS client certs was our E2E test limitation of reusing the same enrollment entry for X509 certificate-based enrollments. Once I modified the tests to generate X509 certificate-based enrollments on-demand, our DPS infrastructure was unable to handle the volume of requests it was receiving from our E2E tests to create enrollments. As a result, most of our tests kept on getting throttled and would subsequently fail.

This PR adds retry logic for provisioning service client failures with error codes 429 (throttling) and 5xx (server errors).

### DIFF
--- a/e2e/test/provisioning/ReprovisioningE2ETests.cs
+++ b/e2e/test/provisioning/ReprovisioningE2ETests.cs
@@ -414,8 +414,8 @@ namespace Microsoft.Azure.Devices.E2ETests.Provisioning
                             string primaryKeyEnrollmentGroup = symmetricKeyAttestation.PrimaryKey;
                             string secondaryKeyEnrollmentGroup = symmetricKeyAttestation.SecondaryKey;
 
-                            string primaryKeyIndividual = ProvisioningE2ETests.ComputeDerivedSymmetricKey(Convert.FromBase64String(primaryKeyEnrollmentGroup), registrationIdSymmetricKey);
-                            string secondaryKeyIndividual = ProvisioningE2ETests.ComputeDerivedSymmetricKey(Convert.FromBase64String(secondaryKeyEnrollmentGroup), registrationIdSymmetricKey);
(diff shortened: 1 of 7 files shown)


timtay-microsoft:
Do we need this for creating enrollment groups as well? Or do we not create enough of them to warrant adding retry to them?

timtay-microsoft (approved):

abhipsaMisra:
> Do we need this for creating enrollment groups as well? Or do we not create enough of them to warrant adding retry to them?

Right now, we don't create as many enrollment groups as we create individual enrollments. That said, I'm working on creating X509 based enrollment groups on-demand as well. With that change I'll be adding the retry policies to enrollment groups as well.

e2e/test/provisioning/ReprovisioningE2ETests.cs
```diff
@@ -542,27 +550,47 @@ private async Task ConfirmDeviceWorksAfterReprovisioning(DeviceRegistrationResul
...
 
-                    if (reprovisionPolicy.MigrateDeviceData)
-                    {
-                        Assert.AreNotEqual(1, twin.Properties.Desired.Count);
-                        Assert.AreNotEqual(0, twin.Properties.Desired.Version);
-                        Assert.AreEqual(ProvisioningRegistrationSubstatusType.DeviceDataMigrated, result.Substatus);
-                    }
-                    else if (reprovisionPolicy.UpdateHubAssignment)
+                    // Reprovision
+                    if (reprovisionPolicy.UpdateHubAssignment)
```
abhipsaMisra, line 560: The pervious assertions seemed confusing to me, since the logic relied on `reprovisionPolicy.MigrateDeviceData` and `reprovisionPolicy.UpdateHubAssignment` being mutually exclusive, while they are not.
I reworked the assertions based on the explanation here: https://docs.microsoft.com/en-us/azure/iot-dps/concepts-device-reprovision

@timtay-microsoft I added this in now; could you give this another look? I'd like to get another pair of eyes on this before I merge it in.

PR #2414, opened 2022-05-25
refactor(iot-device): Add additional logging statement to disposal flow

abhipsaMisra:
image

### DIFF
--- a/iothub/device/src/Authentication/AuthenticationWithTokenRefresh.cs
+++ b/iothub/device/src/Authentication/AuthenticationWithTokenRefresh.cs
@@ -101,6 +101,14 @@ namespace Microsoft.Azure.Devices.Client
         {
             if (_isDisposed)
             {
+                if (Logging.IsEnabled)
+                {
+                    Logging.Error(
+                        this,
+                        $"Encountered {nameof(ObjectDisposedException)} - The authentication method instance has already been disposed, so this client is no longer usable.",
+                        nameof(GetTokenAsync));
+                }
+
                 throw new ObjectDisposedException(GetType().Name, "The authentication method instance has already been disposed, so this client is no longer usable. " +
                     "Please close and dispose your current client instance. To continue carrying out operations from your device/ module, " +
                     "create a new authentication method instance and use it for reinitializing your client.");
@@ -177,15 +185,30 @@ namespace Microsoft.Azure.Devices.Client
         /// <param name="disposing">true to release both managed and unmanaged resources; false to release only unmanaged resources.</param>
         protected virtual void Dispose(bool disposing)
         {
-            if (!_isDisposed)
+            try
             {
-                if (disposing)
+                if (Logging.IsEnabled)
                 {
-                    _lock?.Dispose();
-                    _lock = null;
+                    Logging.Enter(this, $"Disposed={_isDisposed}; disposing={disposing}", $"{nameof(AuthenticationWithTokenRefresh)}.{nameof(Dispose)}");
                 }
 
-                _isDisposed = true;
+                if (!_isDisposed)
+                {
+                    if (disposing)
+                    {
+                        _lock?.Dispose();
+                        _lock = null;
+                    }
+
+                    _isDisposed = true;
+                }
+            }
+            finally
+            {
+                if (Logging.IsEnabled)
+                {
+                    Logging.Exit(this, $"Disposed={_isDisposed}; disposing={disposing}", $"{nameof(AuthenticationWithTokenRefresh)}.{nameof(Dispose)}");
+                }
             }
         }
 
--- a/iothub/device/src/Pipeline/TransportHandler.cs
+++ b/iothub/device/src/Pipeline/TransportHandler.cs
@@ -25,15 +25,30 @@ namespace Microsoft.Azure.Devices.Client.Transport
 
         protected override void Dispose(bool disposing)
         {
-            if (_disposed)
+            try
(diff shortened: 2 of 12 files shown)


iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -616,23 +616,32 @@ private void TwinMessageListener(Twin twin, string correlationId, TwinCollection
...
+            finally
+            {
                 if (Logging.IsEnabled)
-                    Logging.Info(this, nameof(disposing));
-
-                if (disposing)
                 {
-                    _closed = true;
-                    AmqpUnitManager.GetInstance()?.RemoveAmqpUnit(_amqpUnit);
-                    _disposed = true;
```
abhipsaMisra, line 635: Removed setting `_disposed` to `true`. The `_disposed` flag is inherited from the base class `DefaultDelegatingHandler` and is finally set to null there.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -952,38 +952,49 @@ public void Dispose()
...
+                        _messageReceivingCallbackSemaphore?.Dispose();
+                        _eventReceivingLinkSemaphore?.Dispose();
+                        _methodLinkSemaphore?.Dispose();
+                        _twinLinksSemaphore?.Dispose();
+
+                        Logging.Exit(this, disposing, nameof(Dispose));
+                    }
+                }
 
+                _disposed = true;
```
abhipsaMisra, line 990: added `_disposed = true`.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -452,33 +452,46 @@ public override Task RejectAsync(string lockToken, CancellationToken cancellatio
 
         protected override void Dispose(bool disposing)
         {
-            if (_disposed)
-            {
-                return;
-            }
-
-            base.Dispose(disposing);
```
abhipsaMisra, line 460: removed the call to `base.Dispose()`. The transport handler Dispose calls are being made through the delegating handler "inner handler" dispose call. So calling back the base class dispose isn't required.
https://github.com/Azure/azure-iot-sdk-csharp/blob/bf5b6997486fa47691efd5fba999db2f2078b7f8/iothub/device/src/Pipeline/DefaultDelegatingHandler.cs#L229

iothub/device/src/Transport/Amqp/AmqpUnit.cs
abhipsaMisra: moved the `_disposed = true` setter to after disposing block has executed

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: I've reverted this change for now. I'll rework the inheritance disposal pattern of our pipeline (if necessary) in the next PR. I'm running some tests to confirm the behavior.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
abhipsaMisra: I've reverted this change for now. I'll rework the inheritance disposal pattern of our pipeline (if necessary) in the next PR. I'm running some tests to confirm the behavior.

iothub/device/src/Authentication/AuthenticationWithTokenRefresh.cs
```diff
@@ -101,6 +101,14 @@ public async Task<string> GetTokenAsync(string iotHub)
         {
             if (_isDisposed)
             {
+                if (Logging.IsEnabled)
+                {
+                    Logging.Error(
+                        this,
+                        $"Encountered {nameof(ObjectDisposedException)} - The authentication method instance has already been disposed, so this client is no longer usable.",
```
timtay-microsoft, line 108: Is this redundant since we are throwing an ObjectDisposedException right afterwards?

timtay-microsoft (approved):

iothub/device/src/Pipeline/TransportHandler.cs
```diff
@@ -25,15 +25,30 @@ public override Task WaitForTransportClosedAsync()
...
         protected override void Dispose(bool disposing)
         {
-            if (_disposed)
+            try
             {
-                return;
-            }
+                if (Logging.IsEnabled)
+                {
+                    Logging.Enter(this, $"{nameof(DefaultDelegatingHandler)}.Disposed={_disposed}; disposing={disposing}", $"{nameof(TransportHandler)}.{nameof(Dispose)}");
```
drwill-ms, line 32: Why `DefaultDelegatingHandler`?

abhipsaMisra: The `_disposed` parameter referenced throughout our delegating handler pipeline is actually the protected variable defined in `DefaultDelegatingHandler`: https://github.com/Azure/azure-iot-sdk-csharp/blob/929a6d36ee8dbf9f66f0a7829f8c4c0c489a9b29/iothub/device/src/Pipeline/DefaultDelegatingHandler.cs#L15

This, along with the way `base.Dispose()` and `inherited.Dispose()` are mixed in our pipeline disposal flow is something I'm looking to refactor.

PR #2418, opened 2022-05-31
Merge in latest from main

drwill-ms:


brycewang-microsoft (approved):

timtay-microsoft:
@drwill-ms  We typically don't squash when merging from main to preview, FYI. Be careful with that auto-squash

drwill-ms:
> @drwill-ms We typically don't squash when merging from main to preview, FYI. Be careful with that auto-squash

Totally right. Thanks for catching that.

PR #2422, opened 2022-06-02
Fix preprocessor directives to be more platform inclusive

drwill-ms:


common/src/service/StringValidationHelper.cs
```diff
@@ -58,7 +58,7 @@ public static bool IsNullOrBase64String(string value)
 
         public static bool IsBase64String(string value)
         {
-#if NETSTANDARD2_1
+#if NETSTANDARD2_1_OR_GREATER || NET5_0_OR_GREATER
```
abhipsaMisra, line 61: nit: the .NET API surface went from .NET Standard 2.1 -> .NET 5.0. So, `NETSTANDARD2_1` should be the same as `NETSTANDARD2_1_OR_GREATER`

abhipsaMisra (approved):

common/src/service/StringValidationHelper.cs
drwill-ms: I think it is better to keep these preprocessor directives future proof. The API we're enabling below was introduced with netstandard2.1 so anything it and higher should have it. Even if our SDK doesn't target netstandard2.2+ right now, if it ever did we wouldn't have to edit all these directives.

FWIW, I found that net5.0 isn't included in NETSTANDARD2_1_OR_GREATER so I added that too.

Issue #2434, opened 2022-06-09
Constrained module-to-module message rate using AMQP


vipeller:
When we use the SDK to write a module that receives M2M messages (from edgeHub), the message rate is very limited using AMQP protocol, compared to MQTT. It does not go above ~40 msg/sec, and probably it is because of the following issue:

When edgeHub sends an m2m message to a client module, it waits for acknowledgement before sending out the next message. This acknowledgement is delayed and the time it arrives is around 20-30ms. This gives a 1000/20 = 50msg/sec upper limit.

We had the very same problem with edgeHub a while ago with AMQP (when modules sent messages to edgehub using AMQP) and the problem was the following, which I think the SDK also has:

I believe the acknowledgement in the SDK code is sent by this code:

https://cs.github.com/Azure/azure-iot-sdk-csharp/blob/3261abae3abcad23a73cb1763daac1825749fcf6/iothub/service/src/Feedback/AmqpFeedbackReceiver.cs#L142

the last 'true' parameter is a 'batchable' parameter. The following screenshots are from the investigation from a while ago with edgeHub, but it must be very similar with the SDK (sorry if not, and this is misleading). So, the 'batchable' parameter sets a flag, which then later causes a timer to kick-in before the acknowledgement gets sent back, and this timeout has a 20ms value.

![image](https://user-images.githubusercontent.com/51135538/172902403-aead8a70-9037-4cd4-97d4-2fd447e21b66.png)

Because edgeHub always sends a single message at a time, this batching feature never will be used and just slows things down. I advise to turn that flag off to increase M2M message rate.

This issue got highlighted because customers use edgeHub in more and more complex scenarios and we have an ongoing IcM where they hit the 50msg/sec limit.

Please, let me know if the 50msg limit is caused by the problem above and if it can be fixed.

timtay-microsoft:
#2439 should fix this issue. I'll update this thread again once we've released it

timtay-microsoft:
This issue's fix has been released as of version 1.41.1 so I'm closing this thread

PR #2439, opened 2022-06-14
refactor(iot-serv): Stop batching ack's for c2d feedback messages

timtay-microsoft:
The service only sends the next c2d feedback message after the previous one has been ack'd, so it doesn't make sense to delay sending the ack in case another feedback message is received.

We currently do not batch the ack's for file upload notifications, so this brings the feedback receiver to par with the file upload receiver.

#2434


iothub/service/src/Feedback/AmqpFeedbackReceiver.cs
```diff
@@ -145,7 +145,7 @@ public override Task CompleteAsync(FeedbackBatch feedback)
                 _faultTolerantReceivingLink,
                 feedback.LockToken,
                 AmqpConstants.AcceptedOutcome,
-                true);
```
timtay-microsoft, line 148: All of these true -> false changes are for the "batchable" parameter in this method. We don't need to add or remove any logic in how we batch to make this change since it is all already in place.

iothub/service/src/Feedback/AmqpFeedbackReceiver.cs
```diff
@@ -145,7 +145,7 @@ public override Task CompleteAsync(FeedbackBatch feedback)
...
+                false);
```
abhipsaMisra, line 148: I think it would be useful to add the description that you have in the PR topic as a comment on these methods:
> The service only sends the next c2d feedback message after the previous one has been ack'd, so it doesn't make sense to delay sending the ack in case another feedback message is received.

abhipsaMisra (approved):

Issue #2218, opened 2021-10-22
[Bug Report] Calling SetReceiveMessageHandlerAsync(null, null) with Amqp will put connection in bad state


jamdavi:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 Desktop x64
- **Application's .NET Target Framework :** netcore3.1
- **Device:** Laptop
- **SDK version used:** 1.38.0

## Description of the issue
When testing a detach operation by setting the handler to null the device client (and module client) will correctly detach but it will cause the client to disconnect and reconnect.

This is due to the device will call [SafeClose()](https://github.com/Azure/azure-iot-sdk-csharp/blob/5349c3f7b849c6ea65ac521510e5838c63fb532b/iothub/device/src/Transport/Amqp/AmqpUnit.cs#L230-L233) against the session when the receiver link is closed. It does this regardless of the reason it's called.

`Status Disconnected_Retrying, Reason Communication_Error`

## Code sample exhibiting the issue

```csharp
 static async Task Main(string[] args)
 {
        var devClient = DeviceClient.CreateFromConnectionString(<<CONNSTRING>>, TransportType.Amqp);
        await devClient.SetReceiveMessageHandlerAsync(MessageHandlerMethod, null);
        await Task.Delay(10000);
        await devClient.SetReceiveMessageHandlerAsync(null, null);
        await Task.Delay(int.MaxValue);
}
```

## Console log of the issue
```
2021-10-22T16:22:01.6224864>> Trace - [Microsoft-Azure-Devices-Device-Client-Enter] (InternalClient#36038289, SetReceiveMessageHandlerAsync, ((null), (null))).
2021-10-22T16:22:01.6249817>> Trace - [Microsoft-Azure-Devices-Device-Client-Enter] (RetryDelegatingHandler#49972132, DisableReceiveMessageAsync, (CancellationToken#3841804)).
2021-10-22T16:22:01.6264194>> Trace - [Microsoft-Azure-Devices-Device-Client-Enter] (ErrorDelegatingHandler#65204782, ExecuteWithErrorHandlingAsync, ErrorDelegatingHandler.ExecuteWithErrorHandlingAsync).
2021-10-22T16:22:01.6275157>> Trace - [Microsoft-Azure-Devices-Device-Client-Enter] (AmqpTransportHandler#42750725, DisableReceiveMessageAsync, (CancellationToken#3841804)).
[...]

PR #2448, opened 2022-06-16
fix(iot-dev): Fix bug where unsubscribing from twin/methods/c2d caused brief disconnection in AMQP

timtay-microsoft:
Each amqp link has an event handler that closed the owning session whenever the link gets closed. This happened even when just the link was deliberately closed by the client (unsubscribing to twin/method/c2d) so the client would suffer a connection drop and then quickly recover.

#2218

### DIFF
--- a/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
+++ b/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
@@ -1029,7 +1029,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
 
             // This will make the client unsubscribe from the mqtt c2d topic/close the amqp c2d link. Neither event
             // should close the connection as a whole, though.
-            await deviceClient.SetReceiveMessageHandlerAsync(null, null);
+            await deviceClient.SetReceiveMessageHandlerAsync(null, null).ConfigureAwait(false);
 
             await Task.Delay(1000);
 
--- a/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
+++ b/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
@@ -436,6 +436,30 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
             await DoNotReceiveMessagesSentBeforeSubscriptionAsync(TestDeviceType.X509, settings).ConfigureAwait(false);
         }
 
+        [LoggedTestMethod]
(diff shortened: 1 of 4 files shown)


e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
```diff
@@ -980,5 +1004,38 @@ private async Task DoNotReceiveMessagesSentBeforeSubscriptionAsync(TestDeviceTyp
...
+            using var testDeviceCallbackHandler = new TestDeviceCallbackHandler(deviceClient, testDevice, Logger);
+
+            // Subscribe to receive C2D messages over the callback.
+            await testDeviceCallbackHandler.SetMessageReceiveCallbackHandlerAsync().ConfigureAwait(false);
+
+            // This will make the client unsubscribe from the mqtt c2d topic/close the amqp c2d link. Neither event
+            // should close the connection as a whole, though.
+            await deviceClient.SetReceiveMessageHandlerAsync(null, null);
+
+            await Task.Delay(1000);
```
timtay-microsoft, line 1034: The disconnection that would happen prior to this fix would not necessarily cause a connection status callback event within the ```deviceClient.SetReceiveMessageHandlerAsync(null, null)``` call, so this task.delay accounts for that.

drwill-ms: ConfigureAwait(false)

iothub/device/src/Transport/AmqpIot/AmqpIotSendingLink.cs
```diff
@@ -29,6 +29,15 @@ public AmqpIotSendingLink(SendingAmqpLink sendingAmqpLink)
             _sendingAmqpLink.Closed += SendingAmqpLinkClosed;
         }
 
+        /// <summary>
+        /// Removes all events that would have run in the Closed event. This value cannot be set to null
+        /// outside of this class due to how .NET events work.
+        /// </summary>
+        public void ClearClosedEventHandler()
+        {
+            Closed = null;
```
drwill-ms, line 38: I'm concerned about a possible memory leak here. Let's chat over Teams when we're both free.

drwill-ms:
```suggestion
            Closed = null;
            _sendingAmqpLink.Closed -= SendingAmqpLinkClosed
```

drwill-ms (approved):

PR #2453, opened 2022-06-21
Renamed shared package to Authentication

drwill-ms:
Now that we've whittled down what was left of the Shared project to just SecurityProviders, we want to rename the project/package, similar to what we did with Java v2.

However, having talked with Tim, we think Authentication is a better term than Security, so we're using this new term for the project/package and classes.


azureiot.sln
```diff
@@ -7,7 +7,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "iothub", "iothub", "{537FA8
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "provisioning", "provisioning", "{3839DB7F-C289-4C1A-9D70-11DF35F139E0}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "shared", "shared", "{3AA089A9-A035-439E-BAF6-C3975A334379}"
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "authentication", "authentication", "{3AA089A9-A035-439E-BAF6-C3975A334379}"
```
drwill-ms, line 10: Moved to a new root folder.

common/src/HttpMessageHelper.cs
```diff
@@ -5,7 +5,7 @@
 using System.Text;
 using Newtonsoft.Json;
 
-namespace Microsoft.Azure.Devices.Shared
```
drwill-ms, line 8: Getting rid of the last classes with Shared in the namespace. As they are internal, they can just live in the "root" of `Microsoft.Azure.Devices`, IMHO.

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -43,7 +43,7 @@
 
   <ItemGroup>
     <None Include="$(RootDir)\LICENSE" Pack="true" PackagePath="" />
-    <None Include="$(RootDir)\shared\icons\nugetIcon.png" Pack="true" PackagePath="" />
+    <None Include="$(RootDir)\icons\nugetIcon.png" Pack="true" PackagePath="" />
```
drwill-ms, line 46: moved icons to the root

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -1157,32 +1159,32 @@ public static ProvisioningTransportHandler CreateTransportHandlerFromName(Client
                     throw new NotSupportedException("Unrecognized attestation type");
             }
 
-            throw new NotSupportedException($"Unknown security type: '{attestationType}'.");
+            throw new NotSupportedException($"Unknown attestation type: '{attestationType}'.");
         }
 
-        private Client.IAuthenticationMethod CreateAuthenticationMethodFromSecurityProvider(
-            SecurityProvider provisioningSecurity,
+        private Client.IAuthenticationMethod CreateAuthenticationMethodFromAuthhProvider(
```
brycewang-microsoft, line 1165:
```suggestion
        private Client.IAuthenticationMethod CreateAuthenticationMethodFromAuthProvider(
```

brycewang-microsoft (approved):

authentication/src/AuthenticationProviderSymmetricKey.cs
```diff
@@ -1,24 +1,24 @@
...
-    public class SecurityProviderSymmetricKey : SecurityProvider
+    public class AuthenticationProviderSymmetricKey : AuthenticationProvider
     {
         private readonly string _registrationId;
         private readonly string _primaryKey;
         private readonly string _secondaryKey;
 
         /// <summary>
-        /// Initializes a new instance of the SecurityProviderSymmetricKey class.
+        /// Initializes a new instance of the this class.
```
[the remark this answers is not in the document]
drwill-ms: Good eye, Bryce!

PR #2463, opened 2022-06-25
refactor(iot-serv): Remove non-references codes

brycewang-microsoft:
Remove internal/private methods without any references in the SDK. No public-facing API changes.


common/src/service/CommonConstants.cs
```diff
@@ -131,93 +131,5 @@ internal static class CommonConstants
         public const string HttpErrorCodeName = "iothub-errorcode";
 
         public static readonly string[] IotHubAadTokenScopes = new string[] { "https://iothubs.azure.net/.default" };
-
-        //Service Analytics related
-        public static class ServiceAnalytics
```
drwill-ms, line 136: My guess is that these constants are things customers might reference and use when calling our other APIs, which would explain why they don't have any references.

Are we sure we want to remove them? @abhipsaMisra or @timtay-microsoft, any insight here?

common/src/service/QuerySpecification.cs
```diff
@@ -12,11 +12,5 @@ internal class QuerySpecification
     {
         [JsonProperty(PropertyName = "query", Required = Required.Always)]
         public string Sql { get; set; }
-
-        [JsonIgnore]
-        public int? PageSize { get; set; }
```
drwill-ms, line 17: I suspect this is another case of things a customer may reference, and we just don't have any tests for it. Do we know this is for sure dead code?

iothub/service/src/Common/Exceptions/DeviceMaximumQueueDepthExceededException.cs
```diff
@@ -59,12 +59,6 @@ internal DeviceMaximumQueueDepthExceededException()
         {
         }
 
-        private DeviceMaximumQueueDepthExceededException(SerializationInfo info, StreamingContext context)
```
drwill-ms, line 62: We're going to redo exceptions as another step, and remove all of these.

If we kept these, I take it these ctors are used for de/serialization, which would explain why they don't have references. I'd suggest leaving them be for now.

iothub/service/src/DigitalTwin/Generated/DigitalTwinExtensions.cs
```diff
@@ -1,135 +0,0 @@
...
-{
-    using System.Collections;
-    using System.Collections.Generic;
-    using System.Threading;
-    using System.Threading.Tasks;
-
-    /// <summary>
-    /// Extension methods for DigitalTwin.
-    /// </summary>
-    internal static partial class DigitalTwinExtensions
```
drwill-ms, line 17: Why delete this file?

common/src/service/CommonConstants.cs
drwill-ms: Never mind; class is internal.

common/src/service/QuerySpecification.cs
drwill-ms: Never mind; another internal class.

PR #2465, opened 2022-06-27
Drop support for net472 in SDKs

drwill-ms:
net472 can use netstandard2.0, so we don't need an explicit target for it, but we should still test using it. Also, bump versions.


iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -2,8 +2,7 @@
   <PropertyGroup>
     <!--netcoreapp2.1 target framework here is solely so that this library can use the SocketsHttpHandler class to handle connection lease timeout issues.-->
     <!--See #1874 for additional details-->
-    <TargetFrameworks>net5.0;netstandard2.1;netstandard2.0;net472;netcoreapp2.1</TargetFrameworks>
```
drwill-ms, line 5: Dunno why we had netcoreapp2.1 either.

provisioning/service/src/Microsoft.Azure.Devices.Provisioning.Service.csproj
```diff
@@ -21,7 +21,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.18.1</Version>
+    <Version>2.0.0-preview</Version>
```
timtay-microsoft, line 24:
```suggestion
    <Version>2.0.0-preview001</Version>
```

To follow the pattern elsewhere?

PR #2468, opened 2022-06-28
refactor(iot-device): Update Amqp unit to update session references

abhipsaMisra:


iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -123,30 +131,27 @@ internal async Task<AmqpIotSession> EnsureSessionIsOpenAsync(CancellationToken c
...
-                    || _amqpIotSession.IsClosing())
+                if (_amqpIotSession == null || _amqpIotSession.IsClosing())
                 {
                     _amqpIotSession?.SafeClose();
 
-                    _amqpIotSession = await _amqpConnectionHolder
-                        .OpenSessionAsync(_deviceIdentity, cancellationToken)
-                        .ConfigureAwait(false);
+                    _amqpIotSession = await _amqpConnectionHolder.OpenSessionAsync(_deviceIdentity, cancellationToken).ConfigureAwait(false);
+
```
abhipsaMisra, line 139: I've updated these lines to break only if they have >=3 args being passed into the method call. Otherwise, it seemed to me like we were breaking too soon and that was decreasing the readability.

Let me know if you feel the previous indentation was more readable. I don't feel too strongly about it; this just made it more readable for me.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -245,7 +251,7 @@ private async Task EnsureMessageReceivingLinkIsOpenAsync(CancellationToken cance
             if (Logging.IsEnabled)
                 Logging.Enter(this, nameof(EnsureMessageReceivingLinkIsOpenAsync));
 
-            AmqpIotSession amqpIotSession = await EnsureSessionIsOpenAsync(cancellationToken).ConfigureAwait(false);
+            _amqpIotSession = await EnsureSessionIsOpenAsync(cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 254: Instead of saving the reference on a local variable, use the class variable to perform all opening and closing operations. The session object is opened under a lock and the transport layer disconnection and retry is also handled under locks.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.Azure.Devices.Shared;
 using Microsoft.Azure.Devices.Client.Transport.Amqp;
 using Microsoft.Azure.Devices.Client.Exceptions;
+using System.Linq;
```
drwill-ms, line 13: sort

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -100,6 +101,13 @@ public async Task OpenAsync(CancellationToken cancellationToken)
             }
         }
 
+        /// <summary>
+        /// Under a semaphore, fetch the reference to an amqp session that is open and active, and has a reference to an opened telemetry sending link.
```
drwill-ms, line 105:
```suggestion
        /// Under a semaphore, fetch the reference to an AMQP session that is open and active, and has a reference to an opened telemetry sending link.
```
```suggestion
        /// Under a semaphore, fetch the reference to an amqp session that is open and active, and has a reference to an opened telemetry sending link.
```

iothub/device/src/Transport/Amqp/AmqpUnit.cs
drwill-ms: To me it is mostly about line length, but also when exceeding 3 args.

drwill-ms: For this case, I can see the full line in diff split view on my work monitor, so it doesn't require line splitting for the first reason, either.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -123,30 +131,27 @@ internal async Task<AmqpIotSession> EnsureSessionIsOpenAsync(CancellationToken c
 
             try
             {
-                if (_amqpIotSession == null
-                    || _amqpIotSession.IsClosing())
+                if (_amqpIotSession == null || _amqpIotSession.IsClosing())
                 {
                     _amqpIotSession?.SafeClose();
```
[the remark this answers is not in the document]
abhipsaMisra: Discussed offline - adding locks or checks for this isn't possible because the AMQP library doesn't provide any callbacks for notifying us of the state. Instead, we have error handling logic when we open links or try to perform operations on opened links. If the operation throws an exception, the error handling code will determine if it is to be tried, and it will retry, if necessary.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -525,25 +547,31 @@ public async Task EnableEventReceiveAsync(CancellationToken cancellationToken)
...
                     {
                         _eventReceiverLinkDisconnected = (obj, arg) =>
                         {
-                            amqpIotSession.SafeClose();
+                            _amqpIotSession.SafeClose();
                         };
                     }
 
-                    _messageReceivingLink.Closed += _eventReceiverLinkDisconnected;
+                    _eventReceivingLink.Closed += _eventReceiverLinkDisconnected;
```
abhipsaMisra, line 572: This was probably a merge edit at some point. The horton pipeline caught this bug. A win for horton!

drwill-ms (approved):

PR #2469, opened 2022-06-28
Merge DPS transports into DPS device client

drwill-ms:
The hub client has all transports in it. So, there's no space to be gained by having DPS broken out. Instead, we can simplify usage of the DPS client and not require someone to add yet another NuGet package. It also cuts down our NuGet package count by 3.


provisioning/device/src/Microsoft.Azure.Devices.Provisioning.Client.csproj
```diff
@@ -41,20 +41,31 @@
...
       <Link>Common\Logging.Common.cs</Link>
     </Compile>
+    <Compile Include="$(Common)\Logging.ProvisioningTransport.Common.cs">
+      <Link>Common\Logging.ProvisioningTransport.Common.cs</Link>
+    </Compile>
+    <Compile Include="$(Common)\DefaultWebProxySettings.cs" />
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Amqp" Version="2.5.10" />
```
drwill-ms, line 59: The DPS device client now has NuGet dependencies for all the transports.

provisioning/device/src/Models/DeviceRegistrationHttp.cs
```diff
@@ -1,35 +1,33 @@
...
         /// Initializes a new instance of the DeviceRegistration class.
         /// </summary>
-        public DeviceRegistration(string registrationId = default(string), TpmAttestation tpm = default(TpmAttestation), JRaw payload = default(JRaw))
-            : this(payload)
+        public DeviceRegistrationHttp(JRaw payload = default, string registrationId = default, TpmAttestation tpm = default)
+            : base(payload)
         {
             RegistrationId = registrationId;
             Tpm = tpm;
-            CustomInit();
```
drwill-ms, line 22: Some fancy stuff used to be going on to account for partial classes across projects/packages. We don't need to do that anymore. I don't see how `CustomInit()` was even being used, so I removed that too.

Now we rely on class inheritance for these transport-specific classes.

common/src/DefaultWebProxySettings.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.Devices
     /// </summary>
     internal sealed class DefaultWebProxySettings : IWebProxy
     {
-        private static readonly DefaultWebProxySettings s_defaultWebproxy = new DefaultWebProxySettings();
```
timtay-microsoft, line 14: Does this actually work? That's pretty interesting

provisioning/device/src/Models/DeviceRegistrationResult.cs
```diff
@@ -0,0 +1,182 @@
...
+            GenerationId = generationId;
+            LastUpdatedDateTimeUtc = lastUpdatedDateTimeUtc;
+            ErrorCode = errorCode;
+            ErrorMessage = errorMessage;
+            Etag = etag;
+            Payload = jsonPayload;
+        }
+
+        /// <summary>
+        ///. Creates an instance of this class.
```
timtay-microsoft, line 55:
```suggestion
        /// Creates an instance of this class.
```
```suggestion
        ///. Creates an instance of this class.
```
```suggestion
        ///. Creates an instance of this class.
```

provisioning/device/src/Models/SymmetricKeyRegistrationResult.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Provisioning.Client
+{
+    /// <summary>
+    /// Symmetric Key registration result.
+    /// </summary>
+    internal class SymmetricKeyRegistrationResult
+    {
+        // this class has no implementation. does the service not send a payload?
```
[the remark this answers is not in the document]
timtay-microsoft: This may be a bug. I would expect a payload here. A payload should (optionally) be present regardless of attestation type

provisioning/device/src/Models/DeviceRegistrationResult.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// </summary>
+        internal DeviceRegistrationResult()
+        {
+
+        }
+
+        /// <summary>
+        ///. Creates an instance of this class.
+        /// </summary>
+        public DeviceRegistrationResult(
```
abhipsaMisra, line 26: iirc only one of these constructors is actually being used by the SDK but we couldn't remove the unused ones because they were public. We can remove the unused ones and change the visibility to internal.

common/src/DefaultWebProxySettings.cs
drwill-ms: C# 9.0!

I'm most excited about it for class members that are dictionary<string, object> because that gets so long.

https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#fit-and-finish-features

provisioning/device/src/ProvisioningTransportRegisterMessage.cs
```diff
@@ -3,43 +3,14 @@
 
 using Microsoft.Azure.Devices.Authentication;
 
-namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
+namespace Microsoft.Azure.Devices.Provisioning.Client
 {
     /// <summary>
     /// Represents a provisioning registration message.
     /// </summary>
     public class ProvisioningTransportRegisterMessage
```
timtay-microsoft, line 11: I'm not sure why the name of the file and this class were out of sync, but I actually like ```ProvisioningTransportRegisterRequest``` more than ```ProvisioningTransportRegisterMessage```. This message is a registration request after all

provisioning/device/src/Models/SymmetricKeyRegistrationResult.cs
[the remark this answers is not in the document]
drwill-ms: Would you mind checking Java to see if we have anything there?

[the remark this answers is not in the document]
abhipsaMisra: Looks like it should have the enrollment group information: https://github.com/Azure/CertsForIoT-B/blob/a8a11dc23ae577e3cf9f993aeb5257779d9ab0b8/swagger/device.json#L695-L703

provisioning/device/src/Models/DeviceRegistrationResult.cs
```diff
@@ -0,0 +1,182 @@
...
+            Tpm = tpm;
+            X509 = x509;
+            SymmetricKey = symmetricKey;
+        }
+
+        /// <summary>
+        /// The registration id.
+        /// </summary>
+        [JsonProperty(PropertyName = "registrationId")]
+        public string RegistrationId { get; internal set; }
```
abhipsaMisra, line 96: we probably don't need these setters anymore either

timtay-microsoft (approved):

provisioning/device/src/Models/DeviceRegistrationResult.cs
drwill-ms: We need them for deserialization, right?

abhipsaMisra: ah, you're right

provisioning/device/src/Models/SymmetricKeyRegistrationResult.cs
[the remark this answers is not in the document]
timtay-microsoft: The payload I was expecting lives in the [DeviceRegistrationResult](https://github.com/Azure/CertsForIoT-B/blob/a8a11dc23ae577e3cf9f993aeb5257779d9ab0b8/swagger/device.json#L553) object defined in the swagger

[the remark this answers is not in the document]
timtay-microsoft: So this class is fine without a payload. As Abhipsa pointed out, this class should just have a string: "enrollmentGroupId"

abhipsaMisra (approved):
Looks good.

provisioning/device/src/Models/SymmetricKeyRegistrationResult.cs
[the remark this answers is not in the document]
drwill-ms: I'll post another PR for adding that.

PR #2473, opened 2022-06-29
Cleanup remaining netcoreapp targets

drwill-ms:
I need to talk to Tim about this as he added this support originally, I'd guess before we added net5.0 support.

That said, I don't think libraries should target netcoreapp. Maybe we have a good reason for an exception here, but I'll post this PR to start the discussion.


abhipsaMisra (approved):
This not being a part of .NET "standard" until .NET 5.0, I'd say it is fair for us match the support with .NET itself. Us being libraries, we cannot practically extent support for specific .NET implementations.

PR #2474, opened 2022-06-29
Continue removing non-references codes

brycewang-microsoft:
Removing non-references codes for the folders `provisioning` and `security` in particular. No public-facing API changes.


provisioning/service/src/Enrollments/AttestationMechanism.cs
```diff
@@ -48,42 +48,6 @@ internal AttestationMechanism(Attestation attestation)
             }
         }
 
-        [JsonConstructor]
```
drwill-ms, line 51: How do we know this one is safe to remove? It is a constructor for JSON so it wouldn't have references.

provisioning/transport/http/src/Generated/RuntimeRegistrationExtensions.cs
```diff
@@ -12,29 +12,6 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
     /// </summary>
     internal static partial class RuntimeRegistrationExtensions
     {
-        /// <summary>
```
drwill-ms, line 15: Should we avoid editing generated files?

provisioning/transport/http/src/SymmetricKeyCredentials.cs
```diff
@@ -12,30 +12,11 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
...
-        private const string Registration = "registration";
         private readonly string SymmetricKey;
-        private volatile string _sasToken;
 
         public SymmetricKeyCredentials(string symmetricKey) : base()
         {
             SymmetricKey = symmetricKey;
         }
-
-        public override Task ProcessHttpRequestAsync(HttpRequestMessage request, CancellationToken cancellationToken)
```
drwill-ms, line 25: I'm not sure what the bar is for removing code in this PR. Is it simply that a member doesn't have any references? I'm worried we're going too far in some cases.

provisioning/service/src/Enrollments/AttestationMechanism.cs
brycewang-microsoft: Yeah, the failed e2e tests complained about the same. I will undo the `JsonConstructor` removing.

timtay-microsoft (approved):

provisioning/transport/http/src/Generated/RuntimeRegistrationExtensions.cs
drwill-ms: Yeah.

drwill-ms (approved):

PR #2477, opened 2022-06-30
Add DPS response for symmetric key registration

drwill-ms:


provisioning/device/src/Models/SymmetricKeyRegistrationResult.cs
```diff
@@ -1,13 +1,19 @@
...
+    /// Registration result returned when using symmetric key attestation.
     /// </summary>
     internal class SymmetricKeyRegistrationResult
     {
-        // this class has no implementation. does the service not send a payload?
+        /// <summary>
+        /// The Id of the enrollment group.
+        /// </summary>
+        [JsonProperty(PropertyName = "enrollmentGroupId")]
+        public string EnrollmentGroupId { get; internal set; }
```
drwill-ms, line 17: Okay, added, but we don't do anything with it. Does it need to be hooked up somewhere?

drwill-ms: FWIW, neither is `X509RegistrationResult.EnrollmentGroupId`. We deserialize it but don't do anything with it.

Should we?

abhipsaMisra: Looks like we haven't been exposing all fields in the public facing `DeviceRegistrationResult` (eg. `TpmRegistrationResult`, `X509RegistrationResult` etc.). Adding those in the public facing registration result class should fix it.

drwill-ms: Should we make them public get then?

abhipsaMisra (approved):

PR #2483, opened 2022-07-06
Rename C2D methods, misc code cleanup

drwill-ms:
Incremental changes on the way to making DeviceClient and ModuleClient use ctors and be unit testable.

Upcoming changes include removing overrides that don't take cancellation tokens, documenting exceptions, rearranging methods...

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -3,8 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
-using System.IO;
 using System.Linq;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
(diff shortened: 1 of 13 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -284,14 +207,14 @@ public async Task OpenAsync()
...
+        /// Explicitly open the client instance.
         /// </summary>
         /// <param name="cancellationToken">A token to cancel the operation.</param>
-        public Task OpenAsync(CancellationToken cancellationToken)
+        public async Task OpenAsync(CancellationToken cancellationToken)
         {
             try
             {
-                return InnerHandler.OpenAsync(cancellationToken);
+                await InnerHandler.OpenAsync(cancellationToken).ConfigureAwait(false);
```
drwill-ms, line 217: Fixing this bug where the try/catch didn't actually work once the callee went async.

iothub/device/src/InternalClient.cs
```diff
@@ -378,12 +301,12 @@ internal void OnConnectionStatusChanged(ConnectionStatus status, ConnectionStatu
         /// Deletes a received message from the device queue
         /// </summary>
         /// <returns>The lock identifier for the previously received message</returns>
-        public async Task CompleteAsync(string lockToken)
+        public async Task CompleteMessageAsync(string lockToken)
```
drwill-ms, line 304: Putting an object noun in these method names so the meaning is clearer.

iothub/device/src/MessageResponse.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Status of handling a message.
+    /// </summary>
+    public enum MessageResponse
```
drwill-ms, line 9: Moved out from InternalClient to avoid having multiple types in the same file.

andyk-ms (approved):

PR #2484, opened 2022-07-06
Separate out the payload collection types

abhipsaMisra:
This PR addresses the following:
- Separate out the TelemetryCollection/ClientPropertyCollection -> PayloadCollection inheritance. Having a bunch of TryGetValue and Contains method implementations on the PayloadCollection base class was hindering discoverability.
- Separate out the "reported" and "desired" client property collection types. Having them be the same types was causing issues with the receive property update notification -> respond with ack flow.
- Add additional helper methods to retrieve the writable property object.
- Add additional helper methods to retrieve the property value from the writable property object.

API surface doc: https://github.com/Azure/azure-iot-sdk-csharp/blob/abmisr/separatePropertyType/iothub/device/devdoc/Convention-based%20operations.md


iothub/device/src/Convention/ClientPropertyCollection.cs
```diff
@@ -1,38 +1,109 @@
...
+    /// </remarks>
+    public class ClientPropertyCollection : IEnumerable<KeyValuePair<string, object>>
     {
         private const string VersionName = "$version";
 
-        /// <inheritdoc path="/summary" cref="AddInternal(IDictionary{string, object}, string)" />
+        // TODO: Unit-testable and mockable
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="ClientPropertyCollection"/>.
```
drwill-ms, line 25:
```suggestion
        /// Initializes a new instance of this class.
```

iothub/device/tests/Transport/Amqp/MoqableAmqpTransportHandler.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.Azure.Devices.Shared;
 using Microsoft.Azure.Devices.Client.Tests.ConnectionString;
 using Moq;
+using System.Collections.Generic;
```
drwill-ms, line 11: sort

drwill-ms (approved):

PR #2488, opened 2022-07-08
Reorder methods in device clients

drwill-ms:


iothub/device/src/DeviceClient.cs
```diff
@@ -255,6 +255,67 @@ public string ProductInfo
             set => InternalClient.ProductInfo = value;
         }
 
+        /// <summary>
```
drwill-ms, line 258: General order:

- Subscriptions (connection status, property update, methods, C2D)
- RetryPolicy/Open/Close
- Telemetry
- Twins
- Methods
- C2D

drwill-ms: Also publics above internal/protected above private

brycewang-microsoft: > Also publics above internal/protected above private

Looks like we have some methods not following this (such as L793) in InternalClient class. Do we wanna update those as well?

PR #2493, opened 2022-07-13
Update test that validates c2d message can be received after it was sent

drwill-ms:


e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
```diff
@@ -536,47 +536,55 @@ await Task
...
             Logger.Trace($"Sending C2D message from service, messageId={msg.MessageId}");
             await serviceClient.SendAsync(testDevice.Id, msg).ConfigureAwait(false);
 
-            // Subscribe to receive C2D messages over the callback.
-            testDeviceCallbackHandler.ExpectedMessageSentByService = msg;
-            await testDeviceCallbackHandler.SetMessageReceiveCallbackHandlerAsync().ConfigureAwait(false);
+            using DeviceClient deviceClient2 = testDevice.CreateDeviceClient(transportType);
+            // Open the device client - for MQTT, this will connect the device with CleanSession flag set to false.
+            // Also, over MQTT it seems the device must be connected (although not necessarily subscribed for C2D messages)
+            // in order to C2D messages to get to the device. If they are offline, the messages will never be delivered.
```
brycewang-microsoft, line 562:
```suggestion
            // in order for C2D messages to get to the device. If they are offline, the messages will never be delivered.
```

e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
```diff
@@ -536,47 +536,55 @@ await Task
...
-
-            (Message msg, string payload, string p1Value) = ComposeC2dTestMessage(Logger);
-
-            // Subscribe to receive C2D messages over the callback.
-            await testDeviceCallbackHandler.SetMessageReceiveCallbackHandlerAsync().ConfigureAwait(false);
-
-            // Now dispose and reinitialize the client instance.
-            deviceClient.Dispose();
-            deviceClient = null;
+            using DeviceClient deviceClient1 = testDevice.CreateDeviceClient(transportType);
```
brycewang-microsoft, line 542: what is `deviceClient1` doing here? It appears that we just open and close it without any other actions.

drwill-ms: The comment below explains it. With MQTT, if a new device hasn't connected yet, then the service client can't send a C2D message to it. I've found this via trial and error.

PR #2496, opened 2022-07-14
Adjust some test errors to be clearer

drwill-ms:


e2e/test/iothub/service/RegistryManagerE2ETests.cs
```diff
@@ -107,10 +107,16 @@ public async Task RegistryManager_AddDeviceWithTwinWithDeviceCapabilities()
...
             await registryManager.AddDeviceWithTwinAsync(iotEdgeDevice, twin).ConfigureAwait(false);
 
-            Device actual = await registryManager.GetDeviceAsync(deviceId).ConfigureAwait(false);
-            await registryManager.RemoveDeviceAsync(deviceId).ConfigureAwait(false);
-
-            Assert.IsTrue(actual.Capabilities.IotEdge);
+            try
+            {
+                Device actual = await registryManager.GetDeviceAsync(deviceId).ConfigureAwait(false);
+                actual.Should().NotBeNull($"Got null in GET on device {deviceId} to check IotEdge property.");
```
drwill-ms, line 113: I think it is null because of a lag issue with the device registry. I'm going to aim to solve that independently, but in the meantime, I'd like not to see NREs in test results.

e2e/test/helpers/TestDevice.cs
```diff
@@ -112,6 +112,18 @@ await RetryOperationHelper
                     s_logger)
                 .ConfigureAwait(false);
 
+            // Confirm the device exists in the registry before calling it good to avoid downstream test failures.
+            await RetryOperationHelper
```
drwill-ms, line 116: Prevent cases where we create a device but then get it and we get null back from the service.

e2e/test/iothub/AuthenticationWithTokenRefreshDisposalTests.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Azure.Devices.E2ETests
     public class AuthenticationWithTokenRefreshDisposalTests : E2EMsTestBase
     {
         public static readonly TimeSpan MaxWaitTime = TimeSpan.FromSeconds(10);
-        private readonly string _devicePrefix = $"E2E_{nameof(AuthenticationWithTokenRefreshDisposalTests)}_";
+        private readonly string _devicePrefix = $"{nameof(AuthenticationWithTokenRefreshDisposalTests)}_";
```
drwill-ms, line 23: The helper adds the E2E prefix. This is why we have so many test devices prefixed "E2E_E2E_".

e2e/test/iothub/messaging/FaultInjectionPoolAmqpTests.MessageReceiveFaultInjectionPoolAmqpTests.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.Devices.E2ETests
 {
     public partial class FaultInjectionPoolAmqpTests
     {
-        private readonly string MessageReceive_DevicePrefix = $"MessageReceiveFaultInjectionPoolAmqpTests";
+        private readonly string MessageReceive_DevicePrefix = "MessageReceiveFaultInjectionPoolAmqpTests";
```
abhipsaMisra, line 19:
```suggestion
        private readonly string MessageReceive_DevicePrefix = $"{nameof(MessageReceiveFaultInjectionPoolAmqpTests)}_";
```

for consistency with the other tests.

abhipsaMisra (approved):

e2e/test/iothub/messaging/FaultInjectionPoolAmqpTests.MessageReceiveFaultInjectionPoolAmqpTests.cs
drwill-ms: That isn't a name of a type. :(

PR #2497, opened 2022-07-14
refactor(iot-device): Add some more doc comments to public APIs

abhipsaMisra:


iothub/device/src/Convention/CommandRequest.cs
```diff
@@ -42,17 +43,32 @@ internal CommandRequest(PayloadConvention payloadConvention, string commandName,
...
-        /// The command request payload.
+        /// The command request payload, deserialized to the specified type.
         /// </summary>
         /// <typeparam name="T">The type to deserialize the command request payload to.</typeparam>
-        /// <returns>The command request payload.</returns>
-        public T GetPayload<T>()
+        /// <param name="payload">When this method returns true, this contains the value of the command request payload.
+        /// When this method returns false, this contains the default value of the type <c>T</c> passed in.</param>
+        /// <returns><c>true</c> if the command request payload can be deserialized to type <c>T</c>; otherwise, <c>false</c>.</returns>
+        public bool TryGetPayload<T>(out T payload)
```
abhipsaMisra, line 52: Updated this to a "TryGet" method, for consistency with the "TryGet" methods in properties classes.

iothub/device/src/Convention/WritableClientPropertyAcknowledgement.cs
```diff
@@ -16,10 +16,6 @@ public class WritableClientPropertyAcknowledgement
     {
         // TODO: Unit-testable and mockable
 
-        internal WritableClientPropertyAcknowledgement()
```
abhipsaMisra, line 19: This object can be created by the user, so the constructor needs to be public.

iothub/device/src/Convention/CommandRequest.cs
```diff
@@ -42,17 +43,32 @@ internal CommandRequest(PayloadConvention payloadConvention, string commandName,
...
+                }
+
+                payload = _payloadConvention.PayloadSerializer.DeserializeToType<T>(jsonPayload);
+                return true;
+            }
+            catch (Exception)
+            {
+            }
+
+            return false;
```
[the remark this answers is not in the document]
drwill-ms: To avoid an empty catch statement that looks questionable.

iothub/device/src/DeviceClient.ConventionBasedOperations.cs
```diff
@@ -102,10 +106,14 @@ public Task<ClientProperties> GetClientPropertiesAsync(CancellationToken cancell
         /// <param name="callback">The callback to handle all writable property updates for the client.</param>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <remarks>
+        /// Calling this API more than once will result in the listerner set last overwriting any previously set listener.
```
drwill-ms, line 109:
```suggestion
        /// Calling this API more than once will result in the listener set last overwriting any previously set listener.
```

iothub/device/src/ModuleClient.ConventionBasedOperations.cs
```diff
@@ -37,6 +37,10 @@ public Task SendTelemetryAsync(TelemetryMessage telemetryMessage, CancellationTo
         /// <summary>
         /// Sets the listener for all command calls from the service.
         /// </summary>
+        /// <remarks>
+        /// Calling this API more than once will result in the listerner set last overwriting any previously set listener.
```
drwill-ms, line 41:
```suggestion
        /// Calling this API more than once will result in the listener set last overwriting any previously set listener.
```

iothub/device/src/ModuleClient.ConventionBasedOperations.cs
```diff
@@ -96,10 +100,14 @@ public Task<ClientPropertiesUpdateResponse> UpdateClientPropertiesAsync(ClientPr
         /// <param name="callback">The callback to handle all writable property updates for the client.</param>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <remarks>
+        /// Calling this API more than once will result in the listerner set last overwriting any previously set listener.
```
drwill-ms, line 103:
```suggestion
        /// Calling this API more than once will result in the listener set last overwriting any previously set listener.
```

drwill-ms (approved):

iothub/device/src/Convention/CommandRequest.cs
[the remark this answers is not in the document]
drwill-ms: It isn't the default because in this case the code has acted as a gauntlet for catching all the bad cases (above returning false where appropriate), so a `return true` at the end is the point where it has made it.

But if you prefer to just add a comment to the catch block, that's fine with me.

PR #2499, opened 2022-07-14
Migrate samples from .NET SDK sample repo to source code repo

brycewang-microsoft:
This migrates all of our existing device, service and provisioning .NET samples from the separate repo to here, with all commits history together. We will re-organize the folder structure according to our discussion before and categorize samples into each use-case level (100, 200-300, 400+).


brycewang-microsoft:
It's a little bit annoying that GitHub automatically takes the PR numbers (#XXX) within the name of each commit linked with the actual PRs in this repo. However, we can view the code changes for each commit, by clicking the commit ID and then clicking "view the full commit page here".

image

PR #2504, opened 2022-07-19
Move TransportType to ClientOptions

drwill-ms:
In order to reduce the number of factory methods (soon to be ctors), we'll start by moving parameters to `ClientOptions`, starting with `TransportType`.

### DIFF
--- a/e2e/test/iothub/NoRetryE2ETests.cs
+++ b/e2e/test/iothub/NoRetryE2ETests.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.Devices.E2ETests
         public async Task FaultInjection_NoRetry_NoRecovery_OpenAsync()
         {
             using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix, TestDeviceType.Sasl).ConfigureAwait(false);
-            using DeviceClient deviceClient = testDevice.CreateDeviceClient(Client.TransportType.Amqp_Tcp_Only);
+            using DeviceClient deviceClient = testDevice.CreateDeviceClient(new ClientOptions { TransportType = Client.TransportType.Amqp_Tcp_Only });
 
             Logger.Trace($"{nameof(FaultInjection_NoRetry_NoRecovery_OpenAsync)}: deviceId={testDevice.Id}");
             deviceClient.SetRetryPolicy(new NoRetry());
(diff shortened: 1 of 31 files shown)


iothub/device/src/ClientSettings/ClientOptions.cs
```diff
@@ -10,10 +10,16 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public class ClientOptions
     {
+        /// <summary>
+        /// The transport type to use (i.e., AMQP, MQTT, HTTP), including whether to use TCP or web sockets where applicable.
+        /// </summary>
+        public TransportType TransportType { get; set; } = TransportType.Amqp;
```
timtay-microsoft, line 16: Do you want to change the default to MQTT now?

timtay-microsoft (approved):

iothub/device/src/ClientSettings/ClientOptions.cs
timtay-microsoft: Or maybe MQTT_WS since most users use websocket rather than raw MQTT?

e2e/test/iothub/messaging/MessageSendE2ETests.cs
```diff
@@ -344,7 +344,7 @@ public async Task Message_DeviceSendMessageWayOverAllowedSize_Http()
         private async Task SendSingleMessage(TestDeviceType type, Client.TransportType transport, int messageSize = 0)
         {
             using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, _devicePrefix, type).ConfigureAwait(false);
-            using DeviceClient deviceClient = testDevice.CreateDeviceClient(transport);
+            using DeviceClient deviceClient = testDevice.CreateDeviceClient();
```
abhipsaMisra, line 347: this, and below - missing transport information.

iothub/device/src/ClientSettings/ClientOptions.cs
abhipsaMisra, line 16: How are we accepting the transport settings? Is that also going to be a part of client options?

e2e/test/iothub/NoRetryE2ETests.cs
```diff
@@ -87,8 +87,8 @@ public async Task DuplicateDevice_NoRetry_NoPingpong_OpenAsync()
 
             Logger.Trace($"{nameof(DuplicateDevice_NoRetry_NoPingpong_OpenAsync)}: 2 device client instances with the same deviceId={testDevice.Id}.");
 
-            using DeviceClient deviceClient1 = testDevice.CreateDeviceClient(Client.TransportType.Amqp_Tcp_Only);
-            using DeviceClient deviceClient2 = testDevice.CreateDeviceClient(Client.TransportType.Amqp_Tcp_Only);
+            using DeviceClient deviceClient1 = testDevice.CreateDeviceClient();
```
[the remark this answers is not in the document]
abhipsaMisra: If this is omitted because amqp is the default transport type (right now), I think specifying the transport type in our E2E tests explicitly will help with readability. What do you think?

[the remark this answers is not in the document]
drwill-ms: I meant to replace all instances where it was specified with a client options. I just missed this one. Thanks for catching it.

iothub/device/src/ClientSettings/ClientOptions.cs
drwill-ms: Yes, but a next step in this series of PRs.

drwill-ms: Also, I plan to only accept a single one, not an array.

Thoughts?

abhipsaMisra: That sounds good to me. +1 to removing the fallback code.
We'd need a base class for the transport settings then, if we don't have one already.

PR #2506, opened 2022-07-20
Move gatewayHostname param to ClientOptions

drwill-ms:


iothub/device/src/ClientFactory.cs
```diff
@@ -19,7 +19,7 @@ internal class ClientFactory
 
         private static readonly TimeSpan s_regexTimeoutMilliseconds = TimeSpan.FromMilliseconds(500);
 
-        private static readonly Regex s_deviceIdParameterRegex = new Regex(
+        private static readonly Regex s_deviceIdParameterRegex = new(
```
timtay-microsoft, line 22: I know this is better, but this will never not look funny to me

timtay-microsoft (approved):

iothub/device/src/ClientFactory.cs
drwill-ms: Not much different than `var foo = new Foo();` but it solves the problem of long lines in class field definitions, which I really appreciate the lack of redundancy. :)

But I hear you on it looking a bit off. ;)

PR #2509, opened 2022-07-21
Device handlers cleanup

drwill-ms:
InnerHandler means what? It is the next handler in the chain, right? I've renamed it to make more sense.

Removed the timeout methods.

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -465,7 +465,7 @@ namespace Microsoft.Azure.Devices.Client
 
             try
             {
-                await InnerHandler.CompleteAsync(lockToken, cancellationToken).ConfigureAwait(false);
+                await InnerHandler.CompleteMessageAsync(lockToken, cancellationToken).ConfigureAwait(false);
             }
             catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
             {
@@ -509,7 +509,7 @@ namespace Microsoft.Azure.Devices.Client
             // expire or unrecoverable error(authentication, quota exceed) occurs.
             try
             {
-                await InnerHandler.AbandonAsync(lockToken, cancellationToken).ConfigureAwait(false);
+                await InnerHandler.AbandonMessageAsync(lockToken, cancellationToken).ConfigureAwait(false);
             }
             catch (IotHubCommunicationException ex) when (ex.InnerException is OperationCanceledException)
             {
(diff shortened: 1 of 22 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -465,7 +465,7 @@ public async Task CompleteMessageAsync(string lockToken, CancellationToken cance
 
             try
             {
-                await InnerHandler.CompleteAsync(lockToken, cancellationToken).ConfigureAwait(false);
+                await InnerHandler.CompleteMessageAsync(lockToken, cancellationToken).ConfigureAwait(false);
```
drwill-ms, line 468: I renamed this in the clients, but missed the handlers. Now C2D methods make more sense as being related to messaging.

iothub/device/src/Pipeline/DefaultDelegatingHandler.cs
```diff
@@ -10,59 +10,51 @@ namespace Microsoft.Azure.Devices.Client.Transport
 {
     internal abstract class DefaultDelegatingHandler : IDelegatingHandler
     {
-        private volatile IDelegatingHandler _innerHandler;
+        private volatile IDelegatingHandler _nextHandler;
```
andyk-ms, line 13: It is private anyway, but wonder if 'next' does communicate 'delegating'.
Why don't we call it 'delegate' directly (delegating to this)?

andyk-ms (approved):

abhipsaMisra (approved):

iothub/service/src/IotHubClientWebSocket.cs
```diff
@@ -241,7 +241,7 @@ public async Task<int> ReceiveAsync(byte[] buffer, int offset, TimeSpan timeout)
                         State = WebSocketState.Closed;
                         WebSocketStream.Close();
                         TcpClient.Close();
-                        return 0;  // TODO: throw exception?
+                        return 0; // TODO: throw exception?
```
abhipsaMisra, line 244: Another task for us to address

iothub/device/src/Pipeline/DefaultDelegatingHandler.cs
drwill-ms: What is the delegating part? I've never understood that either.

It is a chain of handlers for an operation.

iothub/service/src/IotHubClientWebSocket.cs
drwill-ms: Seems like it.

PR #2511, opened 2022-07-21
Use cancellation tokens instead of timespans in a few more places with AMQP

drwill-ms:


tmahmood-microsoft (approved):

andyk-ms (approved):

drwill-ms:
I have a problem with the AMQP library. I'm talking to Xin about it. I think we need another release to be able to finish my work. I'm putting it on hold while I chat with Xin.

PR #2512, opened 2022-07-22
Remove a cancellation token argument that would never be cancelled

timtay-microsoft:
MethodRequestInternal (instantiated in an MQTT/AMQP callback thread, not by a user thread) was given CancellationToken.None from MQTT and a new CancellationToken that went unreferenced from AMQP, so there is no point in propagating these tokens along.

### DIFF
--- a/iothub/device/src/Transport/AmqpIot/AmqpIotReceivingLink.cs
+++ b/iothub/device/src/Transport/AmqpIot/AmqpIotReceivingLink.cs
@@ -218,9 +218,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIot
 
             try
             {
-                MethodRequestInternal methodRequestInternal = AmqpIotMessageConverter.ConstructMethodRequestFromAmqpMessage(
-                    amqpMessage,
-                    new CancellationToken(false));
+                MethodRequestInternal methodRequestInternal = AmqpIotMessageConverter.ConstructMethodRequestFromAmqpMessage(amqpMessage);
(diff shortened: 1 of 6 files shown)


iothub/device/src/Transport/AmqpIot/AmqpIotReceivingLink.cs
```diff
@@ -218,9 +218,7 @@ private void OnMethodReceived(AmqpMessage amqpMessage)
 
             try
             {
-                MethodRequestInternal methodRequestInternal = AmqpIotMessageConverter.ConstructMethodRequestFromAmqpMessage(
-                    amqpMessage,
-                    new CancellationToken(false));
```
timtay-microsoft, line 223: This was the only spot we created the MethodRequestInternal from AMQP. Note that this token would never be cancelled

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -508,7 +508,7 @@ private async Task HandleIncomingMethodPostAsync(Message message)
             {
                 string[] tokens = Regex.Split(message.MqttTopicName, "/", RegexOptions.Compiled, s_regexTimeoutMilliseconds);
 
-                using var mr = new MethodRequestInternal(tokens[3], tokens[4].Substring(6), message.GetBodyStream(), CancellationToken.None);
```
timtay-microsoft, line 511: And this was the only spot we created the MethodRequestInternal from MQTT.

abhipsaMisra (approved):

andyk-ms (approved):

PR #2513, opened 2022-07-26
Fix a portion of the test build errors

drwill-ms:


tmahmood-microsoft (approved):

iothub/device/src/ClientSettings/AmqpTransportSettings.cs
```diff
@@ -104,13 +98,7 @@ public TimeSpan OperationTimeout
         /// <summary>
         /// The pre-fetch count
         /// </summary>
-        public uint PrefetchCount
```
drwill-ms, line 107: Turns out it can't be negative because the type is unsigned! :)

PR #2518, opened 2022-07-26
refactor(iot-device): Update writable property update subscription API

abhipsaMisra:


iothub/device/src/InternalClient.cs
```diff
@@ -1167,20 +1167,48 @@ public async Task UpdateReportedPropertiesAsync(TwinCollection reportedPropertie
...
                 TwinCollection twinUpdate = JsonConvert.DeserializeObject<TwinCollection>(JsonConvert.SerializeObject(patch));
-                _desiredPropertyUpdateCallback(twinUpdate, _twinPatchCallbackContext);
+                await _desiredPropertyUpdateCallback(twinUpdate, _twinPatchCallbackContext);
             }
             else if (_writableClientPropertyUpdateCallback != null)
             {
                 var writableClientPropertyCollection = new WritableClientPropertyCollection(patch, PayloadConvention);
-                _writableClientPropertyUpdateCallback(writableClientPropertyCollection);
+
+                var propertiesToBeReported = new ClientPropertyCollection();
```
drwill-ms, line 1184: Why give it a default instance instead of `null`?

iothub/device/src/InternalClient.cs
```diff
@@ -1167,20 +1167,48 @@ public async Task UpdateReportedPropertiesAsync(TwinCollection reportedPropertie
...
+                    {
+                        if (Logging.IsEnabled)
+                            Logging.Error(this, $"Received an exception when updating writable client property acknowledgement: {ex}", nameof(OnDesiredStatePatchReceived));
+                    }
+                }
+                else
+                {
+                    if (Logging.IsEnabled)
+                        Logging.Info(this, $"Did not receive any properties to update", nameof(OnDesiredStatePatchReceived));
+                }
```
drwill-ms, line 1211:
```suggestion
                else if (Logging.IsEnabled)
                    Logging.Info(this, $"Did not receive any properties to update", nameof(OnDesiredStatePatchReceived));
```

drwill-ms (approved):

iothub/device/src/InternalClient.cs
drwill-ms, line 1211: Can this block just be at line 1200?

drwill-ms: Avoiding nested try/catch?

iothub/device/src/InternalClient.cs
abhipsaMisra: No reason. This was minimal memory usage, so I didn't necessarily optimize. null should also work, sure.

e2e/test/iothub/properties/PropertiesE2ETests.cs
```diff
@@ -307,7 +307,20 @@ await Task
                 .ConfigureAwait(false);
 
             // Validate the updated properties from the device-client
-            ClientProperties clientProperties = await deviceClient.GetClientPropertiesAsync().ConfigureAwait(false);
+
+            // Since this will depend on the SDK updating the properties asynchronously under-the-cover,
+            // we will need to fetch the properties until we receive an updated version number.
```
abhipsaMisra, line 312: The `UpdateClientPropertiesAysnc` call now happens asynchronously on a different thread, internally. As a result, when the `writablePropertyCallbackSemaphore` is released, we won't have updated the property yet. The customer will now need to loop over the `DeviceClient.GetClientPropertiesAsync()` call and validate against the `Version` to compare and check if that has been updated.

iothub/device/src/InternalClient.cs
abhipsaMisra: Good point, simplified.

abhipsaMisra: It could, yes

PR #2519, opened 2022-07-26
New release pipeline for partner signing

andyk-ms:
New release pipeline
- Build/test/pack release packages for partner signing pipeline


drwill-ms:
Why does the file name have an underscore for the 1st word boundary, and then dashes for the others?

drwill-ms (approved):

andyk-ms:
> Why does the file name have an underscore for the 1st word boundary, and then dashes for the others?

The first part is what it is for, and second part is what it does.
Saw prior name use csharp etc, switching to net.
we can change to different name - just a unique name to identify it.

PR #2520, opened 2022-07-26
Fix service API query string being added twice to device statistics r…

timtay-microsoft:
…equests

Also add a basic e2e test to ensure that these statistics requests can be made

### DIFF
--- a/e2e/test/iothub/service/RegistryE2ETests.cs
+++ b/e2e/test/iothub/service/RegistryE2ETests.cs
@@ -574,6 +574,24 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
             }
         }
 
+        [LoggedTestMethod]
+        public async Task DevicesClient_GetStatistics()
+        {
+            using var serviceClient = new IotHubServiceClient(TestConfiguration.IoTHub.ConnectionString);
+
+            // No great way to test the accuracy of these statistics, but making the request successfully should
+            // be enough to indicate that this API works as intended
+            ServiceStatistics serviceStatistics = await serviceClient.Devices.GetServiceStatisticsAsync().ConfigureAwait(false);
+            serviceStatistics.ConnectedDeviceCount.Should().BeGreaterOrEqualTo(0);
+
+            // No great way to test the accuracy of these statistics, but making the request successfully should
+            // be enough to indicate that this API works as intended
+            RegistryStatistics registryStatistics = await serviceClient.Devices.GetRegistryStatisticsAsync().ConfigureAwait(false);
+            registryStatistics.DisabledDeviceCount.Should().BeGreaterOrEqualTo(0);
+            registryStatistics.EnabledDeviceCount.Should().BeGreaterOrEqualTo(0);
+            registryStatistics.TotalDeviceCount.Should().BeGreaterOrEqualTo(0);
+        }
+
         private static async Task CleanupAsync(IotHubServiceClient serviceClient, string deviceId)
         {
             if (deviceId == null)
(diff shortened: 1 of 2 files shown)


iothub/service/src/Registry2/DevicesClient.cs
```diff
@@ -31,7 +31,7 @@ public class DevicesClient
         private const string JobsListUriFormat = "/jobs";
         private const string JobsCreateUriFormat = "/jobs/create";
         private const string DeviceStatisticsUriFormat = "/statistics/devices";
-        private const string ServiceStatisticsUriFormat = "/statistics/service?" + ClientApiVersionHelper.ApiVersionQueryString;
```
timtay-microsoft, line 34: This service API string is added to every HTTP request by the HTTP client, so there is no need to add it to any request URI format variable

e2e/test/iothub/service/RegistryE2ETests.cs
```diff
@@ -574,6 +574,24 @@ public async Task RegistryManager_DeviceTwinLifecycle()
...
+
+            // No great way to test the accuracy of these statistics, but making the request successfully should
+            // be enough to indicate that this API works as intended
+            ServiceStatistics serviceStatistics = await serviceClient.Devices.GetServiceStatisticsAsync().ConfigureAwait(false);
+            serviceStatistics.ConnectedDeviceCount.Should().BeGreaterOrEqualTo(0);
+
+            // No great way to test the accuracy of these statistics, but making the request successfully should
+            // be enough to indicate that this API works as intended
+            RegistryStatistics registryStatistics = await serviceClient.Devices.GetRegistryStatisticsAsync().ConfigureAwait(false);
+            registryStatistics.DisabledDeviceCount.Should().BeGreaterOrEqualTo(0);
```
drwill-ms, line 590: I worry that things like this will cause the test to fail, depending on the condition of the hub.

If the test doesn't guarantee devices exist in this state, I think we should be wary of asserting on them. I'd say, consider adding these devices then cleaning up afterwards, or simply assert the requests returns without incident (exception) and that the properties exist.

timtay-microsoft: > or simply assert the requests returns without incident (exception) and that the properties exist.

That's what I'm doing here. The only real assertion in this test is that the values are greater than or equal to 0. There is no condition where a hub has negative disabled devices

e2e/test/iothub/service/RegistryE2ETests.cs
```diff
@@ -574,6 +574,24 @@ public async Task RegistryManager_DeviceTwinLifecycle()
...
         }
 
+        [LoggedTestMethod]
+        public async Task DevicesClient_GetStatistics()
+        {
+            using var serviceClient = new IotHubServiceClient(TestConfiguration.IoTHub.ConnectionString);
+
+            // No great way to test the accuracy of these statistics, but making the request successfully should
+            // be enough to indicate that this API works as intended
+            ServiceStatistics serviceStatistics = await serviceClient.Devices.GetServiceStatisticsAsync().ConfigureAwait(false);
```
abhipsaMisra, line 584: Unrelated to the changes in this PR, but what is `Get ServiceStatistics` supposed to return? Looking at the value below (.ConnectedDeviecCount), it looks like it returns device statistics that are stored at service end? Is that accurate? If yes, then should we call it "Devices.ServiceStatistics" or "Devices.Statistics"? The term "ServiceStatistics" made me think that I should expect some "service" related statistics; i.e. hub units, scale, etc.

abhipsaMisra (approved):

e2e/test/iothub/service/RegistryE2ETests.cs
drwill-ms: Okay, fair enough.

drwill-ms (approved):

e2e/test/iothub/service/RegistryE2ETests.cs
timtay-microsoft: Good point. This may be worth renaming. I'll follow up on this in a later PR

PR #2523, opened 2022-07-27
Rename device clients and device client options

drwill-ms:


iothub/device/src/IotHubClientOptions.cs
```diff
@@ -6,14 +6,14 @@
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// Options that allow configuration of the device or module client instance during initialization.
+    /// Options that allow configuration of the IoT hub device or module client instance during initialization.
     /// </summary>
-    public class ClientOptions
+    public class IotHubClientOptions
```
drwill-ms, line 11: Named `IotHubClientOptions`, not IotHub**Device**ClientOptions because it is used both by device and module clients.

One option is to name it `IotHubDeviceClientOptions` and then subclass it just for naming purposes as IotHubModuleClientOptions. Thoughts?

timtay-microsoft: There shouldn't be any options that are exclusive to devices vs modules, so I'd prefer to just have one options class ```IotHubClientOptions```

timtay-microsoft (approved):

brycewang-microsoft (approved):

abhipsaMisra (approved):

iothub/device/src/IotHubClientOptions.cs
brycewang-microsoft: We have `IotHubServiceClientOptions` for the service client, so I am worried just `IotHubClientOptions` sounds like the superclass of it. As keeping a single client option for both device and module client is ideal, maybe we can have a new name for it? With that said, sorry I cannot come up with any better name for now.

drwill-ms: They will be in different namespaces and the doc comments should be clear it is for device clients.

FWIW, I don't mind calling it `IotHubDeviceClientOptions` anyway, and leave the ambiguity with the IotHubModuleClient which takes one as a parameter.

PR #2524, opened 2022-07-27
Rename iot hub device client transport settings

drwill-ms:


e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -96,7 +96,7 @@ public static async Task ActivateFaultInjectionAsync(ITransportSettings transpor
                 logger.Trace($"{nameof(ActivateFaultInjectionAsync)}: {ex}");
 
                 // For quota injection, the fault is only seen for the original HTTP request.
-                if (transportSettings is Client.HttpTransportSettings)
+                if (transportSettings is IotHubClientHttpSettings)
```
abhipsaMisra, line 99: Is there a reason why we're prefixing everything with IotHubClient? Is that to differentiate from IotHubServiceHttpSettings? They being under different assemblies and thus different namespaces, wouldn't that be self-explanatory?

drwill-ms: Differentiate from DPS device client's, and possibly the hub service client's, transport settings, yes.

drwill-ms: We're trying to avoid having the same name of types especially in a device app which uses DPS and hub device clients there are no naming conflicts that have to be differentiated by using more namespace or namespace aliasing.

timtay-microsoft (approved):

PR #2525, opened 2022-07-27
Move TlsVerions properties to transport settings for device client

brycewang-microsoft:


iothub/device/src/ClientSettings/AmqpTransportSettings.cs
```diff
@@ -128,5 +125,58 @@ public override string ToString()
         {
             return $"{GetType().Name}/{Protocol}";
         }
+
+        /// <summary>
+        /// The acceptable versions of TLS to use when the SDK must be explicit.
+        /// </summary>
+        public SslProtocols MinimumTlsVersions { get; set; } = SslProtocols.Tls12;
```
drwill-ms, line 132: Was `SslProtocols.None` not available? That one is to let the OS decide.

iothub/device/src/ClientSettings/AmqpTransportSettings.cs
```diff
@@ -128,5 +125,58 @@ public override string ToString()
...
+            // ensure the preferred TLS version is included
+            if (((protocols & SslProtocols.Tls) != 0
+                    || (protocols & SslProtocols.Tls11) != 0)
+                && (protocols & PreferredProtocol) == 0)
+            {
+                protocols ^= PreferredProtocol;
+            }
+
+            MinimumTlsVersions = Preferred = protocols;
+        }
```
drwill-ms, line 178: Did we still need this? Will a single property alone not do?

PR #2528, opened 2022-07-27
Move TlsVerions properties to IotHubClientTransportSettings for device client

brycewang-microsoft:


iothub/device/src/TransportSettings/IotHubClientTransportSettings.cs
```diff
@@ -72,5 +73,63 @@ public override string ToString()
...
+        /// </summary>
+        /// <remarks>
+        /// Will ignore a version less than TLS 1.0.
+        ///
+        /// Affects:
+        /// 1. MinimumTlsVersions property
+        /// 2. Preferred property
+        /// 3. For .NET framework 4.5.1 over HTTPS or websocket, as it does not offer a "SystemDefault" option, the version must be explicit.
+        /// </remarks>
+        public void SetMinimumTlsVersions(SslProtocols protocols = SslProtocols.None)
```
brycewang-microsoft, line 110: We still need this as it ignore versions less than TLS 1.0 and ensure that users don't use deprecated SslProtocols values.

drwill-ms: What do we care if they set a value they want? The service will reject versions it doesn't accept.

drwill-ms (approved):

PR #2529, opened 2022-07-28
Refactor cs builder to use ctor

drwill-ms:


iothub/device/src/Authentication/DeviceAuthenticationWithSakRefresh.cs
```diff
@@ -10,18 +10,18 @@ namespace Microsoft.Azure.Devices.Client
     // Implementing SAS Token refresh based on a SharedAccessKey (SAK).
     internal class DeviceAuthenticationWithSakRefresh : DeviceAuthenticationWithTokenRefresh
     {
-        private readonly IotHubConnectionString _connectionString;
+        private readonly IotHubConnectionInfo _connectionString;
```
drwill-ms, line 13: I think the name was misleading. It isn't a connection string, rather it parses them, holds the info, produces one, and generates new passwords based on the auth information it was provided.

Also, it was split into 2 files, for no apparent reason. I consolidated them.

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -3,15 +3,16 @@
 
 using System;
 using System.Diagnostics;
+using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client.Extensions;
 
 namespace Microsoft.Azure.Devices.Client
 {
-    internal sealed partial class IotHubConnectionString : IAuthorizationProvider
+    internal class IotHubConnectionInfo : IAuthorizationProvider
```
drwill-ms, line 11: I think the name was misleading. It isn't a connection string, rather it parses them, holds the info, produces one, and generates new passwords based on the auth information it was provided.

Also, it was split into 2 files, for no apparent reason. I consolidated them.

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -57,74 +57,50 @@ internal IotHubConnectionStringBuilder()
...
-            {
-                HostName = "TEMP.HUB",
-            };
+            Argument.AssertNotNullOrWhiteSpace(iotHubConnectionString, nameof(iotHubConnectionString));
 
             if (authenticationMethod == null)
             {
-                csBuilder.Parse(iotHubConnectionString);
-                csBuilder.AuthenticationMethod = AuthenticationMethodFactory.GetAuthenticationMethod(csBuilder);
+                // We'll parse the connection string and use that to build an auth method
```
drwill-ms, line 92: This is weird, so I added some code comments to help future readers.

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -245,8 +221,6 @@ private void Parse(string iotHubConnectionString)
             UsingX509Cert = GetConnectionStringOptionalValueOrDefault<bool>(map, X509CertPropertyName, ParseX509, true)
                 || GetConnectionStringOptionalValueOrDefault<bool>(map, CommonX509CertPropertyName, ParseX509, true);
             GatewayHostName = GetConnectionStringOptionalValue(map, GatewayHostNamePropertyName);
-
-            Validate();
```
drwill-ms, line 249: Removed the implicit call to `Validate()` during `Parse()` because it needed to be called after the object was initialized. Because we assign the auth method (which modifies the object) then parse the connection string in one case, and do it in reverse order in the other case, this meant validation didn't happen when it should.

So now it always happens explicitly (like in the other ctor), which changes the behavior of the object ever so slightly (removed a unit test, changed an expected exception type in another), which I don't think will have much impact.

iothub/device/tests/Authentication/IotHubConnectionStringBuilderTests.cs
```diff
@@ -214,38 +214,30 @@ public void IotHubConnectionStringBuilder_ParamConnectionString_ParsesGatewayHos
...
+        [ExpectedException(typeof(ArgumentNullException))]
         public void IotHubConnectionStringBuilder_ParamConnectionString_MissingDeviceId_Throws()
         {
             var connectionString = $"HostName={HostName};SharedAccessKey={SharedAccessKey}";
-            var csBuilder = IotHubConnectionStringBuilder.Create(connectionString);
-        }
-
-        [TestMethod]
-        [ExpectedException(typeof(ArgumentException))]
-        public void IotHubConnectionStringBuilder_ParamConnectionString_IncludesBothSharedAcccess_Throws()
```
drwill-ms, line 230: Yes, this is an invalid connection string, but once initialized by the builder when an auth method is created, it nulls out the shared access key, making it valid. I find it hard to get very excited about this condition being caught when the code fixes it.

Thoughts?

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -57,74 +57,50 @@ internal IotHubConnectionStringBuilder()
...
-                HostName = "TEMP.HUB",
-            };
+            Argument.AssertNotNullOrWhiteSpace(iotHubConnectionString, nameof(iotHubConnectionString));
 
             if (authenticationMethod == null)
             {
-                csBuilder.Parse(iotHubConnectionString);
-                csBuilder.AuthenticationMethod = AuthenticationMethodFactory.GetAuthenticationMethod(csBuilder);
+                // We'll parse the connection string and use that to build an auth method
+                Parse(iotHubConnectionString);
```
abhipsaMisra, line 93: This class always read a bit weird to me because of the way the constructor, `Create` and `Parse` methods were used. I'd expect a `Parse` method to return the "parsed" object, but that was weirdly managed by this class because the `Parse` method set member variables. Now that you've refactored the "Create" logic, is there a way for us to modify the "Parse" method as well so that we either set member variables in the constructor or have the `Parse` method return the "parsed" object.
Do you think that is something this class can allow?

iothub/device/src/Authentication/DeviceAuthenticationWithSakRefresh.cs
abhipsaMisra, line 13: nit: we'd probably want to rename these variable names as well

iothub/device/tests/Authentication/IotHubConnectionStringBuilderTests.cs
andyk-ms: Looks like this is a set of testing different way of the invalid connection string. Make sense when this is initially created to make sure the API handles them. Maybe we can remove them for v2.

andyk-ms (approved):

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -57,74 +57,50 @@ internal IotHubConnectionStringBuilder()
...
         /// </summary>
-        /// <param name="hostname">The fully-qualified DNS hostname of IoT hub</param>
-        /// <param name="gatewayHostname">The fully-qualified DNS hostname of the gateway</param>
-        /// <param name="authenticationMethod">The authentication method that is used</param>
+        /// <param name="authenticationMethod">The authentication method that is used.</param>
+        /// <param name="hostName">The fully-qualified DNS host name of IoT hub.</param>
+        /// <param name="gatewayHostName">The fully-qualified DNS host name of the gateway.</param>
         /// <returns>A new instance of the <see cref="IotHubConnectionStringBuilder"/> class with a populated connection string.</returns>
-        public static IotHubConnectionStringBuilder Create(string hostname, string gatewayHostname, IAuthenticationMethod authenticationMethod)
+        public IotHubConnectionStringBuilder(IAuthenticationMethod authenticationMethod, string hostName, string gatewayHostName = null)
```
abhipsaMisra, line 67: Do you think this would be a useful helper method for users or could we make this internal?

abhipsaMisra: Helper = IotHubConnectionStringBuilder itself

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
drwill-ms: Parse is just a private method, so I don't really compare it to Create. It isn't as if users of the class would see both Create and Parse as options.

The refactored Create logic was just getting rid of the factory method and using constructors instead.

If we made parse only get (not set) the values, it would need some (new) object to put them on, then the ctor would then assign the properties from that other object. Honestly, I don't see the point of that.

To me, the weirdest part of this whole thing is that we create an auth method and pass `this`. The new auth method object now changes properties on this class. That was unexpected and creates the 2-ways-to-create-this-object.

So, looking at it again, I decided to split the cs + authmethod into 2 methods and found that we only had references for cs only. So, I took out the cs + authmethod overload. That reduces that oddness.

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
drwill-ms: I think we could make it internal and only bring it back if a customer makes a case for why it was valuable to them.

drwill-ms: Well, no because IAuthenticationMethod refers to this type in the impl. So it wouldn't be that easy.

iothub/device/src/InternalClient.cs
```diff
@@ -146,7 +146,7 @@ public string ProductInfo
 
         internal IDelegatingHandler InnerHandler { get; set; }
 
-        internal IotHubConnectionString IotHubConnectionString { get; private set; }
+        internal IotHubConnectionInfo IotHubConnectionString { get; private set; }
```
abhipsaMisra, line 149: nit: rename property

iothub/device/src/Pipeline/PipelineContext.cs
```diff
@@ -11,7 +11,7 @@ internal class PipelineContext
 
         internal ProductInfo ProductInfo { get; set; }
 
-        internal IotHubConnectionString IotHubConnectionString { get; set; }
+        internal IotHubConnectionInfo IotHubConnectionString { get; set; }
```
abhipsaMisra, line 14: nit: rename property

iothub/device/src/Pipeline/TransportHandlerFactory.cs
```diff
@@ -15,7 +15,7 @@ public IDelegatingHandler Create(PipelineContext context)
             // ProtocolRoutingDelegatingHandler configures the TransportSettings configuration
             // which is different from TransportSettings[] element.
             IotHubClientTransportSettings transportSettings = context.TransportSettings;
-            IotHubConnectionString connectionString = context.IotHubConnectionString;
+            IotHubConnectionInfo connectionString = context.IotHubConnectionString;
```
abhipsaMisra, line 18: same nit, here and below :)

iothub/device/tests/Authentication/IotHubConnectionStringBuilderTests.cs
abhipsaMisra: Passing in both a shared access signature and shared access key, you'd expect the more specific credential (sas) to be used. That being the behavior exhibited by our SDK, I don't mind us not throwing an exception. It would be a good bonus if we could call this out somewhere (DeviceClient connection string constructor?).

abhipsaMisra (approved):

PR #2530, opened 2022-07-28
Create release build test pipeline

andyk-ms:
New .NET release build, E2E test, and nuget package pipeline

### DIFF
--- a/vsts/build-test-nuget.yml
+++ b/vsts/build-test-nuget.yml
@@ -1,0 +1,423 @@
+name: $(BuildID)_$(BuildDefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
+resources:
+  - repo: self
+    clean: true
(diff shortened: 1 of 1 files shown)


abhipsaMisra (approved):
We'll also want to retire these configuration files:
https://github.com/Azure/azure-iot-sdk-csharp/blob/main/vsts/build-release-artifacts.yml
https://github.com/Azure/azure-iot-sdk-csharp/blob/main/vsts/test-release-nuget.yaml

andyk-ms:
> We'll also want to retire these configuration files: https://github.com/Azure/azure-iot-sdk-csharp/blob/main/vsts/build-release-artifacts.yml https://github.com/Azure/azure-iot-sdk-csharp/blob/main/vsts/test-release-nuget.yaml

Will retire old pipelines (main/preview) and corresponding yaml as a separate steps.

PR #2534, opened 2022-07-28
DigitalTwins subclient rework

timtay-microsoft:
Removes the dependency on Microsoft.Rest.Runtime and reshapes the digital twin client's API surface to no longer use the Microsoft.Rest.Runtime classes that we used previously

This also removes all auto-generation scripts and other resources used in auto-generating this client in the past. This is being done since we likely won't autogenerate just this client again.

This also removes the folder of classes we had dedicated to implementing Microsoft.Rest.Runtime authentication primitives within the context of our service. Now we'll be using the same authentication classes as the rest of the service client subclients. These authentication methods include the track 2 friendly AzureSasCredential and TokenCredential, and 
 include using a simple connection string.


iothub/service/src/Http2/HttpMessageHelper2.cs
```diff
@@ -48,7 +50,17 @@ internal static async Task ValidateHttpResponseStatus(HttpStatusCode expectedHtt
...
             {
-                throw await ExceptionHandlingHelper.GetDefaultErrorMapping()[responseMessage.StatusCode].Invoke(responseMessage);
+                IReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> defaultErrorMapping =
+                    ExceptionHandlingHelper.GetDefaultErrorMapping();
+                if (defaultErrorMapping.TryGetValue(responseMessage.StatusCode, out Func<HttpResponseMessage, Task<Exception>> mappedException))
+                {
+                    throw await mappedException.Invoke(responseMessage);
+                }
+
+                // Default case for when the mapping of this error code to an exception does not exist yet
```
timtay-microsoft, line 60: This was orthogonal to the digital twin client work, but I noticed we didn't throw an appropriate exception in the default case

iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -8,8 +8,6 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <OutputPath>.\bin\$(Configuration)</OutputPath>
     <DocumentationFile>$(OutputPath)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
-    <!-- Don't auto-compile platform specific files: -->
-    <DefaultItemExcludes>ModernDotNet/**;DigitalTwin/**;$(DefaultItemExcludes)</DefaultItemExcludes>
```
timtay-microsoft, line 12: We no longer have platform specific files

iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -129,11 +124,9 @@
 
   <ItemGroup>
     <Compile Include="$(Common)\DefaultWebProxySettings.cs" />
-    <Compile Include="DigitalTwin\**\*.cs" />
```
timtay-microsoft, line 132: No need to manually include these since they are no longer excluded from the project in the above statement

iothub/service/tests/TokenHelperTests.cs
```diff
@@ -7,7 +7,6 @@
 using System.Threading;
 using FluentAssertions;
 using Microsoft.Azure.Devices.Common;
-using Microsoft.Azure.Devices.DigitalTwin.Authentication;
```
timtay-microsoft, line 10: Unused

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -129,16 +129,15 @@ public async Task DigitalTwinClient_Http_SasCredentialAuth_Success()
...
             string signature = TestConfiguration.IoTHub.GetIotHubSharedAccessSignature(TimeSpan.FromHours(1));
-            using var digitalTwinClient = DigitalTwinClient.Create(
+            using var serviceClient = new IotHubServiceClient(
                 TestConfiguration.IoTHub.GetIotHubHostName(),
                 new AzureSasCredential(signature));
 
             // act
-            HttpOperationResponse<ThermostatTwin, DigitalTwinGetHeaders> response = await digitalTwinClient
-                .GetDigitalTwinAsync<ThermostatTwin>(testDevice.Id)
+            DigitalTwinGetResponse<ThermostatTwin> response = await serviceClient.DigitalTwins
```
timtay-microsoft, line 137: What do you all think of this name? Would ```GetDigitalTwinResponse``` make more sense?

abhipsaMisra: It's been a while since I've looked at these APIs; do you have a sample response that shows what `DigitalTwinGetResponse<ThermostatTwin> response` looks like?

timtay-microsoft: Here is the definition for [GetDigitalTwinResponse](https://github.com/Azure/azure-iot-sdk-csharp/blob/timtay/digitaltwin/iothub/service/src/DigitalTwin/Models/GetDigitalTwinResponse.cs). The HTTP response from the server is defined such that the payload is the digital twin and there is a header that specifies the ETag for the digital twin

e2e/test/iothub/service/DigitalTwinClientE2ETests.cs
```diff
@@ -86,11 +83,14 @@ public async Task DigitalTwinWithOnlyRootComponentOperationsAsync()
 
                 // Invoke the root-level command "getMaxMinReport" on the digital twin.
                 DateTimeOffset since = DateTimeOffset.Now.Subtract(TimeSpan.FromMinutes(1));
-                string payload = JsonConvert.SerializeObject(since);
-                HttpOperationResponse<DigitalTwinCommandResponse, DigitalTwinInvokeCommandHeaders> commandResponse =
-                    await digitalTwinClient.InvokeCommandAsync(deviceId, commandName, payload).ConfigureAwait(false);
-                commandResponse.Body.Status.Should().Be(expectedCommandStatus);
-                commandResponse.Body.Payload.Should().Be(JsonConvert.SerializeObject(commandName));
+                var requestOptions = new DigitalTwinInvokeCommandRequestOptions()
```
drwill-ms, line 86:
```suggestion
                var requestOptions = new DigitalTwinInvokeCommandRequestOptions
```

e2e/test/iothub/service/DigitalTwinClientE2ETests.cs
```diff
@@ -86,11 +83,14 @@ public async Task DigitalTwinWithOnlyRootComponentOperationsAsync()
...
+                {
+                    Payload = JsonConvert.SerializeObject(since)
+                };
+                DigitalTwinCommandResponse commandResponse =
+                    await serviceClient.DigitalTwins.InvokeCommandAsync(deviceId, commandName, requestOptions).ConfigureAwait(false);
```
drwill-ms, line 91:
```suggestion
                DigitalTwinCommandResponse commandResponse = await serviceClient
                    .DigitalTwins.InvokeCommandAsync(deviceId, commandName, requestOptions)
                    .ConfigureAwait(false);
```

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -0,0 +1,363 @@
...
+    /// This client contains methods to retrieve and update digital twin information, and invoke commands
+    /// on a digital twin device.
+    /// </summary>
+    /// <seealso href="https://docs.microsoft.com/en-us/azure/iot-develop/concepts-digital-twin"/>
+    public class DigitalTwinsClient
+    {
+        private string _hostName;
+        private IotHubConnectionProperties _credentialProvider;
+        private HttpClient _httpClient;
+        private HttpRequestMessageFactory _httpRequestMessageFactory;
```
drwill-ms, line 25:
```suggestion
        private readonly IotHubConnectionProperties _credentialProvider;
        private readonly HttpClient _httpClient;
        private readonly HttpRequestMessageFactory _httpRequestMessageFactory;
```

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -0,0 +1,363 @@
...
+        /// Creates an instance of this class. Provided for unit testing purposes only.
+        /// </summary>
+        protected DigitalTwinsClient()
+        {
+        }
+
+        internal DigitalTwinsClient(string hostName, IotHubConnectionProperties credentialProvider, HttpClient httpClient, HttpRequestMessageFactory httpRequestMessageFactory)
+        {
+            _credentialProvider = credentialProvider;
+            _hostName = hostName;
```
drwill-ms, line 45: Assign params in same order as they are passed.

```suggestion
            _hostName = hostName;
            _credentialProvider = credentialProvider;
```

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -0,0 +1,363 @@
...
+                Logging.Enter(this, $"Getting digital twin with Id: {digitalTwinId}", nameof(GetAsync));
+
+            try
+            {
+                Argument.RequireNotNullOrEmpty(digitalTwinId, nameof(digitalTwinId));
+
+                cancellationToken.ThrowIfCancellationRequested();
+
+                using HttpRequestMessage request = _httpRequestMessageFactory.CreateRequest(HttpMethod.Get, GetDigitalTwinRequestUri(digitalTwinId), _credentialProvider);
+                HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken);
```
drwill-ms, line 80:
```suggestion
                HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken).ConfigureAwait(false);
```

drwill-ms: And below...

iothub/service/src/DigitalTwin/Models/DigitalTwinCommandResponse.cs
```diff
@@ -17,5 +17,10 @@ public class DigitalTwinCommandResponse
         /// Command invocation result payload, as supplied by the device.
         /// </summary>
         public string Payload { get; internal set; }
+
+        /// <summary>
+        /// Server Generated Request Id (GUID), to uniquely identify this request in the service.
```
drwill-ms, line 22: Unnecessary capitalization.

```suggestion
        /// Server generated request Id (GUID), to uniquely identify this request in the service.
```

iothub/service/src/DigitalTwin/Models/GetDigitalTwinResponse.cs
```diff
@@ -0,0 +1,33 @@
...
+            ETag = eTag;
+        }
+
+        /// <summary>
+        /// Gets the deserialized digital twin.
+        /// </summary>
+        public T DigitalTwin { get; internal set; }
+
+        /// <summary>
+        /// Gets the weak Etag of the modified resource
```
drwill-ms, line 29:
```suggestion
        /// Gets the weak Etag of the modified resource.
```

iothub/service/src/DigitalTwin/Models/GetDigitalTwinResponse.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// The service response to a get digital twin request.
+    /// </summary>
+    /// <typeparam name="T">The type of the digital twin.</typeparam>
+    public class GetDigitalTwinResponse<T>
+    {
+        /// <summary>
+        /// Initializes a new instance of this class.
+        /// </summary>
+        /// <param name="eTag">Weak Etag of the modified resource.</param>
```
drwill-ms, line 15:
```suggestion
        /// <param name="eTag">Weak ETag of the modified resource.</param>
```

iothub/service/src/DigitalTwin/Models/UpdateDigitalTwinResponse.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// The service response to a digital twin update operation.
+    /// </summary>
+    public class UpdateDigitalTwinResponse
+    {
+        /// <summary>
+        /// Initializes a new instance of the
+        /// DigitalTwinUpdateHeaders class.
```
drwill-ms, line 15: Overly aggressive line length and copy/pasta error (?) of class name.

```suggestion
        /// Initializes a new instance of this class.
```

iothub/service/src/IotHubServiceClient.cs
```diff
@@ -142,7 +149,14 @@ public IotHubServiceClient(string hostName, AzureSasCredential credential, IotHu
         public ConfigurationsClient Configurations { get; protected set; }
 
         /// <summary>
-        /// Dispose this client and all the disposable resources it has. This includes any HTTP clients
+        /// Subclient of <see cref="IotHubServiceClient"/> that handles all digital twin operations including
+        /// getting a digital twin, updating a digital twin, and invoking commands on a digital twin.
+        /// </summary>
+        /// <seealso href="https://docs.microsoft.com/en-us/azure/iot-develop/concepts-digital-twin"/>
```
drwill-ms, line 155: Remove `en-us` part of URI.

```suggestion
        /// <seealso href="https://docs.microsoft.com/azure/iot-develop/concepts-digital-twin"/>
```

e2e/test/iothub/SasCredentialAuthenticationTests.cs
drwill-ms: In track2, a slight equivalent would be the request options for the calls (which we don't expose to the users, for the record).

<https://github.com/Azure/azure-sdk-for-net/blob/f35da107f6d8aa69ff3b0ac64fd3a1853f527212/sdk/digitaltwins/Azure.DigitalTwins.Core/src/Customized/Models/CreateOrReplaceDigitalTwinOptions.cs#L9>

In that case, we use the format: verb-noun-purpose.

So, I'd say GetDigitalTwinResponse would be consistent with that.

timtay-microsoft: I'll rename the request option types in this PR to match that pattern as well then

e2e/test/iothub/service/DigitalTwinClientE2ETests.cs
```diff
@@ -47,13 +45,13 @@ public async Task DigitalTwinWithOnlyRootComponentOperationsAsync()
...
-                using var digitalTwinClient = DigitalTwinClient.CreateFromConnectionString(s_connectionString);
+                using var serviceClient = new IotHubServiceClient(s_connectionString);
 
                 // Retrieve the digital twin.
-                HttpOperationResponse<ThermostatTwin, DigitalTwinGetHeaders> response =
-                    await digitalTwinClient.GetDigitalTwinAsync<ThermostatTwin>(deviceId).ConfigureAwait(false);
-                ThermostatTwin twin = response.Body;
-                twin.Metadata.ModelId.Should().Be(ThermostatModelId);
+                GetDigitalTwinResponse<ThermostatTwin> response =
+                    await serviceClient.DigitalTwins.GetAsync<ThermostatTwin>(deviceId).ConfigureAwait(false);
```
drwill-ms, line 52: Other cases like this too:
```suggestion
                GetDigitalTwinResponse<ThermostatTwin> response = await serviceClient.DigitalTwins
                    .GetAsync<ThermostatTwin>(deviceId)
                    .ConfigureAwait(false);
```

iothub/service/src/DigitalTwin/Models/UpdateDigitalTwinResponse.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <param name="eTag">Weak Etag of the modified resource</param>
+        /// <param name="location">URI of the digital twin</param>
+        public UpdateDigitalTwinResponse(string eTag = default(string), string location = default(string))
+        {
+            ETag = eTag;
+            Location = location;
+        }
+
+        /// <summary>
+        /// Gets or sets weak Etag of the modified resource
```
drwill-ms, line 25:
```suggestion
        /// Gets or sets weak Etag of the modified resource.
```

iothub/service/src/DigitalTwin/Models/UpdateDigitalTwinResponse.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "ETag")]
+        public string ETag { get; internal set; }
+
+        /// <summary>
+        /// Gets the URI of the digital twin
```
drwill-ms, line 31:
```suggestion
        /// Gets the URI of the digital twin.
```

drwill-ms (approved):

e2e/test/iothub/service/DigitalTwinClientE2ETests.cs
```diff
@@ -47,13 +45,15 @@ public async Task DigitalTwinWithOnlyRootComponentOperationsAsync()
...
                 // Retrieve the digital twin.
-                HttpOperationResponse<ThermostatTwin, DigitalTwinGetHeaders> response =
-                    await digitalTwinClient.GetDigitalTwinAsync<ThermostatTwin>(deviceId).ConfigureAwait(false);
-                ThermostatTwin twin = response.Body;
-                twin.Metadata.ModelId.Should().Be(ThermostatModelId);
+                GetDigitalTwinResponse<ThermostatTwin> response = await serviceClient.DigitalTwins
+                    .GetAsync<ThermostatTwin>(deviceId)
+                    .ConfigureAwait(false);
+
+                ThermostatTwin twin = response.DigitalTwin;
```
abhipsaMisra, line 55: not asserting this anymore?
twin.Metadata.ModelId.Should().Be(ThermostatModelId);

iothub/service/src/DigitalTwin/Models/GetDigitalTwinResponse.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// The service response to a get digital twin request.
+    /// </summary>
+    /// <typeparam name="T">The type of the digital twin.</typeparam>
+    public class GetDigitalTwinResponse<T>
```
abhipsaMisra, line 10: Is it just me or does "**Get**DigitalTwinResponse" seem like a weird name for a data object? The "Get" verb in here is throwing me off since on first glance this looks more like an API call than a data object.

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -0,0 +1,363 @@
...
+
+            try
+            {
+                Argument.RequireNotNullOrEmpty(digitalTwinId, nameof(digitalTwinId));
+
+                cancellationToken.ThrowIfCancellationRequested();
+
+                using HttpRequestMessage request = _httpRequestMessageFactory.CreateRequest(HttpMethod.Get, GetDigitalTwinRequestUri(digitalTwinId), _credentialProvider);
+                HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken).ConfigureAwait(false);
+                await HttpMessageHelper2.ValidateHttpResponseStatus(HttpStatusCode.OK, response).ConfigureAwait(false);
```
abhipsaMisra, line 81: Q - why is this response code validator an async call? Don't we have the I/O operation result already, or are we making another call?

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -0,0 +1,363 @@
...
+                T digitalTwin = await HttpMessageHelper2.DeserializeResponse<T>(response, cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 82: Same question for this deserializer as well - why is it async?

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -0,0 +1,363 @@
...
+                    requestOptions?.Payload,
+                    queryStringParameters);
+
+                HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken).ConfigureAwait(false);
+                await HttpMessageHelper2.ValidateHttpResponseStatus(HttpStatusCode.OK, response).ConfigureAwait(false);
+
+                // No need to deserialize here since the user will deserialize this into their expected type
+                // after this function returns.
+                string responsePayload = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                int responseStatusCode = int.Parse(response.Headers.GetValues("x-ms-command-statuscode").FirstOrDefault());
```
abhipsaMisra, line 223: nit: we should save these "x-ms-..." key names into constants.

e2e/test/iothub/service/DigitalTwinClientE2ETests.cs
timtay-microsoft: My mistake, I'll add that line back in

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
timtay-microsoft: The pre-existing exception mapping logic has a single async step in it for reading the contents of the http response. See [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/be46adbbe5a0a82be8b84ead85d13e1c947ef1d3/iothub/service/src/Http2/HttpMessageHelper2.cs#L51)

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
timtay-microsoft: This helper method reads the payload of the http response with ```response.Content.ReadHttpContentAsStringAsync(...)```

Would it be better to do this synchronously?

iothub/service/src/DigitalTwin/Models/GetDigitalTwinResponse.cs
timtay-microsoft: I can see where you are coming from. Do you have any suggestions on a possible rename? DigitalTwinWithVersion?

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
abhipsaMisra: Ah, I see. No, what you have here is good; we don't want to block on an operation when an async version exists. 
Another option is to directly read the response content once, and then pass in the string to the validate and deserialize methods. We should likely also see some improvement with speed, although I think it'll be quite small (since we're only making 2 calls per API).

abhipsaMisra (approved):

iothub/service/src/DigitalTwin/Models/GetDigitalTwinResponse.cs
abhipsaMisra: I'm ok with DigitalTwinWithVersion. We also have the update variant, what would we rename that to? UpdateDigitalTwinResponse -> DigitalTwinUpdatedResponse?

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
timtay-microsoft: The ```ValidateHttpResponseStatus``` call doesn't read the body unless an error code is attached to the response, so we woudn't save any payload read operations by factoring it out here. We either read it once before throwing an exception or we read it once before returning the successful result

iothub/service/src/DigitalTwin/Models/GetDigitalTwinResponse.cs
timtay-microsoft: DigitalTwinUpdatedResponse contains only the ETag and location of the digital twin, so following the naming pattern would give us something like DigitalTwinWithVersionAndLocation which sounds clunky to me.

I originally had these return types named like "DigitalTwinGetResponse" and "DigitalTwinUpdateResponse" and we seem to be almost circling back to them. Do you like "DigitalTwinGetResponse" more than "GetDigitalTwinResponse"?

abhipsaMisra: I personally do like DigitalTwinGetResponse more than GetDigitalTwinResponse. The latter, to me, looks like an API call.

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -0,0 +1,366 @@
...
+                int responseStatusCode = int.Parse(response.Headers.GetValues(StatusCodeHeaderKey).FirstOrDefault());
+                string requestId = response.Headers.GetValues(RequestIdHeaderKey).FirstOrDefault();
+                var commandResponse = new InvokeDigitalTwinCommandResponse()
+                {
+                    Payload = responsePayload,
+                    Status = responseStatusCode,
+                    RequestId = requestId,
+                };
+
+                return commandResponse;
```
drwill-ms, line 235:
```suggestion
                return new InvokeDigitalTwinCommandResponse
                {
                    Payload = responsePayload,
                    Status = responseStatusCode,
                    RequestId = requestId,
                };
```

iothub/service/src/DigitalTwin/Models/UpdateDigitalTwinResponse.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <param name="eTag">Weak Etag of the modified resource.</param>
+        /// <param name="location">URI of the digital twin.</param>
+        public UpdateDigitalTwinResponse(string eTag = default(string), string location = default(string))
+        {
+            ETag = eTag;
+            Location = location;
+        }
+
+        /// <summary>
+        /// Gets or sets weak Etag of the modified resource.
```
drwill-ms, line 25:
```suggestion
        /// Gets or sets weak ETag of the modified resource.
```

iothub/service/src/DigitalTwin/Models/UpdateDigitalTwinResponse.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>
+    /// The service response to a digital twin update operation.
+    /// </summary>
+    public class UpdateDigitalTwinResponse
+    {
+        /// <summary>
+        /// Initializes a new instance of this class.
+        /// </summary>
+        /// <param name="eTag">Weak Etag of the modified resource.</param>
```
drwill-ms, line 16:
```suggestion
        /// <param name="eTag">Weak ETag of the modified resource.</param>
```

iothub/service/src/DigitalTwin/Models/GetDigitalTwinResponse.cs
timtay-microsoft: Let's go with that pattern then. DigitalTwinGetResponse and DigitalTwinUpdateResponse

PR #2537, opened 2022-08-01
Further split out client init into two paths

drwill-ms:
Also, reject x509 cert using a connection string

### DIFF
--- a/e2e/test/iothub/SasCredentialAuthenticationTests.cs
+++ b/e2e/test/iothub/SasCredentialAuthenticationTests.cs
@@ -13,7 +13,7 @@ using Microsoft.Azure.Devices.E2ETests.Helpers;
 using Microsoft.Rest;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
+namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
 {
     /// <summary>
     /// Tests to ensure authentication using SAS credential succeeds in all the clients.
--- a/iothub/device/src/ClientFactory.cs
+++ b/iothub/device/src/ClientFactory.cs
@@ -129,11 +129,11 @@ namespace Microsoft.Azure.Devices.Client
             Argument.AssertNotNull(csBuilder, nameof(csBuilder));
             Argument.AssertNotNull(options, nameof(options));
 
-
             // Clients that derive their authentication method from AuthenticationWithTokenRefresh will need to specify
             // the token time to live and renewal buffer values through the corresponding AuthenticationWithTokenRefresh
             // implementation constructors instead.
-            if (!(csBuilder.AuthenticationMethod is AuthenticationWithTokenRefresh or DeviceAuthenticationWithX509Certificate))
+            if (csBuilder.AuthenticationMethod is not AuthenticationWithTokenRefresh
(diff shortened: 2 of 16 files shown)


e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -13,7 +13,7 @@
 using Microsoft.Rest;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
+namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
```
drwill-ms, line 16: Fixed this while I noticed the mis-cased namespaces in Test Explorer.

timtay-microsoft (approved):

iothub/device/src/ClientFactory.cs
```diff
@@ -22,9 +21,22 @@ internal static class ClientFactory
...
+                throw new ArgumentException("To use X509 certificates, use the initializer with the IAuthenticationMethod parameter.", nameof(connectionString));
+            }
+
+            // Clients that derive their authentication method from AuthenticationWithTokenRefresh will need to specify
+            // the token time to live and renewal buffer values through the corresponding AuthenticationWithTokenRefresh
+            // implementation constructors instead.
+            if (csBuilder.AuthenticationMethod is not AuthenticationWithTokenRefresh)
+            {
+                csBuilder.SasTokenTimeToLive = options?.SasTokenTimeToLive ?? default;
+                csBuilder.SasTokenRenewalBuffer = options?.SasTokenRenewalBuffer ?? default;
```
drwill-ms, line 36: I mostly copied this from below. Not sure how I feel about the TTL default being default(TimeSpan). Probably we should have defaults we apply, but no sure where that would be done. Maybe in `ClientOptions` itself since we have other defaults there too?

iothub/device/src/ClientFactory.cs
```diff
@@ -99,8 +118,7 @@ internal static InternalClient Create(string hostName, IAuthenticationMethod aut
         /// This initializer exists only for unit tests to override the pipeline creation
         /// </summary>
         /// <param name="pipelineBuilder">Should only be specified by SDK tests to override the operation pipeline.</param>
-        /// <param name="connectionString">The device connection string.</param>
-        /// <param name="authenticationMethod">The authentication method.</param>
```
drwill-ms, line 103: Changed this choke point factory method (last stop before calling `new` on `InternalClient` from taking a connection string. Now it takes the parsed `IotHubConnectionStringBuilder` class that Abhipsa will change to `DeviceIdentity`.

Consequently, isnce `IotHubConnectionStringBuilder` also has the `IAuthenticationMethod`, we don't need a separate param for it.

iothub/device/src/ClientFactory.cs
```diff
@@ -207,18 +175,6 @@ private static void EnsureOptionsIsSetup(X509Certificate2 cert, ref IotHubClient
                     options.TransportSettings.ClientCertificate = cert;
                 }
             }
-
-        }
-
-        private static IDeviceClientPipelineBuilder BuildPipeline()
```
drwill-ms, line 213: Moved to `InternalClient`

abhipsaMisra (approved):

PR #2538, opened 2022-08-01
refactor(iot-device): Remove DeviceIdentity and add the information to IotHubConnectionInfo

abhipsaMisra:


iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -97,33 +99,17 @@ public IotHubConnectionInfo(IotHubConnectionStringBuilder builder)
...
-            string hostName = null,
-            Uri httpsEndpoint = null,
-            Uri amqpEndpoint = null,
-            string audience = null,
-            string sharedAccessKeyName = null,
-            string sharedAccessKey = null,
-            string sharedAccessSignature = null,
-            bool isUsingGateway = false)
+        // This contructor is intended for creating an IotHubConnectionInfo that uses an IAuthenticationMethod-based authentication.
+        internal IotHubConnectionInfo()
```
abhipsaMisra, line 109: I've set this constructor skeleton code for us to use when we refactor the implementation of the connection string builder, together with IotHubConnectionInfo.

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -150,6 +136,14 @@ public IotHubConnectionInfo(IotHubConnectionStringBuilder builder)
 
         public bool IsUsingGateway { get; private set; }
 
+        public AuthenticationModel AuthenticationModel { get; }
+
+        public IotHubClientAmqpSettings AmqpTransportSettings { get; internal set; }
+
+        public ProductInfo ProductInfo { get; internal set; }
+
+        public IotHubClientOptions ClientOptions { get; internal set; }
```
abhipsaMisra, line 145: These properties have internal set because they are set on a pre-existing IotHubConnectionInfo a bit further down our transport pipeline.

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -187,5 +181,54 @@ public Uri BuildLinkAddress(string path)
...
+            return obj is IotHubConnectionInfo iotHubConnectionInfo
+                && GetHashCode() == iotHubConnectionInfo.GetHashCode()
+                && Equals(DeviceId, iotHubConnectionInfo.DeviceId)
+                && Equals(HostName, iotHubConnectionInfo.HostName)
+                && Equals(ModuleId, iotHubConnectionInfo.ModuleId)
+                && Equals(AmqpTransportSettings.Protocol, iotHubConnectionInfo.AmqpTransportSettings.Protocol)
+                && Equals(AuthenticationModel.GetHashCode(), iotHubConnectionInfo.AuthenticationModel.GetHashCode());
+        }
+
+        public override int GetHashCode()
```
abhipsaMisra, line 201: The hashcode algorithm used only these fields for creating the hash (in DeviceIdentity), so I left them as-is.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -48,9 +48,14 @@ static AmqpTransportHandler()
         {
             _operationTimeout = transportSettings.OperationTimeout;
             _onDesiredStatePatchListener = onDesiredStatePatchReceivedCallback;
-            IDeviceIdentity deviceIdentity = new DeviceIdentity(connectionInfo, transportSettings, context.ProductInfo, context.ClientOptions);
+
+            // Set the context information required for setting up the connection into IotHubConnectionInfo
+            connectionInfo.AmqpTransportSettings = transportSettings;
+            connectionInfo.ProductInfo = context.ProductInfo;
+            connectionInfo.ClientOptions = context.ClientOptions;
```
abhipsaMisra, line 55: This is the only place where a DeviceIdenity instance was being created, so I modified the code to update the information into the existing IotHubConnectionInfo.

iothub/device/src/Authentication/IIotHubConnectionInfo.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System;
+using Microsoft.Azure.Devices.Client.Transport;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Interface for device configurations and common attributes.
+    /// </summary>
+    internal interface IIotHubConnectionInfo : IAuthorizationProvider
```
abhipsaMisra, line 12: Moq can only mock interfaces and virtual methods, so this interface is meant to replace the IDeviceIdentity interface.

drwill-ms: I'd prefer `IDeviceIdentity` over this name. Thoughts?

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -97,58 +99,50 @@ public IotHubConnectionInfo(IotHubConnectionStringBuilder builder)
...
-            HostName = hostName;
-            HttpsEndpoint = httpsEndpoint;
-            AmqpEndpoint = amqpEndpoint;
-            Audience = audience;
-            SharedAccessKeyName = sharedAccessKeyName;
-            SharedAccessKey = sharedAccessKey;
-            SharedAccessSignature = sharedAccessSignature;
-            IsUsingGateway = isUsingGateway;
+            // if certificate information is present, set:
+            // AuthenticationModel = AuthenticationModel.X509;
```
drwill-ms, line 112: What's the plans here?

drwill-ms (approved):

iothub/device/src/Authentication/IIotHubConnectionInfo.cs
abhipsaMisra: I like IDeviceIdentity for this interface

iothub/device/src/Authentication/IotHubConnectionInfo.cs
abhipsaMisra: My plan is to use this constructor to create the IotHubConnectionInfo using an IAuthenticationMethod. Within this constructor we'll check if the certificate information is present, and set the AuthenticationModel in there.

iothub/device/src/Authentication/IDeviceIdentity.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.Azure.Devices.Client.Transport;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Interface for device configurations and common attributes.
```
drwill-ms, line 10: Want to mention it is included for our own unit test mocking?

iothub/device/src/Authentication/IDeviceIdentity.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    internal interface IDeviceIdentity : IAuthorizationProvider
```
drwill-ms, line 12: Or `IClientIdentity` if you want to disambiguate between devices and modules.

iothub/device/src/Authentication/IotHubConnectionInfo.cs
drwill-ms: So, this is a TODO for an upcoming change?

iothub/device/src/Authentication/IotHubConnectionInfo.cs
drwill-ms: Probably worth a code comment then, otherwise it is unclear why only some properties are included?

iothub/device/src/Authentication/IDeviceIdentity.cs
abhipsaMisra: I like that

iothub/device/src/Authentication/IotHubConnectionInfo.cs
abhipsaMisra: yes.

iothub/device/src/Authentication/IClientIdentity.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        string DeviceId { get; }
+
+        string ModuleId { get; }
+
+        string HostName { get; }
+
+        Uri HttpsEndpoint { get; }
+
+        Uri AmqpEndpoint { get; }
```
timtay-microsoft, line 27: Won't we use the same endpoint for both Https and Amqp? I get that we'll have some http/amqp specific suffixes and/or prefixes, but I'd envision that being known within the amqp/http layer

abhipsaMisra: You're correct in that. This class still has a bunch of things that we expect will change as we refactor the InternalClient implementation. We'll make those changes as we go. I didn't change all of it in this PR so as to avoid merge conflicts with the changes that David is working on.
Having said that, feel free to highlight things that look off so that we can keep those in mind as we refactor the other classes around this implementation.

timtay-microsoft: Gotcha, I didn't realize this was still a temporary field. No objections here then

PR #2539, opened 2022-08-01
v2 ScheduledJobs subclient

tmahmood-microsoft:
## Checklist


iothub/service/src/IotHubServiceClient.cs
```diff
@@ -147,6 +147,12 @@ public IotHubServiceClient(string hostName, AzureSasCredential credential, IotHu
         /// <seealso href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-query-language"/>
         public QueryClient Query { get; protected set; }
 
+        /// <summary>
+        /// Subclient of <see cref="IotHubServiceClient"/> for Scheduled Jobs management.
```
drwill-ms, line 151:
```suggestion
        /// Subclient of <see cref="IotHubServiceClient"/> for scheduled jobs management.
```

iothub/service/src/IotHubServiceClient.cs
```diff
@@ -147,6 +147,12 @@ public IotHubServiceClient(string hostName, AzureSasCredential credential, IotHu
...
+        /// </summary>
+        /// <seealso href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-jobs"/>
```
drwill-ms, line 153: Remember to remove the "en-us" from your URLs when including them.

```suggestion
        /// <seealso href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-jobs"/>
```

iothub/service/src/Http/HttpTransportSettings.cs
```diff
@@ -25,7 +25,7 @@ public HttpTransportSettings()
         /// An instance of a class that implements <see cref="IWebProxy"/>.
         /// </value>
         /// <remarks>
-        /// The <see cref="HttpTransportSettings"/> class is only used for the <see cref="JobClient"/> or the <see cref="RegistryManager"/>; so the proxy set here will only be valid for those clients.
+        /// The <see cref="HttpTransportSettings"/> class is only used for the <see cref="ScheduledJobsClient"/> or the <see cref="RegistryManager"/>; so the proxy set here will only be valid for those clients.
```
drwill-ms, line 28: @timtay-microsoft looks like we have an old reference to `RegistryManager` here.

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Common;
+using Microsoft.Azure.Devices.Common.Exceptions;
+using Microsoft.Azure.Devices.Http2;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Scheduled Jobs management.
```
drwill-ms, line 18:
```suggestion
    /// Scheduled jobs management.
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+    /// </summary>
+    /// <seealso href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-jobs"/>.
```
drwill-ms, line 20: Remove en-us

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+    public class ScheduledJobsClient
+    {
+        private string _hostName;
+        private IotHubConnectionProperties _credentialProvider;
+        private HttpClient _httpClient;
+        private HttpRequestMessageFactory _httpRequestMessageFactory;
```
drwill-ms, line 26: readonly

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+
+        private const string ContinuationTokenHeader = "x-ms-continuation";
+        private const string PageSizeHeader = "x-ms-max-item-count";
+
+        private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromSeconds(100);
+
+        /// <summary>
+        /// Creates scheduled jobs client, provided for unit testing purposes only.
+        /// </summary>
+        public ScheduledJobsClient()
```
drwill-ms, line 40:
```suggestion
        protected ScheduledJobsClient()
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+        {
+        }
+
+        internal ScheduledJobsClient(string hostName, IotHubConnectionProperties credentialProvider, HttpClient httpClient, HttpRequestMessageFactory httpRequestMessageFactory)
+        {
+            _credentialProvider = credentialProvider;
+            _hostName = hostName;
```
drwill-ms, line 47: Initialize in the same order as the parameters, when possible.

```suggestion
            _hostName = hostName;
            _credentialProvider = credentialProvider;
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+            _httpClient = httpClient;
+            _httpRequestMessageFactory = httpRequestMessageFactory;
+        }
+
+        /// <summary>
+        /// Explicitly open the ScheduledJobsClient instance.
```
drwill-ms, line 53: Avoid hard coding the client name.

```suggestion
        /// Explicitly open the client instance.
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+        /// </summary>
+        public virtual Task OpenAsync()
+        {
+            return TaskHelpers.CompletedTask;
+        }
+
+        /// <summary>
+        /// Closes the ScheduledJobsClient instance and disposes its resources.
+        /// </summary>
+        /// <seealso href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-jobs"/>.
```
drwill-ms, line 63: I don't see why the URL is relevant here, and in several other methods below.

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+            if (Logging.IsEnabled)
+                Logging.Enter(this, jobId, nameof(GetAsync));
+
+            try
+            {
+                Argument.RequireNotNullOrEmpty(jobId, nameof(jobId));
+                cancellationToken.ThrowIfCancellationRequested();
+
+                using HttpRequestMessage request = _httpRequestMessageFactory.CreateRequest(HttpMethod.Get, GetJobUri(jobId), _credentialProvider);
+                HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken);
```
drwill-ms, line 99: Add `ConfigureAwait(false)`

drwill-ms: And elsewhere

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,378 @@
...
+        /// request was throttled, <see cref="IotHubThrottledException"/> is thrown. For a complete list of possible
+        /// error cases, see <see cref="Common.Exceptions"/>.
+        /// </exception>
+        /// <exception cref="HttpRequestException">
+        /// If the HTTP request fails due to an underlying issue such as network connectivity, DNS failure, or server
+        /// certificate validation.
+        /// </exception>
+        /// <exception cref="OperationCanceledException">If the provided cancellation token has requested cancellation.</exception>
+        /// <seealso href="https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-jobs"/>.
+        public virtual async Task<JobResponse> ScheduleDeviceMethodAsync(string jobId, string queryCondition, CloudToDeviceMethod cloudToDeviceMethod, DateTime startTimeUtc, long maxExecutionTimeInSeconds, CancellationToken cancellationToken = default)
```
drwill-ms, line 209: Why not Logging.Enter/Exit calls like in other methods?

iothub/service/tests/JobClient/JobClientTests.cs
```diff
@@ -71,15 +69,13 @@ private void TestVerify(JobResponse actualJobResponse)
...
-            jobClient.Dispose();
             httpClientHelperMock.Verify(restOp => restOp.Dispose(), Times.Once());
         }
 
         [TestMethod]
-        public async Task CloseAsyncTest()
+        public void CloseAsyncTest()
         {
             httpClientHelperMock.Setup(restOp => restOp.Dispose());
-            await jobClient.CloseAsync().ConfigureAwait(false);
```
drwill-ms, line 82: Why remove the call to `CloseAsync()` in a test called `CloseAsyncTest()`?

drwill-ms: IOW, either add it back or remove the test if you assert the test is not needed.

iothub/service/src/Utilities/ClientApiVersionHelper.cs
```diff
@@ -14,7 +14,7 @@ internal class ClientApiVersionHelper
 
         /// <summary>
         /// The default API version query string parameter to use for all data-plane service calls in <see cref="RegistryManager"/>,
-        /// <see cref="ServiceClient"/>, and <see cref="JobClient"/>.
+        /// <see cref="ServiceClient"/>, and <see cref="ScheduledJobsClient"/>.
```
timtay-microsoft, line 17: I'd go ahead and delete all the references to specific clients here since we will now have many more.

"The API version used in all service requests." is sufficient here

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,386 @@
...
+        /// Thrown if IoT hub responded to the request with a non-successful status code. For example, if the provided
+        /// request was throttled, <see cref="IotHubThrottledException"/> is thrown. For a complete list of possible
+        /// error cases, see <see cref="Common.Exceptions"/>.
+        /// </exception>
+        /// <exception cref="HttpRequestException">
+        /// If the HTTP request fails due to an underlying issue such as network connectivity, DNS failure, or server
+        /// certificate validation.
+        /// </exception>
+        /// <exception cref="OperationCanceledException">If the provided cancellation token has requested cancellation.</exception>
+        public virtual Task<JobResponse> ScheduleTwinUpdateAsync(string jobId, string queryCondition, Twin twin, DateTime startTimeUtc, long maxExecutionTimeInSeconds, CancellationToken cancellationToken = default)
```
timtay-microsoft, line 264: This API is pretty heavy on parameters. I'd consider making a concrete type to encapsulate a few of these parameters

tmahmood-microsoft: can you give an example?

timtay-microsoft: Something like:

```csharp
public virtual Task<JobResponse> ScheduleTwinUpdateAsync(string jobId, ScheduledTwinUpdate scheduledTwinUpdate, CancellationToken cancellationToken = default);

public class ScheduledTwinUpdate
{
    public string queryCondition { get; set; }

    public Twin twin { get; set; }

    public DateTime startTimeUtc { get; set; }

    public long? maxExecutionTimeInSeconds { get; set; }
}
```

Doing it like this allows us to provide defaults values, leave values null that are optional, etc.

timtay-microsoft: The jobId parameter is interesting because I think it is required, but can be an arbitrary value. We may be able to label it as an optional parameter and provide it for the user if they don't specify a value.

timtay-microsoft: To build on this a bit, the usual pattern we follow for APIs like these is

```csharp
public void SomeMethod(string requiredParam1, string requiredParam2, SomeMethodRequestOptions options = default, CancellationToken cancellationToken = default)
```

Where ```SomeMethodRequestOptions``` contains all the optional parameters for the request. An example of this style can be seen [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/b7e0cb9a4b94dff1a4ed009eeb1ee0ac61c6bb15/iothub/service/src/DigitalTwin/DigitalTwinsClient.cs#L125).

In order to make this API both concise and flexible, you should aim to follow this pattern. In order to do that, you'll need to figure out which of these parameters are required by the service and which are optional.

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -94,13 +94,15 @@ public async Task JobClient_Http_SasCredentialAuth_Success()
             try
             {
                 // act
-                JobResponse createJobResponse = await serviceClient.ScheduledJobsClient
+                ScheduledTwinUpdate twinUpdate = new ScheduledTwinUpdate();
+                twinUpdate.queryCondition = query;
```
drwill-ms, line 98: Why are these properties camelCased?

iothub/service/src/Jobs/ScheduledTwinUpdate.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <summary>
+    /// Contains fields used when creating a job to update twin tags and desired properties on one or multiple devices.
+    /// </summary>
+    public class ScheduledTwinUpdate
+    {
+        /// <summary>
+        /// Query condition to evaluate which devices to run the job on.
+        /// </summary>
+        [JsonProperty(PropertyName = "queryCondition", Required = Required.Always)]
+        public string queryCondition { get; set; }
```
drwill-ms, line 15: Properties in .NET are TitleCased

iothub/service/src/Jobs/ScheduledTwinUpdate.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Twin object to use for the update
```
drwill-ms, line 18: All of these should have ending punctuation.

```suggestion
        /// Twin object to use for the update.
```

iothub/service/src/Jobs/ScheduledTwinUpdate.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "twin", Required = Required.Always)]
+        public Twin twin { get; set; }
+
+        /// <summary>
+        /// Date time in Utc to start the job
```
drwill-ms, line 24: Acronyms in all upper case.

```suggestion
        /// Date time in UTC to start the job.
```

e2e/test/iothub/service/IoTHubServiceProxyE2ETests.cs
```diff
@@ -88,17 +88,23 @@ private async Task JobClient_ScheduleAndRunTwinJob(HttpTransportSettings httpTra
...
                 {
                     string jobId = "JOBSAMPLE" + Guid.NewGuid().ToString();
                     string query = $"DeviceId IN ['{JobDeviceId}']";
-                    JobResponse createJobResponse = await jobClient
-                        .ScheduleTwinUpdateAsync(jobId, query, twin, DateTime.UtcNow, (long)TimeSpan.FromMinutes(2).TotalSeconds)
+                    ScheduledTwinUpdate twinUpdate = new ScheduledTwinUpdate();
+                    twinUpdate.queryCondition = query;
+                    twinUpdate.twin = twin;
+                    twinUpdate.startTimeUtc = DateTime.UtcNow;
+                    twinUpdate.maxExecutionTimeInSeconds = (long)TimeSpan.FromMinutes(2).TotalSeconds;
```
drwill-ms, line 105: Use an initializer list.

```suggestion
                    ScheduledTwinUpdate twinUpdate = new ScheduledTwinUpdate
                    {
                        queryCondition = query,
                        twin = twin,
                        startTimeUtc = DateTime.UtcNow,
                        maxExecutionTimeInSeconds = (long)TimeSpan.FromMinutes(2).TotalSeconds,
                    }
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,382 @@
...
+            return TaskHelpers.CompletedTask;
+        }
+
+        /// <summary>
+        /// Closes the client instance and disposes its resources.
+        /// </summary>
+        public virtual Task CloseAsync()
+        {
+            return TaskHelpers.CompletedTask;
+        }
```
drwill-ms, line 66: Why even have these methods if they do nothing and our clients don't inherit from a base class or interface which requires it?

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,382 @@
...
+
+        /// <summary>
+        /// Gets the job with the specified Id.
+        /// </summary>
+        /// <param name="jobId">Id of the job to retrieve</param>
```
drwill-ms, line 71: punctuation and simpler verbs.

```suggestion
        /// <param name="jobId">Id of the job to get.</param>
```

iothub/service/src/Jobs/ScheduledJobsOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// Unique Job Id for the twin update job.
+        /// </summary>
+        [JsonProperty(PropertyName = "JobId", Required = Required.Default)]
+        public string JobId { get; set; }
+
+        /// <summary>
+        /// Max execution time in seconds, i.e., ttl duration the job can run.
+        /// </summary>
+        [JsonProperty(PropertyName = "MaxExecutionTimeInSeconds", Required = Required.Default)]
+        public long MaxExecutionTimeInSeconds { get; set; }
```
timtay-microsoft, line 21: I see this same property in the non-optional parameter classes like ScheduledDeviceMethod

timtay-microsoft: It can either be optional or required, but not both

iothub/service/src/Jobs/ScheduledDeviceMethod.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Contains fields used when creating a job to run a device method on one or multiple devices.
+    /// </summary>
+    public class ScheduledDeviceMethod
```
timtay-microsoft, line 9:
```suggestion
    public class ScheduledDirectMethod
```

since "Direct Method" is the concept's name, not "Device Method"

iothub/service/src/Utilities/ClientApiVersionHelper.cs
```diff
@@ -5,16 +5,15 @@ namespace Microsoft.Azure.Devices
 {
     /// <summary>
     /// Holds the API version numbers required in data-plane calls to the service for <see cref="RegistryManager"/>,
-    /// <see cref="ServiceClient"/>, and <see cref="JobClient"/>.
+    /// <see cref="ServiceClient"/>, and <see cref="ScheduledJobsClient"/>.
```
timtay-microsoft, line 8: Same comment as before. Go ahead and delete these specific references to clients. "The API version used in all service requests."

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,364 @@
...
+        /// Thrown if IoT hub responded to the request with a non-successful status code. For example, if the provided
+        /// request was throttled, <see cref="IotHubThrottledException"/> is thrown. For a complete list of possible
+        /// error cases, see <see cref="Common.Exceptions"/>.
+        /// </exception>
+        /// <exception cref="HttpRequestException">
+        /// If the HTTP request fails due to an underlying issue such as network connectivity, DNS failure, or server
+        /// certificate validation.
+        /// </exception>
+        /// <exception cref="OperationCanceledException">If the provided <paramref name="cancellationToken"/> has requested cancellation.</exception>
+        public virtual async Task<JobResponse> GetAsync(string jobId, CancellationToken cancellationToken = default)
```
timtay-microsoft, line 70: I don't love the name "JobResponse" in this context. Maybe "ScheduledJob" so that the API reads like:

```csharp
ScheduledJob scheduledJob = await serviceClient.ScheduledJobs.GetAsyc(...);
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,364 @@
...
+        /// Thrown if IoT hub responded to the request with a non-successful status code. For example, if the provided
+        /// request was throttled, <see cref="IotHubThrottledException"/> is thrown. For a complete list of possible
+        /// error cases, see <see cref="Common.Exceptions"/>.
+        /// </exception>
+        /// <exception cref="HttpRequestException">
+        /// If the HTTP request fails due to an underlying issue such as network connectivity, DNS failure, or server
+        /// certificate validation.
+        /// </exception>
+        /// <exception cref="OperationCanceledException">If the provided <paramref name="cancellationToken"/> has requested cancellation.</exception>
+        public virtual async Task<JobResponse> ScheduleDeviceMethodAsync(ScheduledDeviceMethod scheduledDeviceMethod, ScheduledJobsOptions scheduledJobsOptions, CancellationToken cancellationToken = default)
```
timtay-microsoft, line 185:
```suggestion
        public virtual async Task<JobResponse> ScheduleDirectMethodAsync(ScheduledDirectMethod scheduledDirectMethod, ScheduledJobsOptions scheduledJobsOptions, CancellationToken cancellationToken = default)
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,364 @@
...
+            }
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Enter(this, $"queryCondition=[{scheduledTwinUpdate.QueryCondition}]", nameof(ScheduleDeviceMethodAsync));
+            }
+
+        }
+
+        private async Task<JobResponse> CreateAsync(JobRequest jobRequest, CancellationToken cancellationToken)
```
timtay-microsoft, line 281: Having a private function to house what is basically 4 lines of code seems unnecessary to me. Especially since it is only being called from 2 places. I'd move the HTTP operations into the functions that call this one and delete this private function

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,364 @@
...
+            {
+                Argument.RequireNotNull(scheduledTwinUpdate, nameof(scheduledTwinUpdate));
+                Argument.RequireNotNullOrEmpty(scheduledTwinUpdate.QueryCondition, nameof(scheduledTwinUpdate.QueryCondition));
+                Argument.RequireNotNull(scheduledTwinUpdate.Twin, nameof(scheduledTwinUpdate.Twin));
+                Argument.RequireNotNull(scheduledTwinUpdate.StartTimeUtc, nameof(scheduledTwinUpdate.StartTimeUtc));
+                cancellationToken.ThrowIfCancellationRequested();
+
+                var jobRequest = new JobRequest
+                {
+                    JobId = string.IsNullOrWhiteSpace(scheduledJobsOptions.JobId) ? "job_" + Guid.NewGuid().ToString() : scheduledJobsOptions.JobId,
```
timtay-microsoft, line 257:
```suggestion
                    JobId = string.IsNullOrWhiteSpace(scheduledJobsOptions.JobId) ?  Guid.NewGuid().ToString() : scheduledJobsOptions.JobId,
```

The extra job prefix isn't necessary since a GUID is already unique. It would just be extra bits on the wire to make it human readable.

iothub/service/src/Jobs/ScheduledJobsClient.cs
timtay-microsoft: You still need to add ```.ConfigureAwait(false)``` after every async call that gets awaited @tmahmood-microsoft 

Here and everywhere else

```suggestion
                HttpResponseMessage response = await _httpClient.SendAsync(request, cancellationToken).ConfigureAwait(false);
```

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -96,13 +94,21 @@ public async Task JobClient_Http_SasCredentialAuth_Success()
...
+                ScheduledTwinUpdate twinUpdate = new ScheduledTwinUpdate
+                {
+                    QueryCondition = query,
+                    Twin = twin,
+                    StartTimeUtc = DateTime.UtcNow
+                };
+                ScheduledJobsOptions scheduledTwinUpdateOptions = new ScheduledJobsOptions
+                {
+                    JobId = jobId,
+                    MaxExecutionTimeInSeconds = (long)TimeSpan.FromMinutes(2).TotalSeconds
```
drwill-ms, line 106: Can we change this property to be a TimeSpan instead of long with a specified unit of seconds?

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -96,13 +94,21 @@ public async Task JobClient_Http_SasCredentialAuth_Success()
...
             {
                 // act
-                JobResponse createJobResponse = await jobClient
+                ScheduledTwinUpdate twinUpdate = new ScheduledTwinUpdate
+                {
+                    QueryCondition = query,
+                    Twin = twin,
+                    StartTimeUtc = DateTime.UtcNow
+                };
+                ScheduledJobsOptions scheduledTwinUpdateOptions = new ScheduledJobsOptions
```
drwill-ms, line 103: var when the type is obvious (in this case we call `new ScheduledJobOptions` so it is obvious.

```suggestion
                var scheduledTwinUpdateOptions = new ScheduledJobsOptions
```

e2e/test/iothub/service/IoTHubServiceProxyE2ETests.cs
```diff
@@ -88,17 +88,29 @@ private async Task JobClient_ScheduleAndRunTwinJob(HttpTransportSettings httpTra
                 Tags = new TwinCollection(),
             };
             twin.Tags[JobTestTagName] = JobDeviceId;
-
-            using var jobClient = JobClient.CreateFromConnectionString(s_connectionString, httpTransportSettings);
+            IotHubServiceClientOptions options = new IotHubServiceClientOptions();
+            options.Proxy = httpTransportSettings.Proxy;
```
drwill-ms, line 92: Use an initializer list.

```suggestion
            IotHubServiceClientOptions options = new IotHubServiceClientOptions
            {
                Proxy = httpTransportSettings.Proxy,
            }
```

iothub/service/src/Jobs/ScheduledDeviceMethod.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Query condition to evaluate which devices to run the job on.
+        /// </summary>
+        [JsonProperty(PropertyName = "QueryCondition", Required = Required.Always)]
```
drwill-ms, line 14: I'm assuming the JSON property name should be camelCased, and the service JSON is always (AFAIK) cased as such.

e2e/test/iothub/service/IoTHubCertificateValidationE2ETest.cs
```diff
@@ -61,14 +61,21 @@ private static async Task TestServiceClientInvalidServiceCertificate(TransportTy
         [LoggedTestMethod]
         public async Task JobClient_ScheduleTwinUpdateInvalidServiceCertificateHttp_Fails()
         {
-            using var jobClient = JobClient.CreateFromConnectionString(TestConfiguration.IoTHub.ConnectionStringInvalidServiceCertificate);
+            using var sc = new IotHubServiceClient(TestConfiguration.IoTHub.ConnectionStringInvalidServiceCertificate);
+            ScheduledTwinUpdate twinUpdate = new ScheduledTwinUpdate
```
drwill-ms, line 65:
```suggestion
            var twinUpdate = new ScheduledTwinUpdate
```

drwill-ms: also below

iothub/service/src/Jobs/JobRequest.cs
```diff
@@ -26,39 +26,52 @@ public class JobRequest
...
 
 
         /// <summary>
-        /// ISO 8601 date time to start the job
+        /// ISO 8601 date time to start the job.
         /// </summary>
         [JsonProperty(PropertyName = "startTime")]
         [JsonConverter(typeof(IsoDateTimeConverter))]
         // TODO: siport : validate if we want utc on the internal propname but not on the external
         public DateTime StartTime { get; set; }
```
drwill-ms, line 62: Let's rename to StartTimeUtc

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,341 @@
...
+        private static readonly TimeSpan s_defaultOperationTimeout = TimeSpan.FromSeconds(100);
+
+        /// <summary>
+        /// Creates client, provided for unit testing purposes only.
+        /// </summary>
+        protected ScheduledJobsClient()
+        {
+        }
+
+        internal ScheduledJobsClient(string hostName, IotHubConnectionProperties credentialProvider, HttpClient httpClient, HttpRequestMessageFactory httpRequestMessageFactory)
```
drwill-ms, line 44: When lines get long like this, we want to split them up. In this case, we do it by parameters.

```suggestion
        internal ScheduledJobsClient(
            string hostName,
            IotHubConnectionProperties credentialProvider,
            HttpClient httpClient,
            HttpRequestMessageFactory httpRequestMessageFactory)
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,341 @@
...
+        /// If the HTTP request fails due to an underlying issue such as network connectivity, DNS failure, or server
+        /// certificate validation.
+        /// </exception>
+        public virtual IQuery CreateQuery(JobType? jobType = null, JobStatus? jobStatus = null, int? pageSize = null)
+        {
+            return new Query(async (token) => await GetAsync(jobType, jobStatus, pageSize, token, CancellationToken.None).ConfigureAwait(false));
+        }
+
+        /// <summary>
+        /// Cancels/Deletes the job with the specified Id.
```
drwill-ms, line 121:
```suggestion
        /// Cancels/deletes the job with the specified Id.
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,341 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Creates a new Job to update twin tags and desired properties on one or multiple devices.
+        /// </summary>
+        /// <param name="scheduledTwinUpdate">Required parameters for scheduled twin update, i.e: <paramref name="scheduledTwinUpdate.Twin"/>, <paramref name="scheduledTwinUpdate.QueryCondition"/>, <paramref name="scheduledTwinUpdate.StartTimeUtc"/></param>
+        /// <param name="scheduledJobsOptions">Optional parameters for scheduled twin update, i.e: <paramref name="scheduledJobsOptions.JobId"/> and <paramref name="scheduledJobsOptions.MaxExecutionTimeInSeconds"/></param>
+        /// <param name="cancellationToken">Task cancellation token.</param>
+        /// <returns>A JobResponse object.</returns>\
```
drwill-ms, line 230: Trail backslash here

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,341 @@
...
+                if (Logging.IsEnabled)
+                    Logging.Error(this, $"{nameof(ScheduleTwinUpdateAsync)} threw an exception: {ex}", nameof(ScheduleTwinUpdateAsync));
+                throw;
+            }
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Enter(this, $"queryCondition=[{scheduledTwinUpdate.QueryCondition}]", nameof(ScheduleDeviceMethodAsync));
+            }
+
```
drwill-ms, line 280: extra newline here can be removed

iothub/service/src/Jobs/ScheduledJobsOptions.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    ///
```
drwill-ms, line 7: Fill in

iothub/service/src/Jobs/ScheduledJobsOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class ScheduledJobsOptions
+    {
+        /// <summary>
+        /// Unique Job Id for the twin update job.
```
drwill-ms, line 12: lowercase j in job.

```suggestion
        /// Unique job Id for the twin update job.
```

iothub/service/src/Query/IQuery.cs
```diff
@@ -52,17 +52,17 @@ public interface IQuery
         Task<QueryResponse<DeviceJob>> GetNextAsDeviceJobAsync(QueryOptions options);
 
         /// <summary>
-        /// Retrieves the next paged result as <see cref="JobResponse"/> objects
+        /// Retrieves the next paged result as <see cref="ScheduledJob"/> objects
```
drwill-ms, line 55: Type reference not required as the API already refers to it.

```suggestion
        /// Retrieves the next paged result as scheduled jobs.
```

drwill-ms: Also below.

iothub/service/src/Query/Query.cs
```diff
@@ -72,18 +72,18 @@ public async Task<QueryResponse<DeviceJob>> GetNextAsDeviceJobAsync(QueryOptions
         ///     fetch the next paged result as job responses
         /// </summary>
         /// <returns></returns>
```
drwill-ms, line 74: Fill in.

iothub/service/tests/JobClient/JobClientTests.cs
```diff
@@ -19,68 +20,62 @@ namespace Microsoft.Azure.Devices.Api.Test
     public class JobClientTests
     {
         private readonly string jobId = "testJobId";
-        private readonly JobResponse expectedJobResponse = new JobResponse();
+        private readonly ScheduledJob scheduledJob = new ScheduledJob();
```
drwill-ms, line 23:
```suggestion
        private readonly ScheduledJob scheduledJob = new();
```

iothub/service/tests/JobClient/JobClientTests.cs
```diff
@@ -19,68 +20,62 @@ namespace Microsoft.Azure.Devices.Api.Test
...
         private readonly TimeSpan timeout = TimeSpan.FromMinutes(1);
+        private const string HostName = "acme.azure-devices.net";
+        private static Uri HttpUri = new Uri("https://" + HostName);
```
drwill-ms, line 26:
```suggestion
        private static Uri HttpUri = new("https://" + HostName);
```

iothub/service/src/Configurations/ConfigurationsClient.cs
```diff
@@ -42,12 +44,12 @@ internal ConfigurationsClient(string hostName, IotHubConnectionProperties creden
         }
 
         /// <summary>
-        /// Creates a new configuration for Azure IoT Edge in IoT hub
+        /// Creates a new <see cref="Configuration"/> for Azure IoT Edge in IoT hub.
```
drwill-ms, line 47: As the API definition includes a reference to this type, this reference here is unnecessary. I prefer the plain text that it was before. Same elsewhere.

iothub/service/src/Jobs/JobRequest.cs
```diff
@@ -26,39 +26,52 @@ public class JobRequest
...
+        /// </remarks>
         [JsonProperty(PropertyName = "cloudToDeviceMethod")]
         public CloudToDeviceMethod CloudToDeviceMethod { get; set; }
 
         /// <summary>
-        /// Required if jobType is updateTwin.
         /// The Update Twin tags and desired properties.
         /// </summary>
+        /// <remarks>
+        /// Required if jobType is updateTwin.
```
drwill-ms, line 41: jobType is the serialized type.

There are a few ways to approach this. As this is referencing an actual property that isn't otherwise related with this one, I'd see a link

```suggestion
        /// Required if <see cref="JobType"/> is <see cref="JobType.UpdateTwin"/>.
```

Alternatively, you could just use plain English, so it'd be "Required if the job type is 'UpdateTwin'" (or even <c>UpdateTwin</c>).

drwill-ms: Same for CloudToDeviceMethod.

drwill-ms: And below.

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,349 @@
...
+        /// <summary>
+        /// Creates client, provided for unit testing purposes only.
+        /// </summary>
+        protected ScheduledJobsClient()
+        {
+        }
+
+        internal ScheduledJobsClient(string hostName, IotHubConnectionProperties credentialProvider,
+            HttpClient httpClient, HttpRequestMessageFactory httpRequestMessageFactory)
```
drwill-ms, line 45: We do all params or none.

```suggestion
        internal ScheduledJobsClient(
            string hostName,
            IotHubConnectionProperties credentialProvider,
            HttpClient httpClient,
            HttpRequestMessageFactory httpRequestMessageFactory)
```

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,349 @@
...
+        {
+            _hostName = hostName;
+            _credentialProvider = credentialProvider;
+            _httpClient = httpClient;
+            _httpRequestMessageFactory = httpRequestMessageFactory;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="ScheduledJob"/> with the specified Id.
```
drwill-ms, line 54: More cases of adding crefs where not necessary because the API surface includes a link to the type.

iothub/service/src/Query/Query.cs
```diff
@@ -35,7 +35,7 @@ internal Query(Func<string, Task<QueryResult>> queryTaskFunc)
         /// <summary>
         ///     fetch the next paged result as twins
         /// </summary>
-        /// <returns></returns>
+        /// <returns>List of <see cref="Twin"/>.</returns>
```
drwill-ms, line 38: Same, unnecessary links.

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,352 @@
...
+        /// Thrown if IoT hub responded to the request with a non-successful status code. For example, if the provided
+        /// request was throttled, <see cref="IotHubThrottledException"/> is thrown. For a complete list of possible
+        /// error cases, see <see cref="Common.Exceptions"/>.
+        /// </exception>
+        /// <exception cref="HttpRequestException">
+        /// If the HTTP request fails due to an underlying issue such as network connectivity, DNS failure, or server
+        /// certificate validation.
+        /// </exception>
+        /// <exception cref="OperationCanceledException">If the provided <paramref name="cancellationToken"/> has requested cancellation.</exception>
+        public virtual async Task<ScheduledJob> ScheduleDeviceMethodAsync(ScheduledDirectMethod scheduledDirectMethod,
```
timtay-microsoft, line 195:
```suggestion
        public virtual async Task<ScheduledJob> ScheduleDirectMethodAsync(ScheduledDirectMethod scheduledDirectMethod,
```

timtay-microsoft (approved):
One final comment, but otherwise looks good!

iothub/service/src/Jobs/JobRequest.cs
```diff
@@ -26,39 +26,52 @@ public class JobRequest
...
         [JsonProperty(PropertyName = "queryCondition")]
         public string QueryCondition { get; set; }
 
 
         /// <summary>
-        /// ISO 8601 date time to start the job
+        /// ISO 8601 date time to start the job.
         /// </summary>
-        [JsonProperty(PropertyName = "startTime")]
+        [JsonProperty(PropertyName = "startTimeUtc")]
```
drwill-ms, line 59: You can't change the serialized value. The service still expects `startTime`.

The only rename was meant to be to the .NET property.

iothub/service/src/Jobs/ScheduledJobsClient.cs
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        /// <summary>
+        /// Gets <see cref="IQuery"/> through which job responses for specified jobType and jobStatus are retrieved page by page,
+        /// and specify page size.
+        /// </summary>
+        /// <param name="jobType">The job type to query. Could be null if not querying.</param>
+        /// <param name="jobStatus">The job status to query. Could be null if not querying.</param>
+        /// <param name="pageSize">Number of job responses in a page.</param>
+        /// <returns>A <see cref="IQuery"/> object to get results and next pages.</returns>
```
drwill-ms, line 109: Don't need to link to IQuery here.

PR #2540, opened 2022-08-02
Make MQTT the default transport

drwill-ms:
Lorenzo supports making MQTT the default, and suggested we move ahead with that decision.

### DIFF
--- a/iothub/device/tests/Edge/EdgeModuleClientFactoryTest.cs
+++ b/iothub/device/tests/Edge/EdgeModuleClientFactoryTest.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Azure.Devices.Client.Test.Edge
         public async Task TestCreate_FromConnectionStringEnvironment_ShouldCreateClientWithOptions()
         {
             // setup
-            var clientOptions = new IotHubClientOptions
+            var clientOptions = new IotHubClientOptions(new IotHubClientAmqpSettings())
             {
                 ModelId = "tempModuleId"
             };
--- a/iothub/device/tests/Transport/Amqp/MockableAmqpUnit.cs
+++ b/iothub/device/tests/Transport/Amqp/MockableAmqpUnit.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.Amqp;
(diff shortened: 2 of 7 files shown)


iothub/device/tests/Edge/EdgeModuleClientFactoryTest.cs
```diff
@@ -45,7 +45,7 @@ public async Task TestCreate_FromConnectionStringEnvironment_ShouldCreateClient(
         public async Task TestCreate_FromConnectionStringEnvironment_ShouldCreateClientWithOptions()
         {
             // setup
-            var clientOptions = new IotHubClientOptions
+            var clientOptions = new IotHubClientOptions(new IotHubClientAmqpSettings())
```
drwill-ms, line 48: Chose AMQP given this is an Edge module.

iothub/device/tests/Transport/Amqp/MockableAmqpUnit.cs
```diff
@@ -13,17 +14,19 @@ namespace Microsoft.Azure.Devices.Client.Test.Transport
 {
     internal class MockableAmqpUnit : AmqpUnit
     {
+        private static IotHubClientAmqpSettings s_transportSettings = new();
```
drwill-ms, line 17: Need to explicitly initialize AMQP for this mock, and I changed the init below to all refer to the same instance.

iothub/device/tests/Transport/Mqtt/MqttIotHubAdapterTest.cs
```diff
@@ -111,7 +111,7 @@ public void TestAuthenticationChain()
             var passwordProvider = new Mock<IAuthorizationProvider>();
             var mqttIotHubEventHandler = new Mock<IMqttIotHubEventHandler>();
             var productInfo = new ProductInfo();
-            var options = new IotHubClientOptions();
+            var options = new IotHubClientOptions(new IotHubClientMqttSettings());
```
drwill-ms, line 114: Oops, these tests are supposed to be for MQTT, but without being explicit they were testing AMQP before. Fixing that, even though the default is now MQTT.

timtay-microsoft (approved):

andyk-ms (approved):

PR #2543, opened 2022-08-02
refactor(prov-dev): Add ProvisioningClientOptions and move TlsVersions

brycewang-microsoft:
+ Add `ProvisioningClientOptions` to provisioning device client
+ Move TlsVersions properties to `ProvisioningTransportHandler`


provisioning/device/src/ProvisioningClientOptions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client
+{
+    /// <summary>
+    /// Options that allow configuration of the Provisioning device client instance during initialization.
```
drwill-ms, line 8:
```suggestion
    /// Options that allow configuration of the provisioning device client instance during initialization.
```

provisioning/device/src/ProvisioningClientOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    public class ProvisioningClientOptions
+    {
+        /// <summary>
+        /// Creates an instances of this class with the default transport settings.
+        /// </summary>
+        public ProvisioningClientOptions()
+        {
+            ProvisioningTransportSettings = new ProvisioningTransportHandlerAmqp();
```
drwill-ms, line 17: As I just changed the hub device client default to MQTT, can we change the default here too?

provisioning/device/src/Transports/Mqtt/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -185,7 +185,7 @@ protected override void Dispose(bool disposing)
             X509Certificate2 clientCertificate = ((AuthenticationProviderX509)message.Authentication).GetAuthenticationCertificate();
 
             var tlsSettings = new ClientTlsSettings(
-                TlsVersions.Instance.Preferred,
+                this.SslProtocols,
```
drwill-ms, line 188: We don't use `this` unless absolutely necessary, and there's almost always a way to avoid using it.

```suggestion
                SslProtocols,
```

provisioning/device/src/ProvisioningClientOptions.cs
drwill-ms: I think it is more important to be inline with the protocol the hub client uses, considering some customers will need to open up firewall ports and such.

We know MQTT isn't as rich as AMQP, even in hub. The whole team is pushing to improve the MQTT experience, so I think we just need to keep pushing it.

drwill-ms (approved):
I think you should change it to MQTT, but I wouldn't let that hold up this PR.

PR #2544, opened 2022-08-02
Fix NPE issue in DevicesClient unit tests

timtay-microsoft:
Mocked http response had no defined payload so the deserialization logic threw an NPE on some platforms


abhipsaMisra (approved):
Any idea why this was throwing an exception only for certain frameworks?

timtay-microsoft:
> Any idea why this was throwing an exception only for certain frameworks?

It must have something to do with how our mocking framework handles unassigned properties on a mocked object. I'm guessing .NET 5/6 give it a default non-null value whereas other frameworks don't

PR #2548, opened 2022-08-04
refactor(iot-device): Remove DeviceIdentity and add the information to IotHubConnectionInfo

abhipsaMisra:
Rework of #2538


iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -3,22 +3,26 @@
...
+        public IotHubConnectionInfo(
+            IotHubConnectionStringBuilder builder,
+            IotHubClientOptions iotHubClientOptions)
         {
             if (builder == null)
             {
                 throw new ArgumentNullException(nameof(builder));
             }
 
+            builder.Validate();
```
abhipsaMisra, line 24: We were previously used the following to create an IotHubConnectionInfo instance. Since this was called from an IotHubConnectionStringBuilder instance, this will not have the ClientOptions. The important bit in the method call below, however, is the call to Validate. I've moved it to the constructor.

```csharp
        internal IotHubConnectionInfo ToIotHubConnectionInfo()
        {
            Validate();
            return new IotHubConnectionInfo(this);
        }
```

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -30,9 +35,7 @@ public IotHubConnectionInfo(IotHubConnectionStringBuilder builder)
             DeviceId = builder.DeviceId;
             ModuleId = builder.ModuleId;
 
-            HttpsEndpoint = new UriBuilder(Uri.UriSchemeHttps, HostName).Uri;
-
-            AmqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, HostName, DefaultAmqpSecurePort).Uri;
```
abhipsaMisra, line 35: Moved them to the transport specific classes

iothub/device/src/Transport/Http/HttpTransportHandler.cs
```diff
@@ -62,18 +62,18 @@ internal sealed class HttpTransportHandler : TransportHandler
...
             bool isClientPrimaryTransportHandler = false)
             : base(context, transportSettings)
         {
-            ProductInfo productInfo = context.ProductInfo;
-            _deviceId = iotHubConnectionString.DeviceId;
-            _moduleId = iotHubConnectionString.ModuleId;
+            ProductInfo productInfo = context.IotHubConnectionInfo.ClientOptions.ProductInfo;
+            _deviceId = context.IotHubConnectionInfo.DeviceId;
+            _moduleId = context.IotHubConnectionInfo.ModuleId;
+            Uri httpsEndpoint = new UriBuilder(Uri.UriSchemeHttps, context.IotHubConnectionInfo.HostName).Uri;
```
abhipsaMisra, line 73: Moved from IotHubConnectionInfo:

```csharp
HttpsEndpoint = new UriBuilder(Uri.UriSchemeHttps, HostName).Uri;
```

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -13,23 +13,25 @@ namespace Microsoft.Azure.Devices.Client.Transport.Amqp
...
+        internal AmqpAuthenticationRefresher(IClientIdentity clientIdentity, AmqpIotCbsLink amqpCbsLink)
         {
             _amqpIotCbsLink = amqpCbsLink;
-            _connInfo = deviceIdentity.IotHubConnectionInfo;
-            _audience = deviceIdentity.Audience;
-            _amqpIotCbsTokenProvider = new AmqpIotCbsTokenProvider(_connInfo);
+            _clientIdentity = clientIdentity;
+            _audience = clientIdentity.AmqpCbsAudience;
+            _amqpIotCbsTokenProvider = new AmqpIotCbsTokenProvider(_clientIdentity);
+            _amqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, clientIdentity.HostName, CommonConstants.DefaultAmqpSecurePort).Uri;
```
abhipsaMisra, line 30: Moved from IotHubConnectionInfo:

```csharp
AmqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, HostName, DefaultAmqpSecurePort).Uri;
```

iothub/device/src/Transport/AmqpIot/AmqpIotSession.cs
```diff
@@ -342,23 +344,28 @@ internal bool IsClosing()
...
                     deviceTemplate,
-                    WebUtility.UrlEncode(deviceIdentity.IotHubConnectionInfo.DeviceId))
+                    WebUtility.UrlEncode(clientIdentity.DeviceId))
                 : string.Format(
                     CultureInfo.InvariantCulture,
                     moduleTemplate,
-                    WebUtility.UrlEncode(deviceIdentity.IotHubConnectionInfo.DeviceId), WebUtility.UrlEncode(deviceIdentity.IotHubConnectionInfo.ModuleId));
+                    WebUtility.UrlEncode(clientIdentity.DeviceId), WebUtility.UrlEncode(clientIdentity.ModuleId));
 
-            return deviceIdentity.IotHubConnectionInfo.BuildLinkAddress(path).AbsoluteUri;
```
abhipsaMisra, line 361: Moved from IotHubConnectionInfo:

```csharp
public Uri BuildLinkAddress(string path)
{
	var builder = new UriBuilder(AmqpEndpoint)
	{
		Path = path,
	};

	return builder.Uri;
}
```

iothub/device/src/IotHubClientOptions.cs
```diff
@@ -89,6 +89,12 @@ public IotHubClientOptions(IotHubClientTransportSettings transportSettings)
...
         /// Stores custom product information that will be appended to the user agent string that is sent to IoT hub.
         /// </summary>
-        public string ProductInfo { get; set; }
+        public string CustomProductInfo
+        { 
+            get => ProductInfo.Extra;
+            set => ProductInfo.Extra = value;
+        }
+
+        internal ProductInfo ProductInfo { get; } = new ProductInfo();
```
abhipsaMisra, line 98: This internal property is required because the public one stores only the extra custom text that is appended to the SDK defined user-agent string (ProductInfo).

iothub/device/src/Pipeline/PipelineContext.cs
```diff
@@ -7,14 +7,8 @@ namespace Microsoft.Azure.Devices.Client
 {
     internal class PipelineContext
     {
-        internal IotHubClientTransportSettings TransportSettings { get; set; }
```
abhipsaMisra, line 10: PipelineContext holds IotHubConnectionInfo, which in turn holds ClientOptions. TransportSettings and ProductInfo are a part of ClientOptions. So, we don't need PipelineContext to hold them explicitly.

iothub/device/src/Authentication/IClientIdentity.cs
```diff
@@ -0,0 +1,88 @@
...
+    internal interface IClientIdentity
+    {
+        /// <summary>
+        /// Authentication method associated with this client that uses a shared access signature token
+        /// and allows for token refresh.
+        /// </summary>
+        AuthenticationWithTokenRefresh TokenRefresher { get; }
+
+        /// <summary>
+        /// The IoT Hub this client connects to.
```
drwill-ms, line 20:
```suggestion
        /// The IoT hub this client connects to.
```

iothub/device/src/Authentication/IClientIdentity.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// </summary>
+        string IotHubName { get; }
+
+        /// <summary>
+        /// The device ID associated with this client.
```
drwill-ms, line 25:
```suggestion
        /// The device Id associated with this client.
```

iothub/device/src/Pipeline/TransportHandlerFactory.cs
```diff
@@ -12,49 +12,43 @@ internal class TransportHandlerFactory : ITransportHandlerFactory
...
 
-            if (transportSettings is IotHubClientAmqpSettings)
+            if (connInfo.ClientOptions.TransportSettings is IotHubClientAmqpSettings iotHubClientAmqpSettings)
             {
                 return new AmqpTransportHandler(
                     context,
-                    connInfo,
-                    transportSettings as IotHubClientAmqpSettings,
+                    iotHubClientAmqpSettings,
                     new Func<MethodRequestInternal, Task>(onMethodCallback),
```
abhipsaMisra, line 26: PipelineContext holds the callback definitions as well, so they can also be removed from this method signature. However, the delegate -> function transformation needs to be done first, so I'll do that in the next PR.

iothub/device/src/Authentication/IClientIdentity.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// The host service that this client connects to.
+        /// This can either be the IoT hub name or a gateway service name.
+        /// </summary>
+        string HostName { get; }
+
+        // TODO (abmisr): Consolidate with AmqpCbsAudience
+        /// <summary>
+        /// This currently points to the hostname.
+        /// This needs to store the formatted audience (hostname/devices/deviceId)
+        /// </summary>
```
drwill-ms, line 44:
```suggestion
        /// This currently points to the host name.
        /// </summary>
        /// <remarks>
        /// This needs to store the formatted audience (hostname/devices/deviceId)
        /// </remarks>
```

iothub/device/src/Authentication/IClientIdentity.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// The shared access key used to connect to the IoT hub service.
+        /// </summary>
+        string SharedAccessKey { get; }
+
+        /// <summary>
+        /// The shared access signature used to connect to the IoT hub service.
+        /// </summary>
+        string SharedAccessSignature { get; }
```
drwill-ms, line 60: I wonder if we should have a sub-type to hold all 3 of these things, since they go together.

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -97,58 +100,54 @@ public IotHubConnectionInfo(IotHubConnectionStringBuilder builder)
...
+        public IotHubClientOptions ClientOptions { get; }
 
-        public string SharedAccessSignature { get; private set; }
+        // TODO (abmisr): Consolidate with Audience
+        public string AmqpCbsAudience { get; }
 
-        public bool IsUsingGateway { get; private set; }
+        public bool IsPooling()
+        {
+
```
drwill-ms, line 146: extra newline here can be removed

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -97,58 +100,54 @@ public IotHubConnectionInfo(IotHubConnectionStringBuilder builder)
...
+            return AuthenticationModel != AuthenticationModel.X509
+                && ClientOptions.TransportSettings is IotHubClientAmqpSettings iotHubClientAmqpSettings1
```
drwill-ms, line 148: Why iotHubClientAmqpSettings**1**?

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -178,14 +177,63 @@ async Task<string> IAuthorizationProvider.GetPasswordAsync()
             }
         }
 
-        public Uri BuildLinkAddress(string path)
+        public override bool Equals(object obj)
```
drwill-ms, line 180: Is this method even needed? I found one on another object that had no references, so I assumed it was created with no applicable reason in mind.

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -178,14 +177,63 @@ async Task<string> IAuthorizationProvider.GetPasswordAsync()
...
+        private static int UpdateHashCode(int hashCode, object field)
+        {
+            return field == null
+                ? hashCode
+                : hashCode * -1521134295 + field.GetHashCode();
+        }
+
+        private string CreateAmqpCbsAudience()
+        {
+            // If the shared access key name is null then this is an individual sas authenticated client.
```
drwill-ms, line 219:
```suggestion
            // If the shared access key name is null then this is an individual SAS authenticated client.
```

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -178,14 +177,63 @@ async Task<string> IAuthorizationProvider.GetPasswordAsync()
...
+            // SAS tokens granted to an individual sas authenticated client will be scoped to an individual device; for example, myHub.azure-devices.net/devices/device1.
```
drwill-ms, line 220:
```suggestion
            // SAS tokens granted to an individual SAS authenticated client will be scoped to an individual device; for example, myHub.azure-devices.net/devices/device1.
```

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -178,14 +177,63 @@ async Task<string> IAuthorizationProvider.GetPasswordAsync()
...
+                {
+                    clientAudience += $"/modules/{WebUtility.UrlEncode(ModuleId)}";
+                }
 
-            return builder.Uri;
+                return clientAudience;
+            }
+            else
+            {
+                // If the shared access key name is not null then this is a group sas authenticated client.
```
drwill-ms, line 233:
```suggestion
                // If the shared access key name is not null then this is a group SAS authenticated client.
```

iothub/device/src/Authentication/IotHubConnectionInfo.cs
```diff
@@ -178,14 +177,63 @@ async Task<string> IAuthorizationProvider.GetPasswordAsync()
...
+                // SAS tokens granted to a group sas authenticated client will scoped to the IoT hub-level; for example, myHub.azure-devices.net
```
drwill-ms, line 234:
```suggestion
                // SAS tokens granted to a group SAS authenticated client will scoped to the IoT hub-level; for example, myHub.azure-devices.net
```

iothub/device/src/IotHubClientOptions.cs
drwill-ms, line 98:
```suggestion
        internal ProductInfo ProductInfo { get; } = new();
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1118,7 +1116,9 @@ void OnTwinResponse(Message possibleResponse)
                      settings.SslProtocols,
                      settings.CertificateRevocationCheck,
                      certs,
-                     iotHubConnectionString.HostName);
+                     clientIdentity.HostName);
+
+                var iotHubConnectionInfo = clientIdentity as IotHubConnectionInfo;
```
drwill-ms, line 1121: What if it isn't and this comes back null?

iothub/device/src/Authentication/IClientIdentity.cs
abhipsaMisra: That's an idea. We should revisit that when we rework the auth classes.

iothub/device/tests/Transport/Http/HttpTransportHandlerTests.cs
```diff
@@ -91,7 +91,12 @@ public void HttpTransportHandler_DisableReceiveMessageAsync_ThrowsNotSupportedEx
 
         HttpTransportHandler CreateFromConnectionString()
         {
-            return new HttpTransportHandler(new PipelineContext(), IotHubConnectionInfoExtensions.Parse(DumpyConnectionString), new IotHubClientHttpSettings());
+            return new HttpTransportHandler(
+                new PipelineContext
+                {
+                    IotHubConnectionInfo = new IotHubConnectionInfo(new IotHubConnectionStringBuilder(DumpyConnectionString), new IotHubClientOptions(new IotHubClientHttpSettings())),
```
drwill-ms, line 97: Haha dumpy?

iothub/device/src/Authentication/IotHubConnectionInfo.cs
abhipsaMisra: Ah, I was reworking the logic, and this was to avoid conflict with the existing logic. Updated.

drwill-ms (approved):

iothub/device/src/Authentication/IotHubConnectionInfo.cs
abhipsaMisra: I believe this particular one is being referenced in our connection pooling logic to resolve the identity of the AmqpUnit to be removed from the pool. I can add a comment calling that out.

iothub/device/tests/Transport/Http/HttpTransportHandlerTests.cs
abhipsaMisra: iirc we used to call these "dummy" but the code analysis tools kept complaining.

iothub/device/src/Authentication/IClientIdentity.cs
```diff
@@ -0,0 +1,88 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.Devices.Client.Transport;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Interface for device configurations and common attributes.
+    /// </summary>
+    internal interface IClientIdentity
```
abhipsaMisra, line 11: ClientConfiguration?

abhipsaMisra: Note about testing and interface/ abstract class

iothub/device/tests/Transport/Http/HttpTransportHandlerTests.cs
drwill-ms: The word you're looking for is `Fake`.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: That's a good point. I'll update the IClientConfiguration interface to reflect all the properties and methods that any class down the pipeline would expect it to have. We won't need to cast it, then.

PR #2549, opened 2022-08-04
Re-organize Direct Methods APIs a bit

timtay-microsoft:
Before
```csharp
public virtual async Task<CloudToDeviceMethodResult> InvokeAsync(string deviceId, CloudToDeviceMethod cloudToDeviceMethod, CancellationToken cancellationToken = default)
```

After
```csharp
public virtual async Task<DirectMethodResponse> InvokeAsync(string deviceId, string methodName, DirectMethodRequestOptions options = null, CancellationToken cancellationToken = default)
```

This change gives us a minimal method invocation call that looks like

```csharp
DirectMethodResponse response = serviceClient.DirectMethods.InvokeAsync("myDevice", "myMethodName");
```

And moves the payload and connect/response timeouts into an options bundle

The one big problem here is that we will still need to expose a DirectMethodRequest object (containing a mix of required and optional parameters) outside of this subclient such as in the [GetScheduledJob API](https://docs.microsoft.com/en-us/rest/api/iothub/service/jobs/get-scheduled-job#cloudtodevicemethod). My solution to this is to still expose this DirectMethodRequest json serialization type, but only make getters public. That way, users can't create a DirectMethodRequest themselves, but they can still get all the context given to the SDK in APIs like GetScheduledJobs


iothub/service/src/DirectMethod/DirectMethodRequestOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Optional parameters to execute a direct method with.
+    /// </summary>
+    public class DirectMethodRequestOptions
+    {
+        /// <summary>
+        /// The timeout (in seconds) before the direct method request will fail if the device fails to respond to the request.
+        /// This timeout may happen if the target device is slow in handling the direct method.
+        /// </summary>
+        public int? ResponseTimeout { get; set; }
```
timtay-microsoft, line 15: Note that we used to accept this as a timespan. I'm in favor of using int? here instead because the service has pretty strict limits on what this value can be (5-300). On top of that, exposing these as a timespan might falsely convince users that there is a level of granularity beyond seconds. Finally, if users try to pass in something like

```csharp
options.ResponseTimeout = TimeSpan.FromMilliseconds(50);
```

Then the SDK/service will end up interpreting that as 0 seconds and throw an exception that would confuse users.

The typical code for a user would go from

```csharp
options.ResponseTimeout = TimeSpan.FromSeconds(50);
```

to

```csharp
options.ResponseTimeout = 50;
```

iothub/service/src/DirectMethod/DirectMethodRequestOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// This timeout may happen if the target device is not connected to the cloud or if the cloud fails to deliver
+        /// the request to the target device in time. If this value is set to 0 seconds, then the target device must be online
+        /// when this direct method request is made.
+        /// </summary>
+        public int? ConnectionTimeout { get; set; }
+
+        /// <summary>
+        /// The json payload for the request.
+        /// </summary>
+        public string Payload { get; set; }
```
timtay-microsoft, line 28: I want to revisit this payload type later the way we did in Java v2 in a later PR

PR #2551, opened 2022-08-04
MessageFeedbackProcessor, MessagingClient and FileUploadNotificationProcessorClient

tmahmood-microsoft:
## Checklist


iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -59,6 +61,18 @@ public IotHubServiceClientOptions()
         /// </summary>
         public HttpClient HttpClient { get; set; }
 
+        /// <summary>
+        /// Whether to use web sockets or not (Only used for AMQP or MQTT).
+        /// </summary>
```
drwill-ms, line 66:
```suggestion
        /// Whether to use web sockets or not.
        /// </summary>
        /// <remarks>
        /// Only used for AMQP or MQTT.
        /// </remarks>
```

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -59,6 +61,18 @@ public IotHubServiceClientOptions()
...
+        public bool UseWebSocketOnly { get; set; }
+
+        /// <summary>
+        /// Service client transport settings used for AMQP/MQTT operations.
+        /// If provided, all other settings will be ignored. If not provided,
+        /// a transport settings client will be created for you based on the other provided settings.
```
drwill-ms, line 72: Put these into a <remarks> section.

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,104 @@
+﻿using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Subclient of <see cref="IotHubServiceClient"/> for handling cloud to device message feedback.
```
drwill-ms, line 8:
```suggestion
    /// Subclient of <see cref="IotHubServiceClient"/> for handling cloud-to-device message feedback.
```

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// </summary>
+    /// <seealso href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-c2d#message-feedback"/>.
+    public class MessageFeedbackProcessorClient
+    {
+        private readonly string _hostName;
+        private readonly IotHubConnectionProperties _credentialProvider;
+        private readonly IotHubConnection _connection;
+
+        /// <summary>
+        /// Get the <see cref="AmqpFeedbackReceiver"/> which can deliver acknowledgments for messages sent to a device/module from IoT hub.
```
drwill-ms, line 18: Don't link to the type. It is in the property definition and will already be included.

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// </summary>
+        internal AmqpFeedbackReceiver FeedbackReceiver { get; }
```
drwill-ms, line 20: properties after ctors

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,104 @@
...
+            }
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, $"Closing FeedbackReceiver", nameof(CloseAsync));
+            }
+        }
+
+        /// <summary>
+        /// Dispose the FeedbackReceiver instance. This call is made over AMQP.
```
drwill-ms, line 91: No call is made.
```suggestion
        /// Dispose the FeedbackReceiver instance.
```

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// </summary>
+        public virtual void Dispose()
+        {
+            if (Logging.IsEnabled)
+                Logging.Enter(this, $"Disposing FeedbackReceiver", nameof(CloseAsync));
```
drwill-ms, line 96:
```suggestion
                Logging.Enter(this, $"Disposing FeedbackReceiver", nameof(Dispose));
```

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            FeedbackReceiver.Dispose();
+
+            if (Logging.IsEnabled)
+                Logging.Exit(this, $"Disposing FeedbackReceiver", nameof(CloseAsync));
```
drwill-ms, line 101:
```suggestion
                Logging.Exit(this, $"Disposing FeedbackReceiver", nameof(Dispose));
```

iothub/service/src/IotHubServiceClientOptions.cs
timtay-microsoft: There is no MQTT API in the service client

e2e/test/iothub/messaging/MessageReceiveE2EPoolAmqpTests.cs
```diff
@@ -160,45 +160,46 @@ await PoolingOverAmqp
...
-                (Message msg, string payload, string p1Value) = MessageReceiveE2ETests.ComposeC2dTestMessage(Logger);
-
-                await testDeviceCallbackHandler.SetMessageReceiveCallbackHandlerAsync().ConfigureAwait(false);
-                testDeviceCallbackHandler.ExpectedMessageSentByService = msg;
-
-                await serviceClient.SendAsync(testDevice.Id, msg).ConfigureAwait(false);
-            }
-
-            async Task TestOperationAsync(IotHubDeviceClient deviceClient, TestDevice testDevice, TestDeviceCallbackHandler testDeviceCallbackHandler)
+            using (var serviceClient = new IotHubServiceClient(TestConfiguration.IoTHub.ConnectionString))
```
drwill-ms, line 163: Why switch to this style of using?

e2e/test/iothub/service/IoTHubCertificateValidationE2ETest.cs
```diff
@@ -51,11 +51,14 @@ public async Task ServiceClient_SendMessageToDeviceInvalidServiceCertificateAmqp
 
         private static async Task TestServiceClientInvalidServiceCertificate(TransportType transport)
         {
-            using var service = ServiceClient.CreateFromConnectionString(
-                TestConfiguration.IoTHub.ConnectionStringInvalidServiceCertificate,
-                transport);
+            IotHubServiceClientOptions options = new IotHubServiceClientOptions
```
drwill-ms, line 54: Where is the transport being set?

e2e/test/iothub/service/ServiceClientE2ETests.cs
```diff
@@ -71,16 +71,20 @@ private async Task TestTimeout(TimeSpan? timeout)
         public async Task ServiceClient_SendsMessage(TransportType transportType)
         {
             // arrange
+            IotHubServiceClientOptions options = new IotHubServiceClientOptions
```
drwill-ms, line 74: Same, I don't see transport type being set

iothub/service/src/Jobs/JobRequest.cs
```diff
@@ -56,7 +56,7 @@ public class JobRequest
         /// <summary>
         /// ISO 8601 date time to start the job.
         /// </summary>
-        [JsonProperty(PropertyName = "startTimeUtc")]
+        [JsonProperty(PropertyName = "startTime")]
```
drwill-ms, line 59: Is this fixing a previous break? Otherwise, we are breaking the contract with the service by rneaming this.

e2e/test/iothub/messaging/MessageReceiveE2EPoolAmqpTests.cs
drwill-ms: oh, it just need a `using` keyword before `var` and you've added it. Just switch to the other style.

iothub/service/src/Jobs/JobRequest.cs
drwill-ms: Ah, setting it back to what it was before; thanks.

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,311 @@
...
+            IotHubServiceClientOptions options)
+        {
+            _hostName = hostName;
+            _credentialProvider = credentialProvider;
+            _clientOptions = options;
+            _connection = new IotHubConnection(credentialProvider, options.UseWebSocketOnly, options.TransportSettings, options);
+            _openTimeout = IotHubConnection.DefaultOpenTimeout;
+            _operationTimeout = IotHubConnection.DefaultOperationTimeout;
+            _faultTolerantSendingLink = new FaultTolerantAmqpObject<SendingAmqpLink>(CreateSendingLinkAsync, _connection.CloseLink);
+            _sendingPath = "/messages/deviceBound";
```
drwill-ms, line 70: consider inline init and make const instead of readonly

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,311 @@
...
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Dispose the MessagingClient instance.
+        /// </summary>
+        public virtual void Dispose(bool disposing)
```
drwill-ms, line 134: should be protected

drwill-ms: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

drwill-ms (approved):

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Subclient of <see cref="IotHubServiceClient"/> for handling cloud-to-device message feedback.
+    /// </summary>
+    /// <seealso href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-c2d#message-feedback"/>.
+    public class MessageFeedbackProcessorClient : IDisposable
```
timtay-microsoft, line 11: I don't follow how users are expected to use this class to get c2d message feedback. I was expecting to see a callback that executes each time feedback is received, but I don't see anything like that. All I see is the ability to open and close an AMQP connection

timtay-microsoft: Same comment for FileUploadNotificationProcessort

timtay-microsoft: I'd suggest you take a look at the [Java implementation](https://github.com/Azure/azure-iot-sdk-java/blob/main/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/messaging/MessageFeedbackProcessorClient.java) of this class and the [associated sample](https://github.com/Azure/azure-iot-sdk-java/blob/main/service/iot-service-samples/message-feedback-processor-client-sample/src/main/java/samples/com/microsoft/azure/sdk/iot/MessageFeedbackProcessorClientSample.java) to understand what this client is supposed to do.

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        private readonly string _hostName;
+        private readonly IotHubConnectionProperties _credentialProvider;
+        private readonly IotHubConnection _connection;
```
timtay-microsoft, line 15: You'll need to expose a callback in this class that notifies users when the connection is lost. Otherwise, they may successfully call open, wait for feedback, and then never get it because the connection was dropped. See [here](https://github.com/Azure/azure-iot-sdk-java/blob/c35f7782e400d3d7cc341b370117e70ece76236f/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/messaging/MessageFeedbackProcessorClient.java#L43) for the Java equivalent

timtay-microsoft: All these 3 AMQP subclients will need this feature since all 3 rely on a persistent connection.

timtay-microsoft (changes requested):
These aren't usable APIs yet

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
tmahmood-microsoft: @timtay-microsoft added the suggested changes.

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,156 @@
...
+                    Logging.Exit(this, $"Opening FeedbackReceiver", nameof(OpenAsync));
+            }
+        }
+
+        /// <summary>
+        /// Receive cloud-to-device message feedback.
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <exception cref="OperationCanceledException">If the provided <paramref name="cancellationToken"/> has requested cancellation.</exception>
+        public async Task ReceiveAsync(CancellationToken cancellationToken = default)
```
timtay-microsoft, line 88: It looks like you are mixing design patterns a bit here. A class like this could go one of 2 ways.

The first way is that the user gives us a callback to execute each time a feedback message is received on the opened connection.

```csharp
public class SampleCode
{
    public async Task main()
    {
        serviceClient.MessageFeedbackProcessorClient.MessageFeedbackProcessor = 
            HandleMessageFeedback;
        await serviceClient.MessageFeedbackProcessorClient.OpenAsync();
    }
    
    public AcknowledgementType HandleMessageFeedback(FeedbackBatch feedback)
    {
        // customer logic for handling the feedback
        return AcknowledgementType.Complete; 
    }
}
```


The second way is to ask the user to constantly poll for feedback messages.

```csharp
public class SampleCode
{
    public async Task main()
    {
        await serviceClient.MessageFeedbackProcessorClient.OpenAsync();

        while (true)
        {
            FeedbackBatch feedbackBatch = await serviceClient.MessageFeedbackProcessorClient.ReceiveAsync();
            if (feedbackBatch != null)
            {
                // do something with the received feedbackBatch
            }
        }
    }
}
```

timtay-microsoft: I generally prefer the first pattern, but this could be a design discussion you bring up to the broader team if you feel it is worth talking about in depth. If you do go with the first pattern, then there shouldn't be a ```ReceiveAsync()``` call in your class here

tmahmood-microsoft: @timtay-microsoft  where would we invoke the callback in the first case then?

timtay-microsoft: Internally. The customer wouldn't need to do anything other than open the connection

timtay-microsoft: While we've been reticent to maintain e2e tests of these two AMQP receiver classes due to how flakey the feature feels, it may be worthwhile to write some just to test out your code locally and get a feel for how usable your APIs are here.

abhipsaMisra: Azure AMQP does allow you to assign a callback, which would get triggered any time a message is received on the receiving link. This is the API you'd want to use to be able to set a listener: https://github.com/Azure/azure-amqp/blob/43e7b821e4fe119a4a47fa2240a6a1f8871d0848/src/ReceivingAmqpLink.cs#L56-L66

Our device client cloud-to-device callback API makes use of this feature: https://github.com/Azure/azure-iot-sdk-csharp/blob/05a98010ef1daab9dc3208d3aaf97628e2828dc6/iothub/device/src/Transport/Amqp/AmqpUnit.cs#L302

drwill-ms: This sounds like the ReceiveAsync vs SetC2dMessageHandlerAsync design choice. Right?

I'd prefer the callback too.

timtay-microsoft: > This sounds like the ReceiveAsync vs SetC2dMessageHandlerAsync design choice. Right?

Yes, it's the same discussion we had for receiving c2d messages on the device side

tmahmood-microsoft: @timtay-microsoft added the suggested changes.

e2e/test/iothub/service/PurgeMessageQueueE2ETests.cs
```diff
@@ -0,0 +1,52 @@
...
+            for (int i = 0; i < 50; ++i)
+            {
+                await sc.Messaging.SendAsync(deviceId, testMessage);
+            }
+            PurgeMessageQueueResult result = await sc.Messaging.PurgeMessageQueueAsync(deviceId, CancellationToken.None).ConfigureAwait(false);
+            Assert.AreEqual(expectedResult.DeviceId, result.DeviceId);
+            Assert.AreEqual(expectedResult.TotalMessagesPurged, result.TotalMessagesPurged);
+        }
+
+        private Message ComposeD2CTestMessage()
```
timtay-microsoft, line 37: You don't need to go so far as to create a message with all these detailed properties for this test to work. 

```csharp
return var new Message(Encoding.UTF8.GetBytes("some payload"));
```

iothub/service/src/IotHubServiceClient.cs
```diff
@@ -125,6 +127,7 @@ public IotHubServiceClient(string hostName, AzureSasCredential credential, IotHu
             _hostName = hostName;
             _httpClient = HttpClientFactory.Create(_hostName, options);
             _httpRequestMessageFactory = new HttpRequestMessageFactory(_credentialProvider.HttpsEndpoint, ApiVersion);
+            _options = options;
```
timtay-microsoft, line 130: It doesn't look like we need to save this options bundle in this class locally. All you need to do is pass it as an argument into the ```InitializeSubclients``` function below

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -59,6 +61,23 @@ public IotHubServiceClientOptions()
         /// </summary>
         public HttpClient HttpClient { get; set; }
 
+        /// <summary>
+        /// Whether to use web sockets or not.
+        /// </summary>
+        /// <remarks>
+        /// Only used for AMQP or MQTT.
```
timtay-microsoft, line 68: There is no MQTT usage in the service client, so remove it here. On top of that, the user isn't expected to know off-hand which subclients use AMQP, so it may be worth enumerating which ones do here.

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -59,6 +61,23 @@ public IotHubServiceClientOptions()
...
+        /// </remarks>
+        public bool UseWebSocketOnly { get; set; }
+
+        /// <summary>
+        /// Service client transport settings used for AMQP/MQTT operations.
```
timtay-microsoft, line 73: No MQTT

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -59,6 +61,23 @@ public IotHubServiceClientOptions()
...
+        /// </summary>
+        /// <remarks>
+        /// If provided, all other settings will be ignored, otherwise,
+        /// a transport settings client will be created based on other provided settings.
+        /// </remarks>
+        public ServiceClientTransportSettings TransportSettings { get; set; }
```
timtay-microsoft, line 79: This class doesn't appear to add any options that we didn't already have. Why are you adding it here?

timtay-microsoft: This ```IotHubServiceClientOptions``` class already allows users to provide a proxy that the HTTP client and AMQP client can use. The only other option is ```ConnectionLeaseTimeoutMilliseconds``` which we are intentionally no longer exposing. Users who want to configure settings at that level of granularity are expected to pass in an HTTP client they themselves constructed

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// <remarks>
+        /// May not be null.
+        /// </remarks>
+        public Func<FileNotification, DeliveryAcknowledgement> _fileNotificationProcessor;
+
+        /// <summary>
+        /// The callback to be executed when the connection is lost.
+        /// </summary>
+        /// <remarks>
+        /// May not be null.
```
timtay-microsoft, line 36: We can allow this field to be null. The customer won't be notified of connection drops which is bad practice, but we can still allow it.

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+    {
+        private readonly string _hostName;
+        private readonly IotHubConnectionProperties _credentialProvider;
+        private readonly IotHubConnection _connection;
+
+        /// <summary>
+        /// The callback to be executed each time file upload notification is received from the service.
+        /// </summary>
+        /// <remarks>
+        /// May not be null.
```
timtay-microsoft, line 28: While this field should not be null, it does appear to be null when this subclient is initialized. I'd consider throwing an exception if the user calls ```OpenAsync()``` without first setting a value here.

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// </remarks>
+        public Func<FileNotification, DeliveryAcknowledgement> _fileNotificationProcessor;
+
+        /// <summary>
+        /// The callback to be executed when the connection is lost.
+        /// </summary>
+        /// <remarks>
+        /// May not be null.
+        /// </remarks>
+        public Action<Exception> _errorProcessor;
```
timtay-microsoft, line 38: Don't have this action give the user the Exception like this. Wrap it into a class like we did in Java [here](https://github.com/Azure/azure-iot-sdk-java/blob/c35f7782e400d3d7cc341b370117e70ece76236f/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/messaging/FileUploadNotificationProcessorClient.java#L43). The reason we do this is so that we can add additional fields later such as DateTime when it happened, previous state, etc. without causing breaking changes

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+            _hostName = hostName;
+            _credentialProvider = credentialProvider;
+            _connection = new IotHubConnection(credentialProvider, options.UseWebSocketOnly, options.TransportSettings, options);
+            FileNotificationReceiver = new AmqpFileNotificationReceiver(_connection);
+        }
+
+        /// <summary>
+        /// Gets the AmqpFileNotificationReceiver which can deliver notifications for file upload operations.
+        /// </summary>
+        internal AmqpFileNotificationReceiver FileNotificationReceiver { get; }
```
timtay-microsoft, line 61:
```suggestion
        internal AmqpFileNotificationReceiver _fileNotificationReceiver;
```

There's no need to make a property out of this if we aren't exposing it to users.

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+            IotHubServiceClientOptions options)
+        {
+            _hostName = hostName;
+            _credentialProvider = credentialProvider;
+            _connection = new IotHubConnection(credentialProvider, options.UseWebSocketOnly, options.TransportSettings, options);
+            FileNotificationReceiver = new AmqpFileNotificationReceiver(_connection);
+        }
+
+        /// <summary>
+        /// Gets the AmqpFileNotificationReceiver which can deliver notifications for file upload operations.
```
timtay-microsoft, line 59:
```suggestion
        /// Gets the AmqpFileNotificationReceiver which receives notifications for file upload operations.
```

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+                ReceivingAmqpLink receivingLink = await FileNotificationReceiver.FaultTolerantReceivingLink.GetReceivingLinkAsync().ConfigureAwait(false);
+                receivingLink.RegisterMessageListener(OnNotificationMessageReceivedAsync);
+            }
+            catch (Exception ex)
+            {
+                if (Logging.IsEnabled)
+                    Logging.Error(this, $"{nameof(OpenAsync)} threw an exception: {ex}", nameof(OpenAsync));
+                if (ex is IotHubException || ex is IOException)
+                {
+                    _errorProcessor?.Invoke(ex);
```
timtay-microsoft, line 83: There's no need to feed the error processor with this error when we throw it immediately afterwards. This error processor is only for surfacing connection level errors that occur in threads that the user does not maintain. Since the user maintains the thread that called ```OpenAsync()```, there is no need to feed this error to the error processor.

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+            FileNotificationReceiver = new AmqpFileNotificationReceiver(_connection);
+        }
+
+        /// <summary>
+        /// Gets the AmqpFileNotificationReceiver which can deliver notifications for file upload operations.
+        /// </summary>
+        internal AmqpFileNotificationReceiver FileNotificationReceiver { get; }
+
+        /// <summary>
+        /// Open the AmqpFileNotificationReceiver instance.
```
timtay-microsoft, line 64: Customers have no idea what an ```AmqpFileNotificationReceiver``` is.

```suggestion
        /// Open the connection and start receiving file upload notifications.
```

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+            }
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, $"Opening AmqpFileNotificationReceiver", nameof(OpenAsync));
+            }
+        }
+
+        /// <summary>
+        /// Close the AmqpFileNotificationReceiver instance.
```
timtay-microsoft, line 95: Same comment as in the OpenAsync xml comments

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+                Logging.Enter(this, $"Disposing AmqpFileNotificationReceiver", nameof(Dispose));
+
+            FileNotificationReceiver.Dispose();
+
+            if (Logging.IsEnabled)
+                Logging.Exit(this, $"Disposing AmqpFileNotificationReceiver", nameof(Dispose));
+            GC.SuppressFinalize(this);
+        }
+
+        private async void OnNotificationMessageReceivedAsync(AmqpMessage amqpMessage)
```
timtay-microsoft, line 137: This function

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+                if (amqpMessage != null && _fileNotificationProcessor != null)
+                {
+                    using (amqpMessage)
+                    {
+                        AmqpClientHelper.ValidateContentType(amqpMessage, CommonConstants.FileNotificationContentType);
+
+                        FileNotification fileNotification = await AmqpClientHelper.GetObjectFromAmqpMessageAsync<FileNotification>(amqpMessage).ConfigureAwait(false);
+                        fileNotification.LockToken = new Guid(amqpMessage.DeliveryTag.Array).ToString();
+
+                        _fileNotificationProcessor.Invoke(fileNotification);
```
timtay-microsoft, line 153: You are ignoring the AcknowledgementType return value here. This return value decides how the AMQP layer acknowledges this particular file upload notification message.

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+                    }
+                }
+            }
+            catch(Exception ex)
```
timtay-microsoft, line 157:
```suggestion
            catch (Exception ex)
```

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                if (Logging.IsEnabled)
+                    Logging.Error(this, $"{nameof(OnNotificationMessageReceivedAsync)} threw an exception: {ex}", nameof(OnNotificationMessageReceivedAsync));
+                if (ex is IotHubException || ex is IOException)
+                {
+                    _errorProcessor?.Invoke(ex);
```
timtay-microsoft, line 163: This is a good example of a case where we should invoke the error processor but should **not** throw. This is a callback function that executes on the AMQP layer's thread, not the user's application thread.

timtay-microsoft: So keep this line, but don't throw this error below.

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// </remarks>
+        public Func<FeedbackBatch, DeliveryAcknowledgement> _messageFeedbackProcessor;
+
+        /// <summary>
+        /// The callback to be executed when the connection is lost.
+        /// </summary>
+        /// <remarks>
+        /// May not be null.
+        /// </remarks>
+        public Action<Exception> _errorProcessor;
```
timtay-microsoft, line 39: I don't see this error processor being invoked when the AMQP connection is lost. So far, it is only invoked if the handling of a file upload notification fails. The connection can be lost even when no messages are being received.

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Close the FeedbackReceiver instance.
+        /// </summary>
+        public virtual async Task CloseAsync()
+        {
+            if (Logging.IsEnabled)
+                Logging.Enter(this, $"Closing FeedbackReceiver", nameof(CloseAsync));
```
timtay-microsoft, line 100:
```suggestion
                Logging.Enter(this, $"Closing MessageFeedbackProcessorClient", nameof(CloseAsync));
```

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Transport types supported by MessagingClient.
+    /// </summary>
+    /// <remarks>
+    /// Amqp and Amqp over WebSocket only.
+    /// </remarks>
+    public enum TransportType
```
timtay-microsoft, line 23: This enum is applicable to the other subclients you're adding in this PR, so it should live in its own file.

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+    {
+        /// <summary>
+        /// Advanced Message Queuing Protocol transport.
+        /// </summary>
+        Amqp,
+
+        /// <summary>
+        /// Advanced Message Queuing Protocol transport over WebSocket only.
+        /// </summary>
+        Amqp_WebSocket_Only
```
timtay-microsoft, line 33:
```suggestion
        Amqp_WebSocket
```

The "Only" suffix is a relic of when we wanted to implement fallback protocols (for example, try AMQP, but if that doesn't work, try AMQP_WS). We don't support fallback protocols in v2, so we can simplify this name

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    /// <summary>
+    /// Subclient of <see cref="IotHubServiceClient"/> for sending cloud to device and cloud to module messages.
+    /// </summary>
+    /// <seealso href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-c2d#message-feedback"/>.
```
timtay-microsoft, line 39:
```suggestion
    /// <seealso href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-c2d"/>.
```

That subsection of the article is after the more general C2D documentation

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+    public class MessagingClient : IDisposable
+    {
+        private readonly string _hostName;
+        private readonly IotHubConnectionProperties _credentialProvider;
+        private readonly IotHubConnection _connection;
+        private readonly IotHubServiceClientOptions _clientOptions;
+        private readonly HttpClient _httpClient;
+        private readonly HttpRequestMessageFactory _httpRequestMessageFactory;
+        private readonly FaultTolerantAmqpObject<SendingAmqpLink> _faultTolerantSendingLink;
+        private readonly TimeSpan _openTimeout;
```
timtay-microsoft, line 49: We don't need default timeout values. If the user doesn't pass in a cancellation token, then they are fine with the method call potentially taking forever

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+        private readonly TimeSpan _operationTimeout;
```
timtay-microsoft, line 50: Same here

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+
+        private const string _sendingPath = "/messages/deviceBound";
+        private const string PurgeMessageQueueFormat = "/devices/{0}/commands";
+        private int _sendingDeliveryTag;
+
+        /// <summary>
+        /// The callback to be executed when the connection is lost.
+        /// </summary>
+        /// <remarks>
+        /// May not be null.
```
timtay-microsoft, line 60: This can be null even if we don't recommend it

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+            _clientOptions = options;
+            _connection = new IotHubConnection(credentialProvider, options.UseWebSocketOnly, options.TransportSettings, options);
+            _openTimeout = IotHubConnection.DefaultOpenTimeout;
+            _operationTimeout = IotHubConnection.DefaultOperationTimeout;
+            _faultTolerantSendingLink = new FaultTolerantAmqpObject<SendingAmqpLink>(CreateSendingLinkAsync, _connection.CloseLink);
+            
+        }
+
+        /// <summary>
+        /// Open the MessagingClient instance.
```
timtay-microsoft, line 91:
```suggestion
        /// Open this instance. Must be done before any cloud to device messages can be sent.
```

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+        /// </summary>
+        public virtual async Task OpenAsync()
```
timtay-microsoft, line 93: documentation on what exceptions can be thrown?

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+            }
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, $"Opening MessagingClient", nameof(OpenAsync));
+            }
+        }
+
+        /// <summary>
+        /// Close the MessagingClient instance.
```
timtay-microsoft, line 116: Needs documentation on if a client can be re-opened after it was closed.

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+        /// <param name="message">The cloud-to-device message.</param>
+        /// <param name="timeout">The operation timeout, which defaults to 1 minute if unspecified.</param>
+        /// <exception cref="ArgumentNullException">Thrown when the provided <paramref name="deviceId"/> or <paramref name="message"/> is null.</exception>
+        /// <exception cref="ArgumentException">Thrown if the provided <paramref name="deviceId"/> is empty or whitespace.</exception>
+        /// <exception cref="IotHubException">
+        /// Thrown if IoT hub responded to the request with a non-successful status code. For example, if the provided
+        /// request was throttled, <see cref="IotHubThrottledException"/> is thrown. For a complete list of possible
+        /// error cases, see <see cref="Common.Exceptions"/>.
+        /// </exception>
+        public virtual async Task SendAsync(string deviceId, Message message, TimeSpan? timeout = null)
```
timtay-microsoft, line 168: We don't want to take TimeSpans for timeouts anymore. We should only take CancellationTokens

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+            Argument.RequireNotNull(message, nameof(message));
+
+            if (_clientOptions?.SdkAssignsMessageId == SdkAssignsMessageId.WhenUnset && message.MessageId == null)
+            {
+                message.MessageId = Guid.NewGuid().ToString();
+            }
+
+            if (message.IsBodyCalled)
+            {
+                message.ResetBody();
```
timtay-microsoft, line 183: This looks like it is only necessary if we support users passing in Streams as the payload. If we just take byte[], then there is no need for this.

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -0,0 +1,379 @@
...
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, $"Sending message with Id [{message?.MessageId}] for device {deviceId}, module {moduleId}", nameof(SendAsync));
+            }
+        }
+
+        /// <summary>
+        /// Removes all cloud-to-device messages from a device's queue.
+        /// </summary>
+        /// <remarks>
+        /// This call is made over HTTP. Call to <see cref="OpenAsync"/> or <see cref="CloseAsync"/> does not affect this method.
```
timtay-microsoft, line 300:
```suggestion
        /// This call is made over HTTP. There is no need to call <see cref="OpenAsync"/> before calling this method.
```

iothub/service/tests/ServiceClientTests.cs
```diff
@@ -12,83 +12,14 @@ namespace Microsoft.Azure.Devices.Api.Test
     using Microsoft.Azure.Amqp;
     using Microsoft.Azure.Devices;
     using Microsoft.Azure.Devices.Common.Exceptions;
+    using Microsoft.Azure.Devices.Http2;
```
timtay-microsoft, line 15: Probably unused

timtay-microsoft:
e2e tests are missing for feedback message receiver and for file upload notification receiver

iothub/service/src/Messaging/MessagingClient.cs
tmahmood-microsoft: this change is added in the PR for removing Disposable dependencies from Message API. removing it from this PR as well.

iothub/service/tests/ServiceClientTests.cs
tmahmood-microsoft: @timtay-microsoft I have added the suggested changes. Please have a look.

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+            try
+            {
+                if (_fileNotificationProcessor == null)
+                {
+                    throw new Exception("Callback for message feedback {0} must be set before opening the connection.".FormatInvariant(_fileNotificationProcessor));
+                }
+                await FileNotificationReceiver.OpenAsync().ConfigureAwait(false);
+                ReceivingAmqpLink receivingAmqpLink = await FileNotificationReceiver.FaultTolerantReceivingLink.GetReceivingLinkAsync().ConfigureAwait(false);
+                receivingAmqpLink.RegisterMessageListener(OnNotificationMessageReceivedAsync);
+                receivingAmqpLink.Session.Connection.Closed += ConnectionClosed;
```
timtay-microsoft, line 88: We should also add our ```ConnectionClosed``` handler to these events as well:

```csharp
receivingAmqpLink.Session.Closed += ConnectionClosed;
receivingAmqpLink.Closed += ConnectionClosed;
```

For the customer's purposes, losing any of the link and/or session is basically equivalent to losing the connection/connection in that they won't be able to receive anything anymore.

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Amqp;
+using Microsoft.Azure.Devices.Common;
+using Microsoft.Azure.Devices.Common.Exceptions;
+using Microsoft.Azure.Devices.Common.Extensions;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Subclient of <see cref="IotHubServiceClient"/> for handling file upload notications.
```
timtay-microsoft, line 17:
```suggestion
    /// Subclient of <see cref="IotHubServiceClient"/> for receiving file upload notifications.
```

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+        public virtual async Task OpenAsync()
+        {
+            if (Logging.IsEnabled)
+                Logging.Enter(this, $"Opening FileUploadNotificationProcessorClient", nameof(OpenAsync));
+
+            try
+            {
+                if (_fileNotificationProcessor == null)
+                {
+                    throw new Exception("Callback for message feedback {0} must be set before opening the connection.".FormatInvariant(_fileNotificationProcessor));
```
timtay-microsoft, line 83: We'll likely change this exception type later with Bryce's work, so it's fine to leave this as Exception for now. I don't see any benefit to adding the null file upload notification processor to the message, though.

```suggestion
                    throw new Exception("Callback for file upload notifications must be set before opening the connection.");
```

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+            FileNotificationReceiver = new AmqpFileNotificationReceiver(_connection);
+        }
+
+        /// <summary>
+        /// Gets the AmqpFileNotificationReceiver which receives notifications for file upload operations.
+        /// </summary>
+        internal AmqpFileNotificationReceiver FileNotificationReceiver;
+
+        /// <summary>
+        /// Open the connection and start receiving file upload notifications.
```
timtay-microsoft, line 63: Add a note here that the user needs to set the file upload notification callback before calling this

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+            _hostName = hostName;
+            _credentialProvider = credentialProvider;
+            _connection = new IotHubConnection(credentialProvider, options.UseWebSocketOnly, options);
+            FileNotificationReceiver = new AmqpFileNotificationReceiver(_connection);
+        }
+
+        /// <summary>
+        /// Gets the AmqpFileNotificationReceiver which receives notifications for file upload operations.
+        /// </summary>
+        internal AmqpFileNotificationReceiver FileNotificationReceiver;
```
timtay-microsoft, line 60:
```suggestion
        private AmqpFileNotificationReceiver _fileUploadNotificationReceiver;
```

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// </summary>
+        /// <remarks>
+        /// May not be null.
+        /// </remarks>
+        public Func<FileNotification, DeliveryAcknowledgement> _fileNotificationProcessor;
+
+        /// <summary>
+        /// The callback to be executed when the connection is lost.
+        /// </summary>
+        public Action<ErrorContext> _errorProcessor;
```
timtay-microsoft, line 37: Public facing properties and fields should be CamelCase.

We only name private fields with an underscore and camelCase.

```suggestion
        public Action<ErrorContext> ErrorProcessor;
```

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+            IotHubServiceClientOptions options)
+        {
+            _hostName = hostName;
+            _credentialProvider = credentialProvider;
+            _connection = new IotHubConnection(credentialProvider, options.UseWebSocketOnly, options);
+            FileNotificationReceiver = new AmqpFileNotificationReceiver(_connection);
+        }
+
+        /// <summary>
+        /// Gets the AmqpFileNotificationReceiver which receives notifications for file upload operations.
```
timtay-microsoft, line 58: No need to document private fields with xml headers
```
// The AMQP receiver object that listens for file upload notifications.
```

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+        private readonly IotHubConnectionProperties _credentialProvider;
+        private readonly IotHubConnection _connection;
+
+        /// <summary>
+        /// The callback to be executed each time file upload notification is received from the service.
+        /// </summary>
+        /// <remarks>
+        /// May not be null.
+        /// </remarks>
+        public Func<FileNotification, DeliveryAcknowledgement> _fileNotificationProcessor;
```
timtay-microsoft, line 32: This ```DeliveryAcknowledgement``` type isn't what we need from the user here. We need an enum with three possible values:

```csharp
public enum AcknowledgementType
{
    Complete, // maps to Completed in AMQP
    Abandon, // maps to Released in AMQP
    Reject // maps to Rejected in AMQP
}
```

timtay-microsoft: Complete means that the received file upload notification/feedback message was received and handled and the service can remove it from the queue. 

Abandon means the received file upload notifications/feedback message was received but not handled and that the service should requeue it so that this or any other receiver can receive it later.

Rejected means the file upload notification was received but was not handled because it was poorly formatted or otherwise unhandleable. This acknowledgement type makes the service remove the item from the queue like with Completed.

timtay-microsoft: It is very similar to how device clients can abandon/reject/complete cloud to device messages, so [this reading](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-c2d#the-cloud-to-device-message-life-cycle) may be helpful

iothub/service/src/Messaging/MessageFeedbackProcessorClient.cs
```diff
@@ -0,0 +1,222 @@
...
+                    using (amqpMessage)
+                    {
+                        AmqpClientHelper.ValidateContentType(amqpMessage, CommonConstants.BatchedFeedbackContentType);
+                        IEnumerable<FeedbackRecord> records = await AmqpClientHelper
+                            .GetObjectFromAmqpMessageAsync<IEnumerable<FeedbackRecord>>(amqpMessage).ConfigureAwait(false);
+
+                        FeedbackBatch feedbackBatch = new FeedbackBatch
+                        {
+                            EnqueuedTime = (DateTime)amqpMessage.MessageAnnotations.Map[MessageSystemPropertyNames.EnqueuedTime],
+                            LockToken = new Guid(amqpMessage.DeliveryTag.Array).ToString(),
```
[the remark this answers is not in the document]
timtay-microsoft: For context here, "lock token" is an IoT hub level concept that is built on top of the AMQP level concept of delivery tags. Each AMQP message that we receive has a delivery tag attached (expressed as a byte[], but usually comes from an integer like "1", or a guid). When we want to acknowledge a message over AMQP, we basically say "I complete the message with delivery tag ____ "

iothub/service/src/Messaging/ErrorContext.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// For example, if the device has no internet connection.
+        /// </remarks>
+        /// <param name="iOException"></param>
+        public ErrorContext(IOException iOException)
+        {
+            IOException = iOException;
+        }
+
+        IotHubException IotHubException { get; }
+        IOException IOException { get; }
```
timtay-microsoft, line 39: Documentation?

timtay-microsoft: Make sure to specify that exactly one of these fields will always be present, but they will never **both** be present.

iothub/service/src/Messaging/ErrorContext.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
```
timtay-microsoft, line 1: Missing the copyright header that we put at the top of every source code file. I'd setup your Visual Studio instance to do this for you: https://docs.microsoft.com/en-us/visualstudio/ide/reference/add-file-header?view=vs-2022

iothub/service/src/Messaging/ErrorContext.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Microsoft.Azure.Devices.Common.Exceptions;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    ///
```
timtay-microsoft, line 10: Documentation?

iothub/service/src/Messaging/TransportType.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <summary>
+    /// Transport types supported by MessagingClient.
+    /// </summary>
+    /// <remarks>
+    /// Amqp and Amqp over WebSocket.
+    /// </remarks>
+    public enum TransportType
+    {
+        /// <summary>
+        /// Advanced Message Queuing Protocol transport.
```
timtay-microsoft, line 16: Call out what port this uses in the remarks. Users with firewalls may have this port blocked.

iothub/service/src/Messaging/TransportType.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// </summary>
+        Amqp,
+
+        /// <summary>
+        /// Advanced Message Queuing Protocol transport over WebSocket.
```
timtay-microsoft, line 21: Same here

iothub/service/src/Messaging/TransportType.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Transport types supported by MessagingClient.
```
timtay-microsoft, line 8: This should be used by FileUploadNotificationProcessor and MessageFeedbackProcessor as well

iothub/service/src/Messaging/FileUploadNotificationProcessorClient.cs
timtay-microsoft: Although, now that I look at the Java implementation, the service may not support Reject here.

https://github.com/Azure/azure-iot-sdk-java/blob/main/service/iot-service-client/src/main/java/com/microsoft/azure/sdk/iot/service/messaging/AcknowledgementType.java

iothub/service/src/Messaging/ErrorContext.cs
tmahmood-microsoft: @timtay-microsoft  added header to all missing files.

e2e/test/iothub/service/MessageFeedbackReceiverE2ETest.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    /// E2E test class for MessageFeedbackReceiver.
+    /// </summary>
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    public class MessageFeedbackReceiverE2eTest : E2EMsTestBase
+    {
+        private bool messagedFeedbackReceived;
+        [TestMethod]
```
timtay-microsoft, line 22: Needs a timeout attribute. Check another test to see how it is defined normally

e2e/test/iothub/service/MessageFeedbackReceiverE2ETest.cs
```diff
@@ -0,0 +1,55 @@
...
+            await serviceClient.MessageFeedbackProcessor.CloseAsync();
+            messagedFeedbackReceived.Should().BeTrue();
+        }
+
+        private AcknowledgementType feedbackCallback(FeedbackBatch feedback)
+        {
+            messagedFeedbackReceived = true;
+            return AcknowledgementType.Complete;
+        }
+        private async Task receiveMessage()
```
timtay-microsoft, line 48: Newline after a brace

e2e/test/iothub/service/MessageFeedbackReceiverE2ETest.cs
```diff
@@ -0,0 +1,55 @@
...
+            serviceClient.MessageFeedbackProcessor.MessageFeedbackProcessor = feedbackCallback;
+            await serviceClient.MessageFeedbackProcessor.OpenAsync().ConfigureAwait(false);
+            var message = new Message(Encoding.UTF8.GetBytes("some payload"));
+            message.Ack = DeliveryAcknowledgement.Full;
+            messagedFeedbackReceived = false;
+            await serviceClient.Messaging.SendAsync(TestConfiguration.IoTHub.X509ChainDeviceName, message).ConfigureAwait(false);
+            await receiveMessage().ConfigureAwait(false);
+
+            var timer = Stopwatch.StartNew();
+            while (timer.ElapsedMilliseconds < 60000)
```
timtay-microsoft, line 35: Don't just wait until a minute has passed. You should check to see if the feedback message has been received here as well. If the feedback message has been received, then you can exit this loop since the test passed

e2e/test/iothub/service/MessageFeedbackReceiverE2ETest.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                continue;
+            }
+            await serviceClient.MessageFeedbackProcessor.CloseAsync();
+            messagedFeedbackReceived.Should().BeTrue();
+        }
+
+        private AcknowledgementType feedbackCallback(FeedbackBatch feedback)
```
timtay-microsoft, line 43:
```suggestion
        private AcknowledgementType OnFeedbackReceived(FeedbackBatch feedback)
```

e2e/test/iothub/service/MessageFeedbackReceiverE2ETest.cs
timtay-microsoft: It also helps test failures be more descriptive if you explicitly throw an assertion failure saying "Timed out waiting to receive message feedback" once this loop completes and the message feedback wasn't received

iothub/service/src/FileUpload/FileUploadNotificationProcessorClient.cs
```diff
@@ -0,0 +1,215 @@
...
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, amqpMessage, nameof(OnNotificationMessageReceivedAsync));
+            }
+        }
+
+        private void ConnectionClosed(object sender, EventArgs e)
+        {
+            var ex = new IotHubException(e.ToString());
```
timtay-microsoft, line 209: This isn't all that helpful. Ideally, we would parse out any AMQP level error code so that users can interpret this exception. Sections 2.8.16 - 2.8.18 of the [AMQP spec](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-complete-v1.0-os.pdf) enumerate the possible error codes we should expect from the AMQP client here.

Some of these errors are network level issues (and should be given to users as IOExceptions) such as "amqp:connection:framing-error" whereas some errors are Hub level issues such as "amqp:link:message-size-exceeded" when a user sends a message that is too big.

Bryce is working on the broad exception strategies for our v2 SDK, so I'm not asking you to figure that out right now, but I would expect some kind of parsing logic to get the AMQP error code. The device client side should be doing this already if you want to take reference from it.

tmahmood-microsoft: Will the exception be passed as an instance of EventArgs or I would need to convert it into an instance of amqp exception?

tmahmood-microsoft: @timtay-microsoft added the requested changes.

timtay-microsoft (approved):

PR #2555, opened 2022-08-05
Fix some deserialization issues around job requests

timtay-microsoft:
IoT hub is inconsistent with the json key value for these fields across different jobs APIs (schedule update twin, get job, query jobs, etc.) so we need a bit of flexibility when deserializing the ScheduledJob type.

Also marking an optional field (MaxExecutionTime) as nullable


iothub/service/src/Jobs/ScheduledJob.cs
```diff
@@ -24,24 +24,93 @@ public class ScheduledJob
         [JsonProperty(PropertyName = "queryCondition", NullValueHandling = NullValueHandling.Ignore)]
         public string QueryCondition { get; internal set; }
 
+        /// <summary>
+        /// Scheduled job start time in UTC.
+        /// </summary>
+        [JsonProperty(PropertyName = "createdDateTimeUtc", NullValueHandling = NullValueHandling.Ignore)]
```
brycewang-microsoft, line 30: "createdDateTimeUtc" or "CreatedTimeUtc"? seems inconsistent

drwill-ms: Is this a duplicate of the next property?

drwill-ms: https://msazure.visualstudio.com/One/_git/Azure-IoT-Hub-Main?path=/swagger/2021-11-30/Service_2021-04-12.json&version=GBmaster&line=2477&lineEnd=2478&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

drwill-ms (changes requested):
I need to walk through these changes with you as I'm too confused and have some questions.

If different jobs have different fields, perhaps we should have different data model classes for them rather than try to have one that works for all?

timtay-microsoft:
> If different jobs have different fields, perhaps we should have different data model classes for them rather than try to have one that works for all?

David and I chatted about this a bit, and the big drawback to that plan is that we'd expose different return types to the user for job's APIs such as

```csharp
QueriedScheduledJob job1 = await serviceClient.Query.CreateAsync<QueriedJob>("SELECT * FROM devices.jobs");
ScheduledJob job2 = await serviceClient.Jobs.GetAsync("someJobId");
```

where QueriedScheduledJob and ScheduledJob have mostly the same fields (jobId, status, type, queryCondition, etc.) but differ on a few auxiliary fields (which date times are present, mostly).

For our user's sake, I'd rather we return the same ScheduledJob type regardless of if they used "query" or "get" to see it for simplicity. 

If we accept that, then the question becomes if we need two deserialization classes with largely duplicated code or if we just take the approach I've implemented in this PR to allows for 2 different json key's to deserialize into the same field.

iothub/service/src/Jobs/ScheduledJob.cs
timtay-microsoft: I've reworked this a little with David

drwill-ms (approved):

iothub/service/src/Jobs/JobRequest.cs
```diff
@@ -52,26 +52,24 @@ public class JobRequest
...
-        public TimeSpan MaxExecutionTime { get; set; }
+        public TimeSpan? MaxExecutionTime { get; set; }
 
         [JsonProperty(PropertyName = "maxExecutionTimeInSeconds", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        internal long MaxExecutionTimeInSeconds
+        internal long? MaxExecutionTimeInSeconds
         {
-            get => (long)MaxExecutionTime.TotalSeconds;
-            set => MaxExecutionTime = TimeSpan.FromSeconds(value);
+            get => MaxExecutionTime?.Seconds;
```
abhipsaMisra, line 71: Looks like we're changing the logic from this representing the TimeSpan as seconds to this representing the seconds component of the TimeSpan. Is this intentional? If yes, then a comment calling this out would be useful since generally we expect the SDK to use the TimeSpan total milliseconds value, and not just the TimeSpan.Seconds component.

iothub/service/src/Jobs/ScheduledJob.cs
```diff
@@ -27,35 +27,68 @@ public class ScheduledJob
...
         /// Scheduled job start time in UTC.
         /// </summary>
-        [JsonProperty(PropertyName = "createdTime", NullValueHandling = NullValueHandling.Ignore)]
+        [JsonProperty(PropertyName = "createdDateTimeUtc", NullValueHandling = NullValueHandling.Ignore)]
         public DateTime? CreatedTimeUtc { get; internal set; }
 
+        // Some service Jobs APIs use "createdTime" as the key for this value and some others use "createdDateTimeUtc".
+        // This private field is a workaround that allows us to deserialize either "createdTime" or "createdDateTimeUtc"
+        // as the created time value for this class and expose it either way as CreatedTimeUtc.
+        [JsonProperty(PropertyName = "createdTime")]
```
abhipsaMisra, line 36: Does this need any null value handling configuration?

iothub/service/src/Jobs/ScheduledJob.cs
```diff
@@ -27,35 +27,68 @@ public class ScheduledJob
...
+        [JsonConverter(typeof(StringEnumConverter))]
```
abhipsaMisra, line 37: Q - is this StringEnumConverter something we define? How is this being used in the context of this DateTime field?

iothub/service/src/Jobs/JobRequest.cs
timtay-microsoft: This was not intentional. I'll change this to MaxExecutionTime?.**Total**Seconds

iothub/service/src/Jobs/ScheduledJob.cs
timtay-microsoft: I'll add the config to make it ignore if null

iothub/service/src/Jobs/ScheduledJob.cs
timtay-microsoft: This comes from Newtonsoft for converting string json values into enum types that we define. It was a copy-paste accident that this attribute was given to a non-enum type like DateTime, so I'll remove it.

PR #2562, opened 2022-08-08
Remove dead code and unnecessary abstraction

drwill-ms:


iothub/device/src/Common/Fx.cs
```diff
@@ -41,22 +32,6 @@ public static ExceptionTrace Exception
             }
         }
 
-        // Do not call the parameter "message" or else FxCop thinks it should be localized.
-        [Conditional("DEBUG")]
-        public static void Assert(bool condition, string description)
```
drwill-ms, line 46: Silly methods like this (and next) were unnecessary abstractions over a well-known built-in library.

andyk-ms (approved):

iothub/device/tests/Transport/ServerWebSocketChannel.cs
```diff
@@ -249,7 +249,7 @@ async Task HandleCloseAsync()
             {
                 await CloseAsync().ConfigureAwait(false);
             }
-            catch (Exception) //when (!ex.IsFatal())
+            catch (Exception) //when (!Fx.IsFatal(ex))
```
timtay-microsoft, line 252: Remove the comment?

timtay-microsoft (approved):

iothub/device/tests/Transport/ServerWebSocketChannel.cs
drwill-ms: I thought about that. I wonder why it was commented out.

I didn't take the time to investigate why so I figured I might leave it in case there was a reason to restore it.

Thoughts? Feel it is safe to delete?

PR #2564, opened 2022-08-09
Removed .NET 5.0 dependency (Client Tests)

andyk-ms:
Removed NET 5.0 dependency
Removed test cases for Net 5.0
Completed unit tests with 100% pass rate for 613 tests in the group

### DIFF
--- a/authentication/tests/Microsoft.Azure.Devices.Shared.Tests.csproj
+++ b/authentication/tests/Microsoft.Azure.Devices.Shared.Tests.csproj
@@ -1,10 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <RootNamespace>Microsoft.Azure.Devices.Shared.Tests</RootNamespace>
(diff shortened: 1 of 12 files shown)


authentication/tests/Microsoft.Azure.Devices.Shared.Tests.csproj
```diff
@@ -1,10 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <RootNamespace>Microsoft.Azure.Devices.Shared.Tests</RootNamespace>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net6.0;net5.0;netcoreapp3.1;netcoreapp2.1.30;net472</TargetFrameworks>
```
drwill-ms, line 4: Should we remove it from testing too?

I assumed just NuGet targets.

drwill-ms (changes requested):
I only expected NuGet target removal, and any SDK preprocessor #ifs.

PR #2565, opened 2022-08-09
Replace custom delegates with functions, consolidate sas audience logic

abhipsaMisra:
- replace custom delegates with functions
- consolidate sas audience logic


iothub/device/src/InternalClient.cs
```diff
@@ -27,41 +27,33 @@ internal class InternalClient : IDisposable
...
-
-        private volatile ConnectionStateChangeHandler _connectionStateChangeHandler;
-
-        // Count of messages sent by the device/ module. This is used for sending diagnostic information.
-        private int _currentMessageCount;
-
-        private int _diagnosticSamplingPercentage;
-
+        // Connection state change information
+        private volatile Action<ConnectionState, ConnectionStateChangeReason> _connectionStateChangeHandler;
```
timtay-microsoft, line 31: This could be tackled in a later PR, but I'd like to see us create a bundle class for the information we provide in this callback. That way, we can easily add new fields like previousState, the exception that caused it, etc.

https://github.com/Azure/azure-iot-sdk-java/blob/main/device/iot-device-client/src/main/java/com/microsoft/azure/sdk/iot/device/ConnectionStatusChangeContext.java

timtay-microsoft (approved):

iothub/device/src/InternalClient.cs
abhipsaMisra: +1. We'll keep that for another PR. We'll also want to take another look at all our callback args and see if there is any room for improvement in there.

Issue #1137, opened 2019-11-22
Disconnect renders DeviceClient useless - The semaphore has been disposed.


domdeger:
- **OS, version, SKU and CPU architecture used:** <VERSION>  Windows 10 Desktop x64
- **Application's .NET Target Framework :** netcoreapp2.2 
- **Device:** Laptop, Raspberry PI3 
- **SDK version used:** Microsoft.Azure.Devices.Client v1.21.2 

## Description of the issue:
A device client which has been connected to an IoT-Hub and used to transfer telemetry does not recover from longer connection loss. So if I start the linked code with available network connection everything works fine. Telemetry arrives at the hub and method triggers are received.

But if you shutdown the network connection for approximatly three to five minutes the DeviceClient starts throwing System.ObjectDisposedExceptions with attached Stacktrace which does not seem to be expected behaviour.

Furthermore it persists after restoring the network connection.

## Code sample exhibiting the issue:
See https://gist.github.com/domdeger/0f192e46c3e40e3161fc195c827d3e4a

## Console log of the issue:
`   
The semaphore has been disposed.
at System.Threading.SemaphoreSlim.CheckDispose()
   at System.Threading.SemaphoreSlim.WaitAsync(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.SemaphoreSlim.WaitAsync(TimeSpan timeout, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.Transport.Amqp.AmqpConnectionHolder.EnsureConnectionAsync(TimeSpan timeout)
   at Microsoft.Azure.Devices.Client.Transport.Amqp.AmqpConnectionHolder.OpenSessionAsync(DeviceIdentity deviceIdentity, TimeSpan timeout)
   at Microsoft.Azure.Devices.Client.Transport.AmqpIoT.AmqpUnit.EnsureSessionAsync(TimeSpan timeout)
   at Microsoft.Azure.Devices.Client.Transport.AmqpIoT.AmqpUnit.OpenAsync(TimeSpan timeout)
   at Microsoft.Azure.Devices.Client.Transport.Amqp.AmqpTransportHandler.OpenAsync(CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Client.Transport.ProtocolRoutingDelegatingHandler.OpenAsync(CancellationToken cancellationToken)
[...]

PR #2566, opened 2022-08-09
Remove extra tests and limit how long they'll run until timeout

drwill-ms:


e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -71,12 +58,6 @@ public async Task X509_InvalidDeviceId_Throw_UnauthorizedException_Mqtt_WebSocke
...
         }
 
-        [LoggedTestMethod, Timeout(TestTimeoutMilliseconds)]
-        public async Task X509_InvalidDeviceId_Throw_UnauthorizedException_Twice_Amqp()
-        {
-            await X509InvalidDeviceIdOpenAsyncTwiceTest(DeviceTransportType.Amqp).ConfigureAwait(false);
-        }
-
         [LoggedTestMethod, Timeout(TestTimeoutMilliseconds)]
         public async Task X509_InvalidDeviceId_Throw_UnauthorizedException_Twice_Amqp_TCP()
```
drwill-ms, line 62: What's the point of having another test that does it twice? Was there some bug where it wouldn't throw on a second try?

Seems like we should delete the "twice" tests, or remove the singles and just have the doubles.

@timtay-microsoft @abhipsaMisra ?

andyk-ms (approved):

e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
abhipsaMisra: I'm not finding any additional context in the PR associated with this change: https://github.com/Azure/azure-iot-sdk-csharp/pull/1136. It looks like this manual check was added in as a part of a bug fix (#1137 , #1130) but I don't have the context on what the bug/fix was.
In the absence of any additional context, having the "single run" test looks good to me.

abhipsaMisra (approved):

PR #2567, opened 2022-08-09
Remove .NET 5.0 Dependency

andyk-ms:
Removed NET 5.0 dependency
Added NET 6.0 Test
Completed unit tests with 100% pass rate for 613 tests in the group

### DIFF
--- a/vsts/test-release-nuget.yaml
+++ b/vsts/test-release-nuget.yaml
@@ -172,6 +172,8 @@ jobs:
       # Change maxParallel to 1 make builds run in serial rather than in parallel
       maxParallel: 100
       matrix:
+        .Net 6.0:
+          FRAMEWORK: net6.0
         .Net 5.0:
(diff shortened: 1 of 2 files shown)


vsts/test-release-nuget.yaml
```diff
@@ -172,6 +172,8 @@ jobs:
       # Change maxParallel to 1 make builds run in serial rather than in parallel
       maxParallel: 100
       matrix:
+        .Net 6.0:
+          FRAMEWORK: net6.0
```
timtay-microsoft, line 176: Do we need to make this same change in our nightly build yaml as well? Or are we already testing .NET 6.0 there?

timtay-microsoft (approved):

vsts/test-release-nuget.yaml
andyk-ms: Will make that update in the next change.

PR #2569, opened 2022-08-10
Merge latest from main

drwill-ms:


andyk-ms (approved):
The AMQP library was updated on another PR already.

PR #2571, opened 2022-08-10
Remove some common classes, spread helper functions to where they are used

timtay-microsoft:
Several classes in the common folder were only used by one library, so those classes have been moved to that library.

Some classes in the common folder had functions used only by one library and other functions used by a different library. Those functions have been split from the common class and moved to the library where they are used.

This is just a first step to removing the common folder, so we'll have more PRs like this in the future


iothub/device/src/Common/Amqp/ClientWebSocketTransport.cs
```diff
@@ -341,5 +341,24 @@ private void ThrowIfNotOpen()
 
             throw new AmqpException(AmqpErrorCode.IllegalState, null);
         }
+
+        private static void ValidateBufferBounds(byte[] buffer, int offset, int size)
+        {
+            if (buffer == null)
+            {
+                throw new ArgumentNullException(nameof(buffer));
+            }
```
drwill-ms, line 350: `Argument.AssertNotNull()`?

iothub/device/src/DirectMethod/MethodResponse.cs
```diff
@@ -44,12 +44,7 @@ public MethodResponse(int status)
             Justification = "Cannot change property types on public classes.")]
         public byte[] Result
         {
-            private set
-            {
-                Utils.ValidateDataIsEmptyOrJson(value);
-
-                _result = value;
-            }
+            private set => _result = value;
```
drwill-ms, line 47: Odd that set is above get. Mind swapping them?

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -1395,6 +1397,33 @@ internal static string PopulateMessagePropertiesFromMessage(string topicName, Me
             return msg;
         }
 
+        private static IReadOnlyDictionary<TKey, TValue> MergeDictionaries<TKey, TValue>(IDictionary<TKey, TValue>[] dictionaries)
+        {
+            // No item in the array should be null.
+            if (dictionaries == null || dictionaries.Any(item => item == null))
+            {
+                throw new ArgumentNullException(nameof(dictionaries), "Provided dictionaries should not be null");
+            }
```
drwill-ms, line 1406: I think there is an option for this in `Arguments.RequireNotNullOrEmpty<T>()` that I added to the device's Arguments class you could copy to the service one. The source for this was Azure.Core.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -1395,6 +1397,33 @@ internal static string PopulateMessagePropertiesFromMessage(string topicName, Me
...
+
+            var result = dictionaries.SelectMany(dict => dict)
+                .ToLookup(pair => pair.Key, pair => pair.Value)
+                .ToDictionary(group => group.Key, group => group.First());
+
+            return new ReadOnlyDictionary<TKey, TValue>(result);
+        }
+
+        private static DeliveryAcknowledgement ConvertDeliveryAckTypeFromString(string value)
```
drwill-ms, line 1415: We should consider making the enum deserializable so we don't have to maintain this mapping outside of the enum.

iothub/service/src/Http/HttpClientHelper.cs
```diff
@@ -885,5 +886,13 @@ internal static HttpMessageHandler CreateDefaultHttpMessageHandler(IWebProxy web
 
             return httpMessageHandler;
         }
+
+        private static string GetClientVersion()
```
drwill-ms, line 890: I don't see the point of decentralizing this when we now have 2 copies of the same code.

iothub/service/src/Http2/HttpRequestMessageFactory.cs
```diff
@@ -74,5 +75,13 @@ internal HttpRequestMessage CreateRequest(HttpMethod method, Uri relativeUri, Io
 
             return message;
         }
+
+        private static string GetClientVersion()
```
drwill-ms, line 79: 3 copies now.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
timtay-microsoft: This class will disappear once we move to MQTTNet, so I'm not going to bother here

drwill-ms (changes requested):
I see value in some of these changes, but question the DRY violations caused by others.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
timtay-microsoft: This class will disappear once we move to MQTTNet

iothub/service/src/Utilities/Utils.cs
```diff
@@ -0,0 +1,35 @@
...
+            if (offset < 0 || offset > buffer.Length || size <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(offset));
+            }
+
+            int remainingBufferSpace = buffer.Length - offset;
+            if (size > remainingBufferSpace)
+            {
+                throw new ArgumentOutOfRangeException(nameof(size));
+            }
```
drwill-ms, line 32: Would this fit better in Arguments.cs?

drwill-ms (approved):

iothub/service/src/Utilities/Utils.cs
timtay-microsoft: Good idea

PR #2573, opened 2022-08-10
Fix E2E test failures - part 1

abhipsaMisra:
Part 1 of addressing E2E test failures - failures down from [412](https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=113494&view=ms.vss-test-web.build-test-results-tab) to [127](https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=113558&view=ms.vss-test-web.build-test-results-tab).

Yesterday's nightly: https://azure-iot-sdks.visualstudio.com/azure-iot-sdks/_build/results?buildId=113531&view=ms.vss-test-web.build-test-results-tab

### DIFF
--- a/e2e/test/helpers/TestDevice.cs
+++ b/e2e/test/helpers/TestDevice.cs
@@ -167,24 +167,6 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
 
         public Client.IAuthenticationMethod AuthenticationMethod { get; private set; }
 
-        public IotHubDeviceClient CreateDeviceClient(IotHubClientOptions options = default)
-        {
-            IotHubDeviceClient deviceClient = null;
-
-            if (AuthenticationMethod == null)
-            {
-                deviceClient = IotHubDeviceClient.CreateFromConnectionString(ConnectionString, options);
-                s_logger.Trace($"{nameof(CreateDeviceClient)}: Created {nameof(IotHubDeviceClient)} {Device.Id} from connection string: {options.TransportSettings} Id={TestLogger.IdOf(deviceClient)}");
-            }
-            else
-            {
-                deviceClient = IotHubDeviceClient.Create(IotHubHostName, AuthenticationMethod, options);
-                s_logger.Trace($"{nameof(CreateDeviceClient)}: Created {nameof(IotHubDeviceClient)} {Device.Id} from IAuthenticationMethod: {options.TransportSettings} Id={TestLogger.IdOf(deviceClient)}");
-            }
-
-            return deviceClient;
-        }
-
         public IotHubDeviceClient CreateDeviceClient(IotHubClientOptions options = default, ConnectionStringAuthScope authScope = ConnectionStringAuthScope.Device)
         {
             IotHubDeviceClient deviceClient = null;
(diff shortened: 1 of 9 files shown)


iothub/device/src/Pipeline/TransportDelegatingHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+                IDelegatingHandler innerHandler = NextHandler;
+
+                // Ask the ContinuationFactory to attach the proper handler given the Context's ITransportSettings.
+                NextHandler = ContinuationFactory(Context, null);
+
+                innerHandler?.Dispose();
+            }
+        }
+
+        public override async Task WaitForTransportClosedAsync()
```
abhipsaMisra, line 70: This logic is required to recover the MQTT transport handler on reconnections. It doesn't look like AMQP relies on this logic though. Since we can get rid of this implementation once we move to MQTTNet, I'm not spending time optimizing this code.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -174,24 +174,16 @@ public override async Task<Message> ReceiveMessageAsync(CancellationToken cancel
...
                 cancellationToken.ThrowIfCancellationRequested();
-
-                using var ctb = new CancellationTokenBundle(_transportSettings.DefaultReceiveTimeout, cancellationToken);
-                message = await _amqpUnit.ReceiveMessageAsync(ctb.Token).ConfigureAwait(false);
-
-                if (message != null)
-                {
-                    break;
-                }
+                return await _amqpUnit.ReceiveMessageAsync(cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 180: We await this call until the cancellation token expires, so this loop is not necessary.

e2e/test/helpers/TestDevice.cs
```diff
@@ -167,24 +167,6 @@ public string ConnectionString
 
         public Client.IAuthenticationMethod AuthenticationMethod { get; private set; }
 
-        public IotHubDeviceClient CreateDeviceClient(IotHubClientOptions options = default)
```
abhipsaMisra, line 170: CreateDeviceClient had the following overloads:
public IotHubDeviceClient CreateDeviceClient(IotHubClientOptions options = default)
public IotHubDeviceClient CreateDeviceClient(IotHubClientOptions options = default, ConnectionStringAuthScope authScope = ConnectionStringAuthScope.Device)

They compile to the same code, causing ambiguity when CreateDeviceClient() is called. So, I'm removing the redundant overload.

e2e/test/iothub/method/MethodFaultInjectionTests.cs
```diff
@@ -219,7 +219,7 @@ private async Task ServiceSendMethodAndVerifyResponseAsync(string deviceName, st
                     var directMethodRequest = new DirectMethodRequest()
                     {
                         MethodName = methodName,
-                        Payload = respJson,
+                        Payload = reqJson,
```
abhipsaMisra, line 222: The request should have reqJson as the payload. Looks like this was a typo while refactoring.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: No need for the token bundle including the default receive timeout?

drwill-ms (approved):

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
abhipsaMisra: Since we're using the user supplied cancellation token to signal when this API should be cancelled, we shouldn't be needing the default receive timeout anymore, right?

PR #2574, opened 2022-08-10
refactor(iot-device): Add ConnectionInfo class and expose to users

brycewang-microsoft:
Add ConnectionInfo class containing properties "State", "ChangeReason" and "StateLastChangedOnUtc"


timtay-microsoft:
I have mixed feelings about this one. My worry is that customers will use this state to write code like

```csharp
while (true)
{
    await deviceClient.OpenAsync();
    while (deviceClient.LastConnectionState == Connected)
    {
        await deviceClient.SendEventAsync(...)
    }
}
```

without understanding that a connection loss could occur within between the LastConnectionState check and the SendEventAsync() call. This would lead to some confusing errors for users if we remove the implicit Open() call from operations like SendEventAsync().

How exactly do you envision users using these newly exposed "last connection state" properties?

drwill-ms:
> I have mixed feelings about this one. My worry is that customers will use this state to write code like
> 
> ```cs
> while (true)
> {
>     await deviceClient.OpenAsync();
>     while (deviceClient.LastConnectionState == Connected)
>     {
>         await deviceClient.SendEventAsync(...)
>     }
> }
> ```
> 
> without understanding that a connection loss could occur within between the LastConnectionState check and the SendEventAsync() call. This would lead to some confusing errors for users if we remove the implicit Open() call from operations like SendEventAsync().
> 
> How exactly do you envision users using these newly exposed "last connection state" properties?

The notion is that customers effectively need to cache this info anyway (from the callback), so they can avoid calls they know will fail when not connected, like we do in the reconnection sample in the telemetry loop.

iothub/device/src/IotHubDeviceClient.cs
```diff
@@ -109,6 +109,22 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
...
+            get => InternalClient._lastConnectionState;
+        }
+
+        /// <summary>
+        /// Last connection state change reason.
+        /// </summary>
+        public ConnectionStateChangeReason LastConnectionStateChangeReason
+        {
+            get => InternalClient._lastConnectionStateChangeReason;
+        }
```
drwill-ms, line 126: I'd rather expose this as a class that has cohesive properties.

e.g. `IotHubDeviceClient.LatestConnectionState` which would have properties `State`, `ChangeReason`, and `ChangedDateTimeUtc`.

brycewang-microsoft: > I'd rather expose this as a class that has cohesive properties.
> 
> e.g. `IotHubDeviceClient.LatestConnectionState` which would have properties `State`, `ChangeReason`, and `ChangedDateTimeUtc`.

Yeah, I planned to do that in another PR originally, but sure, I can also do that here

iothub/device/src/IotHubDeviceClient.cs
```diff
@@ -109,6 +109,22 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
         public void SetConnectionStateChangeHandler(Action<ConnectionState, ConnectionStateChangeReason> stateChangeHandler)
             => InternalClient.SetConnectionStateChangeHandler(stateChangeHandler);
 
+        /// <summary>
+        /// Last connection state.
+        /// </summary>
+        public ConnectionState LastConnectionState
+        {
+            get => InternalClient._lastConnectionState;
+        }
```
drwill-ms, line 118: Btw, you can write it this way in just one line.

```suggestion
        public ConnectionState LastConnectionState => InternalClient._lastConnectionState;
```

iothub/device/src/IotHubDeviceClient.cs
drwill-ms: Fine, I'm okay with that, but it seems like more work to make such a small number of changes in 2 PRs. Up to you.

drwill-ms (approved):

iothub/device/src/LastConnectionState.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        /// <remark>
+        /// Defaults to <see cref="ConnectionStateChangeReason.ClientClose"/>.
+        /// </remark>
+        public ConnectionStateChangeReason ChangeReason { get; set; } = ConnectionStateChangeReason.ClientClose;
+
+        /// <summary>
+        /// Timestamp in UTC when the last connection state was changed.
+        /// </summary>
+        public DateTime ChangedDateTimeUtc { get; set; } = DateTime.UtcNow;
```
drwill-ms, line 31:
```suggestion
        public DateTimeOffset ChangedOnUtc { get; internal set; } = DateTime.UtcNow;
```

iothub/device/src/LastConnectionState.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class LastConnectionState
```
drwill-ms, line 10:
```suggestion
    public class ConnectionInfo
```

iothub/device/src/LastConnectionState.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    ///
```
drwill-ms, line 8: The connection information since the last state change.

iothub/device/src/LastConnectionState.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class LastConnectionState
+    {
+        /// <summary>
+        /// Connection state supported by the client.
```
drwill-ms, line 13:
```suggestion
        /// The current connection state of the device or module.
```

iothub/device/src/LastConnectionState.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        /// <remark>
+        /// Defaults to <see cref="ConnectionState.Disconnected"/>.
+        /// </remark>>
+        public ConnectionState State { get; set; } = ConnectionState.Disconnected;
+
+        /// <summary>
+        /// Connection state change reason supported by the client.
```
drwill-ms, line 21:
```suggestion
        /// The reason for the current connection state change.
```

iothub/device/src/IotHubModuleClient.cs
```diff
@@ -138,6 +138,21 @@ public int DiagnosticSamplingPercentage
         public void SetConnectionStateChangeHandler(Action<ConnectionState, ConnectionStateChangeReason> stateChangeHandler)
             => InternalClient.SetConnectionStateChangeHandler(stateChangeHandler);
 
+        /// <summary>
+        /// Last connection state.
+        /// </summary>
+        public ConnectionState LastConnectionState => InternalClient._lastConnectionState.State;
```
drwill-ms, line 144: Store as one property: ConnectionInfo. Set it at the same time so no timing issues if someone checked between assignment of the 3 properties.

iothub/device/src/InternalClient.cs
```diff
@@ -1186,16 +1184,17 @@ internal void OnConnectionStateChanged(ConnectionState state, ConnectionStateCha
                 if (Logging.IsEnabled)
                     Logging.Enter(this, state, reason, nameof(OnConnectionStateChanged));
 
-                if (_lastConnectionState != state
-                    || _lastConnectionStateChangeReason != reason)
+                if (_lastConnectionState.State != state
+                    || _lastConnectionState.ChangeReason != reason)
                 {
                     _connectionStateChangeHandler?.Invoke(state, reason);
```
drwill-ms, line 1190: Change callback to take the ConnectionInfo class?

drwill-ms (changes requested):
Update the sample as well.

iothub/device/src/InternalClient.cs
```diff
@@ -1186,16 +1184,16 @@ internal void OnConnectionStateChanged(ConnectionState state, ConnectionStateCha
...
+                    || _connectionInfo.ChangeReason != reason)
                 {
                     _connectionStateChangeHandler?.Invoke(state, reason);
                 }
             }
             finally
             {
-                _lastConnectionState = state;
-                _lastConnectionStateChangeReason = reason;
+                _connectionInfo = new ConnectionInfo(state, reason, DateTime.UtcNow);
```
brycewang-microsoft, line 1195: Set properties in `ConnectionInfo` at the same time so no timing issues if someone checked between assignment of multiple properties.

iothub/device/src/ConnectionInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// The connection information since the last state change.
+    /// </summary>
+    public class ConnectionInfo
+    {
+        internal ConnectionInfo() { }
```
drwill-ms, line 12: not needed?

iothub/device/src/ConnectionInfo.cs
```diff
@@ -0,0 +1,42 @@
...
+            ChangedOnUtc = changedOnUtc;
+        }
+
+        /// <summary>
+        /// The current connection state of the device or module.
+        /// </summary>
+        /// <remark>
+        /// Defaults to <see cref="ConnectionState.Disconnected"/>.
+        /// </remark>>
+        public ConnectionState State { get; internal set; } = ConnectionState.Disconnected;
```
drwill-ms, line 27: Can these be readonly now?

```suggestion
        public ConnectionState State { get; } = ConnectionState.Disconnected;
```

iothub/device/src/InternalClient.cs
drwill-ms, line 1195:
```suggestion
                _connectionInfo = new ConnectionInfo(state, reason, DateTimeOffset.UtcNow);
```

drwill-ms (approved):

PR #2575, opened 2022-08-10
Move more common classes into the only library that uses them

timtay-microsoft:
Remove some unused constants from CommonConstants, remove some unnecessary client-side validation of user generated device/module symmetric keys


provisioning/service/src/Microsoft.Azure.Devices.Provisioning.Service.csproj
```diff
@@ -46,9 +46,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="$(common)\service\CommonResources.cs" Link="Common\CommonResources.cs" />
     <Compile Include="$(common)\service\Resources.Designer.cs" Link="Common\Resources.Designer.cs" />
-    <Compile Include="$(common)\service\StringValidationHelper.cs" Link="Common\StringValidationHelper.cs" />
```
timtay-microsoft, line 51: Both these classes were unused in the provisioning service client

drwill-ms (approved):

PR #2576, opened 2022-08-11
E2E test failures - part 2

abhipsaMisra:
The only fix here is increasing the cancelation token expiration time for our fault injection message.

I'm removing some ignored tests that we've never implemented the feature for, eg. device client allowing the connection pool size to change post initialization. This feature has never been supported.


e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -82,7 +82,7 @@ public static async Task ActivateFaultInjectionAsync(IotHubClientTransportSettin
 
             try
             {
-                using var cts = new CancellationTokenSource(delay);
+                using var cts = new CancellationTokenSource(LatencyTimeBuffer);
```
abhipsaMisra, line 85: In some cases, it is taking over 5secs for the fault injection message to be ack'ed; causing tests to fail with the error "fault was not injected". Increasing the cancellation token validity to 10 secs.

timtay-microsoft (approved):

PR #2577, opened 2022-08-11
Remove HttpContentExtensions class

timtay-microsoft:
This class provided overloads of HttpContent read/copy as stream/byte/string that took cancellation tokens, but the overloads didn't actually do anything with the cancellation token.

It used to pass the cancellation token into .NET standard 2.1 overloads of the HttpContent methods, but since we have removed .NET standard 2.1, this class no longer has any benefit


iothub/service/src/Http2/HttpMessageHelper2.cs
```diff
@@ -69,11 +69,10 @@ internal static async Task ValidateHttpResponseStatusAsync(HttpStatusCode expect
         /// </summary>
         /// <typeparam name="T">The type of object to deserialize the response payload into.</typeparam>
         /// <param name="response">The HTTP response containing the payload to deserialize.</param>
-        /// <param name="cancellationToken">Cancellation token.</param>
         /// <returns>The deserialized object.</returns>
-        internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage response, CancellationToken cancellationToken)
+        internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage response)
```
timtay-microsoft, line 73: Most of the files changed in this PR are because this function no longer needs a cancellation token.

abhipsaMisra (approved):

drwill-ms (approved):

tmahmood-microsoft (approved):

PR #2578, opened 2022-08-11
Move more classes out of common, delete unused common classes

timtay-microsoft:
FxCop.cs and Resources.Designer.cs appear to be unused. StreamDisposalResponsibility is used in both IoT Hub device and IoT Hub service clients, and TaskCompletionSource is only used in IoT Hub device client's MQTT layer.

It should be noted that TaskCompletionSource will be deleted when we move to MQTTNet and StreamDisposalResponsibility will disappear when we make the Message classes not implement IDisposable


iothub/service/src/Messaging/StreamDisposalResponsibility.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices
+{
+    // Enum that indicates if the Stream used to initialize an IoT hub message
+    // should be disposed by the application or by the SDK.
+    internal enum StreamDisposalResponsibility
```
drwill-ms, line 8: FYI @tmahmood-microsoft this will go away when you work on removing IDisposable from the Message class.

drwill-ms (approved):

PR #569, opened 2018-07-27
Add proxy support

abhipsaMisra:
Issue: #477 - Proxy support for Azure IoT Hub and Provisioning SDKs

With the release of .NET Core 2.1, we are implementing proxy support for the C# sdk. 
Currently, the sdk picks up the proxy settings as per the OS default proxy settings. With this change, we will be exposing an `IWebProxy` member to all our IoT SDK clients, allowing for setting the proxy explicitly in the application.

The following 3 scenarios will be supported:

1. Proxy not set in the code (sdk will pick the OS default proxy settings)
2. Proxy = null (sdk will not use proxy)
3. Proxy = set in the code (sdk will use custom proxy specified).

#### Device Client, Module Client
```C# 
using System.Net;

// s_proxyServerAddress = ...
// s_deviceConnectionString = …
// s_moduleConnectionString = ...
Http1TransportSettings transportSettings = new Http1TransportSettings();
transportSettings.Proxy = new WebProxy(s_proxyServerAddress);
ITransportSettings[] transportSettings = new ITransportSettings[] { transportSettings };

DeviceClient deviceClient = DeviceClient.CreateFromConnectionString(s_deviceConnectionString, transportSettings);
ModuleClient moduleClient = ModuleClient.CreateFromConnectionString(s_moduleConnectionString, transportSettings);
```
#### Job Client, Registry Manager
```C#
using System.Net;
// s_proxyServerAddress= ...
// s_connectionString = ...
HttpTransportSettings customTransportSettings = new HttpTransportSettings();
customTransportSettings.Proxy = new WebProxy(s_proxyServerAddress);

JobClient jobClient = JobClient.CreateFromConnectionString(s_connectionString, customTransportSettings);
RegistryManager registryManager = RegistryManager.CreateFromConnectionString(s_connectionString, customTransportSettings);
```
#### Service Client
```C#
using System.Net;
// s_proxyServerAddressAmqp = ...
// s_proxyServerAddressHttp = …
// s_connectionString = ...
ServiceClientTransportSettings transportSettings = new ServiceClientTransportSettings();
transportSettings.AmqpProxy = new WebProxy(s_proxyServerAddressAmqp);
transportSettings.HttpProxy = new WebProxy(s_proxyServerAddressHttp);

ServiceClient serviceClient = ServiceClient.CreateFromConnectionString(s_connectionString, transportSettings);
```
#### Provisioning Client
```C#
using System.Net;
// s_proxyServerAddress = ...
[...]

PR #2581, opened 2022-08-11
Move DefaultWebProxySettings from Common to each package that was using it

timtay-microsoft:
~~This class served as a stubbed out IWebProxy implementation that we used as the value to signal that no proxy is configured. To simplify this, we can use "null" as that value instead and delete this class~~

To preserve the currently functionality outlined by @abhipsaMisra below, I've opted to simply clone this DefaultProxySettings class to each library that was using it.


abhipsaMisra:
Make sure that all three scenarios are still supported:

> Currently, the sdk picks up the proxy settings as per the OS default proxy settings. With this change, we will be exposing an IWebProxy member to all our IoT SDK clients, allowing for setting the proxy explicitly in the application.
>
> The following 3 scenarios will be supported:
>
> Proxy not set in the code (sdk will pick the OS default proxy settings)
> Proxy = null (sdk will not use proxy)
> Proxy = set in the code (sdk will use custom proxy specified).

https://github.com/Azure/azure-iot-sdk-csharp/pull/569

timtay-microsoft:
> Make sure that all three scenarios are still supported:
> 
> > Currently, the sdk picks up the proxy settings as per the OS default proxy settings. With this change, we will be exposing an IWebProxy member to all our IoT SDK clients, allowing for setting the proxy explicitly in the application.
> > The following 3 scenarios will be supported:
> > Proxy not set in the code (sdk will pick the OS default proxy settings)
> > Proxy = null (sdk will not use proxy)
> > Proxy = set in the code (sdk will use custom proxy specified).
> 
> #569

Interesting... I didn't know that we supported a "Let the OS decide" proxy setting here. Let me rework this a bit then

iothub/service/src/Http2/HttpClientFactory.cs
```diff
@@ -53,9 +53,9 @@ internal static HttpClient Create(string hostName, IotHubServiceClientOptions op
             httpMessageHandler.SslProtocols = options.SslProtocols;
             httpMessageHandler.CheckCertificateRevocationList = options.CertificateRevocationCheck;
 
-            if (options.Proxy != DefaultWebProxySettings.Instance)
+            if (options.Proxy != null)
```
timtay-microsoft, line 56:
```suggestion
            if (options.Proxy != DefaultWebProxySettings.Instance)
```

iothub/service/src/Http2/HttpClientFactory.cs
```diff
@@ -55,7 +55,7 @@ internal static HttpClient Create(string hostName, IotHubServiceClientOptions op
 
             if (options.Proxy != DefaultWebProxySettings.Instance)
             {
-                httpMessageHandler.UseProxy = options.Proxy != null;
+                httpMessageHandler.UseProxy = true;
```
timtay-microsoft, line 58:
```suggestion
                httpMessageHandler.UseProxy = options.Proxy != null;
```

drwill-ms (approved):

PR #2582, opened 2022-08-12
Move classes out of IoT Hub device client's "common" folder

timtay-microsoft:
These classes are moved to the most appropriate spot based on what set of folders contained classes that referenced them. Some classes were deleted because they were unused.

None of these classes were used outside of the IoT Hub device client package, so I'm not sure why this "common" folder existed in the first place


iothub/device/src/Transport/Amqp/AccessRights.cs
```diff
@@ -6,7 +6,7 @@
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 
-namespace Microsoft.Azure.Devices.Client
+namespace Microsoft.Azure.Devices.Client.Transport.Amqp
```
timtay-microsoft, line 9:
```suggestion
namespace Microsoft.Azure.Devices.Client
```

iothub/device/src/Transport/Http/CustomHeaderConstants.cs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.Azure.Devices.Client
+namespace Microsoft.Azure.Devices.Client.Transport
```
drwill-ms, line 4: I'd like to not have namespaces match the folders. Unless the files in there already do, can we leave this the same as the root namespace?

drwill-ms (approved):

iothub/device/src/Transport/Http/CustomHeaderConstants.cs
timtay-microsoft: Sure I can revert this change

PR #2583, opened 2022-08-12
Move classes out of IoT Hub service client's "common" folder

timtay-microsoft:
These classes are moved to the most appropriate spot based on what set of folders contained classes that referenced them. Some classes were deleted because they were unused.

None of these classes were used outside of the IoT Hub service client package, so I'm not sure why this "common" folder existed in the first place


iothub/service/src/Resources/Fx.cs
```diff
@@ -64,7 +64,7 @@ public static void AssertAndThrow(bool condition, string description)
         public static Exception AssertAndThrow(string description)
         {
             Assert(description);
-            throw Exception.AsError(new AssertionFailedException(description));
+            throw Exception.AsError(new FatalException(description));
```
drwill-ms, line 67: What does this change mean?

timtay-microsoft: I'm not completely sure, but we don't use this AssertionFailedException type anywhere else and we do use this FatalException type elsewhere in this class. I'd like to delete the AssertionFailedException class since it only has this one use.

drwill-ms: Will it impact customers?

timtay-microsoft: Not at all. The only place this function is called is by the Message class if the user attempts to clone a message that uses a Stream for its body. As I understand, that use case will disappear when we remove streams from Message classes, so this function won't even be called anymore.

drwill-ms (approved):

PR #2586, opened 2022-08-12
Add FaultInjection as a separate test category

abhipsaMisra:


timtay-microsoft (approved):

build.ps1
```diff
@@ -62,7 +62,7 @@ Param(
     [switch] $clean,
     [switch] $build,
     [switch] $unittests,
-    [switch] $prtests,
```
drwill-ms, line 65: I think PR Tests was a better name for how we're using it. They aren't BVTs, strictly speaking. Our integration build and nightly builds run all the tests that validate the build.

drwill-ms (approved):

build.ps1
abhipsaMisra: I don't mind changing it back to PR tests.

PR #2590, opened 2022-08-15
Reorganize last of the common code

timtay-microsoft:
Logging.Common was used by several libraries, so it was copied to each.

ExceptionHandlingHelper was only used by IoT hub service client, so it was moved there.


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -74,7 +67,6 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.1" />
```
timtay-microsoft, line 77: This appears to be unused, but correct me if I'm wrong here.

brycewang-microsoft (approved):

PR #2593, opened 2022-08-16
Remove flakey assertions from query test

timtay-microsoft:
start time and end time may not be set yet if the job has not started and/or ended yet


brycewang-microsoft (approved):

e2e/test/iothub/service/QueryClientE2ETests.cs
```diff
@@ -130,8 +130,6 @@ public async Task JobQuery_QueryWorks()
             // That's why these checks aren't more specific.
             queriedJob.JobId.Should().NotBeNull();
             queriedJob.JobType.Should().NotBeNull();
-            queriedJob.StartTimeUtc.Should().NotBeNull();
-            queriedJob.EndTimeUtc.Should().NotBeNull();
```
drwill-ms, line 134: Why does this make the test flaky?

drwill-ms: Maybe we could base this check on the job state?

PR #2594, opened 2022-08-16
Fix build error in Tpm simulator

abhipsaMisra:
The TPM simulato sample was referencing a nuget package instead of the sourcecode project. This was causing the build to fail since Microsoft.Azure.Devices.Authentication nuget isn't released yet.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -259,8 +259,13 @@ try
         # We must disable package testing here as the E2E csproj may reference new APIs that are not available in existing NuGet packages.
         $packageTempPath = $env:AZURE_IOT_LOCALPACKAGES
         $env:AZURE_IOT_LOCALPACKAGES = ""
+        
         # SDK binaries
         BuildProject . "Azure IoT C# SDK Solution"
+
+        # Samples
+        BuildProject security\tpm\samples "SecurityProvider for TPM Samples"
+
         $env:AZURE_IOT_LOCALPACKAGES = $packageTempPath
     }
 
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -365,9 +370,6 @@ try
         RunTests "E2E tests" -filterTestCategory $testCategory -framework $framework
 
         $verbosity = $oldVerbosity
-
-        # Samples
-        BuildProject security\tpm\samples "SecurityProvider for TPM Samples"
```
abhipsaMisra, line 370: moving this from the E2E test section to the project build section.

build.ps1
```diff
@@ -259,8 +259,13 @@ try
         # We must disable package testing here as the E2E csproj may reference new APIs that are not available in existing NuGet packages.
         $packageTempPath = $env:AZURE_IOT_LOCALPACKAGES
         $env:AZURE_IOT_LOCALPACKAGES = ""
+        
         # SDK binaries
         BuildProject . "Azure IoT C# SDK Solution"
+
+        # Samples
+        BuildProject security\tpm\samples "SecurityProvider for TPM Samples"
```
abhipsaMisra, line 267: We need to specify this project explicitly since it isn't a part of our sourcecode solution

timtay-microsoft (approved):
Why haven't we released the TPM package? Does it just contain the TPM simulator code and not the actual TPM code?

abhipsaMisra:
> Why haven't we released the TPM package? Does it just contain the TPM simulator code and not the actual TPM code?

Microsoft.Azure.Devices.Provisioning.Security.Tpm has been released and it contains the TPM HSM code. The released version doesn't contain authentication classes rework that we've done in v2.

PR #2599, opened 2022-08-17
Update Timeout for failed tests

andyk-ms:
Updated timeout for failed tests from last pipeline run.

### DIFF
--- a/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
+++ b/e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
@@ -42,9 +42,10 @@ namespace Microsoft.Azure.Devices.E2ETests.Iothub.Service
             JobStatus.Unknown,
         };
 
-        [LoggedTestMethod, Timeout(TestTimeoutMilliseconds)]
+        [LoggedTestMethod, Timeout(LongRunningTestTimeoutMilliseconds)]
         [TestCategory("LongRunning")]
         [DoNotParallelize] // the number of jobs that can be run at a time are limited anyway
+        [Ignore]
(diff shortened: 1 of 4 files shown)


timtay-microsoft (approved):

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
```diff
@@ -42,9 +42,10 @@ public class RegistryManagerExportDevicesTests : E2EMsTestBase
             JobStatus.Unknown,
         };
 
-        [LoggedTestMethod, Timeout(TestTimeoutMilliseconds)]
+        [LoggedTestMethod, Timeout(LongRunningTestTimeoutMilliseconds)]
         [TestCategory("LongRunning")]
         [DoNotParallelize] // the number of jobs that can be run at a time are limited anyway
+        [Ignore]
```
drwill-ms, line 48: Why ignore?

abhipsaMisra (approved):

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
andyk-ms: Cannot run in the matrix tests in the build pipeline run because the [DoNotParallelize] flag do not work in that setup.
We can either skip this test or need to create an separate job to run these type of tests on each framework in sequence.

drwill-ms (approved):

e2e/test/iothub/service/RegistryManagerExportDevicesTests.cs
andyk-ms: Created a task for us to improve these ignored tests.

PR #2600, opened 2022-08-17
Remove HTTP as a valid client initialization transport

drwill-ms:


iothub/device/src/IotHubClientOptions.cs
```diff
@@ -21,16 +21,21 @@ public IotHubClientOptions()
...
         /// </summary>
-        /// <param name="transportSettings">The transport settings to use (i.e., <see cref="IotHubClientMqttSettings"/>,
-        /// <see cref="IotHubClientAmqpSettings"/>, or <see cref="IotHubClientHttpSettings"/>).</param>
+        /// <param name="transportSettings">The transport settings to use (i.e., <see cref="IotHubClientMqttSettings"/> or
+        /// <see cref="IotHubClientAmqpSettings"/>.</param>
         /// <exception cref="ArgumentNullException">When <paramref name="transportSettings"/> is null.</exception>
         public IotHubClientOptions(IotHubClientTransportSettings transportSettings)
         {
             TransportSettings = transportSettings ?? throw new ArgumentNullException(nameof(transportSettings));
+            if (transportSettings is not IotHubClientMqttSettings
```
drwill-ms, line 30: Choosing, for now, to simply reject any other option. HttpTransportSettings still inherits from TransportSettings, so if we chose to bring support back it would not be a breaking change.

timtay-microsoft (approved):

abhipsaMisra (approved):

PR #2603, opened 2022-08-18
refactor(iot-device): Add enum RecommendedAction and use it as a new property in ConnectionInfo class

brycewang-microsoft:
For a specific combination of `ConnectionStatus` and `ConnectionStatusChangeReason`, there will be a `RecommendedAction` exposed to the users so that they are aware of the recommended actions to taken further, like what we are doing in the device reconnection sample [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/device/samples/DeviceReconnectionSample/DeviceReconnectionSample.cs#L190-L258).


iothub/device/src/ConnectionInfo.cs
```diff
@@ -44,5 +46,50 @@ internal ConnectionInfo(ConnectionStatus status, ConnectionStatusChangeReason ch
...
+
+        /// <summary>
+        /// Recommended actions for users to take upon different ConnectionStatus and ConnectionStatusChangeReason.
+        /// </summary>
+        /// <remark>
+        /// Defaults to <see cref="RecommendedAction.DisposeAndOpenIfWish"/>.
+        /// </remark>>
+        public RecommendedAction RecommendedAction { get; internal set; }
+
+        private RecommendedAction GetRecommendedActionBasedOnConnectionStatusAndChangeReason(ConnectionStatus status, ConnectionStatusChangeReason changeReason)
```
brycewang-microsoft, line 58: This is the mapping between <<`ConnectionStatus`, `ConnectionStatusChangeReason`>, `RecommendedAction`>

iothub/device/src/ConnectionInfo.cs
```diff
@@ -15,13 +15,15 @@ internal ConnectionInfo()
             Status = ConnectionStatus.Disconnected;
             ChangeReason = ConnectionStatusChangeReason.ClientClose;
             StatusLastChangedOnUtc = DateTimeOffset.UtcNow;
+            RecommendedAction = RecommendedAction.DefaultAction;
         }
 
-        internal ConnectionInfo(ConnectionStatus status, ConnectionStatusChangeReason changeReason, DateTimeOffset changedOnUtc)
```
brycewang-microsoft, line 20: It's unnecessary to pass changedOnUtc as a parameter into the constructor here, instead we can simply set it to `DateTimeOffset.UtcNow`

iothub/device/src/RecommendedAction.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Recommended actions for users to take upon different ConnectionStatus and ConnectionStatusChangeReason
```
drwill-ms, line 7:
```suggestion
    /// Recommended actions for device applications to take upon different connection change events.
```

iothub/device/src/RecommendedAction.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    public enum RecommendedAction
+    {
+        /// <summary>
+        /// It's recommended to take no more actions as all operations on the device or module client will be carried out as normal.
+        /// This is expected when <see cref="ConnectionStatus.Connected"/> and <see cref="ConnectionStatusChangeReason.ConnectionOk"/>.
+        /// </summary>
+        NoActionWhenNormal,
```
drwill-ms, line 15: NoAction seems wrong given the action they should be taking is resuming normal device operations, like sending telemetry.

It is only NoAction with respect to addressing the connection.

iothub/device/src/RecommendedAction.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Recommended actions for users to take upon different ConnectionStatus and ConnectionStatusChangeReason
+    /// </summary>
+    public enum RecommendedAction
```
drwill-ms, line 9: I think the default value (first enum) should be what the state the SDK is in before OpenAsync is called.

iothub/device/src/RecommendedAction.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <list type="bullet">
+        /// <item><description>The client has been closed gracefully with <see cref="ConnectionStatus.Disabled"/> and <see cref="ConnectionStatusChangeReason.ClientClose"/>.
+        /// </description></item>
+        /// <item><description>The client has been disconnected because the retry policy expired with <see cref="ConnectionStatus.Disconnected"/> and <see cref="ConnectionStatusChangeReason.RetryExpired"/>.
+        /// </description></item>
+        /// <item><description>The client has been disconnected due to a non-retry-able exception with <see cref="ConnectionStatus.Disconnected"/> and <see cref="ConnectionStatusChangeReason.CommunicationError"/>.
+        /// Inspect the exception for details.</description></item>
+        /// </list>
+        /// </summary>
+        DisposeAndOpenIfWish,
```
drwill-ms, line 36: ReinitializeClient?

iothub/device/src/RecommendedAction.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// The supplied credentials are invalid. Use a valid one and run again.
+        /// This is expected when <see cref="ConnectionStatus.Disconnected"/> and <see cref="ConnectionStatusChangeReason.BadCredential"/>.
+        /// </summary>
+        UseValidCredential,
+
+        /// <summary>
+        /// Fix the device status in the Iot hub and then create a new device client instance as the device has been deleted or marked as disabled (on your hub instance).
+        /// This is expected when <see cref="ConnectionStatus.Disconnected"/> and <see cref="ConnectionStatusChangeReason.DeviceDisabled"/>.
+        /// </summary>
+        FixDeviceStatus,
```
drwill-ms, line 48: InvestigateDeviceEnabledState?

iothub/device/src/RecommendedAction.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// The combination of ConnectionStatus and ConnectionStatusChangeReason is not expected, contact the client library team with logs.
+        /// </summary>
+        ContactUs,
+
+        /// <summary>
+        /// The default recommended action. This is being used for the initialization of ConnectionInfo only.
+        /// </summary>
+        DefaultAction,
```
drwill-ms, line 58: Should be the first enum (default value) then. But in a previous comment, I advocate for a different default.

iothub/device/samples/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -186,27 +186,17 @@ private async void ConnectionStatusChangeHandler(ConnectionInfo connectionInfo)
             var status = connectionInfo.Status;
             var reason = connectionInfo.ChangeReason;
             Console.WriteLine($"Connection status changed: status={status}, reason={reason}");
+            Console.WriteLine($"The recommended actions upon the current status and change reason is {connectionInfo.RecommendedAction}");
```
andyk-ms, line 189: "recommended action"

iothub/device/src/ConnectionInfo.cs
```diff
@@ -15,13 +15,15 @@ internal ConnectionInfo()
...
-        { 
+        internal ConnectionInfo(ConnectionStatus status, ConnectionStatusChangeReason changeReason)
+        {
             Status = status;
             ChangeReason = changeReason;
-            StatusLastChangedOnUtc = changedOnUtc;
+            StatusLastChangedOnUtc = DateTimeOffset.UtcNow;
+            RecommendedAction = GetRecommendedActionBasedOnConnectionStatusAndChangeReason(status, changeReason);
```
andyk-ms, line 26: Long and clear method..... but what about a shorter one?
GetRecommendedAction(connectionStatus, changeReason)

iothub/device/src/RecommendedAction.cs
andyk-ms: NoRecommendation?

iothub/device/src/RecommendedAction.cs
andyk-ms: Do we need to consider the status is always retrying and need to recommend checking other connection related settings?

Also, do we consider something like a "retry count" somewhere?

iothub/device/src/RecommendedAction.cs
andyk-ms: Do we have a defaultAction when it is normal and another defaultAction when it is more catch-all?
Isn't the catch-all is the case having ConnectionStatus and ConnectionStatusChangeReason being not expected?

iothub/device/src/RecommendedAction.cs
brycewang-microsoft: Changed this to `PerformNormally` as per our discussion

iothub/device/src/RecommendedAction.cs
brycewang-microsoft: I don't have a better name of the default one, so keep it as-is. Open to suggestions. :)

iothub/device/src/RecommendedAction.cs
brycewang-microsoft: `DisposeAndOpenIfWish` was not for retriable exceptions, and has been removed now to shorten the list of enum as per our discussion.

iothub/device/src/RecommendedAction.cs
brycewang-microsoft: this has been removed

iothub/device/src/RecommendedAction.cs
brycewang-microsoft: I thought the default value was a "placeholder" for the property `RecommendedAction` when initialize a `ConnectionInfo` instance. I do have `ContactUs` in enum for the catch-all cases.

iothub/device/samples/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -181,79 +181,47 @@ private async Task OnC2dMessageReceivedAsync(Message message, object context)
...
+                // If we had a backup, we can try using that.
+                _deviceConnectionStrings.RemoveAt(0);
+                Console.WriteLine($"The current connection string is invalid. Trying another.");
+                await InitializeAndSetupClientAsync(s_cancellationTokenSource.Token);
+                return;
+            }
 
-            switch (status)
+            // Otherwise, we follow the SDK's recommendation.
+            switch (connectionInfo.RecommendedAction)
```
drwill-ms, line 205: This is what @timtay-microsoft was looking for.

iothub/device/src/ConnectionStatus.cs
```diff
@@ -4,32 +4,36 @@
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// Connection status supported by the client.
+    /// The connection status from a connection change event.
```
drwill-ms, line 7: Reordered this and the other connection status classes to report enums in the general lifecycle order.

Comments all updated too.

iothub/device/src/ConnectionStatusChangeReason.cs
```diff
@@ -4,58 +4,61 @@
...
         /// <summary>
-        /// The client is connected, and ready to be used.
-        /// <para>This is returned with a connection status of <see cref="ConnectionStatus.Connected"/>.</para>
+        /// The client has not been opened or has been closed gracefully.
         /// </summary>
-        ConnectionOk,
+        /// <remarks>
+        /// To perform more operations on the client, call <see cref="IotHubDeviceClient.Dispose()"/> and then re-initialize the client.
+        /// </remarks>
+        ClientClosed,
```
drwill-ms, line 17: Renamed to ClientClose**d**

iothub/device/src/ConnectionStatus.cs
```diff
@@ -4,32 +4,36 @@
...
         /// <summary>
-        /// The device connection was closed.
-        /// <para>If you want to perform more operations on the device client, you should <see cref="IotHubDeviceClient.Dispose()"/> and then re-initialize the client.</para>
+        /// The device connection was closed gracefully.
         /// </summary>
-        Disabled,
+        /// <remarks>
+        /// To perform more operations on the client, call <see cref="IotHubDeviceClient.Dispose()"/> and then re-initialize the client.
+        /// </remarks>
+        Closed,
```
drwill-ms, line 37: Renamed from Disabled, which frankly made no sense to me.

drwill-ms (approved):

iothub/device/src/ConnectionInfo.cs
```diff
@@ -36,13 +36,51 @@ internal ConnectionInfo(ConnectionStatus status, ConnectionStatusChangeReason ch
...
+                            return RecommendedAction.OpenConnection;
+
+                        case ConnectionStatusChangeReason.BadCredential:
+                        case ConnectionStatusChangeReason.DeviceDisabled:
+                            return RecommendedAction.Quit;
+                    }
+                    break;
+            }
+
+            return RecommendedAction.OpenConnection;
```
abhipsaMisra, line 83: Q - other than disconnected-retry_expired, communication_error, what other status and reason combination gives us the Open recommended action?

iothub/device/src/ConnectionStatus.cs
```diff
@@ -4,32 +4,36 @@
...
+        /// <remarks>The client is attempting to recover the connection. Do NOT close or open the client instance when it is retrying.</remarks>
         DisconnectedRetrying,
 
         /// <summary>
-        /// The device connection was closed.
-        /// <para>If you want to perform more operations on the device client, you should <see cref="IotHubDeviceClient.Dispose()"/> and then re-initialize the client.</para>
+        /// The device connection was closed gracefully.
         /// </summary>
-        Disabled,
+        /// <remarks>
```
abhipsaMisra, line 34: Looks like we unintentionally lost the summary description

iothub/device/src/ConnectionStatus.cs
abhipsaMisra, line 37: Make sure you run these tests and make any changes as necessary: https://github.com/Azure/azure-iot-sdk-csharp/blob/main/e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs

iothub/device/src/RecommendedAction.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// This can occur when the client has been closed gracefully (by calling <see cref="IotHubDeviceClient.CloseAsync(CancellationToken)"/>
+        /// or the <see cref="IotHubModuleClient"/> equivalent).
+        /// Other terminal states include when the client has been disconnected due to non-retryable exceptions. The disconnection reasons include:
+        /// <list type="bullet">
+        /// <item><see cref="ConnectionStatusChangeReason.BadCredential"/></item>
+        /// <item><see cref="ConnectionStatusChangeReason.DeviceDisabled"/></item>
+        /// </list>
+        /// </remarks>
+        Quit,
+    }
```
abhipsaMisra, line 53: I like this set of recommended actions

iothub/device/src/ConnectionInfo.cs
brycewang-microsoft: these 2 should be all. `OpenConnection` is also the default recommended action when a client is first initialized though

abhipsaMisra: you could add a comment calling that out in this sample

iothub/device/samples/DeviceReconnectionSample/DeviceReconnectionSample.cs
timtay-microsoft: Indeed

timtay-microsoft (approved):

e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs
```diff
@@ -14,32 +14,32 @@ namespace Microsoft.Azure.Devices.E2ETests
...
-            await IotHubDeviceClient_Gives_ConnectionStatus_DeviceDisabled_Base(
+            await IotHubDeviceClient_Gives_ConnectionStatus_Disconnected_ChangeReason_DeviceDisabled_Base(
                 new IotHubClientAmqpSettings(IotHubClientTransportProtocol.WebSocket),
                 async (r, d) => await r.Devices.DeleteAsync(d).ConfigureAwait(false)).ConfigureAwait(false);
         }
 
         [LoggedTestMethod, Timeout(ConnectionStateChangeTestTimeoutMilliseconds)] // This test always takes more than 5 minutes for service to return. Needs investigation.
         [TestCategory("LongRunning")]
-        public async Task IotHubDeviceClient_DeviceDisabled_Gives_ConnectionStatus_DeviceDisabled_AmqpTcp()
+        public async Task IotHubDeviceClient_DeviceDisabled_Gives_ConnectionStatus_Disconnected_ChangeReason_DeviceDisabled_AmqpTcp()
```
abhipsaMisra, line 40: It looks like both deleting and disabling devices on IoT hub now gives the same response -> Disconnected with DeviceDisabled. This seems a bit off considering now we don't have any way to help the user understand if their device was indeed disabled or deleted.
I'm not sure what the previous behavior was, but I think this is something we can improve. What's the error code from service on both of these operations?

abhipsaMisra (approved):

e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs
brycewang-microsoft: I will look into this and try to bring improvements in another PR

iothub/device/samples/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -181,79 +181,47 @@ private async Task OnC2dMessageReceivedAsync(Message message, object context)
...
-            var status = connectionInfo.Status;
-            var reason = connectionInfo.ChangeReason;
-            Console.WriteLine($"Connection status changed: status={status}, reason={reason}");
+            ConnectionStatus status = connectionInfo.Status;
+            ConnectionStatusChangeReason reason = connectionInfo.ChangeReason;
+            Console.WriteLine($"Connection status changed: status={status}, reason={reason}, recommendation={connectionInfo.RecommendedAction}");
+
+            // In our case, we can operate with more than 1 shared access key and attempt to fall back to a secondary.
+            // We'll disregard the SDK's recommendation and attempt to connect with the second one.
+            if (status == ConnectionStatus.Disconnected
```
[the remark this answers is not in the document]
drwill-ms: Because this is an exception to using the recommended action.

Putting it inside recommended action means if the mapping was incorrect, the code would never run.

I prefer it up top for that reason.

PR #2605, opened 2022-08-18
IoHubConnectionStringBuilder refactoring - part 1

abhipsaMisra:


iothub/device/src/Authentication/AuthenticationMethodFactory.cs
```diff
@@ -53,63 +48,5 @@ internal static IAuthenticationMethod GetAuthenticationMethod(IotHubConnectionSt
...
         }
-
-        /// <summary>
-        /// Creates a <see cref="DeviceAuthenticationWithSharedAccessPolicyKey"/> instance based on the parameters.
-        /// </summary>
-        /// <param name="deviceId">Device Identifier.</param>
-        /// <param name="policyName">Name of the shared access policy to use.</param>
-        /// <param name="key">Key associated with the shared access policy.</param>
-        /// <returns>A new instance of the <see cref="DeviceAuthenticationWithSharedAccessPolicyKey"/> class.</returns>
-        public static IAuthenticationMethod CreateAuthenticationWithSharedAccessPolicyKey(string deviceId, string policyName, string key)
```
abhipsaMisra, line 64: All of the below factory methods were used

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -37,25 +33,6 @@ public sealed class IotHubConnectionStringBuilder
 
         private const string CommonX509CertPropertyName = "x509";
 
-        private const RegexOptions CommonRegexOptions = RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.CultureInvariant;
```
abhipsaMisra, line 40: I'm removing the regex checks in here, since these character restrictions could be updated by service at any time, and we would be required to keep them updated. Instead, we'll only check the following:
- device Id is set
- IoT hub hostname is set
- the shared access key supplied is a valid base64 string
- the shred access signature supplied follows the expected format

iothub/device/src/Authentication/Security/SharedAccessSignature.cs
```diff
@@ -2,152 +2,36 @@
...
-using System.Collections.Generic;
-using System.Globalization;
 using System.Net;
-using System.Security.Cryptography;
-using System.Text;
 
 namespace Microsoft.Azure.Devices.Client
 {
-    internal sealed class SharedAccessSignature : ISharedAccessSignatureCredential
+    internal sealed class SharedAccessSignature
```
abhipsaMisra, line 9: This class now represents only the parsed out shared access signature. The parsing logic has been moved to its own class. I'll check in a subsequent PR if this parsing is really required.

iothub/device/tests/Authentication/IotHubConnectionStringBuilderTests.cs
```diff
@@ -133,28 +133,6 @@ public void IotHubConnectionStringBuilder_ParamConnectionString_ParsesSharedAcce
             csBuilder.UsingX509Cert.Should().BeFalse("SharedAccessSignature and X509 are mutually exclusive");
         }
 
-        [TestMethod]
-        public void IotHubConnectionStringBuilder_OverrideAuthMethodToken()
```
abhipsaMisra, line 137: The IAuthenticationMethod property is now readonly and can only be set in the constructor. Hence, this test is no longer required.

Same for the test below.

drwill-ms (approved):

PR #2608, opened 2022-08-19
IotHubConnectionStringBuilder - part 2

abhipsaMisra:
This PR primarily removes x509 certificate information from the connection string flow.
Our device client constructors don't allow you to set both the connection string and client certificates.

### DIFF
--- a/iothub/device/src/Authentication/AuthenticationMethodFactory.cs
+++ b/iothub/device/src/Authentication/AuthenticationMethodFactory.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public sealed class AuthenticationMethodFactory
     {
-        internal static IAuthenticationMethod GetAuthenticationMethod(IotHubConnectionStringBuilder csBuilder)
+        internal static IAuthenticationMethod GetAuthenticationMethodFromConnectionString(IotHubConnectionStringBuilder csBuilder)
         {
             if (csBuilder.SharedAccessKeyName != null)
             {
@@ -40,10 +40,6 @@ namespace Microsoft.Azure.Devices.Client
                         csBuilder.DeviceId,
                         csBuilder.ModuleId,
                         csBuilder.SharedAccessSignature);
-            }
-            else if (csBuilder.UsingX509Cert)
-            {
-                return new DeviceAuthenticationWithX509Certificate(csBuilder.DeviceId, csBuilder.Certificate);
             }
 
             throw new InvalidOperationException($"Unsupported authentication method in '{csBuilder}'.");
--- a/iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
+++ b/iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
@@ -191,9 +191,12 @@ namespace Microsoft.Azure.Devices.Client
             }
 
             // Either shared access key, shared access signature or X.509 certificate is required for authenticating the client with IoT hub.
+            // These values should be populated in the constructor. The only exception to this scenario is when the authentication method is
+            // AuthenticationWithTokenRefresh, in which case the shared access signature is initially null and is generated on demand during client authentication.
             if (Certificate == null
                 && SharedAccessKey.IsNullOrWhiteSpace()
-                    && SharedAccessSignature.IsNullOrWhiteSpace())
+                && SharedAccessSignature.IsNullOrWhiteSpace()
+                && AuthenticationMethod is not AuthenticationWithTokenRefresh)
             {
                 throw new ArgumentException(
                         "Should specify either SharedAccessKey, SharedAccessSignature or X.509 certificate for authenticating the client with IoT hub.");
--- a/iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
+++ b/iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
@@ -24,15 +24,6 @@ namespace Microsoft.Azure.Devices.Client
         private const string SharedAccessSignaturePropertyName = "SharedAccessSignature";
         private const string GatewayHostNamePropertyName = "GatewayHostName";
 
-        // For some reason, the .NET SDK originally expected "X509Cert=true" in a connection string to inform the SDK that it would not
-        // include a shared access key, and to not error when parsing a connection string. However, the other SDK languages all followed
-        // the same key/value pair of "x509=true".
-        // So now we're adding support for it to work either way, so we stay compliant with the past but can improve documentation so all
-        // SDK languages refer to the same key/value pair naming.
-        private const string X509CertPropertyName = "X509Cert";
-
-        private const string CommonX509CertPropertyName = "x509";
-
         /// <summary>
         /// Creates an instnace of this class based on an authentication method and the hostname of the IoT hub.
         /// </summary>
@@ -65,7 +56,7 @@ namespace Microsoft.Azure.Devices.Client
 
             // We'll parse the connection string and use that to build an auth method
             ExtractPropertiesFromConnectionString(iotHubConnectionString);
-            AuthenticationMethod = AuthenticationMethodFactory.GetAuthenticationMethod(this);
+            AuthenticationMethod = AuthenticationMethodFactory.GetAuthenticationMethodFromConnectionString(this);
 
             Validate();
         }
@@ -73,32 +64,32 @@ namespace Microsoft.Azure.Devices.Client
         /// <summary>
         /// Gets or sets the value of the fully-qualified DNS hostname of the IoT hub service.
         /// </summary>
-        public string HostName { get; internal set; }
+        public string HostName { get; private set; }
 
         /// <summary>
         /// Gets the optional name of the gateway to connect to
         /// </summary>
-        public string GatewayHostName { get; internal set; }
+        public string GatewayHostName { get; private set; }
 
         /// <summary>
         /// Gets the device identifier of the device connecting to the service.
         /// </summary>
-        public string DeviceId { get; internal set; }
+        public string DeviceId { get; set; }
 
         /// <summary>
         /// Gets the module identifier of the module connecting to the service.
         /// </summary>
-        public string ModuleId { get; internal set; }
+        public string ModuleId { get; set; }
 
         /// <summary>
         /// Gets the shared access key name used to connect the device to the IoT hub service.
         /// </summary>
-        public string SharedAccessKeyName { get; internal set; }
+        public string SharedAccessKeyName { get; set; }
 
         /// <summary>
         /// Gets the shared access key used to connect to the IoT hub service.
         /// </summary>
-        public string SharedAccessKey { get; internal set; }
+        public string SharedAccessKey { get; set; }
 
         /// <summary>
         /// Gets the shared access signature used to connect to the IoT hub service.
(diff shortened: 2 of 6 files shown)


iothub/device/src/Authentication/AuthenticationMethodFactory.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public sealed class AuthenticationMethodFactory
     {
-        internal static IAuthenticationMethod GetAuthenticationMethod(IotHubConnectionStringBuilder csBuilder)
```
abhipsaMisra, line 13: This method gets called only from the connection string flow, so renaming it to make it more specific

iothub/device/src/Authentication/AuthenticationMethodFactory.cs
```diff
@@ -41,10 +41,6 @@ internal static IAuthenticationMethod GetAuthenticationMethod(IotHubConnectionSt
                         csBuilder.ModuleId,
                         csBuilder.SharedAccessSignature);
             }
-            else if (csBuilder.UsingX509Cert)
```
abhipsaMisra, line 44: connection string flow isn't expected to have certificate information

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -24,15 +24,6 @@ public sealed class IotHubConnectionStringBuilder
...
         private const string GatewayHostNamePropertyName = "GatewayHostName";
 
-        // For some reason, the .NET SDK originally expected "X509Cert=true" in a connection string to inform the SDK that it would not
-        // include a shared access key, and to not error when parsing a connection string. However, the other SDK languages all followed
-        // the same key/value pair of "x509=true".
-        // So now we're adding support for it to work either way, so we stay compliant with the past but can improve documentation so all
-        // SDK languages refer to the same key/value pair naming.
-        private const string X509CertPropertyName = "X509Cert";
-
-        private const string CommonX509CertPropertyName = "x509";
```
abhipsaMisra, line 34: The Azure portal no longer provides connection strings for X509 authenticated clients. Removing this support.

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -65,40 +56,40 @@ public IotHubConnectionStringBuilder(string iotHubConnectionString)
...
         /// <summary>
         /// Gets or sets the value of the fully-qualified DNS hostname of the IoT hub service.
         /// </summary>
-        public string HostName { get; internal set; }
+        public string HostName { get; private set; }
 
         /// <summary>
         /// Gets the optional name of the gateway to connect to
         /// </summary>
-        public string GatewayHostName { get; internal set; }
```
abhipsaMisra, line 81: These properties are expected to be set only through the constructor

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -65,40 +56,40 @@ public IotHubConnectionStringBuilder(string iotHubConnectionString)
...
+        public string GatewayHostName { get; private set; }
 
         /// <summary>
         /// Gets the device identifier of the device connecting to the service.
         /// </summary>
-        public string DeviceId { get; internal set; }
```
abhipsaMisra, line 86: This and below public properties - these can be set through a custom IAuthenticationMethod implementation through the `Populate` method. So, making them public.

The constructor still calls validate, so the properties would get validated there.

```csharp
AuthenticationMethod.Populate(this);
Validate();
```

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -109,28 +100,22 @@ public IotHubConnectionStringBuilder(string iotHubConnectionString)
         /// SAS token, when that token expires, the client must be disposed, and if desired, recreated
         /// with a newly derived SAS token.
         /// </remarks>
-        public string SharedAccessSignature { get; internal set; }
+        public string SharedAccessSignature { get; set; }
 
         /// <summary>
-        /// Gets or sets the authentication method to be used with the IoT hub service.
-        /// </summary>
-        public IAuthenticationMethod AuthenticationMethod { get; }
```
abhipsaMisra, line 117: This isn't required to be exposed to the user. Making this internal

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -169,14 +153,12 @@ private void ExtractPropertiesFromConnectionString(string iotHubConnectionString
             IDictionary<string, string> map = iotHubConnectionString.ToDictionary(ValuePairDelimiter, ValuePairSeparator);
 
             HostName = GetConnectionStringValue(map, HostNamePropertyName);
+            GatewayHostName = GetConnectionStringOptionalValue(map, GatewayHostNamePropertyName);
             DeviceId = GetConnectionStringOptionalValue(map, DeviceIdPropertyName);
             ModuleId = GetConnectionStringOptionalValue(map, ModuleIdPropertyName);
             SharedAccessKeyName = GetConnectionStringOptionalValue(map, SharedAccessKeyNamePropertyName);
             SharedAccessKey = GetConnectionStringOptionalValue(map, SharedAccessKeyPropertyName);
             SharedAccessSignature = GetConnectionStringOptionalValue(map, SharedAccessSignaturePropertyName);
-            UsingX509Cert = GetConnectionStringOptionalValueOrDefault<bool>(map, X509CertPropertyName, ParseX509, true)
```
abhipsaMisra, line 177: removing x509 information from connection string flow

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -201,28 +183,29 @@ internal void Validate()
...
-            if (!string.IsNullOrWhiteSpace(SharedAccessSignature))
+            if (!SharedAccessSignature.IsNullOrWhiteSpace())
             {
                 // Parse the supplied shared access signature string
                 // and throw exception if the string is not in the expected format.
                 _ = SharedAccessSignatureParser.Parse(SharedAccessSignature);
             }
 
-            if (!(SharedAccessKey.IsNullOrWhiteSpace() ^ SharedAccessSignature.IsNullOrWhiteSpace()))
+            // Either shared access key, shared access signature or X.509 certificate is required for authenticating the client with IoT hub.
```
abhipsaMisra, line 193: Added some more comments describing the logic

iothub/device/src/ClientFactory.cs
```diff
@@ -27,11 +27,6 @@ internal static class ClientFactory
             }
 
             var csBuilder = new IotHubConnectionStringBuilder(connectionString);
-            if (csBuilder.UsingX509Cert)
```
abhipsaMisra, line 30: connection string flow doesn't allow x509 information to be set

iothub/device/tests/Authentication/IotHubConnectionStringBuilderTests.cs
```diff
@@ -130,53 +130,7 @@ public void IotHubConnectionStringBuilder_ParamConnectionString_ParsesSharedAcce
...
 
             csBuilder.SharedAccessKey.Should().BeNull("SharedAccessSignature and SharedAccessKey are mutually exclusive");
-            csBuilder.UsingX509Cert.Should().BeFalse("SharedAccessSignature and X509 are mutually exclusive");
-        }
-
-        [TestMethod]
-        [DataRow("true")]
-        [DataRow("True")]
-        [DataRow("TRUE")]
-        public void IotHubConnectionStringBuilder_ParamConnectionString_ParsesX509BoolCaseInsensitive(string value)
```
abhipsaMisra, line 140: This is no longer being supported

iothub/device/tests/Authentication/DeviceClientConnectionStringExceptionTests.cs
```diff
@@ -120,22 +120,6 @@ public void DeviceClientConnectionStringEmptyConnectionStringExceptionTest()
             var deviceClient = IotHubDeviceClient.CreateFromConnectionString("");
         }
 
-        [TestMethod]
-        [ExpectedException(typeof(ArgumentException))]
-        public void DeviceClientConnectionStringSASKeyX509CertExceptionTest()
```
abhipsaMisra, line 125: We only parse out the expected contents of a connection string. Any superfluous data in the connection string will be ignored

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
drwill-ms: Azure IoT explorer does

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
drwill-ms: readonly then?

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
drwill-ms: Why not internal?

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -109,28 +100,22 @@ public IotHubConnectionStringBuilder(string iotHubConnectionString)
...
-        /// Gets or sets the authentication method to be used with the IoT hub service.
-        /// </summary>
-        public IAuthenticationMethod AuthenticationMethod { get; }
-
-        /// <summary>
-        /// Indicates if the connection string indicates if an x509 certificate is specified for authentication.
+        /// The client X509 certificates used for authenticating with IoT hub.
         /// </summary>
-        public bool UsingX509Cert { get; internal set; }
+        public X509Certificate2 Certificate { get; set; }
```
drwill-ms, line 108: Why does a "connection string builder" need cert info?

abhipsaMisra: This class holds the authentication information for any client to connect to IoT hub. It had been poorly named in the past. The next PR addresses the renaming: https://github.com/Azure/azure-iot-sdk-csharp/pull/2609

drwill-ms (approved):

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
abhipsaMisra: I know. I thought about that but since Azure portal doesn't, I took that as the best recommended official approach (since that is where everyone is going to start). Do you think this flow is worth preserving then? Or maybe, should we sync up with the rest of the SDKs and see how we collectively feel about it?

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
abhipsaMisra: The constructor itself calls a static method to do most of the settings. So, I made the setter private.

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
abhipsaMisra: The Populate method class is on the public IAUthenticationMethod interface that anyone can implement.

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
abhipsaMisra: Our constructors support only 2 flows today - either pass in a connection string, or pass in a IAUthenticationMethod that has either sas key/sas token information or the certificate information. There isn't a way for someone to supply both the certificates and a connection string. So, we'd have to go out of our way and modify our client initializers to support "x509=true" in connection string.

drwill-ms: The only reason for preserving it is for validating a connection string and throwing when it is invalid. If it doesn't have a SAK, it is invalid, unless it says, "oh but i'm using x509, trust me".

But we're not doing that, so I don't mind losing it.

I'm just saying it isn't gone from everywhere.

iothub/device/src/Authentication/IotHubConnectionStringBuilder.cs
drwill-ms: Ah

PR #2613, opened 2022-08-22
Disable test for net6.0 for gwv2 for now

drwill-ms:
Fails during TLS handshake but only for MQTT TCP to gateway v2. As the test is for a disabled/deleted device, this scenario seems to be a reasonable regression, that hopefully will be different with MQTTNet.

### DIFF
--- a/e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
+++ b/e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
@@ -43,17 +43,18 @@ namespace Microsoft.Azure.Devices.E2ETests
             await X509InvalidDeviceIdOpenAsyncTest(DeviceTransportType.Amqp_WebSocket_Only).ConfigureAwait(false);
         }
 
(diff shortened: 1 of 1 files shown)


e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -43,17 +43,18 @@ public async Task X509_InvalidDeviceId_Throw_UnauthorizedException_AmqpWs()
             await X509InvalidDeviceIdOpenAsyncTest(DeviceTransportType.Amqp_WebSocket_Only).ConfigureAwait(false);
         }
 
+// TODO: there is a problem with DotNetty on net6.0 with gatewayv2. Wait for transition to MqttNet to re-enable and try this test out.
+#if !NET6_0_OR_GREATER
         [LoggedTestMethod]
-        [Timeout(LongRunningTestTimeoutMilliseconds)]
-        [TestCategory("LongRunning")]
```
drwill-ms, line 48: These tests all run within seconds.

timtay-microsoft (approved):

andyk-ms (approved):

PR #2619, opened 2022-08-23
Remove ServicePointHelpers.cs

timtay-microsoft:
Move the relevant http client settings to HttpClientFactory.cs


patilsnr (approved):

iothub/service/src/Http/HttpClientFactory.cs
```diff
@@ -20,6 +20,18 @@ internal class HttpClientFactory
...
+        // If this value is set to any negative value, the connection lease will be infinite. If this value is set to 0, then the TCP connection will close after
+        // each HTTP request and a new TCP connection will be opened upon the next request.
+        //
+        // By closing cached TCP connections and opening a new one upon the next request, the underlying HTTP client has a chance to do a DNS lookup
+        // to validate that it will send the requests to the correct IP address. While it is atypical for a given IoT hub to change its IP address, it does
+        // happen when a given IoT hub fails over into a different region.
+        //
+        // This default value is consistent with the default value used in Azure.Core
+        // https://github.com/Azure/azure-sdk-for-net/blob/7e3cf643977591e9041f4c628fd4d28237398e0b/sdk/core/Azure.Core/src/Pipeline/ServicePointHelpers.cs#L29
+        private const int DefaultConnectionLeaseTimeoutMinutes = 5;
```
drwill-ms, line 33: Why not make the property a TimeSpan? That way the unit doesn't matter and it makes more sense to readers.

PR #2623, opened 2022-08-24
Some job query client test cleanup

drwill-ms:


e2e/test/iothub/service/QueryClientE2ETests.cs
```diff
@@ -132,23 +136,36 @@ public async Task JobQuery_QueryWorks()
             queriedJob.JobType.Should().NotBeNull();
             queriedJob.CreatedTimeUtc.Should().NotBeNull();
             queriedJob.Status.Should().NotBeNull();
+            if (queriedJob.Status != JobStatus.Queued
+                && queriedJob.Status != JobStatus.Scheduled
+                && queriedJob.Status != JobStatus.Cancelled
+                && queriedJob.Status != JobStatus.Unknown)
+            {
+                queriedJob.StartTimeUtc.Should().NotBeNull();
```
drwill-ms, line 144: Optionally check these values when they should be there, so we know we are deserializing them fine.

e2e/test/iothub/service/QueryClientE2ETests.cs
```diff
@@ -184,13 +202,13 @@ private async Task WaitForDevicesToBeQueryable(QueryClient queryClient, string q
             QueryResponse<Twin> queryResponse = await queryClient.CreateAsync<Twin>(query);
             while (queryResponse.CurrentPage.Count() < expectedCount)
             {
-                Thread.Sleep(100);
+                await Task.Delay(100).ConfigureAwait(false);
```
drwill-ms, line 205: Should use the async sleep

e2e/test/iothub/service/QueryClientE2ETests.cs
```diff
@@ -18,6 +18,7 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
     [TestClass]
     [TestCategory("E2E")]
     [TestCategory("IoTHub")]
+    [DoNotParallelize] // creating jobs limits to running one at a time anyway, so avoid throttling conflicts
```
timtay-microsoft, line 21: These tests may be run on different .NET frameworks simultaneously, so this doesn't protect us completely from throttling

drwill-ms: Not completely, but we can at least reduce it.

drwill-ms: If we make other changes here, we could probably remove this constraint.

Maybe class initialize queries to see if at least one job is there, and if not, creates and immediately cancels one, for example.

andyk-ms: This set of tests are for querying devices, the problem likely from other test that is creating device or removing devices that is running in parallel.

There is a idPrefix string that may add uniqueness to the runs. What about we add the host-info (like hostname, or agentId etc) to the prefix so parallel run on different agent have its own scope of devices prefix?

drwill-ms: > This set of tests are for querying devices, the problem likely from other test that is creating devices or removing devices in parallel.

Tests shouldn't be removing devices with Ids they didn't create, so I don't anticipate that being a problem.

andyk-ms (approved):

timtay-microsoft (approved):

PR #2624, opened 2022-08-24
Code cleanup after review with Tim/Abhipsa

drwill-ms:


iothub/service/src/ApiResources.resx
```diff
@@ -120,9 +120,6 @@
   <data name="ConnectionStringIsNotWellFormed" xml:space="preserve">
     <value>The connection string is not well formed.</value>
   </data>
-  <data name="HostNameIsNull" xml:space="preserve">
```
drwill-ms, line 123: Removed one case where we spot checked was not used. We should go through this file and remove all the others that aren't referenced.

iothub/service/src/Jobs/DeviceJobParameters.cs
```diff
@@ -27,23 +25,12 @@ public DeviceJobParameters(JobType jobType, string deviceId)
         public DeviceJobParameters(JobType jobType, IEnumerable<string> deviceIds)
             : base(jobType)
         {
-            IList<string> deviceList = deviceIds?.ToListSlim();
```
drwill-ms, line 30: We can let the service validate the input, and no longer need to convert IEnumerable to IList.

iothub/service/src/Utilities/ExceptionExtensions.cs
```diff
@@ -1,61 +0,0 @@
...
-    public static class ExceptionExtensions
-    {
-        private const string ExceptionIdentifierName = "ExceptionId";
-
-        /// <summary>
-        /// Indicates whether the exception is considered fatal.
-        /// </summary>
-        /// <param name="exception">The exception to test.</param>
-        /// <returns>True if the exception is considered fatal; otherwise, false.</returns>
-        internal static bool IsFatal(this Exception exception)
```
drwill-ms, line 25: They can just as easily make this call directly.

iothub/service/src/Utilities/ExceptionExtensions.cs
```diff
@@ -1,61 +0,0 @@
...
-        {
-            return Fx.IsFatal(exception);
-        }
-
-        /// <summary>
-        /// Stringify the exception, containing all relevant fields.
-        /// </summary>
-        /// <param name="exception">The exception to stringify.</param>
-        /// <returns>The stringified exception.</returns>
-        internal static string ToStringSlim(this Exception exception)
```
drwill-ms, line 35: Moved to the one place it is used.

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -20,18 +20,18 @@ public class IotHubServiceClientOptions
         /// </summary>
         public IotHubServiceClientOptions()
         {
-            Proxy = DefaultWebProxySettings.Instance;
```
drwill-ms, line 23: Let default be null.

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -20,18 +20,18 @@ public class IotHubServiceClientOptions
...
-            UseWebSocketOnly = false;
```
drwill-ms, line 24: false is the default; no reason to initialize it.

timtay-microsoft (approved):

abhipsaMisra (approved):

iothub/service/src/Configurations/ConfigurationsClient.cs
```diff
@@ -66,7 +66,7 @@ public virtual async Task<Configuration> CreateAsync(Configuration configuration
 
             try
             {
-                Argument.RequireNotNull(configuration, nameof(configuration));
+                Argument.AssertNotNull(configuration, nameof(configuration));
```
drwill-ms, line 69: I think we'll find it confusing when the Arguments classes per NuGet package don't work the same. I've modeled the names after what is in Track2, again for consistency.

iothub/service/tests/ClientWebSocketTransportTests.cs
```diff
@@ -29,17 +29,17 @@ public class ClientWebSocketTransportTests
         private static HttpListener s_listener;
         private static byte[] s_byteArray = new byte[10] { 0x5, 0x6, 0x7, 0x8, 0xA, 0xB, 0xC, 0xD, 0xE, 0xF };
 
-        [AssemblyInitialize()]
```
drwill-ms, line 32: Why would this be required for all unit tests in the assembly?

Also, use the async version.

iothub/device/src/Authentication/HsmAuthentication/ModuleAuthenticationWithHsm.cs
```diff
@@ -39,7 +39,7 @@ protected override async Task<string> SafeCreateNewTokenAsync(string iotHub, Tim
             string audience = SasTokenBuilder.BuildAudience(iotHub, DeviceId, ModuleId);
             string expiresOn = SasTokenBuilder.BuildExpiresOn(startTime, suggestedTimeToLive);
             string data = string.Join("\n", new string[] { audience, expiresOn });
-            string signature = await _signatureProvider.SignAsync(ModuleId, _generationId, data);
+            string signature = await _signatureProvider.SignAsync(ModuleId, _generationId, data).ConfigureAwait(false);
```
drwill-ms, line 42: Broke this in a previous PR.

iothub/device/src/Transport/Mqtt/ClientWebSocketChannel.cs
```diff
@@ -98,7 +98,6 @@ public ClientWebSocketChannel Option<T>(ChannelOption<T> option, T value)
         public void Dispose()
         {
             _webSocket?.Dispose();
-            _webSocket = null;
```
drwill-ms, line 101: Found an NRE from a test I un-ignored.

iothub/device/tests/Transport/Mqtt/ClientWebSocketChannelTests.cs
```diff
@@ -14,46 +14,43 @@ namespace Microsoft.Azure.Devices.Client.Test.Mqtt
...
         {
-            listener = new HttpListener();
-            listener.Prefixes.Add("http://+:" + Port + WebSocketConstants.UriSuffix + "/");
-            listener.Start();
+            s_listener = new HttpListener();
+            s_listener.Prefixes.Add("http://+:" + Port + WebSocketConstants.UriSuffix + "/");
+            s_listener.Start();
 
-            RunWebSocketServer().ContinueWith(t => t, TaskContinuationOptions.OnlyOnFaulted);
+            _ = RunWebSocketServer();
```
drwill-ms, line 46: We don't await this because it runs for the duration of the test, and is signalled to quit in ClassCleanup.

abhipsaMisra (approved):

PR #2625, opened 2022-08-24
Update the internal authentication flow of the client

abhipsaMisra:
We had a lot of "configuration" type classes that held client information (credentials, transport settings, options etc), and they were all inter-related in a messy way.
This PR refactors it so that we have clear separation between the concepts.

We'll pass in the pipeline context to our transport layer.
The pipeline context has the following properties:
- connection credentials (contains the required params to authenticate a client - sas tokens/ certificate information).
- product information
- model Id
- transport settings (the transport settings still holds certificate information, but I'll refactor that out in a separate PR)
- all the callbacks that a user sets on the client

This PR also refactors out the parsing logic for a device connection string and shared access signature tokens.


e2e/test/helpers/IotHubConnectionCredentialsExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text;
+using Microsoft.Azure.Devices.Client;
+
+namespace Microsoft.Azure.Devices.E2ETests.Helpers
+{
+    internal static class IotHubConnectionCredentialsExtensions
+    {
+        public static string GetIotHubConnectionString(this IotHubConnectionCredentials iotHubConnectionCredentials)
```
abhipsaMisra, line 11: This logic is used by a single edge related test that tests edge capabilities. Moved this logic to a test helper instead of keeping it in our sourcecode.

e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
```diff
@@ -115,14 +115,11 @@ public async Task IotHubDeviceClient_CreateFromConnectionString_TokenIsRefreshed
             using var deviceDisconnected = new SemaphoreSlim(0);
 
             TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, DevicePrefix).ConfigureAwait(false);
+            var auth = new DeviceAuthenticationWithConnectionString(testDevice.ConnectionString, sasTokenTimeToLive, sasTokenRenewalBuffer);
```
abhipsaMisra, line 118: We now have a specific IAuthenticationMethod that accepts connection string and the sas ttl settings. The regular CreateFromConnectionString no longer accepts configurable sas ttl.

iothub/device/src/AdditionalClientInformation.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    // Holder for additional client properties such as product information, PnP model Id.
+    // This class does not hold any client credential information.
+    internal class AdditionalClientInformation
```
abhipsaMisra, line 12: This is a holder for additional client information that is passed on to the transport layer. Previously we were passing in the complete client options but this now exposes only the specific required information.

iothub/device/src/Authentication/IConnectionCredentials.cs
```diff
@@ -0,0 +1,94 @@
...
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Azure.Devices.Client.Transport;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Interface for client connection credentials.
+    /// This has been included for our own unit testing.
```
abhipsaMisra, line 12: Purpose of this class is primarily for unit testing

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -2,30 +2,20 @@
...
 
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// Holder for client credentials that will be used for authenticating the client with IoT hub service.
     /// </summary>
-    public sealed class IotHubConnectionCredentials
+    public sealed class IotHubConnectionCredentials : IConnectionCredentials
     {
-        private const char ValuePairDelimiter = ';';
```
abhipsaMisra, line 17: Moved the parsing logic to a dedicated parser class.

iothub/device/src/ClientFactory.cs
```diff
@@ -105,23 +105,12 @@ internal static InternalClient Create(string hostName, IAuthenticationMethod aut
             Argument.AssertNotNull(iotHubConnectionCredentials, nameof(iotHubConnectionCredentials));
             Argument.AssertNotNull(options, nameof(options));
 
-            // Clients that derive their authentication method from AuthenticationWithTokenRefresh will need to specify
```
abhipsaMisra, line 108: This flow is no longer supported

iothub/device/src/Edge/EdgeModuleClientFactory.cs
```diff
@@ -92,13 +92,14 @@ private async Task<IotHubModuleClient> CreateInternalClientFromEnvironmentAsync(
...
                 ISignatureProvider signatureProvider = new HttpHsmSignatureProvider(edgedUri, DefaultApiVersion);
 
-                TimeSpan sasTokenTimeToLive = _options.SasTokenTimeToLive;
-                int sasTokenRenewalBuffer = _options.SasTokenRenewalBuffer;
-
-#pragma warning disable CA2000 // Dispose objects before losing scope - IDisposable ModuleAuthenticationWithHsm is disposed when the client is disposed.
-                // Since the sdk creates the instance of disposable ModuleAuthenticationWithHsm, the sdk needs to dispose it once the client is disposed.
-                var authMethod = new ModuleAuthenticationWithHsm(signatureProvider, deviceId, moduleId, generationId, sasTokenTimeToLive, sasTokenRenewalBuffer);
-#pragma warning restore CA2000 // Dispose objects before losing scope - IDisposable ModuleAuthenticationWithHsm is disposed when the client is disposed.
+                // TODO: environment variables need to be added to accept SasTokenTimeToLive and SasTokenRenewalBuffer.
```
abhipsaMisra, line 95: This is an open #TODO item

iothub/device/src/IotHubClientOptions.cs
```diff
@@ -67,30 +67,6 @@ public IotHubClientOptions(IotHubClientTransportSettings transportSettings)
...
-        /// The suggested time to live value for tokens generated for SAS authenticated clients.
-        /// The <see cref="TimeSpan"/> provided should be a positive value, signifying that it is not possible to generate tokens that have already expired.
-        /// If unset the generated SAS tokens will be valid for 1 hour.
-        /// </summary>
-        /// <remarks>
-        /// This is used only for SAS token authenticated clients through either the
-        /// <see cref="IotHubDeviceClient.CreateFromConnectionString(string, IotHubClientOptions)"/> flow, the <see cref="IotHubModuleClient.CreateFromConnectionString(string, IotHubClientOptions)"/> flow
-        /// or the <see cref="IotHubModuleClient.CreateFromEnvironmentAsync(IotHubClientOptions)"/> flow.
-        /// </remarks>
-        public TimeSpan SasTokenTimeToLive { get; set; }
```
abhipsaMisra, line 80: The connection string flow has an associated IAuthenticationMethod that can be used for this feature.

iothub/device/src/Pipeline/TransportHandlerFactory.cs
```diff
@@ -12,43 +11,31 @@ internal class TransportHandlerFactory : ITransportHandlerFactory
...
-            Func<Message, Task> onDeviceMessageReceivedCallback = context.DeviceEventCallback;
+            IotHubClientTransportSettings transportSettings = context.IotHubClientTransportSettings;
 
-            if (clientConfiguration.ClientOptions.TransportSettings is IotHubClientAmqpSettings iotHubClientAmqpSettings)
+            if (transportSettings is IotHubClientAmqpSettings iotHubClientAmqpSettings)
             {
                 return new AmqpTransportHandler(
                     context,
-                    iotHubClientAmqpSettings,
-                    onMethodCallback,
```
abhipsaMisra, line 26: These callbacks are a part of the pipeline context which is passed into the transport layer. It is not required for us to extract these here and pass them in again.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -17,22 +17,22 @@ internal class AmqpAuthenticationRefresher : IAmqpAuthenticationRefresher
         private static readonly string[] s_accessRightsStringArray = AccessRightsHelper.AccessRightsToStringArray(AccessRights.DeviceConnect);
         private readonly Uri _amqpEndpoint;
         private readonly AmqpIotCbsLink _amqpIotCbsLink;
-        private readonly IClientConfiguration _clientConfiguration;
```
abhipsaMisra, line 20: The amqp changes rea only to refactor the layer to use IotHubConnectionCredentials instead of ClientConfiguration.

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -127,10 +128,10 @@ public void IotHubDeviceClient_ParamsHostNameAuthMethodTransportType_Works()
             var transportSettings = new IotHubClientAmqpSettings(IotHubClientTransportProtocol.WebSocket);
             var options = new IotHubClientOptions(transportSettings);
 
-            // TODO (abmisr): IotHubClientOptions shouldn't be required in both
```
abhipsaMisra, line 130: This TODO has now been addressed

iothub/device/src/Authentication/AuthenticationModel.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// For example, myHub.azure-devices.net/devices/device1.
+        /// </summary>
+        SasIndividual,
+
+        /// <summary>
+        /// This is the authentication model where the SAS tokens generated for a client are scoped to IoT hub level; for example, myHub.azure-devices.net
+        /// This is generally not as secure as X.509 certificates or individually authenticated client authentication, as it opens up 
+        /// the IoT hub service instance to vulnerabilities in case the SAS tokens are compromised.
+        /// </summary>
+        SasGrouped,
```
timtay-microsoft, line 27: Have we considered removing this support for v2?

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -112,61 +117,188 @@ public IotHubConnectionCredentials(string iotHubConnectionString)
...
-            // Host name
-            if (HostName.IsNullOrWhiteSpace())
+            // IoT Hub host name
+            if (IotHubHostName.IsNullOrWhiteSpace())
+            {
+                throw new FormatException("IoT hub service hostname to connect to must be specified.");
+            }
+
+            // Gateway host name
+            if (GatewayHostName.IsNullOrWhiteSpace())
```
timtay-microsoft, line 299: Why is this now mandatory? Users don't need a gateway to connect if they want to connect to IoT hub directly

iothub/device/src/Authentication/AuthenticationModel.cs
abhipsaMisra: David and I spoke about this briefly this morning, and although we both agree that this is not a secure mechanism for authenticating clients, we're wary of removing the support outright since people might have taken a dependency on this feature with implementing their custom gateway solutions with multiplexing. 
That said, the plan is to dissuade people from adopting this flow, and instead showing them a flow using AAD with IoT hub service to create multiple client identities and connect as them (as a part of a level 400 sample).

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
abhipsaMisra:
```csharp
        /// <summary>
        /// The host service that this client connects to.
        /// This can either be the IoT hub name or a gateway service name.
        /// </summary>
        string GatewayHostName { get; }
`

File: iothub/device/src/Authentication/IotHubConnectionStringParser.cs [hunk 14]
```diff
@@ -0,0 +1,52 @@
...
+            {
+                throw new ArgumentException($"The connection string is missing the property: {propertyName}.");
+            }
+
+            return value;
+        }
+
+        private static string GetConnectionStringOptionalValue(IDictionary<string, string> map, string propertyName)
+        {
+            map.TryGetValue(propertyName, out string value);
```
timtay-microsoft, line 48: If we are ignoring the returned boolean here, can't we just make this

```csharp
return map[propertyName];
```

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
abhipsaMisra:
```csharp
            IsUsingGateway = !iotHubConnectionString.GatewayHostName.IsNullOrWhiteSpace();
            GatewayHostName = IsUsingGateway
                ? iotHubConnectionString.GatewayHostName
                : IotHubHostName;
```

iothub/device/src/Authentication/IotHubConnectionStringParser.cs
timtay-microsoft: And if we do that, we won't need a helper function to do one line of code

iothub/device/src/Edge/EdgeModuleClientFactory.cs
timtay-microsoft: Is there any reason why you want to take them as environment variables rather than in ClientOptions? Are these env vars set up for the user in Edge scenarios?

iothub/device/src/Authentication/IotHubConnectionStringParser.cs
abhipsaMisra: using an indexer still leaves us open to KeyNotFoundException. This will just return "null" instead.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -38,22 +38,26 @@ static AmqpTransportHandler()
 
         internal AmqpTransportHandler(
             PipelineContext context,
-            IotHubClientAmqpSettings transportSettings,
-            Func<MethodRequestInternal, Task> onMethodCallback = null,
-            Action<TwinCollection> onDesiredStatePatchReceivedCallback = null,
-            Func<string, Message, Task> onModuleMessageReceivedCallback = null,
-            Func<Message, Task> onDeviceMessageReceivedCallback = null)
+            IotHubClientAmqpSettings transportSettings)
```
timtay-microsoft, line 41: I love how much simpler this parameter list is now

timtay-microsoft (approved):

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
timtay-microsoft: This seems a bit confusing to me. If the user didn't pass in a gatewayhostname in their connection string, why would the gatewayhostname be anything but null in our SDK?

abhipsaMisra: That's fair. I think the name "gateway" is what's misleading there. We essentially need 3 fields - IotHubHostName, GatewayHostName and a "connecting-service-endpoint" that holds the value for this expression:
```c#
connecting-service-endpoint = iotHubConnectionString.GatewayHostName.IsNullOrWhiteSpace()
                ? IotHubHostName
                : iotHubConnectionString.GatewayHostName;
```

abhipsaMisra: Any suggestions for "connecting-service-endpoint"?

timtay-microsoft: "connecting-service-endpoint" as in the server that this client _directly_ connects to? "hostname" seems like a good enough name to me

iothub/device/src/Authentication/DeviceAuthenticationWithConnectionString.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// The default behavior is that the token will be renewed when it has 15% or less of its lifespan left.
+        ///</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown if <paramref name="suggestedTimeToLive"/> is a negative timespan, or if
+        /// <paramref name="timeBufferPercentage"/> is outside the range 0-100.</exception>
+        public DeviceAuthenticationWithConnectionString(
+            string connectionString,
+            TimeSpan suggestedTimeToLive = default,
+            int timeBufferPercentage = default)
+        {
+            if (suggestedTimeToLive.Ticks < 0)
```
drwill-ms, line 40: Why not `< TimeSpan.Zero`?

iothub/device/src/Authentication/DeviceAuthenticationWithRegistrySymmetricKey.cs
```diff
@@ -12,18 +12,56 @@ namespace Microsoft.Azure.Devices.Client
...
+        private readonly int _timeBufferPercentage;
+
         private string _deviceId;
         private byte[] _key;
 
         /// <summary>
         /// Initializes a new instance of the class.
         /// </summary>
-        /// <param name="deviceId">Device identifier.</param>
+        /// <param name="deviceId">Device Identifier.</param>
```
drwill-ms, line 27: Why capitalize identifier? Is it a property noun?

iothub/device/src/Authentication/IotHubConnectionString.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text;
+using Microsoft.Azure.Devices.Client.Extensions;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    internal class IotHubConnectionString
```
drwill-ms, line 9: Meaningful class comments, please.

iothub/device/src/Authentication/IotHubConnectionStringParser.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.Devices.Client.Extensions;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    internal sealed class IotHubConnectionStringParser
```
drwill-ms, line 10: Did you add unit tests?

iothub/device/src/Authentication/ModuleAuthenticationWithSakRefresh.cs
```diff
@@ -10,33 +10,36 @@ namespace Microsoft.Azure.Devices.Client
...
+            string sharedAccessKey,
+            string sharedAccessKeyName = default,
             TimeSpan sasTokenTimeToLive = default,
             int sasTokenRenewalBuffer = default)
             : base(
-                deviceId,
-                moduleId,
-                sasTokenTimeToLive,
-                sasTokenRenewalBuffer)
+                  deviceId,
```
drwill-ms, line 24: This was already indented 4 spaces. Why add more?

iothub/device/src/Edge/EdgeModuleClientFactory.cs
drwill-ms: I concur. As a rule, I don't like using environment variables for client library input, aside from the Azure Edge case.

iothub/device/tests/Authentication/DeviceClientConnectionStringExceptionTests.cs
```diff
@@ -37,7 +37,7 @@ public void DeviceClientConnectionStringDefaultScopeDefaultCredentialTypeMissing
         }
 
         [TestMethod]
-        [ExpectedException(typeof(ArgumentException))]
+        [ExpectedException(typeof(InvalidOperationException))]
```
drwill-ms, line 40: Is the client documented with this new exception type?

iothub/device/src/Authentication/DeviceAuthenticationWithConnectionString.cs
abhipsaMisra: This will also work. I'll change it to this to keep it clear.

iothub/device/src/Authentication/DeviceAuthenticationWithRegistrySymmetricKey.cs
abhipsaMisra: This looks like an unintentional copy-paste issue. I'll update it.

iothub/device/src/Authentication/IotHubConnectionString.cs
abhipsaMisra: sure, I'll add some comments

iothub/device/src/Authentication/ModuleAuthenticationWithSakRefresh.cs
abhipsaMisra: Probably another copy-paste side-effect. I'll take another look at the spacing here.

iothub/device/tests/Authentication/DeviceClientConnectionStringExceptionTests.cs
abhipsaMisra: Good point, I'll check to make sure.

iothub/device/src/Edge/EdgeModuleClientFactory.cs
abhipsaMisra: The environmental variable is intended to be for Azure Edge case only. That is the only scenario where ModuleAuthenticationWithHsm is used. This is the internal IAuthenticationMethod the SDK creates for you when the user initializes a ModuleClient through the "CreateFromEnvironmentAsync" flow. 
Since we ask the user to provide all other credential params through environment variables, i.e. iot hub name, device Id, module Id, edge URI Id, etc.), it seems reasonable to me to have them pass the sas ttl through env variables as well. That said, I do mean to confirm with the Edge team before introducing any additional env variables.

iothub/device/src/Authentication/DeviceAuthenticationWithConnectionString.cs
abhipsaMisra: This, and below: addressed in https://github.com/Azure/azure-iot-sdk-csharp/pull/2630/files

PR #2614, opened 2022-08-22
Service client AMQP layer redo

timtay-microsoft:
The v1 AMQP layer implementation was confusing and poorly abstracted from the service clients that used it. Service clients would have access to AMQP links for example. With this re-implementation, service clients will only have access to the AMQP connection and that connection will have the needed abstractions for sessions, sender links, and receiver links.

This rework also makes it so the AMQP layer is no longer disposable which lends itself to the open/close/re-open pattern that we need here.

PR #2628, opened 2022-08-25
Update service client E2E tests

abhipsaMisra:


e2e/test/iothub/service/ServiceClientE2ETests.cs
```diff
@@ -34,29 +35,29 @@ public async Task Message_NoTimeoutPassed()
 
         private async Task FastTimeout()
         {
-            TimeSpan? timeout = TimeSpan.FromTicks(10).Negate();
-            await TestTimeout(timeout).ConfigureAwait(false);
+            TimeSpan timeout = TimeSpan.FromTicks(10).Negate();
+            using var cts = new CancellationTokenSource(timeout);
```
timtay-microsoft, line 39: ```new CancellationToken(true);``` is all you need if you want a pre-cancelled cancellation token

e2e/test/iothub/service/ServiceClientE2ETests.cs
```diff
@@ -34,29 +35,29 @@ public async Task Message_NoTimeoutPassed()
...
+        private async Task TestTimeout(CancellationToken token = default)
         {
             using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(Logger, DevicePrefix).ConfigureAwait(false);
             using var sender = new IotHubServiceClient(TestConfiguration.IoTHub.ConnectionString);
 
             var sw = new Stopwatch();
             sw.Start();
 
-            Logger.Trace($"Testing ServiceClient SendAsync() timeout in ticks={timeout?.Ticks}");
+            Logger.Trace($"Testing ServiceClient SendAsync() cancellation.");
```
timtay-microsoft, line 56: David has [expressed some interest](https://github.com/Azure/azure-iot-sdk-csharp/pull/2614#discussion_r953150296) in removing these tests

timtay-microsoft (approved):

e2e/test/iothub/service/ServiceClientE2ETests.cs
abhipsaMisra: There's some value in ensuring that the cancellation token is propagated to the transport layer, but we could probably do that through unit tests as well.

drwill-ms: Absolutely, should be a unit test to confirm a pre-canceled token aborts the operational call before it is made.

PR #2629, opened 2022-08-25
Remove certificate information from transport settings

abhipsaMisra:


iothub/device/src/Authentication/IConnectionCredentials.cs
```diff
@@ -3,31 +3,31 @@
...
         /// </summary>
         string GatewayHostName { get; }
 
         /// <summary>
-        /// Indicates if the client is connecting to IoT hub service through a gateway service.
+        /// The host service that this client connects to.
+        /// This can either be the IoT hub name or a gateway service name.
         /// </summary>
-        bool IsUsingGateway { get; }
+        string HostName { get; }
```
abhipsaMisra, line 30: Added another field to store the host service name that the client connects to.

iothub/device/src/ClientFactory.cs
```diff
@@ -131,19 +130,6 @@ private static void EnsureOptionsIsSetup(X509Certificate2 cert, ref IotHubClient
             {
                 options.FileUploadTransportSettings = new();
             }
-
-            if (cert != null)
```
abhipsaMisra, line 135: We don't need to propagate the certificate information this way anymore. Certs will be stored in IotHubConnectionCredentials and will be propagated to the transport layer.

iothub/device/src/IotHubModuleClient.cs
```diff
@@ -456,12 +456,6 @@ private async Task<MethodResponse> InvokeMethodAsync(Uri uri, MethodRequest meth
                     IotHubConnectionCredentials = InternalClient.IotHubConnectionCredentials,
                 };
 
-                // We need to add the certificate to the httpTransport if DeviceAuthenticationWithX509Certificate
-                if (InternalClient.Certificate != null)
-                {
-                    transportSettings.ClientCertificate = InternalClient.Certificate;
```
abhipsaMisra, line 462: no longer required

iothub/device/src/Pipeline/TransportHandlerFactory.cs
```diff
@@ -31,8 +31,7 @@ public IDelegatingHandler Create(PipelineContext context)
             {
                 return new HttpTransportHandler(
                     context,
-                    iotHubClientHttpSettings,
-                    isClientPrimaryTransportHandler: true);
```
abhipsaMisra, line 35: remove some more remnants of AuthenticationWithTokenRefresh disposal

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -28,7 +28,7 @@ internal AmqpAuthenticationRefresher(IConnectionCredentials connectionCredential
             _connectionCredentials = connectionCredentials;
             _audience = CreateAmqpCbsAudience(_connectionCredentials);
             _amqpIotCbsTokenProvider = new AmqpIotCbsTokenProvider(_connectionCredentials);
-            _amqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, _connectionCredentials.GatewayHostName, CommonConstants.DefaultAmqpSecurePort).Uri;
+            _amqpEndpoint = new UriBuilder(CommonConstants.AmqpsScheme, _connectionCredentials.HostName, CommonConstants.DefaultAmqpSecurePort).Uri;
```
abhipsaMisra, line 31: GatewayHostName -> HostName

timtay-microsoft (approved):

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -295,10 +296,11 @@ private void Validate()
                 throw new FormatException("IoT hub service hostname to connect to must be specified.");
             }
 
-            // Gateway host name
-            if (GatewayHostName.IsNullOrWhiteSpace())
+            // Host name
+            if (HostName.IsNullOrWhiteSpace())
             {
-                throw new FormatException("Gateway service hostname to connect to must be specified.");
+                throw new FormatException("Service hostname to connect to must be specified." +
```
drwill-ms, line 302:
```suggestion
                throw new FormatException("Service host name to connect to must be specified." +
```

iothub/device/src/ClientFactory.cs
```diff
@@ -131,19 +130,6 @@ private static void EnsureOptionsIsSetup(X509Certificate2 cert, ref IotHubClient
             {
                 options.FileUploadTransportSettings = new();
```
drwill-ms, line 131: We can change the definition of IotHubClientOptions to create an instance of this class by default, so this whole method will no longer be needed.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapterFactory.cs
```diff
@@ -23,8 +23,8 @@ public MqttIotHubAdapterFactory(IotHubClientMqttSettings settings)
             return new MqttIotHubAdapter(
                 connectionCredentials.DeviceId,
                 connectionCredentials.ModuleId,
-                connectionCredentials.GatewayHostName,
-                connectionCredentials.Certificate == null ? connectionCredentials : null,
+                connectionCredentials.HostName,
+                connectionCredentials,
```
drwill-ms, line 27: If we pass in the connecitonCredentials object, no reason to pass the previous 3 as the class can get those values directly now.

iothub/device/tests/Authentication/DeviceAuthenticationWithTokenRefreshTests.cs
```diff
@@ -154,7 +154,7 @@ public async Task DeviceAuthenticationWithSakRefresh_SharedAccessKeyConnectionSt
             Assert.IsNotNull(iotHubConnectionCredentials.SasTokenRefresher);
             Assert.IsInstanceOfType(iotHubConnectionCredentials.SasTokenRefresher, typeof(DeviceAuthenticationWithSakRefresh));
 
-            var auth = (IAuthorizationProvider)iotHubConnectionCredentials;
+            var auth = iotHubConnectionCredentials;
```
drwill-ms, line 157: Can we do away with the alias and just use `iotHubConnectionCredentials` directly in line 160 below?

iothub/device/tests/Authentication/ModuleAuthenticationWithTokenRefreshTests.cs
```diff
@@ -90,7 +90,7 @@ public async Task ModuleAuthenticationWithSakRefresh_SharedAccessKeyConnectionSt
             Assert.IsNotNull(iotHubConnectionCredentials.SasTokenRefresher);
             Assert.IsInstanceOfType(iotHubConnectionCredentials.SasTokenRefresher, typeof(ModuleAuthenticationWithSakRefresh));
 
-            var auth = (IAuthorizationProvider)iotHubConnectionCredentials;
+            var auth = iotHubConnectionCredentials;
```
drwill-ms, line 93: same here

iothub/device/src/ClientFactory.cs
abhipsaMisra: good point, I'll update it

iothub/device/src/Transport/Mqtt/MqttIotHubAdapterFactory.cs
abhipsaMisra: good catch, sure

iothub/device/src/ClientFactory.cs
abhipsaMisra: looks like we do this already, so I'll go ahead and remove this method: https://github.com/Azure/azure-iot-sdk-csharp/blob/0ced7919b601a2aee98bd30e46bb8fdaee017984/iothub/device/src/IotHubClientOptions.cs#L46

PR #2630, opened 2022-08-25
Update MqttIotHubAdapter to not duplicate values

abhipsaMisra:
Feedback from #2629 .

### DIFF
--- a/doc/coding-style.md
+++ b/doc/coding-style.md
@@ -57,6 +57,13 @@ To enforce this, we make use of an .editorconfig file, and [CodeMaid](https://ma
             param3)
         .ConfigureAwait(false);
     ```
+1. Order of tags in documentation comments:
+  - `<summary>` - Describe a type or a member of a type
+  - `<remarks>` - Describe additional information about a type
+  - `<param>` - Describe a parameter for a method or constructor
+  - `<returns>` - Describe the return value of a method
+  - `<exception>` - Identifies the exceptions a method can throw
+  - `<example>` - Indicate an example
 
 ## Examples
 
--- a/e2e/test/iothub/service/ServiceClientE2ETests.cs
+++ b/e2e/test/iothub/service/ServiceClientE2ETests.cs
@@ -35,9 +35,8 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
 
         private async Task FastTimeout()
         {
-            TimeSpan timeout = TimeSpan.FromTicks(10).Negate();
-            using var cts = new CancellationTokenSource(timeout);
-            await TestTimeout(cts.Token).ConfigureAwait(false);
+            var cancellationToken = new CancellationToken(true);
+            await TestTimeout(cancellationToken).ConfigureAwait(false);
         }
 
         private async Task DefaultTimeout()
--- a/iothub/device/src/Authentication/AuthenticationMethodFactory.cs
+++ b/iothub/device/src/Authentication/AuthenticationMethodFactory.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Azure.Devices.Client
                         iotHubConnectionString.SharedAccessSignature);
             }
 
-            throw new InvalidOperationException($"Unsupported authentication method in '{iotHubConnectionString}'.");
+            throw new ArgumentException($"Unsupported authentication method in '{iotHubConnectionString}'.");
         }
     }
 }
--- a/iothub/device/src/Authentication/AuthenticationMethodFactory.cs
+++ b/iothub/device/src/Authentication/AuthenticationMethodFactory.cs
@@ -42,7 +42,8 @@ namespace Microsoft.Azure.Devices.Client
                         iotHubConnectionString.SharedAccessSignature);
             }
 
-            throw new ArgumentException($"Unsupported authentication method in '{iotHubConnectionString}'.");
+            throw new ArgumentException($"Should specify either SharedAccessKey or SharedAccessSignature" +
+                $" if connection string is used for authenticating the client with IoT hub.");
         }
     }
 }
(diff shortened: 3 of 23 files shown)


drwill-ms (approved):

doc/coding-style.md
```diff
@@ -57,6 +57,13 @@ To enforce this, we make use of an .editorconfig file, and [CodeMaid](https://ma
             param3)
         .ConfigureAwait(false);
     ```
+1. Order of tags in documentation comments:
```
abhipsaMisra, line 60: Added our tags ordering guideline to our code-style readme

e2e/test/iothub/service/ServiceClientE2ETests.cs
```diff
@@ -35,9 +35,8 @@ public async Task Message_NoTimeoutPassed()
 
         private async Task FastTimeout()
         {
-            TimeSpan timeout = TimeSpan.FromTicks(10).Negate();
-            using var cts = new CancellationTokenSource(timeout);
-            await TestTimeout(cts.Token).ConfigureAwait(false);
+            var cancellationToken = new CancellationToken(true);
```
abhipsaMisra, line 38: creating new CancellationToken(true) for an already cancelled cancellation token.

iothub/device/src/Authentication/AuthenticationMethodFactory.cs
```diff
@@ -42,7 +42,7 @@ internal static IAuthenticationMethod GetAuthenticationMethodFromConnectionStrin
                         iotHubConnectionString.SharedAccessSignature);
             }
 
-            throw new InvalidOperationException($"Unsupported authentication method in '{iotHubConnectionString}'.");
+            throw new ArgumentException($"Unsupported authentication method in '{iotHubConnectionString}'.");
```
abhipsaMisra, line 45: We're throwing an ArgumentException for all other cases where our input did not match the expected value, so it made sense to update this one to fit too. 
Let me know what you think.

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -20,7 +20,11 @@ public sealed class IotHubConnectionCredentials : IConnectionCredentials
         /// <param name="authenticationMethod">The authentication method that is used.</param>
         /// <param name="iotHubHostName">The fully-qualified DNS host name of IoT hub.</param>
         /// <param name="gatewayHostName">The fully-qualified DNS host name of the gateway (optional).</param>
-        /// <returns>A new instance of the <see cref="IotHubConnectionCredentials"/> class with a populated connection string.</returns>
+        /// <returns>A new instance of the <c>IotHubConnectionCredentials</c> class with a populated connection string.</returns>
```
abhipsaMisra, line 23: documented the possible exceptions thrown by this method

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -291,23 +299,13 @@ private void SetAuthenticationModel()
         private void Validate()
         {
             // IoT Hub host name
-            if (IotHubHostName.IsNullOrWhiteSpace())
-            {
-                throw new FormatException("IoT hub service hostname to connect to must be specified.");
-            }
+            Argument.AssertNotNullOrWhiteSpace(IotHubHostName, nameof(IotHubHostName));
```
abhipsaMisra, line 302: Updated the FormatException to ArgumentException/ArgumentNullException. We encounter this even in the non-connection string flow, i.e. through an IAuthenticationMethod flow, so it made sense to throw the generic argument exceptions.

iothub/device/tests/Authentication/DeviceClientConnectionStringExceptionTests.cs
```diff
@@ -37,7 +37,7 @@ public void DeviceClientConnectionStringDefaultScopeDefaultCredentialTypeMissing
         }
 
         [TestMethod]
-        [ExpectedException(typeof(InvalidOperationException))]
+        [ExpectedException(typeof(ArgumentException))]
```
abhipsaMisra, line 40: These tests are now back to asserting an ArgumentException

abhipsaMisra: Note on why this changed even though we didn't change the exception type in the previous PR (#2625 ):
Previously, the SDK would first throw an InvalidOperationException, and then, in trying to throw the exception, would do a ".ToString()" on the class (IotHUbConnectionStringBuilder). The "ToString()" method here would again throw an ArgumentException, which was returned to the caller.

This flow is now updated to throw an ArgumentException right from the beginning.

iothub/device/src/Authentication/AuthenticationMethodFactory.cs
drwill-ms: But we're outside the path of a customer sending this argument, right?

What value do you use for the `paramName` parameter that would make sense to the user?

drwill-ms (approved):

iothub/device/src/Authentication/AuthenticationMethodFactory.cs
abhipsaMisra: I'm leaning towards not throwing this exception in this method, but instead delegating the check to IotHubConnectionCredentials.Validate() method. This method will call out the specific property that is unset, in this case that being:

https://github.com/Azure/azure-iot-sdk-csharp/blob/327e7b5ecfc791e19efaca9792dd32fb401e73e7/iothub/device/src/Authentication/IotHubConnectionCredentials.cs#L325-L335

drwill-ms: As long as when the user gets the exception, it matches what they see in the API surface they called.

If needed, you can catch this exception, grab the exception message, and create a new one with the parameter name you know about.

abhipsaMisra: Updated to:
```c#
throw new ArgumentException($"Should specify either SharedAccessKey or SharedAccessSignature" +
    $" if connection string is used for authenticating the client with IoT hub.");
```

PR #2632, opened 2022-08-25
RetryDelegationHandler removed implicit OpenAsync()

tmahmood-microsoft:
## Checklist


iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -607,41 +600,12 @@ await _internalRetryPolicy
             }
         }
 
-        public override Task OpenAsync(CancellationToken cancellationToken)
+        public override async Task OpenAsync(CancellationToken cancellationToken)
         {
-            return EnsureOpenedAsync(true, cancellationToken);
+            await OpenAsync(cancellationToken, true).ConfigureAwait(false);
```
drwill-ms, line 605: By convention, cancellation tokens always go last in the parameter list.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -699,6 +663,55 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
...
+                _handlerSemaphore?.Release();
+                Dispose(true);
+            }
+        }
+
+        private async Task EnsureOpenedAsync()
+        {
+            try
+            {
+                await _handlerSemaphore.WaitAsync().ConfigureAwait(false);
```
drwill-ms, line 696: Is a semaphore still required just to check connection state?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -607,41 +600,12 @@ await _internalRetryPolicy
             }
         }
 
-        public override Task OpenAsync(CancellationToken cancellationToken)
+        public override async Task OpenAsync(CancellationToken cancellationToken)
```
drwill-ms, line 603: What is the purpose of this method?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -699,6 +663,55 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
...
+        private async Task EnsureOpenedAsync()
+        {
+            try
+            {
+                await _handlerSemaphore.WaitAsync().ConfigureAwait(false);
+                if (!_opened)
+                {
+                    if (!_openCalled)
+                    {
+                        throw new InvalidOperationException($"Explicit call to {nameof(OpenAsync)} is missing before this operation.");
```
drwill-ms, line 701:
```suggestion
                        throw new InvalidOperationException($"The client connection must be opened before operations can begin. Call '{nameof(OpenAsync)}' and try again.");
```

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -699,6 +663,55 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
...
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"Transport disconnected, call to {nameof(OpenAsync)} is required before this operation.");
```
drwill-ms, line 705:
```suggestion
                        throw new InvalidOperationException($"The transport has disconnected; call {'nameof(OpenAsync)}' to reconnect.");
```

drwill-ms: FWIW, I don't think we support a call to OpenAsync again (yet). It is in our list of things to do, but I'd be surprised if it works already.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
tmahmood-microsoft: To keep override implementation. As its used in all unit tests. 
https://github.com/Azure/azure-iot-sdk-csharp/blob/95532a519f6081e6987b04b7f8a571e7e1e98671/iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs#L65

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
tmahmood-microsoft: few [unit tests](https://github.com/Azure/azure-iot-sdk-csharp/blob/95532a519f6081e6987b04b7f8a571e7e1e98671/iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs#L21) were failing due to it. 
While digging into it, I found [this](https://github.com/Azure/azure-iot-sdk-csharp/blob/95532a519f6081e6987b04b7f8a571e7e1e98671/iothub/device/src/Pipeline/RetryDelegatingHandler.cs#L808) call which is invoked [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/95532a519f6081e6987b04b7f8a571e7e1e98671/iothub/device/src/Pipeline/RetryDelegatingHandler.cs#L649)
Looks like it was due to semaphore behavior.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
tmahmood-microsoft: fixed in latest commit 
https://github.com/Azure/azure-iot-sdk-csharp/pull/2632/commits/47c9ce49828b188a309c818f8486485e2a38adbe

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
tmahmood-microsoft: Fixed in latest commit
https://github.com/Azure/azure-iot-sdk-csharp/pull/2632/commits/47c9ce49828b188a309c818f8486485e2a38adbe

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
tmahmood-microsoft: Fixed in latest commit
https://github.com/Azure/azure-iot-sdk-csharp/pull/2632/commits/47c9ce49828b188a309c818f8486485e2a38adbe

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: So, yes?

Any way to find a way not to require it? Perhaps not in this PR, but in a future task?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: Can't the tests just be updated?

drwill-ms: Maybe what I really mean is, do we need both OpenAsync(CancellationToken) and OpenAsync(bool, CancellationToken). Can we combine them and just have one?

@timtay-microsoft and @abhipsaMisra, what does Java do? Can we have with retry be the default because the client has a retry policy?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
tmahmood-microsoft: Yes, a semaphore is required in this case to check the connection state.

Removing the dependency sounds like a good task. I'll look more into it for a future change.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
timtay-microsoft: Java just has open(bool) since we agreed that it would be best for the user to be aware if their open call had retry baked in or not.

drwill-ms: If all the other methods have retry built-in, why not OpenAsync?

abhipsaMisra: I forget if there were any implementation related concerns in having retry baked into the Open call. If not, then I agree that it makes sense to have the Open call retry until it succeeds/the cancellation token expires.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -699,6 +663,55 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
...
+            {
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, cancellationToken, nameof(CloseAsync));
+
+                _handlerSemaphore?.Release();
+                Dispose(true);
+            }
+        }
+
+        private async Task EnsureOpenedAsync()
```
drwill-ms, line 692: Why not pass in the cancellation token? If the user wants to cancel, this should cancel too.

tmahmood-microsoft: Added requested changes.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
tmahmood-microsoft: removed extra OpenSync() implementation and retry logic built-in as default.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -699,64 +657,86 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
...
-                    .RunWithRetryAsync(
-                        async () =>
-                        {
-                            try
-                            {
-                                if (Logging.IsEnabled)
-                                    Logging.Enter(this, cancellationToken, nameof(OpenAsync));
+                if (!_openCalled)
+                {
+                    return;
```
drwill-ms, line 667:
```suggestion
                    // Already closed so gracefully exit, instead of throw.
                    return;
```

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -699,64 +657,86 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
...
+                }
 
-                                // Will throw on error.
-                                await base.OpenAsync(cancellationToken).ConfigureAwait(false);
+                if (Logging.IsEnabled)
```
drwill-ms, line 670: Should logging happen before the previous check? I think probably yes.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -699,64 +657,86 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
...
-                    if (Logging.IsEnabled)
-                        Logging.Exit(this, cancellationToken, nameof(OpenAsync));
+                    if (!_openCalled)
+                    {
+                        throw new InvalidOperationException($"The client connection must be opened before operations can begin. Call '{nameof(OpenAsync)}' and try again.");
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"The transport has disconnected; call '{nameof(OpenAsync)}' to reconnect.");
+                    }
```
drwill-ms, line 700: Consider:
```suggestion
                    throw new InvalidOperationException(_openCalled
                        ? $"The transport has disconnected; call '{nameof(OpenAsync)}' to reconnect."
                        : $"The client connection must be opened before operations can begin. Call '{nameof(OpenAsync)}' and try again.");
```

drwill-ms (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
tmahmood-microsoft: updated

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.Azure.Devices.Client;
 using Microsoft.Azure.Devices.Common.Exceptions;
 using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.Rest;
```
timtay-microsoft, line 13: Unused?

iothub/device/src/InternalClient.cs
```diff
@@ -181,7 +181,7 @@ public void SetConnectionStatusChangeHandler(Action<ConnectionStatusInfo> status
         {
             if (Logging.IsEnabled)
                 Logging.Enter(this, callback, userContext, nameof(SetDesiredPropertyUpdateCallbackAsync));
-            
+
```
timtay-microsoft, line 184: No documentation changes for either DeviceClient or ModuleClient about the behavior of SendEventAsync requiring users to call OpenAsync first?

timtay-microsoft: Also missing documentation on what exception to expect if the user tries to perform a send operation when the client is closed

timtay-microsoft (changes requested):
Needs documentation changes so users know to expect this new behavior, but otherwise looks good

iothub/device/src/InternalClient.cs
tmahmood-microsoft: Currently, re-opening after close does not work but it should after this: https://msazure.visualstudio.com/One/_workitems/edit/14095155

Added documentation that client can be re-opened.

timtay-microsoft (approved):
Still missing documentation explaining what exception is thrown when SendEventAsync and similar APIs are called without opening first, but otherwise looks good

PR #2637, opened 2022-08-26
Remove TaskHelpers from several libraries, and use Task.Completed tas…

drwill-ms:
TaskHelpers was a utility to help with NuGet targets that didn't support Task.CompletedTask. Now that we don't target those .NET versions anymore, we don't need that.

Also, moved the remaining methods to the classes that uses them (if singular) so I could remove the TaskHelpers.cs file.


brycewang-microsoft (commented):
"provisioning/device/src/Transports/Amqp/TaskHelpers.cs" is not removed as a whole class, is this expected?

drwill-ms:
> "provisioning/device/src/Transports/Amqp/TaskHelpers.cs" is not removed as a whole class, is this expected?

I removed others, but not that one because it is referenced from multiple classes in that project.

brycewang-microsoft (approved):

timtay-microsoft (approved):

PR #2640, opened 2022-08-26
More code cleanup with Abhipsa and Tim

drwill-ms:


iothub/service/src/Amqp/AmqpsConstants.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices.Common
+{
+    internal static class AmqpsConstants
```
drwill-ms, line 6: Moved from CommonConstants to the AMQP folder where they are relevant.

iothub/service/src/Utilities/Extensions/HttpHeadersExtensions.cs
```diff
@@ -1,17 +1,16 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http.Headers;
 
-namespace Microsoft.Azure.Devices.Utilities
```
drwill-ms, line 9: Folders give us a way to organize files, but we don't want to pay the cost of including lots of extra namespaces.

iothub/service/src/Utilities/Extensions/StringFormattingExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    internal static class StringFormattingExtensions
+    {
+        /// <summary>
+        /// Format string to be displayed.
+        /// </summary>
+        /// <param name="format">A composite format string.</param>
+        /// <param name="args">The object to format.</param>
+        /// <returns>A copy of format in which the format item or items have been replaced by the string representation of arg0.</returns>
+        internal static string FormatInvariant(this string format, params object[] args)
```
drwill-ms, line 20: This was left as a TODO. Tim will follow up with someone.

iothub/service/src/Utilities/Utils.cs
```diff
@@ -1,14 +1,13 @@
...
 
-using System;
 using System.Reflection;
 
 namespace Microsoft.Azure.Devices
 {
     internal class Utils
     {
-        public static string GetClientVersion()
+        internal static string GetClientVersion()
```
drwill-ms, line 10: We've decided to make public methods on internal classes internal instead. They aren't really public and it can be confusing in code reviews.

iothub/device/tests/Pipeline/RetryDelegatingHandlerImplicitOpenTests.cs
```diff
@@ -134,7 +134,7 @@ public async Task OpenAsyncNextFailedSutIsOpenAndCanBeReopen()
                 }
                 else
                 {
-                    return Task.CompletedTask; ;
+                    return Task.CompletedTask;
```
timtay-microsoft, line 137:
```suggestion
                    return Task.CompletedTask;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```

Just in case

timtay-microsoft (approved):

abhipsaMisra (approved):

andyk-ms (approved):

PR #2641, opened 2022-08-26
Enable user to provide ClientWebSocket instance

tmahmood-microsoft:
## Checklist

### DIFF
--- a/iothub/device/src/Edge/CustomCertificateValidator.cs
+++ b/iothub/device/src/Edge/CustomCertificateValidator.cs
@@ -41,15 +41,7 @@ namespace Microsoft.Azure.Devices.Client.Edge
         {
             Debug.WriteLine("CustomCertificateValidator.SetupCertificateValidation()");
 
-            if (_transportSettings is IotHubClientAmqpSettings amqpTransportSettings)
-            {
-                if (amqpTransportSettings.RemoteCertificateValidationCallback == null)
-                {
-                    amqpTransportSettings.RemoteCertificateValidationCallback =
-                        (sender, certificate, chain, sslPolicyErrors) => ValidateCertificate(_certs.First(), certificate, chain, sslPolicyErrors);
-                }
-            }
-            else if (_transportSettings is IotHubClientMqttSettings mqttTransportSettings)
+            if (_transportSettings is IotHubClientMqttSettings mqttTransportSettings)
             {
                 if (mqttTransportSettings.RemoteCertificateValidationCallback == null)
                 {
--- a/iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
+++ b/iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
@@ -358,7 +358,6 @@ namespace Microsoft.Azure.Amqp.Transport
             }
         }
 
-
         private static IAsyncResult ToAsyncResult(Task task, AsyncCallback callback, object state)
         {
             if (task.AsyncState == state)
--- a/iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
+++ b/iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
@@ -6,7 +6,6 @@ using System.Linq;
 using System.Net;
 using System.Net.Security;
 using System.Net.WebSockets;
-using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
@@ -49,8 +48,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIot
             {
                 TargetHost = hostName,
                 Certificate = null,
-                CertificateValidationCallback = _amqpTransportSettings.RemoteCertificateValidationCallback
-                    ?? OnRemoteCertificateValidation,
+                CertificateValidationCallback = OnRemoteCertificateValidation,
                 Protocols = amqpTransportSettings.SslProtocols,
             };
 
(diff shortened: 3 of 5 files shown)


iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
```diff
@@ -358,7 +358,6 @@ private static void ValidateBufferBounds(byte[] buffer, int offset, int size)
             }
         }
 
-
```
abhipsaMisra, line 361: I also recommend updating the namespace this class is under from an Azure.Amqp namespace to one under Microsoft.Azure.Devices.Client.

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -124,38 +122,13 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
                 if (Logging.IsEnabled)
                     Logging.Enter(this, nameof(CreateClientWebSocketAsync));
 
-                var websocket = new ClientWebSocket();
+                bool customWebSocket = _amqpTransportSettings.ClientWebSocket != null;
+                ClientWebSocket websocket = customWebSocket ? _amqpTransportSettings.ClientWebSocket : new ClientWebSocket();
+                IWebProxy webProxy = customWebSocket ? websocket.Options.Proxy : _amqpTransportSettings.Proxy;
```
abhipsaMisra, line 127: Where is this webProxy being used?

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -124,38 +122,13 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
...
-                    }
-                }
-                catch (PlatformNotSupportedException)
-                {
-                    // .NET Core 2.0 doesn't support proxy. Ignore this setting.
-                    if (Logging.IsEnabled)
-                        Logging.Error(this, $"{nameof(CreateClientWebSocketAsync)} PlatformNotSupportedException thrown as .NET Core 2.0 doesn't support proxy");
-                }
-
-                if (_amqpTransportSettings.WebSocketKeepAlive.HasValue)
```
abhipsaMisra, line 152: Is this no longer being set on the webSocket? The customer likely won't know to set this, so we should check their supplied websocket instance to see if they have set it, else set it for them. In the case where we create the web socket for them, we still need to set it.

iothub/device/src/TransportSettings/IotHubClientAmqpSettings.cs
```diff
@@ -72,25 +73,19 @@ public TimeSpan OperationTimeout
         }
 
         /// <summary>
-        /// A keep-alive for the transport layer in sending ping/pong control frames when using web sockets.
-        /// </summary>
-        /// <seealso href="https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocketoptions.keepaliveinterval"/>
-        public TimeSpan? WebSocketKeepAlive { get; set; }
```
abhipsaMisra, line 78: Removed?

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -49,8 +48,7 @@ internal class AmqpIotTransport : IDisposable
             {
                 TargetHost = hostName,
                 Certificate = null,
-                CertificateValidationCallback = _amqpTransportSettings.RemoteCertificateValidationCallback
-                    ?? OnRemoteCertificateValidation,
+                CertificateValidationCallback = OnRemoteCertificateValidation,
```
abhipsaMisra, line 51: Ah, this looks like a gap in our plan. The remote certificate validation callback is being set into the TLS settings, so we'll need a way for users to be able to pass this to us for tcp-based connections as well.

iothub/device/src/TransportSettings/IotHubClientAmqpSettings.cs
drwill-ms: Are we taking the path that if they want this stuff set, they provide the whole instance?

drwill-ms: And is that consistent with our similar situations, e.g., HttpClient?

iothub/device/src/Edge/CustomCertificateValidator.cs
```diff
@@ -41,15 +41,7 @@ private void SetupCertificateValidation()
...
-            if (_transportSettings is IotHubClientAmqpSettings amqpTransportSettings)
-            {
-                if (amqpTransportSettings.RemoteCertificateValidationCallback == null)
-                {
-                    amqpTransportSettings.RemoteCertificateValidationCallback =
-                        (sender, certificate, chain, sslPolicyErrors) => ValidateCertificate(_certs.First(), certificate, chain, sslPolicyErrors);
-                }
-            }
-            else if (_transportSettings is IotHubClientMqttSettings mqttTransportSettings)
+            if (_transportSettings is IotHubClientMqttSettings mqttTransportSettings)
```
drwill-ms, line 44: Why is AMQP path being removed and not MQTT?

drwill-ms (changes requested):
ClientWebSocket inherits from WebSocket which is IDisposable. What is our plan for disposing this when our client is disposed? Do we always dispose it? Do we leave that to the user to do? Do we tell them that? Do we give them an option, like other .NET cases do?

https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocket?view=net-6.0

PR #2644, opened 2022-08-30
TransportType instead of useWebSocketOnly

tmahmood-microsoft:
## Checklist


iothub/service/src/TransportType.cs
```diff
@@ -7,24 +7,18 @@ namespace Microsoft.Azure.Devices
     /// Transport types supported by MessagingClient, FileUploadNotificationProcessorClient and MessageFeedbackProcessorClient.
     /// </summary>
     /// <remarks>
-    /// Amqp and Amqp over WebSocket.
+    /// Only supports Amqp and Amqp over WebSocket.
```
drwill-ms, line 10:
```suggestion
    /// Only supports AMQP over TCP and AMQP over web socket.
```

iothub/service/src/TransportType.cs
```diff
@@ -7,24 +7,18 @@ namespace Microsoft.Azure.Devices
...
     /// </remarks>
     public enum TransportType
```
drwill-ms, line 12: Can we adopt the naming scheme we came up with in the device client? See `IotHubClientTransportProtocol.cs`. For service I'd probably name it `IotHubTransportProtocol`.

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -57,12 +57,12 @@ public class IotHubServiceClientOptions
...
 
         /// <summary>
-        /// Whether to use web sockets or not.
+        /// The configured transport protocol.
         /// </summary>
         /// <remarks>
         /// Only used for AMQP. Can only be used for <see cref="MessagingClient"/> and <see cref="MessageFeedbackProcessorClient"/> and <see cref="FileUploadNotificationProcessorClient"/>.
         /// </remarks>
-        public bool UseWebSocketOnly { get; set; }
+        public TransportType Transport { get; set; } = TransportType.Tcp;
```
drwill-ms, line 65: In my view, the transport is HTTP or AMQP, and the protocol is TCP or web socket. Unless I got that wrong, I'd name this property `Protocol`.

drwill-ms (approved):

iothub/service/src/FileUpload/FileUploadNotificationProcessorClient.cs
```diff
@@ -53,7 +53,7 @@ protected FileUploadNotificationProcessorClient()
         {
             _hostName = hostName;
             _credentialProvider = credentialProvider;
-            _connection = new IotHubConnection(credentialProvider, options.UseWebSocketOnly, options);
+            _connection = new IotHubConnection(credentialProvider, options.Transport == TransportType.WebSocket, options);
```
drwill-ms, line 56: Should IotHubConnection take the enum instead of a bool?

drwill-ms (approved):

iothub/service/src/IotHubServiceClientOptions.cs
timtay-microsoft: In the device client side, this enum was called ```IotHubClientTransportProtocol``` FYI

timtay-microsoft: Since it is the same concept here in the service client side, we should match the name used

iothub/service/src/TransportType.cs
```diff
@@ -7,24 +7,18 @@ namespace Microsoft.Azure.Devices
...
     {
         /// <summary>
-        /// Advanced Message Queuing Protocol transport.
+        /// Communicate over TCP using the port 5671.
```
timtay-microsoft, line 15:
```suggestion
        /// Communicate over TCP using port 5671.
```

iothub/service/src/TransportType.cs
```diff
@@ -7,24 +7,18 @@ namespace Microsoft.Azure.Devices
     /// Transport types supported by MessagingClient, FileUploadNotificationProcessorClient and MessageFeedbackProcessorClient.
```
timtay-microsoft, line 7:
```suggestion
    /// Transport protocol types supported by MessagingClient, FileUploadNotificationProcessorClient and MessageFeedbackProcessorClient.
```

timtay-microsoft (approved):

PR #2646, opened 2022-08-31
Refactor DevicesClient ETag usage to match track 2

timtay-microsoft:
This PR is scoped to the DevicesClient. All other clients will follow afterwards.

This work is based on the pattern shown by the Azure SDK team's model client [ConfigurationClient](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs).

The basic pattern is:

```csharp
public async Task<Foo> SetAsync(Foo foo, bool onlyIfUnchanged = false, CancellationToken cancellationToken = default);

// Constructs a Foo instance for the user, then calls DeleteAsync overload with onlyIfUnchanged set to false
public async Task DeleteAsync(string fooId, CancellationToken cancellationToken = default);

public async Task DeleteAsync(Foo foo, bool onlyIfUnchanged = false, CancellationToken cancellationToken = default);
```

Notably, the Azure SDK team recommends that the default case is to ignore ETag and perform unconditionally. They advocate for _opting in_ to precondition checking rather than _opting out_. 

Also notably, they ask for a bool to decide what If-Match header to send as opposed to asking for the string itself to insert. I like this because it spares users from understanding everything about the If-Match header.

Finally, we are encouraged to use the Azure.Core ETag type instead of strings to represent ETags. This resulted in me needing to pull in a special serialization helper class from Microsoft.Azure.Core.NewtonsoftJson, but the dependency is quite small 

Also added some tests for this behavior since we did not have any yet


iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -91,36 +92,47 @@ internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage re
         /// <param name="eTag">The If-Match header value to sanitize before adding.</param>
         /// <exception cref="ArgumentNullException">Thrown when the provided <paramref name="requestMessage"/> or <paramref name="requestMessage"/> is null.</exception>
         /// <exception cref="ArgumentException">Thrown if the <paramref name="eTag"/> is empty or white space.</exception>
-        /// <exception cref="IotHubException">
-        /// Thrown if IoT hub responded to the request with a non-successful status code. For example, if the provided
-        /// request was throttled, <see cref="IotHubThrottledException"/> is thrown. For a complete list of possible
-        /// error cases, see <see cref="Common.Exceptions"/>.
-        /// </exception>
-        /// <exception cref="HttpRequestException">
```
timtay-microsoft, line 99: I'm not sure why we had these exceptions documented here when this method doesn't make any http calls. Likely just copy + paste error

iothub/service/src/Configurations/ConfigurationsClient.cs
```diff
@@ -241,7 +241,7 @@ public virtual async Task<Configuration> SetAsync(Configuration configuration, b
                 cancellationToken.ThrowIfCancellationRequested();
 
                 using HttpRequestMessage request = _httpRequestMessageFactory.CreateRequest(HttpMethod.Put, GetConfigurationRequestUri(configuration.Id), _credentialProvider, configuration);
-                HttpMessageHelper.InsertETag(request, configuration.ETag);
+                HttpMessageHelper.ConditionallyInsertETag(request, configuration.ETag);
```
drwill-ms, line 244: You aren't passing along the `forceUpdate` parameter.

timtay-microsoft: This PR is just for DevicesClient. We'll fix this later

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -91,36 +92,47 @@ internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage re
...
-        /// If the HTTP request fails due to an underlying issue such as network connectivity, DNS failure, or server
-        /// certificate validation.
-        /// </exception>
-        public static void InsertETag(HttpRequestMessage requestMessage, string eTag)
+        public static void ConditionallyInsertETag(HttpRequestMessage requestMessage, string eTag)
```
drwill-ms, line 95: internal?

drwill-ms, line 95: Why have this overload rather than force all callers to be explicit?

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -91,36 +92,47 @@ internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage re
...
+        /// <exception cref="ArgumentException">Thrown if the <paramref name="eTag"/> is empty or white space.</exception>
+        public static void ConditionallyInsertETag(HttpRequestMessage requestMessage, ETag eTag, bool onlyIfUnchanged = false)
         {
             Argument.AssertNotNullOrWhiteSpace(eTag, nameof(eTag));
             Argument.AssertNotNull(requestMessage, nameof(requestMessage));
 
-            // All ETag values need to be wrapped in escaped quotes, but the "forced" value
-            // is hardcoded with quotes so it can be skipped here
-            if (!ETagForce.Equals(eTag))
+            if (!onlyIfUnchanged)
```
drwill-ms, line 115: My mind feels warped. If not only if unchanged. Wait, what is the condition?

iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -52,6 +52,7 @@
   <!--Nuget package dependencies-->
   <ItemGroup>
     <PackageReference Include="Azure.Core" Version="1.22.0" />
+    <PackageReference Include="Microsoft.Azure.Core.NewtonsoftJson" Version="1.0.0" />
```
drwill-ms, line 55: We have to take this one, huh?

iothub/service/src/Registry/DevicesClient.cs
```diff
@@ -262,6 +239,12 @@ public virtual async Task DeleteAsync(string deviceId, CancellationToken cancell
         /// To force the operation to execute regardless of ETag, set the device identity's ETag to "*" or
         /// use <see cref="DeleteAsync(string, CancellationToken)"/>.
         /// </param>
+        /// <param name="onlyIfUnchanged">
+        /// If false, this delete operation will be performed even if the provided device identity has
+        /// an out of date ETag. If true, the operation will throw a <see cref="PreconditionFailedException"/>
```
drwill-ms, line 244: Does the service return an error that we map to this exception?

iothub/service/src/Registry/Models/Device.cs
```diff
@@ -49,8 +51,8 @@ public Device(string id)
         /// <summary>
         /// Device's ETag.
         /// </summary>
-        [JsonProperty(PropertyName = "etag")]
-        public string ETag { get; set; }
+        [JsonProperty(PropertyName = "etag"), JsonConverter(typeof(NewtonsoftJsonETagConverter))] // NewtonsoftJsonETagConverter is used here because otherwise the ETag isn't serialized properly
```
drwill-ms, line 54: Can we make them separate attributes?
```suggestion
        [JsonProperty(PropertyName = "etag")]
        [JsonConverter(typeof(NewtonsoftJsonETagConverter))] // NewtonsoftJsonETagConverter is used here because otherwise the ETag isn't serialized properly
```

iothub/service/src/Http/HttpMessageHelper.cs
timtay-microsoft: To not break existing clients

iothub/service/src/Http/HttpMessageHelper.cs
timtay-microsoft: I can remove the double negative in the logic here, sure

iothub/service/src/Microsoft.Azure.Devices.csproj
timtay-microsoft: The [class we need](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Microsoft.Azure.Core.NewtonsoftJson/src/NewtonsoftJsonETagConverter.cs) is rather small but so is the dependency. I could go either way on this one

iothub/service/src/Registry/DevicesClient.cs
timtay-microsoft: Yes, but the exception's shape is subject to change based on Bryce's work

iothub/service/src/Registry/Models/Device.cs
timtay-microsoft: Sure, I'll do that here and elsewhere

iothub/service/tests/ExportImportDeviceTests.cs
```diff
@@ -89,22 +48,22 @@ public void ExportImportDeviceTakingDeviceWithNullEtag()
             // nothing to Act on
 
             // Assert
-            Assert.IsTrue(exportimportDevice.ETag == null, "ETag was not set correctly");
-            Assert.IsTrue(exportimportDevice.TwinETag == null, "Twin ETag was not set correctly");
+            Assert.IsTrue(exportimportDevice.ETag.ToString() == "", "ETag was not set correctly");
```
drwill-ms, line 51: Consider taking the time to convert to fluent assertions
```suggestion
            exportimportDevice.ETag.Should().Be("", "ETag was not set correctly");
```

drwill-ms: Also, this way it prints out the value if it isn't what was expected.

iothub/service/tests/Registry/DevicesTests.cs
```diff
@@ -92,7 +93,7 @@ public async Task RegisterDeviceAsyncWithNullDeviceTest()
         public async Task RegisterDevicesAsyncWithETagsSetTest()
         {
             var goodDevice = new Device("123") { ConnectionState = DeviceConnectionState.Connected };
-            var badDevice = new Device("234") { ConnectionState = DeviceConnectionState.Connected, ETag = "234" };
+            var badDevice = new Device("234") { ConnectionState = DeviceConnectionState.Connected, ETag = new ETag("234") };
```
drwill-ms, line 96:
```suggestion
            var badDevice = new Device("234") { ConnectionState = DeviceConnectionState.Connected, ETag = new("234") };
```

iothub/service/src/Http/HttpMessageHelper.cs
drwill-ms: Until you finish migration, then it can go away?

iothub/service/src/Http/HttpMessageHelper.cs
drwill-ms: A comment would help too. Remind me what `onlyIfUnchanged` means.

iothub/service/src/Microsoft.Azure.Devices.csproj
drwill-ms: Not exposed to the user, though?

drwill-ms: Might just copy that code.

iothub/service/src/Http/HttpMessageHelper.cs
timtay-microsoft: Correct

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -91,36 +92,60 @@ internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage re
...
 
-                if (!eTag.EndsWith("\"", StringComparison.OrdinalIgnoreCase))
-                {
-                    eTag += "\"";
-                }
+        // ETag values other than "*" need to be wrapped in escaped quotes if they are not
+        // already.
+        private static string EscapeETag(string eTag)
+        {
+            StringBuilder escapedETagBuilder = new StringBuilder();
```
drwill-ms, line 134:
```suggestion
            var escapedETagBuilder = new StringBuilder();
```

iothub/service/src/Utilities/NewtonsoftJsonETagConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    /// <summary>
+    /// A <see cref="JsonConverter"/> implementation for <see cref="ETag"/>.
+    /// </summary>
+    internal class NewtonsoftJsonETagConverter : JsonConverter
+    {
+        /// <inheritdoc/>
+        public override bool CanConvert(Type objectType) => objectType == typeof(ETag) || objectType == typeof(ETag?);
+
+#nullable enable
```
drwill-ms, line 18: We don't use nullable in our projects, so you can remove these directives.

timtay-microsoft: Removing these lines breaks the implementation here though. This method is an override of a method that returns a ```object?``` type.

PR #2649, opened 2022-08-31
Move model classes into Model folders

timtay-microsoft:
Also remove some unused classes (ETagHolder, IEtagHolder, DeviceJob classes)


iothub/service/src/Configurations/ConfigurationsClient.cs
```diff
@@ -288,9 +288,9 @@ public virtual async Task DeleteAsync(string configurationId, CancellationToken
                 cancellationToken.ThrowIfCancellationRequested();
 
                 // use wild-card ETag
-                var eTag = new ETagHolder { ETag = "*" };
+                var eTag = "*";
```
patilsnr, line 291: Should we still use var if not declaring an explicit type? Why not use string? Just curious.

patilsnr (approved):

brycewang-microsoft (approved):

andyk-ms (approved):

iothub/service/src/Configurations/ConfigurationsClient.cs
timtay-microsoft: My understanding is that we are allowed and encouraged to use var when making an assignment where it is immediately obvious what type is being assigned to it. For instance:

```csharp
var myClass = new MyClass(); // okay because the assignment is obviously an instance of MyClass

var myClass = client.PerformOperation(); // not okay because it is not obvious what "PerformOperation" returns
```

In this case, although we aren't explicitly saying ```new string()```, it is still obvious that a string is being assigned, so it is okay to use var here

PR #2653, opened 2022-08-31
Remove TimeoutException from hub device client

brycewang-microsoft:
Also update some unit tests to check the status code when IotHubClientException is thrown


iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -183,7 +183,7 @@ public async Task<AmqpIotConnection> EnsureConnectionAsync(CancellationToken can
             }
             catch (OperationCanceledException)
             {
-                throw new TimeoutException();
+                throw new IotHubClientException(true, IotHubStatusCode.NetworkErrors);
```
timtay-microsoft, line 186: If the operation was cancelled, why would we label this as a network error?

timtay-microsoft: For this particular case, the service may just be very busy and is too slow to respond to the connect request even if the underlying network is fine

brycewang-microsoft: I am using `IotHubStatusCode.NetworkErrors` to replace `IotHubCommunicationException` (which we had before) which was thrown when an attempt to communicate with the IoT hub service fails due to transient network errors after exhausting all the retries based on the retry policy set on the client or due to operation timeouts. As the original exception thrown here was `TimeoutException`, I think `NetworkErrors` fits the place here

timtay-microsoft: > I am using `IotHubStatusCode.NetworkErrors` to replace `IotHubCommunicationException` (which we had [before](https://github.com/Azure/azure-iot-sdk-csharp/blob/fe800fcbbadb627db689e4d01566f03bde1ae9ba/iothub/device/src/Exceptions/IotHubCommunicationException.cs)) which was thrown when an attempt to communicate with the IoT hub service fails due to transient network errors after exhausting all the retries based on the retry policy set on the client or due to operation timeouts. 

I'm fine with this, sure

> As the original exception thrown here was `TimeoutException`, I think `NetworkErrors` fits the place here

This doesn't follow from the above to me. I'd rather we have an IotHubStatusCode for ```Timeout``` for cases like these where the SDK has no way of knowing if the request failed because of network issues or because the service was too slow

brycewang-microsoft: I considered about using a new `IotHubStatusCode.Timeout` before, but then I thought `NetworkErrors` would cover the case of operation timeouts.
Additional question for handling the case of " service was too slow": we do have `IotHubStatusCode.ServerBusy` now, if using a new code `Timeout` here, will it be sort of overlapping?

timtay-microsoft: The point of a ```Timeout``` enum value would be to signal that the SDK does not know if it was caused by network issues or by service being slow

> Additional question for handling the case of " service was too slow": we do have IotHubStatusCode.ServerBusy now, if using a new code Timeout here, will it be sort of overlapping?

We only use ```ServerBusy``` when the service explicitly notifies the SDK that it is busy. Timeout would be for cases when the SDK doesn't receive any sort of notification from the service

timtay-microsoft: Or, to put it another way, if a user sees the SDK emitting ```NetworkErrors``` reasons for failure when the customer has validated that their network to the hub is fine, they will likely file a bug.

brycewang-microsoft: that's fair!

iothub/device/src/Exceptions/IotHubStatusCode.cs
```diff
@@ -62,6 +62,12 @@ public enum IotHubStatusCode
         /// </remark>
         Suspended,
 
+        /// <summary>
+        /// The request failed because the operation times out which can be caused by underlying network issues
```
timtay-microsoft, line 66:
```csharp
        /// The request failed because the operation timed. This can be caused by underlying network issues 
        /// or by the server being too busy to handle the request.
```

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -413,8 +413,7 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
                 }
                 else
                 {
-                    // Timeout happen
-                    throw new TimeoutException();
+                    throw new IotHubClientException("Timeout happen before task completes.", true, IotHubStatusCode.Timeout);
```
timtay-microsoft, line 416: This error message doesn't really explain much. Did the client time out waiting for the service to send the twin response here?

iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs
```diff
@@ -363,14 +363,16 @@ public async Task RetrySetRetryPolicyVerifyInternalsSuccess()
...
+            exception.StatusCode.Should().Be(IotHubStatusCode.NetworkErrors);
             nextHandlerCallCounter.Should().Be(2);
             retryPolicy.Counter.Should().Be(2);
 
             var noretry = new NoRetry();
             sut.SetRetryPolicy(noretry);
-            await sut.OpenAsync(CancellationToken.None).ExpectedAsync<IotHubClientException>().ConfigureAwait(false);
 
+            exception = await sut.OpenAsync(CancellationToken.None).ExpectedAsync<IotHubClientException>().ConfigureAwait(false);
+            exception.StatusCode.Should().Be(IotHubStatusCode.NetworkErrors);
```
timtay-microsoft, line 375: Should the expected error code be ```Timeout``` now?

iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs
```diff
@@ -444,6 +446,7 @@ public bool ShouldRetry(int currentRetryCount, Exception lastException, out Time
             {
                 Counter++;
                 lastException.Should().BeOfType(typeof(IotHubClientException));
+                ((IotHubClientException)lastException).StatusCode.Should().Be(IotHubStatusCode.NetworkErrors);
```
timtay-microsoft, line 449: Same question here

timtay-microsoft (approved):

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
brycewang-microsoft: Yes, it should be. I will explain with more details in the error message.

iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs
brycewang-microsoft: No. This test is intended to throw `IotHubClientException` with `NetworkErrors` when opening the RetryDelegatingHandler to mimic the transient network errors, so `NetworkErrors` is expected.

iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs
brycewang-microsoft: No, similar answer as above. `NetworkErrors` is expected.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -413,8 +413,7 @@ public override async Task SendTwinPatchAsync(TwinCollection reportedProperties,
                 }
```
drwill-ms, line 413: In line 401 we have a default timeout being used. Can we remove that and rely simply on the cancellation token provided?

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -134,7 +137,10 @@ internal async Task<AmqpIotSession> EnsureSessionIsOpenAsync(CancellationToken c
             }
             catch (OperationCanceledException)
             {
-                throw new TimeoutException("Failed to enter the semaphore required for opening an AMQP session.");
+                throw new IotHubClientException(
```
drwill-ms, line 140: Why remap the exception? Seems we should let OperationCanceledException bubble up to the user.

iothub/device/src/Transport/ClientWebSocketTransport.cs
```diff
@@ -153,7 +154,7 @@ private async Task<int> ReadImplAsync(TransportAsyncCallbackArgs args)
             }
             catch (TaskCanceledException taskCanceledException)
```
drwill-ms, line 155: Same as above. Why remap?

iothub/device/src/Transport/Amqp/AmqpUnit.cs
brycewang-microsoft: Yes, I blindly converted the original `TimeoutException` into `IotHubClientException` with new status Timeout here. I will throw `OperationCanceledException` as-is instead.

PR #2656, opened 2022-09-01
Rename/reformat device-side direct method models to eliminate duplicates

timtay-microsoft:
```MethodInvokeRequest``` + ```MethodRequest``` + ```MethodRequestInternal``` are collapsed into just ```DirectMethodRequest```

```MethodInvokeResponse``` + ```MethodResponse``` + ```MethodResponseInternal``` are collapsed into just ```DirectMethodResponse```

Also delete IotHubSuspendedException and FatalException because they are unused.

Note that we should still give a closer look at what types we take and give the payload in, but this PR isn't about that, so it will need to be addressed later.

### CODE REVIEW

File: iothub/device/src/DirectMethod/DirectMethodRequest.cs [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+        /// <summary>
+        /// Connection timeout, in seconds.
+        /// </summary>
+        [JsonProperty("connectTimeoutInSeconds", NullValueHandling = NullValueHandling.Ignore)]
+        public int? ConnectionTimeoutInSeconds => ConnectionTimeout.HasValue && ConnectionTimeout > TimeSpan.Zero
+            ? (int)ConnectionTimeout.Value.TotalSeconds
+            : (int?)null;
+
+        [JsonProperty("payload", NullValueHandling = NullValueHandling.Include)]
+        internal JRaw JsonPayload { get; set; }
```
brycewang-microsoft, line 57: How about moving this further down so that we have all internal properties together?

iothub/device/src/Fx/Fx.cs
```diff
@@ -40,20 +40,12 @@ public static void AssertAndThrow(bool condition, string description)
...
-        }
-
         public static bool IsFatal(Exception ex)
         {
             while (ex != null)
             {
                 // FYI, CallbackException is-a FatalException
-                if (ex is FatalException
-                    || ex is OutOfMemoryException
+                if (ex is OutOfMemoryException
```
brycewang-microsoft, line 48: How about just one line as it is not too long here?

iothub/device/tests/Transport/Amqp/MockableAmqpUnit.cs
```diff
@@ -1,4 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Copyright (c) Microsoft. All rights reserved.
```
brycewang-microsoft, line 3: Extra

brycewang-microsoft (approved):
Glad to see we simplify this!

iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Parameters to execute a direct method on a device or module.
+    /// </summary>
+    public class DirectMethodRequest
+    {
+        /// <summary>
+        /// Initialize and instance of this class.
```
drwill-ms, line 16: **an** instance

iothub/device/src/DirectMethod/DirectMethodResponse.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <summary>
+        /// The optional direct method payload.
+        /// </summary>
+        [JsonIgnore]
+        public byte[] Payload { get; set; }
+
+        /// <summary>
+        /// The request Id for the transport layer.
+        /// </summary>
+        internal string RequestId { get; set; }
```
drwill-ms, line 35: No JsonProperty?

timtay-microsoft: This field is set in the topic string over MQTT or in properties over AMQP, so it isn't part of the serialized payload

drwill-ms: Could you add JsonIgnore attribute and include a code comment to this effect?

timtay-microsoft: Sure thing

PR #2658, opened 2022-09-01
Removed unused HTTP stack from Device Client

tmahmood-microsoft:


timtay-microsoft:
```CustomHeaderConstants.cs``` in that HTTP folder appears to be only used in one spot. Could you move the one referenced constant into ```HttpTransportHandler```?

**Comment by timtay-microsoft** [comment]
In ```HttpClientHelper``` there are several overloads of PostAsync and ExecuteAsync that just add on a parameter that can be simplified here. Ideally there would only be one ```PostAsync``` method and no ```ExecuteAsync``` method since we only ever post from this layer now.

**Review by andyk-ms — APPROVED** [review]

File: iothub/device/src/Transport/Http/HttpClientHelper.cs [hunk 1]
```diff
@@ -300,99 +147,20 @@ private static void InsertEtag(HttpRequestMessage requestMessage, IETagHolder en
             return result;
         }
 
-        private Task PostAsyncHelper<T1>(
+        private async Task PostAsyncHelper<T1>(
```
timtay-microsoft, line 150: If ```PostAsync``` is a one liner that just calls this helper, why not delete the helper and move its contents into the ```PostAsync``` method?

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -300,99 +147,20 @@ private static void InsertEtag(HttpRequestMessage requestMessage, IETagHolder en
...
             Uri requestUri,
             T1 entity,
             IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> errorMappingOverrides,
```
timtay-microsoft, line 153: The callers of this method always pass in the same default error mapping for this parameter, so you don't need this to be a parameter anymore

iothub/device/src/Transport/Http/HttpTransportHandler.cs
```diff
@@ -365,7 +92,7 @@ await _httpClientHelper
             _ = GetInvokeDeviceMethodOperationTimeout(methodInvokeRequest);
```
timtay-microsoft, line 92: This appears to be the only place ```GetInvokeDeviceMethodOperationTimeout``` is called, and its result is immediately discarded. It should be safe to remove this method

timtay-microsoft: Unresolving this because I still see ```GetInvokeDeviceMethodOperationTimeout``` in ```HttpTransportHandler```

timtay-microsoft:
```MergeErrorMapping``` appears to be pointless now that we aren't passing in any error mapping overrides. The function seems to exist to give the union between the dictionary of default mappings and the dictionary of override mappings. Now that there are no overrides, we don't need this function

**Comment by timtay-microsoft** [comment]
The ```modifyRequestMessageAsync``` locally defined function in ```HttpClientHelper``` also seems odd. It is a defined function that is immediately called after it is defined and it isn't used anywhere else.

```csharp
Func<HttpRequestMessage, CancellationToken, Task> modifyRequestMessageAsync = (requestMsg, token) =>
{
    AddCustomHeaders(requestMsg, customHeaders);
    if (entity != null)
    {
        if (typeof(T1) == typeof(byte[]))
        {
            requestMsg.Content = new ByteArrayContent((byte[])(object)entity);
        }
        else if (typeof(T1) == typeof(string))
        {
            // only used to send batched messages on Http runtime
            requestMsg.Content = new StringContent((string)(object)entity);
            requestMsg.Content.Headers.ContentType = new MediaTypeHeaderValue(CommonConstants.BatchedMessageContentType);
        }
        else
        {
            requestMsg.Content = CreateContent(entity);
        }
    }

    return Task.CompletedTask;
};

if (modifyRequestMessageAsync != null)
{
    await modifyRequestMessageAsync(msg, cancellationToken).ConfigureAwait(false);
}
```

It could probably be simplified to 

```csharp
    AddCustomHeaders(msg, customHeaders);
    if (entity != null)
    {
        if (typeof(T1) == typeof(byte[]))
        {
            msg.Content = new ByteArrayContent((byte[])(object)entity);
        }
        else if (typeof(T1) == typeof(string))
        {
            // only used to send batched messages on Http runtime
            msg.Content = new StringContent((string)(object)entity);
            msg.Content.Headers.ContentType = new MediaTypeHeaderValue(CommonConstants.BatchedMessageContentType);
        }
        else
        {
            msg.Content = CreateContent(entity);
        }
    }
```

timtay-microsoft:
```preRequestActionForAllRequests``` parameter for the ```HttpClientHelper``` class appears to always be null, so that parameter can be removed

timtay-microsoft:
```AddCustomHeaders``` in HttpClientHelper can be made sync since there is no async call in it

timtay-microsoft (approved):

PR #2661, opened 2022-09-02
Copy the default http settings from IoT hub service client to other clients

timtay-microsoft:
Connection lease timeout and max connections per server are defaults we set in the IoT hub service client to better handle failover scenarios, but the other client libraries that use HTTP would benefit from having them, too

### DIFF
--- a/iothub/device/src/Transport/Http/HttpClientHelper.cs
+++ b/iothub/device/src/Transport/Http/HttpClientHelper.cs
@@ -28,6 +28,22 @@ namespace Microsoft.Azure.Devices.Client.Transport
         private HttpClientHandler _httpClientHandler;
         private readonly AdditionalClientInformation _additionalClientInformation;
 
+        // These default values are consistent with Azure.Core default values:
+        // https://github.com/Azure/azure-sdk-for-net/blob/7e3cf643977591e9041f4c628fd4d28237398e0b/sdk/core/Azure.Core/src/Pipeline/ServicePointHelpers.cs#L28
+        private const int DefaultMaxConnectionsPerServer = 50;
+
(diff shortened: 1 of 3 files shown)


iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -59,6 +75,10 @@ internal sealed class HttpClientHelper
                 _httpClientHandler.Proxy = proxy;
             }
 
+            _httpClientHandler.MaxConnectionsPerServer = DefaultMaxConnectionsPerServer;
+            ServicePoint servicePoint = ServicePointManager.FindServicePoint(_baseAddress);
+            servicePoint.ConnectionLeaseTimeout = DefaultConnectionLeaseTimeout.Milliseconds;
```
abhipsaMisra, line 80: Do you see any value in us making this configurable?

abhipsaMisra (approved):

iothub/device/src/Transport/Http/HttpClientHelper.cs
timtay-microsoft: I'd rather we just allow users to pass in the entire HttpClient if they want to configure settings at this level

abhipsaMisra: That's fair. Remind me, do we configure the user-supplied http client in any way, or use it as-is?

timtay-microsoft: We use it as-is. However, we currently only allow users to pass in a HttpClient in our IoT hub service client, so we'll need to do that work in all the other clients still.

andyk-ms (approved):

PR #2662, opened 2022-09-02
Add synchronization logic to open/close amqp operations in service client

timtay-microsoft:
Also add null checks to allow close to be called even when already closed

### DIFF
--- a/iothub/service/src/Amqp/AmqpConnectionHandler.cs
+++ b/iothub/service/src/Amqp/AmqpConnectionHandler.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Amqp
     /// <remarks>
     /// This class intentionally abstracts away details about sessions and links for simplicity at the service client level.
     /// </remarks>
-    internal class AmqpConnectionHandler
+    internal class AmqpConnectionHandler : IDisposable
     {
         private AmqpConnection _connection;
         private AmqpCbsSessionHandler _cbsSession;
@@ -39,6 +39,9 @@ namespace Microsoft.Azure.Devices.Amqp
 
         // The current delivery tag. Increments after each send operation to give a unique value.
         private int _sendingDeliveryTag;
+
+        // The lock that prevents simultaneous open/close, open/open, and close/close operations
+        private SemaphoreSlim _openCloseSemaphore = new SemaphoreSlim(1, 1);
 
         public AmqpConnectionHandler(
             IotHubConnectionProperties credential,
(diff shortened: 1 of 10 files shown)


iothub/service/src/Feedback/MessageFeedbackProcessorClient.cs
```diff
@@ -116,11 +115,6 @@ public virtual async Task OpenAsync(CancellationToken cancellationToken = defaul
                     throw new Exception("Callback for message feedback must be set before opening the connection.");
                 }
 
-                if (_amqpConnection.IsOpen())
```
timtay-microsoft, line 119: Moving this check into the AmqpConnectionHandler instead since all 3 subclients call into it and all 3 subclients need this check

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -40,6 +40,9 @@ internal class AmqpConnectionHandler
         // The current delivery tag. Increments after each send operation to give a unique value.
         private int _sendingDeliveryTag;
 
+        // The lock that prevents simultaneous open/close, open/open, and close/close operations
+        private SemaphoreSlim _openCloseSemaphore = new SemaphoreSlim(1, 1);
```
timtay-microsoft, line 44: It's a bit regrettable that this is the sole reason why we need these amqp subclients to be disposable. Is there some reason why we don't tend to use the non-disposable [semaphore](https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphore?view=net-6.0) class for cases like these?

abhipsaMisra: The general guidance is generally that we should use SemaphoreSlim when we want to use local semaphores. For semaphores that are used across processes (system-wide) we'd want to use Semaphore. As for SemaphoreSlim being disposable and Semaphore not being disposable, that's a good point that I hadn't considered before. Something worth looking into for sure.

abhipsaMisra (approved):
We'll also need to update our tests to ensure these clients are now getting disposed.

timtay-microsoft:
> We'll also need to update our tests to ensure these clients are now getting disposed.

The e2e tests should be good here since IotHubServiceClient was already disposable, but I'll check for unit tests

iothub/service/src/Amqp/AmqpConnectionHandler.cs
timtay-microsoft: Apparently ```Semaphore``` doesn't have cancellation token support, so I suppose we'll need to stick with ```SemaphoreSlim```

andyk-ms (approved):

iothub/service/src/Amqp/AmqpConnectionHandler.cs
drwill-ms: Semaphore is IDisposable (via inheritance).

PR #2670, opened 2022-09-06
Reorder device and module client operations

abhipsaMisra:
This PR doesn't make any changes other than reordering the client APIs. This is in preparation for the change where we update the device and module clients to inherit from InternalClient, instead of being composed of the InternalClient (as they currently are).

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -4,11 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client.Exceptions;
-using Microsoft.Azure.Devices.Client.Extensions;
 using Microsoft.Azure.Devices.Client.Transport;
 using Microsoft.Azure.Devices.Client.Utilities;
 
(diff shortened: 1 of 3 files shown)


timtay-microsoft (approved):

iothub/device/src/InternalClient.cs
```diff
@@ -474,65 +371,120 @@ public async Task UpdateReportedPropertiesAsync(TwinCollection reportedPropertie
...
-        /// <param name="message">The message to complete</param>
-        /// <param name="cancellationToken">A token to cancel the operation. </param>
-        /// <returns>The previously received message</returns>
-        public async Task CompleteMessageAsync(Message message, CancellationToken cancellationToken = default)
+        /// <param name="methodHandler">The delegate to be used when a method is called by the cloud service and there is no
+        /// delegate registered for that method name.</param>
+        /// <param name="userContext">Generic parameter to be interpreted by the client code.</param>
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice
+        /// of cancellation.</param>
+        public async Task SetMethodDefaultHandlerAsync(
```
drwill-ms, line 434: I had reordered these previously and aimed to put them in the rough order I thought someone would need to call them. IOW, setting handlers would come before telemetry or completing C2D messages etc.

You have a different order in mind. Can you elaborate on that?

abhipsaMisra: Ah, I see your point. The order I used was simply based on the ordering of operations that I see in our docs (which I feel could also corelate to increasing order of complexity of operations) - send event, receive event, direct method, twins. I can see the value in placing the callback APIs before the "sending" APIs, but I'd also like to have the same or similar operations grouped together, i.e. all twin operations grouped together. How do you feel about that? 
Also, with the inheritance model, since we'll be removing a bunch of APIs from either the InternalClient or (Device/Module)Client we can do the re-ordering again. The idea here was to make understanding the next PR's diff a bit better.

PR #2672, opened 2022-09-07
refactor(iot-service): throw a more general IotHubServiceException with the property StatusCode and ErrorCode

brycewang-microsoft:
Like what we did for the hub **device** client before, we stop throwing particular exceptions ([ref](https://github.com/Azure/azure-iot-sdk-csharp/tree/main/iothub/service/src/Common/Exceptions)), but a more general exception called `IotHubServiceException` with the property "StatusCode" and "ErrorCode" instead. 

The property "StatusCode" is a 3-digit HTTP status code returned in the response, and "ErrorCode" is a 6-digit code representing a more specific error with details (usually returned in the response content).

A usage sample is following:

//v1
```c#
try
{
    await serviceClient.DirectMethods.InvokeAsync(invalidDeviceName, directMethodRequest).ConfigureAwait(false);
}
catch (DeviceNotFoundException ex) { }
```

//v2
```c#
try
{
    await serviceClient.DirectMethods.InvokeAsync(invalidDeviceName, directMethodRequest).ConfigureAwait(false);
}
catch (IotHubServiceException ex) when (ex.StatusCode is HttpStatusCode.NotFound && ex.ErrorCode is IotHubErrorCode.DeviceNotFound) { }
```


e2e/test/helpers/RetryOperationHelper.cs
```diff
@@ -51,5 +52,43 @@ public static async Task RetryOperationsAsync(Func<Task> asyncOperation, IRetryP
...
+
+        /// <summary>
+        /// Rety an async operation based on the retry strategy supplied.
+        /// </summary>
+        /// <param name="asyncOperation">The async operation to be retried.</param>
+        /// <param name="retryPolicy">The retry policy to be applied.</param>
+        /// <param name="retryableStatusCodes">The errors to be retried on.</param>
+        /// <param name="logger">The <see cref="MsTestLogger"/> instance to be used.</param>
+        /// <returns></returns>
+        public static async Task RetryOperationsAsync1(Func<Task> asyncOperation, IRetryPolicy retryPolicy, HashSet<IotHubStatusCode> retryableStatusCodes, MsTestLogger logger)
```
brycewang-microsoft, line 64: As we don't have particular exceptions thrown any more, but "IotHubServiceException" with "StatusCode", updating this method accordingly and renaming it to `RetryOperationsAsync1()` for now, because the original `RetryOperationsAsync()` is still used by DPS clients E2E tests.

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -313,7 +310,7 @@ await deviceClient
                     .InvokeAsync(deviceId, directMethodRequest)
                     .ConfigureAwait(false);
             }
-            catch (DeviceNotFoundException)
```
brycewang-microsoft, line 316: This was incorrect as the root cause here was unsubscribing the direct method from the device, so [DeviceNotOnline ](https://docs.microsoft.com/en-us/azure/iot-hub/troubleshoot-error-codes#404103-devicenotonline)should be expected instead.

brycewang-microsoft: `DeviceNotOnline` was not existing before, so adding within this PR.

e2e/test/iothub/service/IoTHubCertificateValidationE2ETest.cs
```diff
@@ -20,7 +22,7 @@ public class IotHubCertificateValidationE2ETest : E2EMsTestBase
         public async Task ServiceClient_QueryDevicesInvalidServiceCertificateHttp_Fails()
         {
             using var sc = new IotHubServiceClient(TestConfiguration.IoTHub.ConnectionStringInvalidServiceCertificate);
-            IotHubCommunicationException exception = await Assert.ThrowsExceptionAsync<IotHubCommunicationException>(
+            var exception = await Assert.ThrowsExceptionAsync<IotHubServiceException>(
```
brycewang-microsoft, line 25: Tests here are skipped currently, so not checking ErrorCode and IotHubStatusCode properties here

e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -232,5 +235,60 @@ public async Task MessageIdDefaultSetToGuid_SendEventSetMessageIdIfNotSet()
             messageWithoutId.MessageId.Should().NotBeNullOrEmpty();
             messageWithId.MessageId.Should().Be(messageId);
         }
+
+        [LoggedTestMethod, Timeout(TestTimeoutMilliseconds)]
+        [DataRow(IotHubTransportProtocol.Tcp)]
+        [DataRow(IotHubTransportProtocol.WebSocket)]
+        public async Task MessagingClient_SendToNonexistentDevice_ThrowIotHubServiceException(IotHubTransportProtocol protocol)
```
brycewang-microsoft, line 242: Add E2E tests for validating IotHubServiceException for AMQP operations

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -130,41 +127,41 @@ public static Exception ToIotHubClientContract(Error error)
...
             }
             else if (error.Condition.Equals(IotHubAmqpErrorCode.PreconditionFailed))
             {
-                retException = new PreconditionFailedException(message, (Exception)null);
+                retException = new IotHubServiceException(HttpStatusCode.PreconditionFailed, IotHubStatusCode.PreconditionFailed, message, (Exception)null);
             }
             else if (error.Condition.Equals(IotHubAmqpErrorCode.IotHubSuspended))
             {
-                retException = new IotHubSuspendedException(message);
+                retException = new IotHubServiceException(IotHubStatusCode.HubSuspended, message);
```
brycewang-microsoft, line 154: to-do: check the error code returned by the hub service while sending requests to a suspended hub.

brycewang-microsoft: ideally, we should pass the 3digit ErrorCode property into IotHubServiceException ctor

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,102 +17,30 @@ internal class ExceptionHandlingHelper
         private const string MessageFieldErrorCode = "errorCode";
         private const string HttpErrorCodeName = "iothub-errorcode";
 
-        private static readonly IReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> s_mappings =
```
brycewang-microsoft, line 21: We don't have to map the returned 3-digit http status code in response to IotHubStatusCode here, so delete

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -149,7 +76,16 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
                                 if (messageField.IndexOf(errorCodeDelimiter) >= 0)
                                 {
                                     string[] errorCodeFields = messageField.Split(errorCodeDelimiter);
-                                    if (Enum.TryParse(errorCodeFields[1], out ErrorCode errorCode))
+
+                                    string returnedErrorCode = errorCodeFields[1];
+
+                                    // When the returned error iotHubStatusCode is numeric, only taing the first 6 characters as the numeric iotHubStatusCode contains 6 digits.
+                                    if (int.TryParse(returnedErrorCode.Substring(0, 6), out int code))
```
brycewang-microsoft, line 83: fix the logic of getting 6-digit status code from the response content

iothub/service/src/Exceptions/IotHubStatusCode.cs
```diff
@@ -123,6 +118,17 @@ public enum ErrorCode
...
+        /// <summary>
+        /// The operation failed because the requested device isn't online or hasn't registered the direct method callback.
+        /// </summary>
+        /// <para>
+        /// May be thrown by operations such as <see cref="DirectMethodsClient.InvokeAsync(string, DirectMethodRequest, System.Threading.CancellationToken)"/>
+        /// </para>
+        /// <remarks>
+        /// For more information, see <see href="https://docs.microsoft.com/en-us/azure/iot-hub/troubleshoot-error-codes#404103-devicenotonline"/>.
+        /// </remarks>
+        DeviceNotOnline = 404103,
```
brycewang-microsoft, line 130: newly add enum `DeviceNotOnline`

iothub/service/src/Exceptions/IotHubStatusCode.cs
```diff
@@ -178,14 +184,6 @@ public enum ErrorCode
         /// </summary>
         ThrottlingException = 429001,
 
-        /// <summary>
-        /// IoT hub throttling limits have been exceeded for the requested operation.
-        /// <para>
-        /// For more information, see <see href="https://aka.ms/iothubthrottling">IoT hub quotas and throttling</see>.
-        /// </para>
-        /// </summary>
-        ThrottleBacklogLimitExceeded = 429002,
```
brycewang-microsoft, line 187: unused

iothub/service/src/Exceptions/IotHubStatusCode.cs
```diff
@@ -204,5 +202,13 @@ public enum ErrorCode
...
         /// </summary>
         ServiceUnavailable = 503001,
+
+        /// <summary>
+        /// The operation failed because the IoT hub has been suspended. 
+        /// </summary>
+        /// <remarks>
+        /// This is likely due to exceeding Azure spending limits. To resolve the error, check the Azure bill and ensure there are enough credits.
+        /// </remarks>
+        HubSuspended, // to-do: brycewang - check the error code returned by the hub service while sending requests to a suspended hub.
```
brycewang-microsoft, line 212: to-do: I am checking this with the service team for a 6-digit code returned back in the response content

e2e/test/helpers/RetryOperationHelper.cs
```diff
@@ -51,5 +52,43 @@ public static async Task RetryOperationsAsync(Func<Task> asyncOperation, IRetryP
...
         }
+
+        /// <summary>
+        /// Rety an async operation based on the retry strategy supplied.
+        /// </summary>
+        /// <param name="asyncOperation">The async operation to be retried.</param>
+        /// <param name="retryPolicy">The retry policy to be applied.</param>
+        /// <param name="retryableStatusCodes">The errors to be retried on.</param>
+        /// <param name="logger">The <see cref="MsTestLogger"/> instance to be used.</param>
+        /// <returns></returns>
```
drwill-ms, line 63: Fill in doc comment or remove if it is not interesting (as is the case with `Task`).

e2e/test/helpers/RetryOperationHelper.cs
drwill-ms, line 64: Why a suffix of 1?

e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -64,7 +65,7 @@ public async Task RegistryManager_Http_SasCredentialAuth_Renewed_Success()
                 await serviceClient.Devices.CreateAsync(device).ConfigureAwait(false);
                 Assert.Fail("The SAS token is expired so the call should fail with an exception");
             }
-            catch (UnauthorizedException)
+            catch (IotHubServiceException ex) when (ex.ErrorCode is System.Net.HttpStatusCode.Unauthorized)
```
drwill-ms, line 68: Move System.Net to usings

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -142,18 +143,22 @@ public async Task Method_ServiceInvokeDeviceMethodWithUnknownDeviceThrows()
...
             {
-                actualErrorCode = ex.Code;
+                errorCode = ex.ErrorCode;
+                iotHubStatusCode = ex.IotHubStatusCode;
             }
 
-            Assert.AreEqual(ErrorCode.DeviceNotFound, actualErrorCode);
+            // assert
+            Assert.AreEqual(HttpStatusCode.NotFound, errorCode);
+            Assert.AreEqual(IotHubStatusCode.DeviceNotFound, iotHubStatusCode);
```
drwill-ms, line 161: Please take the time to change these to fluent assertions:

```suggestion
            errorCode.Should().Be(HttpStatusCode.NotFound);
            iotHubStatusCode.Should().Be(IotHubStatusCode.DeviceNotFound);
```

e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -232,5 +235,60 @@ public async Task MessageIdDefaultSetToGuid_SendEventSetMessageIdIfNotSet()
...
+        {
+            // arrange
+            IotHubServiceClientOptions options = new IotHubServiceClientOptions
```
drwill-ms, line 245:
```suggestion
            var options = new IotHubServiceClientOptions
```

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -125,7 +52,7 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
                     if (messageFields != null
                         && messageFields.TryGetValue(MessageFieldErrorCode, out string errorCodeObj))
                     {
-                        // The result of TryParse is not being tracked since errorCodeValue has already been initialized to a default value of InvalidErrorCode.
+                        // The result of TryParse is not being tracked since errorCodeValue has already been initialized to a default value of Unknown.
                         _ = int.TryParse(errorCodeObj, NumberStyles.Any, CultureInfo.InvariantCulture, out errorCodeValue);
```
drwill-ms, line 56: It doesn't matter if it was initialized before this call because any `out` parameter must be set by the callee.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -149,7 +76,16 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
                                 if (messageField.IndexOf(errorCodeDelimiter) >= 0)
                                 {
                                     string[] errorCodeFields = messageField.Split(errorCodeDelimiter);
-                                    if (Enum.TryParse(errorCodeFields[1], out ErrorCode errorCode))
+
+                                    string returnedErrorCode = errorCodeFields[1];
+
+                                    // When the returned error iotHubStatusCode is numeric, only taing the first 6 characters as the numeric iotHubStatusCode contains 6 digits.
```
drwill-ms, line 82:
```suggestion
                                    // When the returned error iotHubStatusCode is numeric, only take the first 6 characters as the numeric iotHubStatusCode contains 6 digits.
```

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,102 +17,30 @@ internal class ExceptionHandlingHelper
...
 
         /// <summary>
-        /// Get the fully-qualified error code from the HTTP response message, if exists.
+        /// Get the fully-qualified error iotHubStatusCode from the HTTP response message, if exists.
         /// </summary>
         /// <param name="response">The HTTP response message</param>
-        /// <returns>The fully-qualified error code, or the response status code, if no error code was provided.</returns>
-        public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage response)
+        /// <returns>The fully-qualified error iotHubStatusCode, or the response status iotHubStatusCode, if no error iotHubStatusCode was provided.</returns>
+        public static async Task<IotHubStatusCode> GetExceptionCodeAsync(HttpResponseMessage response)
```
drwill-ms, line 30: Do we have unit tests for this?

Might need to have one method get the response content and call another one that processes it. It'll be easier to unit test a method that doesn't require passing a response.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,102 +17,30 @@ internal class ExceptionHandlingHelper
...
         }
 
         /// <summary>
-        /// Get the fully-qualified error code from the HTTP response message, if exists.
+        /// Get the fully-qualified error iotHubStatusCode from the HTTP response message, if exists.
         /// </summary>
         /// <param name="response">The HTTP response message</param>
-        /// <returns>The fully-qualified error code, or the response status code, if no error code was provided.</returns>
-        public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage response)
+        /// <returns>The fully-qualified error iotHubStatusCode, or the response status iotHubStatusCode, if no error iotHubStatusCode was provided.</returns>
```
drwill-ms, line 29: Lines in this method are getting quite long. Can you make them shorter where excessively long?

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -142,18 +143,22 @@ public async Task Method_ServiceInvokeDeviceMethodWithUnknownDeviceThrows()
...
-            ErrorCode actualErrorCode = ErrorCode.InvalidErrorCode;
+            HttpStatusCode errorCode = HttpStatusCode.OK;
+            IotHubStatusCode iotHubStatusCode = IotHubStatusCode.Unknown;
             try
             {
                 // Invoke the direct method asynchronously and get the response from the simulated device.
                 await serviceClient.DirectMethods.InvokeAsync("SomeNonExistantDevice", methodInvocation);
             }
-            catch (DeviceNotFoundException ex)
+            catch (IotHubServiceException ex) when (ex.StatusCode is HttpStatusCode.NotFound)
```
andyk-ms, line 153: With "when", don't we need a catch all to cover other cases?

brycewang-microsoft: No, we intentionally catch `IotHubServiceException` with `StatusCode` is `NotFound` here.  

That's the only expected exception thrown out here, otherwise we should let it throw out unexpected exceptions.

iothub/service/src/Amqp/AmqpClientHelper.cs
brycewang-microsoft: Added "IotHubSuspended = 400020" enum to IotHubErrorCode

iothub/service/src/Exceptions/IotHubStatusCode.cs
brycewang-microsoft: Added "IotHubSuspended = 400020" enum to IotHubErrorCode. Complete!

e2e/test/helpers/RetryOperationHelper.cs
brycewang-microsoft: The original RetryOperationsAsync() is still used by DPS clients E2E tests, which should be replaced by this method after we re-work the exception thrown by DPS clients.

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -142,18 +143,22 @@ public async Task Method_ServiceInvokeDeviceMethodWithUnknownDeviceThrows()
...
             {
-                actualErrorCode = ex.Code;
+                errorCode = ex.StatusCode;
+                iotHubStatusCode = ex.ErrorCode;
             }
 
-            Assert.AreEqual(ErrorCode.DeviceNotFound, actualErrorCode);
+            // assert
+            Assert.AreEqual(HttpStatusCode.NotFound, errorCode);
+            Assert.AreEqual(IotHubErrorCode.DeviceNotFound, iotHubStatusCode);
```
drwill-ms, line 161: Fluent assertions please

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -21,11 +22,11 @@ public static Exception ToIotHubClientContract(Exception exception)
         {
             if (exception is TimeoutException)
             {
-                return new IotHubCommunicationException(exception.Message);
+                return new IotHubServiceException(HttpStatusCode.RequestTimeout, IotHubErrorCode.Unknown, exception.Message);
```
drwill-ms, line 25: Long lines

```suggestion
                return new IotHubServiceException(
                    HttpStatusCode.RequestTimeout,
                    IotHubErrorCode.Unknown,
                    exception.Message);
```

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -21,11 +22,11 @@ public static Exception ToIotHubClientContract(Exception exception)
...
             }
             else if (exception is UnauthorizedAccessException)
```
drwill-ms, line 27: Don't need else ifs as previous block returns

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -130,41 +127,41 @@ public static Exception ToIotHubClientContract(Error error)
...
             }
 
             if (trackingId != null
-                && retException is IotHubException exHub)
+                && retException is IotHubServiceException exHub)
             {
-                IotHubException iotHubException = exHub;
+                IotHubServiceException iotHubException = exHub;
                 iotHubException.TrackingId = trackingId;
                 // This is created but not assigned to `retException`. If we change that now, it might be a
```
drwill-ms, line 166: Here's the breaking change note I mentioned in Teams

These code is executing but does nothing. It should be put on the exception that is returned

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -130,41 +127,41 @@ public static Exception ToIotHubClientContract(Error error)
...
-                retException = new IotHubSuspendedException(message);
+                retException = new IotHubServiceException(HttpStatusCode.BadRequest, IotHubErrorCode.IotHubSuspended, message);
             }
             else
             {
-                retException = new IotHubException(message);
+                retException = new IotHubServiceException(message);
             }
 
             if (trackingId != null
```
drwill-ms, line 161: if (trackingId != null
    && retException is IotHubServiceException exHub)
{
     exHub.TrackingId = trackingId;
}

iothub/service/src/Exceptions/IotHubServiceException.cs
```diff
@@ -0,0 +1,128 @@
...
+            : base(message, innerException)
+        {
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="IotHubServiceException"/> with the supplied error message and tracking Id, and marks it as non-transient.
+        /// </summary>
+        /// <param name="message">The message that describes the error.</param>
+        /// <param name="trackingId">The service returned tracking Id associated with this particular error.</param>
+        public IotHubServiceException(string message, string trackingId)
```
drwill-ms, line 34: Any luck changing the calls to these extra ctors to use the basic one and an initializer list?

iothub/service/src/Exceptions/IotHubServiceException.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// The exception thrown when the client receives an error while communicating with IoT hub service.
+    /// </summary>
+    [Serializable]
+    public class IotHubServiceException : Exception
+    {
+        private const string IsTransientValueSerializationStoreName = "IotHubServiceException-IsTransient";
+        private const string TrackingIdSerializationStoreName = "IoTHubException-TrackingId";
+
+        /// <summary>
+        /// Creates an instance of <see cref="IotHubServiceException"/> with the supplied error message and marks it as non-transient.
```
drwill-ms, line 20: No need to reference this type in doc comments.

```suggestion
        /// Creates an instance of this class with the specified error message and optional inner exception.
```

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -85,6 +76,18 @@ internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage re
             return JsonConvert.DeserializeObject<T>(str);
         }
 
+        /// <summary>
```
drwill-ms, line 79: Not a public method so documenting it is only for code readers, mainly us.

All upstream methods that are public need to have these exceptions documented, however.

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -95,7 +98,7 @@ internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage re
         /// </param>
         /// <exception cref="ArgumentNullException">Thrown when the provided <paramref name="requestMessage"/> or <paramref name="requestMessage"/> is null.</exception>
         /// <exception cref="ArgumentException">Thrown if the <paramref name="eTag"/> is empty or white space.</exception>
-        internal static void ConditionallyInsertETag(HttpRequestMessage requestMessage, ETag eTag, bool onlyIfUnchanged)
+        internal static void ConditionallyInsertETag(HttpRequestMessage requestMessage, ETag eTag, bool onlyIfUnchanged = false)
```
drwill-ms, line 101: Setting default to false intentional?

iothub/service/src/Registry/DevicesClient.cs
```diff
@@ -61,10 +61,10 @@ internal DevicesClient(string hostName, IotHubConnectionProperties credentialPro
         /// <param name="cancellationToken">The token which allows the operation to be canceled.</param>
         /// <returns>The registered device with the generated keys and ETags.</returns>
         /// <exception cref="ArgumentNullException">Thrown when the provided device is null.</exception>
-        /// <exception cref="IotHubException">
+        /// <exception cref="IotHubServiceException">
```
drwill-ms, line 64: Device client's exception shouldn't be IotHub**Service**Exception, should it?

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -99,100 +25,112 @@ public static Task<string> GetExceptionMessageAsync(HttpResponseMessage response
         /// Get the fully-qualified error code from the HTTP response message, if exists.
         /// </summary>
         /// <param name="response">The HTTP response message</param>
-        /// <returns>The fully-qualified error code, or the response status code, if no error code was provided.</returns>
-        public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage response)
+        /// <returns>
+        /// The fully-qualified error code, or one derived from the response HTTP status code if no error code was provided.
+        /// </returns>
+        internal static async Task<IotHubErrorCode> GetIotHubErrorCodeAsync(HttpResponseMessage response)
```
drwill-ms, line 31: Do you plan to change the callers to use the other 2 methods directly so we can remove this?

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -99,100 +25,112 @@ public static Task<string> GetExceptionMessageAsync(HttpResponseMessage response
...
+        // the error description. Ideally, when a backend service returns an error, both of these fields are set
+        // in the same place. However, IoT hub is returning the 6-digit code in the response content, while
+        // the error description in the response header. Therefore, there is a chance that the 6-digit error code
+        // does not match the error description. For that reason, the SDK will do its best to decide what to surface to the user.
+        // The SDK will attempt to retrieve the integer error code from the response content and the error description
+        // from the response header. Through a 'description' to 'error code' enum mapping, the SDK will check if
+        // both values are a match. If so, the SDK will populate the exception with the proper Code. In the case where
+        // there is a mismatch between the error code and the description, the SDK returns
+        // IotHubStatusCode.Unknown and log a warning.
+        internal static IotHubErrorCode GetIotHubErrorCode(string responseBody)
```
drwill-ms, line 48: Still want to see more unit tests for these, especially the positive cases.

iothub/service/src/Http/HttpMessageHelper.cs
brycewang-microsoft: This is weird, I am not aware of this, so changing it back.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
brycewang-microsoft: Yes, we do have existing units tests for this, and I also add some new ones to validate cases of different response payload.

e2e/test/helpers/RetryOperationHelper.cs
```diff
@@ -51,5 +52,42 @@ public static async Task RetryOperationsAsync(Func<Task> asyncOperation, IRetryP
...
         }
+
+        /// <summary>
+        /// Rety an async operation based on the retry strategy supplied.
+        /// </summary>
+        /// <param name="asyncOperation">The async operation to be retried.</param>
+        /// <param name="retryPolicy">The retry policy to be applied.</param>
+        /// <param name="retryableStatusCodes">The errors to be retried on.</param>
+        /// <param name="logger">The <see cref="MsTestLogger"/> instance to be used.</param>
+        public static async Task RetryOperationsAsync1(Func<Task> asyncOperation, IRetryPolicy retryPolicy, HashSet<IotHubErrorCode> retryableStatusCodes, MsTestLogger logger)
```
drwill-ms, line 63: Can a user's cancellation token be passed along and checked throughout retries?

e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -232,5 +235,60 @@ public async Task MessageIdDefaultSetToGuid_SendEventSetMessageIdIfNotSet()
...
+            // act
+            var message = new Message(new byte[10]); // arbitrary payload since it shouldn't matter
+            Func<Task> act = async () => await sender.Messaging.SendAsync("nonexistent-device-id", message).ConfigureAwait(false);
+
+            // assert
+            var error = await act.Should().ThrowAsync<IotHubServiceException>();
+            error.And.StatusCode.Should().Be(HttpStatusCode.NotFound);
+            error.And.ErrorCode.Should().Be(IotHubErrorCode.DeviceNotFound);
+
+            await sender.Messaging.CloseAsync().ConfigureAwait(false);
```
drwill-ms, line 261: Perhaps this should be in a finally statement, because otherwise it won't run if the asserts fire.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,181 +16,131 @@ internal class ExceptionHandlingHelper
...
-                    else
-                    {
-                        if (Logging.IsEnabled)
-                            Logging.Error(null, $"Failed to deserialize error message into a dictionary and could not parse ';' delimited string either: {ex}." +
-                                $" Message body: '{responseContentStr}.'");
 
-                        return ErrorCode.InvalidErrorCode;
+                            // Otherwise the error code might be a string (e.g., PreconditionFailed) in which case we'll try to
+                            // find the matching IotHubErrorCode enum with that same name.
+                            if (Enum.TryParse(returnedErrorCode, out IotHubErrorCode errorCode))
```
drwill-ms, line 108: I feel like this code block should be outside of the message fields foreach. We should only do this if that attempt to find a best match fails.

iothub/service/tests/ExceptionHandlingHelperTests.cs
```diff
@@ -29,17 +31,36 @@ public async Task GetExceptionCodeAsync_ContentAndHeadersMatch_ValidErrorCode()
...
+
+            // act
+            IotHubErrorCode errorCode = await ExceptionHandlingHelper.GetIotHubErrorCodeAsync(httpResponseMessage);
+
+            // assert
+            errorCode.Should().Be(IotHubErrorCode.PreconditionFailed);
+        }
+
+        [TestMethod]
+        public async Task GetExceptionCodeAsync_ContentAndHeadersMisMatch_UnknownErrorCode()
```
drwill-ms, line 60: Mismatch one word, so no second capital M.

```suggestion
        public async Task GetExceptionCodeAsync_ContentAndHeadersMismatch_UnknownErrorCode()
```

iothub/service/tests/ExceptionHandlingHelperTests.cs
```diff
@@ -29,17 +31,36 @@ public async Task GetExceptionCodeAsync_ContentAndHeadersMatch_ValidErrorCode()
...
         [TestMethod]
-        public async Task GetExceptionCodeAsync_ContentAndHeadersMisMatch_InvalidErrorCode()
+        public async Task GetExceptionCodeAsync_ContentAndHeadersMatch_NonNumericErrorCode_ValidErrorCode()
         {
             // arrange
-            var httpResponseMessage = new HttpResponseMessage(System.Net.HttpStatusCode.BadRequest);
+            var httpResponseMessage = new HttpResponseMessage(HttpStatusCode.BadRequest);
+            var exceptionResult = new IoTHubExceptionResult
+            {
+                Message = "errorCode:PreconditionFailed"
```
drwill-ms, line 47: Doesn't the message often have more to it than just this? We should ensure the message parsed here is what the SDK would see real-world.

iothub/service/tests/ExceptionHandlingHelperTests.cs
```diff
@@ -48,17 +69,17 @@ public async Task GetExceptionCodeAsync_ContentAndHeadersMisMatch_InvalidErrorCo
...
         [TestMethod]
-        public async Task GetExceptionCodeAsync_NoContentErrorCode_InvalidErrorCode()
+        public async Task GetExceptionCodeAsync_NoContentErrorCode_UnknownErrorCode()
         {
             // arrange
-            var httpResponseMessage = new HttpResponseMessage(System.Net.HttpStatusCode.BadRequest);
+            var httpResponseMessage = new HttpResponseMessage(HttpStatusCode.BadRequest);
             var exceptionResult = new IoTHubExceptionResult
             {
                 Message = ""
```
drwill-ms, line 85: Should still be a string with content, just not in a format we expect so we can't parse it.

iothub/service/tests/ExceptionHandlingHelperTests.cs
```diff
@@ -67,28 +88,28 @@ public async Task GetExceptionCodeAsync_NoContentErrorCode_InvalidErrorCode()
...
             };
             httpResponseMessage.Content = new StringContent(JsonConvert.SerializeObject(exceptionResult));
 
             // act
-            ErrorCode errorCode = await ExceptionHandlingHelper.GetExceptionCodeAsync(httpResponseMessage);
+            IotHubErrorCode errorCode = await ExceptionHandlingHelper.GetIotHubErrorCodeAsync(httpResponseMessage);
 
             // assert
-            Assert.AreEqual(ErrorCode.InvalidErrorCode, errorCode);
+            errorCode.Should().Be(IotHubErrorCode.Unknown);
```
drwill-ms, line 112: Why shouldn't this one be IotHubErrorCode.DeviceNotFound?

e2e/test/helpers/RetryOperationHelper.cs
brycewang-microsoft: No, neither for the original `RetryOperationsAsync()` call.
https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/e2e/test/helpers/RetryOperationHelper.cs#L25

brycewang-microsoft: Do we wanna this happen?

drwill-ms: I would think yes, we'd want retries to stop if the user's cancellation token has been signaled.

drwill-ms: Maybe it is already being checked by the operation, but we do have code in here to wait an interval, and we'd want that to cancel sooner.

drwill-ms: Feel free to take it as a task for another PR. Just please add it to a PBI with enough context for someone to be able to solve.

brycewang-microsoft: added a task [here](https://msazure.visualstudio.com/One/_boards/board/t/IoT-Developers%20and%20Devices-Managed%20SDKs/Backlog%20items/?workitem=15520439)

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
brycewang-microsoft: This code block is for the case non-numeric error codes in the response content (i.e., "PreconditionFailed"), so I prefer to leave it as-is. If both attempts of matching numeric and non-numeric codes fail, we will simply return `IotHubErrorCode.Unknown`.

iothub/service/tests/ExceptionHandlingHelperTests.cs
brycewang-microsoft: We don't add `DeviceNotFound` into the response **header**, so it won't pass the check when comparing the error code from the response content to the one from response header ([ref](https://github.com/Azure/azure-iot-sdk-csharp/blob/0b950a257129594403c9b7e4e78a6499759262eb/iothub/service/src/Exceptions/ExceptionHandlingHelper.cs#L125)).

brycewang-microsoft: this check is actually what we deleted near the bottom of the class "ExceptionHandlingHelper" by mistake.

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -130,42 +133,41 @@ internal static Exception ToIotHubClientContract(Error error)
...
-                && retException is IotHubException exHub)
+                && retException is IotHubServiceException exHub)
             {
-                IotHubException iotHubException = exHub;
-                iotHubException.TrackingId = trackingId;
+                exHub.TrackingId = trackingId;
                 // This is created but not assigned to `retException`. If we change that now, it might be a
```
drwill-ms, line 171: Can remove this comment

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -22,20 +23,26 @@ internal static class AmqpClientHelper
...
 
-            return exception;
+                case UnauthorizedAccessException:
+                    return new IotHubServiceException(
+                        HttpStatusCode.Unauthorized,
+                        IotHubErrorCode.IotHubUnauthorizedAccess,
+                        exception.Message);
+
+                case AmqpException amqpException:
+                    return ToIotHubClientContract(amqpException.Error);
```
drwill-ms, line 41: Pass the exception, so the callee can at least make it an inner exception.

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -58,7 +65,7 @@ internal static Exception GetExceptionFromOutcome(Outcome outcome)
         {
             if (outcome == null)
             {
-                return new IotHubException("Unknown error.");
+                return new IotHubServiceException("Unknown error.");
```
drwill-ms, line 68: Needs inner exception set.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,181 +16,131 @@ internal class ExceptionHandlingHelper
...
-            // to 'error code' enum mapping, the SDK will check if both values are a match. If so, the SDK will populate the exception with the proper Code. In the case where
-            // there is a mismatch between the error code and the description, the SDK returns ErrorCode.InvalidErrorCode and log a warning.
-
-            int errorCodeValue = (int)ErrorCode.InvalidErrorCode;
+            string responseBody = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+            IoTHubExceptionResult responseContent = null;
             try
             {
-                IoTHubExceptionResult responseContent = JsonConvert.DeserializeObject<IoTHubExceptionResult>(responseContentStr);
+                responseContent = JsonConvert.DeserializeObject<IoTHubExceptionResult>(responseBody);
```
drwill-ms, line 40: We should be able to improve this type so that the various fields of this object are defined, and we can access the property we want directly, rather than having to deserialize again to a dictionary (below).

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,181 +16,131 @@ internal class ExceptionHandlingHelper
...
-                    string[] messageFields = responseContent.Message?.Split(errorFieldsDelimiter);
+            // In some scenarios, the error response string is a semicolon delimited string with the service-returned error code
+            // embedded in a string response.
+            const char errorFieldsDelimiter = ';';
+            string[] messageFields = responseContent.Message?.Split(errorFieldsDelimiter);
 
-                    if (messageFields != null)
+            if (messageFields == null && Logging.IsEnabled)
+            {
+                Logging.Error(
```
drwill-ms, line 81: Now a warning.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,181 +16,131 @@ internal class ExceptionHandlingHelper
...
+            else
+            {
+                foreach (string messageField in messageFields)
+                {
+                    if (messageField.IndexOf(MessageFieldErrorCode, StringComparison.OrdinalIgnoreCase) >= 0)
                     {
-                        foreach (string messageField in messageFields)
+                        const char errorCodeDelimiter = ':';
+
+                        if (messageField.IndexOf(errorCodeDelimiter) >= 0)
```
drwill-ms, line 94:
```suggestion
                        if (messageField.IndexOf(errorCodeDelimiter) <= 0)
                        {
                            continue;
                        }
```

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,181 +16,131 @@ internal class ExceptionHandlingHelper
...
-                        if (Logging.IsEnabled)
-                            Logging.Error(null, $"Failed to deserialize error message into a dictionary and could not parse ';' delimited string either: {ex}." +
-                                $" Message body: '{responseContentStr}.'");
 
-                        return ErrorCode.InvalidErrorCode;
+                            // Otherwise the error code might be a string (e.g., PreconditionFailed) in which case we'll try to
+                            // find the matching IotHubErrorCode enum with that same name.
+                            if (Enum.TryParse(returnedErrorCode, out IotHubErrorCode errorCode))
+                            {
+                                return CompareToResponseHeader(response, errorCode);
```
drwill-ms, line 110: Why compare?

iothub/service/tests/ExceptionHandlingHelperTests.cs
```diff
@@ -17,29 +19,57 @@ public class ExceptionHandlingHelperTests
...
             // arrange
-            var httpResponseMessage = new HttpResponseMessage(System.Net.HttpStatusCode.BadRequest);
+            var httpResponseMessage = new HttpResponseMessage(HttpStatusCode.BadRequest);
             var exceptionResult = new IoTHubExceptionResult
             {
-                Message = "{\"errorCode\":404001}"
+                // A read-world message in the response content which includes the numeric error code returned by the hub service.
+                Message = 
+                "{" +
+                    "\"errorCode\":404103," +
```
drwill-ms, line 31: This is the type we can define that I referred to above, so we can simply index in to `ErrorCode`.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
brycewang-microsoft: I just realized that we don't have Warning level logging, do we want to add it if it doesn't break anything?

drwill-ms: Meh, I guess we'll stick with Error then.

iothub/service/src/Exceptions/IotHubServiceException.cs
brycewang-microsoft: reworked the ctors here

e2e/test/iothub/service/ConfigurationsClientE2ETests.cs
```diff
@@ -139,7 +141,8 @@ public async Task ConfigurationsClient_SetETag_Works()
                     // If this fails, it won't fail the test
                     await serviceClient.Configurations.DeleteAsync(configurationId).ConfigureAwait(false);
                 }
-                catch (DeviceNotFoundException)
+                catch (IotHubServiceException ex) 
+                    when (ex.StatusCode is HttpStatusCode.NotFound && ex.ErrorCode is IotHubErrorCode.DeviceNotFound)
```
drwill-ms, line 145:
```suggestion
                    when (ex.ErrorCode is IotHubErrorCode.DeviceNotFound)
```

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,181 +16,117 @@ internal class ExceptionHandlingHelper
...
 
-                    if (messageFields != null
-                        && messageFields.TryGetValue(MessageFieldErrorCode, out string errorCodeObj))
+                    if (structuredMessageFields != null)
                     {
-                        // The result of TryParse is not being tracked since errorCodeValue has already been initialized to a default value of InvalidErrorCode.
-                        _ = int.TryParse(errorCodeObj, NumberStyles.Any, CultureInfo.InvariantCulture, out errorCodeValue);
+                        if (structuredMessageFields.TrackingId != null)
+                        {
+                            s_trackingId = structuredMessageFields.TrackingId;
```
drwill-ms, line 58: Can't save this as a static member, otherwise it is no longer thread safe.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -18,181 +16,117 @@ internal class ExceptionHandlingHelper
...
-                    // In some scenarios, the error response string is a ';' delimited string with the service-returned error code.
-                    const char errorFieldsDelimiter = ';';
-                    string[] messageFields = responseContent.Message?.Split(errorFieldsDelimiter);
+            // In some scenarios, the error response string is a semicolon delimited string with the service-returned error code
+            // embedded in a string response.
+            const char errorFieldsDelimiter = ';';
+            string[] messageFields = responseContent.Message?.Split(errorFieldsDelimiter);
 
-                    if (messageFields != null)
+            if (messageFields == null && Logging.IsEnabled)
```
drwill-ms, line 84:
```suggestion
            if (messageFields == null || messageFields.Count < 2)
            {
                if (Logging.IsEnabled)
```

iothub/service/src/Exceptions/IoTHubExceptionResult.cs
```diff
@@ -3,7 +3,7 @@
 
 using Newtonsoft.Json;
 
-namespace Microsoft.Azure.Devices
+namespace Microsoft.Azure.Devices.Exceptions
```
drwill-ms, line 6:
```suggestion
namespace Microsoft.Azure.Devices
```

iothub/service/src/Exceptions/IotHubServiceException.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// error message, a flag indicating if the error was transient, an optional tracking id and an optional reference
+        /// to the inner exception that caused this exception.
+        /// </summary>
+        /// <param name="errorCode">The 3-digit error iotHubStatusCode returned back in the hub service response.</param>
+        /// <param name="iotHubStatusCode">The 6-digit error iotHubStatusCode representing a more specific error in details.</param>
+        /// <param name="message">The message that describes the error.</param>
+        /// <param name="isTransient">Indicates if the error is transient and should be retried.</param>
+        /// <param name="trackingId">The service returned tracking Id associated with this particular error.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception.</param>
+        public IotHubServiceException(HttpStatusCode errorCode, IotHubErrorCode iotHubStatusCode, string message, bool isTransient, string trackingId = null, Exception innerException = null)
```
drwill-ms, line 40: Should be able to deprecate this ctor overload as well.

drwill-ms: If you choose to keep this, and I understand why, reorder the params to make message first, inner exception second (to maintain the same order as the other).

drwill-ms: Okay, except optional last.

iothub/service/src/Exceptions/ResponseMessage.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.Exceptions
```
drwill-ms, line 6:
```suggestion
namespace Microsoft.Azure.Devices
```

iothub/service/src/Exceptions/ResponseMessage.cs
```diff
@@ -0,0 +1,25 @@
...
+        internal string ErrorCode { get; set; }
+
+        [JsonProperty("trackingId")]
+        internal string TrackingId { get; set; }
+
+        [JsonProperty("message")]
+        internal string Message { get; set; }
+
+        [JsonProperty("timestampUtc")]
+        internal string TimestampUtc { get; set; }
```
drwill-ms, line 23:
```suggestion
        internal string OccurredOnUtc { get; set; }
```

drwill-ms: Could the type be DateTimeOffset?

iothub/service/src/Feedback/MessageFeedbackProcessorClient.cs
```diff
@@ -95,15 +96,15 @@ protected MessageFeedbackProcessorClient()
...
+        /// <exception cref="IotHubServiceClient"> with <see cref="HttpStatusCode.RequestTimeout"/>Thrown when the operation has been canceled. The inner exception will be
         /// <see cref="OperationCanceledException"/>.</exception>
         /// <exception cref="SocketException">Thrown if a socket error occurs.</exception>
         /// <exception cref="WebSocketException">Thrown if an error occurs when performing an operation on a WebSocket connection.</exception>
         /// <exception cref="IOException">Thrown if an I/O error occurs.</exception>
-        /// <exception cref="IotHubException">Thrown if an error occurs when communicating with IoT hub service.
-        /// If <see cref="IotHubException.IsTransient"/> is set to <c>true</c> then it is a transient exception.
-        /// If <see cref="IotHubException.IsTransient"/> is set to <c>false</c> then it is a non-transient exception.</exception>
+        /// <exception cref="IotHubServiceException">Thrown if an error occurs when communicating with IoT hub service.
+        /// If <see cref="IotHubServiceException.IsTransient"/> is set to <c>true</c> then it is a transient exception.
```
drwill-ms, line 106: Recommend getting rid of this all over the place. Just say the exception type and the user will go check out what the interesting properties are.

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -134,5 +127,28 @@ private static string EscapeETag(string eTag)
 
             return escapedETagBuilder.ToString();
         }
+
+        private static bool GetIsTransientFlag(HttpStatusCode statusCode, IotHubErrorCode errorCode)
```
drwill-ms, line 131: DetermineIfTransient

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -134,5 +127,28 @@ private static string EscapeETag(string eTag)
...
+
+                case IotHubErrorCode.Unknown:
+                    if (statusCode == HttpStatusCode.RequestTimeout)
+                    {
+                        return true;
+                    }
+                    else
+                    {
+                        return false;
+                    }
```
drwill-ms, line 147:
```suggestion
                    return statusCode == HttpStatusCode.RequestTimeout;
```

iothub/service/tests/ExceptionHandlingHelperTests.cs
```diff
@@ -17,78 +21,137 @@ public class ExceptionHandlingHelperTests
...
         {
             // arrange
-            var httpResponseMessage = new HttpResponseMessage(System.Net.HttpStatusCode.BadRequest);
+            var httpResponseMessage = new HttpResponseMessage(HttpStatusCode.BadRequest);
             var exceptionResult = new IoTHubExceptionResult
             {
-                Message = "{\"errorCode\":404001}"
+                // A read-world message in the response content which includes the numeric error code returned by the hub service.
+                Message = 
+                "{" +
```
drwill-ms, line 32: Can you create this as that new type and serialize it?

drwill-ms: Rather delete this one in favor of the next one that does it.

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -29,7 +28,7 @@ internal static Task<string> GetExceptionMessageAsync(HttpResponseMessage respon
         // the error description in the response header. The SDK will attempt to retrieve the integer error code
         // in the field of ErrorCode from the response content. If it works, the SDK will populate the exception
         // with the proper Code. Otherwise the SDK returns IotHubStatusCode.Unknown and log an error.
-        internal static async Task<IotHubErrorCode> GetIotHubErrorCodeAsync(HttpResponseMessage response)
+        internal static async Task<KeyValuePair<string, IotHubErrorCode>> GetErrorCodeAndTrackingIdAsync(HttpResponseMessage response)
```
drwill-ms, line 31: Not keyvalue pair, a tuple (string, IotHubErrorCode).

iothub/service/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -108,25 +109,25 @@ internal static async Task<IotHubErrorCode> GetIotHubErrorCodeAsync(HttpResponse
                         // When the returned error code is numeric, only take the first 6 characters as it contains 6 digits.
                         if (int.TryParse(returnedErrorCode.Substring(0, 6), out int code))
                         {
-                            return (IotHubErrorCode)code;
+                            return new KeyValuePair<string, IotHubErrorCode>(string.Empty, (IotHubErrorCode)code);
```
drwill-ms, line 112: return Tuple.Create(string.Empty, (IotHubErrorCode)code);

drwill-ms (approved):

PR #2673, opened 2022-09-07
Refactor ServiceClient ETag usage to match track 2

tmahmood-microsoft:
Extending work from the following PR to ConfigurationsClient, ModulesClient and TwinsClient:
https://github.com/Azure/azure-iot-sdk-csharp/pull/2646

### DIFF
--- a/e2e/test/iothub/service/ConfigurationsClientE2ETests.cs
+++ b/e2e/test/iothub/service/ConfigurationsClientE2ETests.cs
@@ -4,7 +4,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Azure;
 using FluentAssertions;
+using Microsoft.Azure.Devices.Common.Exceptions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
@@ -54,7 +56,7 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
                 addResult.TargetCondition.Should().Be(expected.TargetCondition);
                 addResult.Content.DeviceContent.First().Should().Be(expected.Content.DeviceContent.First());
                 addResult.Metrics.Queries.First().Should().Be(expected.Metrics.Queries.First());
-                addResult.ETag.Should().NotBeNullOrEmpty();
+                addResult.ETag.ToString().Should().NotBeNullOrEmpty();
 
                 Configuration getResult = await sc.Configurations.GetAsync(configurationId).ConfigureAwait(false);
                 getResult.Id.Should().Be(configurationId);
@@ -75,7 +77,7 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
                 updateResult.TargetCondition.Should().Be(expected.TargetCondition);
                 updateResult.Content.DeviceContent.First().Should().Be(expected.Content.DeviceContent.First());
                 updateResult.Metrics.Queries.First().Should().Be(expected.Metrics.Queries.First());
-                updateResult.ETag.Should().NotBeNullOrEmpty().And.Should().NotBe(getResult.ETag, "The ETag should have changed after update");
+                updateResult.ETag.ToString().Should().NotBeNullOrEmpty().And.Should().NotBe(getResult.ETag, "The ETag should have changed after update");
             }
             finally
             {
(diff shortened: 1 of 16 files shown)


e2e/test/iothub/twin/TwinE2ETests.cs
```diff
@@ -654,6 +656,58 @@ await deviceClient
             Twin serviceTwin = await serviceClient.Twins.GetAsync(testDevice.Id).ConfigureAwait(false);
             Assert.IsFalse(serviceTwin.Properties.Reported.Contains(propName1));
         }
+
+        [DataTestMethod, Timeout(LongRunningTestTimeoutMilliseconds)]
+        [DataRow(IotHubClientTransportProtocol.Tcp)]
+        [DataRow(IotHubClientTransportProtocol.WebSocket)]
+        [TestCategory("LongRunning")]
+        public async Task Twin_Client_SetETag_Works(IotHubClientTransportProtocol transportProtocol)
```
tmahmood-microsoft, line 664: This test is failing with the error:
Newtonsoft.Json.JsonWriterException: Unsupported type: Azure.ETag. Use the JsonSerializer class to get the object's JSON representation. Path ''.

Any suggestions for it?

e2e/test/iothub/service/ConfigurationsClientE2ETests.cs
```diff
@@ -86,5 +88,162 @@ public async Task ConfigurationOperations_Work()
...
+
+                configuration = await serviceClient.Configurations.SetAsync(configuration).ConfigureAwait(false);
+
+                configuration.ETag = oldEtag;
+
+                try
+                {
+                    // set the 'onlyIfUnchanged' flag to true to check that, with an out of date ETag, the request throws a PreconditionFailedException.
+                    await serviceClient.Configurations.SetAsync(configuration, true).ConfigureAwait(false);
+                    throw new AssertFailedException("Expected test to throw a precondition failed exception since it updated a configuration with an out of date ETag");
```
drwill-ms, line 129: You can model this differently/easier with [FluentAssertions](https://fluentassertions.com/exceptions/).

drwill-ms (approved):

e2e/test/iothub/twin/TwinE2ETests.cs
timtay-microsoft: Did you add [this tag](https://github.com/Azure/azure-iot-sdk-csharp/blob/7a9ccb4d95c5ff65d811563a8696f1b82f70ae9d/iothub/service/src/Registry/Models/Device.cs#L55) to your ETag field that fails to serialize?

e2e/test/iothub/service/ConfigurationsClientE2ETests.cs
```diff
@@ -86,5 +88,162 @@ public async Task ConfigurationOperations_Work()
...
+                    Priority = 2,
+                    Labels = { { "labelName", "labelValue" } },
+                    TargetCondition = "deviceId='fakeDevice'",
+                    Content = new ConfigurationContent
+                    {
+                        DeviceContent = { { "properties.desired.x", 4L } },
+                    },
+                    Metrics = new ConfigurationMetrics
+                    {
+                        Queries = { { "successfullyConfigured", "select deviceId from devices where properties.reported.x = 4" } }
```
timtay-microsoft, line 111: Is there any risk this configuration could affect other tests' devices?

e2e/test/iothub/service/ConfigurationsClientE2ETests.cs
```diff
@@ -86,5 +88,162 @@ public async Task ConfigurationOperations_Work()
...
+                catch (DeviceMessageLockLostException)
+                {
+                    throw new AssertFailedException("Did not expect test to throw a precondition failed exception since 'onlyIfUnchanged' was set to false");
+                }
+            }
+            finally
+            {
+                try
+                {
+                    if (configCreated)
```
timtay-microsoft, line 233: No need to check this. Just attempt the delete operation and let it throw a not found exception that the test ignores

iothub/service/src/Configurations/ConfigurationsClient.cs
```diff
@@ -227,22 +210,22 @@ public virtual Task<Configuration> SetAsync(Configuration configuration, Cancell
...
         {
             if (Logging.IsEnabled)
-                Logging.Enter(this, $"Updating configuration: {configuration?.Id} - Force update: {forceUpdate}", nameof(SetAsync));
+                Logging.Enter(this, $"Updating configuration: {configuration?.Id} - only if changed: {onlyIfUnchanged}", nameof(SetAsync));
 
             try
             {
                 Argument.AssertNotNull(configuration, nameof(configuration));
-                if (string.IsNullOrWhiteSpace(configuration.ETag) && !forceUpdate)
+                if (string.IsNullOrWhiteSpace(configuration.ETag.ToString()) && onlyIfUnchanged)
```
timtay-microsoft, line 221: Go ahead and remove this validation here and elsewhere. The service will do this validation for us.

iothub/service/src/Twin/TwinsClient.cs
```diff
@@ -169,7 +175,7 @@ public virtual async Task<Twin> GetAsync(string deviceId, string moduleId, Cance
         /// certificate validation.
         /// </exception>
         /// <exception cref="OperationCanceledException">If the provided <paramref name="cancellationToken"/> has requested cancellation.</exception>
-        public virtual async Task<Twin> UpdateAsync(string deviceId, Twin twinPatch, string etag, CancellationToken cancellationToken = default)
+        public virtual async Task<Twin> UpdateAsync(string deviceId, Twin twinPatch, ETag etag, bool onlyIfUnchanged = false, CancellationToken cancellationToken = default)
```
timtay-microsoft, line 178: The ETag for this twin is already in the ```twinPatch``` argument, so we don't need to ask for it as a separate parameter here

timtay-microsoft:
```suggestion
        public virtual async Task<Twin> UpdateAsync(string deviceId, Twin twinPatch, bool onlyIfUnchanged = false, CancellationToken cancellationToken = default)
```

timtay-microsoft:
You should be able to remove [this function](https://github.com/Azure/azure-iot-sdk-csharp/blob/7a9ccb4d95c5ff65d811563a8696f1b82f70ae9d/iothub/service/src/Http/HttpMessageHelper.cs#L95) now that we use [this function](https://github.com/Azure/azure-iot-sdk-csharp/blob/7a9ccb4d95c5ff65d811563a8696f1b82f70ae9d/iothub/service/src/Http/HttpMessageHelper.cs#L110) everywhere. The ```onlyIfUnchanged``` flag in this function should also be mandatory now

e2e/test/iothub/service/ConfigurationsClientE2ETests.cs
```diff
@@ -86,5 +88,123 @@ public async Task ConfigurationOperations_Work()
...
+                FluentActions
+                .Invoking(async () => { await serviceClient.Configurations.SetAsync(configuration, true).ConfigureAwait(false); })
+                .Should()
+                .NotThrow<DeviceMessageLockLostException>("Did not expect test to throw a precondition failed exception since 'onlyIfUnchanged' was set to true");
+            }
+            finally
+            {
+                try
+                {
+                    // If this fails, we shall let it throw an exception and fail the test
```
timtay-microsoft, line 139: It looks like we actually don't care if this throws, FYI

timtay-microsoft:
```suggestion
                    // If this fails, it won't fail the test
```

timtay-microsoft (approved):

PR #2678, opened 2022-09-08
Update device and module client to inherit from internal client

abhipsaMisra:


iothub/device/src/ConnectionStatus.cs
```diff
@@ -32,7 +32,7 @@ public enum ConnectionStatus
         /// The device connection was closed gracefully.
         /// </summary>
         /// <remarks>
-        /// To perform more operations on the client, call <see cref="IotHubDeviceClient.Dispose()"/> and then re-initialize the client.
```
abhipsaMisra, line 35: Cannot resolve the inherited method in doc comments. This is supposed to link to IotHubDeviceClient.Dispose() and IotHubModuleClient.Dispose().

iothub/device/src/InternalClient.cs
```diff
@@ -3,33 +3,34 @@
...
     /// <summary>
-    /// Contains methods that a device can use to send messages to and receive messages from the service,
+    /// Contains methods that a client can use to send messages to and receive messages from the service,
     /// respond to direct method invocations from the service, and send and receive twin property updates.
     /// </summary>
-    internal class InternalClient : IDisposable
+    public class InternalClient : IDisposable
     {
+        private protected readonly PipelineContext _pipelineContext;
+        private protected readonly IotHubClientOptions _clientOptions;
```
abhipsaMisra, line 26: these are referenced by IotHubDeviceClient and IotHubModuleClient, so making them private protected.

iothub/device/src/InternalClient.cs
```diff
@@ -3,33 +3,34 @@
...
 
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// Contains methods that a device can use to send messages to and receive messages from the service,
+    /// Contains methods that a client can use to send messages to and receive messages from the service,
     /// respond to direct method invocations from the service, and send and receive twin property updates.
     /// </summary>
-    internal class InternalClient : IDisposable
+    public class InternalClient : IDisposable
```
abhipsaMisra, line 23: This base class needs to be public

iothub/device/src/InternalClient.cs
```diff
@@ -3,33 +3,34 @@
...
     {
+        private protected readonly PipelineContext _pipelineContext;
+        private protected readonly IotHubClientOptions _clientOptions;
+
         private readonly SemaphoreSlim _methodsSemaphore = new(1, 1);
-        private readonly SemaphoreSlim _deviceReceiveMessageSemaphore = new(1, 1);
```
abhipsaMisra, line 22: Moved the device and module specific variables to their respective classes instead

iothub/device/src/InternalClient.cs
```diff
@@ -57,58 +50,46 @@ internal class InternalClient : IDisposable
...
-                && iotHubClientOptions.TransportSettings is IotHubClientHttpSettings)
+            if (!_clientOptions.ModelId.IsNullOrWhiteSpace()
+                && _clientOptions.TransportSettings is IotHubClientHttpSettings)
             {
                 throw new InvalidOperationException("Plug and Play is not supported over the HTTP transport.");
             }
 
-            var pipelineContext = new PipelineContext
+            _pipelineContext = new PipelineContext
             {
```
abhipsaMisra, line 76: The internal client now sets up the common properties in the pipeline context. The device client and module client will set up their specific properties and then build the pipeline.

iothub/device/src/InternalClient.cs
```diff
@@ -213,9 +219,17 @@ public async Task SendEventAsync(Message message, CancellationToken cancellation
         }
 
         /// <summary>
-        /// Sends a batch of events to device hub
+        /// Sends a batch of events to IoT hub. Use AMQP or HTTPs for a true batch operation. MQTT will just send the messages
```
abhipsaMisra, line 222: Moved the doc comments from the client class

iothub/device/src/InternalClient.cs
```diff
@@ -587,6 +643,61 @@ public async Task CloseAsync(CancellationToken cancellationToken = default)
             }
         }
 
+        /// <summary>
+        /// Releases the unmanaged resources used by the client and optionally disposes of the managed resources.
+        /// </summary>
+        /// <remarks>
+        /// The method <see cref="CloseAsync(CancellationToken)"/> should be called before disposing.
+        /// </remarks>
+        public void Dispose()
```
abhipsaMisra, line 652: The public dispose method is now exposed on the base class

iothub/device/src/InternalClient.cs
```diff
@@ -712,159 +823,6 @@ private async Task HandleMethodEnableAsync(CancellationToken cancellationToken =
             _isDeviceMethodEnabled = true;
         }
 
-        #region Device Specific API
```
abhipsaMisra, line 715: These have been moved to IotHubDeviceClient

iothub/device/src/InternalClient.cs
```diff
@@ -879,326 +837,10 @@ private async Task HandleMethodDisableAsync(CancellationToken cancellationToken
             _isDeviceMethodEnabled = false;
         }
 
-        #endregion Device Specific API
-
-        #region Module Specific API
```
abhipsaMisra, line 884: These have been moved to IotHubModuleClient

iothub/device/src/Pipeline/ClientPipelineBuilder.cs
```diff
@@ -6,11 +6,11 @@
 
 namespace Microsoft.Azure.Devices.Client
 {
-    internal class DeviceClientPipelineBuilder : IDeviceClientPipelineBuilder
```
abhipsaMisra, line 9: This doesn't really require an interface to be defined

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -329,7 +329,7 @@ public async Task IotHubDeviceClient_OnMethodCalled_NullMethodRequest()
                 return Task.FromResult(directMethodResponseWithPayload);
             }, "custom data").ConfigureAwait(false);
 
-            await deviceClient.InternalClient.OnMethodCalledAsync(null).ConfigureAwait(false);
```
abhipsaMisra, line 332: The method is available on the client directly

iothub/device/src/InternalClient.cs
```diff
@@ -3,33 +3,34 @@
...
-        private readonly SemaphoreSlim _deviceReceiveMessageSemaphore = new(1, 1);
-        private readonly SemaphoreSlim _moduleReceiveMessageSemaphore = new(1, 1);
         private readonly SemaphoreSlim _twinDesiredPropertySemaphore = new(1, 1);
-        private readonly HttpTransportHandler _fileUploadHttpTransportHandler;
-        private readonly IotHubClientOptions _clientOptions;
 
         // Connection status change information
         private volatile Action<ConnectionStatusInfo> _connectionStatusChangeHandler;
 
-        internal ConnectionStatusInfo _connectionStatusInfo { get; private set; } = new ConnectionStatusInfo();
```
timtay-microsoft, line 31: Why is this disappearing from InternalClient? Wouldn't both Device and Modules have this?

iothub/device/src/InternalClient.cs
```diff
@@ -57,58 +50,46 @@ internal class InternalClient : IDisposable
...
+                iotHubClientOptions = new();
+            }
 
             IotHubConnectionCredentials = iotHubConnectionCredentials;
             _clientOptions = iotHubClientOptions;
 
-            if (!iotHubClientOptions.ModelId.IsNullOrWhiteSpace()
-                && iotHubClientOptions.TransportSettings is IotHubClientHttpSettings)
+            if (!_clientOptions.ModelId.IsNullOrWhiteSpace()
+                && _clientOptions.TransportSettings is IotHubClientHttpSettings)
```
timtay-microsoft, line 70: Aren't we already not allowing users to use HTTP? We can get rid of this check, if so

iothub/device/src/InternalClient.cs
timtay-microsoft: Oh, never mind. I see it further below

iothub/device/src/InternalClient.cs
```diff
@@ -154,7 +139,8 @@ public void SetRetryPolicy(IRetryPolicy retryPolicy)
 
         /// <summary>
         /// Sets a new delegate for the connection status changed callback. If a delegate is already associated,
-        /// it will be replaced with the new delegate.
+        /// it will be replaced with the new delegate. Note that this callback will never be called if the client is configured to use
+        /// HTTP, as that protocol is stateless.
```
timtay-microsoft, line 143: If we don't support HTTP, then we can remove this statement

iothub/device/src/InternalClient.cs
```diff
@@ -265,11 +281,12 @@ public async Task CompleteMessageAsync(string lockToken, CancellationToken cance
         }
 
         /// <summary>
-        /// Deletes a received message from the device queue
+        /// Deletes a received message from the client queue.
```
timtay-microsoft, line 284:
```suggestion
        /// Deletes a received message from the client's service-side queue.
```

iothub/device/src/InternalClient.cs
```diff
@@ -351,9 +389,16 @@ public async Task RejectMessageAsync(string lockToken, CancellationToken cancell
         }
 
         /// <summary>
-        /// Deletes a received message from the device queue and indicates to the server that the message could not be processed.
+        /// Deletes a received message from the client queue and indicates to the server that the message could not be processed.
```
timtay-microsoft, line 392:
```suggestion
        /// Deletes a received message from the client's service-side queue and indicates to the server that the message could not be processed.
```

iothub/device/src/InternalClient.cs
```diff
@@ -473,9 +518,14 @@ public async Task RejectMessageAsync(Message message, CancellationToken cancella
         }
 
         /// <summary>
-        /// Retrieve the device twin properties for the current device.
-        /// For the complete device twin object, use Microsoft.Azure.Devices.RegistryManager.GetTwinAsync(string deviceId).
+        /// Retrieve the twin properties for the current client. The client instance must be opened already.
+        /// For the complete twin object, use Microsoft.Azure.Devices.IotHubServiceClient.Twins.GetAsync(string deviceId, CancellationToken cancellationToken)
```
timtay-microsoft, line 522: It feels odd to recommend users to call a service client API here in the device client docs. Maybe we need more nuance saying that the device-side view of the twin is different from the service-side view and link to some msdoc explaining further?

timtay-microsoft (approved):

iothub/device/src/InternalClient.cs
abhipsaMisra: good point

andyk-ms (approved):

iothub/device/src/InternalClient.cs
abhipsaMisra: How about this:
```c#
/// <summary>
/// Retrieve the twin properties for the current client. The client instance must be opened already.
/// This API gives you the client's view of the twin. For more information on twins in IoT hub see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-device-twins"/>.
/// </summary>
```

iothub/device/src/ConnectionStatus.cs
```diff
@@ -32,7 +32,7 @@ public enum ConnectionStatus
...
+        /// To perform more operations on the client, call <see cref="InternalClient.Dispose()"/> and then re-initialize the client.
```
drwill-ms, line 35: Maybe InternalClient is the wrong name now. It feels weird to expose that type name to users.

drwill-ms: Also, even though it is inherited, shouldn't there be an override on each of these classes that calls the base?

iothub/device/src/InternalClient.cs
drwill-ms: Maybe they should be properties with a protected getter and private setter?

iothub/device/src/ConnectionStatus.cs
timtay-microsoft: I feel the same way. ```IotHubBaseClient```?

iothub/device/src/InternalClient.cs
```diff
@@ -57,58 +50,40 @@ internal class InternalClient : IDisposable
...
+                IotHubClientTransportSettings = _clientOptions.TransportSettings,
+                ModelId = _clientOptions.ModelId,
                 MethodCallback = OnMethodCalledAsync,
                 DesiredPropertyUpdateCallback = OnDesiredStatePatchReceived,
                 ConnectionStatusChangeHandler = OnConnectionStatusChanged,
-                ModuleEventCallback = OnModuleEventMessageReceivedAsync,
-                DeviceEventCallback = OnDeviceMessageReceivedAsync,
             };
 
-            pipelineBuilder ??= BuildPipeline();
```
drwill-ms, line 92: Could have abstract methods for deriving classes to add code for init, that the base calls, so it can init the pipeline as best it can, let a deriving class add on, and then call build (so derived classes don't have to know to do that).

iothub/device/src/InternalClient.cs
```diff
@@ -57,58 +50,40 @@ internal class InternalClient : IDisposable
...
-
-            if (Logging.IsEnabled)
-                Logging.Exit(this, iotHubClientOptions.TransportSettings, pipelineBuilder, nameof(InternalClient) + "_ctor");
+                Logging.Exit(this, _clientOptions.TransportSettings, nameof(InternalClient) + "_ctor");
         }
 
         /// <summary>
         /// Diagnostic sampling percentage value, [0-100];
-        /// A value of 0 means no message will carry on diagnostics info.
+        /// 0 means no message will carry on diagnostics info
```
drwill-ms, line 86: I don't think this change is better.

iothub/device/src/InternalClient.cs
```diff
@@ -130,17 +105,21 @@ public int DiagnosticSamplingPercentage
             }
         }
 
+        /// <summary>
+        /// The latest connection status information since the last status change.
+        /// </summary>
+        public ConnectionStatusInfo ConnectionStatusInfo { get; private set; } = new ConnectionStatusInfo();
```
drwill-ms, line 111:
```suggestion
        public ConnectionStatusInfo ConnectionStatusInfo { get; private set; } = new();
```

iothub/device/src/ConnectionStatus.cs
abhipsaMisra: I agree that internal client here feels wrong, however, the doc comments aren't able to resolve the inherited method. We aren't overridding the implementations in the derived class; the device client and module client instance will call the implementations in the base internal client class for the "common" APIs.

iothub/device/src/InternalClient.cs
abhipsaMisra: Ah, I copied this over from the public class. I agree that this change isn't better than what we previously had on this class. I'll revert it.

iothub/device/src/InternalClient.cs
```diff
@@ -185,9 +165,28 @@ public async Task OpenAsync(CancellationToken cancellationToken = default)
...
+        /// <remarks>
+        /// In case of a transient issue, retrying the operation should work. In case of a non-transient issue, inspect
+        /// the error details and take steps accordingly.
+        /// Please note that the list of exceptions is not exhaustive.
+        /// </remarks>
+        /// <param name="message">The message to send.</param>
+        /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <exception cref="ArgumentNullException">Thrown when a required parameter is null.</exception>
+        /// <exception cref="IotHubClientException">Thrown and <see cref="IotHubClientException.StatusCode"/> is set to <see cref="IotHubStatusCode.NetworkErrors"/>
+        /// when the operation has been canceled. The inner exception will be <see cref="OperationCanceledException"/>.</exception>
```
drwill-ms, line 179: @brycewang-microsoft you'll be changing this to throw OperationCanceledException natively instead of as an inner exception, right?

iothub/device/src/InternalClient.cs
```diff
@@ -185,9 +165,28 @@ public async Task OpenAsync(CancellationToken cancellationToken = default)
...
+        /// <exception cref="IotHubClientException">Thrown and <see cref="IotHubClientException.StatusCode"/> is set to <see cref="IotHubStatusCode.NetworkErrors"/>
+        /// if the client encounters a transient retryable exception. </exception>
+        /// <exception cref="InvalidOperationException">Thrown if the client instance is not opened already.</exception>
+        /// <exception cref="SocketException">Thrown if a socket error occurs.</exception>
+        /// <exception cref="WebSocketException">Thrown if an error occurs when performing an operation on a WebSocket connection.</exception>
+        /// <exception cref="IOException">Thrown if an I/O error occurs.</exception>
+        /// <exception cref="ClosedChannelException">Thrown if the MQTT transport layer closes unexpectedly.</exception>
+        /// <exception cref="IotHubClientException">Thrown if an error occurs when communicating with IoT hub service.
+        /// If <see cref="IotHubClientException.IsTransient"/> is set to <c>true</c> then it is a transient exception.
+        /// If <see cref="IotHubClientException.IsTransient"/> is set to <c>false</c> then it is a non-transient exception.</exception>
```
drwill-ms, line 189:
```suggestion
        /// If <see cref="IotHubClientException.IsTransient"/> is set to <c>true</c> then it is a transient exception and should be retried,
        /// but if <c>false</c> then it is a non-transient exception and should probably not be retried.</exception>
```

iothub/device/src/InternalClient.cs
```diff
@@ -265,11 +274,12 @@ public async Task CompleteMessageAsync(string lockToken, CancellationToken cance
         }
 
         /// <summary>
-        /// Deletes a received message from the device queue
+        /// Deletes a received message from the client's service-side queue.
```
drwill-ms, line 277: service-side queue?

iothub/device/src/InternalClient.cs
```diff
@@ -311,9 +328,16 @@ public async Task AbandonMessageAsync(string lockToken, CancellationToken cancel
         }
 
         /// <summary>
-        /// Puts a received message back onto the device queue
+        /// Puts a received message back onto the client queue.
         /// </summary>
-        /// <returns>The lock identifier for the previously received message</returns>
+        /// <remarks>
+        /// You cannot reject or abandon messages over MQTT protocol. For more details, see
```
drwill-ms, line 334: Prefer non-personal documentation, such as:

```suggestion
        /// Messages cannot be rejected or abandoned over the MQTT protocol. For more details, see
```

iothub/device/src/ConnectionStatus.cs
drwill-ms: We're using the suggested pattern for dispose, so never mind it being an override. I still don't love the name InternalClient being what is shown to users, though. It feels misleading when they are calling it on their client instance.

iothub/device/src/IotHubDeviceClient.cs
```diff
@@ -18,8 +13,15 @@ namespace Microsoft.Azure.Devices.Client
     /// Contains methods that a device can use to send messages to and receive from the service.
     /// </summary>
     /// <threadsafety static="true" instance="true" />
-    public class IotHubDeviceClient : IDisposable
+    public class IotHubDeviceClient : InternalClient, IDisposable
```
drwill-ms, line 16: It does not need IDisposable on its own as it inherits it.

iothub/device/src/IotHubDeviceClient.cs
```diff
@@ -57,131 +59,44 @@ public IotHubDeviceClient(string hostName, IAuthenticationMethod authenticationM
         }
 
         private IotHubDeviceClient(IotHubConnectionCredentials iotHubConnectionCredentials, IotHubClientOptions options)
+            : base(iotHubConnectionCredentials, options)
         {
+            // Validate
             if (iotHubConnectionCredentials.ModuleId != null)
             {
                 throw new ArgumentException("A module Id was specified in the connection string - please use IotHubModuleClient for modules.");
```
drwill-ms, line 67: ArgumentException should specify the argument, however this isn't a parameter sent by the user. Maybe we should use InvalidOperationException instead?

iothub/device/src/IotHubDeviceClient.cs
```diff
@@ -191,175 +106,89 @@ public Task SendEventBatchAsync(IEnumerable<Message> messages, CancellationToken
...
+                    // If a null delegate is passed, it will disable the callback triggered on receiving messages from the service.
+                    _deviceReceiveMessageCallback = null;
+                    await DisableReceiveMessageAsync(cancellationToken).ConfigureAwait(false);
+                }
+            }
+            catch (IotHubClientException ex) when (ex.InnerException is OperationCanceledException)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                throw;
+            }
```
drwill-ms, line 177: Can we go upstream to find what is converting an OperationCanceledException into an IotHubClientException and not do that?

iothub/device/src/InternalClient.cs
timtay-microsoft: The service is the party that maintains the queue that this operation removes an entry from

timtay-microsoft: Maybe "Deletes a received message from the service's cloud to device message queue for this client"?

iothub/device/src/IotHubDeviceClient.cs
abhipsaMisra: Yeah, I like that

abhipsaMisra: IotHubModuleClient, however, throws an ArgumentException/ArgumentNullException if the module Id isn't specified. For the sake of keeping exceptions similar, what do think about us throwing ArgumentException here and passing in the moduleId as arg name. Even though not set directly by the user, they do set it indirectly. The connection string flow might still be a bit confusing to corelate with this error though.

iothub/device/src/IotHubDeviceClient.cs
abhipsaMisra: I'm not seeing an OperationCancelledException being converted to an IotHubException in the amqp flow. This is possibly leftover code from our previous exception handling logic. Since we're rewriting the mqtt flow anyway, I'll go ahead and remove this block.

abhipsaMisra: Actually, scratch that, this block will do the conversion in our delegating handler pipeline: 
https://github.com/Azure/azure-iot-sdk-csharp/blob/f512c192d2b5b9f55a7e4a7c433f4969cd0e96a4/iothub/device/src/Pipeline/ErrorDelegatingHandler.cs#L202-L205
https://github.com/Azure/azure-iot-sdk-csharp/blob/f512c192d2b5b9f55a7e4a7c433f4969cd0e96a4/iothub/device/src/Pipeline/ErrorDelegatingHandler.cs#L29-L38

We should remove OperationCanceledException from the list of expected network exceptions.

iothub/device/src/InternalClient.cs
abhipsaMisra: Sure, that'll work. I'll rework this with the change to make the device client/ module client only add to the context and not necessarily build it entirely.

iothub/device/src/InternalClient.cs
abhipsaMisra: I wanted to go with this approach initially, but this being set up in the constructor makes our implementation options slightly limited. I'll try it again to see if I can find a way to make this work.

iothub/device/src/InternalClient.cs
drwill-ms: Suggest putting the second sentence in <remarks>.

iothub/device/src/IotHubDeviceClient.cs
drwill-ms: Yeah, I'd rather not specify an argument that isn't immediately obvious to the user in context of the parameter names they passed.

drwill-ms: IOW, I think we should also change the module client to be consistent with this approach.

iothub/device/src/IotHubDeviceClient.cs
drwill-ms: Are we going to add a task for that?

iothub/device/src/InternalClient.cs
abhipsaMisra: done

iothub/device/src/IotHubDeviceClient.cs
abhipsaMisra: Syncing up with Bryce to get this added to a task.

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -587,6 +645,63 @@ public async Task CloseAsync(CancellationToken cancellationToken = default)
...
+        {
+            if (disposing)
+            {
+                InnerHandler?.Dispose();
+                _methodsSemaphore?.Dispose();
+                _twinDesiredPropertySemaphore?.Dispose();
+            }
+        }
+
+        internal abstract void AddToPipelineContext();
```
drwill-ms, line 676: protected?

drwill-ms (approved):

PR #2680, opened 2022-09-08
Recycle a single HttpClient for all provisioning service client's requests

timtay-microsoft:
The query APIs used to create a new instance of ContractApiHttp each request.

Also removes overloads of query APIs that took ProvisioningServiceHttpSettings since there is no reason to use a different HTTP client per query request

### DIFF
--- a/provisioning/service/src/Query.cs
+++ b/provisioning/service/src/Query.cs
@@ -56,7 +56,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
     ///     the point where you stopped. Just recreating the query with the same <see cref="QuerySpecification"/> and calling
     ///     the <see cref="NextAsync(string)"/> passing the stored <c>ContinuationToken</c>.
     /// </remarks>
-    public class Query : IDisposable
+    public class Query
     {
         private const string ContinuationTokenHeaderKey = "x-ms-continuation";
         private const string ItemTypeHeaderKey = "x-ms-item-type";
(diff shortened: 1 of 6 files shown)


provisioning/service/src/Query.cs
```diff
@@ -97,10 +97,7 @@ public class Query : IDisposable
                 throw new ArgumentException($"{nameof(pageSize)} cannot be negative.");
             }
 
-            // TODO: Refactor ContractApiHttp being created again
-            _contractApiHttp = new ContractApiHttp(
-                serviceConnectionString.HttpsEndpoint,
-                serviceConnectionString, httpTransportSettings);
```
timtay-microsoft, line 103: This block of code is the important bit here. Now we will pass around the single ```contractApiHttp``` instance everywhere instead of constructing a new one for each query

abhipsaMisra: Does this introduce any change with respect to processing limitations, i.e. will parallel requests on separate threads still execute asynchronously?

abhipsaMisra (approved):

provisioning/service/src/Query.cs
timtay-microsoft: No changes, no. We were already using the same httpClient to make every other request (createEnrollment, GetEnrollment, etc.) so this just brings the query requests in with them.

provisioning/service/src/Query.cs
```diff
@@ -97,10 +97,7 @@ public class Query : IDisposable
...
+            _contractApiHttp = contractApiHttp;
```
drwill-ms, line 100: It looks like it is still being disposed in this class. Should it be?

timtay-microsoft: Good catch. It shouldn't be disposed here. It should only be disposed when the service client is disposed.

drwill-ms (approved):

PR #2681, opened 2022-09-08
Allow users to pass in an HttpClient for ProvisioningServiceClient 

timtay-microsoft:
To construct a client with a custom Httpclient:
```csharp
var options = new ProvisioningServiceClientOptions()
{
    HttpClient = myCustomHttpClient
};
var provServiceClient = new ProvisioningServiceClient(connString, options);
```

To construct a client with some custom settings:
```csharp
var options = new ProvisioningServiceClientOptions()
{
    ProvisioningServiceHttpSettings = new ProvisioningServiceHttpSettings()
    {
        Proxy = myCustomProxy,
        SslProtocols = SslProtocols.None,
    }
};
var provServiceClient = new ProvisioningServiceClient(connString, options);
```


provisioning/service/src/ProvisioningServiceClientOptions.cs
```diff
@@ -13,8 +16,36 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
...
+        /// </remarks>
+        public SslProtocols SslProtocols { get; set; } = SslProtocols.None;
+
+        /// <summary>
+        /// To enable certificate revocation check.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to false.
+        /// </remarks>
+        public bool CertificateRevocationCheck { get; set; }
```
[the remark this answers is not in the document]
timtay-microsoft: Okay, I can revert this particular change then. The ProvisioningServiceClientOptions will contain the settable http client and a settable httptransportsettings

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -46,25 +46,32 @@ internal class ContractApiHttp : IContractApiHttp
         public ContractApiHttp(
             Uri baseAddress,
             IAuthorizationHeaderProvider authenticationHeaderProvider,
-            ProvisioningServiceHttpSettings httpSettings)
+            ProvisioningServiceClientOptions options)
         {
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
 
+            if (options.HttpClient != null)
```
drwill-ms, line 54: Are you sure `options` can't be null here?

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -46,25 +46,32 @@ internal class ContractApiHttp : IContractApiHttp
...
+            {
+                _httpClientObj = options.HttpClient;
+                _httpClientHandler = null;
```
drwill-ms, line 57: What does assigning null do here? Was there a previous instance? Did it need to be disposed?

timtay-microsoft: It doesn't do anything since this is the default value. I can remove this line

provisioning/service/src/Contract/ContractApiHttp.cs
timtay-microsoft: Yes. We deliberately ensure that options is never null even if the service client is constructed like

```csharp
new ProvisioningServiceClient("connString", null);
```

drwill-ms (approved):

abhipsaMisra (approved):

PR #2685, opened 2022-09-12
Unwrap OperationCanceledException from IotHubClientException

brycewang-microsoft:
For [this](https://github.com/Azure/azure-iot-sdk-csharp/pull/2678#discussion_r966512261), stop converting an `OperationCanceledException` into an `IotHubClientException`

### DIFF
--- a/e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
+++ b/e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
@@ -204,10 +204,6 @@ namespace Microsoft.Azure.Devices.E2ETests
             {
                 // It should always throw IotHubClientException with status code Unauthorized
             }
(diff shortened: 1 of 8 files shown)


drwill-ms:
I see all the places where we stop translating an IoT exception to an OperationCanceledException, but I don't see where we stop it from being thrown as an IoT exception in the first place. Am I missing it?

brycewang-microsoft:
> Is it not categorizing it as a network exception?

Yes, which it is addressed [here](https://github.com/Azure/azure-iot-sdk-csharp/pull/2685/files#diff-a571d81fa8211b471ff7876fda85aba25c2dbf0323a072dbe5baa6e06b6dba5b).

By removing from here, it won't be taken as "network" exception, nor wrapped into IotHubClientException as following:
```c#
else if (IsNetworkExceptionChain(ex)) 
 { 
     throw new IotHubClientException("Transient network error occurred, please retry.", ex, true, IotHubStatusCode.NetworkErrors); 
 }
```

drwill-ms (approved):

PR #2686, opened 2022-09-12
Ensure all model types are mockable

tmahmood-microsoft:
Followed the following Azure Design guidelines: https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-mocking

### DIFF
--- a/iothub/device/src/FileUpload/FileUploadCompletionNotification.cs
+++ b/iothub/device/src/FileUpload/FileUploadCompletionNotification.cs
@@ -11,9 +11,16 @@ namespace Microsoft.Azure.Devices.Client.Transport
     public class FileUploadCompletionNotification
     {
         /// <summary>
+        /// Initialize an instance of this class. Provided for unit testing purposes only.
+        /// </summary>
+        public FileUploadCompletionNotification()
+        {
+        }
+
+        /// <summary>
         /// The correlation id that maps this completion notification to the file upload.
         /// The value should equal the <see cref="FileUploadSasUriResponse.CorrelationId">correlation id </see>
-        /// returned from IoT hub when first getting the SAS Uri for this file upload 
+        /// returned from IoT hub when first getting the SAS Uri for this file upload
         /// </summary>
         [JsonProperty(PropertyName = "correlationId")]
         public string CorrelationId { get; set; }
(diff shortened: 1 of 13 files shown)


iothub/device/src/FileUpload/FileUploadCompletionNotification.cs
```diff
@@ -10,10 +10,17 @@ namespace Microsoft.Azure.Devices.Client.Transport
     /// </summary>
     public class FileUploadCompletionNotification
     {
+        /// <summary>
+        /// Initialize an instance of this class. Provided for unit testing purposes only.
+        /// </summary>
+        public FileUploadCompletionNotification()
```
drwill-ms, line 16:
```suggestion
        protected internal FileUploadCompletionNotification()
```

drwill-ms: If it is for unit testing only, make it protected internal so it can be called by our tests, or by end-users who need to mock it.

iothub/service/src/Messaging/Models/PurgeMessageQueueResult.cs
```diff
@@ -10,6 +10,21 @@ namespace Microsoft.Azure.Devices
...
+        /// <summary>
+        /// Initializes an instance of this class. Provided for unit testing purposes only.
+        /// </summary>
+        /// <param name="deviceId">
+        /// The Id of the device whose messages are being purged.
+        /// </param>
+        /// <param name="totalMessagesPurged">
+        /// The total number of messages that were purged from the device's queue.
+        /// </param>
+        public PurgeMessageQueueResult(string deviceId, int totalMessagesPurged)
```
drwill-ms, line 22: This class is public with public setters on the properties. It does not need a public ctor in its current state.

Also, if you add a ctor with params but don't offer one without, it can't be used to deserialize.

drwill-ms (changes requested):
Visibility spec is all over the place on various classes here. I'm looking for consistency.

timtay-microsoft:
What was the exact guidance from the Azure SDK team here? Protected, empty ctor on every class?

tmahmood-microsoft:
Krzysztof is out of the office till 9/26, I was not able to get a response.

drwill-ms:
> Krzysztof is out of the office till 9/26, I was not able to get a response.

Try Ted Glaza

tmahmood-microsoft:
Closing this for now as we need to have a discussion with Krzysztof about the azure SDK guidelines as the suggested changes to have Model Factory seems like a overkill for our case.

PR #2687, opened 2022-09-13
More code cleanup

drwill-ms:
More hub service code cleanup


e2e/test/iothub/SasCredentialAuthenticationTests.cs
```diff
@@ -98,12 +98,12 @@ public async Task JobClient_Http_SasCredentialAuth_Success()
                 {
                     QueryCondition = query,
                     Twin = twin,
-                    StartOn = DateTime.UtcNow
+                    StartOnUtc = DateTime.UtcNow,
```
drwill-ms, line 101: .NET track2 standards for naming date times are to use the keyword **on**.

e2e/test/iothub/service/ExportDevicesTests.cs
```diff
@@ -199,7 +199,7 @@ public async Task DevicesClient_ExportDevices(StorageAuthenticationType storageA
             {
                 await Task.Delay(s_waitDuration).ConfigureAwait(false);
                 jobProperties = await serviceClient.Devices.GetJobAsync(jobProperties.JobId).ConfigureAwait(false);
-                Logger.Trace($"Job {jobProperties.JobId} is {jobProperties.Status} with progress {jobProperties.Progress}% after {sw.Elapsed}.");
```
drwill-ms, line 202: progress seems to be only 0 or 100, so removing from the API surface so the user doesn't look to use it. Instead, they should use status. If progress ever gets interesting, we can make it public.

iothub/service/src/Amqp/ClientWebSocketTransport.cs
```diff
@@ -57,8 +58,8 @@ public override bool WriteAsync(TransportAsyncCallbackArgs args)
         {
             ThrowIfNotOpen();
 
-            Fx.AssertAndThrow(args.Buffer != null || args.ByteBufferList != null, "must have a buffer to write");
-            Fx.AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
+            Debug.Assert(args.Buffer != null || args.ByteBufferList != null, "must have a buffer to write");
```
drwill-ms, line 61: Deprecated Fx assertions, which were just calling Debug.Assert anyway.

iothub/service/src/Configurations/ConfigurationsClient.cs
```diff
@@ -19,18 +19,18 @@ namespace Microsoft.Azure.Devices
     /// <seealso href="https://docs.microsoft.com/azure/iot-hub/iot-hub-automatic-device-management"/>
     public class ConfigurationsClient
     {
-        private readonly string _hostName;
-        private readonly IotHubConnectionProperties _credentialProvider;
-        private readonly HttpClient _httpClient;
-        private readonly HttpRequestMessageFactory _httpRequestMessageFactory;
-
         private const string ConfigurationRequestUriFormat = "/configurations/{0}";
```
drwill-ms, line 22: fields order:
1. const
2. static readonly
3. static
4. readonly
5. instance

iothub/service/src/Configurations/ConfigurationsClient.cs
```diff
@@ -118,11 +119,13 @@ public virtual async Task<Configuration> GetAsync(string configurationId, Cancel
         {
             if (Logging.IsEnabled)
                 Logging.Enter(this, $"Getting configuration: {configurationId}", nameof(GetAsync));
+
+            Argument.AssertNotNullOrWhiteSpace(configurationId, nameof(configurationId));
```
drwill-ms, line 123: As mentioned before, check preconditions outside of the try/catch.

iothub/service/src/Configurations/Models/Configuration.cs
```diff
@@ -52,7 +42,7 @@ internal Configuration()
         /// The schema version of the configuration.
         /// </summary>
         [JsonProperty(PropertyName = "schemaVersion", NullValueHandling = NullValueHandling.Ignore)]
-        public string SchemaVersion { get; }
+        public string SchemaVersion { get; } = "1.0";
```
drwill-ms, line 45: Initialize inline

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -343,30 +341,46 @@ private static Uri GetDigitalTwinCommandRequestUri(string digitalTwinId, string
...
 
         // Root level commands and component level commands append the connect and read timeout values as query string values such as:
         // /digitaltwins/myDigitalTwin/commands/myCommand?api-version="2021-04-12"&connectTimeoutInSeconds=10&readTimeoutInSeconds=20
         private static string BuildCommandRequestQueryStringParameters(InvokeDigitalTwinCommandOptions requestOptions)
         {
             string queryStringParameters = "";
+
             if (requestOptions?.ConnectTimeout != null)
             {
-                queryStringParameters += string.Format($"&connectTimeoutInSeconds={(int)requestOptions.ConnectTimeout.Value.TotalSeconds}");
```
drwill-ms, line 364: No reason to call format

iothub/service/src/Feedback/MessageFeedbackProcessorClient.cs
```diff
@@ -240,7 +246,7 @@ private void OnConnectionClosed(object sender, EventArgs e)
                 Exception exceptionToLog = errorContext.IOException != null ? errorContext.IOException : errorContext.IotHubException;
 
                 if (Logging.IsEnabled)
-                    Logging.Error(this, $"{nameof(sender) + '.' + nameof(OnConnectionClosed)} threw an exception: {exceptionToLog}", nameof(OnConnectionClosed));
+                    Logging.Error(this, $"{nameof(sender)}.{nameof(OnConnectionClosed)} threw an exception: {exceptionToLog}", nameof(OnConnectionClosed));
```
drwill-ms, line 249: Readability

iothub/service/src/Http/HttpClientFactory.cs
```diff
@@ -30,7 +30,7 @@ internal class HttpClientFactory
         //
         // This default value is consistent with the default value used in Azure.Core
         // https://github.com/Azure/azure-sdk-for-net/blob/7e3cf643977591e9041f4c628fd4d28237398e0b/sdk/core/Azure.Core/src/Pipeline/ServicePointHelpers.cs#L29
-        private static readonly TimeSpan DefaultConnectionLeaseTimeout = TimeSpan.FromMinutes(5);
+        private static readonly TimeSpan s_defaultConnectionLeaseTimeout = TimeSpan.FromMinutes(5);
```
drwill-ms, line 33: Tim, this is still used in service point manager.

iothub/service/src/Http/HttpClientFactory.cs
```diff
@@ -73,9 +72,9 @@ internal static HttpClient Create(string hostName, IotHubServiceClientOptions op
 
             httpMessageHandler.MaxConnectionsPerServer = DefaultMaxConnectionsPerServer;
             ServicePoint servicePoint = ServicePointManager.FindServicePoint(httpsEndpoint);
-            servicePoint.ConnectionLeaseTimeout = DefaultConnectionLeaseTimeout.Milliseconds;
+            servicePoint.ConnectionLeaseTimeout = s_defaultConnectionLeaseTimeout.Milliseconds;
```
drwill-ms, line 75: Tim, down here.

drwill-ms: Will this go away?

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -120,14 +121,14 @@ private static string EscapeETag(string eTag)
         {
             var escapedETagBuilder = new StringBuilder();
 
-            if (!eTag.ToString().StartsWith("\"", StringComparison.OrdinalIgnoreCase))
```
drwill-ms, line 123: eTag was already a string.

iothub/service/src/Jobs/Models/DeviceJobStatistics.cs
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// The job counts, e.g., number of failed/succeeded devices.
+    /// </summary>
+    public class DeviceJobStatistics
```
drwill-ms, line 11: Moved to its own file.

iothub/service/src/Jobs/Models/JobQueryOptions.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Specifies the options associated with job queries.
+    /// </summary>
+    public class JobQueryOptions : QueryOptions
```
drwill-ms, line 9: Inherits from the base and adds 2 more optional values, so we go from 3 to 1 params.

iothub/service/src/Jobs/Models/ScheduledJob.cs
```diff
@@ -28,72 +53,55 @@ public class ScheduledJob
...
-        /// <summary>
-        /// [Required] The type of job to execute.
+        /// Max execution time.
         /// </summary>
-        [JsonProperty(PropertyName = "jobType", NullValueHandling = NullValueHandling.Ignore)]
-        [JsonConverter(typeof(StringEnumConverter))]
-        public JobType JobType { get; internal set; }
+        /// <remarks>The precision on this is in seconds.</remarks>
+        [JsonIgnore]
+        public TimeSpan MaxExecutionTime { get; set; }
```
drwill-ms, line 96: Present properties with TimeSpan, instead of seconds. Serializable property is now internal (just below).

iothub/service/src/Query/QueryClient.cs
```diff
@@ -20,17 +17,16 @@ namespace Microsoft.Azure.Devices
...
         private const string ContinuationTokenHeader = "x-ms-continuation";
         private const string PageSizeHeader = "x-ms-max-item-count";
-        private const string DevicesQueryUriFormat = "/devices/query";
-        private const string JobsQueryFormat = "/jobs/v2/query";
 
-        private string _hostName;
-        private IotHubConnectionProperties _credentialProvider;
-        private HttpClient _httpClient;
-        private HttpRequestMessageFactory _httpRequestMessageFactory;
+        private static readonly Uri s_jobsQueryFormat = new("/jobs/v2/query", UriKind.Relative);
```
drwill-ms, line 23: Cache URIs for non-dynamic paths.

iothub/service/src/Utilities/Fx.cs
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.ObjectModel;
+using System.Reflection;
+
+namespace Microsoft.Azure.Devices.Common
+{
+    internal static class Fx
```
drwill-ms, line 10: Moved from another dir, with everything but this method removed.

@brycewang-microsoft, this should be on your list of things to update for the exceptions work.

iothub/service/src/Configurations/ConfigurationsClient.cs
```diff
@@ -309,14 +319,17 @@ public virtual async Task DeleteAsync(Configuration configuration, bool onlyIfUn
         {
             if (Logging.IsEnabled)
                 Logging.Enter(this, $"Deleting configuration: {configuration?.Id} - only if changed: {onlyIfUnchanged}", nameof(DeleteAsync));
+
+            Argument.AssertNotNull(configuration, nameof(configuration));
+            if (string.IsNullOrWhiteSpace(configuration.ETag.ToString()) && onlyIfUnchanged)
```
timtay-microsoft, line 324: We talked about removing this validation, so you can remove this

iothub/service/src/Http/HttpClientFactory.cs
timtay-microsoft: Yes, that is by design. We use the service point manager to set this timeout so that users avoid problems around failover scenarios. Did you want to change this in some way?

iothub/service/src/Http/HttpClientFactory.cs
timtay-microsoft: No, this was the fix we provided so that a single http client instance would eventually start making requests to the correct endpoint if failover occurs.

drwill-ms: I thought we were going to let customers give us an HttpClient instance where this would already have been set. But I know we prefer to have smart defaults, so if we're capable of doing this, then I'm fine.

drwill-ms: Obviously, this one has nothing to do with the HttpClient instance, we just set it in the same method. I thought there was some problem netstandard2.0 where we couldn't set this value, though. Memory is hazy.

iothub/service/src/Registry/DevicesClient.cs
```diff
@@ -1078,58 +881,65 @@ public virtual async Task<ServiceStatistics> GetServiceStatisticsAsync(Cancellat
...
-            return new Uri(string.Format(CultureInfo.InvariantCulture, DeviceRequestUriFormat, string.Empty), UriKind.Relative);
+            return new Uri(
+                string.Format(
+                    CultureInfo.InvariantCulture,
+                    DeviceRequestUriFormat,
+                    string.Empty),
+                UriKind.Relative);
         }
 
+        // TODO: why no references?
```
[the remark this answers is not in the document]
drwill-ms: It is unreferenced.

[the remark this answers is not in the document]
drwill-ms: The original functions for starting an import/export job used this path: /$admin/{0} where {0} is exportRegistry or importRegistry. I can't find that URI in the swagger.



For some reason, the newer API call (the one I added) uses what is in the swagger of /jobs/create.

iothub/service/src/Http/HttpClientFactory.cs
timtay-microsoft: > I thought we were going to let customers give us an HttpClient instance where this would already have been set. But I know we prefer to have smart defaults, so if we're capable of doing this, then I'm fine.

This setting fell under the category of a "smart default" last we talked about it. We agreed that our client's should be resilient to failover, and this same setting is applied to the default http client in track 2

> Obviously, this one has nothing to do with the HttpClient instance, we just set it in the same method. I thought there was some problem netstandard2.0 where we couldn't set this value, though. Memory is hazy.

The problem was that we had different HttpClientHandlers depending on the .NET target framework. The track 2 logic [here](https://github.com/Azure/azure-sdk-for-net/blob/cc96e605365ebfd30f9dd00a6dff47f89feb82d8/sdk/core/Azure.Core/src/Pipeline/ServicePointHelpers.cs#L87) demonstrates the problem.

timtay-microsoft (approved):

iothub/service/src/Registry/DevicesClient.cs
```diff
@@ -1147,7 +950,7 @@ private static IEnumerable<ExportImportDevice> GenerateExportImportDeviceListFor
                     case ImportMode.UpdateIfMatchETag:
                         if (string.IsNullOrWhiteSpace(device.ETag.ToString()))
```
drwill-ms, line 951: @timtay-microsoft @abhipsaMisra what do you think about removing this (and delete if match etag) validation is removed (like we have for non-bulk operations)?

To me, the downside here is that they won't get immediate errors, rather they have to check the errors blob, which is much more costly.

iothub/service/src/Messaging/MessagingClient.cs
```diff
@@ -80,7 +62,24 @@ protected MessagingClient()
         }
 
         /// <summary>
-        /// Open this instance. Must be done before any cloud-to-device messages can be sent.
+        /// The callback to be executed when the connection is lost.
```
patilsnr, line 65: When would we uncomment this code? Could maybe benefit with an explanation of that scenario -- kind of subjective though.

iothub/service/src/Feedback/Models/FeedbackRecord.cs
```diff
@@ -43,7 +43,7 @@ public class FeedbackRecord
         /// Date and time that indicates when the outcome of the message happened in UTC (for example, the hub received the feedback message or the original message expired).
         /// </summary>
         [JsonProperty(PropertyName = "enqueuedTimeUtc", DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public DateTime EnqueuedTimeUtc { get; internal set; }
+        public DateTimeOffset EnqueuedTimeUtc { get; internal set; }
```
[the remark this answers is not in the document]
drwill-ms: Not a nit. Thanks for catching this.

iothub/service/src/Registry/Models/JobProperties.cs
```diff
@@ -154,11 +138,21 @@ public JobProperties()
         [JsonIgnore]
         public bool IsFinished => s_finishedStates.Contains(Status);
 
+        /// <summary>
+        /// System generated. Ignored at creation.
+        /// Represents the percentage of completion.
+        /// </summary>
+        /// <remarks>The service doesn't actually seem to set this, so not exposing it.</remarks>
```
abhipsaMisra, line 145: 👍🏽

abhipsaMisra (approved):

iothub/service/src/Registry/DevicesClient.cs
timtay-microsoft: I'm fine with keeping this validation here for the reasons you outlined above. In the non-bulk APIs, it is easy enough for the customer to handle a 412

abhipsaMisra: +1 to that. No point in increasing an additional out-of-the-way step for the user to discover the error.

iothub/service/src/Registry/Models/JobProperties.cs
```diff
@@ -20,14 +20,61 @@ public class JobProperties
...
+        /// <list type="bullet">
+        /// <item>InputBlobName, to override the default.</item>
+        /// <item>StorageAuthenticationType, if using identity-based instead of SAS in InputBlobContainerUri.</item>
+        /// <item>ManagedIdentity, if the IoT hub is not configured with a managed identity and the service app will specify
+        /// which identity has permissions to the storage account.</item>
+        /// <item>IncludeConfigurations, if the import job should also include configurations.</item>
+        /// <item>ConfigurationsBlobName, if the import job includes configurations and to override the default blob name.</item>
+        /// </list>
+        /// </remarks>
+        public JobProperties(Uri inputBlobContainerUri, Uri outputBlobContainerUri)
```
drwill-ms, line 42: Instead of factory methods, we have 2 ctors, one for import and one for export. Comments indicate which are for which, as well as what additional properties are relevant, but optional.

iothub/service/src/Registry/Models/JobProperties.cs
```diff
@@ -77,17 +124,17 @@ public class JobProperties
         /// <summary>
         /// URI to a blob container that contains registry data to sync. Including a SAS token is dependent on the <see cref="StorageAuthenticationType" /> property.
         /// </summary>
-        /// <remarks>
-        /// For import job, if there are errors they will be written to OutputBlobContainerUri to a file called "importerrors.log"
-        /// </remarks>
         [JsonProperty(PropertyName = "inputBlobContainerUri", NullValueHandling = NullValueHandling.Ignore)]
-        public string InputBlobContainerUri { get; set; }
+        public Uri InputBlobContainerUri { get; set; }
```
drwill-ms, line 128: Works as a URI.

e2e/test/iothub/service/ImportDevicesTests.cs
```diff
@@ -191,14 +190,14 @@ private static async Task UploadFileAndConfirmAsync(StorageContainer storageCont
...
         {
-            JobProperties jobProperties = JobProperties.CreateForImportJob(
-                containerUri,
-                containerUri,
-                devicesFileName,
-                storageAuthenticationType,
-                identity);
-            jobProperties.ConfigurationsBlobName = configsFileName;
-            jobProperties.IncludeConfigurations = true;
+            var jobProperties = new JobProperties(containerUri, containerUri)
```
abhipsaMisra, line 193: we're passing in containerUri twice to the constructor?

abhipsaMisra: ah, I see your comment below about these being input and output container URIs.

iothub/service/src/Registry/Models/JobProperties.cs
abhipsaMisra, line 42: no params for doc comments?

drwill-ms: Adding, thanks.

iothub/service/src/Registry/DevicesClient.cs
```diff
@@ -226,6 +226,7 @@ public virtual async Task DeleteAsync(string deviceId, CancellationToken cancell
             try
             {
                 using HttpRequestMessage request = _httpRequestMessageFactory.CreateRequest(HttpMethod.Delete, GetRequestUri(deviceId), _credentialProvider);
+                HttpMessageHelper.ConditionallyInsertETag(request, ETag.All, false);
```
drwill-ms, line 229: This call fails if I don't include this.

Interestingly enough, I can't pass null, as ETag is not nullable. So the code in this method which checks for null is wrong (or we need to make the param nullable.

Thoughts @timtay-microsoft ?

e2e/test/iothub/service/MessageFeedbackReceiverE2ETest.cs
```diff
@@ -31,40 +31,51 @@ public async Task MessageFeedbackReceiver_Operation(IotHubTransportProtocol prot
...
-            using IotHubDeviceClient deviceClient = testDevice.CreateDeviceClient(new IotHubClientOptions(new IotHubClientAmqpSettings()));
-            await deviceClient.OpenAsync().ConfigureAwait(false);
-            Client.Message receivedMessage = await deviceClient.ReceiveMessageAsync().ConfigureAwait(false);
-            await deviceClient.CompleteMessageAsync(receivedMessage.LockToken).ConfigureAwait(false);
 
-            var timer = Stopwatch.StartNew();
-            while (!messagedFeedbackReceived && timer.ElapsedMilliseconds < 200000)
+            try
             {
-                await Task.Delay(800);
```
drwill-ms, line 54: Rather than looping, we can just await a task completion source.

e2e/test/iothub/service/MessageFeedbackReceiverE2ETest.cs
```diff
@@ -31,40 +31,51 @@ public async Task MessageFeedbackReceiver_Operation(IotHubTransportProtocol prot
...
+                {
+                    Ack = DeliveryAcknowledgement.Full,
+                    MessageId = Guid.NewGuid().ToString(),
+                };
 
-            messagedFeedbackReceived.Should().BeTrue("Timed out waiting to receive message feedback.");
+                var messageReceived = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+                serviceClient.MessageFeedbackProcessor.MessageFeedbackProcessor = (FeedbackBatch feedback) =>
+                {
+                    if (feedback.Records.Any(x => x.OriginalMessageId == message.MessageId))
```
drwill-ms, line 49: Validate we got the message that was sent.

iothub/service/src/Feedback/MessageFeedbackProcessorClient.cs
```diff
@@ -196,7 +196,7 @@ private async void OnFeedbackMessageReceivedAsync(AmqpMessage amqpMessage)
 
                         var feedbackBatch = new FeedbackBatch
                         {
-                            EnqueuedOnUtc = (DateTimeOffset)amqpMessage.MessageAnnotations.Map[MessageSystemPropertyNames.EnqueuedOn],
+                            EnqueuedOnUtc = (DateTime)amqpMessage.MessageAnnotations.Map[MessageSystemPropertyNames.EnqueuedOn],
```
drwill-ms, line 199: DateTimeOffset was not supported, so I rolled that back.

patilsnr (approved):
Everything looks fine to me! Also glad that Tim took a look though -- maybe wait for another's approval as well.

iothub/service/src/Registry/DevicesClient.cs
timtay-microsoft: For this API, there is never an ETag since the user is only giving us the device Id. There is no reason to call ```HttpMessageHelper.ConditionallyInsertETag``` because of that.

timtay-microsoft: Why are you replacing what was here? Calling the overload with a new ```Device``` instance with an empty etag seemed fine to me

drwill-ms: If I don't make that call, it fails.

drwill-ms: Because the device one ran through extra code this overload didn't need.

drwill-ms: Except, apparently, adding the ETag.

PR #2688, opened 2022-09-13
Implement single listener for all callback APIs

abhipsaMisra:
- Remove the method name specific direct method API
- Remove input name specific module message receive API

### DIFF
--- a/iothub/device/src/IotHubBaseClient.cs
+++ b/iothub/device/src/IotHubBaseClient.cs
@@ -30,9 +30,7 @@ namespace Microsoft.Azure.Devices.Client
 
         // Method callback information
         private bool _isDeviceMethodEnabled;
-
         private volatile Tuple<Func<DirectMethodRequest, object, Task<DirectMethodResponse>>, object> _deviceDefaultMethodCallback;
-        private readonly Dictionary<string, Tuple<Func<DirectMethodRequest, object, Task<DirectMethodResponse>>, object>> _deviceMethods = new();
 
         // Twin property update request callback information
         private bool _twinPatchSubscribedWithService;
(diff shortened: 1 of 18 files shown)


iothub/device/src/IotHubBaseClient.cs
```diff
@@ -587,41 +533,9 @@ internal async Task OnMethodCalledAsync(DirectMethodRequest directMethodRequest)
                 return;
             }
 
-            Tuple<Func<DirectMethodRequest, object, Task<DirectMethodResponse>>, object> callbackContextPair = null;
             DirectMethodResponse directMethodResponse = null;
 
-            await _methodsSemaphore.WaitAsync().ConfigureAwait(false);
```
abhipsaMisra, line 593: The local variable saving a reference to the user supplied callback, I'm not seeing much value in having this semaphore here. Let me know if you feel there's a reason to have this semaphore here.
If we really wanted to ensure that there are no rtace conditions between a user setting this callback and us invoking it, we'll need to create a deep-copy of this callback-context pair.

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -672,7 +672,7 @@ public async Task<AmqpIotOutcome> SendEventAsync(Message message, CancellationTo
 
         public void OnEventsReceived(Message message)
         {
-            _onModuleMessageReceivedCallback?.Invoke(message.InputName, message);
+            _onModuleMessageReceivedCallback?.Invoke(message);
```
abhipsaMisra, line 675: this callback no longer needs to input name to be parsed out in advance

drwill-ms (approved):

PR #2696, opened 2022-09-15
Add throttling and traffic shaping comment to sample, and avoid deadlock

drwill-ms:


iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -169,6 +169,11 @@ private async void ConnectionStatusChangeHandlerAsync(ConnectionStatus status, C
 
                     // Call GetTwinAndDetectChangesAsync() to retrieve twin values from the server once the connection status changes into Connected.
                     // This can get back "lost" twin updates in a device reconnection from status like Disconnected_Retrying or Disconnected.
+                    //
+                    // Howevever, considering how a fleet of devices connected to a hub may behave together, one must consider the implication of performing
+                    // work on a device when it comes online. If all the devices go offline and then come online at the same time (for example,
```
drwill-ms, line 174: How do you feel about this blurb?

samples/helpers/ColorConsoleLogger/ColorConsoleLogger.cs
```diff
@@ -57,23 +57,20 @@ public bool IsEnabled(LogLevel logLevel)
         /// <param name="formatter">The formatter to be used for formatting the log message.</param>
         public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
         {
-            lock (s_lockObject)
```
drwill-ms, line 60: We were locking too early.

Also, the sample was stopping, and it looked like a deadlock.

I moved the lock lower and cut out one feature (multiple colors on the same line) and have not witnessed the deadlock since.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
massand, line 174:
```suggestion
                    // work (e.g. get twin) on a device when it comes online. If all the devices go offline and then come online at the same time (for example,
```

samples/helpers/ColorConsoleLogger/ColorConsoleLogger.cs
patilsnr: I don't have much knowledge on the ColorConsoleLogger so this question might not make sense, but would there be a way to move the operation somewhere else instead of getting rid of it altogether?

patilsnr (approved):

samples/helpers/ColorConsoleLogger/ColorConsoleLogger.cs
andyk-ms: I am uncertain why moving the lock change the behavior, but if that is the case, one of the console logging step may have issue during testing?

andyk-ms (approved):

samples/helpers/ColorConsoleLogger/ColorConsoleLogger.cs
drwill-ms: No chance to move elsewhere. We aren't getting rid of the logging, just the distinct color for the timestamp prefix on each message.

PR #2697, opened 2022-09-15
Update payload type of direct methods for device and service clients

tmahmood-microsoft:


iothub/device/src/DirectMethod/DirectMethodResponse.cs
```diff
@@ -26,8 +27,11 @@ public DirectMethodResponse()
         /// <summary>
         /// The optional direct method payload.
         /// </summary>
+        /// <remarks>
+        /// The payload can be null or Primitive type (e.g., String, Int)/Array/List/Map/custom type.
```
drwill-ms, line 31:
```suggestion
        /// The payload can be null or primitive type (e.g., string, int/array/list/dictionary/custom type).
```

iothub/device/tests/CustomType.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    internal class CustomType
+    {
+        [JsonProperty]
+        internal string stringAttri;
```
drwill-ms, line 15: Follow .NET casing standards, and use attributes to fix serialization casing.

```suggestion
        [JsonProperty("stringAtri")]
        internal string StringAttri;
```

iothub/device/tests/CustomType.cs
```diff
@@ -0,0 +1,45 @@
...
+            this.NestedCustomType = nestedCustomType;
+        }
+
+        public override string ToString()
+        {
+            Dictionary<string, object> dict = new Dictionary<string, object>();
+            dict.Add("stringAttri", stringAttri);
+            dict.Add("intAttri", intAttri);
+            dict.Add("boolAttri", boolAttri);
+            dict.Add("NestedCustomType", NestedCustomType);
```
drwill-ms, line 40:
```suggestion
            var dict = new Dictionary<string, object>
            {
                { "stringAttri", stringAttri },
                { "intAttri", intAttri },
                { "boolAttri", boolAttri },
                { "NestedCustomType", NestedCustomType },
            };
```

iothub/service/src/DirectMethod/Models/DirectMethodResponse.cs
```diff
@@ -18,13 +18,13 @@ public class DirectMethodResponse
...
         /// </summary>
+        /// <remarks>
+        /// The payload can be null or Primitive type (e.g., String, Int)/Array/List/Map/custom type.
+        /// </remarks>
         [JsonIgnore]
-        public string Payload
-        {
-            get => (string)JsonPayload.Value;
-        }
+        public object Payload { get; set; }
```
drwill-ms, line 27: We should be looking at the code @abhipsaMisra has in the PnP preview branch and copy what we did there. We finely honed that API and wished to bring it here.

You can merge this in, but we're going to redo it, so you might want to look at that first and get us closer to the mark, at least.

drwill-ms (approved):

iothub/service/src/DirectMethod/Models/DirectMethodResponse.cs
tmahmood-microsoft: @drwill-ms  you mean this class: https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/pnpApi/iothub/service/src/DirectMethod/CloudToDeviceMethodResult.cs

iothub/device/tests/CustomType.cs
```diff
@@ -0,0 +1,45 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    internal class CustomType
```
timtay-microsoft, line 12: Missing documentation on what this class is and what it is used for

iothub/device/tests/NestedCustomType.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    internal class NestedCustomType
```
timtay-microsoft, line 10: Missing comments on what this class is and where it is used

iothub/service/src/DirectMethod/Models/DirectMethodRequest.cs
```diff
@@ -57,41 +67,42 @@ public DirectMethodRequest()
...
+        /// Get the serialized JSON payload. May be null or empty.
+        /// </summary>
+        [JsonIgnore]
+        public string PayloadAsJsonString { get; internal set; }
+
+        /// <summary>
+        /// Get the JSON payload in JRaw type.
+        /// </summary>
+        [JsonProperty("payload")]
+        public JRaw JsonPayload { get; internal set; }
```
timtay-microsoft, line 98: If possible, I'd consider making this internal. If we ever allow users to pass in their own JSON serailization library, then we'd have trouble constructing and exposing this Newtonsoft type

e2e/test/helpers/TestDeviceCallbackHandler.cs
```diff
@@ -62,7 +62,7 @@ public async Task SetDeviceReceiveMethodAsync(string methodName, string deviceRe
                         var response = new Client.DirectMethodResponse()
                         {
                             Status = 200,
-                            Payload = Encoding.UTF8.GetBytes(deviceResponseJson)
```
timtay-microsoft, line 65: I'd like to see a few e2e tests that demonstrate sending and receiving custom JSON objects (type T) so that we know for sure there isn't any unexpected escaping of payloads that is done by the service

iothub/device/src/DirectMethod/DirectMethodResponse.cs
patilsnr: If we're providing examples of primitive types, why mention custom types at all? Also, should we consider adding a link to the [built-in C# types](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types)?

iothub/service/src/DirectMethod/Models/DirectMethodResponse.cs
drwill-ms: Yes, for this class, but there are several other changes and associated classes.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -799,7 +799,7 @@ public override async Task SendMethodResponseAsync(DirectMethodResponse methodRe
             cancellationToken.ThrowIfCancellationRequested();
             EnsureValidState();
 
-            var message = new Message(methodResponse.Payload)
+            var message = new Message((byte[])methodResponse.Payload)
```
andyk-ms, line 802: Do we need to worry about exception on explicit type casting?

iothub/device/src/DirectMethod/DirectMethodResponse.cs
tmahmood-microsoft: I think adding a link a to built-in c# types would be an overkill.

We do need to mention custom types as its an important feature to support it

iothub/service/src/DirectMethod/Models/DirectMethodResponse.cs
tmahmood-microsoft: Will discuss the exact changes with @abhipsaMisra when she is back on Monday

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
tmahmood-microsoft: Message constructor expects byte[] as the parameter, In the latest commit, I have also updated the code to explicitly convert the payload to byte[].

patilsnr (approved):
LGTM, but I'd also wait for someone more experienced to weigh in as well.

andyk-ms (approved):

timtay-microsoft (approved):

PR #2699, opened 2022-09-19
[minor] Updated service client to use Azure.ETag

tmahmood-microsoft:


iothub/service/src/DigitalTwin/Models/DigitalTwinGetResponse.cs
```diff
@@ -17,7 +19,7 @@ public class DigitalTwinGetResponse<T>
         internal DigitalTwinGetResponse(T digitalTwin, string eTag = default)
         {
             DigitalTwin = digitalTwin;
-            ETag = eTag;
+            ETag = new ETag(eTag);
```
drwill-ms, line 22: Should the ctor be modified to take an ETag instead of a string?

iothub/service/src/DigitalTwin/Models/DigitalTwinUpdateResponse.cs
```diff
@@ -15,7 +17,7 @@ public class DigitalTwinUpdateResponse
         /// <param name="location">URI of the digital twin.</param>
         internal DigitalTwinUpdateResponse(string eTag = default, string location = default)
         {
-            ETag = eTag;
+            ETag = new ETag(eTag);
```
drwill-ms, line 20: same

iothub/service/src/DigitalTwin/Models/DigitalTwinGetResponse.cs
tmahmood-microsoft: Updated ctor accordingly.

timtay-microsoft (approved):

PR #2707, opened 2022-09-20
Revamp exceptions thrown by the hub service- part 2

brycewang-microsoft:
+ stop throwing `TimeoutException`
+ let `TaskCanceledException` bubble up to users
+ revamp the `IsTransient` property in `IotHubServiceException`


iothub/device/src/Transport/AmqpIot/AmqpIotExceptionAdapter.cs
```diff
@@ -40,9 +40,7 @@ internal static Exception ConvertToIotHubException(Exception exception)
         internal static Exception ConvertToIotHubException(Exception exception, AmqpObject source)
         {
             Exception ex = ConvertToIotHubException(exception);
-            if (source.IsClosing() &&
-                (ex is InvalidOperationException
-                || ex is OperationCanceledException))
+            if (source.IsClosing() && ex is InvalidOperationException)
```
brycewang-microsoft, line 43: missed this from a previous PR of unwrapping OperationCanceledException from other exceptions in the hub device client

iothub/service/src/Amqp/ClientWebSocketTransport.cs
```diff
@@ -224,9 +224,9 @@ private async Task<int> ReadImplAsync(TransportAsyncCallbackArgs args)
             {
                 throw new IOException(httpListenerException.Message, httpListenerException);
             }
-            catch (TaskCanceledException taskCanceledException)
+            catch (TaskCanceledException)
             {
-                throw new TimeoutException(taskCanceledException.Message, taskCanceledException);
+                throw;
```
timtay-microsoft, line 229: Why catch the exception just to throw it again?

iothub/service/src/Amqp/ClientWebSocketTransport.cs
```diff
@@ -191,9 +191,9 @@ await _webSocket
...
```
timtay-microsoft, line 196: Same question here

iothub/service/src/Amqp/ClientWebSocketTransport.cs
brycewang-microsoft: Ah good catch, I should simply stop catching it here.

iothub/service/src/Exceptions/IotHubServiceException.cs
```diff
@@ -79,5 +78,21 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
             info.AddValue(IsTransientValueSerializationStoreName, IsTransient);
             info.AddValue(TrackingIdSerializationStoreName, TrackingId);
         }
+
+        private bool DetermineIfTransient(HttpStatusCode statusCode, IotHubErrorCode errorCode)
+        {
+            switch (errorCode)
```
timtay-microsoft, line 84: I'd say we should label any 500 level error (ServerError, ServiceUnavailable) as transient, and I'd argue "DeviceNotOnline" is retriable too since the device's state will likely return to online soon if the customer has a retry policy in place

timtay-microsoft (approved):

iothub/service/src/Exceptions/IotHubServiceException.cs
brycewang-microsoft: > "DeviceNotOnline" is retriable too since the device's state will likely return to online soon if the customer has a retry policy in place

"DeviceNotOnline" can also happen when the direct method callback isn't registered on the device, which seems not able to recover via the custom retry policy. Should we still consider this as retriable?

drwill-ms: Yeah, probably retriable for DeviceNotOnline.

patilsnr (approved):

iothub/device/src/Transport/AmqpIot/AmqpIotExceptionAdapter.cs
```diff
@@ -40,9 +40,7 @@ internal static Exception ConvertToIotHubException(Exception exception)
         internal static Exception ConvertToIotHubException(Exception exception, AmqpObject source)
         {
             Exception ex = ConvertToIotHubException(exception);
```
drwill-ms, line 42: What is this doing? Did it convert it?

iothub/device/src/Transport/AmqpIot/AmqpIotExceptionAdapter.cs
```diff
@@ -40,9 +40,7 @@ internal static Exception ConvertToIotHubException(Exception exception)
...
-            if (source.IsClosing() &&
-                (ex is InvalidOperationException
-                || ex is OperationCanceledException))
+            if (source.IsClosing() && ex is InvalidOperationException)
             {
                 return new IotHubClientException("Amqp resource is disconnected.", null, IotHubStatusCode.NetworkErrors);
```
drwill-ms, line 45: Upper case AMQP, and also we should set the inner exception.
```suggestion
                return new IotHubClientException("AMQP resource is disconnected.", null, IotHubStatusCode.NetworkErrors);
```

drwill-ms (approved):
Optimistic sign-off.

iothub/device/src/Transport/AmqpIot/AmqpIotExceptionAdapter.cs
brycewang-microsoft: I just left this method as-is. With another glance, it can be improved though.

PR #2711, opened 2022-09-21
Return updated twin version from reported property update call

timtay-microsoft:
The service returns this new version after a successful reported properties update, but the SDK wasn't bubbling it up to the user in v1

### DIFF
--- a/iothub/device/src/IotHubBaseClient.cs
+++ b/iothub/device/src/IotHubBaseClient.cs
@@ -392,7 +392,7 @@ namespace Microsoft.Azure.Devices.Client
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <returns>The new version of the updated twin if the update was successful.</returns>
         /// <exception cref="OperationCanceledException">Thrown when the operation has been canceled.</exception>
-        public async Task<int> UpdateReportedPropertiesAsync(TwinCollection reportedProperties, CancellationToken cancellationToken = default)
+        public async Task<long> UpdateReportedPropertiesAsync(TwinCollection reportedProperties, CancellationToken cancellationToken = default)
         {
             Argument.AssertNotNull(reportedProperties, nameof(reportedProperties));
             cancellationToken.ThrowIfCancellationRequested();
--- a/iothub/device/src/IotHubBaseClient.cs
+++ b/iothub/device/src/IotHubBaseClient.cs
@@ -11,6 +11,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client.Exceptions;
 using Microsoft.Azure.Devices.Client.Transport;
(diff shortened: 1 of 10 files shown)


iothub/device/src/Transport/AmqpIot/AmqpIotReceivingLink.cs
```diff
@@ -292,9 +292,11 @@ private void OnTwinChangesReceived(AmqpMessage amqpMessage)
                     else if (correlationId.StartsWith(AmqpTwinMessageType.Patch.ToString(), StringComparison.OrdinalIgnoreCase))
                     {
                         // This can be used to coorelate success response with updating reported properties
-                        // However currently we do not have it as request response style implementation
                         if (Logging.IsEnabled)
                             Logging.Info("Updated twin reported properties successfully", nameof(OnTwinChangesReceived));
+
+                        twin = new Twin();
+                        twin.Version = Convert.ToInt32(amqpMessage.MessageAnnotations.Map["version"]);
```
drwill-ms, line 299:
```suggestion
                        twin = new Twin
                        {
                            Version = Convert.ToInt32(amqpMessage.MessageAnnotations.Map["version"]),
                        };
```

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -388,14 +390,15 @@ public async Task<Twin> GetTwinAsync(CancellationToken cancellationToken = defau
         /// </summary>
         /// <param name="reportedProperties">Reported properties to push</param>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <returns>The new version of the updated twin if the update was successful.</returns>
         /// <exception cref="OperationCanceledException">Thrown when the operation has been canceled.</exception>
-        public async Task UpdateReportedPropertiesAsync(TwinCollection reportedProperties, CancellationToken cancellationToken = default)
+        public async Task<int> UpdateReportedPropertiesAsync(TwinCollection reportedProperties, CancellationToken cancellationToken = default)
```
abhipsaMisra, line 395: Our pnp work chose to expose this as a long, although I forget why. It is possible the Version in a Twin is saved as a long.
Also, even though we chose to expose this with the request Id for pnp, I think the Version alone is what is going to be useful to the user.

abhipsaMisra (approved):

iothub/device/src/IotHubBaseClient.cs
timtay-microsoft: I'll try to dig some more into if this value should be a long or an int

timtay-microsoft: Currently, the user doesn't get to pick the request id when they make this request, so I don't think there is any value in returning the request id to the user here. Even if we did allow them to pick the request id, the service returns the same request id in its response message so there is no reason to expose it then either.

PR #2713, opened 2022-09-21
Split ScheduledJobs into relevant classes

tmahmood-microsoft:


iothub/service/src/Jobs/Models/CloudToDeviceMethodScheduledJob.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Contains properties of a cloud-to-device method scheduled job.
+    /// </summary>
+    public class CloudToDeviceMethodScheduledJob : ScheduledJob
+    {
+        /// <summary>
+        /// [Required] The method type and parameters.
```
drwill-ms, line 14: If it is required, we should add it to a ctor.

iothub/service/src/Jobs/Models/ScheduledJob.cs
```diff
@@ -36,13 +36,6 @@ internal JobType AlternateJobType
             set => JobType = value;
         }
 
-        /// <summary>
```
drwill-ms, line 39: Should CraetedOnUtc also be pulled into the base class? It is part of the response, no?

iothub/service/src/Jobs/Models/TwinScheduledJob.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Contains properties of twin scheduled job.
+    /// </summary>
+    public class TwinScheduledJob
+    {
+        /// <summary>
+        /// [Required] The update twin tags and desired properties.
```
drwill-ms, line 14: same, if required put in ctor

iothub/service/src/Jobs/Models/IotHubScheduledJobResponse.cs
```diff
@@ -0,0 +1,83 @@
...
+        // as the created time value for this class and expose it either way as EndTimeUtc.
+        [JsonProperty(PropertyName = "endTime")]
+        internal DateTime? AlternateEndedOnUtc
+        {
+            get => EndedOnUtc;
+            set => EndedOnUtc = value;
+        }
+
+        /// <summary>
+        /// System generated. Ignored at creation.
```
drwill-ms, line 68: Update these comments.

drwill-ms (approved):

PR #2715, opened 2022-09-21
Fix DPS device samples

drwill-ms:


provisioning/device/samples/Getting Started/X509Sample/Parameters.cs
```diff
@@ -9,6 +9,12 @@
 
 namespace Microsoft.Azure.Devices.Provisioning.Client.Samples
 {
+    public enum Transport
```
drwill-ms, line 12: No longer have an SDK enum to use, so I created this enum to make the options clear. Each sample can have its own enum for the params file.

patilsnr (approved):

ngastelum-ms (approved):

PR #2717, opened 2022-09-21
Documenting exceptions thrown in DPS clients

tmahmood-microsoft:


provisioning/service/src/Authentication/SecurityHelper.cs
```diff
@@ -10,15 +10,8 @@ internal static class SecurityHelper
...
-            if (string.IsNullOrWhiteSpace(serviceHostName))
-            {
-                throw new ArgumentNullException(nameof(serviceHostName));
-            }
-
-            if (string.IsNullOrWhiteSpace(serviceName))
-            {
-                throw new ArgumentNullException(nameof(serviceName));
-            }
+            Argument.AssertNotNullOrWhiteSpace(serviceName, nameof(serviceName));
```
drwill-ms, line 13: The danger here is that this is not a public API. Which API calls this which then inherits an ArgumentException and ArgumentNullException? Will the parameter name match what parameters they have?

I'll be looking for the public API that calls this method and making sure it translates them appropriately, so they make sense to the user who called a public method.

provisioning/service/src/Authentication/ServiceAuthenticationWithSharedAccessPolicyKey.cs
```diff
@@ -52,21 +52,13 @@ public ServiceConnectionStringBuilder Populate(ServiceConnectionStringBuilder pr
 
         private void SetPolicyName(string policyName)
         {
-            if (string.IsNullOrWhiteSpace(policyName))
-            {
-                throw new ArgumentNullException(nameof(policyName));
-            }
-
+            Argument.AssertNotNullOrWhiteSpace(policyName, nameof(policyName));
```
drwill-ms, line 55: Same with this. A property that calls this doesn't have a parameter. The "parameter", if you will, is `value`.

In this case I'd suggest we stop validating these values, at least at this time. We can either let it go to the service who will then throw an error, or we can check the values at a point where it matters and throw an InvalidOperationException.

provisioning/service/src/Authentication/ServiceAuthenticationWithSharedAccessPolicyToken.cs
```diff
@@ -38,10 +38,7 @@ public string Token
 
         public ServiceConnectionStringBuilder Populate(ServiceConnectionStringBuilder provisioningConnectionStringBuilder)
         {
-            if (provisioningConnectionStringBuilder == null)
-            {
-                throw new ArgumentNullException(nameof(provisioningConnectionStringBuilder));
-            }
+            Argument.AssertNotNull(provisioningConnectionStringBuilder, nameof(provisioningConnectionStringBuilder));
```
drwill-ms, line 41: Same with this, unfortunately. Internal class, so not a public API. Where this might bubble up to a customer will make no sense to them.

I know you didn't change the behavior, just using a library we have, but it isn't the right long-term change we need.

provisioning/service/src/Enrollments/X509Attestation.cs
```diff
@@ -247,7 +247,8 @@ public static X509Attestation CreateFromRootCertificates(string primary, string
         /// </remarks>
         /// <param name="primary">the <c>string</c> with the primary certificate. It cannot be <c>null</c> or empty.</param>
         /// <returns>The new instance of the X509Attestation.</returns>
-        /// <exception cref="ArgumentException">if the provide primary certificate is invalid.</exception>
+        /// <exception cref="ArgumentException">if the provided primary certificate is invalid.</exception>
+        /// <exception cref="CryptographicException">if the provided certificate is invalid.</exception>
         public static X509Attestation CreateFromCAReferences(string primary)
         {
             if (string.IsNullOrWhiteSpace(primary))
```
drwill-ms, line 254: This is the first case we can actually change to Argument.AssertNotNullOrWhiteSpace.

provisioning/service/src/Enrollments/X509Attestation.cs
```diff
@@ -270,6 +271,7 @@ public static X509Attestation CreateFromCAReferences(string primary)
         /// <param name="secondary">the <c>string</c> with the secondary certificate. It can be <c>null</c> or empty (ignored).</param>
         /// <returns>The new instance of the X509Attestation.</returns>
         /// <exception cref="ArgumentException">if the provide primary certificate is invalid.</exception>
+        /// <exception cref="CryptographicException">if the one of the provided certificate is invalid.</exception>
         public static X509Attestation CreateFromCAReferences(string primary, string secondary)
         {
             if (string.IsNullOrWhiteSpace(primary))
```
drwill-ms, line 277: Same here.

provisioning/service/src/Enrollments/X509Attestation.cs
```diff
@@ -247,7 +247,8 @@ public static X509Attestation CreateFromRootCertificates(string primary, string
         /// </remarks>
         /// <param name="primary">the <c>string</c> with the primary certificate. It cannot be <c>null</c> or empty.</param>
         /// <returns>The new instance of the X509Attestation.</returns>
-        /// <exception cref="ArgumentException">if the provide primary certificate is invalid.</exception>
+        /// <exception cref="ArgumentException">if the provided primary certificate is invalid.</exception>
```
drwill-ms, line 250: Then add ArgumentNullException too.

provisioning/service/src/ProvisioningServiceClient.cs
```diff
@@ -74,7 +74,6 @@ public class ProvisioningServiceClient : IDisposable
         /// The Device Provisioning Service Client is created based on a <b>Provisioning Connection string</b>.
         /// Once you create a Device Provisioning Service on Azure, you can get the connection string on the Azure portal.
         /// </remarks>
-        ///
         /// <param name="connectionString">The <c>string</c> that cares the connection string of the Device Provisioning Service.</param>
         /// <param name="options"> The options that allow configuration of the provisioning service client instance during initialization.</param>
         /// <returns>The <c>ProvisioningServiceClient</c> with the new instance of this object.</returns>
```
drwill-ms, line 79: Add ArgumentNullException, because this helper method throws both.

provisioning/device/src/ProvisioningTransportHandler.cs
```diff
@@ -37,6 +38,7 @@ public abstract class ProvisioningTransportHandler : IDisposable
         /// <summary>
         /// Gets or sets the inner handler.
         /// </summary>
+        /// <exception cref="ArgumentNullException">When the provided inner handler is null.</exception>
         public ProvisioningTransportHandler InnerHandler
         {
             get => _innerHandler;
```
drwill-ms, line 44: Consider rewriting the setter as:

```csharp
set
{
    _innerHandler = value ?? throw new ArgumentNullException(nameof(value));
    Logging.Associate(this, value);
}
```

drwill-ms: Consider InvalidOperationException instead.

provisioning/service/src/Enrollments/X509Attestation.cs
```diff
@@ -290,6 +286,7 @@ public static X509Attestation CreateFromCAReferences(string primary, string seco
         ///     primary certificate.
         /// </remarks>
         /// <returns>The <see cref="X509CertificateInfo"/> with the returned certificate information. it can be <c>null</c>.</returns>
+        /// <exception cref="ArgumentException">if no valid certificate information is available.</exception>
```
drwill-ms, line 289:
```suggestion
        /// <exception cref="InvalidOperationException">If no valid certificate information was provided on initialization.</exception>
```

provisioning/service/src/Authentication/SecurityHelper.cs
```diff
@@ -2,14 +2,17 @@
...
 
 namespace Microsoft.Azure.Devices
 {
     internal static class SecurityHelper
     {
-        public static void ValidateServiceHostName(string serviceHostName, string serviceName)
+        internal static void ValidateServiceHostName(string serviceHostName, string serviceName)
         {
+            Debug.Assert(!string.IsNullOrWhiteSpace(serviceHostName));
+            Debug.Assert(!string.IsNullOrWhiteSpace(serviceName));
```
drwill-ms, line 15: These will need some comments otherwise it won't be clear why it failed. Comment is the second parameter.

provisioning/service/src/Authentication/ServiceAuthenticationWithSharedAccessPolicyKey.cs
```diff
@@ -38,25 +28,28 @@ public string Key
...
-            }
-
             provisioningConnectionStringBuilder.SharedAccessKey = Key;
             provisioningConnectionStringBuilder.SharedAccessKeyName = PolicyName;
             provisioningConnectionStringBuilder.SharedAccessSignature = null;
 
             return provisioningConnectionStringBuilder;
         }
 
+        internal ServiceAuthenticationWithSharedAccessPolicyKey(string policyName, string key)
```
drwill-ms, line 38: ctors above properties and other methods, even if internal

provisioning/service/src/Authentication/ServiceConnectionStringBuilder.cs
```diff
@@ -93,20 +79,11 @@ private void Parse(string serviceConnectionString)
 
         private void Validate()
         {
-            if (string.IsNullOrWhiteSpace(SharedAccessKeyName))
-            {
-                throw new ArgumentException("Should specify SharedAccessKeyName");
-            }
+            Debug.Assert(!string.IsNullOrWhiteSpace(SharedAccessKeyName), "Should specify SharedAccessKeyName");
```
drwill-ms, line 82: We can't change these to asserts if it doesn't give feedback to the customer.

My earlier point was, if you thrown an ArgumentException or ArgumentNullException in a private/internal method based on the parameter names it knows about, that won't make sense to an end user who called a public API that doesn't have an argument with that name.

So, you either need to move exception throwing to the public API (it catches exceptions thrown earlier and rethrows specifically on an argument it knows has failed (like connection string).

Or you throw `InvalidOperationException` from deep in the code, and simply document that in the public API.

Which is better? I think it depends. Generally, I'd like to be as specific as possible, so the customer has a very actionable error, so I'd favor the former when possible.

provisioning/service/src/Enrollments/QueryResultType.cs
```diff
@@ -35,7 +35,7 @@ public enum QueryResultType
         /// Enrollment group result type.
         /// </summary>
         /// <remarks>
-        /// The query result in a list of enrollmentGroup. You shall cast 
+        /// The query result in a list of enrollmentGroup. You shall cast
         /// the Objects in the items using <see cref="Service.EnrollmentGroup"/>.
```
drwill-ms, line 39:
```suggestion
        /// the objects in the items using <see cref="Service.EnrollmentGroup"/>.
```

provisioning/service/src/Enrollments/TpmAttestation.cs
```diff
@@ -25,15 +23,15 @@ public sealed class TpmAttestation : Attestation
...
+        /// <param name="endorsementKey">the string with the TPM endorsement key. It cannot be null or empty.</param>
+        /// <param name="storageRootKey">the string with the TPM storage root key. It can be null or empty.</param>
+        /// <exception cref="ArgumentException">if the endorsementKey is null or empty.</exception>
         [JsonConstructor]
         public TpmAttestation(string endorsementKey, string storageRootKey = null)
         {
             try
             {
-                EndorsementKey = endorsementKey;
+                EndorsementKey = endorsementKey ?? throw new ArgumentNullException(nameof(endorsementKey));
```
drwill-ms, line 34: If it can't be empty, a null check (??) won't cut it.

How about:

```suggestion
                Argument.AssertNotNullOrEmpty(endorsementKey, nameof(endorsementKey));
                EndorsementKey = endorsementKey;
```

provisioning/service/src/Enrollments/TpmAttestation.cs
```diff
@@ -25,15 +23,15 @@ public sealed class TpmAttestation : Attestation
         /// with both endorsement and storage root keys. Only the endorsement
         /// key is mandatory.
         /// </remarks>
-        /// <param name="endorsementKey">the <c>string</c> with the TPM endorsement key. It cannot be <c>null</c> or empty.</param>
-        /// <param name="storageRootKey">the <c>string</c> with the TPM storage root key. It can be <c>null</c> or empty.</param>
-        /// <exception cref="ArgumentException">if the endorsementKey is <c>null</c> or empty.</exception>
+        /// <param name="endorsementKey">the string with the TPM endorsement key. It cannot be null or empty.</param>
+        /// <param name="storageRootKey">the string with the TPM storage root key. It can be null or empty.</param>
+        /// <exception cref="ArgumentException">if the endorsementKey is null or empty.</exception>
```
drwill-ms, line 28: End then document both ArgumentNullException and ArgumentException.

provisioning/service/src/Enrollments/X509Attestation.cs
```diff
@@ -92,17 +91,16 @@ public static X509Attestation CreateFromClientCertificates(X509Certificate2 prim
...
         /// Creates a new instance of the X509Attestation with the primary and secondary certificates in X509Certificate2 objects.
         /// </remarks>
-        /// <param name="primary">the <c>X509Certificate2</c> with the primary certificate. It cannot be <c>null</c>.</param>
-        /// <param name="secondary">the <c>X509Certificate2</c> with the secondary certificate. It can be <c>null</c> (ignored).</param>
+        /// <param name="primary">The X509Certificate2 with the primary certificate. It cannot be null.</param>
+        /// <param name="secondary">The X509Certificate2 with the secondary certificate. It can be null (ignored).</param>
         /// <returns>The new instance of the X509Attestation.</returns>
-        /// <exception cref="ArgumentException">if the primary certificate is <c>null</c>.</exception>
-        /// <exception cref="CryptographicException">if the one of the provided certificate is invalid.</exception>
+        /// <exception cref="ArgumentException">If the primary certificate is null.</exception>
```
drwill-ms, line 97: ArgumentNullException

drwill-ms (changes requested):
Comments inline.

provisioning/service/src/Authentication/SecurityHelper.cs
```diff
@@ -2,25 +2,16 @@
...
         {
-            if (string.IsNullOrWhiteSpace(serviceHostName))
-            {
-                throw new ArgumentNullException(nameof(serviceHostName));
-            }
-
-            if (string.IsNullOrWhiteSpace(serviceName))
-            {
-                throw new ArgumentNullException(nameof(serviceName));
-            }
```
tmahmood-microsoft, line 21: These checks are performed in ServiceConnectionStringBuilder where connection string is parsed and broken in different parameters.

provisioning/device/src/ProvisioningTransportHandler.cs
```diff
@@ -38,33 +38,29 @@ public ProvisioningTransportHandler()
...
             set
             {
-                if (value == null)
-                {
-                    throw new ArgumentNullException(nameof(value));
-                }
-
+                _innerHandler = value ?? throw new InvalidOperationException(nameof(value));
```
drwill-ms, line 47:
```suggestion
                _innerHandler = value ?? throw new InvalidOperationException($"{nameof(InnerHandler)} cannot be null.");
```

provisioning/device/src/ProvisioningTransportHandler.cs
```diff
@@ -38,33 +38,29 @@ public ProvisioningTransportHandler()
...
         /// <summary>
         /// Gets or sets the port number.
         /// </summary>
+        /// <exception cref="InvalidOperationException">When the provided port value is out of range.</exception>
         public int Port
         {
             get => _port;
             set
             {
                 if (value < 1 || value > 65535)
```
drwill-ms, line 61: make constants of these, to refer to in a meaningful error message on line 63.

provisioning/service/src/Authentication/ServiceAuthenticationWithSharedAccessPolicyKey.cs
```diff
@@ -38,11 +34,6 @@ public string Key
 
         public ServiceConnectionStringBuilder Populate(ServiceConnectionStringBuilder provisioningConnectionStringBuilder)
         {
```
drwill-ms, line 36: Debug.Assert(provisioningConnectionStringBuilder != null, $"nameof(provisioningConnectionStringBuilder)} cannot be null. Validate parameters upstream.");

provisioning/service/src/Authentication/ServiceConnectionStringBuilder.cs
```diff
@@ -167,9 +133,8 @@ private static void ValidateFormat(string value, string propertyName, Regex rege
         {
             if (!regex.IsMatch(value))
             {
-                throw new ArgumentException(
-                    $"The connection string has an invalid value for property: {propertyName}",
-                    nameof(value));
+                throw new InvalidOperationException(
```
drwill-ms, line 136: Consider FormatException, and on 137 and perhaps more.

provisioning/service/src/Enrollments/QueryResult.cs
```diff
@@ -87,12 +87,12 @@ public class QueryResult
         /// <summary>
         /// CONSTRUCTOR
         /// </summary>
```
drwill-ms, line 89: Worth a team discussion for doc style.

provisioning/service/src/Enrollments/X509Attestation.cs
```diff
@@ -36,16 +36,16 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
     ///     </item>
     /// </list>
     ///
-    /// The provisioning service allows user to create <see cref="IndividualEnrollment"/> and <see cref="EnrollmentGroup"/>. 
-    ///     For all operations over <see cref="IndividualEnrollment"/> with <b>X509</b>, user must provide a 
-    ///     <b>clientCertificates</b>, and for operations over <see cref="EnrollmentGroup"/>, user must provide a 
+    /// The provisioning service allows user to create <see cref="IndividualEnrollment"/> and <see cref="EnrollmentGroup"/>.
+    ///     For all operations over <see cref="IndividualEnrollment"/> with <b>X509</b>, user must provide a
+    ///     <b>clientCertificates</b>, and for operations over <see cref="EnrollmentGroup"/>, user must provide a
```
drwill-ms, line 41: Consider also removing bold tags, if we're getting rid of code tags too.

provisioning/service/src/Enrollments/TwinState.cs
```diff
@@ -95,10 +95,10 @@ public class TwinState
         ///         }
         ///     }
         /// }
-        /// </c>
+        /// </code>
         /// </example>
-        /// <param name="tags">the <see cref="TwinCollection"/> with the initial tags state. It can be <c>null</c>.</param>
-        /// <param name="desiredProperties">the <see cref="TwinCollection"/> with the initial desired properties. It can be <c>null</c>.</param>
+        /// <param name="tags">the <see cref="TwinCollection"/> with the initial tags state. It can be null.</param>
```
drwill-ms, line 100:
```suggestion
        /// <param name="tags">The twin collection with the initial tags state. It can be null.</param>
```

provisioning/service/src/Enrollments/X509Attestation.cs
```diff
@@ -74,16 +74,15 @@ public sealed class X509Attestation : Attestation
         /// <remarks>
         /// Creates a new instance of the X509Attestation with the primary certificate in X509Certificate2 object.
         /// </remarks>
-        /// <param name="primary">the <c>X509Certificate2</c> with the primary certificate. It cannot be <c>null</c>.</param>
+        /// <param name="primary">The X509Certificate2 with the primary certificate. It cannot be null.</param>
         /// <returns>The new instance of the X509Attestation.</returns>
-        /// <exception cref="ArgumentException">if the primary certificate is <c>null</c>.</exception>
-        /// <exception cref="CryptographicException">if the one of the provided certificate is invalid.</exception>
+        /// <exception cref="ArgumentException">If the primary certificate is null.</exception>
```
drwill-ms, line 79:
```suggestion
        /// <exception cref="ArgumentException">If <paramref cref="primary"/> is null.</exception>
```

provisioning/service/src/Enrollments/X509Attestation.cs
```diff
@@ -74,16 +74,15 @@ public sealed class X509Attestation : Attestation
...
+        /// <exception cref="CryptographicException">If the one of the provided certificate is invalid.</exception>
```
drwill-ms, line 80:
```suggestion
        /// <exception cref="CryptographicException">If the provided certificate is invalid.</exception>
```

provisioning/service/src/Enrollments/X509Attestation.cs
drwill-ms: Also should be `ArgumentNullException`.

provisioning/service/src/Twin/TwinCollection.cs
```diff
@@ -232,12 +232,12 @@ public IEnumerator GetEnumerator()
...
-        /// If this method is used with a <see cref="TwinCollection"/> returned from a <c>DeviceClient</c> it will always return a
+        /// If this method is used with a <see cref="TwinCollection"/> returned from a DeviceClient it will always return a
         /// <see cref="JToken"/>. However, if you are using this method with a <see cref="TwinCollection"/> returned from a
-        /// <c>RegistryManager</c> client, it will return the corresponding type depending on what is stored in the properties collection.
-        /// 
+        /// RegistryManager client, it will return the corresponding type depending on what is stored in the properties collection.
+        ///
         /// For example a <see cref="List{T}"/> would return a <see cref="TwinCollectionArray"/>, with the metadata intact, when used with
-        /// a <see cref="TwinCollection"/> returned from a <c>RegistryManager</c> client. If you need this method to always return a
+        /// a <see cref="TwinCollection"/> returned from a RegistryManagerclient. If you need this method to always return a
```
drwill-ms, line 240:
```suggestion
        /// a <see cref="TwinCollection"/> returned from a <see cref="IotHubServcieClient.Twins"/> client. If you need this method to always return a
```

provisioning/service/src/Authentication/SharedAccessSignature.cs
```diff
@@ -47,19 +43,9 @@ private SharedAccessSignature(string shareAccessSignatureName, DateTime expiresO
 
         public string Signature { get; private set; }
 
-        public static SharedAccessSignature Parse(string shareAccessSignatureName, string rawToken)
+        internal static SharedAccessSignature Parse(string serviceName, string sharedAccessSignature)
         {
-            if (string.IsNullOrWhiteSpace(shareAccessSignatureName))
```
drwill-ms, line 52: Debug.Asserts?

provisioning/service/src/Query.cs
```diff
@@ -36,25 +36,25 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
...
     ///     contains 0, the Device Provisioning Service will ignore it and use a standard page size.
     ///
-    /// You can use this Object as a standard iterator, just using the <c>HasNext</c> and <c>NextAsync</c> in a
-    ///     <c>while</c> loop, up to the point where the <c>HasNext</c> contains <c>false</c>. But, keep
-    ///     in mind that the <see cref="QueryResult"/> can contain a empty list, even if the <c>HasNext</c> contained
-    ///     <c>true</c>. For example, image that you have 10 IndividualEnrollment in the Device Provisioning Service
-    ///     and you created new query with the <c>PageSize</c> equals 5. In the first iteration, <c>HasNext</c>
-    ///     will contains <c>true</c>, and the first <c>NextAsync</c> will return a <c>QueryResult</c> with
-    ///     5 items. After, your code will check the <c>HasNext</c>, which will contains <c>true</c> again. Now,
+    /// You can use this Object as a standard iterator, just using the <code>HasNext</code> and <code>NextAsync</code> in a
```
drwill-ms, line 43: These aren't multi-line so <c> is the correct tag.

drwill-ms: Although one could argue some of them don't need a code tag at all, like "while" for while loop.

drwill-ms (approved):

PR #2720, opened 2022-09-21
Revamp exceptions thrown by the dps device & service clients

brycewang-microsoft:
+ Pass `cancellationToken` into "RetryOperationsAsync" calls for e2e test ([ref](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/e2e/test/helpers/RetryOperationHelper.cs))
+ Throw `TaskCanceledException` and `OperationCanceledException` as-is
+ Stop throwing `TimeoutException`

### DIFF
--- a/e2e/test/helpers/RetryOperationHelper.cs
+++ b/e2e/test/helpers/RetryOperationHelper.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client;
 
@@ -15,14 +16,22 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
     public class RetryOperationHelper
     {
         /// <summary>
-        /// Rety an async operation based on the retry strategy supplied.
+        /// Retry an async operation based on the retry strategy supplied.
         /// </summary>
+        /// <remarks>
+        /// This is for E2E tests of provisioning service clients .
+        /// </remarks>
         /// <param name="asyncOperation">The async operation to be retried.</param>
         /// <param name="retryPolicy">The retry policy to be applied.</param>
         /// <param name="retryableExceptions">The exceptions to be retried on.</param>
         /// <param name="logger">The <see cref="MsTestLogger"/> instance to be used.</param>
-        /// <returns></returns>
-        public static async Task RetryOperationsAsync(Func<Task> asyncOperation, IRetryPolicy retryPolicy, HashSet<Type> retryableExceptions, MsTestLogger logger)
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        public static async Task RetryOperationsAsync(
+            Func<Task> asyncOperation,
+            IRetryPolicy retryPolicy,
+            HashSet<Type> retryableExceptions,
+            MsTestLogger logger,
+            CancellationToken cancellationToken = default)
         {
             int counter = 0;
             bool shouldRetry;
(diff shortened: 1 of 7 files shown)


e2e/test/helpers/RetryOperationHelper.cs
```diff
@@ -15,14 +16,22 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
...
         /// <param name="logger">The <see cref="MsTestLogger"/> instance to be used.</param>
-        /// <returns></returns>
-        public static async Task RetryOperationsAsync(Func<Task> asyncOperation, IRetryPolicy retryPolicy, HashSet<Type> retryableExceptions, MsTestLogger logger)
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
+        public static async Task RetryOperationsAsync(
+            Func<Task> asyncOperation,
+            IRetryPolicy retryPolicy,
+            HashSet<Type> retryableExceptions,
+            MsTestLogger logger,
+            CancellationToken cancellationToken = default)
```
brycewang-microsoft, line 34: with this change, now it's able to stop retries if the user's cancellation token has been signaled.

e2e/test/helpers/RetryOperationHelper.cs
```diff
@@ -47,19 +56,28 @@ public static async Task RetryOperationsAsync(Func<Task> asyncOperation, IRetryP
                 }
 
                 logger.Trace($"Will retry operation in {retryInterval}.");
-                await Task.Delay(retryInterval).ConfigureAwait(false);
+                await Task.Delay(retryInterval, cancellationToken);
             }
-            while (shouldRetry);
+            while (shouldRetry && !cancellationToken.IsCancellationRequested);
```
drwill-ms, line 61: Probably shouldn't just end the loop without some kind of exception that says why the operation was cancelled.

I think for the cancellation token being signalled, that is handled in line 59 - it'll produce an OperationCancelledException.

But what about `shouldRetry` being false?

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -45,6 +45,7 @@ public class ProvisioningE2ETests : E2EMsTestBase
 
         private static DirectoryInfo s_x509CertificatesFolder;
         private static string s_intermediateCertificateSubject;
+        private static CancellationTokenSource s_cancellationTokenSource = new CancellationTokenSource();
```
drwill-ms, line 48: Why do this instead of just passing `CancellationToken.None` in the tests?

This type is IDisposable, too.

provisioning/service/src/Contract/ContractApiHttp.cs
```diff
@@ -171,7 +171,7 @@ internal class ContractApiHttp : IContractApiHttp
                     // Unfortunately TaskCanceledException is thrown when HttpClient times out.
                     if (cancellationToken.IsCancellationRequested)
                     {
-                        throw new ProvisioningServiceClientException(ex.Message, ex);
+                        throw new OperationCanceledException(ex.Message);
```
drwill-ms, line 174: We should consider including the caught exception as an inner.

drwill-ms (changes requested):
Waiting on the test updates to not leak cancellation token sources.

e2e/test/helpers/RetryOperationHelper.cs
brycewang-microsoft: When `shouldRetry` is false, the retriable `IotHubServiceException` will be thrown out and we log that as well.

drwill-ms (approved):

PR #2721, opened 2022-09-22
Remove all 'ReceiveAsync' and 'CompleteAsync' style APIs

timtay-microsoft:
Users will be expected to solely use callbacks to receive methods, messages, feedback, etc.


iothub/device/src/IotHubBaseClient.cs
```diff
@@ -521,56 +421,56 @@ internal void OnConnectionStatusChanged(ConnectionStatusInfo connectionStatusInf
...
-            if (Logging.IsEnabled)
-                Logging.Exit(this, directMethodRequest.MethodName, directMethodRequest, nameof(OnMethodCalledAsync));
+                        directMethodResponse.RequestId = directMethodRequest.RequestId;
+                    }
+                    catch (Exception ex)
+                    {
+                        if (Logging.IsEnabled)
+                            Logging.Error(this, ex, nameof(OnMethodCalledAsync));
+
+                        directMethodResponse = new DirectMethodResponse()
```
drwill-ms, line 459:
```suggestion
                        directMethodResponse = new DirectMethodResponse
```

iothub/device/src/IotHubDeviceClient.cs
```diff
@@ -215,37 +188,38 @@ internal override void AddToPipelineContext()
...
             try
             {
-                Func<Message, object, Task> callback = _deviceReceiveMessageCallback?.Item1;
+                Func<Message, object, Task<MessageResponse>> callback = _deviceReceiveMessageCallback?.Item1;
                 object callbackContext = _deviceReceiveMessageCallback?.Item2;
 
                 if (callback != null)
                 {
-                    _ = callback.Invoke(message, callbackContext);
+                    return await callback.Invoke(message, callbackContext);
```
drwill-ms, line 207: configureawait

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -522,14 +521,28 @@ public async Task DisableMessageReceivingLinkAsync(CancellationToken cancellatio
...
+                    {
+                        acknowledgementType = AmqpIotDisposeActions.Released;
+                    }
+                    else if (response == MessageResponse.None)
+                    {
+                        acknowledgementType = AmqpIotDisposeActions.Rejected;
+                    }
+
+                    await DisposeMessageAsync(message.LockToken, acknowledgementType, CancellationToken.None).ConfigureAwait(false);
+                }
```
drwill-ms, line 545:
```suggestion
                if (_onDeviceMessageReceivedCallback == null)
                {
                    return;
                }
                
                MessageResponse response = await _onDeviceMessageReceivedCallback?.Invoke(message);
                AmqpIotDisposeActions acknowledgementType = AmqpIotDisposeActions.Accepted;
                if (response == MessageResponse.Abandoned)
                {
                    acknowledgementType = AmqpIotDisposeActions.Released;
                }
                else if (response == MessageResponse.None)
                {
                    acknowledgementType = AmqpIotDisposeActions.Rejected;
                }

                await DisposeMessageAsync(message.LockToken, acknowledgementType, CancellationToken.None).ConfigureAwait(false);
            }
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -795,12 +763,18 @@ private async Task HandleReceivedCloudToDeviceMessageAsync(MqttApplicationMessag
...
             {
-                // We are intentionally not awaiting _deviceMessageReceivedListener callback.
-                // This is a user-supplied callback that isn't required to be awaited by us. We can simply invoke it and continue.
-                _ = _deviceMessageReceivedListener.Invoke(receivedCloudToDeviceMessage);
+                MessageResponse acknowledgementType = await _deviceMessageReceivedListener.Invoke(receivedCloudToDeviceMessage).ConfigureAwait(false);
+
+                // Note that MQTT does not support Abandon or Reject
+                if (acknowledgementType != MessageResponse.Completed)
+                {
+                    if (Logging.IsEnabled)
```
drwill-ms, line 771: I guess we can't throw an exception because the customer won't even observe it, right?

Is there an error processor like what we have for file notifications?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -820,9 +794,9 @@ private async Task HandleReceivedCloudToDeviceMessageAsync(MqttApplicationMessag
             }
         }
 
-        private void HandleReceivedDirectMethodRequest(MqttApplicationMessageReceivedEventArgs receivedEventArgs)
+        private async Task HandleReceivedDirectMethodRequestAsync(MqttApplicationMessageReceivedEventArgs receivedEventArgs)
         {
-            receivedEventArgs.AutoAcknowledge = true;
+            await receivedEventArgs.AcknowledgeAsync(CancellationToken.None);
```
drwill-ms, line 799: Debug.Assert(receivedEventArgs != null)?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Correct.

> I guess we can't throw an exception because the customer won't even observe it, right?

Correct

> Is there an error processor like what we have for file notifications?

We have the connection status callback, but it wouldn't be appropriate to feed it an event like this since it isn't necessarily a connection level problem. My take on cases like this is that users will likely need to re-send the direct method because the service does seem to block on receiving this ack and then block on receiving the response

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -466,9 +468,22 @@ public static async Task<Task> SetModuleReceiveMethodAsync(IotHubModuleClient mo
...
+                {
+                    serviceSendTask,
+                    methodReceivedTask
+                },
+                TimeSpan.FromMinutes(1));
 
             await deviceClient.CloseAsync().ConfigureAwait(false);
+
+            // If the Task.WaitAll timed out, then one or both of these tasks timed out. This block clarifies which task(s) timed out
+            using (new AssertionScope())
```
timtay-microsoft, line 482: Here's a neat trick I picked up. In times like these where you have multiple assertions to make, blocks like this allow you to throw for **all** of the assertions that failed instead of just the **first** assertion that failed.

timtay-microsoft:
Closing this PR to take it in smaller chunks

PR #2722, opened 2022-09-22
Update digital twins temperature controller sample, and update servic…

drwill-ms:


iothub/service/samples/solutions/DigitalTwinClientSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -65,33 +65,34 @@ private async Task UpdateDigitalTwinComponentPropertyAsync()
             int desiredTargetTemperature = s_random.Next(0, 100);
 
             const string targetTemperaturePropertyName = "targetTemperature";
-            var updateOperation = new UpdateOperationsUtility();
+            var updateOperation = new JsonPatchDocument();
```
drwill-ms, line 68: @timtay-microsoft why not make the update API take this type rather than string? Do we not want to rely on an external library (maybe we already link to it?)?

iothub/service/samples/solutions/DigitalTwinClientSamples/TemperatureController/TemperatureControllerSample.cs
```diff
@@ -101,15 +102,16 @@ private async Task UpdateDigitalTwinComponentPropertyAsync()
...
 
             _logger.LogDebug($"Update the {targetTemperaturePropertyName} property under component {Thermostat1Component} on the {_digitalTwinId} `" +
                 $"digital twin to {desiredTargetTemperature}.");
-            HttpOperationHeaderResponse<DigitalTwinUpdateHeaders> updateDigitalTwinResponse = await _digitalTwinClient
-                .UpdateDigitalTwinAsync(_digitalTwinId, updateOperation.Serialize());
+            DigitalTwinUpdateResponse updateDigitalTwinResponse = await _hubClient
+                .DigitalTwins
+                .UpdateAsync(_digitalTwinId, updateOperation.ToString());
 
-            _logger.LogDebug($"Update {_digitalTwinId} digital twin response: {updateDigitalTwinResponse.Response.StatusCode}.");
```
drwill-ms, line 112: @timtay-microsoft we lost the status code on this response type. Was that intended?

timtay-microsoft: This is just the HTTP status, not some application layer field of interest. We don't expose the HTTP status in other service client operations either. It is assumed that the service client method will throw if it gets a non-200 status code

iothub/service/samples/solutions/DigitalTwinClientSamples/TemperatureController/TemperatureControllerSample.cs
timtay-microsoft: I don't know how intentional this was when we originally created the v1 digital twins client. I suspect we took it as a string to allow users to construct this field however they want rather than forcing them to use the UpdateOperationsUtility/JsonPatchDocument type

iothub/service/samples/solutions/DigitalTwinClientSamples/TemperatureController/TemperatureControllerSample.cs
drwill-ms: Other responses still have it, which is why I thought about it. But I realized they are direct method ones, and the status is part of the device's response.

iothub/service/samples/solutions/DigitalTwinClientSamples/TemperatureController/TemperatureControllerSample.cs
drwill-ms: Fair. Do you think we should consider taking this type directly, like we do in the ADT client? Or just reference in the doc comments that it can be useful and demonstrate it with samples?

iothub/service/samples/solutions/DigitalTwinClientSamples/TemperatureController/TemperatureControllerSample.cs
timtay-microsoft: Yeah, the direct method case is the application-level status code, not the HTTP-level status code.

PR #2724, opened 2022-09-22
Bring pnp changes to direct method

abhipsaMisra:


iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -105,5 +69,64 @@ public object Payload
...
+            : (int?)null;
+
+        /// <summary>
+        /// The direct method request payload, deserialized to the specified type.
+        /// </summary>
+        /// <typeparam name="T">The type to deserialize the direct method request payload to.</typeparam>
+        /// <param name="payload">When this method returns true, this contains the value of the direct method request payload.
+        /// When this method returns false, this contains the default value of the type <c>T</c> passed in.</param>
+        /// <returns><c>true</c> if the direct method request payload can be deserialized to type <c>T</c>; otherwise, <c>false</c>.</returns>
+        public bool TryGetPayload<T>(out T payload)
```
abhipsaMisra, line 108: Updated this method to TryGet format

iothub/device/src/DirectMethod/DirectMethodResponse.cs
```diff
@@ -11,11 +11,13 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public class DirectMethodResponse
     {
+
         /// <summary>
         /// Initializes an instance of this class.
         /// </summary>
-        public DirectMethodResponse()
+        public DirectMethodResponse(int status)
```
abhipsaMisra, line 18: Making status a mandatory field for direct method response

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -31,23 +31,23 @@ internal sealed class MqttTransportHandler : TransportHandler, IDisposable
 
         private const string DeviceToCloudMessagesTopicFormat = "devices/{0}/messages/events/";
         private const string ModuleToCloudMessagesTopicFormat = "devices/{0}/modules/{1}/messages/events/";
-        private string _deviceToCloudMessagesTopic;
-        private string _moduleToCloudMessagesTopic;
+        private readonly string _deviceToCloudMessagesTopic;
```
abhipsaMisra, line 34: The only changes in this file are making some fields readonly and moving the const and readonly fields to the top of the class definition

timtay-microsoft (approved):

iothub/device/src/DirectMethod/DirectMethodResponse.cs
drwill-ms: It was mandatory for PnP, but not otherwise mandatory, right? Or was it?

drwill-ms (approved):

iothub/device/src/DirectMethod/DirectMethodResponse.cs
abhipsaMisra: The public document doesn't call out the status being required. However, a direct method response without an associated status looks logically incomplete. What do you think?
For an additional data point, our public API had also expected the direct method response to accompany a status code: https://github.com/Azure/azure-iot-sdk-csharp/blob/main/iothub/device/src/DirectMethod/MethodResponse.cs

iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -105,5 +69,64 @@ public object Payload
         /// </remarks>
         [JsonIgnore]
         internal string RequestId { get; set; }
```
drwill-ms, line 71: internals below publics

iothub/device/src/DirectMethod/DirectMethodResponse.cs
drwill-ms: I agree that as a best practice this is a good idea.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -212,7 +213,7 @@ internal MqttTransportHandler(PipelineContext context, IotHubClientMqttSettings
                 }
             }
 
-            MqttClientOptionsBuilderTlsParameters tlsParameters = new MqttClientOptionsBuilderTlsParameters();
+            var tlsParameters = new MqttClientOptionsBuilderTlsParameters();
```
drwill-ms, line 216: var is not obvious

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -30,20 +30,15 @@ public class IotHubDeviceClientTests
 
         private static readonly IotHubConnectionCredentials s_iotHubConnectionCredentials = new(FakeConnectionString);
 
-        private DirectMethodResponse directMethodResponseWithPayload = new DirectMethodResponse()
+        private DirectMethodResponse directMethodResponseWithPayload = new DirectMethodResponse(200)
         {
-            Status = 200,
             Payload = Encoding.UTF8.GetBytes("{\"name\":\"ABC\"}")
```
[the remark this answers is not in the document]
drwill-ms: Even better use a custom type, or just a primitive (int, DateTimeOffset).

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -17,8 +18,8 @@ namespace Microsoft.Azure.Devices.E2ETests.Methods
     [TestCategory("IoTHub")]
     public class MethodE2ETests : E2EMsTestBase
     {
-        public static readonly object DeviceResponseJson = new { name = "e2e_test" };
```
abhipsaMisra, line 20: @drwill-ms The "ServiceSendMethodAndVerifyResponseAsync" method isn't able to make use of these changes because the custom serialization implementation isn't available on the service client yet. So, that assertion will still require us to serialize the payload and then compare.
If we plan on adding custom serialization to the service client eventually, then we'll be able to add this there.

PR #2725, opened 2022-09-22
Add serialization classes

abhipsaMisra:


iothub/device/src/IotHubClientOptions.cs
```diff
@@ -45,6 +45,19 @@ public IotHubClientOptions(IotHubClientTransportSettings transportSettings)
...
+        /// </summary>
+        /// <remarks>
+        /// The <see cref="PayloadConvention"/> defines both the serializer and encoding to be used.
+        /// You will only need to set this if you have objects that have special serialization rules or require a specific byte encoding.
+        /// <para>
+        /// The default value is set to <see cref="DefaultPayloadConvention"/> which uses the <see cref="NewtonsoftJsonPayloadSerializer"/> serializer
+        /// and <see cref="Utf8PayloadEncoder"/> encoder.
+        /// </para>
+        /// </remarks>
+        public PayloadConvention PayloadConvention { get; set; } = DefaultPayloadConvention.Instance;
```
abhipsaMisra, line 59: It might be worth changing the name from "convention" to something more serialization centric. Since we're not putting pnp in the forefront anymore, calling this a convention might not be the best idea.

iothub/device/src/Serialization/NewtonsoftJsonPayloadSerializer.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <remarks>
+        /// This class is usedto attempt to get a property of the underlying JSON object.
+        /// </remarks>
+        /// <typeparam name="T">The type to convert the retrieved property to.</typeparam>
+        /// <param name="nestedJsonObject">The object that might contain the nested property.
+        /// This needs to be in the json object equivalent format as required by the serializer or the string representation of it.</param>
+        /// <param name="propertyName">The name of the property to be retrieved.</param>
+        /// <param name="outValue">The retrieved value.</param>
+        /// <returns>True if the nested object contains an element with the specified key, otherwise false.</returns>
+        internal bool TryGetNestedJsonObjectValue<T>(JObject nestedJsonObject, string propertyName, out T outValue)
```
abhipsaMisra, line 66: We might not need this method since we won't be doing fancy component-property parsing. I've left it in for now for us to delete it later once the rest of the serialization code is in.

timtay-microsoft (approved):

iothub/device/src/Serialization/Utf8PayloadEncoder.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Text;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// A UTF-8 PayloadEncoder implementation.
+    /// </summary>
+    public class Utf8PayloadEncoder : PayloadEncoder
+    {
+        private Utf8PayloadEncoder()
```
drwill-ms, line 13: private? Why not internal or even protected internal?

drwill-ms (approved):

iothub/device/src/Serialization/Utf8PayloadEncoder.cs
abhipsaMisra: sure, that makes sense

abhipsaMisra: Looks like we're exposing a public static instance of this class which doesn't require any additional customization before being used. That should be enough for both users wanting to use this class or test something that uses this class. So, I'm inclined to not add additional visibility on this private constructor. What do you think?

PR #2726, opened 2022-09-22
Merge TPM package into Authentication package

tmahmood-microsoft:


security/tpm/src/AuthenticationProviderTpmHsm.cs
```diff
@@ -3,10 +3,9 @@
 
 using System;
 using System.Runtime.InteropServices;
-using Microsoft.Azure.Devices.Authentication;
 using Tpm2Lib;
 
-namespace Microsoft.Azure.Devices.Provisioning.Security
+namespace Microsoft.Azure.Devices.Authentication
 {
     /// <summary>
```
timtay-microsoft, line 10: It looks like this file wasn't moved to the Authentication directory like the others

timtay-microsoft (approved):

security/tpm/src/AuthenticationProviderTpmHsm.cs
tmahmood-microsoft: It was moved but I did not delete a copy of it from the TPM directory.

Updated it in recent change.

PR #2733, opened 2022-09-22
Add retry to a flakey step in an e2e test

timtay-microsoft:
The first call to get the module often fails with "The device does not have this module registered" because we very recently registered it and the service is slow. Adding retry here so that we call get twin until we don't see an error

### DIFF
--- a/e2e/test/iothub/service/RegistryE2ETests.cs
+++ b/e2e/test/iothub/service/RegistryE2ETests.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
         private readonly string _idPrefix = $"{nameof(RegistryE2ETests)}_";
 
         private static readonly IRetryPolicy s_provisioningServiceRetryPolicy = new ProvisioningServiceRetryPolicy();
+        // In particular, this should retry on "module not registered on this device" errors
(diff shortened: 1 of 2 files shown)


abhipsaMisra (approved):

brycewang-microsoft (approved):

e2e/test/iothub/service/RegistryE2ETests.cs
```diff
@@ -24,6 +25,9 @@ public class RegistryE2ETests : E2EMsTestBase
     {
         private readonly string _idPrefix = $"{nameof(RegistryE2ETests)}_";
 
+        private static readonly IRetryPolicy s_provisioningServiceRetryPolicy = new ProvisioningServiceRetryPolicy();
+        private static readonly HashSet<Type> s_retryableExceptions = new HashSet<Type> { typeof(IotHubServiceException) };
```
drwill-ms, line 29: This concept will need to be revamped now that exception type isn't the deciding factor, rather the ErrorCode or even the IsTransient property.

drwill-ms (approved):

andyk-ms (approved):

e2e/test/iothub/service/RegistryE2ETests.cs
timtay-microsoft: Agreed

PR #2738, opened 2022-09-23
Rename MessageResponse to MessageAcknowledgementType

timtay-microsoft:
Also remove unused DeliveryAcknowledgement class. The concept it demonstrated is a service client only concept


iothub/device/src/MessageAcknowledgementType.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using static System.Net.WebRequestMethods;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// The options for acknowledging a cloud to device message.
+    /// </summary>
+    /// <seealso href="https://learn.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-c2d#the-cloud-to-device-message-life-cycle"/>
+    public enum MessageAcknowledgementType
```
timtay-microsoft, line 12: Does anyone feel this is still too generic sounding? Should I go with ```CloudToDeviceMessageAcknowledgementType``` instead?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -995,18 +995,6 @@ private static object ConvertToSystemProperty(KeyValuePair<string, string> prope
                 return DateTime.ParseExact(property.Value, "o", CultureInfo.InvariantCulture);
             }
 
-            if (property.Key == MessageSystemPropertyNames.Ack)
```
timtay-microsoft, line 998: There was no reason for the device client side to look for this property as the service would never send it. This was the only use of DeliveryAcknowledgement, so we can safely delete that class

SDK v2 migration guide.md
```diff
@@ -41,6 +41,7 @@ to migrate to version 2.x when they have the chance. For more details on LTS rel
 | `ReceiveAsync` | `ReceiveMessageAsync` |
 | `CompleteAsync` | `CompleteMessageAsync` |
 | `SetConnectionStatusChangesHandler` | `SetConnectionStatusChangeHandler` |
+| `MessageResponse` | `MessageAcknowledgementType` |
```
drwill-ms, line 44: Do you think we need the word Type?

iothub/device/src/IotHubModuleClient.cs
```diff
@@ -328,10 +328,10 @@ internal async Task OnModuleEventMessageReceivedAsync(Message message)
 
             try
             {
-                MessageResponse response = MessageResponse.Completed;
+                MessageAcknowledgementType response = MessageAcknowledgementType.Complete;
```
drwill-ms, line 331:
```suggestion
                var response = MessageAcknowledgementType.Complete;
```

iothub/device/src/MessageAcknowledgementType.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using static System.Net.WebRequestMethods;
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// The options for acknowledging a cloud to device message.
```
drwill-ms, line 9:
```suggestion
    /// The options for acknowledging a cloud-to-device (C2D) message.
```

iothub/device/src/MessageAcknowledgementType.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    /// <seealso href="https://learn.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-c2d#the-cloud-to-device-message-life-cycle"/>
+    public enum MessageAcknowledgementType
+    {
+        /// <summary>
+        /// The message will be positively acknowledged. This removes the message from the queue
+        /// and it will not be sent again.
+        /// </summary>
```
drwill-ms, line 17: Use remarks for extra info. Also below.

```suggestion
        /// <summary>
        /// The message will be positively acknowledged.
        /// </summary>
        /// <remarks>
        /// This removes the message from the queue; it will not be sent again.
        /// </remarks>
```

iothub/device/src/MessageAcknowledgementType.cs
```diff
@@ -0,0 +1,37 @@
...
+        Complete,
+
+        /// <summary>
+        /// The message will be re-added to the queue and will be sent again.
```
drwill-ms, line 21:
```suggestion
        /// The message will be re-queued to be sent again.
```

iothub/device/src/MessageAcknowledgementType.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This option is not supported over MQTT or MQTT websockets
```
drwill-ms, line 24: If you're going to call out web sockets, I'd specify TCP specifically as well. But honestly, I don't think the protocol is worth mentioning.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: Even a user wouldn't specify it?

drwill-ms: image

iothub/device/src/MessageAcknowledgementType.cs
drwill-ms: I commented independently that I think we could drop the word "Type", but I wouldn't be against CloudToDeviceMessageAcknowledgment. It is a mouthful, though.

SDK v2 migration guide.md
timtay-microsoft: I like it here, yeah. It fits for an enum.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: The service client message chooses this enum, yes. But the value isn't forwarded to the device/module recipient. The service just uses this value to decide under what circumstances the service enqueues a message feedback record for the service client to consume

iothub/device/src/MessageAcknowledgementType.cs
timtay-microsoft: Let's go with ```CloudToDeviceMessageAcknowledgment``` then. Drop the "type" to spare the length, but still sounds descriptive enough

iothub/device/src/IotHubModuleClient.cs
```diff
@@ -32,7 +32,7 @@ public class IotHubModuleClient : IotHubBaseClient
         private readonly SemaphoreSlim _moduleReceiveMessageSemaphore = new(1, 1);
 
         // Cloud-to-module message callback information
-        private volatile Tuple<Func<Message, object, Task<MessageResponse>>, object> _defaultEventCallback;
+        private volatile Tuple<Func<Message, object, Task<CloudToDeviceMessageAcknowledgement>>, object> _defaultEventCallback;
```
abhipsaMisra, line 35: I like CloudToDeviceMessageAcknowledgement for being descriptive of its function, but we're going to run into the same device vs module issue again. CloudToClient...?

timtay-microsoft: Good point...

timtay-microsoft: I'd rather revert to ```MessageAcknowledgement``` or ```MessageAcknowledgementType``` than use "Client" as a stand-in for both devices and modules. "Client" could mean device client, module client, or service client so it is less than helpful

brycewang-microsoft: I'd vote for `MessageAcknowledgement` then

timtay-microsoft: Let's go with that then

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: Ah

iothub/device/src/MessageAcknowledgement.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// The message will be negatively acknowledged.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This removes the message from the queue; it will not be sent again.
+        /// </para>
+        /// <para>
+        /// This option is not supported over MQTT.
```
drwill-ms, line 36: What will the SDK do if a user returns it? Ignore and send Complete anyway?

drwill-ms (approved):

iothub/device/src/MessageAcknowledgement.cs
timtay-microsoft: Yes, and log an error saying "not supported"

brycewang-microsoft (approved):

PR #2739, opened 2022-09-23
Catch MqttCommunicationTimedOutException and rethrow as OperationCanceledException

timtay-microsoft:
MQTTNet appears to throw MqttCommunicationTimedOutException in cases when cancellation is requested ([for example](https://github.com/dotnet/MQTTnet/blob/ef85ef676a20fcd85ce1d6632053070d84c65381/Source/MQTTnet/PacketDispatcher/MqttPacketAwaitable.cs#L34))

In order to have more uniformity, I'd rather we threw this as an operation canceled exception. In order to do that, all cancellable calls to the mqtt client are now wrapped in try/catch like

```csharp
try
{
    await _mqttClient.ConnectAsync(..., cancellationToken);
}
catch (MqttCommunicationTimedOutException e)
{
    throw new OperationCanceledException("Timed out waiting to open the MQTT connection", e); //nest the exception
}
```

Note that this change is made for both the IoT Hub device client MQTT layer and the DPS device client MQTT layer

This PR also makes OperationCanceledException a transient exception for the sake of the SDK's retry logic


iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -51,6 +51,11 @@ private class TransientErrorStrategy : ITransientErrorDetectionStrategy
         {
             public bool IsTransient(Exception ex)
             {
+                if (ex is OperationCanceledException)
+                {
+                    return true;
+                }
```
timtay-microsoft, line 57: @drwill-ms I believe this is what we were looking for this morning. Now the retry policy will retry if the encountered error is an OperationCanceledException.

tmahmood-microsoft (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: I'm not sure we want this to be transient. If the user signals to cancel, but we say, ok sure but let me just retry first, that would be a problem. No?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -336,6 +337,12 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
                         throw new IotHubClientException("Failed to open the MQTT connection", cfe);
                 }
             }
+            catch (MqttCommunicationTimedOutException e)
```
drwill-ms, line 340: ex

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -432,6 +448,12 @@ public override async Task SendMethodResponseAsync(DirectMethodResponse methodRe
                     throw new IotHubClientException($"Failed to send direct method response with reason code {result.ReasonCode}", true);
                 }
             }
+            catch (MqttCommunicationTimedOutException e)
+            {
+                // MQTTNet throws MqttCommunicationTimedOutException instead of OperationCanceledException
+                // when the cancellation token requests cancellation.
+                throw new OperationCanceledException("Timed out waiting for MQTT message to be acknowledged.", e);
```
drwill-ms, line 455: Might make the string a constant, or even make throwing this method a private helper so you don't have to repeat it so much.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -716,22 +759,31 @@ private async Task UnsubscribeAsync(string topic, CancellationToken cancellation
...
+                MqttClientUnsubscribeResult unsubscribeResults = await _mqttClient.UnsubscribeAsync(unsubscribeOptions, cancellationToken).ConfigureAwait(false);
 
-            MqttClientUnsubscribeResultItem unsubscribeResult = unsubscribeResults.Items.FirstOrDefault();
-            if (!unsubscribeResult.TopicFilter.Equals(fullTopic))
-            {
-                throw new IotHubClientException($"Received unexpected unsubscription from topic {unsubscribeResult.TopicFilter}", true);
-            }
+                if (unsubscribeResults?.Items == null || unsubscribeResults.Items.Count != 1)
+                {
+                    throw new IotHubClientException($"Failed to unsubscribe to topic {fullTopic}", true);
```
drwill-ms, line 768: Can we set a StatusCode? Users need to be able to distinguish.

timtay-microsoft: This is more indicative of an MQTTNet bug than anything, so we don't have a status code for a case like this.

drwill-ms: Can we associate it with an exising one? Timeout?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
timtay-microsoft: True. This may be a better fit for application level retry logic rather than SDK level retry logic.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: Maybe network error or server error. The only times I'd expect this exception to be thrown are if MQTTNet has a bug and returns a null/empty unsubscribeResults object (no status code for this case), if the hub acknowledges the incorrect unsubscriptions (server error), or if the packets sent over the wire experienced some loss and the ack sent by the service comes in malformed (network error).

timtay-microsoft: Of those scenarios, the only one I'd expect we'd ever see is packet loss messing with this field's results, so I'm leaning towards network errors as the status code

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: If you think this only happens when a ct is signalled, then yes. But I'm pretty sure the cases I ran into there was no cancellation token.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: Let's make a best match.

drwill-ms (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
timtay-microsoft: I'll dig into MQTTNet more to see if there is some default timeout that they observe alongside the cancellationtoken we pass in, then. If they have an independent timeout, then I could see the SDK's retry logic marking those cases as retryable, sure.

timtay-microsoft: Looks like there is a default operation timeout of 100 seconds in MQTTNet regardless of cancellation token:

https://github.com/dotnet/MQTTnet/blob/ef85ef676a20fcd85ce1d6632053070d84c65381/Source/MQTTnet/Client/Options/MqttClientOptions.cs#L93

We can override this to TimeSpan.MaxValue though so that we only ever get MqttCommunicationTimedOutException when the cancellation token requests cancellation. With that, we can leave OperationCanceledException marked as a non-transient exception. Thoughts?

abhipsaMisra (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
timtay-microsoft: I'll address this point in #2743 since I'm updating the ```isTransient``` flag on these same exceptions in that PR

PR #2740, opened 2022-09-23
Periodically check for expired messages in MQTT & AMQP

tmahmood-microsoft:
For Device client, when the device requests twin, the service acknowledges the request and sends an Ack message back. The service then at some point later responds with the actual twin. We keep a track of each request using Dictionary with pairs of request-Id and Twin Response completion task. Upon receiving the actual twin from service, the corresponding pair is removed from the dictionary.

There might be cases where the service never responds with the twin after Ack message. This issue over multiple messages will cause dictionary to grow and as a worst-case scenario, we get out of memory exception.

This PR is to make sure we remove the requestId - twin response pair from the dictionary after an hour as the service has a timeout for such case. 
The implementation is to have another dictionary with requestId-datetime pairs. We start checking this dictionary after an hour of opening the client and then periodically check every hour. If the request was sent more than an hour ago with no response yet, we remove it from both the dictionaries and stop waiting for the response.


iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -18,6 +19,9 @@ internal class AmqpTransportHandler : TransportHandler
         private readonly Action<TwinCollection> _onDesiredStatePatchListener;
         private readonly object _lock = new();
         private readonly ConcurrentDictionary<string, TaskCompletionSource<Twin>> _twinResponseCompletions = new ConcurrentDictionary<string, TaskCompletionSource<Twin>>();
+        private readonly ConcurrentDictionary<string, DateTime> _twinResponseTimeouts = new();
```
drwill-ms, line 22: Can we use DateTimeOffset?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -18,6 +19,9 @@ internal class AmqpTransportHandler : TransportHandler
...
+        private Timer _twinTimeoutTimer;
```
drwill-ms, line 23: Timer is IDisposable. Is it being disposed when this class closes?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -98,6 +102,9 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
             {
                 using var ctb = new CancellationTokenBundle(_operationTimeout, cancellationToken);
                 await _amqpUnit.OpenAsync(ctb.Token).ConfigureAwait(false);
+
+                // Create a timer to remove any expired messages. The timer would invoke callback after first hour of execution and then after every minute.
```
timtay-microsoft, line 106: Checking for expired messages every minute seems a bit aggressive. Why not check once per hour? We don't need this map to be up-to-date so much as we need it to just not grow over time.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: Unless I have the wrong class. Which one is this?

https://learn.microsoft.com/en-us/dotnet/api/system.timers.timer.-ctor?view=net-6.0

tmahmood-microsoft: Yes, disposing it when we close the client.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -400,7 +428,14 @@ public override async Task<long> SendTwinPatchAsync(TwinCollection reportedPrope
             }
             finally
             {
-                _twinResponseCompletions.TryRemove(correlationId, out _);
+                if (_twinResponseCompletions.ContainsKey(correlationId))
+                {
+                    _twinResponseCompletions.TryRemove(correlationId, out _);
```
timtay-microsoft, line 433: You can combine these two calls into one like:
```
_ = _twinResponseCompletions.TryRemove(correlationId, out _)
```

TryRemove will remove it if it is present and will just return false if it was not present

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -106,6 +113,26 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
             }
         }
 
+        private void checkTimeout(Object _)
```
drwill-ms, line 116:
```suggestion
        private void CheckTimeout(Object _)
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -99,8 +99,13 @@ internal sealed class MqttTransportHandler : TransportHandler, IDisposable
         private readonly ConcurrentDictionary<string, TaskCompletionSource<GetTwinResponse>> _getTwinResponseCompletions = new();
         private readonly ConcurrentDictionary<string, TaskCompletionSource<PatchTwinResponse>> _reportedPropertyUpdateResponseCompletions = new();
 
+        private readonly ConcurrentDictionary<string, DateTime> _twinResponseTimeouts = new();
+
         private bool _isSubscribedToTwinResponses;
 
+        private Timer _twinTimeoutTimer;
+        private TimeSpan _twinResponseTimeout = TimeSpan.FromMinutes(60);
```
timtay-microsoft, line 107: Add a comment explaining what this is used for

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -106,6 +113,26 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
+        {
+            if (_twinResponseTimeouts.Any())
+            {
+                var currentDateTime = DateTime.UtcNow;
+                TimeSpan difference;
```
drwill-ms, line 121: Why declare this outside the foreach?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -106,6 +113,26 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
+                foreach (KeyValuePair<string, DateTime> entry in _twinResponseTimeouts)
+                {
+                    difference = currentDateTime - entry.Value;
+                    if (difference >= _twinResponseTimeout)
+                    {
+                        if (_twinResponseCompletions.ContainsKey(entry.Key))
+                            _twinResponseCompletions.TryRemove(entry.Key, out TaskCompletionSource<Twin> _);
```
drwill-ms, line 128: Except for logging, we always use braces.
```suggestion
                        if (_twinResponseCompletions.ContainsKey(entry.Key))
                        {
                            _twinResponseCompletions.TryRemove(entry.Key, out TaskCompletionSource<Twin> _);
                        }
```

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -18,6 +19,9 @@ internal class AmqpTransportHandler : TransportHandler
         private readonly Action<TwinCollection> _onDesiredStatePatchListener;
         private readonly object _lock = new();
         private readonly ConcurrentDictionary<string, TaskCompletionSource<Twin>> _twinResponseCompletions = new ConcurrentDictionary<string, TaskCompletionSource<Twin>>();
+        private readonly ConcurrentDictionary<string, DateTime> _twinResponseTimeouts = new();
+        private Timer _twinTimeoutTimer;
+        private TimeSpan _twinResponseTimeout = TimeSpan.FromMinutes(60);
```
drwill-ms, line 24:
```suggestion
        private static readonly TimeSpan s_twinResponseTimeout = TimeSpan.FromMinutes(60);
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -312,6 +317,9 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
                 await _mqttClient.ConnectAsync(_mqttClientOptions, cancellationToken).ConfigureAwait(false);
                 _mqttClient.DisconnectedAsync += HandleDisconnectionAsync;
                 _mqttClient.ApplicationMessageReceivedAsync += HandleReceivedMessageAsync;
+
+                // Create a timer to remove any expired messages. The timer would invoke callback after first hour of execution and then after every minute.
+                _twinTimeoutTimer = new Timer(checkTimeout, null, _twinResponseTimeout.Milliseconds, TimeSpan.FromMinutes(1).Milliseconds);
```
timtay-microsoft, line 322: FYI, 

```csharp
TimeSpan.FromMinutes(1).Milliseconds
```

evaluates to 0 because that field means the Milliseconds section of the broader timespan. For instance,

0:1:00:25 -> 25
(hours):(minutes):(seconds):(milliseconds)

What you are looking for here is
```csharp
TimeSpan.FromMinutes(1).TotalMilliseconds
```

which evaluates to the total number of milliseconds in a minute.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -106,6 +113,26 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
+            if (_twinResponseTimeouts.Any())
+            {
+                var currentDateTime = DateTime.UtcNow;
+                TimeSpan difference;
+                foreach (KeyValuePair<string, DateTime> entry in _twinResponseTimeouts)
+                {
+                    difference = currentDateTime - entry.Value;
+                    if (difference >= _twinResponseTimeout)
+                    {
+                        if (_twinResponseCompletions.ContainsKey(entry.Key))
```
drwill-ms, line 127: Do we even need a check to ContainsKey? TryRemove won't fault if it isn't there.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -116,10 +143,10 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
                 _closed = true;
             }
 
+            cancellationToken.ThrowIfCancellationRequested();
             try
             {
-                cancellationToken.ThrowIfCancellationRequested();
-
+                _twinTimeoutTimer.Dispose();
```
drwill-ms, line 149: If we create it on ctor and dispose on closeasync, we won't be able to call openasync again. We need to build/destroy it at the same type of event.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -400,7 +428,14 @@ public override async Task<long> SendTwinPatchAsync(TwinCollection reportedPrope
             }
             finally
             {
-                _twinResponseCompletions.TryRemove(correlationId, out _);
+                if (_twinResponseCompletions.ContainsKey(correlationId))
+                {
+                    _twinResponseCompletions.TryRemove(correlationId, out _);
+                    if (_twinResponseTimeouts.ContainsKey(correlationId))
```
timtay-microsoft, line 434: I would attempt to remove the entry from the timeouts dictionary even if it wasn't present in the completions dictionary.

drwill-ms, line 434: Same, don't think we need the ContainsKey check.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: so maybe dispose in the dispose method, and simply stop the timer and start it on open/close?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -116,10 +143,10 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
                 _closed = true;
             }
 
+            cancellationToken.ThrowIfCancellationRequested();
```
timtay-microsoft, line 146: Why is this being moved outside of the try block?

drwill-ms (changes requested):
Comments inline.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: Irrelevant if you remove the ConstainsKey check.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: We chose to do that in public API surface. Not in internal methods?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
tmahmood-microsoft: Just to avoid create a new variable for each key value pair. This way we can re-use it.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
tmahmood-microsoft: Copying the pattern we had in service cleint

ConsoleApp6/Program.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Collections.Concurrent;
+
+public class Program
+{
+    private static ConcurrentDictionary<string, DateTimeOffset> _twinResponseTimeouts = new();
+
+    public static void Main()
```
abhipsaMisra, line 9: Has this been checked in accidentally?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
tmahmood-microsoft: Disposing it in Dispose method.
There is no way to close the time though

ConsoleApp6/Program.cs
tmahmood-microsoft: Yes, my bad. I will remove it

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
tmahmood-microsoft: reverted it

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms: Not really saving anything, I think.

It makes it look like we're saving the value between iterations, so for code readability reasons, I'd inline it.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -106,6 +116,25 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
             }
         }
 
+        private void CheckTimeout(Object _)
```
drwill-ms, line 119:
```suggestion
        private void CheckTimeout(object _)
```

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -106,6 +116,25 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
+        {
+            if (_twinResponseTimeouts.Any())
+            {
+                var currentDateTime = DateTime.UtcNow;
+                TimeSpan difference;
+                foreach (KeyValuePair<string, DateTimeOffset> entry in _twinResponseTimeouts)
+                {
+                    difference = currentDateTime - entry.Value;
+                    if (difference >= s_twinResponseTimeout)
```
drwill-ms, line 128:
```suggestion
                    if (s_twinResponseTimeout < currentDateTime - entry.Value)
```

drwill-ms: Inlined, and > is cheaper than >=

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -95,6 +95,20 @@ internal sealed class MqttTransportHandler : TransportHandler, IDisposable
...
+
+        // Timer to check if any expired messages exist. The timer is executed after each hour of execution.
+        private Timer _twinTimeoutTimer;
+
+        private static TimeSpan s_twinResponseTimeout = TimeSpan.FromMinutes(60);
+
+        // Used to correlate back to a received message when the user wants to acknowledge it. This is not a value
+        // that is sent over the wire, so we increment this value locally instead.
+        private int _nextLockToken;
+
```
drwill-ms, line 110: nit: extra newline here

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -98,6 +105,9 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
             {
                 using var ctb = new CancellationTokenBundle(_operationTimeout, cancellationToken);
                 await _amqpUnit.OpenAsync(ctb.Token).ConfigureAwait(false);
+
+                // Create a timer to remove any expired messages. The timer would invoke callback after every hour.
+                _twinTimeoutTimer = new Timer(CheckTimeout, null, s_twinResponseTimeout, s_twinResponseTimeout);
```
drwill-ms, line 110: I feel like we should tie the lifetime to open/close instead, and double check disposal in Dispose.

That way it doesn't fire when we're not connected. Thoughts?

tmahmood-microsoft: Agree, didn't consider it from this point of view earlier.
Added change.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
tmahmood-microsoft: good to know.

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
drwill-ms, line 110: We're leaking here. Can you initialize an instance in ctor, and then use the `Change` method to get it going here?

tmahmood-microsoft: Good catch. Thanks.
Added the change.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -646,13 +662,32 @@ public override async Task<long> SendTwinPatchAsync(TwinCollection reportedPrope
...
+                foreach (KeyValuePair<string, DateTimeOffset> entry in _twinResponseTimeouts)
+                {
+                    if (currentDateTime - entry.Value > s_twinResponseTimeout)
+                    {
+                        _getTwinResponseCompletions.TryRemove(entry.Key, out TaskCompletionSource<GetTwinResponse> _);
+                        _reportedPropertyUpdateResponseCompletions.TryRemove(entry.Key, out TaskCompletionSource<PatchTwinResponse> _);
+                        _twinResponseTimeouts.TryRemove(entry.Key, out DateTimeOffset _);
+                    }
+                }
             }
```
drwill-ms, line 683: If you handle error cases early, then the rest of the code can be outdented (line length doesn't get impacted as much.
```suggestion
            if (!_twinResponseTimeouts.Any())
            {
                return;
            }
            
            var currentDateTime = DateTime.UtcNow;
            foreach (KeyValuePair<string, DateTimeOffset> entry in _twinResponseTimeouts)
            {
                if (currentDateTime - entry.Value > s_twinResponseTimeout)
                {
                    _getTwinResponseCompletions.TryRemove(entry.Key, out TaskCompletionSource<GetTwinResponse> _);
                    _reportedPropertyUpdateResponseCompletions.TryRemove(entry.Key, out TaskCompletionSource<PatchTwinResponse> _);
                    _twinResponseTimeouts.TryRemove(entry.Key, out DateTimeOffset _);
                }
            }
```

drwill-ms (approved):

PR #2741, opened 2022-09-23
Revamp exceptions thrown by the dps device - part2

brycewang-microsoft:
+ rename `ProvisioningTransportException` to `DeviceProvisioningClientException` to align naming format of exception thrown by the hub device & service clients
+ add "StatusCode" property into `DeviceProvisioningClientException`
+ remove unused ctors from `DeviceProvisioningClientException`


iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -209,10 +209,6 @@ private static void AddCustomHeaders(HttpRequestMessage requestMessage, IDiction
             {
                 throw new IotHubClientException(ex.Message, IotHubStatusCode.NetworkErrors, ex);
             }
-            catch (OperationCanceledException)
```
brycewang-microsoft, line 212: catching and just throwing again is weird, so just not catching anymore

provisioning/device/src/DeviceProvisioningClientException.cs
```diff
@@ -2,107 +2,62 @@
...
 using System.Runtime.Serialization;
 
 namespace Microsoft.Azure.Devices.Provisioning.Client
 {
     /// <summary>
-    /// Represents errors reported by the Provisioning Transport Handlers.
+    /// The exception that is thrown when an error occurs during device provisioning client operation.
     /// </summary>
-    public class ProvisioningTransportException : Exception
+    public class DeviceProvisioningClientException : Exception
```
brycewang-microsoft, line 13: Do we like the naming here `DeviceProvisioningClientException`? Maybe I should change this to `ProvisioningDeviceClientException` to align the naming style for dps device & service clients?

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -185,7 +185,7 @@ internal IotHubClientException(bool isTransient, IotHubStatusCode statusCode) :
         public string TrackingId { get; set; }
 
         /// <summary>
-        /// The HTTP status statusCode returned by the IoT hub service.
+        /// The HTTP status code returned by the IoT hub service.
         /// </summary>
         public IotHubStatusCode StatusCode { get; internal set; }
```
drwill-ms, line 190: ErrorCode to match service SDK?

provisioning/device/src/DeviceProvisioningClientException.cs
drwill-ms: Can modules register with DPS, @timtay-microsoft / @abhipsaMisra ?

provisioning/device/src/DeviceProvisioningClientException.cs
```diff
@@ -127,6 +82,11 @@ protected internal ProvisioningTransportException(SerializationInfo info, Stream
         /// </summary>
         public ProvisioningErrorDetails ErrorDetails { get; private set; }
```
drwill-ms, line 83: Odd that on provisioning we have this other class, whereas on hub we have the properties directly on the exception. We should look into making them look the same.

iothub/device/src/Transport/Http/HttpClientHelper.cs
drwill-ms: As long as it doesn't get caught by the Exception catch block below.

iothub/device/src/Exceptions/IotHubClientException.cs
brycewang-microsoft: No, this is property in `IotHubClientException` thrown by the **hub device**. The changes here is to fix the doc comments only.

drwill-ms: And the code will be simply 200, 400, 500? Or the 6-digit kind?

andyk-ms: Discussed with Bryce offline and we need to gather the code from DPS which are 6-digits.

brycewang-microsoft: > And the code will be simply 200, 400, 500? Or the 6-digit kind?

it's more like 6-digit kind. FYI, I am revamping this `IotHubClientException` to be more aligned with the `IotHubServiceException` by using 2 properties `StatusCode`(3-digit) and `ErrorCode`(6-digit).

brycewang-microsoft: "revert" the changes related to "hub client exceptions" in this PR as I am having another PR for `IotHubClientException` in particular.

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -515,16 +515,14 @@ public async Task DPS_Registration_MqttWs_SymmetricKey_GroupEnrollment_CustomAll
...
-            catch (ProvisioningTransportException ex)
-            {
-                // Exception message must contain the errorCode value as below
-                Assert.IsTrue(ex.Message.Contains("404201"));
-            }
+            // act
+            Func<Task> act = async () => await ProvisioningDeviceClient_InvalidRegistrationId_TpmRegister_Fail(new IotHubClientAmqpSettings()).ConfigureAwait(false);
+
+            // assert
+            var error = await act.Should().ThrowAsync<DeviceProvisioningClientException>();
```
drwill-ms, line 522:
```suggestion
            var error = await act.Should().ThrowAsync<DeviceProvisioningClientException>().ConfigureAwait(false);
```

provisioning/device/src/DeviceProvisioningClientException.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        /// <param name="info">The SerializationInfo that holds the serialized object data about the exception being thrown.</param>
+        /// <param name="context">The StreamingContext that contains contextual information about the source or destination.</param>
+        public override void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            base.GetObjectData(info, context);
+            info.AddValue(IsTransientValueSerializationStoreName, IsTransient);
+        }
+
+        private bool DetermineIfTransient(HttpStatusCode statusCode)
```
drwill-ms, line 102: When is this getting called?

drwill-ms (approved):

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
```diff
@@ -52,61 +53,61 @@ public async Task ProvisioningServiceClient_QueryInvalidServiceCertificateHttp_F
         public async Task ProvisioningDeviceClient_RegisterAsyncInvalidServiceCertificateAmqpTcp_Fails()
         {
             using var transport = new ProvisioningTransportHandlerAmqp(ProvisioningClientTransportProtocol.Tcp);
-            ProvisioningTransportException exception = await Assert.ThrowsExceptionAsync<ProvisioningTransportException>(
-                () => TestInvalidServiceCertificate(transport)).ConfigureAwait(false);
+            Func<Task> act = async () => await TestInvalidServiceCertificate(transport);
 
-            Assert.IsInstanceOfType(exception.InnerException, typeof(AuthenticationException));
+            var error = await act.Should().ThrowAsync<DeviceProvisioningClientException>().ConfigureAwait(false);
```
patilsnr, line 58: I'm new to our fluent assertions, but is there a reason we're using `var` instead of an explicit type here? Would the actual type just be too verbose and hard to read?

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
```diff
@@ -52,61 +53,61 @@ public async Task ProvisioningServiceClient_QueryInvalidServiceCertificateHttp_F
...
         [LoggedTestMethod, Timeout(TestTimeoutMilliseconds)]
         public async Task ProvisioningDeviceClient_RegisterAsyncInvalidServiceCertificateMqttTcp_Fails()
         {
             using var transport = new ProvisioningTransportHandlerMqtt(ProvisioningClientTransportProtocol.Tcp);
-            ProvisioningTransportException exception = await Assert.ThrowsExceptionAsync<ProvisioningTransportException>(
-                () => TestInvalidServiceCertificate(transport)).ConfigureAwait(false);
+            Func<Task> act = async () => await TestInvalidServiceCertificate(transport);
 
-            if (exception.InnerException == null)
+            var error = await act.Should().ThrowAsync<DeviceProvisioningClientException>().ConfigureAwait(false);
```
patilsnr, line 68: same comment here

patilsnr (approved):
I like the new name! I'd wait until also getting a sign off from someone who knows this part of our code a bit better than I do, though.

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
brycewang-microsoft: Yes. In other words, it's not necessary to specify the exception type again as we already include it with `ThrowAsync<>`.

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
brycewang-microsoft: same as above

PR #2742, opened 2022-09-23
Reduce delay from 5 to 1 second, remove initial operation call, remov…

drwill-ms:
Reduce delay from 5 to 1 second, remove initial operation call, remove checking after close of device client, and remove wait after

Fault injection tests down to 10 minutes (from 18).

### DIFF
--- a/e2e/test/helpers/templates/FaultInjection.cs
+++ b/e2e/test/helpers/templates/FaultInjection.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
 {
     public static class FaultInjection
     {
-        public static readonly TimeSpan DefaultFaultDelay = TimeSpan.FromSeconds(5); // Time in seconds after service initiates the fault.
+        public static readonly TimeSpan DefaultFaultDelay = TimeSpan.FromSeconds(1); // Time in seconds after service initiates the fault.
         public static readonly TimeSpan DefaultFaultDuration = TimeSpan.FromSeconds(5); // Duration in seconds
         public static readonly TimeSpan LatencyTimeBuffer = TimeSpan.FromSeconds(10); // Buffer time waiting fault occurs or connection recover
 
@@ -32,8 +32,8 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
         public static Client.Message ComposeErrorInjectionProperties(
             string faultType,
             string reason,
-            TimeSpan delayInSecs,
-            TimeSpan durationInSecs)
+            TimeSpan faultDelay,
+            TimeSpan faultDuration)
         {
             return new Client.Message(Encoding.UTF8.GetBytes(Guid.NewGuid().ToString()))
             {
@@ -41,8 +41,8 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
                 {
                     ["AzIoTHub_FaultOperationType"] = faultType,
                     ["AzIoTHub_FaultOperationCloseReason"] = reason,
-                    ["AzIoTHub_FaultOperationDelayInSecs"] = delayInSecs.TotalSeconds.ToString(CultureInfo.InvariantCulture),
-                    ["AzIoTHub_FaultOperationDurationInSecs"] = durationInSecs.TotalSeconds.ToString(CultureInfo.InvariantCulture)
+                    ["AzIoTHub_FaultOperationDelayInSecs"] = faultDelay.TotalSeconds.ToString(CultureInfo.InvariantCulture),
+                    ["AzIoTHub_FaultOperationDurationInSecs"] = faultDuration.TotalSeconds.ToString(CultureInfo.InvariantCulture)
                 }
             };
         }
@@ -62,12 +62,12 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
             Client.TransportType transport,
             string faultType,
             string reason,
-            TimeSpan delayInSec,
-            TimeSpan durationInSec,
+            TimeSpan faultDelay,
+            TimeSpan faultDuration,
             DeviceClient deviceClient,
             MsTestLogger logger)
         {
-            logger.Trace($"{nameof(ActivateFaultInjectionAsync)}: Requesting fault injection type={faultType} reason={reason}, delay={delayInSec}s, duration={DefaultFaultDuration}s");
+            logger.Trace($"{nameof(ActivateFaultInjectionAsync)}: Requesting fault injection type={faultType} reason={reason}, delay={faultDelay}s, duration={DefaultFaultDuration}s");
 
             uint oldTimeout = deviceClient.OperationTimeoutInMilliseconds;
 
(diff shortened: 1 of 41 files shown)


e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -153,29 +153,14 @@ public static bool FaultShouldDisconnect(string faultType)
 
             try
             {
-                await deviceClient.OpenAsync().ConfigureAwait(false);
```
drwill-ms, line 156: The test helpers always call Open.

e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -153,29 +153,14 @@ public static bool FaultShouldDisconnect(string faultType)
...
-                if (transport != Client.TransportType.Http1)
```
drwill-ms, line 157: Remove spurious validation not really related to the test.

e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -153,29 +153,14 @@ public static bool FaultShouldDisconnect(string faultType)
...
-                        1,
-                        $"The expected connection status change should be equal or greater than 1 but was {connectionStatusChangeCount}");
-
-                    lastConnectionStatus.Should().Be(ConnectionStatus.Connected, $"Reason: {lastConnectionStatusChangeReason}");
-                    lastConnectionStatusChangeReason.Should().Be(ConnectionStatusChangeReason.Connection_Ok);
-                }
-
                 await initOperation(deviceClient, testDevice).ConfigureAwait(false);
 
-                logger.Trace($"{nameof(FaultInjection)} Testing baseline");
```
drwill-ms, line 170: Remove initial test operation. We have other tests for that.

e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -235,42 +220,13 @@ public static bool FaultShouldDisconnect(string faultType)
                 }
 
                 await deviceClient.CloseAsync().ConfigureAwait(false);
-
```
drwill-ms, line 238: Remove after close validation. Other tests should validate that.

e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -235,42 +220,13 @@ public static bool FaultShouldDisconnect(string faultType)
...
-                }
             }
             finally
             {
                 await cleanupOperation().ConfigureAwait(false);
 
                 deviceClient.Dispose();
                 await testDevice.RemoveDeviceAsync().ConfigureAwait(false);
-
-                faultInjectionDuration.Stop();
```
drwill-ms, line 266: Waiting for the remaining duration doesn't seem to matter. These devices are deleted just after, so it shouldn't matter.

e2e/test/iothub/FileUploadFaultInjectionTests.cs
```diff
@@ -92,13 +95,11 @@ public async Task FileUploadSuccess_QuotaExceed_Amqp()
...
-            using (var fileStreamSource = new FileStream(filename, FileMode.Open, FileAccess.Read))
-            {
-                // UploadToBlobAsync is obsolete, added [Obsolete] attribute to suppress CS0618 message
-                Task fileUploadTask = deviceClient.UploadToBlobAsync(filename, fileStreamSource);
-                Task errorInjectionTask = SendErrorInjectionMessageAsync(deviceClient, faultType, reason, delayInSec, durationInSec);
-                await Task.WhenAll(fileUploadTask, errorInjectionTask).ConfigureAwait(false);
-            }
+            using var fileStreamSource = new FileStream(filename, FileMode.Open, FileAccess.Read);
+            // UploadToBlobAsync is obsolete, added [Obsolete] attribute to suppress CS0618 message
+            Task fileUploadTask = deviceClient.UploadToBlobAsync(filename, fileStreamSource);
```
timtay-microsoft, line 100: Why not just delete this test since the feature disappeared in v2?

drwill-ms: It is deleted in v2.

drwill-ms: The test file.

e2e/test/iothub/method/MethodFaultInjectionTests.cs
```diff
@@ -271,7 +287,7 @@ await FaultInjection
                     faultType,
                     reason,
                     FaultInjection.DefaultFaultDelay,
-                    FaultInjection.DefaultFaultDelay,
+                    FaultInjection.DefaultFaultDelay, // we want a quick one because we need time to recover
```
timtay-microsoft, line 290: Should one of these be ```DefaultFaultDuration```?

e2e/test/provisioning/ReprovisioningE2ETests.cs
```diff
@@ -58,170 +58,198 @@ public static void TestClassSetup(TestContext _)
...
+        [LoggedTestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
         public async Task ProvisioningDeviceClient_ReprovisioningBlockingWorks_AmqpWs_SymmetricKey_RegisterOk_Group()
         {
             await ProvisioningDeviceClient_ReprovisioningFlow_DoNotReprovision(Client.TransportType.Amqp_WebSocket_Only, AttestationMechanismType.SymmetricKey, EnrollmentType.Group, false).ConfigureAwait(false);
         }
 
-        [LoggedTestMethod, Timeout(TestTimeoutMilliseconds)]
+        [LoggedTestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
```
timtay-microsoft, line 252: I'm worried that splitting this attribute in main is just going to make for a headache merging into previews/v2. Is there some compelling reason to make this change in main right now?

timtay-microsoft (approved):

e2e/test/iothub/method/MethodFaultInjectionTests.cs
drwill-ms: No, this test chooses a smaller value for duration so it recovers in time. That's why I added the comment.

I too thought it was a bug and changed it, only to find it made the tests fail.

e2e/test/provisioning/ReprovisioningE2ETests.cs
drwill-ms: I'm going to make the same change in previews/v2 and do the merge. I may be a glutton for punishment, but I plan to pay the price.

drwill-ms: Combining attributes is a kind of an antipattern. It just bugs me.

e2e/test/iothub/method/MethodFaultInjectionTests.cs
timtay-microsoft: I'm more pointing out that we are passing in "DefaultFaultDelay" into what I assume is a "FaultDuration" parameter here

timtay-microsoft: I'd be fine with a magic number here with the comment you left, but otherwise it _looks_ like a test bug

drwill-ms: Yep. I just commented what looked like a test bug to me, too.

PR #2745, opened 2022-09-23
Reduce delay from 5 to 1 second, remove initial operation call, remov…

drwill-ms:
…e checking after close of device client, and remove wait after

### DIFF
--- a/e2e/test/helpers/templates/FaultInjection.cs
+++ b/e2e/test/helpers/templates/FaultInjection.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers.Templates
 {
     public static class FaultInjection
     {
-        public static readonly TimeSpan DefaultFaultDelay = TimeSpan.FromSeconds(5); // Time in seconds after service initiates the fault.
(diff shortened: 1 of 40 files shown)


e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -243,15 +213,6 @@ public static bool FaultShouldDisconnect(string faultType)
...
                 deviceClient.Dispose();
                 await testDevice.RemoveDeviceAsync().ConfigureAwait(false);
-
-                faultInjectionDuration.Stop();
-
-                TimeSpan timeToFinishFaultInjection = duration.Subtract(faultInjectionDuration.Elapsed);
-                if (timeToFinishFaultInjection > TimeSpan.Zero)
-                {
-                    logger.Trace($"{nameof(FaultInjection)}: Waiting {timeToFinishFaultInjection} to ensure that FaultInjection duration passed.");
-                    await Task.Delay(timeToFinishFaultInjection).ConfigureAwait(false);
```
abhipsaMisra, line 253: the hub-wide faults might still need this delay.

drwill-ms: Which are hub wide?

abhipsaMisra: iirc that would be quota exceeded, throttled and bad auth.

PR #2751, opened 2022-09-26
ensure mockability of service client models

patilsnr:


timtay-microsoft (approved):

timtay-microsoft:
There are other models in the service client (device, module, etc.). Are those classes already done?

patilsnr:
> There are other models in the service client (device, module, etc.). Are those classes already done?

@ngastelum-ms should be handling the rest of the service client models.

iothub/service/src/DigitalTwin/Models/DigitalTwinGetResponse.cs
```diff
@@ -14,9 +14,12 @@ public class DigitalTwinGetResponse<T>
         /// <summary>
         /// Creates an instance of this class.
         /// </summary>
+        /// <remarks>
+        /// This class can be inherited from and set by unit tests for mocking purposes.
+        /// </remarks>
         /// <param name="eTag">Weak ETag of the modified resource.</param>
         /// <param name="digitalTwin">The deserialized digital twin.</param>
-        internal DigitalTwinGetResponse(T digitalTwin, ETag eTag = default)
+        protected internal DigitalTwinGetResponse(T digitalTwin, ETag eTag = default)
```
drwill-ms, line 22: A step further would be to make the properties that are initialized via a ctor as readonly (meaning no setter at all).

drwill-ms: Like in `DigitalTwinUpdateResponse`

drwill-ms (approved):
Optimistic sign off.

PR #2759, opened 2022-09-27
Remove internal ReceiveAsync style methods for c2d messaging

timtay-microsoft:
CompleteAsync/AbandonAsync/RejectAsync methods are also removed

Also removed lock token and delivery tag fields from the message class. Now that we use callback everywhere, we don't need these correlation fields in the message itself. The MQTT/AMQP layer can correlate this message without them.

### DIFF
--- a/iothub/device/src/IotHubBaseClient.cs
+++ b/iothub/device/src/IotHubBaseClient.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Net.Sockets;
 using System.Net.WebSockets;
@@ -563,20 +564,7 @@ namespace Microsoft.Azure.Devices.Client
             if (Logging.IsEnabled)
                 Logging.Enter(this, message, nameof(OnMessageReceivedAsync));
 
-            if (message == null)
-            {
-                if (Logging.IsEnabled)
-                    Logging.Error(this, "Received a null message. Abandoning message.", nameof(OnMessageReceivedAsync));
-
-                if (_clientOptions.TransportSettings is IotHubClientMqttSettings)
-                {
-                    // MQTT does not support abandoning received messages, so we have no choice
-                    // but to complete it
-                    return MessageAcknowledgement.Complete;
-                }
-
-                return MessageAcknowledgement.Abandon;
-            }
+            Debug.Assert(message != null, "Received a null message");
 
             // Grab this semaphore so that there is no chance that the _receiveMessageCallback instance is set in between the read of the
             // item1 and the read of the item2
--- a/iothub/device/src/IotHubBaseClient.cs
+++ b/iothub/device/src/IotHubBaseClient.cs
@@ -565,6 +565,16 @@ namespace Microsoft.Azure.Devices.Client
 
             if (message == null)
             {
+                if (Logging.IsEnabled)
+                    Logging.Error(this, "Received a null message. Abandoning message.", nameof(OnMessageReceivedAsync));
+
+                if (_clientOptions.TransportSettings is IotHubClientMqttSettings)
+                {
+                    // MQTT does not support abandoning received messages, so we have no choice
+                    // but to complete it
+                    return MessageAcknowledgement.Complete;
+                }
+
                 return MessageAcknowledgement.Abandon;
             }
 
@@ -583,7 +593,7 @@ namespace Microsoft.Azure.Devices.Client
 
                 // The SDK should only receive messages when the user sets a listener, so this should never happen.
                 if (Logging.IsEnabled)
-                    Logging.Info(this, "Received a message when no listener was set. Abandoning message.", nameof(OnMessageReceivedAsync));
+                    Logging.Error(this, "Received a message when no listener was set. Abandoning message.", nameof(OnMessageReceivedAsync));
 
                 return MessageAcknowledgement.Abandon;
             }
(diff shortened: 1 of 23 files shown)


iothub/device/src/IotHubBaseClient.cs
```diff
@@ -558,14 +558,14 @@ private protected static ClientPipelineBuilder BuildPipeline()
...
-        internal async Task OnMessageReceivedAsync(Message message)
+        internal async Task<MessageAcknowledgement> OnMessageReceivedAsync(Message message)
         {
             if (Logging.IsEnabled)
                 Logging.Enter(this, message, nameof(OnMessageReceivedAsync));
 
             if (message == null)
             {
-                return;
+                return MessageAcknowledgement.Abandon;
```
drwill-ms, line 568: Will this work for MQTT?

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -558,14 +558,14 @@ private protected static ClientPipelineBuilder BuildPipeline()
             return !isFound ? default : (T)handler;
         }
 
-        internal async Task OnMessageReceivedAsync(Message message)
+        internal async Task<MessageAcknowledgement> OnMessageReceivedAsync(Message message)
         {
             if (Logging.IsEnabled)
                 Logging.Enter(this, message, nameof(OnMessageReceivedAsync));
 
             if (message == null)
```
drwill-ms, line 566: maybe combine message and callback being null.

drwill-ms (approved):

iothub/device/src/IotHubBaseClient.cs
timtay-microsoft: Good point. For MQTT, we may as well complete here instead

iothub/device/src/IotHubBaseClient.cs
timtay-microsoft: That would be a bit wonky since I'd like to log in each case what happened, so I'd rather not combine these

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -558,14 +558,24 @@ private protected static ClientPipelineBuilder BuildPipeline()
...
+                    Logging.Error(this, "Received a null message. Abandoning message.", nameof(OnMessageReceivedAsync));
+
+                if (_clientOptions.TransportSettings is IotHubClientMqttSettings)
+                {
+                    // MQTT does not support abandoning received messages, so we have no choice
+                    // but to complete it
+                    return MessageAcknowledgement.Complete;
+                }
+
+                return MessageAcknowledgement.Abandon;
```
abhipsaMisra, line 578: If there is no message received, what would complete or abandon mean in that context? 
I think it would be better that we check for null message upstream and not invoke this callback if the message is null. We could add a DebugAssert for our internal validation.

timtay-microsoft: This case should never happen, so I can switch this to a debug.assert, sure

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -665,9 +655,10 @@ public async Task<AmqpIotOutcome> SendEventAsync(Message message, CancellationTo
             }
         }
 
-        public void OnEventsReceived(Message message)
+        public Task OnEventsReceived(Message message)
         {
             _onMessageReceivedCallback?.Invoke(message);
+            return Task.CompletedTask;
```
abhipsaMisra, line 661: Why are we returning a generic completed task? If the caller expects a task in return, then shouldn't that be the invoked task?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -918,12 +911,15 @@ private async Task HandleReceivedCloudToDeviceMessageAsync(MqttApplicationMessag
...
+                MessageAcknowledgement acknowledgementType = await _messageReceivedListener.Invoke(receivedCloudToDeviceMessage);
+
+                if (acknowledgementType != MessageAcknowledgement.Complete)
+                {
+                    if (Logging.IsEnabled)
+                        Logging.Error(this, "Cannot 'reject' or 'abandon' a received message over MQTT. Message will be acknowledged as 'complete' instead.");
+                }
 
-                // note that MQTT does not support Abandon or Reject, so always Complete by acknowledging the message like this.
-                // Also note that
```
abhipsaMisra, line 926: mysterious comment

tmahmood-microsoft (approved):

iothub/device/src/Transport/Amqp/AmqpUnit.cs
timtay-microsoft: I've fixed this in my latest commit to this PR.

iothub/device/src/Messaging/Message.cs
```diff
@@ -185,25 +185,11 @@ public string ComponentName
             set => SystemProperties[MessageSystemPropertyNames.ComponentName] = value;
         }
 
-        /// <summary>
-        /// Unique string for identifying a received message when acknowledging it.
-        /// </summary>
-        internal string LockToken
```
abhipsaMisra, line 191: Since the complete "receive message - ack it" is now a single flow (as they have been combined), we don't need to hold on to the delivery tag or lock token anymore. Is that correct?

timtay-microsoft: Correct

PR #2760, opened 2022-09-27
Use the term callback for non-event-handler delegates

abhipsaMisra:
The Azure SDK team recommends the following naming conventions: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#names-of-common-types

Of interest to us is the recommendadtion for System.Delegate implementing types, i.e. the callbacks that we expose on our public API for receiving cloud-to-device "events".

<html>
<body>

System.Delegate | ✔️ DO add the suffix "EventHandler" to names of delegates that are used in events. </br>✔️ DO add the suffix "Callback" to names of delegates other than those used as event handlers. </br>❌ DO NOT add the suffix "Delegate" to a delegate. </br>
-- | --

</body>
</html>

Based on this recommendation, this PR updates our method names, param names and doc comments to use the term "callback".

### DIFF
--- a/e2e/test/helpers/TestDeviceCallbackHandler.cs
+++ b/e2e/test/helpers/TestDeviceCallbackHandler.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         public async Task SetDeviceReceiveMethodAsync<T>(string methodName, object deviceResponseJson, T expectedServiceRequestJson)
         {
             await _deviceClient.OpenAsync().ConfigureAwait(false);
-            await _deviceClient.SetMethodCallbackAsync(
+            await _deviceClient.SetDirectMethodCallbackAsync(
                 (request) =>
                 {
                     try
--- a/e2e/test/helpers/TestDeviceCallbackHandler.cs
+++ b/e2e/test/helpers/TestDeviceCallbackHandler.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         public async Task SetDeviceReceiveMethodAsync<T>(string methodName, object deviceResponseJson, T expectedServiceRequestJson)
         {
             await _deviceClient.OpenAsync().ConfigureAwait(false);
-            await _deviceClient.SetMethodHandlerAsync(
+            await _deviceClient.SetMethodCallbackAsync(
                 (request) =>
                 {
                     try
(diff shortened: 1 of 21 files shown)


timtay-microsoft (approved):

e2e/test/helpers/TestDeviceCallbackHandler.cs
```diff
@@ -50,7 +50,7 @@ public Message ExpectedMessageSentByService
         public async Task SetDeviceReceiveMethodAsync<T>(string methodName, object deviceResponseJson, T expectedServiceRequestJson)
         {
             await _deviceClient.OpenAsync().ConfigureAwait(false);
-            await _deviceClient.SetMethodHandlerAsync(
+            await _deviceClient.SetMethodCallbackAsync(
```
drwill-ms, line 53: DirectMethodCallback?

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -190,45 +190,43 @@ public async Task SendEventBatchAsync(IEnumerable<Message> messages, Cancellatio
...
+        /// A method callback can be unset by setting <paramref name="messageCallback"/> to null.
         /// </remarks>
-        /// <param name="messageHandler">The delegate to be used when a could to device message is received by the client.</param>
+        /// <param name="messageCallback">The callback to be invoked when a cloud-to-device message is received by the client.</param>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <exception cref="InvalidOperationException">Thrown if instance is not opened already.</exception>
         /// <exception cref="OperationCanceledException">Thrown when the operation has been canceled.</exception>
-        public async Task SetMessageHandlerAsync(
-            Func<Message, Task<MessageAcknowledgement>> messageHandler,
+        public async Task SetMessageCallbackAsync(
```
drwill-ms, line 204: I feel like "message" isn't descriptive enough, and we probably should say "CloudToDeviceMessage". Thoughts?

drwill-ms (approved):

e2e/test/helpers/TestDeviceCallbackHandler.cs
abhipsaMisra: I like that, sure

iothub/device/src/IotHubBaseClient.cs
abhipsaMisra: This is being used for both device and module clients, so we won't be able to use CloudTo"Device"... We could use CloudToClient, but as Tim pointed out in a previous comment, there's room for this being mis-interpreted since our service client is also a "client".

drwill-ms: SetMessageCallbackAsync is used for C2D and D2C? By the customer?

abhipsaMisra: SetMessageCallbackAsync is being used by both device and module clients to receive messages from service

drwill-ms: Oh, so in the module case, it isn't always CloudToDevice. Gotcha.

SetIncomingMessageHandlerAsync?

abhipsaMisra: I like that, yeah

PR #2761, opened 2022-09-27
Catch TpmException or TssException and rethrow them as other types

brycewang-microsoft:
For the hub device client and DPS device client, it can potentially throw `TpmException` or `TssException` if the users' TPM hardware does not support the relevant API call ([ref](https://github.com/Azure/azure-iot-sdk-csharp/blob/brycewang/inner-exception/authentication/src/AuthenticationProviderTpmHsm.cs#L38)). This PR is remapping these 2 types of exception to our new exception types - `IotHubClientException` for the hub device & `DeviceProvisioningClientException` for the DPS device.

### DIFF
--- a/iothub/device/src/Authentication/DeviceAuthenticationWithTpm.cs
+++ b/iothub/device/src/Authentication/DeviceAuthenticationWithTpm.cs
@@ -7,6 +7,7 @@ using System.Net;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Authentication;
+using Tpm2Lib;
 
 namespace Microsoft.Azure.Devices.Client
 {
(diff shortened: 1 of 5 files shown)


iothub/device/src/Authentication/DeviceAuthenticationWithTpm.cs
```diff
@@ -68,7 +69,16 @@ protected override string Sign(string requestString, string key)
                 Debug.Assert(key == null);
 
                 byte[] encodedBytes = Encoding.UTF8.GetBytes(requestString);
-                byte[] hmac = _authenticationProvider.Sign(encodedBytes);
+                byte[] hmac = Array.Empty<byte>();
+                try
+                {
+                    hmac = _authenticationProvider.Sign(encodedBytes);
+                }
+                catch (Exception ex) when (ex is TssException || ex is TpmException)
```
abhipsaMisra, line 77: I believe one of these exceptions inherits from the other, so your check can be against only the base type.

abhipsaMisra (approved):

timtay-microsoft (approved):

iothub/device/src/Authentication/DeviceAuthenticationWithTpm.cs
```diff
@@ -68,7 +69,16 @@ protected override string Sign(string requestString, string key)
...
+                {
+                    throw new IotHubClientException(ex.Message, false, ex);
```
drwill-ms, line 79: Error Code?

drwill-ms, line 79: Why not retriable?

drwill-ms (approved):
I'd like customers to expect error codes set on every exception, and we should be careful when setting isTransient to false, because the recommendation to the customer is to stop trying. So let's be sure that is really a terminal state.

PR #2763, opened 2022-09-27
When using TCS, use them appropriately

drwill-ms:
Following guidance at https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously.

I looked for opportunity to remove these, but for the most part we're stuck using them for 2 reasons:
* AMQP/MQTT pub/sub pattern requires using a TCS to complete it when the ack message comes in.
* AMQP library has some base classes abstract methods that need to be async but aren't defined that way.

### DIFF
--- a/iothub/device/src/Transport/ClientWebSocketTransport.cs
+++ b/iothub/device/src/Transport/ClientWebSocketTransport.cs
@@ -63,13 +63,14 @@ namespace Microsoft.Azure.Amqp.Transport
             Fx.AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
             args.Exception = null; // null out any exceptions
 
+            // We can't await this properly because the abstract method, defined by the AMQP layer, is not Task<bool>.
             Task taskResult = WriteImplAsync(args);
             if (WriteTaskDone(taskResult, args))
             {
                 return false;
             }
 
-            ToAsyncResult(taskResult, s_onWriteComplete, args);
+            _ = ToAsyncResult(taskResult, s_onWriteComplete, args);
             return true;
         }
 
@@ -188,13 +189,13 @@ namespace Microsoft.Azure.Amqp.Transport
 
                 succeeded = true;
             }
-            catch (WebSocketException webSocketException)
+            catch (WebSocketException ex)
             {
-                throw new IOException(webSocketException.Message, webSocketException);
+                throw new IOException(ex.Message, ex);
             }
-            catch (HttpListenerException httpListenerException)
+            catch (HttpListenerException ex)
             {
-                throw new IOException(httpListenerException.Message, httpListenerException);
+                throw new IOException(ex.Message, ex);
             }
             finally
             {
@@ -360,14 +361,14 @@ namespace Microsoft.Azure.Amqp.Transport
                     task.ContinueWith(
                         t => callback(task),
                         CancellationToken.None,
-                        TaskContinuationOptions.ExecuteSynchronously,
+                        TaskContinuationOptions.RunContinuationsAsynchronously,
                         TaskScheduler.Default);
                 }
 
                 return task;
             }
 
-            var tcs = new TaskCompletionSource<object>(state);
+            var tcs = new TaskCompletionSource<object>(state, TaskCreationOptions.RunContinuationsAsynchronously);
             task.ContinueWith(
                 _ =>
                 {
@@ -387,7 +388,7 @@ namespace Microsoft.Azure.Amqp.Transport
                     callback?.Invoke(tcs.Task);
                 },
                 CancellationToken.None,
-                TaskContinuationOptions.ExecuteSynchronously,
+                TaskContinuationOptions.RunContinuationsAsynchronously,
                 TaskScheduler.Default);
 
             return tcs.Task;
@@ -402,14 +403,14 @@ namespace Microsoft.Azure.Amqp.Transport
                     task.ContinueWith(
                         t => callback(task),
                         CancellationToken.None,
-                        TaskContinuationOptions.ExecuteSynchronously,
+                        TaskContinuationOptions.RunContinuationsAsynchronously,
                         TaskScheduler.Default);
                 }
 
                 return task;
             }
 
-            var tcs = new TaskCompletionSource<TResult>(state);
+            var tcs = new TaskCompletionSource<TResult>(state, TaskCreationOptions.RunContinuationsAsynchronously);
             task.ContinueWith(
                 _ =>
                 {
@@ -429,7 +430,7 @@ namespace Microsoft.Azure.Amqp.Transport
                     callback?.Invoke(tcs.Task);
                 },
                 CancellationToken.None,
-                TaskContinuationOptions.ExecuteSynchronously,
+                TaskContinuationOptions.RunContinuationsAsynchronously,
                 TaskScheduler.Default);
 
             return tcs.Task;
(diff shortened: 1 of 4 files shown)


iothub/device/src/Pipeline/TransportHandler.cs
```diff
@@ -18,7 +18,7 @@ protected TransportHandler(PipelineContext context, IotHubClientTransportSetting
 
         public override Task WaitForTransportClosedAsync()
         {
-            _transportShouldRetry = new TaskCompletionSource<bool>();
+            _transportShouldRetry ??= new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```
drwill-ms, line 21: This looks like it may have had a bug. If someone called this multiple times, we'd overwrite the previous TCS, and previous waiters would be left hanging.

But do we need to reset and add a new one sometime? I'm not sure.

@timtay-microsoft and @abhipsaMisra, thoughts here?

iothub/device/src/Transport/Amqp/AmqpTransportHandler.cs
```diff
@@ -121,23 +122,6 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
             }
         }
 
-        private void CheckTimeout(object _)
```
drwill-ms, line 124: Moved to private methods area, and renamed.

iothub/device/src/Transport/ClientWebSocketTransport.cs
```diff
@@ -63,13 +63,14 @@ public override bool WriteAsync(TransportAsyncCallbackArgs args)
             Fx.AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
             args.Exception = null; // null out any exceptions
 
+            // We can't await this properly because the abstract method, defined by the AMQP layer, is not Task<bool>.
             Task taskResult = WriteImplAsync(args);
             if (WriteTaskDone(taskResult, args))
```
drwill-ms, line 68: This code looks fundamentally broken to me. I really didn't touch it, but I have big concerns.

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
```diff
@@ -34,121 +38,149 @@ internal AmqpClientConnection(Uri uri, AmqpSettings amqpSettings)
...
             RemoteCertificateValidationCallback remoteCerificateValidationCallback,
             CancellationToken cancellationToken)
         {
             if (Logging.IsEnabled)
-            {
                 Logging.Enter(this, $"{nameof(AmqpClientConnection)}.{nameof(OpenAsync)}");
-            }
 
-            string hostName = _uri.Host;
+            await _connectionSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
drwill-ms, line 61: Protect open/close with a semaphore.

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
```diff
@@ -34,121 +38,149 @@ internal AmqpClientConnection(Uri uri, AmqpSettings amqpSettings)
...
-            IsConnectionClosed = false;
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, $"{nameof(AmqpClientConnection)}.{nameof(OpenAsync)}");
+            }
         }
 
-        public async Task CloseAsync(CancellationToken cancellationToken)
+        internal async Task CloseAsync(CancellationToken cancellationToken)
         {
-            AmqpConnection connection = AmqpConnection;
```
drwill-ms, line 129: I don't understand this logic. Why get the property and then close, but not set it as null after?

timtay-microsoft: What you've changed it to makes more sense to me than it did previously. I think it was really just trying to not call close if it was already null.

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
```diff
@@ -282,47 +302,17 @@ private void OnReadHeaderComplete(TransportAsyncCallbackArgs args)
...
-                return;
-            }
-
-            if (disposing)
-            {
-                // We don't know if the transport object is instantiated as a disposable or not
-                // We check and dispose if it is.
-                if (_transport is IDisposable disposable)
-                {
-                    disposable.Dispose();
```
timtay-microsoft, line 321: Are none of the transports disposable anymore?

iothub/device/src/Pipeline/TransportHandler.cs
timtay-microsoft: I don't have a lot of context on this one. It does seem like the CloseAsync() call should end when the connection has been closed, and that would spare this layer from using TCS. But that is a problem for another PR

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
drwill-ms: I moved the public methods higher in the file to be above the internal/private methods.

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
```diff
@@ -34,121 +38,149 @@ internal AmqpClientConnection(Uri uri, AmqpSettings amqpSettings)
...
-        {
-            AmqpConnection connection = AmqpConnection;
-            if (connection != null)
+            await _connectionSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+
+            try
             {
-                connection.Close();
+                await AmqpConnection.CloseAsync(cancellationToken).ConfigureAwait(false);
+                AmqpConnection = null;
```
abhipsaMisra, line 143: Was this ever being set to null anywhere else?

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
abhipsaMisra: +1

iothub/device/src/Pipeline/TransportHandler.cs
drwill-ms: I think things like this need to be reconceived a bit when we tackle open/close/open, too.

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
drwill-ms: Nope.

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
abhipsaMisra: Our transports inherit from our a delegating handler class, which is IDisposable. So both amqp and mqtt transports should be IDisposable.

abhipsaMisra: Oh, this is provisioning service, nevermind

abhipsaMisra: TransportBase (from amqp library) isn't IDisposable but our websocket implementation is

abhipsaMisra (approved):

iothub/device/src/Pipeline/TransportHandler.cs
abhipsaMisra: I agree that this entire class needs to be tackled on its own.

timtay-microsoft (approved):

PR #2764, opened 2022-09-27
Make AMQP provisioning call end if connection is lost partway through

timtay-microsoft:
Same strategy we took in MQTT. Link the user's cancellation token to a cancellation token that is cancelled when the connection is lost. Then be careful to throw ProvisioningDeviceException if only the connection loss token was cancelled


drwill-ms (approved):

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
```diff
@@ -151,7 +151,10 @@ public AmqpClientSession CreateSession()
 
         private void OnConnectionClosed(object o, EventArgs args)
         {
-            IsConnectionClosed = true;
```
timtay-microsoft, line 154: This flag wasn't used anywhere

provisioning/device/src/Transports/Amqp/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -24,10 +24,10 @@ public class ProvisioningTransportHandlerAmqp : ProvisioningTransportHandler
         // DPS will generally send a retry-after header that overrides this default value though.
         private static readonly TimeSpan s_defaultOperationPollingInterval = TimeSpan.FromSeconds(2);
 
-        private static readonly TimeSpan s_timeoutConstant = TimeSpan.FromMinutes(1);
```
timtay-microsoft, line 27: We no longer need this timeout. Users can just pass in a cancellation token

PR #2768, opened 2022-09-28
Retain the delay after some faults that are thought to be hub-wide

drwill-ms:


e2e/test/helpers/templates/FaultInjection.cs
```diff
@@ -227,6 +227,18 @@ public static bool FaultShouldDisconnect(string faultType)
 
                 deviceClient.Dispose();
                 await testDevice.RemoveDeviceAsync().ConfigureAwait(false);
+
+                if (FaultShouldDisconnect(faultType))
```
abhipsaMisra, line 231: From what I recall, the faults which were hub-wide were the faults that didn't trigger a disconnect, i.e. auth error, quota exceeded and throttling. 
In any case, I think a comment here calling the condition out would help prevent future confusion.

abhipsaMisra (approved):

e2e/test/helpers/templates/FaultInjection.cs
drwill-ms: I no, I had it backwards. Fixed.

PR #2773, opened 2022-09-28
Make models mockable

ngastelum-ms:


drwill-ms (approved):

iothub/service/src/Registry/Models/ServiceStatistics.cs
```diff
@@ -10,10 +10,19 @@ namespace Microsoft.Azure.Devices
     /// </summary>
     public class ServiceStatistics
     {
+        /// <summary>
+        /// This constructor is for deserialization and unit test mocking purposes.
+        /// </summary>
```
brycewang-microsoft, line 15: How about unifying the summaries across these classes?
```suggestion
        /// <summary>
        /// Creates an instance of this class. Provided for unit testing purposes only.
        /// </summary>
```

iothub/service/src/Registry/Models/RegistryStatistics.cs
```diff
@@ -10,22 +10,31 @@ namespace Microsoft.Azure.Devices
     /// </summary>
     public class RegistryStatistics
     {
+        /// <summary>
+        /// This constructor is for deserialization and unit test mocking purposes.
+        /// </summary>
```
brycewang-microsoft, line 15: same here
```suggestion
        /// <summary>
        /// Creates an instance of this class. Provided for unit testing purposes only.
        /// </summary>
```

brycewang-microsoft (approved):

PR #2775, opened 2022-09-28
Unify IsFatal() between the hub device and service, and remove the FatalException class

brycewang-microsoft:
Appears that we are using `FatalException` at all, so removing in this PR.


iothub/device/src/Fx/Fx.cs
```diff
@@ -44,15 +43,18 @@ public static bool IsFatal(Exception ex)
             while (ex != null)
             {
                 // FYI, CallbackException is-a FatalException
```
brycewang-microsoft, line 45: why we have this comment here but no include `CallbackException` within this method?

drwill-ms: Sounds like something worth removing.

drwill-ms (approved):

PR #2783, opened 2022-09-29
Merge main into previews/v2 part 1

timtay-microsoft:
Merging main into previews/v2 would be massive, so I'm taking this a few commits at a time


brycewang-microsoft (approved):

e2e/test/docker/readme.md
```diff
@@ -1,7 +1,6 @@
 # Docker for WSL setup for local test
 
 ## Intro
-
```
drwill-ms, line 4: there are supposed to be newlines before/after headers and lists.

provisioning/device/samples/Getting Started/X509Sample/Parameters.cs
```diff
@@ -41,6 +41,7 @@ internal class Parameters
         public string CertificatePassword { get; set; }
 
         [Option(
+            'g',
```
drwill-ms, line 44: Why give something that won't likely be changed a shortcut?

Our take has generally evolved to only having aliases for the required/commonly changed params.

drwill-ms (approved):

drwill-ms:
My comments about the code changes have more to do with how they are in main, and not about the merge commit.

PR #2786, opened 2022-09-29
Clean up IotHubConnectionString files so that each class has a distin…

tmahmood-microsoft:
Added IotHubConnectionStringParser and IotHubConnectionStringConstants.

Also, removed usage of Device/Module Id in service client connection string.


iothub/service/src/Authentication/AuthenticationMethodFactory.cs
```diff
@@ -12,17 +12,6 @@ public sealed class AuthenticationMethodFactory
...
-                    return new ServiceAuthenticationWithDeviceSharedAccessPolicyKey(iotHubConnectionStringBuilder.DeviceId, iotHubConnectionStringBuilder.SharedAccessKey);
-                }
-                if (!string.IsNullOrWhiteSpace(iotHubConnectionStringBuilder.SharedAccessSignature))
-                {
-                    return new ServiceAuthenticationWithDeviceSharedAccessPolicyToken(iotHubConnectionStringBuilder.DeviceId, iotHubConnectionStringBuilder.SharedAccessSignature);
-                }
-            }
             if (string.IsNullOrWhiteSpace(iotHubConnectionStringBuilder.SharedAccessKey))
             {
                 return new ServiceAuthenticationWithSharedAccessPolicyToken(iotHubConnectionStringBuilder.SharedAccessKeyName, iotHubConnectionStringBuilder.SharedAccessSignature);
```
abhipsaMisra, line 17: Were you able to check if ServiceAuthenticationWithSharedAccessPolicyToken is relevant for service clients? Or, are you planning to look into that as a part of a separate task?

iothub/service/src/Authentication/AuthenticationMethodFactory.cs
```diff
@@ -12,17 +12,6 @@ public sealed class AuthenticationMethodFactory
     {
         internal static IAuthenticationMethod GetAuthenticationMethod(IotHubConnectionStringBuilder iotHubConnectionStringBuilder)
         {
-            if (!string.IsNullOrWhiteSpace(iotHubConnectionStringBuilder.DeviceId))
```
abhipsaMisra, line 15: a service connection string shouldn't have a device Id, so this change looks good

iothub/service/src/Authentication/IotHubConnectionProperties.cs
```diff
@@ -33,8 +29,8 @@ protected IotHubConnectionProperties(string hostName)
 
             HostName = hostName;
             IotHubName = GetIotHubName(hostName);
-            AmqpEndpoint = new UriBuilder(AmqpsScheme, HostName, AmqpConstants.DefaultSecurePort).Uri;
-            HttpsEndpoint = new UriBuilder(HttpsEndpointPrefix, HostName).Uri;
+            AmqpEndpoint = new UriBuilder(IotHubConnectionStringConstants.AmqpsScheme, HostName, AmqpConstants.DefaultSecurePort).Uri;
+            HttpsEndpoint = new UriBuilder(IotHubConnectionStringConstants.HttpsEndpointPrefix, HostName).Uri;
```
abhipsaMisra, line 33: I'd recommend seeing if these can be moved to the amqp and http specific files directly.

abhipsaMisra: Same for AmqpAudience as well. Try to move it to the amqp specific class.

iothub/service/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -15,24 +15,11 @@ namespace Microsoft.Azure.Devices
...
-
         private static readonly TimeSpan s_regexTimeoutMilliseconds = TimeSpan.FromMilliseconds(500);
-        private static readonly Regex s_hostNameRegex = new Regex(@"[a-zA-Z0-9_\-\.]+$", RegexOptions.Compiled | RegexOptions.IgnoreCase, s_regexTimeoutMilliseconds);
-        private static readonly Regex s_sharedAccessKeyNameRegex = new Regex(@"^[a-zA-Z0-9_\-@\.]+$", RegexOptions.Compiled | RegexOptions.IgnoreCase, s_regexTimeoutMilliseconds);
-        private static readonly Regex s_sharedAccessKeyRegex = new Regex(@"^.+$", RegexOptions.Compiled | RegexOptions.IgnoreCase, s_regexTimeoutMilliseconds);
-        private static readonly Regex s_sharedAccessSignatureRegex = new Regex(@"^.+$", RegexOptions.Compiled | RegexOptions.IgnoreCase, s_regexTimeoutMilliseconds);
-        private static readonly Regex s_idRegex = new Regex(@"^[A-Za-z0-9\-:.+%_#*?!(),=@;$']{1,128}$", RegexOptions.Compiled | RegexOptions.IgnoreCase, s_regexTimeoutMilliseconds);
+        private static readonly Regex s_hostNameRegex = new(@"[a-zA-Z0-9_\-\.]+$", RegexOptions.Compiled | RegexOptions.IgnoreCase, s_regexTimeoutMilliseconds);
```
abhipsaMisra, line 19: I'd like us to not own any such validations. We can delegate that responsibility to service directly.
If we add validations like this in our client code, we'll need to keep updating them as and when service choses to update them.

iothub/service/src/Authentication/IotHubConnectionStringBuilder.cs
```diff
@@ -41,6 +28,15 @@ internal IotHubConnectionStringBuilder()
         {
         }
 
+        internal IotHubConnectionStringBuilder(string hostName, string sharedAccessKeyName, string sharedAccessKey, string sharedAccessSignature, string gatewayHostName)
```
abhipsaMisra, line 31: Is the SDK building a connection string for some scenario? If not, then we should require only a parser and not a builder. Generally, we'd expect the users to provide us with a formatted connection string that they would have simply copied off either portal or CLI.

iothub/service/src/Authentication/IotHubConnectionStringConstants.cs
```diff
@@ -0,0 +1,20 @@
...
+        internal const char ValuePairSeparator = '=';
+        internal const string HostNameSeparator = ".";
+
+        internal const string HostNamePropertyName = "HostName";
+        internal const string SharedAccessKeyNamePropertyName = "SharedAccessKeyName";
+        internal const string SharedAccessKeyPropertyName = "SharedAccessKey";
+        internal const string SharedAccessSignaturePropertyName = "SharedAccessSignature";
+        internal const string GatewayHostNamePropertyName = "GatewayHostName";
+        internal const string HttpsEndpointPrefix = "https";
+        internal const string AmqpsScheme = "amqps";
```
abhipsaMisra, line 18: same for these as well - try to get these moved to the transport class directly

iothub/service/src/Authentication/AuthenticationMethodFactory.cs
tmahmood-microsoft: Its being asserted in unit tests [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/e9511059a5a02126fd7d4354deb2919a27c76ee6/iothub/service/tests/ConnectionString/ConnectionStringTests.cs#L63)
It seemed relevant but digging more into it, looks like its not very useful.
I'll remove it

iothub/service/src/Authentication/IotHubConnectionStringBuilder.cs
tmahmood-microsoft: The SDK was not building connection string so removed the class.
The class was however being used in a couple e2e tests to retrieve host name value from the builder.
I updated the e2e tests to parse the conn string on the fly and fetch the host name value.

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -1428,6 +1428,23 @@ public static ProvisioningClientOptions CreateProvisioningClientOptionsFromName(
             return auth;
         }
 
+        private string GetHostName(string connectionString)
```
abhipsaMisra, line 1431: Since this logic is being used in more than 1 place, you could add this into a helper class

iothub/service/samples/getting started/InvokeDeviceMethod/Parameters.cs
```diff
@@ -27,20 +27,5 @@ internal class Parameters
             HelpText = "The Id of the device to receive the direct method." +
             "\nDefaults to 'MyDotnetDevice'.")]
         public string DeviceId { get; set; } = "MyDotnetDevice";
-
-        public static void ValidateConnectionString(string hubConnectionString)
-        {
-            try
-            {
-                _ = IotHubConnectionStringBuilder.Create(hubConnectionString);
```
abhipsaMisra, line 35: @drwill-ms This class is no longer public, so we can't use this validation in here directly.  We have an "IotHubConnectionStringParser" that we could make public and comment that it is only used for validation. The user isn't ever required to interact with this type otherwise though.

iothub/service/src/Authentication/IotHubConnectionProperties.cs
abhipsaMisra: Could these not be moved to the transport classes?

iothub/service/src/Authentication/IotHubConnectionStringParser.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            return new IotHubConnectionString(hostName, sharedAccessKeyName, sharedAccessKey, sharedAccessSignature);
+        }
+
+        private static void Validate(string hostName, string sharedAccessKeyName, string sharedAccessKey, string sharedAccessSignature)
+        {
+            string iotHubName = GetIotHubName(hostName);
+            if (string.IsNullOrWhiteSpace(sharedAccessKeyName))
+            {
+                throw new ArgumentException("Should specify either SharedAccessKeyName.");
```
abhipsaMisra, line 31: Is the shared access key name mandatory? The message there seems weird - either shared access key name or...?

abhipsaMisra (approved):
Added a couple of comments

iothub/service/src/Authentication/IotHubConnectionStringParser.cs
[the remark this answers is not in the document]
tmahmood-microsoft: Added the change. 
I have created a task to update device client with similar changes.
https://msazure.visualstudio.com/One/_boards/board/t/IoT-Developers%20and%20Devices-Managed%20SDKs/Backlog%20items/?workitem=15690721

tmahmood-microsoft: This message was to validate either SharedAccessKeyName or deviceId is always present. 
Updated the message.

tmahmood-microsoft:
Added code review changes.

iothub/service/src/Authentication/IotHubConnectionStringParser.cs
```diff
@@ -0,0 +1,103 @@
...
+                : hostName;
+            return iotHubName;
+        }
+
+        private static IDictionary<string, string> ToDictionary(string valuePairString, char kvpDelimiter, char kvpSeparator)
+        {
+            if (string.IsNullOrWhiteSpace(valuePairString))
+            {
+                throw new ArgumentException("Malformed token.");
+            }
```
tmahmood-microsoft, line 70: This private method is being used after validation that the argument (valuePairString) is not null. 
We have this check again to make sure any future reference to this method still validates it.

Not sure if the message in the exception is appropriate. Thoughts?

abhipsaMisra (approved):

iothub/service/src/Authentication/IotHubConnectionStringParser.cs
tmahmood-microsoft: Shared access key name is mandatory. I have checked it in Java world (service side) and also verified that it is required when creating SAS tokens.

tmahmood-microsoft: Confirmed it with Tim that we dont actually need to validate presence of shared access key name.
Removed the validation check.

iothub/service/samples/getting started/InvokeDeviceMethod/Parameters.cs
drwill-ms: Just validate it isn't null or whitespace and then rely on the client init to validate it the remainder.

drwill-ms: We don't make it required via CommandLine [Option] attribute, because it can be set via environment variable too. So checking it here is simply meant to ensure the user passed something.

tmahmood-microsoft: Updated

PR #2790, opened 2022-09-29
Multiple fixes for the hub exception, dps exception, and e2e tests

brycewang-microsoft:
What's notable in this PR is that we remove the property `StatusCode` from `IotHubClientException` as per our v2 meeting.

### DIFF
--- a/e2e/test/helpers/TestDevice.cs
+++ b/e2e/test/helpers/TestDevice.cs
@@ -122,11 +122,10 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
                         device = await serviceClient.Devices.GetAsync(requestDevice.Id).ConfigureAwait(false);
                         if (device is null)
                         {
-                            throw new IotHubServiceException($"Created device {requestDevice.Id} not yet gettable from IoT hub.")
-                            {
-                                StatusCode = HttpStatusCode.NotFound,
-                                ErrorCode = IotHubServiceErrorCode.DeviceNotFound,
-                            };
+                            throw new IotHubServiceException(
+                                $"Created device {requestDevice.Id} not yet gettable from IoT hub.",
+                                HttpStatusCode.NotFound,
+                                IotHubServiceErrorCode.DeviceNotFound);
                         }
                     },
                     s_exponentialBackoffRetryStrategy,
--- a/e2e/test/iothub/service/RegistryE2ETests.cs
+++ b/e2e/test/iothub/service/RegistryE2ETests.cs
@@ -131,7 +131,7 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
 
         [LoggedTestMethod]
         [Timeout(TestTimeoutMilliseconds)]
-        public async Task DevicesClient_AddDevices2Async_Works()
+        public async Task DevicesClient_AddDevicesAsync_Works()
         {
             // arrange
 
@@ -182,7 +182,7 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
 
         [LoggedTestMethod]
         [Timeout(TestTimeoutMilliseconds)]
-        public async Task DevicesClient_UpdateDevices2Async_Works()
+        public async Task DevicesClient_UpdateDevicesAsync_Works()
         {
             // arrange
 
@@ -232,7 +232,7 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
 
         [LoggedTestMethod]
         [Timeout(TestTimeoutMilliseconds)]
-        public async Task RegistryManager_UpdateTwins2Async_Works()
+        public async Task RegistryManager_UpdateTwinsAsync_Works()
         {
             // arrange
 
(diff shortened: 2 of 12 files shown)


e2e/test/helpers/TestDevice.cs
```diff
@@ -122,11 +122,10 @@ await RetryOperationHelper
                         device = await serviceClient.Devices.GetAsync(requestDevice.Id).ConfigureAwait(false);
                         if (device is null)
                         {
-                            throw new IotHubServiceException($"Created device {requestDevice.Id} not yet gettable from IoT hub.")
-                            {
```
brycewang-microsoft, line 126: it's weird to use property setters here, so use ctor instead

e2e/test/iothub/service/RegistryE2ETests.cs
```diff
@@ -131,7 +131,7 @@ public async Task DevicesClient_AddDeviceWithTwinWithDeviceCapabilities()
 
         [LoggedTestMethod]
         [Timeout(TestTimeoutMilliseconds)]
-        public async Task DevicesClient_AddDevices2Async_Works()
```
brycewang-microsoft, line 134: not sure what the number 2 in the test name stands for, so delete it

e2e/test/iothub/service/RegistryE2ETests.cs
```diff
@@ -311,7 +311,7 @@ public async Task DevicesClient_RemoveDevices2Async_Works()
                 error1.And.StatusCode.Should().Be(HttpStatusCode.NotFound);
                 error1.And.ErrorCode.Should().Be(IotHubServiceErrorCode.DeviceNotFound);
 
-                Func<Task> act2 = async () => await serviceClient.Devices.GetAsync(device1.Id).ConfigureAwait(false);
+                Func<Task> act2 = async () => await serviceClient.Devices.GetAsync(device2.Id).ConfigureAwait(false);
```
brycewang-microsoft, line 314: we already call `GetAsync` on device1 above (on L308), so correct this to device2 here

e2e/test/iothub/service/RegistryE2ETests.cs
drwill-ms: There was a v1 API called AddDevices2Async (because there used to be an AddDevicesAsync and they wanted to rev it without breaking the existing API).

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -172,33 +172,33 @@ internal IotHubClientException(IotHubClientErrorCode errorCode) : base()
...
         /// Indicates if the error is transient and should be retried.
         /// </summary>
-        public bool IsTransient { get; private set; }
+        public bool IsTransient { get; }
 
         /// <summary>
         /// The service returned tracking Id associated with this particular error.
         /// </summary>
-        public string TrackingId { get; set; }
+        public string TrackingId { get; internal set; }
```
drwill-ms, line 188: protected internal (everywhere it is not settable by code outside this assembly

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -35,7 +35,7 @@ public class IotHubClientException : Exception
         // the hub service, so it is not always possible to obtain the HTTP status code and a specific error code.
         // With a best effort match, we map the error codes here to what the hub service exception has, and then match 
         // an appropriate HTTP status code for each of them via the dictionary below.
-        private static readonly Dictionary<IotHubClientErrorCode, HttpStatusCode> httpStatusCodes = new()
+        private static readonly Dictionary<IotHubClientErrorCode, HttpStatusCode> s_httpStatusCodes = new()
```
drwill-ms, line 38: Looks like we can get rid of HttpStatus code, per our v2 meeting. Sorry for asking you to do this work.

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -215,7 +215,7 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
 
         private bool DetermineIfTransient(IotHubClientErrorCode errorCode)
```
drwill-ms, line 216:
```suggestion
        private static bool DetermineIfTransient(IotHubClientErrorCode errorCode)
```

iothub/service/src/Exceptions/IotHubServiceException.cs
```diff
@@ -53,23 +53,23 @@ public IotHubServiceException(string message, Exception innerException = default
...
         /// Indicates if the error is transient and should be retried.
         /// </summary>
-        public bool IsTransient { get; set; }
+        public bool IsTransient { get; }
 
         /// <summary>
         /// The service returned tracking Id associated with this particular error.
         /// </summary>
-        public string TrackingId { get; set; }
+        public string TrackingId { get; internal set; }
```
drwill-ms, line 61: if we have a setter, then protected internal

drwill-ms (approved):

iothub/device/src/Exceptions/IotHubClientException.cs
brycewang-microsoft: No worries! We can always add it back in the future if necessary, better than keeping it now but deleting it further as a breaking change.

PR #2791, opened 2022-09-29
Remove local AMQP websocket implementation in IoT hub service client

timtay-microsoft:
Just use the implementation provided by the AMQP library instead

Edit: The only problem with this proposal is that the current out-of-the-box websocket implementation doesn't allow us to configure the websocket keep alive. I've created a [PR on the AMQP library](https://github.com/Azure/azure-amqp/pull/225) to get this support added. In the meantime, we should probably hold off on merging this.


timtay-microsoft:
If we go with this approach, then I'll follow up with the IoT hub device client and the DPS device client as well.

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -95,26 +95,42 @@ internal async Task OpenAsync(CancellationToken cancellationToken)
...
+                        Host = _credential.HostName,
+                        Port = _credential.AmqpEndpoint.Port,
+                    };
+                }
 
-                        throw;
-                    }
+                var tlsTransportSettings = new TlsTransportSettings(transportSettings)
+                {
+                    TargetHost = _credential.HostName,
```
abhipsaMisra, line 118: do we need to set this in tcp settings again if we're setting it up here anyway?

abhipsaMisra (approved):

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -95,26 +95,42 @@ internal async Task OpenAsync(CancellationToken cancellationToken)
 
                 AmqpSettings amqpSettings = CreateAmqpSettings();
 
+                TransportSettings transportSettings = null; //either WebSocket or TCP
```
abhipsaMisra, line 98: @tmahmood-microsoft FYI

iothub/service/src/Amqp/AmqpConnectionHandler.cs
timtay-microsoft: We do need to set it in both places otherwise the AMQP library throws

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -95,26 +95,47 @@ internal async Task OpenAsync(CancellationToken cancellationToken)
 
                 AmqpSettings amqpSettings = CreateAmqpSettings();
 
+                AmqpTransportInitiator amqpTransportInitiator;
                 if (_useWebSocketOnly)
                 {
-                    // Try only AMQP transport over WebSocket
-                    _transport = await CreateClientWebSocketTransportAsync(cancellationToken).ConfigureAwait(false);
+                    var websocketUri = new Uri($"{AmqpsConstants.Scheme}{_credential.HostName}:{AmqpsConstants.SecurePort}{AmqpsConstants.UriSuffix}");
+                    var websocketTransportSettings = new WebSocketTransportSettings()
```
drwill-ms, line 102:
```suggestion
                    var websocketTransportSettings = new WebSocketTransportSettings
```

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -95,26 +95,47 @@ internal async Task OpenAsync(CancellationToken cancellationToken)
...
-                        throw;
-                    }
+                    amqpTransportInitiator = new AmqpTransportInitiator(amqpSettings, tlsTranpsortSettings);
+                }
+
+                try
+                {
+                    _transport = await amqpTransportInitiator.ConnectAsync(cancellationToken).ConfigureAwait(false);
+                }
+                catch (Exception e) when (!(e is AuthenticationException))
```
drwill-ms, line 133:
```suggestion
                catch (Exception ex) when (ex is not AuthenticationException)
```

drwill-ms (approved):

abhipsaMisra:
Inputs from Xin (in case it helps):

test: https://github.com/Azure/azure-amqp/blob/hotfix/test/TestCases/AmqpWebSocketTests.cs

You can use AmqpConnectionFactory for simple setup, or you can also do the same thing as the factory does (using transport settings for more control). https://github.com/Azure/azure-amqp/blob/9f78add64fea837ed377f6f9afdf611cd21fe5fa/Microsoft.Azure.Amqp/Amqp/AmqpConnectionFactory.cs#L58

see https://github.com/Azure/azure-amqp/blob/hotfix/Microsoft.Azure.Amqp/Amqp/Transport/WebSocketTransportSettings.cs for everything you can set for web socket transport

andyk-ms (approved):

brycewang-microsoft (approved):

drwill-ms:
Please remove TaskHelpers.cs along with this.

timtay-microsoft:
> Please remove TaskHelpers.cs along with this.

I'll take care of that in the PR where I remove the local websocket implementation from the DPS device client. There is no TaskHelpers class associated with the IoT hub service client (which is the package this PR affects)

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -95,26 +95,47 @@ internal async Task OpenAsync(CancellationToken cancellationToken)
...
-                    _transport = await CreateClientWebSocketTransportAsync(cancellationToken).ConfigureAwait(false);
+                    var websocketUri = new Uri($"{AmqpsConstants.Scheme}{_credential.HostName}:{AmqpsConstants.SecurePort}{AmqpsConstants.UriSuffix}");
+                    var websocketTransportSettings = new WebSocketTransportSettings
+                    {
+                        Uri = websocketUri,
+                        Proxy = _options.Proxy,
+                        SubProtocol = AmqpsConstants.Amqpwsb10,
+                    };
+
+                    amqpTransportInitiator = new AmqpTransportInitiator(amqpSettings, websocketTransportSettings);
```
abhipsaMisra, line 109: Doesn't the websocket connection require/accept the certificate information and the remote certificate validation callback? I was expecting the tls settings to be common for both.

timtay-microsoft: This appears to be a gap in feature parity between TCP and Websocket in the AMQP library. I'll investigate this a bit more

timtay-microsoft: So this is the problem that we've already discussed as a team. Currently, the AMQP library doesn't allow us to pass in the client web socket instance. Since the AMQP library targets .netstandard2.0 which doesn't support setting a remote certificate validation callback, there is no way for us to pass in just the callback for the client websocket to use.

To get around this, we can ask the AMQP library to take in the client web socket instance. That way, users can construct on themselves and set the remote certificate validation callback on that instance.

https://github.com/Azure/azure-amqp/pull/226

timtay-microsoft: We currently don't support setting this option on AMQP web sockets, though, so this PR isn't breaking anything in that regard

PR #2797, opened 2022-09-30
Simplify DPS mqtt connection loss handling

timtay-microsoft:
Move class level cancellation token to method level


drwill-ms (approved):

provisioning/device/src/Transports/Mqtt/ProvisioningTransportHandlerMqtt.cs
```diff
@@ -83,21 +80,32 @@ internal ProvisioningTransportHandlerMqtt(ProvisioningClientOptions options)
...
+                    connectionLostCancellationToken.Token);
+
+            Task HandleDisconnectionAsync(MqttClientDisconnectedEventArgs disconnectedEventArgs)
+            {
+                _connectionLossCause = disconnectedEventArgs.Exception;
+
+                if (Logging.IsEnabled)
+                    Logging.Error(this, $"MQTT connection was lost '{_connectionLossCause}'.");
+
+                // If it was an unexpected disconnect. Ignore cases when the user intentionally closes the connection.
```
abhipsaMisra, line 105: What does a user intentionally closing the connection mean this in context? Do we expect HandleDisconnectionAsync to be called even when the user-supplied cancelation token expires?

PR #2800, opened 2022-09-30
Updated e2e tests failing due to JobProperties casting issue

tmahmood-microsoft:


iothub/service/src/Registry/DevicesClient.cs
```diff
@@ -567,7 +567,7 @@ public virtual async Task<IotHubJobResponse> ImportAsync(ImportJobProperties job
 
             try
             {
-                return await CreateJobAsync<ImportJobProperties>(jobParameters, cancellationToken).ConfigureAwait(false);
```
timtay-microsoft, line 570: I'm a little confused, now. I thought the point of the ImportJobProperties class was to be returned in this method. Are we actually using the importJobProperties class anymore?

tmahmood-microsoft: Fixed the change. 
The return type of importAsync / exportAsync should be ImportJobProperties / ExportJobProperties.
GetJobAsync() should return instance of IotHubJobResponse

timtay-microsoft (approved):

drwill-ms (approved):

PR #2809, opened 2022-10-03
Allow users to set a remote certificate validation callback for the IoT hub service client

timtay-microsoft:
currently supported over HTTP and AMQP TCP. Will add AMQP WS support later

### DIFF
--- a/iothub/service/src/IotHubServiceClientOptions.cs
+++ b/iothub/service/src/IotHubServiceClientOptions.cs
@@ -4,7 +4,9 @@
 using System;
 using System.Net;
 using System.Net.Http;
+using System.Net.Security;
 using System.Security.Authentication;
+using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.Azure.Devices
 {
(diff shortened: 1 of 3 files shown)


iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -101,5 +103,28 @@ public class IotHubServiceClientOptions
         /// </para>
         /// </remarks>
         public TimeSpan AmqpConnectionKeepAlive { get; set; } = TimeSpan.FromMinutes(2);
+
+        /// <summary>
+        /// A callback for remote certificate validation.
+        /// </summary>
+        /// <remarks>
+        /// If incorrectly implemented, your device may fail to connect to IoT Hub and/or be open to security vulnerabilities.
```
drwill-ms, line 111:
```suggestion
        /// If incorrectly implemented, your device may fail to connect to IoT hub and/or be open to security vulnerabilities.
```

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -101,5 +103,28 @@ public class IotHubServiceClientOptions
...
+        /// <para>
+        /// This feature is only applicable for HTTP connections and for AMQP TCP connections. AMQP web socket communication
+        /// does not support this feature.
+        /// </para>
+        /// </remarks>
+        public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; } = DefaultRemoteCertificateValidation;
```
drwill-ms, line 117: Does this apply to both device and module client?

drwill-ms (approved):

iothub/service/src/IotHubServiceClientOptions.cs
timtay-microsoft: Yes, but MQTT also supports this in device/module client

abhipsaMisra (approved):

PR #2811, opened 2022-10-03
Remove local AMQP websocket implementation for DPS device client

timtay-microsoft:
Use the out-of-the-box implementation from the AMQP library instead

Also add remote certificate callback support for all protocols except for AMQP web socket

Edit: WebsocketTransportSettings doesn't seem to allow us to pass in client certificates, so we may not be able to do this after all. Marking as a draft while I investigate


provisioning/device/src/TransportSettings/ProvisioningClientAmqpSettings.cs
```diff
@@ -35,20 +35,5 @@ public ProvisioningClientAmqpSettings(ProvisioningClientTransportProtocol transp
         /// </para>
         /// </remarks>
         public TimeSpan IdleTimeout { get; set; } = TimeSpan.FromMinutes(2);
-
-        /// <summary>
-        /// A keep-alive for the transport layer in sending ping/pong control frames when using web sockets.
-        /// </summary>
-        /// <seealso href="https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocketoptions.keepaliveinterval"/>
-        public TimeSpan? WebSocketKeepAlive { get; set; }
```
timtay-microsoft, line 43: Moving this to the more generic TransportSettings class that this class inherits from so that every protocol gets this

provisioning/device/src/TransportSettings/ProvisioningClientAmqpSettings.cs
```diff
@@ -35,20 +35,5 @@ public ProvisioningClientAmqpSettings(ProvisioningClientTransportProtocol transp
...
-
-        /// <summary>
-        /// A callback for remote certificate validation.
-        /// </summary>
-        /// <remarks>
-        /// If incorrectly implemented, your device may fail to connect to Device Provisioning Service
-        /// and/or be open to security vulnerabilities.
-        /// </remarks>
-        public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; }
```
timtay-microsoft, line 52: Moving this to the more generic TransportSettings class that this class inherits from so that every protocol gets this

provisioning/device/src/TransportSettings/ProvisioningClientAmqpSettings.cs
timtay-microsoft, line 43: Removing this for now

provisioning/device/src/TransportSettings/ProvisioningClientTransportSettings.cs
```diff
@@ -58,6 +60,29 @@ public abstract class ProvisioningClientTransportSettings
         /// </remarks>
         public SslProtocols SslProtocols { get; set; } = SslProtocols.None;
 
+        /// <summary>
+        /// A callback for remote certificate validation.
+        /// </summary>
+        /// <remarks>
+        /// If incorrectly implemented, your device may fail to connect to IoT hub and/or be open to security vulnerabilities.
+        /// <para>
+        /// This feature is only applicable for HTTP connections and for AMQP TCP connections. AMQP web socket communication
```
timtay-microsoft, line 69:
```suggestion
        /// This feature is only applicable for HTTP connections, MQTT TCP connections, MQTT web socket connections, and for AMQP TCP connections. AMQP web socket communication
```

provisioning/device/src/Transports/Amqp/AmqpClientConnection.cs
```diff
@@ -17,6 +16,13 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
 {
     internal sealed class AmqpClientConnection : IDisposable
     {
+        private const string Amqpwsb10 = "AMQPWSB10";
+        private const string UriSuffix = "/$iothub/websocket";
+        private const string Scheme = "wss";
+        private const string Version = "13";
+        private const int WebSocketPort = 443;
+        private const int TcpPort = 5671;
```
timtay-microsoft, line 24: Deleted the constants file that these lived in before since this was the only class referencing them

timtay-microsoft:
Closing for now since we can't do x509 auth with the out-of-the-box websocket implementation

PR #2813, opened 2022-10-03
Add remote certificate validation callback to DPS device client

timtay-microsoft:
For all protocols but AMQP web socket for now.

Also simplify some parameters passed around the AMQP layer

Also remove web socket keep alive interval option for now


provisioning/device/src/TransportSettings/ProvisioningClientAmqpSettings.cs
```diff
@@ -35,20 +35,5 @@ public ProvisioningClientAmqpSettings(ProvisioningClientTransportProtocol transp
...
-        public TimeSpan? WebSocketKeepAlive { get; set; }
-
-        /// <summary>
-        /// A callback for remote certificate validation.
-        /// </summary>
-        /// <remarks>
-        /// If incorrectly implemented, your device may fail to connect to Device Provisioning Service
-        /// and/or be open to security vulnerabilities.
-        /// </remarks>
-        public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; }
```
timtay-microsoft, line 52: Moved this to the more general class this class inherits from so that all transport protocols can use this setting

provisioning/device/src/Transports/Amqp/ProvisioningTransportHandlerAmqp.cs
```diff
@@ -95,18 +95,12 @@ internal ProvisioningTransportHandlerAmqp(ProvisioningClientOptions options)
                 cancellationToken.ThrowIfCancellationRequested();
 
                 bool useWebSocket = _settings.Protocol == ProvisioningClientTransportProtocol.WebSocket;
-                var builder = new UriBuilder
```
timtay-microsoft, line 98: This URI was being picked apart later, so there was never any need to construct it in the first place.

abhipsaMisra (approved):

PR #2814, opened 2022-10-03
Deep copy client options instance for all clients

tmahmood-microsoft:


iothub/device/src/IotHubBaseClient.cs
```diff
@@ -47,11 +47,18 @@ public abstract class IotHubBaseClient : IDisposable
...
-            // Make sure client options is initialized.
-            if (iotHubClientOptions == default)
-            {
-                iotHubClientOptions = new();
-            }
+            
+            _clientOptions = iotHubClientOptions != null
+                ? new (iotHubClientOptions.TransportSettings)
+                {
+                    FileUploadTransportSettings = iotHubClientOptions.FileUploadTransportSettings,
```
timtay-microsoft, line 54: We'll need a new instance of this settings class as well

iothub/device/src/IotHubModuleClient.cs
```diff
@@ -87,14 +87,20 @@ internal IotHubModuleClient(IotHubConnectionCredentials iotHubConnectionCredenti
...
         {
-            // Make sure client options is initialized.
-            if (options == default)
-            {
-                options = new();
-            }
+            IotHubClientOptions clientOptions = options != null
+                ? new ()
+                {
+                    FileUploadTransportSettings = options.FileUploadTransportSettings,
```
timtay-microsoft, line 93: Is this used anywhere? Modules can't upload files, but they can make http requests to invoke methods

abhipsaMisra, line 93: I don't think you can do file upload from a module client, so this can be removed.
Right, @timtay-microsoft ?

iothub/service/src/IotHubServiceClient.cs
```diff
@@ -49,20 +49,28 @@ protected IotHubServiceClient()
         public IotHubServiceClient(string connectionString, IotHubServiceClientOptions options = default)
         {
             Argument.AssertNotNullOrWhiteSpace(connectionString, nameof(connectionString));
-
-            if (options == null)
-            {
-                options = new IotHubServiceClientOptions();
-            }
+            IotHubServiceClientOptions clientOptions = options != null
```
timtay-microsoft, line 52: Consider making this a private function since this clone logic is written multiple times in this file

timtay-microsoft: Maybe even have an internal method ```clone``` on each options class that returns this copy

iothub/device/src/IotHubModuleClient.cs
timtay-microsoft: We don't use this when building an HTTP client to invoke methods, right? If not, then by all means remove this line

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -47,11 +47,18 @@ public abstract class IotHubBaseClient : IDisposable
             if (Logging.IsEnabled)
                 Logging.Enter(this, iotHubClientOptions?.TransportSettings, nameof(IotHubBaseClient) + "_ctor");
 
-            // Make sure client options is initialized.
-            if (iotHubClientOptions == default)
-            {
-                iotHubClientOptions = new();
-            }
+            
+            _clientOptions = iotHubClientOptions != null
```
drwill-ms, line 51: How about making an internal method on the options class to clone it? That way it can be responsible for making a clone and knowing what all fields to copy.

drwill-ms: Along with some unit tests to ensure all the fields are cloned and not the same instance.

drwill-ms (changes requested):
Comment inline.

iothub/device/src/IotHubClientOptions.cs
```diff
@@ -89,5 +89,22 @@ public string AdditionalUserAgentInfo
         }
 
         internal virtual ProductInfo UserAgentInfo { get; } = new();
+
+        internal IotHubClientOptions Clone()
+        {
+            IotHubClientTransportSettings transport = TransportSettings is IotHubClientMqttSettings settings
+                ? settings.Clone()
+                : ((IotHubClientAmqpSettings)TransportSettings).Clone();
```
drwill-ms, line 97: Can you put the clone method on the base class so they both get it and we don't have to cast?

iothub/device/tests/IotHubDeviceClientOptionsCloneTest.cs
```diff
@@ -0,0 +1,47 @@
...
+                IdleTimeout = TimeSpan.FromSeconds(1),
+            };
+            var options = new IotHubClientOptions(settings)
+            {
+                GatewayHostName = "sampleHost",
+                SdkAssignsMessageId = SdkAssignsMessageId.WhenUnset,
+            };
+            var clone = options.Clone();
+
+            Assert.AreNotSame(options, clone);
```
drwill-ms, line 27: Fluent assertions, please.

`Should().NotBeSameAs()`

iothub/service/tests/IotHubServiceClientOptionsCloneTests.cs
```diff
@@ -0,0 +1,24 @@
...
+        [TestMethod]
+        public void IotHubServiceClientOptions()
+        {
+            var options = new IotHubServiceClientOptions()
+            {
+                Protocol = IotHubTransportProtocol.WebSocket,
+            };
+            var clone = options.Clone();
+            Assert.AreNotSame(clone, options);
+
```
drwill-ms, line 21: nit: extra newline can be removed

provisioning/device/src/ProvisioningClientOptions.cs
```diff
@@ -54,5 +54,19 @@ public string AdditionalUserAgentInfo
         }
 
         internal virtual ProductInfo UserAgentInfo { get; } = new();
+
+        internal ProvisioningClientOptions Clone()
+        {
+            ProvisioningClientTransportSettings transport = TransportSettings is ProvisioningClientAmqpSettings amqpSettings 
+                ? amqpSettings.Clone()
```
drwill-ms, line 61: Don't use ternary if more than a simple if/then/else.

provisioning/device/src/ProvisioningDeviceClient.cs
```diff
@@ -37,27 +37,18 @@ public class ProvisioningDeviceClient
                 CertificateInstaller.EnsureChainIsInstalled(x509Auth.GetAuthenticationCertificateChain());
             }
 
-            if (options == default)
-            {
-                options = new();
-            }
+            _options = options != default
+                ? options.Clone()
+                : new ();
```
drwill-ms, line 42: Seems the style is no space between new and parens.

```suggestion
                : new();
```

provisioning/device/src/TransportSettings/ProvisioningClientHttpSettings.cs
```diff
@@ -15,5 +15,15 @@ public ProvisioningClientHttpSettings()
         {
             Protocol = ProvisioningClientTransportProtocol.WebSocket;
         }
+
+        internal ProvisioningClientHttpSettings Clone()
+        {
+            return new ProvisioningClientHttpSettings()
```
drwill-ms, line 21: not: parens not needed when using initializer list.

```suggestion
            return new ProvisioningClientHttpSettings
```

provisioning/service/tests/ProvisioningServiceClientOptionsCloneTests.cs
```diff
@@ -0,0 +1,24 @@
...
+    [TestClass]
+    [TestCategory("Unit")]
+    public class ProvisioningServiceClientOptionsCloneTests
+    {
+        [TestMethod]
+        public void ProvisioningServiceClientOptions()
+        {
+            var options = new ProvisioningServiceClientOptions()
+            {
+                HttpClient = new System.Net.Http.HttpClient(),
```
drwill-ms, line 17: Move `System.Net` to usings.

drwill-ms (changes requested):
Much better than before.

Besides inline comments, I'd like to see unit tests for cloned members having all the expected values as the cloned object, to ensure we don't forget any properties.

drwill-ms (approved):

PR #2820, opened 2022-10-04
Dispose created X509 certificates

tmahmood-microsoft:
https://github.com/Azure/azure-iot-sdk-csharp/blob/672b447887df5faf0018d279d4508462a761c388/iothub/device/src/Edge/TrustBundleProvider.cs#L84

These created X509Certificate2 instances don't appear to be disposed anywhere, but they need to be.


iothub/device/src/Edge/InstalledCertificateValidator.cs
```diff
@@ -48,4 +57,4 @@ private void SetupCertificateValidation()
             }
         }
     }
-}
```
patilsnr, line 51: nit: newline at end of file

patilsnr (approved):

abhipsaMisra (approved):

iothub/device/src/Edge/CustomCertificateValidator.cs
```diff
@@ -10,11 +10,20 @@
...
 namespace Microsoft.Azure.Devices.Client
 {
-    internal class CustomCertificateValidator : ICertificateValidator
+    internal class CustomCertificateValidator : ICertificateValidator, IDisposable
     {
         private readonly IEnumerable<X509Certificate2> _certs;
         private readonly IotHubClientTransportSettings _transportSettings;
 
+        /// <inheritdoc/>
+        public void Dispose()
```
drwill-ms, line 19: methods below ctors, even if they are public and the ctor isn't.

iothub/device/src/Edge/ICertificateValidator.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.Azure.Devices.Client
 {
-    internal interface ICertificateValidator
+    internal interface ICertificateValidator : IDisposable
```
drwill-ms, line 10: No need to make this IDisposable too.

iothub/device/src/IotHubModuleClient.cs
```diff
@@ -280,6 +280,7 @@ private async Task<DirectMethodResponse> InvokeMethodAsync(Uri uri, DirectMethod
             finally
             {
                 httpClientHandler?.Dispose();
+                _certValidator.Dispose();
```
drwill-ms, line 283: Does this need NRE protection?

drwill-ms (approved):

iothub/device/src/IotHubModuleClient.cs
tmahmood-microsoft: It does not. We create an instance NullCertificateValidator in such case.

PR #2821, opened 2022-10-04
Consolidate the device and module specific IAuthenticationMethod into same classes with module Id optional

brycewang-microsoft:
The way I consolidated these IAuthenticationMethod classes was: 
+ move module-related stuff (properties, param in ctor, etc.) from **Module**AuthenticationWith* to **Device**AuthenticationWith* classes
+ then rename **Device**AuthenticationWith* to **Client**AuthenticationWith*
+ finally remove **Module**AuthenticationWith* classes.


iothub/device/src/Authentication/ClientAuthenticationWithRegistrySymmetricKey.cs
```diff
@@ -138,5 +152,10 @@ private void SetDeviceId(string deviceId)
 
             _deviceId = deviceId;
         }
+
+        private void SetModuleId(string moduleId)
+        {
+            _moduleId = moduleId;
```
brycewang-microsoft, line 158: not checking null here for module Id as it's optional while passing to the ctor

iothub/device/tests/Authentication/ClientAuthenticationWithTokenRefreshTests.cs
```diff
@@ -109,50 +148,62 @@ public async Task DeviceAuthenticationWithTokenRefresh_Populate_DefaultParameter
...
 
         [TestMethod]
-        public async Task DeviceAuthenticationWithTokenRefresh_GetTokenAsync_NewTtl_Ok()
+        public async Task ClientAuthenticationWithSakRefresh_WithDevice_SharedAccessKeyConnectionString_HasRefresher()
         {
-            TimeSpan ttl = TimeSpan.FromSeconds(1);
-
-            var refresher = new TestImplementation(TestDeviceId, ttl, 90);
-            await refresher.GetTokenAsync(TestIotHubName).ConfigureAwait(false);
+            IConnectionCredentials iotHubConnectionCredentials = new IotHubConnectionCredentials(
```
patilsnr, line 178: consider using `var` here

iothub/device/tests/Authentication/ClientAuthenticationWithTokenRefreshTests.cs
```diff
@@ -109,50 +148,62 @@ public async Task DeviceAuthenticationWithTokenRefresh_Populate_DefaultParameter
...
+            Assert.IsNotNull(token1);
+            Assert.IsNotNull(token2);
+            Assert.AreEqual(token1, token2.TokenValue);
         }
 
         [TestMethod]
-        public async Task DeviceAuthenticationWithSakRefresh_SharedAccessKeyConnectionString_HasRefresher()
+        public async Task ClientAuthenticationWithSakRefresh_WithModule_SharedAccessKeyConnectionString_HasRefresher()
         {
             IConnectionCredentials iotHubConnectionCredentials = new IotHubConnectionCredentials(
```
patilsnr, line 201: another possible use of `var`

iothub/device/src/Authentication/ClientAuthenticationWithConnectionString.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Azure.Devices.Client
     /// <summary>
     /// Authentication method that uses the device connection string to generate SAS tokens for authenticating with service.
     /// </summary>
-    public sealed class DeviceAuthenticationWithConnectionString : IAuthenticationMethod
```
brycewang-microsoft, line 11: FYI, We didn't have ModuleAuthenticationWithConnectionString before, but this DeviceAuthenticationWithConnectionString only. However, this auth method is supposed to apply for both device and module, so rename it and fix the comments accordingly.

patilsnr (approved):

tmahmood-microsoft (approved):

iothub/device/src/Authentication/ClientAuthenticationWithRegistrySymmetricKey.cs
drwill-ms: Probably should not be whitespace, though, right?

iothub/device/src/Authentication/ClientAuthenticationWithTokenRefresh.cs
```diff
@@ -34,13 +36,20 @@ public abstract class DeviceAuthenticationWithTokenRefresh : AuthenticationWithT
...
+            ModuleId = moduleId;
         }
 
         /// <summary>
         /// Gets the DeviceId.
         /// </summary>
         public string DeviceId { get; private set; }
 
+        /// <summary>
+        /// Gets the ModuleId.
```
drwill-ms, line 49:
```suggestion
        /// Gets the module Id.
```

iothub/device/src/Authentication/ClientAuthenticationWithTokenRefresh.cs
```diff
@@ -34,13 +36,20 @@ public abstract class DeviceAuthenticationWithTokenRefresh : AuthenticationWithT
             Argument.AssertNotNullOrWhiteSpace(deviceId, nameof(deviceId));
 
             DeviceId = deviceId;
+
+            ModuleId = moduleId;
         }
 
         /// <summary>
         /// Gets the DeviceId.
```
drwill-ms, line 44:
```suggestion
        /// Gets the device Id.
```

iothub/device/src/Authentication/ClientAuthenticationWithRegistrySymmetricKey.cs
brycewang-microsoft: Yes, good catch

iothub/device/src/Authentication/ClientAuthenticationWithX509Certificate.cs
```diff
@@ -87,5 +101,10 @@ private void SetDeviceId(string deviceId)
 
             _deviceId = deviceId;
         }
+
+        private void SetModuleId(string moduleId)
```
drwill-ms, line 105: This could be pulled directly into the setter. No need for a private method.

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -262,28 +262,28 @@ private void SetTokenRefresherIfApplicable()
             {
                 if (ModuleId.IsNullOrWhiteSpace())
                 {
-                    SasTokenRefresher = new DeviceAuthenticationWithSakRefresh(
-                        DeviceId,
-                        SharedAccessKey,
-                        SharedAccessKeyName,
-                        SasTokenTimeToLive,
-                        SasTokenRenewalBuffer);
+                    SasTokenRefresher = new ClientAuthenticationWithSakRefresh(
```
drwill-ms, line 265: Why use parameter names?

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -405,24 +405,24 @@ private static IAuthenticationMethod GetAuthenticationMethodFromConnectionString
             else if (iotHubConnectionString.SharedAccessKey != null)
             {
                 return iotHubConnectionString.ModuleId == null
-                    ? new DeviceAuthenticationWithRegistrySymmetricKey(
+                    ? new ClientAuthenticationWithRegistrySymmetricKey(
+                        iotHubConnectionString.SharedAccessKey,
+                        iotHubConnectionString.DeviceId)
+                    : new ClientAuthenticationWithRegistrySymmetricKey(
```
drwill-ms, line 411: Are there 2 different ctors? If so, why not have just one with an optional module Id parameter, which you can pass in anyway. If it is null, no problem.

drwill-ms (approved):

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
brycewang-microsoft: so we don't have to pass null as `moduleId` in here. Or maybe we like to pass null?

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
brycewang-microsoft: Ah, they are using the same ctor. I missed this part, will change it.

iothub/device/tests/Authentication/ClientAuthenticationWithTokenRefreshTests.cs
brycewang-microsoft: same here

PR #2826, opened 2022-10-05
debug.writeline to SDK logger

patilsnr:


drwill-ms (changes requested):
The plan was to use our SDK logger.

PR #2827, opened 2022-10-05
Update the reconnection sample to use the retry policy fully, rather than wrap another retry policy around method calls

drwill-ms:
The additional retry policy on top of the SDK calls always bugged me. While investigating how retry policy works, I realized we can simply override it and then the SDK does what we want.

Other code cleanup. Comments inline.


iothub/device/samples/how to guides/DeviceReconnectionSample/CustomRetryPolicy.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.Azure.Devices.Client.Exceptions;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Devices.Client.Samples
+{
+    /// <summary>
+    /// An exponential backoff with jitter based retry policy that retries on additional exceptions
+    /// not covered by the SDK, because in this application we wish to run as long as possible.
+    /// </summary>
+    internal class CustomRetryPolicy : IRetryPolicy
```
drwill-ms, line 15: Rename to make it clearer that we're customizing, rather than name it what it is. We'll rely on comments to make that clearer.

iothub/device/samples/how to guides/DeviceReconnectionSample/CustomRetryPolicy.cs
```diff
@@ -0,0 +1,74 @@
...
+                || _exceptionsToBeRetried.Contains(lastException.GetType()))
+            {
+                double jitter;
+                // Because Random is not threadsafe
+                lock (_randLock)
+                {
+                    int plusOrMinus = _rng.Next(0, 2) * 2 - 1;
+
+                    // a random double from 0 to 999, positive or negative
+                    jitter =  plusOrMinus * _rng.NextDouble() * 1000;
```
drwill-ms, line 55: simpler jitter logic

iothub/device/samples/how to guides/DeviceReconnectionSample/CustomRetryPolicy.cs
```diff
@@ -0,0 +1,74 @@
...
+        internal CustomRetryPolicy(HashSet<Type> exceptionsToBeRetried, ILogger logger)
+        {
+            _exceptionsToBeRetried = exceptionsToBeRetried ?? new();
+            _logger = logger;
+        }
+
+        public bool ShouldRetry(int currentRetryCount, Exception lastException, out TimeSpan retryInterval)
+        {
+            retryInterval = TimeSpan.Zero;
+            _logger.LogInformation($"Retry requested #{currentRetryCount} exception [{lastException.GetType()}: {lastException.Message}].");
```
drwill-ms, line 35: Add logging statements to make that evident in the running of the app.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -1,43 +1,46 @@
...
         private readonly TransportType _transportType;
-        private readonly ClientOptions _clientOptions = new() { SdkAssignsMessageId = SdkAssignsMessageId.WhenUnset };
+        private static readonly ClientOptions s_clientOptions = new() { SdkAssignsMessageId = SdkAssignsMessageId.WhenUnset };
+
+        private readonly ILogger _logger;
 
         // An UnauthorizedException is handled in the connection status change handler through its corresponding status change event.
         // We will ignore this exception when thrown by client API operations.
-        private readonly Dictionary<Type, string> _exceptionsToBeIgnored = new()
+        private static readonly HashSet<Type> s_exceptionsToBeRetried = new()
```
drwill-ms, line 30: Dunno why this was a dictionary, as we only used the key.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -48,6 +51,7 @@ public class DeviceReconnectionSample
         public DeviceReconnectionSample(List<string> deviceConnectionStrings, TransportType transportType, ILogger logger)
         {
             _logger = logger;
+            _customRetryPolicy = new CustomRetryPolicy(s_exceptionsToBeRetried, _logger);
```
drwill-ms, line 54: Now passing the logger to the retry policy.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -70,38 +74,40 @@ public DeviceReconnectionSample(List<string> deviceConnectionStrings, TransportT
 
         public async Task RunSampleAsync(TimeSpan sampleRunningTime)
         {
-            s_cancellationTokenSource = new CancellationTokenSource(sampleRunningTime);
+            s_appCancellation = new CancellationTokenSource(sampleRunningTime);
```
drwill-ms, line 77: Renamed the CTS to make it clearer what it is used for, not just what it is.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -115,40 +121,29 @@ private async Task InitializeAndSetupClientAsync(CancellationToken cancellationT
...
                             }
-                            catch (UnauthorizedException) { } // if the previous token is now invalid, this call may fail
+                            catch (UnauthorizedException) { } // If the previous token is now invalid, this call may fail
                             s_deviceClient.Dispose();
                         }
 
-                        s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionStrings.First(), _transportType, _clientOptions);
+                        s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionStrings.First(), _transportType, s_clientOptions);
                         s_deviceClient.SetConnectionStatusChangesHandler(ConnectionStatusChangeHandlerAsync);
+                        s_deviceClient.SetRetryPolicy(_customRetryPolicy);
```
drwill-ms, line 130: Set retry policy.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -115,40 +121,29 @@ private async Task InitializeAndSetupClientAsync(CancellationToken cancellationT
...
                         _logger.LogDebug("Initialized the client instance.");
+
+                        // Force connection now.
+                        // We have set the "shouldExecuteOperation" function to always try to open the connection.
+                        // OpenAsync() is an idempotent call, it has the same effect if called once or multiple times on the same client.
+                        await s_deviceClient.OpenAsync(cancellationToken);
```
drwill-ms, line 136: Open should be inside the semaphore, because whether we're connected or not determines if we go in. So, another thread waiting on this semaphore should not enter after we've initialized and opened, thus changing the state.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -115,40 +121,29 @@ private async Task InitializeAndSetupClientAsync(CancellationToken cancellationT
...
                 finally
                 {
-                    _initSemaphore.Release();
+                    s_initSemaphore.Release();
                 }
 
-                // Force connection now.
-                // We have set the "shouldExecuteOperation" function to always try to open the connection.
-                // OpenAsync() is an idempotent call, it has the same effect if called once or multiple times on the same client.
-                await RetryOperationHelper.RetryTransientExceptionsAsync(
```
drwill-ms, line 135: No need for this outer retry policy being applied now.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -256,70 +251,45 @@ private async void ConnectionStatusChangeHandlerAsync(ConnectionStatus status, C
...
+            {
                 // For the purpose of this sample, we'll blindly accept all twin property write requests.
-                await RetryOperationHelper.RetryTransientExceptionsAsync(
-                    operationName: "UpdateReportedProperties",
-                    asyncOperation: async () => await s_deviceClient.UpdateReportedPropertiesAsync(reportedProperties, cancellationToken),
-                    shouldExecuteOperation: () => IsDeviceConnected,
-                    logger: _logger,
-                    exceptionsToBeIgnored: _exceptionsToBeIgnored,
-                    cancellationToken: cancellationToken);
+                await s_deviceClient.UpdateReportedPropertiesAsync(reportedProperties, s_appCancellation.Token);
```
drwill-ms, line 288: Found an issue that the SDK does not protect against exceptions thrown by this callback (and probably others). Until that is dealt with, we should at least model not letting exceptions get through, at least the one we know can happen.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -351,11 +313,6 @@ private async Task SendMessagesAsync(CancellationToken cancellationToken)
 
         private async Task ReceiveMessagesAsync(CancellationToken cancellationToken)
         {
-            var c2dReceiveExceptionsToBeIgnored = new Dictionary<Type, string>(_exceptionsToBeIgnored)
```
drwill-ms, line 354: Actually, we shouldn't be retrying CompleteAsync on this exception. It should gracefully fail, and the message will be redelivered.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -403,28 +353,33 @@ private async Task ReceiveMessageAndCompleteAsync()
                 return;
             }
 
-            try
+            using (receivedMessage)
```
drwill-ms, line 356: More elegant disposal solution

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -403,28 +353,33 @@ private async Task ReceiveMessageAndCompleteAsync()
...
-            }
-            finally
-            {
-                receivedMessage.Dispose();
+                try
+                {
+                    await s_deviceClient.CompleteAsync(receivedMessage, cancellationToken);
+                    _logger.LogInformation($"Completed message '{receivedMessage.MessageId}'.");
+                }
+                catch (DeviceMessageLockLostException)
```
drwill-ms, line 372: Ignore message lock lost.

iothub/device/samples/how to guides/DeviceReconnectionSample/ExceptionHelper.cs
```diff
@@ -29,34 +29,71 @@ internal class ExceptionHelper
...
+                if (singleException.HResult == 60)
+                {
+                    return true;
+                }
+            }
+
             return false;
         }
+
+        private static IEnumerable<Exception> Unwind(Exception exception, bool unwindAggregate = false)
```
drwill-ms, line 71: Consolidated exception helper stuff to one file.

iothub/device/samples/how to guides/DeviceReconnectionSample/ExceptionHelper.cs
```diff
@@ -29,34 +29,71 @@ internal class ExceptionHelper
...
-                (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && (singleException.HResult == unchecked((int)0x80072F8F))) ||
-                // CURLE_SSL_CACERT (60): Peer certificate cannot be authenticated with known CA certificates.
-                (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && (singleException.HResult == 60)) ||
-                singleException is AuthenticationException)
+            if (singleException is AuthenticationException)
             {
                 return true;
             }
 
+            if (s_isWindows)
```
drwill-ms, line 51: Improved for simple readability.

iothub/device/tests/RetryPolicies/ExponentialBackoffTests.cs
```diff
@@ -10,26 +11,29 @@ namespace Microsoft.Azure.Devices.Client.Test
     [TestClass]
     public class ExponentialBackoffTests
     {
-        private const int MAX_RETRY_ATTEMPTS = 5000;
+        private const int MaxRetryAttempts = 50;
```
drwill-ms, line 14: I printed out the values and it topped out after 3. With my change it takes no more than a handful still, but 50 is plenty of iterations.

iothub/device/tests/RetryPolicies/ExponentialBackoffTests.cs
```diff
@@ -10,26 +11,29 @@ namespace Microsoft.Azure.Devices.Client.Test
...
+                deltaBackoff: TimeSpan.FromMilliseconds(50));
+
+            TimeSpan previousDelay = TimeSpan.Zero;
+
             ShouldRetry shouldRetry = exponentialBackoff.GetShouldRetry();
-            for (int i = 1; i < MAX_RETRY_ATTEMPTS; i++)
+            for (int i = 1; i < MaxRetryAttempts; i++)
             {
-                shouldRetry(i, new Exception(), out TimeSpan delay);
+                shouldRetry(i, new Exception(), out TimeSpan delay).Should().BeTrue();
```
drwill-ms, line 31: Added more validation.

patilsnr (approved):

iothub/device/samples/how to guides/DeviceReconnectionSample/ExceptionHelper.cs
```diff
@@ -29,34 +29,71 @@ internal class ExceptionHelper
...
         }
 
-        internal static bool IsSecurityExceptionChain(Exception exceptionChain)
+        private static bool IsNetwork(Exception singleException)
         {
-            return exceptionChain.Unwind(true).Any(e => IsTlsSecurity(e));
+            return s_networkExceptions.Any(baseExceptionType => baseExceptionType.IsInstanceOfType(singleException));
         }
 
         private static bool IsTlsSecurity(Exception singleException)
```
timtay-microsoft, line 44: I don't love having a class like this in our sample code. Maybe we should consider making flags in our IotHubClientException class that tells the user this information.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -1,43 +1,46 @@
...
 
         // An UnauthorizedException is handled in the connection status change handler through its corresponding status change event.
         // We will ignore this exception when thrown by client API operations.
-        private readonly Dictionary<Type, string> _exceptionsToBeIgnored = new()
+        private static readonly HashSet<Type> s_exceptionsToBeRetried = new()
         {
-            { typeof(UnauthorizedException), "Unauthorized exceptions are handled by the ConnectionStatusChangeHandler." }
+            // Unauthorized exception conditions are handled by the ConnectionStatusChangeHandler in this sample and the sample will try
+            // to reconnect indefinitely, so don't give up on an operation that sees this exception.
+            typeof(UnauthorizedException),
```
timtay-microsoft, line 34: Do we need to start checking for the status code on the IotHubClientException here instead?

iothub/device/samples/how to guides/DeviceReconnectionSample/CustomRetryPolicy.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <summary>
+    /// An exponential backoff with jitter based retry policy that retries on additional exceptions
+    /// not covered by the SDK, because in this application we wish to run as long as possible.
+    /// </summary>
+    internal class CustomRetryPolicy : IRetryPolicy
+    {
+        private const int MaxRetryCount = int.MaxValue;
+        private const int MaxExponent = 22; // Avoid integer overlow (max of 30) and clamp max wait to just over 1 hour (2^22 = 1.16 hours).
+
+        private readonly Random _rng = new();
```
timtay-microsoft, line 20: This may be a bit too complex of a custom policy for 101 level samples. Is it just because you want this to be a good policy in case a user copy/pastes this?

timtay-microsoft (approved):

iothub/device/samples/how to guides/DeviceReconnectionSample/ExceptionHelper.cs
timtay-microsoft: Not in this PR though, of course

drwill-ms: I agree, but I think any improvements on this front should be saved for v2.

@brycewang-microsoft do we have any new logic in v2 to determine if transient based on what we have here?

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
drwill-ms: Can you elaborate?

iothub/device/samples/how to guides/DeviceReconnectionSample/CustomRetryPolicy.cs
drwill-ms: Pretty much. I could simply do a constant retry interval, though. Thoughts?

drwill-ms: We do use this sample with customers to tell them how to write their app. It is definitely more than a 100-level sample. That is why it is not in the getting started folder.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
abhipsaMisra: These are "additional" exceptions that will be retried in this sample. The app's retry policy already retries based on IotHubClientException.Transient.

iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs
```diff
@@ -115,40 +121,29 @@ private async Task InitializeAndSetupClientAsync(CancellationToken cancellationT
...
                         }
 
-                        s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionStrings.First(), _transportType, _clientOptions);
+                        s_deviceClient = DeviceClient.CreateFromConnectionString(_deviceConnectionStrings.First(), _transportType, s_clientOptions);
                         s_deviceClient.SetConnectionStatusChangesHandler(ConnectionStatusChangeHandlerAsync);
+                        s_deviceClient.SetRetryPolicy(_customRetryPolicy);
                         _logger.LogDebug("Initialized the client instance.");
+
+                        // Force connection now.
+                        // We have set the "shouldExecuteOperation" function to always try to open the connection.
```
abhipsaMisra, line 134: This particular statement is no longer required (the "shouldExecuteOperation" logic has been removed)

abhipsaMisra (approved):

iothub/device/samples/how to guides/DeviceReconnectionSample/ExceptionHelper.cs
brycewang-microsoft: No, I don't think so. The `IsTransient` flag in "IotHubClientException" is determined based on 
1) the `ErrorCode` when it is set 
OR 
2) how `IsTransient` was set in the original "[IotHubException](https://github.com/Azure/azure-iot-sdk-csharp/blob/283631c16e93bdcd4f63b44bd7fcc54e61e8a670/iothub/device/src/Common/Exceptions/IotHubException.cs)" which is the predecessor of "IotHubClientException"

drwill-ms: @abhipsaMisra when do we think we'd see an exception (or perhaps an inner exception) that we'd want the SDK to check for this condition and use that to set IsTransient to true?

abhipsaMisra: The only scenario I'd expect the SDK to run into a security exception is if it was thrown by the TLS stack underneath, i.e. something off about the certificate we're using, or if the certificate store wasn't accessible etc. I'd expect us to run into this during the initial connect, so if we wrap the OpenAsync call with this logic and transform it into an AuthenticationException in there, then we could get rid of this here. (Our SDK currently owns only an UnauthorizedException, but not an AuthenticationException type).

PR #2828, opened 2022-10-05
Allow user to provide ClientWebSocket instance

tmahmood-microsoft:
- This change allows user to provider custom instance of ClientWebSocket which is directly used in web socket transport layer and passed on to AMQP library. 

- We still keep the implementation to construct the instance within the SDK using the certificate, proxy and Keep Alive Interval values.

- Added doc comments to specify that remote certificate validation is not available for AMQP web socket implementation.


patilsnr (approved):

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -89,7 +90,7 @@ internal async Task<TransportBase> InitializeAsync(CancellationToken cancellatio
                     break;
 
                 default:
-                    throw new InvalidOperationException("AmqpTransportSettings must specify WebSocketOnly or TcpOnly");
+                    throw new InvalidOperationException("AmqpTransportSettings must specify either web socket or Tcp");
```
drwill-ms, line 93:
```suggestion
                    throw new InvalidOperationException("AmqpTransportSettings must specify either web socket or TCP");
```

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -139,39 +141,43 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
...
-                    Logging.Exit(this, nameof(CreateClientWebSocketAsync));
+                    Logging.Exit(this, nameof(CreateClientWebSocket));
             }
         }
 
+        /// <summary>
+        /// Remote certificate validation callback.
+        /// </summary>
+        /// <remarks>
+        /// Remote certificate validation is not supported for web socket implementation.
```
drwill-ms, line 179: Why not? Sorry, I'm slow at this.

provisioning/device/src/TransportSettings/ProvisioningClientAmqpSettings.cs
```diff
@@ -36,6 +37,14 @@ public ProvisioningClientAmqpSettings(ProvisioningClientTransportProtocol transp
         /// </remarks>
         public TimeSpan IdleTimeout { get; set; } = TimeSpan.FromMinutes(2);
 
+        /// <summary>
+        /// An instance of client web socket to be used when transport protocol is set to web socket.
+        /// </summary>
+        /// <remarks>
+        /// If not provided, an instance will be created from provided values for Proxy and Certificate.
```
drwill-ms, line 44:
```suggestion
        /// If not provided, an instance will be created from provided values for <see cref="Proxy"/> and <see cref="Certificate"/>.
```

provisioning/device/src/TransportSettings/ProvisioningClientAmqpSettings.cs
```diff
@@ -36,6 +37,14 @@ public ProvisioningClientAmqpSettings(ProvisioningClientTransportProtocol transp
...
+        /// </remarks>
+        public ClientWebSocket ClientWebSocket { get; set; }
+
         internal override ProvisioningClientTransportSettings Clone()
```
drwill-ms, line 48: Did you add this new property to Clone?

iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
```diff
@@ -8,10 +8,10 @@
...
 using System.Threading.Tasks;
+using Microsoft.Azure.Amqp;
+using Microsoft.Azure.Amqp.Transport;
 
-using Microsoft.Azure.Devices.Client;
-
-namespace Microsoft.Azure.Amqp.Transport
+namespace Microsoft.Azure.Devices.Client
 {
     internal sealed class ClientWebSocketTransport : TransportBase, IDisposable
```
drwill-ms, line 16: Will the cases of `TaskContinuationOptions.ExecuteSynchronously` be changed to `TaskContinuationOptions.RunContinuouslyAsynchronously` in this or a follow-up PR?

Also, the TCS ctors need to be changed to take the same parameter.

See https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

timtay-microsoft:
> Added doc comments to specify that remote certificate validation is not available for AMQP web socket implementation.

We'll need to call this out in the migration docs since this is a breaking change from v1

timtay-microsoft:
Did you mean to do this for both the IoT hub and DPS device clients? If so, you can take the changes I made in #2829 for the DPS side if you'd like. I didn't realize you were owning this change for that package as well.

tmahmood-microsoft:
> u mean to do this for both the IoT hub and DPS

I will revert my change to DPS client in this branch. You can keep the changes

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
tmahmood-microsoft: AMQP library does not support callback for web socket

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -139,39 +141,44 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
...
+        /// </remarks>
```
timtay-microsoft, line 181: I'd add a note explaining how to get this feature through providing a client web socket instance. See my PR #2829 for how I wrote this up

timtay-microsoft (approved):

e2e/test/iothub/DeviceClientX509AuthenticationE2ETests.cs
```diff
@@ -23,6 +25,7 @@ public class DeviceClientX509AuthenticationE2ETests : E2EMsTestBase
         private static X509Certificate2 s_selfSignedCertificateWithPrivateKey = TestConfiguration.IotHub.GetCertificateWithPrivateKey();
         private static X509Certificate2 s_chainCertificateWithPrivateKey = TestConfiguration.IotHub.GetChainDeviceCertificateWithPrivateKey();
         private readonly string _hostName = GetHostName(TestConfiguration.IotHub.ConnectionString);
+        private const string Amqpwsb10 = "AMQPWSB10";
```
abhipsaMisra, line 28: nit: const variable declarations at the top

SDK v2 migration guide.md
```diff
@@ -51,6 +51,11 @@ to migrate to version 2.x when they have the chance. For more details on LTS rel
 - Some options that were previously set in the `DeviceClient` constructor are now in the optional `IotHubClientOptions` parameter.
 - This method has been split into the three individual steps that this method used to take. See [this file upload sample](./iothub/device/samples/getting%20started/FileUploadSample/) for an example of how to do file upload using these discrete steps.
 - Cloud-to-device messages can be received by calling `SetMessageHandlerAsync` and providing a callback. Users no longer need to poll for messages with `ReceiveAsync`.
+- Remote certificate validation is no longer supported for AMQP web socket connections.
```
drwill-ms, line 54: Natively supported, right? Workaround is to set the ClientWebSocket instance in options?

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -139,36 +141,35 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
...
-                            Logging.Info(this, $"{nameof(CreateClientWebSocketAsync)} Set ClientWebSocket.Options.Proxy to {_amqpTransportSettings.Proxy}");
+                            Logging.Info(this, $"{nameof(CreateClientWebSocket)} Set ClientWebSocket.Options.Proxy to {_amqpTransportSettings.Proxy}");
                     }
                     catch (PlatformNotSupportedException)
                     {
                         // Some .NET runtimes don't support this property.
                         if (Logging.IsEnabled)
-                            Logging.Error(this, $"{nameof(CreateClientWebSocketAsync)} PlatformNotSupportedException thrown as this framework doesn't support proxy.");
+                            Logging.Error(this, $"{nameof(CreateClientWebSocket)} PlatformNotSupportedException thrown as this framework doesn't support proxy.");
+                        throw new InvalidOperationException("The current .NET runtime does not support setting the proxy.");
```
drwill-ms, line 151: Add the inner exception.

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -120,14 +122,14 @@ private async Task<TransportBase> CreateClientWebSocketTransportAsync(Cancellati
...
+        private ClientWebSocket CreateClientWebSocket()
         {
             try
             {
                 if (Logging.IsEnabled)
-                    Logging.Enter(this, nameof(CreateClientWebSocketAsync));
+                    Logging.Enter(this, nameof(CreateClientWebSocket));
 
-                var websocket = new ClientWebSocket();
+                using var websocket = new ClientWebSocket();
```
drwill-ms, line 132: If you say using, it'll be disposed after we return it. We don't want to do that.

iothub/device/src/TransportSettings/IotHubClientAmqpSettings.cs
```diff
@@ -57,9 +58,24 @@ public IotHubClientAmqpSettings(IotHubClientTransportProtocol transportProtocol
...
+        /// </para>
         /// </remarks>
         public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; }
 
+        /// <summary>
+        /// An instance of client web socket to be used when transport protocol is set to web socket.
+        /// </summary>
+        /// <remarks>
+        /// If not provided, an instance will be created from provided values for
+        /// <see cref="IotHubClientTransportSettings.Proxy"/> and <see cref="IotHubConnectionCredentials.Certificate"/>.
```
abhipsaMisra, line 75: The certificate information in IotHubConnectionCredentials isn't always set directly by the user, eg. in cases where they use an SDK implemented IAuthenticationMethod. I'd rather you just broadly specify that websocket options in transport settings will be used to initialize the ClientWebSocket, instead of calling out the specific options explicitly.

abhipsaMisra (approved):

iothub/device/src/TransportSettings/IotHubClientAmqpSettings.cs
```diff
@@ -78,6 +94,7 @@ internal override IotHubClientTransportSettings Clone()
                 PrefetchCount = PrefetchCount,
                 RemoteCertificateValidationCallback = RemoteCertificateValidationCallback,
                 ConnectionPoolSettings = ConnectionPoolSettings,
+                ClientWebSocket = ClientWebSocket,
```
drwill-ms, line 97: Will the test need to be updated to have a value here to validate?

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
abhipsaMisra, line 132: you'll want to dispose this when AmqpIotTransport is disposed

iothub/device/src/TransportSettings/IotHubClientAmqpSettings.cs
tmahmood-microsoft: The test does not necessarily need to be updated but I have still updated it, just to be on the safe,

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -139,36 +142,36 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
...
+                            Logging.Info(this, $"{nameof(CreateClientWebSocket)} Set ClientWebSocket.Options.Proxy to {_amqpTransportSettings.Proxy}");
                     }
                     catch (PlatformNotSupportedException)
                     {
+                        websocket.Dispose();
                         // Some .NET runtimes don't support this property.
                         if (Logging.IsEnabled)
-                            Logging.Error(this, $"{nameof(CreateClientWebSocketAsync)} PlatformNotSupportedException thrown as this framework doesn't support proxy.");
+                            Logging.Error(this, $"{nameof(CreateClientWebSocket)} PlatformNotSupportedException thrown as this framework doesn't support proxy.");
+                        throw new PlatformNotSupportedException("The current .NET runtime does not support setting the proxy.");
```
drwill-ms, line 153: This was closer before.

```suggestion
                        throw new InvalidOperationException("The current .NET runtime does not support setting the proxy.", ex);
```

iothub/device/src/Transport/AmqpIot/AmqpIotTransport.cs
```diff
@@ -139,36 +142,36 @@ private async Task<ClientWebSocket> CreateClientWebSocketAsync(Uri websocketUri,
                         // Configure proxy server
                         websocket.Options.Proxy = _amqpTransportSettings.Proxy;
                         if (Logging.IsEnabled)
-                            Logging.Info(this, $"{nameof(CreateClientWebSocketAsync)} Set ClientWebSocket.Options.Proxy to {_amqpTransportSettings.Proxy}");
+                            Logging.Info(this, $"{nameof(CreateClientWebSocket)} Set ClientWebSocket.Options.Proxy to {_amqpTransportSettings.Proxy}");
                     }
                     catch (PlatformNotSupportedException)
```
drwill-ms, line 147:
```suggestion
                    catch (PlatformNotSupportedException ex)
```

iothub/device/tests/IotHubDeviceClientOptionsCloneTest.cs
```diff
@@ -66,6 +67,7 @@ public void IotHubClientAmqpSettings()
                     MaxPoolSize = 120,
                     UsePooling = true,
                 },
+                ClientWebSocket = new ClientWebSocket(),
```
drwill-ms, line 70: Use local logic, not the SDK's method, and make sure it is disposed after

drwill-ms (approved):
With comments.

PR #2831, opened 2022-10-05
Adding samples to SDK pipeline

patilsnr:


vsts/gatedBuild.ps1
```diff
@@ -58,7 +58,7 @@ else
 {
 	#Likely a nightly or CI build
 	Write-Host "Not a pull request build, will run all tests"
```
drwill-ms, line 60:
```suggestion
	Write-Host "Not a pull request build, will run all tests and samples"
```

vsts/vsts.yaml
```diff
@@ -466,12 +479,17 @@ jobs:
       - powershell: ./vsts/gatedBuild.ps1
         displayName: build
         env:
-          # Environment variables for IoT Hub E2E tests
+                   # Environment variables for IoT Hub E2E tests
```
drwill-ms, line 482: Fix tabbing

vsts/vsts.yaml
```diff
@@ -494,7 +512,7 @@ jobs:
           E2E_IKEY: $(E2E-IKEY)
 
           # Environment variables for the DevOps pipeline
-          PROXY_SERVER_ADDRESS: 127.0.0.1:3128
+          PROXY_SERVER_ADDRESS: 127.0.0.1:8888
```
drwill-ms, line 515: I think we have the proxy set in a keyvault setting, but I could be wrong. Should we pull it out of there?

drwill-ms (approved):

vsts/vsts.yaml
abhipsaMisra: you'll want to check the proxy server address for both linux (docker image) and windows (squid proxy setup). If they're the same then we can put the value in keyvault. If they're different then we should hardcode it in our yaml directly and add a note explaining why.

andyk-ms: Linux and windows have different proxy and they have different ports.
For windows, we install SQUID and use the default setup. The proxy port should be 3128.
If indeed we want to change this port, squid/squid.conf need to setup with this port number.

azureiot.sln
```diff
@@ -162,40 +162,40 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SimulatedDevice", "iothub\d
...
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ComputeDerivedSymmetricKeySample", "provisioning\device\samples\Getting Started\ComputeDerivedSymmetricKeySample\ComputeDerivedSymmetricKeySample.csproj", "{A2D72CEF-7CB6-439D-A875-22DB39899969}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ComputeDerivedSymmetricKeySample", "provisioning\device\samples\getting started\ComputeDerivedSymmetricKeySample\ComputeDerivedSymmetricKeySample.csproj", "{A2D72CEF-7CB6-439D-A875-22DB39899969}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{8F9D9145-B786-4A2C-9A59-E1C33E07F5A2}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "X509Sample", "provisioning\device\samples\Getting Started\X509Sample\X509Sample.csproj", "{CAD27793-E63B-4579-B6A8-31FAA9AC8BE4}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "X509Sample", "provisioning\device\samples\getting started\X509Sample\X509Sample.csproj", "{CAD27793-E63B-4579-B6A8-31FAA9AC8BE4}"
```
patilsnr, line 169: Fix inconsistencies with directory naming, e.g. 'how to`, `How To`, `how to guides`. All folder names are now lowercase and level 2 samples are all in directories named `how to guides`

build.ps1
```diff
@@ -70,7 +71,8 @@ Param(
     [string] $framework = "*",
     [switch] $skipIotHubTests,
     [switch] $skipDPSTests,
-    [switch] $noBuildBeforeTesting
+    [switch] $noBuildBeforeTesting,
+    [switch] $runSamples
```
patilsnr, line 75: add a new function to the build script that builds and runs all samples as part of nightly runs

vsts/vsts.yaml
```diff
@@ -313,7 +318,7 @@ jobs:
           E2E_IKEY: $(E2E-IKEY)
 
           # Environment variables for the DevOps pipeline
-          PROXY_SERVER_ADDRESS: 127.0.0.1:8888
+          PROXY_SERVER_ADDRESS: 127.0.0.1:3128
```
schoims, line 321: port needs to be updated [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/ac3e71036edd0a099c5c7283628085a252d4c01a/vsts/vsts.yaml#L284)?

PR #2832, opened 2022-10-05
Continue refactoring the auth methods for the hub device and module

brycewang-microsoft:
+ consolidate `ClientAuthenticationWithSharedAccessPolicy` for the hub device and module
+ rename `ModuleAuthenticationWithHsm` to `EdgeModuleAuthenticationWithHsm`


iothub/device/src/Authentication/ClientAuthenticationWithSharedAccessPolicy.cs
```diff
@@ -49,10 +65,13 @@ public string Key
         /// <summary>
         /// Name of the shared access policy.
         /// </summary>
-        public string PolicyName
+        /// <remarks>
+        /// The name of relevant field in a connection string is "SharedAccessKeyName", so we call the property "KeyName" here.
```
brycewang-microsoft, line 69: Although the auth method here is using shared access policy, this property is supposed to be "SharedAccessKeyName" within a connection string. Also, we use "SharedAccessKeyName" everywhere else, so rename it.

iothub/device/src/Authentication/HsmAuthentication/EdgeModuleAuthenticationWithHsm.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.Azure.Devices.Client.HsmAuthentication
     /// <summary>
     /// Authentication method that uses HSM to get a SAS token.
     /// </summary>
-    internal class ModuleAuthenticationWithHsm : ClientAuthenticationWithTokenRefresh
+    internal class EdgeModuleAuthenticationWithHsm : ClientAuthenticationWithTokenRefresh
```
brycewang-microsoft, line 13: this is a special case applicable for the edge module only, so rename this class

iothub/device/src/Authentication/ClientAuthenticationWithSharedAccessPolicy.cs
abhipsaMisra, line 69: This explanation is useful for us internally. However, for the customer this naming justification isn't really required. I think it'll be useful if we point to a connection string format MS doc instead. That'll help users understand which value to pass in here.

abhipsaMisra (approved):

PR #2834, opened 2022-10-06
Allow user to pass in ClientWebSocket instance to IoT hub service client

timtay-microsoft:
We need to add the custom websocket implementation back in to accomplish this, unfortunately.

Also re-add web socket keep alive as an option for users to set

Edit: I'm running into some issues around the open/close/open scenario here. I'm not sure that client web socket instances can be re-used the way we imagined. I'm still investigating, though.


e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -16,9 +16,9 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
     [TestClass]
     [TestCategory("E2E")]
     [TestCategory("IoTHub")]
-    public class MessagingClientE2ETests : E2EMsTestBase
```
timtay-microsoft, line 19: Renaming the test to match the class it tests

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -110,22 +111,35 @@ public class IotHubServiceClientOptions
...
+        /// does not support this feature. For users who want this support over AMQP websocket, you
+        /// must instead provide a <see cref="ClientWebSocket"/> instance with the desired callback
+        /// and other websocket options (eg. proxy, keep-alive etc.) set.
         /// </para>
         /// </remarks>
         public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; } = DefaultRemoteCertificateValidation;
 
-        // The default remote certificate validation callback. It returns false if any SSL level exceptions occurred
-        // during the handshake.
-        private static bool DefaultRemoteCertificateValidation(
```
timtay-microsoft, line 121: moved this private function to the bottom

e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -16,9 +16,9 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
     [TestClass]
```
drwill-ms, line 16: rename the file, too?

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -250,6 +252,83 @@ private static AmqpSettings CreateAmqpSettings()
...
+        {
+            if (Logging.IsEnabled)
+                Logging.Enter(this, nameof(CreateClientWebSocket));
+
+            try
+            {
+                // If the user provided a client web socket instance, just use it as is.
+                if (_options.ClientWebSocket != null)
+                {
+                    return _options.ClientWebSocket;
```
drwill-ms, line 290: How do we handle disposing the client web socket?

Do we dispose it no matter what, or only if we created it? I can see customers wanting to share the instance across devices (gateway scenario).

Thoughts?

tmahmood-microsoft (approved):

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -250,6 +252,83 @@ private static AmqpSettings CreateAmqpSettings()
...
+                {
+                    websocket.Options.KeepAliveInterval = _options.WebSocketKeepAlive.Value;
+                }
+
+                // Check if we're configured to use a proxy server
+                IWebProxy webProxy = _options.Proxy;
+
+                try
+                {
+                    if (webProxy != null)
```
drwill-ms, line 308: put if outside the try block?

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -250,6 +252,83 @@ private static AmqpSettings CreateAmqpSettings()
...
+                        // Configure proxy server
+                        websocket.Options.Proxy = webProxy;
+                        if (Logging.IsEnabled)
+                            Logging.Info(this, "Setting ClientWebSocket.Options.Proxy", nameof(CreateClientWebSocket));
+                    }
+                }
+                catch (PlatformNotSupportedException)
+                {
+                    // .NET Core 2.0 doesn't support proxy. Ignore this setting.
+                    if (Logging.IsEnabled)
```
drwill-ms, line 319: I asked this question on Tahir's PR. Should we silently fail if the proxy set fails? Seems to me we should throw an exception.

iothub/service/src/Amqp/ClientWebSocketTransport.cs
```diff
@@ -0,0 +1,412 @@
...
+            ToAsyncResult(taskResult, s_onReadComplete, args);
+            return true;
+        }
+
+        public void Dispose()
+        {
+            if (!_isDisposed)
+            {
+                _webSocket.Abort();
+                _webSocket.Dispose();
```
drwill-ms, line 101: If you dispose the user's instance from options, they can't reopen the client, right?

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -110,22 +111,35 @@ public class IotHubServiceClientOptions
         /// <remarks>
         /// If incorrectly implemented, your device may fail to connect to IoT hub and/or be open to security vulnerabilities.
         /// <para>
-        /// This feature is only applicable for HTTP connections and for AMQP TCP connections. AMQP web socket communication
-        /// does not support this feature.
+        /// This feature is only applicable for HTTP and AMQP over TCP. AMQP web socket communication
+        /// does not support this feature. For users who want this support over AMQP websocket, you
+        /// must instead provide a <see cref="ClientWebSocket"/> instance with the desired callback
+        /// and other websocket options (eg. proxy, keep-alive etc.) set.
```
patilsnr, line 117: nit: `e.g.`

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -110,22 +111,35 @@ public class IotHubServiceClientOptions
...
+        /// and <see cref="FileUploadNotificationProcessorClient"/>.
+        /// </remarks>
+        /// <seealso href="https://docs.microsoft.com/dotnet/api/system.net.websockets.clientwebsocketoptions.keepaliveinterval"/>
+        public TimeSpan? WebSocketKeepAlive { get; set; }
+
+        /// <summary>
+        /// An instance of client web socket to be used when transport protocol is set to web socket.
+        /// </summary>
+        /// <remarks>
+        /// If not provided, an instance will be created from provided websocket options (eg. proxy, keep-alive etc.)
```
patilsnr, line 136: `e.g.`

patilsnr (approved):

iothub/service/src/Amqp/ClientWebSocketTransport.cs
```diff
@@ -0,0 +1,412 @@
...
+            {
+                return;
+            }
+
+            if (webSocketState == WebSocketState.Aborted
+                || webSocketState == WebSocketState.Closed
+                || webSocketState == WebSocketState.CloseReceived
+                || webSocketState == WebSocketState.CloseSent)
+            {
+                throw new ObjectDisposedException(GetType().Name);
```
drwill-ms, line 322: Have we documented this at the public API level?

iothub/service/src/Amqp/ClientWebSocketTransport.cs
```diff
@@ -0,0 +1,412 @@
...
+            }
+
+            throw new AmqpException(AmqpErrorCode.IllegalState, null);
```
drwill-ms, line 325: Will this exception type reach the end user?

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -110,22 +111,35 @@ public class IotHubServiceClientOptions
...
+        /// <para>
+        /// Only used for communications over AMQP, used in <see cref="MessagesClient"/>, <see cref="MessageFeedbackProcessorClient"/>,
+        /// and <see cref="FileUploadNotificationProcessorClient"/>.
+        /// </para>
+        /// </remarks>
+        public ClientWebSocket ClientWebSocket { get; set; }
```
drwill-ms, line 142: Who owns disposal?

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -141,5 +155,16 @@ internal IotHubServiceClientOptions Clone()
                 RemoteCertificateValidationCallback = RemoteCertificateValidationCallback,
```
drwill-ms, line 155: Needs ClientWebSocket copy.

drwill-ms (changes requested):
Mostly concerns around disposal.

iothub/service/src/Amqp/AmqpConnectionHandler.cs
timtay-microsoft: Let's add a flag to the options class that allows users to opt out of us disposing the websocket for them when they dispose the service client

timtay-microsoft: Like HttpClient has for disposing HttpClientHandlers

iothub/service/src/Amqp/AmqpConnectionHandler.cs
timtay-microsoft: Yeah we should throw here instead

iothub/service/src/Amqp/ClientWebSocketTransport.cs
timtay-microsoft: I'll get rid of this case by collapsing it with the exception above

iothub/service/src/IotHubServiceClientOptions.cs
timtay-microsoft: Us unless the user opts out

iothub/service/src/Amqp/ClientWebSocketTransport.cs
timtay-microsoft: Right. I'll make it so users can opt out of that behavior

iothub/service/src/Amqp/AmqpConnectionHandler.cs
timtay-microsoft: Allowing the PlatformNotSupportedException to bubble up to the user seems appropriate here to me. What do you think @tmahmood-microsoft ?

tmahmood-microsoft: Yeah, throwing an exception makes sense. 
I am throwing InvalidOperationException for such case in device side.

iothub/service/src/Amqp/AmqpConnectionHandler.cs
drwill-ms: I'd rather it not be a separate but related property. Do we bother making a class to hold the bool and the instance?

iothub/service/src/Amqp/AmqpConnectionHandler.cs
drwill-ms: I'd say throw InvalidOperationException with the PNSE as the inner.

iothub/service/src/Amqp/AmqpConnectionHandler.cs
timtay-microsoft: ```ClientWebSocketOptions``` class that holds them? I could get behind that

iothub/service/src/Amqp/ClientWebSocketTransport.cs
drwill-ms: So, if we put our own instance in options (if the user does not provide one) in the client ctor, we can free it optionally on dispose, rather than in this class, right?

timtay-microsoft: Why dispose the websocket at the service client level when this is the only level that uses it?

iothub/service/src/Amqp/AmqpConnectionHandler.cs
timtay-microsoft: That seems like overkill to me

timtay-microsoft: I think having a separate option here is fine as long as the docs call out that the SDK will dispose the web socket for them by default and that there is a flag they can set if they want to opt out

iothub/service/src/Amqp/ClientWebSocketTransport.cs
drwill-ms: Because it should last for the duration of the client despite how many times it is opened/closed?

timtay-microsoft: What if we re-use the web socket AMQP transport class (which holds this client web socket) for each open/close and only dispose it when the service client is disposed?

service client dispose -> amqp websocket layer dispose -> client web socket dispose

service client open -> create websocket layer if not already created

drwill-ms: If we do, then I'm fine with it.

However, it seems to me that the lifetime of the options class (and what it holds) is more congruent with the client than a single class in the transport layer.

e2e/test/iothub/service/MessageClientE2ETests.cs
```diff
@@ -16,9 +16,9 @@ namespace Microsoft.Azure.Devices.E2ETests.IotHub.Service
     [TestClass]
```
drwill-ms, line 16: Message**s**Client...

iothub/device/src/TransportSettings/IotHubClientAmqpSettings.cs
```diff
@@ -62,7 +62,7 @@ public IotHubClientAmqpSettings(IotHubClientTransportProtocol transportProtocol
         /// This feature is only applicable for HTTP, MQTT over TCP, MQTT over web socket, AMQP
         /// over TCP. AMQP web socket communication does not support this feature. For users who want
         /// this support over AMQP websocket, you must instead provide a <see cref="ClientWebSocket"/>
-        /// instance with the desired callback and other websocket options (eg. proxy, keep-alive etc.) set.
+        /// instance with the desired callback and other websocket options (e.g. proxy, keep-alive etc.) set.
```
drwill-ms, line 65: nit

```suggestion
        /// instance with the desired callback and other web socket options (e.g., proxy, keep-alive etc.) set.
```

iothub/service/src/Amqp/AmqpConnectionHandler.cs
```diff
@@ -121,19 +112,19 @@ internal async Task OpenAsync(CancellationToken cancellationToken)
...
+                    {
+                        _transport = await amqpTransportInitiator.ConnectAsync(cancellationToken).ConfigureAwait(false);
+                    }
+                    catch (Exception ex) when (ex is not AuthenticationException)
+                    {
+                        if (Logging.IsEnabled)
+                            Logging.Error(this, ex, nameof(OpenAsync));
 
-                    throw;
+                        throw;
```
drwill-ms, line 126: Will AuthenticationException be bubbled up to the user or wrapped in an IotHubServiceException?

drwill-ms: well any of the other types

iothub/service/src/Amqp/ClientWebSocketTransport.cs
```diff
@@ -0,0 +1,406 @@
...
+
+        public void Dispose()
+        {
+            if (!_isDisposed)
+            {
+                // A user may opt to not dispose the websocket just because the service client is being
+                // disposed.
+                if (_disposeClientWebSocket)
+                {
+                    _webSocket.Abort();
```
drwill-ms, line 101: What does Abort do, and do we not care about doing it if we don't own it?

iothub/service/src/Amqp/ClientWebSocketTransport.cs
```diff
@@ -0,0 +1,406 @@
...
+                }
+                else
+                {
+                    foreach (ByteBuffer byteBuffer in args.ByteBufferList)
+                    {
+                        await _webSocket
+                            .SendAsync(
+                                new ArraySegment<byte>(byteBuffer.Buffer, byteBuffer.Offset, byteBuffer.Length),
+                                WebSocketMessageType.Binary, true, _writeCancellationTokenSource.Token)
+                            .ConfigureAwait(false);
```
drwill-ms, line 182:
```suggestion
                            .SendAsync(
                                new ArraySegment<byte>(byteBuffer.Buffer, byteBuffer.Offset, byteBuffer.Length),
                                WebSocketMessageType.Binary,
                                true,
                                _writeCancellationTokenSource.Token)
                            .ConfigureAwait(false);
```

iothub/service/src/Feedback/MessageFeedbackProcessorClient.cs
```diff
@@ -98,7 +98,7 @@ protected MessageFeedbackProcessorClient()
         /// <exception cref="IotHubServiceException"> with <see cref="HttpStatusCode.RequestTimeout"/>If the client operation times out before the response is returned.</exception>
         /// <exception cref="OperationCanceledException">If the provided cancellation token has requested cancellation.</exception>
         /// <exception cref="SocketException">If a socket error occurs.</exception>
-        /// <exception cref="WebSocketException">If an error occurs when performing an operation on a WebSocket connection.</exception>
+        /// <exception cref="WebSocketException">If an error occurs when performing an operation on a web socket connection.</exception>
         /// <exception cref="IOException">If an I/O error occurs.</exception>
```
drwill-ms, line 102: I feel like all these (not ArgumentException, ArgumentNullException, InvalidOperationException, ObjectDisposedException) should be wrapped in an IotHubServiceException. @brycewang-microsoft

iothub/service/src/IotHubServiceClientOptions.cs
```diff
@@ -110,22 +111,49 @@ public class IotHubServiceClientOptions
...
+        /// <para>
+        /// Only used for communications over AMQP, used in <see cref="MessagesClient"/>, <see cref="MessageFeedbackProcessorClient"/>,
+        /// and <see cref="FileUploadNotificationProcessorClient"/>.
+        /// </para>
+        /// <para>
+        /// By default, this client web socket will be disposed when you dispose the service client.
+        /// However, you can opt out of this behavior by setting the <see cref="DisposeClientWebSocket"/> flag.
+        /// </para>
+        /// </remarks>
+        public ClientWebSocket ClientWebSocket { get; set; }
```
drwill-ms, line 146: Why have it on both the whole options class and on the AMQP transport setting class?

provisioning/device/src/Transports/Amqp/ClientWebSocketTransport.cs
```diff
@@ -20,23 +20,21 @@ internal sealed class ClientWebSocketTransport : TransportBase, IDisposable
...
 
         private readonly ClientWebSocket _webSocket;
-        private readonly EndPoint _localEndPoint;
-        private readonly EndPoint _remoteEndPoint;
+        private readonly bool _disposeClientWebSocket;
         private CancellationTokenSource _writeCancellationTokenSource;
         private bool _disposed;
 
-        public ClientWebSocketTransport(ClientWebSocket clientwebSocket, EndPoint localEndpoint, EndPoint remoteEndpoint)
+        public ClientWebSocketTransport(ClientWebSocket clientwebSocket, bool disposeClientWebSocket)
```
drwill-ms, line 27:
```suggestion
        public ClientWebSocketTransport(ClientWebSocket clientWebSocket, bool disposeClientWebSocket)
```

iothub/service/src/Amqp/AmqpConnectionHandler.cs
drwill-ms: When you just look at these two properties, it doesn't seem overkill, but complexity comes a little at a time. I'd urge you to make them cohesive.

iothub/service/src/IotHubServiceClientOptions.cs
timtay-microsoft: It's not. I'm making this change for both the IoT hub service client (above) and the DPS device client (specific to AMQP settings)

drwill-ms (approved):
Remaining comments still active.

iothub/service/src/Amqp/ClientWebSocketTransport.cs
drwill-ms: > service client dispose -> amqp websocket layer dispose -> client web socket dispose

But not if you disposed the one provided by the user, and then can't recreate it. Right?

iothub/service/src/Feedback/MessageFeedbackProcessorClient.cs
brycewang-microsoft: This appears overwhelmed by lots of PR comments so I just notice it. I am looking into this.

drwill-ms (approved):

iothub/service/src/Amqp/ClientWebSocketTransport.cs
timtay-microsoft: Right. We won't dispose the client web socket instance if the user provides it and opts out of us disposing it for them.

iothub/service/src/Amqp/ClientWebSocketTransport.cs
timtay-microsoft: It cancels any connection opened on that web socket and cancels any pending I/O operations. In the case of a client web socket that we created, it is safe to abort because it wouldn't be involved in any connections other than the AMQP connection that is being disposed here. In the case of a client web socket created by the user, we don't know if any other connections are being fed into that web socket, so we should avoid aborting it.

timtay-microsoft:
We'll re-evaluate this feature at another time

PR #2835, opened 2022-10-06
Removed HTTP support from DPS device client

tmahmood-microsoft:
- Removed support for HTTP in DPS device client.
- [minor] Using latest API version for EdgeModuleClientHelper


SDK v2 migration guide.md
```diff
@@ -175,6 +175,7 @@ to migrate to version 2.x when they have the chance. For more details on LTS rel
 - The security providers that are used in conjunction with this client have changed. See [this section](#security-provider-clients) for more details.
 - The previous way of providing transport level settings (`ProvisioningTransportHandler`) has been replaced with `ProvisioningClientTransportSettings`.
 - TPM support removed. The library used for TPM operations is broken on Linux and support for it is being shutdown. We'll reconsider how to support HSM.
+- HTTP has been removed as a transport option to keep the provisioning SDK consistent with IoT hub device SDK.
```
timtay-microsoft, line 178:
```suggestion
- HTTP has been removed as a transport option to keep the provisioning device SDK consistent with IoT hub device SDK.
```

provisioning/device/src/Transports/Amqp/ClientWebSocketTransport.cs
```diff
@@ -88,10 +88,6 @@ private async Task<bool> WriteAsyncCoreAsync(TransportAsyncCallbackArgs args)
             {
                 throw new IOException(webSocketException.Message, webSocketException);
             }
-            catch (HttpListenerException httpListenerException)
```
timtay-microsoft, line 91: This change is in the AMQP stack, FYI. I don't think we need to remove this catch

timtay-microsoft (approved):

provisioning/device/src/Utilities/ClientApiVersionHelper.cs
```diff
@@ -5,7 +5,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
 {
     internal class ClientApiVersionHelper
     {
-        // TODO: Split ApiVersionName as it is only used by HTTP and AMQP.
+        // TODO: Split ApiVersionName as it is only used by AMQP.
```
drwill-ms, line 8: Why not MQTT?

drwill-ms (approved):

PR #2838, opened 2022-10-07
Rework retry strategy and policy for v2

drwill-ms:


iothub/device/src/Authentication/HsmAuthentication/HttpHsmSignatureProvider.cs
```diff
@@ -90,15 +84,12 @@ public async Task<string> SignAsync(string moduleId, string generationId, string
...
-                .RunWithRetryAsync(() => hsmHttpClient.SignAsync(_apiVersion, moduleId, generationId, signRequest))
+            var retryHandler = new RetryHandler(s_retryPolicy);
+            return await retryHandler
+                .RunWithRetryAsync(
+                    () => hsmHttpClient.SignAsync(_apiVersion, moduleId, generationId, signRequest),
+                    (Exception ex) => ex is SwaggerException se && se.StatusCode >= 500)
                 .ConfigureAwait(false);
         }
-
-        private class ErrorDetectionStrategy : ITransientErrorDetectionStrategy
```
drwill-ms, line 99: Made a flavor of the RetryHandler API to pass code-path-specific overrides for what is transient.

iothub/device/src/RetryPolicies/ExponentialBackoffRetryPolicy.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (!base.ShouldRetry(currentRetryCount, lastException, out retryInterval))
+            {
+                return false;
+            }
+
+            // Avoid integer overlow and clamp max wait.
+            uint exponent = Math.Min(MaxExponent, currentRetryCount);
+
+            // 2 to the power of the retry count gives us exponential back-off.
+            double exponentialIntervalMs = Math.Pow(2.0, exponent);
```
drwill-ms, line 51: What I don't love about this is that it starts too low and takes 5 or so iterations to get over 1 second delay.

iothub/device/src/RetryPolicies/RetryHandler.cs
```diff
@@ -0,0 +1,152 @@
...
+
+namespace Microsoft.Azure.Devices.Client
+{
+    /// <summary>
+    /// Provides the base implementation of the retry mechanism for unreliable actions and transient conditions.
+    /// </summary>
+    /// <remarks>
+    /// Jitter can be under 1 second, plus or minus.
+    /// </remarks>
+    internal class RetryHandler
```
drwill-ms, line 17: Name suggestions? `OperationRetrier`?

iothub/device/src/RetryPolicies/RetryHandler.cs
```diff
@@ -0,0 +1,152 @@
...
+                }
+
+                // If we expect to wait until retry, calculate the remaining wait time, considering how much time the operation already
+                // took, and the minimum delay.
+                if (retryDelay > TimeSpan.Zero
+                    && !cancellationToken.IsCancellationRequested)
+                {
+                    // Don't pass in the cancellation token, because we'll handle that
+                    // condition specially in the catch blocks above.
+                    await Task.Delay(retryDelay, CancellationToken.None).ConfigureAwait(false);
```
drwill-ms, line 142: Vastly simplified this. Got rid of a min wait and a timer to see how long the operation took and subtracting that from the next delay.

iothub/device/src/RetryPolicies/RetryPolicy.cs
```diff
@@ -1,371 +0,0 @@
...
-                            Logging.Info(
-                                $"{cancellationToken.GetHashCode()} Last execution time was {stopwatch.Elapsed}. Adjusting back-off time to {retryDelay} to avoid high CPU/Memory spikes.");
-                    }
-
-                    if (!cancellationToken.IsCancellationRequested)
-                    {
-                        // Don't pass in the cancellation token, because we'll handle that
-                        // condition specially in the catch blocks above.
-                        await Task.Delay(retryDelay, CancellationToken.None).ConfigureAwait(false);
-                    }
```
drwill-ms, line 361: See here for old logic.

patilsnr (approved):
Optimistic sign off.

iothub/device/src/RetryPolicies/RetryPolicyBase.cs
```diff
@@ -0,0 +1,65 @@
...
+            // Don't calculate jitter if the value is very small
+            if (baseTimeMs < 10)
+            {
+                return TimeSpan.FromMilliseconds(baseTimeMs);
+            }
+
+            double jitterMs;
+
+            // Because Random is not threadsafe
+            lock (_rngLock)
```
timtay-microsoft, line 56: > Because Random is not threadsafe

In the sense that two calls to "rng.Next" might return the same value? If that's the only reason, I don't see the need for a lock here. I understand we'd want multiple devices to not retry with the same exact jitter, but within the scope of a single device, there is no reason I'm aware of that we need to avoid same rng values here

timtay-microsoft (approved):

iothub/device/src/RetryPolicies/ExponentialBackoffRetryPolicy.cs
drwill-ms: I updated it to start at 2^7 so it starts at 128 milliseconds, and gets to 1 second on the 4th retry.

iothub/device/src/RetryPolicies/RetryPolicyBase.cs
drwill-ms: If you had a gateway scenario and the same retry policy was being used with several devices, you'd risk them retrying with similar jitter values, negating the value of jitter.

I did a quick test to see how often I'd get dupes (0-99) running in a tight forloop with Task.Run:

> 33
45
6
20
86
73
16
5
83
1
3
5
33
33
38
98
93
61
64
13
97
57
57
18
67
26
57
59
5
27
74
15
62
96
56
46
70
76
95
82
75
82
87
65
83
35
40
1
82
25
82
86
72
64
39
38
92
66
49
26
44
44
98
49
18
51
83
13
32
28
17
55
88
38
97
56
98
34
19
66
37
57
52
96
88
12
36
40
46
46
8
34
70
28
63
83
96
11
35
35

I spot 4 dupes.

drwill-ms: So, probably not a huge deal, but this feels like something one should be more cautious about, given .NET is used for gateway.

drwill-ms: Adjacent dupes, I should have said.

I tried it with a lock, ran several times, only saw 1 pair of dupes in those runs. Could be statistical.

iothub/device/samples/how to guides/DeviceReconnectionSample/CustomRetryPolicy.cs
```diff
@@ -11,57 +11,35 @@ namespace Microsoft.Azure.Devices.Client.Samples
     /// An exponential backoff with jitter based retry policy that retries on additional exceptions
     /// not covered by the SDK, because in this application we wish to run as long as possible.
     /// </summary>
-    internal class CustomRetryPolicy : IRetryPolicy
+    internal class CustomRetryPolicy : ExponentialBackoffRetryPolicy
     {
-        private const int MaxRetryCount = int.MaxValue;
-        private const int MaxExponent = 22; // Avoid integer overlow (max of 30) and clamp max wait to just over 1 hour (2^22 = 1.16 hours).
```
abhipsaMisra, line 17: 👍🏽

iothub/device/samples/how to guides/DeviceReconnectionSample/CustomRetryPolicy.cs
```diff
@@ -11,57 +11,35 @@ namespace Microsoft.Azure.Devices.Client.Samples
...
             _logger.LogInformation($"Retry requested #{currentRetryCount} exception [{lastException.GetType()}: {lastException.Message}].");
 
-            if (currentRetryCount > MaxRetryCount)
+            shouldRetry = base.ShouldRetry(currentRetryCount, lastException, out retryInterval);
+            if (!shouldRetry
+                && _exceptionsToBeRetried.Contains(lastException.GetType()))
             {
-                _logger.LogError($"Retry requested #{currentRetryCount} of {MaxRetryCount} so giving up.");
-                return false;
+                retryInterval = TimeSpan.FromSeconds(15);
```
abhipsaMisra, line 36: is this so that we have a wait time when shouldRetry resolves to false however the exception is marked as "additional" retryable? I think a comment would help here.

abhipsaMisra (approved):

iothub/device/samples/how to guides/DeviceReconnectionSample/CustomRetryPolicy.cs
drwill-ms: yes, k

iothub/device/src/RetryPolicies/RetryPolicyBase.cs
drwill-ms: Improved the test. Used task completion source to synchronize 1000 tries. Stored results (0 to 100k random value) in a concurrent dictionary.

With lock, had 6 dupes, all in pairs.

Without lock we get 51, the max of which is 4 dupes.

I'd say that is a significant improvement, but one could certainly argue that even for gateway, having handfuls of devices retry with the same jitter the first time around is not impactful.

I could be convinced either way. Abhipsa says it isn't complicated, so leave it in. I'm torn, but I'll leave it in for now. It won't impact single device apps in a measurable way.

brycewang-microsoft (approved):

brycewang-microsoft (commented):
Is this PR the one you want to throw IotHubClientException with a particular ErrorCode for reaching retry limit?

drwill-ms:
Only failures are DPS custom allocation registrations (known external issue), so merging.

PR #2841, opened 2022-10-10
Fix hub service client query paging, and change device clean-up

drwill-ms:
Devices clean-up broken due to 2 bugs, one in query paging, the other in the clean-up app.

Also simplified clean-up to work on an exclusion list for deletion instead of an inclusion list for deletion.

PR #2843, opened 2022-10-10
Change how we clean-up devices and enrollments

drwill-ms:
#2841 but for main


timtay-microsoft (approved):

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
```diff
@@ -221,7 +221,7 @@ $verificationCertPath = "$PSScriptRoot/verification.cer";
 
 $iotHubX509DeviceCertCommonName = "iothubx509device1";
```
abhipsaMisra, line 222: This common name is used for the device Id for self-signed x509 enrollment. We'll want to append the "save" prefix here as well.

abhipsaMisra (approved):

e2e/test/prerequisites/E2ETestsSetup/e2eTestsSetup.ps1
drwill-ms: This is just the cert name. 2 lines down is the name we use for the device.

Does it matter?

PR #2845, opened 2022-10-10
Decrement event handlers in AMQP layers

tmahmood-microsoft:


drwill-ms (approved):

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -73,6 +73,7 @@ public void Shutdown()
 
             _amqpAuthenticationRefresher?.StopLoop();
             _amqpIotConnection?.SafeClose();
+            _amqpIotConnection.Closed -= OnConnectionClosed;
```
abhipsaMisra, line 76: you'll want to decrement the event handler before you make an explicit close call. Otherwise, the close call will end up invoking the handler (which is not required in this scenario)

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -372,6 +373,10 @@ internal async Task DisableMessageReceivingLinkAsync(CancellationToken cancellat
             if (_eventReceiverLinkDisconnected != null)
             {
                 _messageReceivingLink.Closed -= _eventReceiverLinkDisconnected;
```
abhipsaMisra, line 375: does this messaging link require a null check as well?

abhipsaMisra (approved):

iothub/device/src/Transport/Amqp/AmqpUnit.cs
tmahmood-microsoft: we already have null checks for _messageReceivingLink in code above.

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
tmahmood-microsoft: updated

patilsnr (approved):

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -72,7 +72,11 @@ public void Shutdown()
                 Logging.Enter(this, _amqpIotConnection, nameof(Shutdown));
 
             _amqpAuthenticationRefresher?.StopLoop();
-            _amqpIotConnection?.SafeClose();
+            if (_amqpIotConnection != null)
```
tmahmood-microsoft, line 75: doing a if condition because the following is not allowed:
_amqpIotConnection?.Closed -= OnConnectionClosed;

PR #2847, opened 2022-10-11
Fix leaks in DPS enrollment tests

drwill-ms:
Fix enrollment leaks in DPS tests.

For the most part, the action wasn't happening inside the `try`, or the cleanup wasn't happening inside the `finally`.

### DIFF
--- a/e2e/test/provisioning/ProvisioningServiceClientE2ETests.cs
+++ b/e2e/test/provisioning/ProvisioningServiceClientE2ETests.cs
@@ -213,7 +213,10 @@ namespace Microsoft.Azure.Devices.E2ETests.Provisioning
             }
             finally
             {
-                await DeleteCreatedEnrollmentAsync(EnrollmentType.Individual, individualEnrollment.RegistrationId, null, Logger);
+                if (individualEnrollment != null)
+                {
+                    await DeleteCreatedEnrollmentAsync(EnrollmentType.Individual, individualEnrollment.RegistrationId, null, Logger);
+                }
             }
         }
 
@@ -268,7 +271,10 @@ namespace Microsoft.Azure.Devices.E2ETests.Provisioning
             }
             finally
             {
-                await DeleteCreatedEnrollmentAsync(EnrollmentType.Group, null, enrollmentGroup.EnrollmentGroupId, Logger);
+                if (enrollmentGroup != null)
+                {
+                    await DeleteCreatedEnrollmentAsync(EnrollmentType.Group, null, enrollmentGroup.EnrollmentGroupId, Logger);
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


e2e/test/provisioning/ProvisioningServiceClientE2ETests.cs
```diff
@@ -159,21 +159,23 @@ public async Task ProvisioningServiceClient_GetIndividualEnrollmentAttestation(A
...
-                    null,
-                    null,
-                    null,
-                    Logger)
-                .ConfigureAwait(false);
+            IndividualEnrollment individualEnrollment = null;
 
             try
             {
+                individualEnrollment = await CreateIndividualEnrollmentAsync(
```
timtay-microsoft, line 166: Why is this moved into the try block? If this call fails, then there is nothing to cleanup, right?

e2e/test/provisioning/ReprovisioningE2ETests.cs
```diff
@@ -936,9 +949,9 @@ public static void CleanupCertificates()
             {
                 s_x509CertificatesFolder.Delete(true);
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                // In case of an exception, silently exit. All systems images on Microsoft hosted agents will be cleaned up by the system.
+                Console.Write($"Client certificate cleanup failed due to {ex}");
```
timtay-microsoft, line 954: Logger?

e2e/test/provisioning/ProvisioningServiceClientE2ETests.cs
drwill-ms: In some cases I found, the code that creates it has checks and may fail. So, we just want to be sure we clean it up.

e2e/test/provisioning/ReprovisioningE2ETests.cs
drwill-ms: ClassCleanup is static and no access to the logger. :(

e2e/test/provisioning/ProvisioningServiceClientE2ETests.cs
drwill-ms: I checked this one, and it doesn't, but I think it is safe to keep it in here. We check for null, just in case.

drwill-ms: Adding the null check, because it wasn't actually done in this case.

e2e/test/provisioning/ReprovisioningE2ETests.cs
timtay-microsoft: That's a shame, but fair enough.

timtay-microsoft (approved):

drwill-ms:
Only custom allocation policy DPS tests failed (expected), so merging.

PR #2848, opened 2022-10-11
Reprovisioning DPS leaks

patilsnr:


e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -736,6 +736,8 @@ public async Task DPS_Registration_AmqpWs_X509_IndividualEnrollment_InvalidGloba
                     : _idPrefix + AttestationTypeToString(attestationType) + "-" + Guid.NewGuid();
             }
 
+            bool shouldCleanupEnrollment = groupId == null || groupId == TestConfiguration.Provisioning.X509GroupEnrollmentName;
```
drwill-ms, line 739: It shouldn't equal the hardcoded name if we should cleanup, right?

```suggestion
            bool shouldCleanupEnrollment = groupId == null || groupId != TestConfiguration.Provisioning.X509GroupEnrollmentName;
```

e2e/test/provisioning/ReprovisioningE2ETests.cs
```diff
@@ -95,7 +95,7 @@ public async Task ProvisioningDeviceClient_ReprovisionedDeviceResetsTwin_AmqpWs_
         [Timeout(TestTimeoutMilliseconds)]
         public async Task ProvisioningDeviceClient_ReprovisionedDeviceResetsTwin_AmqpTcp_SymmetricKey_RegisterOk_Individual()
         {
-            await ProvisioningDeviceClient_ReprovisioningFlow_KeepTwin(
+            await ProvisioningDeviceClient_ReprovisioningFlow_ResetTwin(
```
drwill-ms, line 98: Can you speak to the test name changes?

e2e/test/provisioning/ProvisioningE2ETests.cs
drwill-ms: This is actually a problem in my changes. I don't know why your PR has my changes in it too.

drwill-ms: Either way, you can fix this for me, please.

e2e/test/provisioning/ReprovisioningE2ETests.cs
```diff
@@ -451,69 +451,84 @@ public async Task ProvisioningDeviceClient_ReprovisioningBlockingWorks_MqttWs_Sy
...
-            }
+                auth = await CreateAuthenticationProviderFromNameAsync(
+                   attestationType,
+                   enrollmentType,
+                   groupId,
+                   reprovisionPolicy,
+                   allocationPolicy,
+                   customAllocationDefinition,
+                   iotHubsToStartAt)
+               .ConfigureAwait(false);
```
drwill-ms, line 469:
```suggestion
                       attestationType,
                       enrollmentType,
                       groupId,
                       reprovisionPolicy,
                       allocationPolicy,
                       customAllocationDefinition,
                       iotHubsToStartAt)
                   .ConfigureAwait(false);
```

e2e/test/provisioning/ReprovisioningE2ETests.cs
```diff
@@ -711,16 +726,14 @@ public async Task ProvisioningDeviceClient_ReprovisioningBlockingWorks_MqttWs_Sy
 
             if (provisioningAuth is AuthenticationProviderX509 x509Auth)
             {
-                X509Certificate2 cert = x509Auth.GetAuthenticationCertificate();
+                using X509Certificate2 cert = x509Auth.GetAuthenticationCertificate();
```
drwill-ms, line 729: If you put `using` here, it'll be disposed after line 731 and will not work when being used when referenced by `auth` in subsequent lines.

It has to be disposed in the `finally` block.

e2e/test/provisioning/ProvisioningE2ETests.cs
patilsnr: On it

e2e/test/provisioning/ReprovisioningE2ETests.cs
patilsnr: These addressed what we spoke about IRL regarding how different tests have the same functionality. After looking at more, it appeared to be that `..._ReprovisioningFlow_KeepTwin()` was being used in tests that should have been calling `ResetTwin()` instead, and vice versa.

drwill-ms (approved):

PR #2850, opened 2022-10-11
Print any errors from the devices cleanup job

drwill-ms:
Same as what was put in previews/v2 today.

### DIFF
--- a/iothub/service/samples/how to guides/CleanupDevicesSample/ImportError.cs
+++ b/iothub/service/samples/how to guides/CleanupDevicesSample/ImportError.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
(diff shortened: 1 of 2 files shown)


iothub/service/samples/how to guides/CleanupDevicesSample/ImportError.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices.Samples
+{
+    internal class ImportError
```
drwill-ms, line 8: I'm thinking about adding this to the SDK for v2, like we have for ImportExportDevice. Thoughts?

timtay-microsoft: I'm surprised to hear we don't already have this feature. I'm in favor of adding it, sure.

timtay-microsoft (approved):

PR #2853, opened 2022-10-12
Removed public property to get shared access key as byte[] 

tmahmood-microsoft:


iothub/device/src/Authentication/ClientAuthenticationWithRegistrySymmetricKey.cs
```diff
@@ -87,25 +87,12 @@ public string ModuleId
...
-        }
-
-        /// <summary>
-        /// Gets or sets the Base64 formatted key associated with the device.
-        /// </summary>
-        public string KeyAsBase64String
-        {
-            get => Convert.ToBase64String(Key);
-            set => SetKeyFromBase64String(value);
+            get => Convert.ToBase64String(_key);
```
timtay-microsoft, line 94: Do we even need to save the key privately as a byte[]? I imagine we could just take a string as is and not need to convert to/from base64 here

iothub/device/src/Authentication/ClientAuthenticationWithRegistrySymmetricKey.cs
```diff
@@ -87,25 +87,12 @@ public string ModuleId
         }
 
         /// <summary>
-        /// Gets or sets the key associated with the device.
+        /// Gets or sets the Base64 formatted shared access key associated with the device.
```
drwill-ms, line 90:
```suggestion
        /// Gets or sets the BBase64-formatted shared access key associated with the device.
```

iothub/device/src/Authentication/ClientAuthenticationWithRegistrySymmetricKey.cs
```diff
@@ -118,14 +105,14 @@ public void Populate(ref IotHubConnectionCredentials iotHubConnectionCredentials
...
-            iotHubConnectionCredentials.SharedAccessKey = KeyAsBase64String;
+            iotHubConnectionCredentials.SharedAccessKey = Key;
             iotHubConnectionCredentials.SharedAccessKeyName = null;
             iotHubConnectionCredentials.SharedAccessSignature = null;
             iotHubConnectionCredentials.SasTokenTimeToLive = _suggestedTimeToLive;
             iotHubConnectionCredentials.SasTokenRenewalBuffer = _timeBufferPercentage;
         }
 
-        private void SetKeyFromBase64String(string key)
+        private void SetKey(string key)
```
drwill-ms, line 115: Do we need this validation? Can we just use an auto-property.

drwill-ms (approved):

timtay-microsoft (approved):

PR #2858, opened 2022-10-12
return hub exception only from ToIotHubClientContract methods

brycewang-microsoft:
In the AMQP layer, this PR is for both the hub device and service.

### DIFF
--- a/iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs
+++ b/iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs
@@ -52,142 +52,113 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIot
             return retException;
         }
 
-        public static Exception ToIotHubClientContract(AmqpException amqpException)
+        public static IotHubClientException ToIotHubClientContract(AmqpException amqpException)
         {
             Error error = amqpException.Error;
             AmqpSymbol amqpSymbol = error.Condition;
             string message = error.ToString();
 
+            IotHubClientException retException;
+
             // Generic AMQP error
             if (Equals(AmqpErrorCode.InternalError, amqpSymbol))
             {
-                return new IotHubClientException(message, IotHubClientErrorCode.NetworkErrors, amqpException);
+                retException = new(message, IotHubClientErrorCode.NetworkErrors, amqpException);
             }
             else if (Equals(AmqpErrorCode.NotFound, amqpSymbol))
             {
-                return new IotHubClientException(message, IotHubClientErrorCode.DeviceNotFound, amqpException);
+                retException = new(message, IotHubClientErrorCode.DeviceNotFound, amqpException);
             }
             else if (Equals(AmqpErrorCode.UnauthorizedAccess, amqpSymbol))
             {
-                return new IotHubClientException(message, IotHubClientErrorCode.Unauthorized, amqpException);
-            }
-            else if (Equals(AmqpErrorCode.DecodeError, amqpSymbol))
-            {
-                return new IotHubClientException(message, amqpException);
+                retException = new(message, IotHubClientErrorCode.Unauthorized, amqpException);
             }
             else if (Equals(AmqpErrorCode.ResourceLimitExceeded, amqpSymbol))
             {
-                return new IotHubClientException(message, IotHubClientErrorCode.DeviceMaximumQueueDepthExceeded, amqpException);
-            }
-            else if (Equals(AmqpErrorCode.NotAllowed, amqpSymbol))
-            {
-                return new InvalidOperationException(message, amqpException);
-            }
-            else if (Equals(AmqpErrorCode.InvalidField, amqpSymbol))
-            {
-                return new InvalidOperationException(message, amqpException);
-            }
-            else if (Equals(AmqpErrorCode.NotImplemented, amqpSymbol))
-            {
-                return new NotSupportedException(message, amqpException);
+                retException = new(message, IotHubClientErrorCode.DeviceMaximumQueueDepthExceeded, amqpException);
             }
             else if (Equals(AmqpErrorCode.ResourceLocked, amqpSymbol))
             {
-                return new IotHubClientException(message, true, amqpException);
+                retException = new(message, true, amqpException);
             }
             else if (Equals(AmqpErrorCode.PreconditionFailed, amqpSymbol))
             {
-                return new IotHubClientException(message, IotHubClientErrorCode.PreconditionFailed, amqpException);
-            }
-            else if (Equals(AmqpErrorCode.ResourceDeleted, amqpSymbol))
-            {
-                return new IotHubClientException(message, amqpException);
-            }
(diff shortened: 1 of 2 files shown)


iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs
```diff
@@ -52,142 +52,113 @@ public static Exception GetExceptionFromOutcome(Outcome outcome)
...
-            {
-                return new InvalidOperationException(message, amqpException);
-            }
-            else if (Equals(AmqpErrorCode.InvalidField, amqpSymbol))
-            {
-                return new InvalidOperationException(message, amqpException);
-            }
-            else if (Equals(AmqpErrorCode.NotImplemented, amqpSymbol))
-            {
-                return new NotSupportedException(message, amqpException);
```
brycewang-microsoft, line 92: these will be addressed in the "else" block below and returned as `IotHubClientException`.

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -114,69 +109,52 @@ internal static Exception ToIotHubClientContract(Error error)
...
             {
-                retException = new IotHubServiceException(message, HttpStatusCode.NotFound, IotHubServiceErrorCode.DeviceNotFound);
-            }
-            else if (error.Condition.Equals(AmqpErrorCode.NotImplemented))
-            {
-                retException = new NotSupportedException(message);
-            }
-            else if (error.Condition.Equals(AmqpErrorCode.NotAllowed))
-            {
-                retException = new InvalidOperationException(message);
```
brycewang-microsoft, line 129: similarly here, these will be addressed in the "else" block below and returned as `IotHubServiceException`.

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -114,69 +109,52 @@ internal static Exception ToIotHubClientContract(Error error)
...
             {
-                retException = new IotHubServiceException(message, HttpStatusCode.Unauthorized, IotHubServiceErrorCode.IotHubUnauthorizedAccess);
-            }
-            else if (error.Condition.Equals(IotHubAmqpErrorCode.ArgumentError))
-            {
-                retException = new ArgumentException(message);
-            }
-            else if (error.Condition.Equals(IotHubAmqpErrorCode.ArgumentOutOfRangeError))
-            {
-                retException = new ArgumentOutOfRangeException(message);
```
brycewang-microsoft, line 141: same thing here

iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs
```diff
@@ -52,142 +52,113 @@ public static Exception GetExceptionFromOutcome(Outcome outcome)
...
             {
-                return new IotHubClientException(message, true, amqpException);
+                retException = new(message, true, amqpException);
             }
             else if (Equals(AmqpErrorCode.Stolen, amqpSymbol))
             {
-                return new IotHubClientException(message, true, amqpException);
+                retException = new(message, true, amqpException);
             }
             // AMQP Transaction Error
```
drwill-ms, line 139: Does this comment need to move or be removed?

iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs
```diff
@@ -252,21 +207,20 @@ public static Exception ToIotHubClientContract(Error error)
             }
             else if (error.Condition.Equals(DeviceContainerThrottled))
             {
-                retException = new IotHubClientException(message, IotHubClientErrorCode.Throttled);
+                retException = new(message, IotHubClientErrorCode.Throttled);
             }
             else if (error.Condition.Equals(IotHubSuspended))
             {
-                retException = new IotHubClientException("Iothub {0} is suspended".FormatInvariant(message), IotHubClientErrorCode.Suspended);
+                retException = new("Iothub {0} is suspended".FormatInvariant(message), IotHubClientErrorCode.Suspended);
```
drwill-ms, line 214:
```suggestion
                retException = new("IoT hub {0} is suspended".FormatInvariant(message), IotHubClientErrorCode.Suspended);
```

iothub/service/src/Amqp/AmqpClientHelper.cs
```diff
@@ -36,18 +36,13 @@ internal static Exception ToIotHubClientContract(Exception exception)
...
-                        // pass amqpException as the inner exception of IotHubServiceException
-                        return new IotHubServiceException(
-                            hubEx.Message,
-                            hubEx.StatusCode,
-                            hubEx.ErrorCode,
-                            string.Empty,
-                            amqpException);
-                    }
-                    return ex;
+                    IotHubServiceException ex = ToIotHubClientContract(amqpException.Error);
```
drwill-ms, line 39: Bummer that we have to create a whole new exception when we have one. Perhaps we pass in the inner to ToIotHubClientContract?

drwill-ms (approved):

drwill-ms:
Please get sign-off from Tim before merge.

iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs
brycewang-microsoft: It seems that this comment is for several transaction-error-related AmqpErrorCode in the following else-ifs, so no, we can just keep it as-is.

iothub/service/src/Amqp/AmqpClientHelper.cs
brycewang-microsoft: Sure, I like this idea!

iothub/device/src/Transport/AmqpIot/AmqpIotErrorAdapter.cs
```diff
@@ -52,142 +52,113 @@ public static Exception GetExceptionFromOutcome(Outcome outcome)
...
             AmqpSymbol amqpSymbol = error.Condition;
             string message = error.ToString();
 
+            IotHubClientException retException;
+
             // Generic AMQP error
             if (Equals(AmqpErrorCode.InternalError, amqpSymbol))
             {
-                return new IotHubClientException(message, IotHubClientErrorCode.NetworkErrors, amqpException);
+                retException = new(message, IotHubClientErrorCode.NetworkErrors, amqpException);
```
timtay-microsoft, line 66: @drwill-ms how do you feel about using ```new()``` in contexts like these where the variable's type isn't immediately obvious?

Should it be
```csharp
retException = new IotHubClientException(...);
```
instead?

drwill-ms: Yeah, we should basically just use it for class fields where the type must be on the left or for inline init of properties. I agree, in this case it isn't as obvious so we should be explicit.

timtay-microsoft (approved):
Just the one comment. Otherwise looks good

PR #2860, opened 2022-10-13
Fix enrollments requiring attestations

timtay-microsoft:
Attestations are allowed to be null in an enrollment


timtay-microsoft:
> what's the use-case for this?

https://github.com/Azure/azure-iot-sdk-csharp/blob/cc1575e2ee47a4fc2981089a25827c8606dec983/provisioning/service/src/IndividualEnrollmentsClient.cs#L116

brycewang-microsoft (approved):

abhipsaMisra:
How, if at all, does this affect the create enrollment flow? I'm concerned that we'll allow users to create enrollments without an associated attestation mechanism, and that seems wrong to me.

PR #2864, opened 2022-10-14
Use the Azure.Core ETag type in the DPS service client

timtay-microsoft:
This is the first class we've added from the Azure.Core package to this library, so we now have a dependency on that package


iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -99,19 +99,7 @@ private static async Task<T> ReadResponseMessageAsync<T>(HttpResponseMessage mes
                 return (T)(object)message;
             }
 
-            T entity = await ReadAsAsync<T>(message.Content, token).ConfigureAwait(false);
-
-            // ETag in the header is considered authoritative
-            if (entity is IETagHolder eTagHolder)
```
timtay-microsoft, line 105: No class still implemented this interface in our IoT hub device client, so this if block was never entered.

timtay-microsoft: It's worth remembering that HTTP never supported twin operations, and twin operations are the only ones that use ETags in the IoT hub device client

provisioning/service/src/Models/EnrollmentGroup.cs
```diff
@@ -122,86 +123,10 @@ public EnrollmentGroup(string enrollmentGroupId, Attestation attestation)
...
-        /// <param name="initialTwinState">The <see cref="TwinState"/> with the initial Twin condition. This is optional and can be null.</param>
-        /// <param name="provisioningStatus">The <see cref="ProvisioningStatus"/> that determine the initial status of the device. This is optional and can be null.</param>
-        /// <param name="createdDateTimeUtc">The DateTime with the date and time that the enrollment was created. This is optional and can be null.</param>
-        /// <param name="lastUpdatedDateTimeUtc">The DateTime with the date and time that the enrollment was updated. This is optional and can be null.</param>
-        /// <param name="eTag">The string with the eTag that identify the correct instance of the enrollment in the service. It cannot be null or empty.</param>
-        /// <param name="capabilities">The capabilities of the device (ie: is it an edge device?)</param>
-        /// <exception cref="DeviceProvisioningServiceException">If the received JSON is invalid.</exception>
-        [JsonConstructor]
-        internal EnrollmentGroup(
```
timtay-microsoft, line 173: We only need an empty ctor to make serialization work. I don't know why we had this uber ctor, but it isn't referenced anywhere so it seems safe to replace with an empty ctor

provisioning/service/src/DeviceRegistrationsClient.cs
```diff
@@ -8,6 +8,7 @@
 using System.Threading.Tasks;
 using System.Threading;
 using Newtonsoft.Json;
+using Azure;
```
drwill-ms, line 11: sort

provisioning/service/src/EnrollmentGroupsClient.cs
```diff
@@ -9,6 +9,7 @@
 using System.Threading.Tasks;
 using System.Threading;
 using Newtonsoft.Json;
+using Azure;
```
drwill-ms, line 12: sort

provisioning/service/src/Http/ContractApiHttp.cs
```diff
@@ -127,7 +128,12 @@ internal class ContractApiHttp : IContractApiHttp
                         msg.Headers.Add(header.Key, header.Value);
                     }
                 }
-                InsertIfMatch(msg, ifMatch);
+                
+                if (!string.IsNullOrWhiteSpace(eTag.ToString()))
```
drwill-ms, line 132: Can ETag be null?

provisioning/service/src/IndividualEnrollmentsClient.cs
```diff
@@ -9,6 +9,7 @@
 using System.Threading.Tasks;
 using System.Threading;
 using Newtonsoft.Json;
+using Azure;
```
drwill-ms, line 12: sort

drwill-ms (approved):

provisioning/service/src/Http/ContractApiHttp.cs
timtay-microsoft: No

PR #2868, opened 2022-10-14
Rename ProvisioningRegistrationAdditionalData to RegistrationRequestPayload

timtay-microsoft:


provisioning/device/src/Models/ProvisioningRegistrationAdditionalData.cs
```diff
@@ -4,10 +4,9 @@
 namespace Microsoft.Azure.Devices.Provisioning.Client
 {
     /// <summary>
-    /// Additional parameters to be passed over registartion instance
+    /// Optional data to be included in the registration request.
     /// </summary>
-    /// <returns>The registration result.</returns>
-    public class ProvisioningRegistrationAdditionalData
+    public class RegistrationRequestPayload
```
drwill-ms, line 9: Need to call it out in the migration doc.

Also, file name needs renaming.

PR #2869, opened 2022-10-14
Update api diff script

drwill-ms:


tmahmood-microsoft (approved):

tools/diffscripts/diffapi.ps1
```diff
@@ -124,32 +124,32 @@ if ($null -eq $asmToolExecutableCommand)
...
 {
     $internalRootPath = Join-Path -Path $baseRootPath -Child "\iot-sdks-internals"
 }
 
 Write-Verbose "Using $internalRootPath for the internals sdk repository base directory."
 
 # If we specify to use the preview directory on the command line we will set it as such
-$compareDirectory = Join-Path -Path $internalRootPath -Child "\sdk_design_docs\CSharp\main"
-if ($IsPreview) 
+$compareDirectory = Join-Path -Path $internalRootPath -Child "\sdk_design_docs\CSharp\previews-v2"
```
abhipsaMisra, line 140: shouldn't this be main and the one in L143 be previews-v2?

drwill-ms: This will become main. It isn't a preview of our GA SDK v1.

I think we'll need to rename the current main directory to v1, and main will become v2.

Or something like that.

Thoughts?

abhipsaMisra: I see, that makes sense. We should probably also cut another preview at some point soon, else this previews/v2 -> preview diff will end up showing us a lot of outdated diffs (the preview branch hasn't been updated in a long long time). The only features in "preview" are dps client certificates and device streaming, both of which have their individual preview feature branches.

drwill-ms: We have nothing to preview besides v2, right?

abhipsaMisra: Nothing else right now, no.

PR #2871, opened 2022-10-14
Fix namespaces

abhipsaMisra:
These public types don't require their own namespaces

### DIFF
--- a/iothub/device/src/FileUpload/FileUploadCompletionNotification.cs
+++ b/iothub/device/src/FileUpload/FileUploadCompletionNotification.cs
@@ -3,7 +3,7 @@
 
 using Newtonsoft.Json;
 
-namespace Microsoft.Azure.Devices.Client.Transport
+namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// The request payload to send to IoT hub to notify it when a file upload is completed, whether successful or not.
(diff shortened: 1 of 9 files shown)


iothub/device/src/Transport/Mqtt/IWillMessage.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
+namespace Microsoft.Azure.Devices.Client
```
abhipsaMisra, line 4: The amqp public types are under the "root" namespace, so moving this there as well

iothub/device/src/FileUpload/FileUploadCompletionNotification.cs
```diff
@@ -3,7 +3,7 @@
 
 using Newtonsoft.Json;
 
-namespace Microsoft.Azure.Devices.Client.Transport
+namespace Microsoft.Azure.Devices.Client
```
abhipsaMisra, line 6: This, and below - they aren't transport specific types really.

drwill-ms (approved):

tmahmood-microsoft (approved):

andyk-ms (approved):

patilsnr (approved):

PR #2876, opened 2022-10-17
Collapse x509 auth provider classes

drwill-ms:
We had an abstract base class with only one inheritor. It didn't seem to serve any purpose, so we removed it and renamed the inheritor to use that name, which seemed more concise.


authentication/src/AuthenticationProviderX509.cs
```diff
@@ -1,15 +1,36 @@
...
 using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.Azure.Devices.Authentication
 {
     /// <summary>
-    /// The device authentication provider interface for X.509-based hardware security modules.
+    /// The device authentication for using an X509 certificate object.
     /// </summary>
-    public abstract class AuthenticationProviderX509 : AuthenticationProvider
+    public class AuthenticationProviderX509 : AuthenticationProvider
```
abhipsaMisra, line 11: extremely small nit: why not "AuthenticationProviderX509Certificate", since this class deals with x509 certificates.

abhipsaMisra (approved):

authentication/src/AuthenticationProviderX509.cs
timtay-microsoft: To me, x509 already implies "x509 certificates", so that would be a bit redundant

timtay-microsoft (approved):

patilsnr (approved):

PR #2880, opened 2022-10-17
Add link to v2 migration document in readme for main branch

brycewang-microsoft:


timtay-microsoft (approved):

readme.md
```diff
@@ -1,5 +1,13 @@
 # Microsoft Azure IoT SDK for .NET
 
+## 2.0.0 clients release notice
+
+A [2.0.0 preview release](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2022-10-14-v2preview) has been published for each package in this library. 
+With this release, all future features will be brought only to this new major version, so users are highly encouraged to migrate from the 1.X.X releases to 
+the new 2.0.0 releases.
```
abhipsaMisra, line 7: We might want to word this a bit differently, given that the v2 is currently in preview. 
How about we say something like - A 2.0.0 preview prelease has been published for each package in this library. These packages simplify usage and handling of our clients, and we highly encourage you to try out these packages and share your feedback. Once we promote these changes to `main` in the future, all future features will be brought only to this new major version.

Thoughts?

readme.md
```diff
@@ -1,5 +1,13 @@
 # Microsoft Azure IoT SDK for .NET
 
+## 2.0.0 clients release notice
```
drwill-ms, line 3: I don't think it is necessary to specify the minor and patch version. Just v2 should be sufficient.

readme.md
```diff
@@ -1,5 +1,13 @@
...
+
+A [2.0.0 preview release](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2022-10-14-v2preview) has been published for each package in this library.
```
drwill-ms, line 5: If you do want to specify this exact release, then it should be 2.0.0-preview001. And then I'd say keep an eye out for updates.

readme.md
```diff
@@ -1,5 +1,13 @@
 # Microsoft Azure IoT SDK for .NET
 
+## 2.0.0 clients release notice
+
+A [2.0.0 preview release](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2022-10-14-v2preview) has been published for each package in this library. 
+With this release, all future features will be brought only to this new major version, so users are highly encouraged to migrate from the 1.X.X releases to
```
drwill-ms, line 6: Don't encourage users to migrate while it is in preview.

Instead, we want them to try it out and give us feedback.

drwill-ms (changes requested):
Comments inline

readme.md
brycewang-microsoft: Yeah, I like it!

drwill-ms (approved):

PR #2885, opened 2022-10-17
API accessors and naming updates

drwill-ms:


iothub/service/src/Registry/Models/BulkRegistryOperationResult.cs
```diff
@@ -1,29 +1,26 @@
...
-        /// If the operation was not successful, this contains an array of DeviceRegistryOperationError objects.
+        /// If the operation was not successful, this contains DeviceRegistryOperationError objects.
         /// </summary>
         [JsonProperty(PropertyName = "errors", DefaultValueHandling = DefaultValueHandling.Ignore)]
-        [System.Diagnostics.CodeAnalysis.SuppressMessage(
-            "Performance",
-            "CA1819:Properties should not return arrays",
-            Justification = "Cannot change property types on public classes.")]
-        public DeviceRegistryOperationError[] Errors { get; internal set; }
+        public IList<DeviceRegistryOperationError> Errors { get; internal set; } = new List<DeviceRegistryOperationError>();
```
patilsnr, line 24: Out of curiosity, why do we use `IList` instead of `List`? On another note, if we used `List`, the expression could be simplified with just a `new()` statement, but I assume you had a reason for the alternative instead.

provisioning/service/src/Models/QueryResult.cs
```diff
@@ -41,17 +41,17 @@ public class QueryResult
...
+        public QueryResultType Type { get; protected private set; }
 
         /// <summary>
         /// Getter for the list of query result Items.
         /// </summary>
-        public IEnumerable<object> Items { get; private set; }
+        public IEnumerable<object> Items { get; protected private set; }
 
         /// <summary>
         /// Getter for the query result continuationToken.
```
patilsnr, line 52: nit: => `...result continuation token.`

patilsnr (approved):
Optimistic sign off :)

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -140,6 +140,7 @@ internal IotHubClientException(IotHubClientErrorCode errorCode) : base()
         /// </summary>
         /// <param name="info">The <see cref="SerializationInfo"/> that holds the serialized object data about the exception being thrown.</param>
         /// <param name="context">The <see cref="StreamingContext"/> that contains contextual information about the source or destination.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public override void GetObjectData(SerializationInfo info, StreamingContext context)
```
timtay-microsoft, line 144: Do we even need this ctor? Last I checked, it was unreferenced

timtay-microsoft (approved):

SDK v2 migration guide.md
```diff
@@ -38,10 +38,13 @@ but users are still encouraged to migrate to version 2 when they have the chance
 | Version 1 API | Equivalent version 2 API |
 |:---|:---|
 | `DeviceClient` | `IotHubDeviceClient` |
-| `SetConnectionStatusChangesHandler` | `SetConnectionStatusChangeHandler` |
+| `DeviceClient.SendEventAsync(...)` | `IotHubDeviceClient.SendTelemetryAsync(...)` |
```
abhipsaMisra, line 41: On this note, should we also update (Incoming/Outgoing)Message to (Incoming/Outgoing)Telemetry?

iothub/service/src/Registry/Models/BulkRegistryOperationResult.cs
drwill-ms: Because we prefer to abstract API interfaces away to interfaces instead of concrete classes when possible. We could change the underlying type, as long as it inherits from IList, and not break the customer.

iothub/device/src/Exceptions/IotHubClientException.cs
drwill-ms: Methods and ctors for de/serialization won't be referenced, but they are called at runtime.

I don't know the point of this `GetObjectData` method, though.

SDK v2 migration guide.md
drwill-ms: Incoming wouldn't be telemetry, but OutgoingMessage could be renamed to Telemetry. Like that?

abhipsaMisra: That makes sense. We do call it "SetIncomingMessageCallbackAsync".

abhipsaMisra (approved):

SDK v2 migration guide.md
drwill-ms: On second thought, I think TelemetryMessage instead of just Telemetry. Good?

PR #2887, opened 2022-10-17
Consolidate exceptions thrown out by different clients

brycewang-microsoft:
Consolidate the following types of exception thrown out by different clients, and update doc comments for public APIs and relevant tests.
+ `IOException`
+ `SocketException`
+ `WebSocketException`
+ `UnauthorizedAccessException`
+ `SwaggerException`
+ `HttpRequestException`
+ `IndexOutOfRangeException`
+ `CryptographicException`


iothub/device/src/Authentication/AuthenticationWithTokenRefresh.cs
```diff
@@ -32,20 +32,20 @@ public abstract class AuthenticationWithTokenRefresh : IAuthenticationMethod
         /// The time buffer before expiry when the token should be renewed, expressed as a percentage of the time to live.
         /// The default behavior is that the token will be renewed when it has <see cref="DefaultSasRenewalBufferPercentage"/> percent or less of its lifespan left.
         ///</param>
-        /// <exception cref="ArgumentOutOfRangeException">Thrown if <paramref name="suggestedTimeToLive"/> is a negative timespan, or if
```
drwill-ms, line 35: I could see an argument for keeping this type of exception. It is more specific.

drwill-ms: It inherits from ArgumentException, btw.

iothub/service/src/Twin/Models/TwinCollection.cs
```diff
@@ -160,7 +160,7 @@ public TwinMetadata GetMetadata()
         /// Gets the last updated time for this property.
         /// </summary>
         /// <returns>Date-time instance representing the last updated time for this property.</returns>
-        /// <exception cref="System.NullReferenceException">Thrown when the metadata object is null.
+        /// <exception cref="NullReferenceException">Thrown when the metadata object is null.
```
drwill-ms, line 163: This was a bug we couldn't fix, but we should fix now. Not in this PR, but we should fix it so we don't return an NRE.

drwill-ms (approved):
I agree with all the changes except removing ArgumentOutOfRangeException, but I wouldn't die on that hill.

iothub/device/src/Authentication/AuthenticationWithTokenRefresh.cs
brycewang-microsoft: Yeah, throwing more specifically is better than maintaining a shorter list of exceptions we have in such a case. Sure, I will change it back.

tmahmood-microsoft (approved):

iothub/service/src/Twin/Models/TwinCollection.cs
drwill-ms: An SDK should never throw a NullReferenceException. That is a bug.

drwill-ms (approved):

iothub/device/src/Authentication/HsmAuthentication/Transport/HttpRequestResponseSerializer.cs
```diff
@@ -120,9 +120,17 @@ internal static async Task<HttpResponseMessage> DeserializeResponseAsync(HttpBuf
...
+                            throw new IotHubClientException($"Header value is invalid for {headerName}.", IotHubClientErrorCode.NetworkErrors);
+                        }
+
+                        try
+                        {
+                            await httpResponse.Content.LoadIntoBufferAsync(contentLength).ConfigureAwait(false);
+                        }
+                        catch (HttpRequestException ex)
+                        {
+                            throw new IotHubClientException(ex.Message, IotHubClientErrorCode.NetworkErrors, ex);
```
drwill-ms, line 132: is transient?

drwill-ms: Implied from network errors?

iothub/service/src/Utilities/Argument.cs
```diff
@@ -120,13 +120,13 @@ internal static void ValidateBufferBounds(byte[] buffer, int offset, int size)
 
             if (offset < 0 || offset > buffer.Length || size <= 0)
             {
-                throw new ArgumentOutOfRangeException(nameof(offset));
+                throw new ArgumentOutOfRangeException(nameof(offset), "The buffer bound is invalid.");
```
drwill-ms, line 123: buffer bounds are invalid

iothub/device/src/Authentication/HsmAuthentication/Transport/HttpRequestResponseSerializer.cs
brycewang-microsoft: Yes, `IotHubClientErrorCode.NetworkErrors` will be taken as transient within ctor of `IotHubClientException`

drwill-ms (approved):

PR #2892, opened 2022-10-19
Remove ToJson() methods

drwill-ms:


iothub/service/src/Twin/Models/Twin.cs
```diff
@@ -175,16 +175,6 @@ public Twin(TwinProperties twinProperties)
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
         public virtual IReadOnlyList<string> ParentScopes { get; internal set; } = new List<string>();
 
-        /// <summary>
-        /// Gets the Twin as a JSON string.
-        /// </summary>
-        /// <param name="formatting">Optional. Formatting for the output JSON string.</param>
-        /// <returns>JSON string.</returns>
-        public string ToJson(Formatting formatting = Formatting.None)
```
drwill-ms, line 183: These included a type from the Newtonsoft.Json library, which we wanted to remove. In doing so, we realized these methods aren't necessary, so we just removed them.

patilsnr (approved):

abhipsaMisra (approved):

PR #2893, opened 2022-10-19
Get rid of a previous NRE, and nearby updates

drwill-ms:


iothub/service/src/Twin/Models/TwinCollection.cs
```diff
@@ -115,6 +115,7 @@ public int Count
         /// </summary>
         /// <param name="propertyName">Name of the property to get.</param>
         /// <returns>Value for the given property name.</returns>
+        /// <exception cref="InvalidOperationException">When the specified <paramref name="propertyName"/> does not exist in the collection.</exception>
         public dynamic this[string propertyName]
```
drwill-ms, line 119: We've chosen this as the exception of choice when the user does something the SDK doesn't support at runtime.

iothub/service/src/Twin/Models/TwinCollection.cs
```diff
@@ -157,14 +158,18 @@ public TwinMetadata GetMetadata()
...
-        /// Gets the last updated time for this property.
+        /// Gets the last time this property was updated in UTC.
         /// </summary>
-        /// <returns>Date-time instance representing the last updated time for this property.</returns>
-        /// <exception cref="System.NullReferenceException">Thrown when the metadata object is null.
-        /// An example would be when the this class is created with the default constructor.</exception>
         public DateTimeOffset GetLastUpdatedOnUtc()
         {
-            return (DateTime)_metadata[LastUpdatedName];
+            if (_metadata != null
```
drwill-ms, line 165: Prevent NRE

iothub/service/src/Twin/Models/TwinCollectionArray.cs
```diff
@@ -62,7 +70,7 @@ public DateTimeOffset GetLastUpdatedOnUtc()
         /// <returns>Last updated version if present, null otherwise.</returns>
         public long? GetLastUpdatedVersion()
         {
-            return (long?)_metadata[TwinCollection.LastUpdatedVersionName];
+            return (long?)_metadata?[LastUpdatedVersionName];
```
drwill-ms, line 73: Prevent NRE

tmahmood-microsoft (approved):

PR #2899, opened 2022-10-19
Tags property on twin should just be a dictionary

drwill-ms:
If you edit a device's twin's tags in the portal, they show up in the twin document as a simple object with key/value pairs where you can't have 2 entries with the same key. IOW, it is just a dictionary. We don't need all the other stuff that comes with a TwinCollection, like metadata.


timtay-microsoft (approved):
You are correct. It seems that tags don't have metadata, so a simple map<string, object> works fine here.

abhipsaMisra (approved):

PR #2905, opened 2022-10-21
Create IoT hub devices within samples

patilsnr:
Our device samples currently rely on existing devices in an IoT hub to run. This can lead to cluttering and possible frustration for users who don't expect to already need an existing device to run our code. This change creates the option for creating a device within the sample, and then running it normally.

### DIFF
--- a/iothub/device/samples/getting started/FileUploadSample/Program.cs
+++ b/iothub/device/samples/getting started/FileUploadSample/Program.cs
@@ -3,6 +3,7 @@
 
 using CommandLine;
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace Microsoft.Azure.Devices.Client.Samples
@@ -34,13 +35,43 @@ namespace Microsoft.Azure.Devices.Client.Samples
                     Environment.Exit(1);
                 });
 
-            using var deviceClient = DeviceClient.CreateFromConnectionString(
-                parameters.PrimaryConnectionString,
-                parameters.TransportType);
-            var sample = new FileUploadSample(deviceClient);
-            await sample.RunSampleAsync();
+            // If both IoT hub and device connection strings exist, default behavior uses existing device
+            if (parameters.DeviceConnectionString != null)
+            {
+                // run sample like normal
+                using var deviceClient = DeviceClient.CreateFromConnectionString(
+                    parameters.DeviceConnectionString,
+                    parameters.TransportType);
 
-            await deviceClient.CloseAsync();
+                var sample = new FileUploadSample(deviceClient);
+                await sample.RunSampleAsync();
+
+                await deviceClient.CloseAsync();
+            } 
+            else if (parameters.HubConnectionString != null)
+            {
+                // Deploy new device to IoT hub
+                using var registryManager = RegistryManager.CreateFromConnectionString(parameters.HubConnectionString);
+
+                string deviceId = $"FileUploadSample_{Guid.NewGuid()}";
+                var tempDevice = new Device(deviceId);
+                devicesToDelete.Add(tempDevice);
+
+                await registryManager.AddDeviceAsync(tempDevice);
+                using var deviceClient = DeviceClient.CreateFromConnectionString(
+                    parameters.HubConnectionString,
+                    deviceId,
+                    parameters.TransportType);
+
+                var sample = new FileUploadSample(deviceClient);
+                await sample.RunSampleAsync();
+
+                await deviceClient.CloseAsync();
+            }
+            else
+            {
+                throw new ArgumentNullException("Must specify either a device or IoT hub connection string.");
+            }
 
             Console.WriteLine("Done.");
             return 0;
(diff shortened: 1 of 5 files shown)


iothub/device/samples/getting started/FileUploadSample/Program.cs
```diff
@@ -34,13 +35,42 @@ public static async Task<int> Main(string[] args)
...
+            else if (parameters.HubConnectionString != null)
+            {
+                // Deploy new device to IoT hub
+                using var registryManager = RegistryManager.CreateFromConnectionString(parameters.HubConnectionString);
+
+                string deviceId = $"FileUploadSample_{Guid.NewGuid()}";
+                var tempDevice = new Device(deviceId);
+
+                await registryManager.AddDeviceAsync(tempDevice);
+                using var deviceClient = DeviceClient.CreateFromConnectionString(
```
abhipsaMisra, line 60: 'd add in a log statement that logs out the created device Id

iothub/device/samples/getting started/FileUploadSample/Program.cs
```diff
@@ -34,13 +35,42 @@ public static async Task<int> Main(string[] args)
...
+                    parameters.HubConnectionString,
+                    deviceId,
+                    parameters.TransportType);
+
+                var sample = new FileUploadSample(deviceClient);
+                await sample.RunSampleAsync();
+
+                await deviceClient.CloseAsync();
```
abhipsaMisra, line 68: we'll also want to clean up (delete) the created device for in this scenario

iothub/device/samples/getting started/FileUploadSample/Program.cs
```diff
@@ -34,13 +35,42 @@ public static async Task<int> Main(string[] args)
...
+            }
+            else
+            {
+                throw new ArgumentNullException("Must specify either a device or IoT hub connection string.");
```
abhipsaMisra, line 72: nit: we can probably add a reference recommending users to look at --help

abhipsaMisra (commented):
I do like this approach

iothub/device/samples/getting started/FileUploadSample/Program.cs
```diff
@@ -34,13 +35,42 @@ public static async Task<int> Main(string[] args)
...
+                await sample.RunSampleAsync();
+
+                await deviceClient.CloseAsync();
+            } 
+            else if (parameters.HubConnectionString != null)
+            {
+                // Deploy new device to IoT hub
+                using var registryManager = RegistryManager.CreateFromConnectionString(parameters.HubConnectionString);
+
+                string deviceId = $"FileUploadSample_{Guid.NewGuid()}";
```
schoims, line 56: Let's get and set deviceId as a parameter

iothub/device/samples/getting started/FileUploadSample/Program.cs
schoims, line 68: Line 65 - 68 are repeated maybe we could refactor these? what do you think?

timtay-microsoft:
I have mixed feelings about this approach. I like not having hardcoded devices in our test hubs, but this feels like it is solving a problem for us while making sample code more complicated than it needs to be. The fewer parameters we expose, the easier it will be for a first-time user to grasp what is being demonstrated.

Maybe it's time we re-evaluated running these samples nightly? Now that we have the samples living in our main repo, and the features they demonstrate are covered by e2e tests, what are we testing when we run these samples?

iothub/device/samples/getting started/FileUploadSample/Program.cs
patilsnr: Do we have other samples that allow users to pass in a desired device Id as a parameter? Also, would it make sense to allow a user to specify a device Id if it's just going to be deleted at the end of the sample?

schoims: https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/provisioning/service/samples/Getting%20Started/IndividualEnrollmentX509Sample/Parameters.cs#L29

patilsnr:
> I have mixed feelings about this approach. I like not having hardcoded devices in our test hubs, but this feels like it is solving a problem for us while making sample code more complicated than it needs to be. The fewer parameters we expose, the easier it will be for a first-time user to grasp what is being demonstrated.
> 
> Maybe it's time we re-evaluated if it is even worth running these samples nightly? Now that we have the samples living in our main repo, and the features they demonstrate are covered by e2e tests, what are we testing when we run these samples?

This is a fair point.

iothub/device/samples/getting started/FileUploadSample/Program.cs
schoims: Ohh we are removing the device after, makes sense

abhipsaMisra:
> Now that we have the samples living in our main repo, and the features they demonstrate are covered by e2e tests, what are we testing when we run these samples?

We'll still want to make sure that any samples we put out publicly run as expected.

iothub/device/samples/getting started/FileUploadSample/Program.cs
andyk-ms: most tests has a GUID as the deviceID.
what about a generic one like this?
nameof(<testname>)_{Guid.NewGuid()}

PR #2909, opened 2022-10-24
Remove app insights logging from tests in main

ngastelum-ms:


drwill-ms (changes requested):
Needs the e2e script+json+bicep updates to remove AppInsights resource from being created.

drwill-ms (approved):

PR #2910, opened 2022-10-24
Disambiguate twin classes from hub and dps service clients

drwill-ms:
Especially for customers that use our SDK clients together in the same file, we prefer to avoid naming conflicts that they otherwise have to disambiguate by using the full namespace (or aliasing types).

For this reason, I've renamed most twin types in hub service client to be `ClientTwin*` and in dps service client to be `ProvisioningTwin*`.

I also took a look at the other type names and resolved some wrongness (like DeviceStatus when it also applies to modules, so renamed to ClientStatus).

The other change was TwinCollection and TwinProperties. I spent more time on this than anything. The user doesn't really interact with the TwinCollection type except off of the Twin class. It contains the Desired and Reported properties. So, I renamed that to ClientTwinDocument, which isn't a great name but not one the user should ever use directly. That freed up the TwinProperties name for what was TwinCollection, so it is now ClientTwinProperties. The other related types followed suit, so ClientTwinPropertyValue and ClientTwinPropertyArray.

Finally, the hub device client has some small number of classes for twin. I renamed it from ClientTwin to just Twin.

Thoughts?

### DIFF
--- a/e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs
+++ b/e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                     async (r, d) =>
                     {
                         Device device = await r.Devices.GetAsync(d).ConfigureAwait(false);
-                        device.Status = DeviceStatus.Disabled;
+                        device.Status = ClientStatus.Disabled;
                         await r.Devices.SetAsync(device).ConfigureAwait(false);
                     })
                 .ConfigureAwait(false);
@@ -67,7 +67,7 @@ namespace Microsoft.Azure.Devices.E2ETests
                 async (r, d) =>
                 {
                     Device device = await r.Devices.GetAsync(d).ConfigureAwait(false);
-                    device.Status = DeviceStatus.Disabled;
+                    device.Status = ClientStatus.Disabled;
                     await r.Devices.SetAsync(device).ConfigureAwait(false);
                 })
                 .ConfigureAwait(false);
@@ -126,7 +126,7 @@ namespace Microsoft.Azure.Devices.E2ETests
 
             // Receiving the module twin should succeed right now.
             VerboseTestLogger.WriteLine($"{nameof(IotHubDeviceClient_Gives_ConnectionStatus_Disconnected_ChangeReason_DeviceDisabled_Base)}: DeviceClient GetTwinAsync.");
-            ClientTwin twin = await deviceClient.GetTwinAsync().ConfigureAwait(false);
+            Client.ClientTwin twin = await deviceClient.GetTwinAsync().ConfigureAwait(false);
             Assert.IsNotNull(twin);
 
             // Delete/disable the device in IoT hub. This should trigger the ConnectionStatusChangeHandler.
@@ -179,7 +179,7 @@ namespace Microsoft.Azure.Devices.E2ETests
 
             // Receiving the module twin should succeed right now.
             VerboseTestLogger.WriteLine($"{nameof(IotHubModuleClient_Gives_ConnectionStatus_Disconnected_ChangeReason_DeviceDisabled_Base)}: ModuleClient GetTwinAsync.");
-            ClientTwin twin = await moduleClient.GetTwinAsync().ConfigureAwait(false);
+            Client.ClientTwin twin = await moduleClient.GetTwinAsync().ConfigureAwait(false);
             Assert.IsNotNull(twin);
 
             // Delete/disable the device in IoT hub.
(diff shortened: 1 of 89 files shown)


provisioning/service/src/Twin/ProvisioningTwinState.cs
```diff
@@ -9,10 +9,10 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
     /// Representation of a single Twin initial state for the Device Provisioning Service.
     /// </summary>
     /// <remarks>
-    /// The TwinState can contain one <see cref="TwinCollection"/> of Tags, and one
-    /// <see cref="TwinCollection"/> of properties.desired.
+    /// The TwinState can contain one <see cref="ProvisioningTwinProperties"/> of Tags, and one
+    /// <see cref="ProvisioningTwinProperties"/> of properties.desired.
```
patilsnr, line 13: should `properties.desired` have some kind of reference link? It isn't clear to me what that means.

patilsnr (approved):
Optimistic sign off

tmahmood-microsoft (approved):

e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs
```diff
@@ -126,7 +126,7 @@ void StatusChangeHandler(ConnectionStatusInfo c)
 
             // Receiving the module twin should succeed right now.
             VerboseTestLogger.WriteLine($"{nameof(IotHubDeviceClient_Gives_ConnectionStatus_Disconnected_ChangeReason_DeviceDisabled_Base)}: DeviceClient GetTwinAsync.");
-            ClientTwin twin = await deviceClient.GetTwinAsync().ConfigureAwait(false);
+            Twin twin = await deviceClient.GetTwinAsync().ConfigureAwait(false);
```
abhipsaMisra, line 129: Are we going from ClientTwin -> Twin here because we're reserving ClientTwin for the service's view of the "twin"?

e2e/test/iothub/service/BulkOperationsE2ETests.cs
```diff
@@ -27,13 +27,13 @@ public async Task BulkOperations_UpdateTwins2Device_Ok()
             using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
             using var serviceClient = new IotHubServiceClient(TestConfiguration.IotHub.ConnectionString);
 
-            Twin twin = await serviceClient.Twins.GetAsync(testDevice.Id).ConfigureAwait(false);
+            ClientTwin twin = await serviceClient.Twins.GetAsync(testDevice.Id).ConfigureAwait(false);
```
abhipsaMisra, line 30: I'm probably nitpicking here but do you think the "Twins" client of the service client will cause confusion with the "device" counterpart? Should we rename it to "ClientTwin"?

iothub/device/samples/getting started/TwinSample/TwinSample.cs
```diff
@@ -34,14 +34,14 @@ public async Task RunSampleAsync(TimeSpan sampleRunningTime)
...
-            ClientTwin twin = await _deviceClient.GetTwinAsync();
+            Twin twin = await _deviceClient.GetTwinAsync();
 
             Console.WriteLine("\tInitial twin value received:");
             Console.WriteLine($"\tDesired properties: {twin.RequestsFromService.GetSerializedString()}");
             Console.WriteLine($"\tReported properties: {twin.ReportedByClient.GetSerializedString()}");
 
             Console.WriteLine("Sending sample start time as reported property");
-            var reportedProperties = new ReportedPropertyCollection
+            var reportedProperties = new ReportedProperties
```
abhipsaMisra, line 44: This looks good

provisioning/service/src/Models/IndividualEnrollment.cs
```diff
@@ -189,7 +189,7 @@ public Attestation Attestation
         /// Capabilities of the device.
         /// </summary>
         [JsonProperty(PropertyName = "capabilities", DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public DeviceCapabilities Capabilities { get; set; }
+        public ProvisioningDeviceCapabilities Capabilities { get; set; }
```
abhipsaMisra, line 192: Are capabilities only relevant for devices? (The comment seems to suggest that they're only for devices).

abhipsaMisra (approved):

provisioning/service/src/Models/IndividualEnrollment.cs
drwill-ms: The only property in it is IsIotEdge, which only applies to devices.

Sure, one day they could extend this to be other things that might also apply to modules, but at this point I doubt that.

drwill-ms: That said, I'm not against renaming this to `*ClientCapabilities`.

e2e/test/iothub/ConnectionStatusChangeHandlerTests.cs
drwill-ms: Yes, for the device client itself, it seemed more natural to use Twin here, as opposed to anywhere else.

drwill-ms: Renamed to TwinProperties on the device-side.

e2e/test/iothub/service/BulkOperationsE2ETests.cs
drwill-ms: Saving grace is that the data type for TwinsClient is probably never explicitly used. As discussed, we'll wait for negative feedback to make any changes.

PR #2913, opened 2022-10-24
Turn SDK log collection off by default

drwill-ms:
Same, but for `main`.

### DIFF
--- a/e2e/test/E2EMsTestBase.cs
+++ b/e2e/test/E2EMsTestBase.cs
@@ -33,10 +33,25 @@ namespace Microsoft.Azure.Devices.E2ETests
         // The test timeout for e2e tests that involve testing token refresh
         protected const int TokenRefreshTestTimeoutMilliseconds = 20 * 60 * 1000; // 20 minutes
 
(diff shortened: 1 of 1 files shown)


e2e/test/E2EMsTestBase.cs
```diff
@@ -33,10 +33,24 @@ public class E2EMsTestBase : IDisposable
         // The test timeout for e2e tests that involve testing token refresh
         protected const int TokenRefreshTestTimeoutMilliseconds = 20 * 60 * 1000; // 20 minutes
 
+        private const string CollectSdkLogsEnvVar = "COLLECT_SDK_LOGS";
+        public static readonly bool CollectSdkLogs;
```
brycewang-microsoft, line 37: `s_collectSdkLogs` ?

brycewang-microsoft (approved):

e2e/test/E2EMsTestBase.cs
drwill-ms: You're right. It should have been s_.

PR #2915, opened 2022-10-25
Remove unnecessary Fx.cs class method

ngastelum-ms:


brycewang-microsoft (commented):
Do you plan to remove the method `AssertAndThrow` from Fx.cs in this PR?

iothub/device/src/Fx/Fx.cs
```diff
@@ -10,14 +10,6 @@ namespace Microsoft.Azure.Devices.Client
...
     {
-        public static void AssertAndThrow(bool condition, string description)
-        {
-            if (!condition)
-            {
-                throw new InvalidOperationException(description);
-            }
-        }
-
         public static bool IsFatal(Exception ex)
```
drwill-ms, line 13: Can we move this method to the exception class as an internal/private static too?

iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
```diff
@@ -433,5 +433,13 @@ private static IAsyncResult ToAsyncResult<TResult>(Task<TResult> task, AsyncCall
 
             return tcs.Task;
         }
+
+        private static void AssertAndThrow(bool condition, string description)
```
drwill-ms, line 437: I don't see the point of this helper method. I'd rather just see the code inline.

Worst case, we should add it to the Argument.cs file.

iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
```diff
@@ -59,8 +59,8 @@ public override bool WriteAsync(TransportAsyncCallbackArgs args)
         {
             ThrowIfNotOpen();
 
-            Fx.AssertAndThrow(args.Buffer != null || args.ByteBufferList != null, "must have a buffer to write");
-            Fx.AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
+            AssertAndThrow(args.Buffer != null || args.ByteBufferList != null, "must have a buffer to write");
+            AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
```
drwill-ms, line 63:
```suggestion
            AssertAndThrow(args.CompletedCallback != null, "Must have a valid callback.");
```

iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
```diff
@@ -59,8 +59,8 @@ public override bool WriteAsync(TransportAsyncCallbackArgs args)
         {
             ThrowIfNotOpen();
```
drwill-ms, line 61: Given we're in an internal file for transport, when would these conditions fail and what would an exception mean to the user if they observed it? What would they do differently in their code to avoid it?

drwill-ms (changes requested):
Comments and questions inline.

iothub/device/src/Fx/Fx.cs
brycewang-microsoft: We can if we want to, but the codes calling it will be something like 
```c#
catch (Exception ex) when (!IotHubClientException.IsFatal(ex))
{
    ......
}
```
kindof weird IMO. 

I would like to keep this method within a separate file but maybe we can move this file into the "Exceptions" folder for the hub device and service packages.

drwill-ms: Does `ex.IsTransient == !IsFatal(ex)`?

If so, can we simply use ex.IsTransient instead?

drwill-ms: Oh, a fatal exception may not always be IotHubClientException, right? I forgot about that.

brycewang-microsoft: Yes, that's what I am worried about.

brycewang-microsoft (approved):

iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
```diff
@@ -59,8 +59,14 @@ public override bool WriteAsync(TransportAsyncCallbackArgs args)
...
 
-            Fx.AssertAndThrow(args.Buffer != null || args.ByteBufferList != null, "must have a buffer to write");
-            Fx.AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
+            if (args.Buffer == null || args.ByteBufferList == null)
+            {
+                throw new InvalidOperationException("Must have a buffer to write");
+            }
+            if (args.CompletedCallback == null)
+            {
+                throw new InvalidOperationException("Must have a valid callback");
```
drwill-ms, line 68:
```suggestion
                throw new InvalidOperationException("Must have a valid callback.");
```

iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
```diff
@@ -78,8 +84,14 @@ public override bool ReadAsync(TransportAsyncCallbackArgs args)
             ThrowIfNotOpen();
 
             // Read with buffer list not supported
-            Fx.AssertAndThrow(args.Buffer != null, "must have buffer to read");
-            Fx.AssertAndThrow(args.CompletedCallback != null, "must have a valid callback");
+            if (args.Buffer == null)
+            {
+                throw new InvalidOperationException("Must have buffer to read");
```
drwill-ms, line 89:
```suggestion
                throw new InvalidOperationException("Must have buffer to read.");
```

drwill-ms (approved):

PR #2925, opened 2022-10-27
Remove TPM tests from PR gate

ngastelum-ms:
Tpm tests seem to be flakey and are blocking PRs, they pass when run individually but not as a group

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -24,7 +24,6 @@ Parameters:
     -framework: Select which framework to run tests on. Allowed values examples include, but are not limited to, "netcoreapp3.1", "net47", "net451"
     -skipIotHubTests: Provide this flag if you want to skip all IoT Hub integration tests
     -skipDPSTests: Provide this flag if you want to skip all DPS integration tests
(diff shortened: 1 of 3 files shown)


build.ps1
```diff
@@ -70,6 +71,7 @@ Param(
     [string] $framework = "*",
     [switch] $skipIotHubTests,
     [switch] $skipDPSTests,
+    [switch] $skipTPMTests,
```
abhipsaMisra, line 74: How about we mark them as flaky, as their behavior is currently flaky.
You could add a comment explaining that TPM only accepts a single connection at a time, and we don't have a way to effectively ensure that only a single test will make a call to TPM since we have class-level test parallelization. This makes the test behavior non-deterministic and flaky.

abhipsaMisra: There is an existing test category called "Flaky"

ngastelum-ms: Makes sense, changed.

drwill-ms (approved):

PR #2927, opened 2022-10-28
Add retry policies into DPS service

brycewang-microsoft:
Also add `Logging` for DPS service client in this PR.


provisioning/service/src/ProvisioningServiceClient.cs
```diff
@@ -46,9 +48,13 @@ public ProvisioningServiceClient(string connectionString, ProvisioningServiceCli
                 _provisioningConnectionString,
                 clientOptions);
 
-            IndividualEnrollments = new IndividualEnrollmentsClient(_provisioningConnectionString, _contractApiHttp);
-            EnrollmentGroups = new EnrollmentGroupsClient(_provisioningConnectionString, _contractApiHttp);
-            DeviceRegistrationStates = new DeviceRegistrationStatesClient(_provisioningConnectionString, _contractApiHttp);
+            _retryPolicy = clientOptions.RetryPolicy ?? new ProvisioningServiceNoRetry();
```
drwill-ms, line 51: Why no retry by default?

drwill-ms (approved):

provisioning/service/src/ProvisioningServiceClient.cs
brycewang-microsoft: No, the [default](https://github.com/Azure/azure-iot-sdk-csharp/blob/brycewang/retry-dps-service/provisioning/service/src/ProvisioningServiceClientOptions.cs#L39) retry policy would be exponential backoff.

`ProvisioningServiceNoRetry` would be used when RetryPolicy is set to `null`.

PR #2928, opened 2022-10-31
Catch TLS exceptions during connect

brycewang-microsoft:
Addressing [this](https://github.com/Azure/azure-iot-sdk-csharp/pull/2827#discussion_r988415081)

### DIFF
--- a/iothub/device/src/Exceptions/IotHubClientErrorCode.cs
+++ b/iothub/device/src/Exceptions/IotHubClientErrorCode.cs
@@ -125,5 +125,10 @@ namespace Microsoft.Azure.Devices.Client
         /// guide for more information.
         /// </remarks>
         Unauthorized,
+
+        /// <summary>
+        /// The request failed because of TLS authentication error.
+        /// </summary>
+        AuthenticationError,
     }
 }
--- a/iothub/device/src/Exceptions/IotHubClientErrorCode.cs
+++ b/iothub/device/src/Exceptions/IotHubClientErrorCode.cs
@@ -129,6 +129,11 @@ namespace Microsoft.Azure.Devices.Client
         /// <summary>
         /// The request failed because of TLS authentication error.
         /// </summary>
-        AuthenticationError,
+        /// <remarks>
+        /// This error may happen when the remote certificate presented could not be validated, the TLS version is different between client requested
+        /// auth and service minimum requirement, cipher suites to be used could not be agreed upon, etc. The best course of action is to check your
+        /// device certificates and ensure they are up-to-date.
+        /// </remarks>
+        TlsAuthenticationError,
     }
 }
--- a/iothub/device/src/Pipeline/ErrorDelegatingHandler.cs
+++ b/iothub/device/src/Pipeline/ErrorDelegatingHandler.cs
@@ -150,8 +150,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
 
                     if (IsSecurityExceptionChain(ex))
                     {
-                        Exception innerException = (ex is IotHubClientException) ? ex.InnerException : ex;
-                        throw new AuthenticationException("TLS authentication error.", innerException);
+                        throw new IotHubClientException("TLS authentication error.", IotHubClientErrorCode.AuthenticationError, ex);
                     }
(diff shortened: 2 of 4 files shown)


iothub/device/src/Pipeline/ErrorDelegatingHandler.cs
```diff
@@ -150,8 +150,7 @@ private async Task<T> ExecuteWithErrorHandlingAsync<T>(Func<Task<T>> asyncOperat
 
                     if (IsSecurityExceptionChain(ex))
                     {
-                        Exception innerException = (ex is IotHubClientException) ? ex.InnerException : ex;
```
abhipsaMisra, line 153: We'll want to hold on to this unwrapping logic. IsSecurityExceptionChain unwinds the entire exception chain to see if any exception underneath is a security exception. 
In case the top level exception is an IotHubClientException to begin with (because we did some transformation in the transport layer), then we'll have an IotHubClientException wrapping another IotHubClientException.

iothub/device/tests/Pipeline/ErrorDelegatingHandlerTests.cs
```diff
@@ -103,7 +103,7 @@ public async Task ErrorHandler_TransientErrorOccuredChannelIsAlive_ChannelIsTheS
         [TestMethod]
         public async Task ErrorHandler_SecurityErrorOccured_ChannelIsAborted()
         {
-            await TestExceptionThrown(typeof(TestSecurityException), typeof(AuthenticationException)).ConfigureAwait(false);
+            await TestExceptionThrown(typeof(TestSecurityException), typeof(IotHubClientException)).ConfigureAwait(false);
```
abhipsaMisra, line 106: Can we add an error code check to this exception assertion? That'll make the assertion specific to this test case.

iothub/device/src/Exceptions/IotHubClientErrorCode.cs
```diff
@@ -125,5 +125,10 @@ public enum IotHubClientErrorCode
         /// guide for more information.
         /// </remarks>
         Unauthorized,
+
+        /// <summary>
+        /// The request failed because of TLS authentication error.
```
abhipsaMisra, line 130: Throwing a question out there - what would be our recommendation for this case? Check your device certificates to ensure they are up-to-date?

abhipsaMisra (approved):
I'm assuming the linked sample changes are going to come in another PR?

iothub/device/src/Exceptions/IotHubClientErrorCode.cs
```diff
@@ -125,5 +125,10 @@ public enum IotHubClientErrorCode
...
+        /// </summary>
+        AuthenticationError,
```
timtay-microsoft, line 132: I'd consider renaming this ```TlsAuthenticationError``` to be more specific. Otherwise, at a glance, I would be confused why we'd have both ```Unauthorized``` and ```AuthenticationError```

brycewang-microsoft:
> I'm assuming the linked sample changes are going to come in another PR?

That linked sample is actually on `main`, and the reconnection sample on `previews/v2` here can leverage the changes in this PR.

iothub/device/tests/Pipeline/ErrorDelegatingHandlerTests.cs
```diff
@@ -96,95 +98,71 @@ public async Task ErrorHandler_TransientErrorOccuredChannelIsAlive_ChannelIsTheS
...
-                di => di.Received(2).OpenAsync(Arg.Any<CancellationToken>()),
-                thrownExceptionType, expectedExceptionType).ConfigureAwait(false);
-        }
-
-        private static async Task OperationAsync_ExceptionThrownAndThenSucceed_OperationSuccessfullyCompleted(
-            Func<IDelegatingHandler, Task<Message>> mockSetup,
-            Func<IDelegatingHandler, Task<Message>> act,
-            Func<IDelegatingHandler, Task<Message>> assert,
-            Type thrownExceptionType,
-            Type expectedExceptionType)
```
brycewang-microsoft, line 149: remove unused private method

PR #2931, opened 2022-11-02
Fixed M2M message (MQTT) failure after Direct Method from Edge Hub

tmahmood-microsoft:
Module-to-module message (MQTT) failed after receiving a direct method from EdgeHub. The issue was occurring due to mis-matched Ack for the M2M message after direct method. The Mqtt client was not processing the Ack for direct method properly that caused sending incorrect Ack for each M2M message after receiving direct method.


tmahmood-microsoft:
Tested against MQTT (TCP and Web Socket)

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -719,6 +719,10 @@ private async void ProcessMessageAsync(IChannelHandlerContext context, Packet pa
 
                     case PacketType.PUBLISH:
                         ProcessPublish(context, (PublishPacket)packet);
+                        if (((PublishPacket)packet).TopicName.StartsWith(MqttTransportHandler.MethodPostTopicPrefix, StringComparison.OrdinalIgnoreCase))
```
drwill-ms, line 722: Please add a code comment as to why we do this conditionally.

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -43,7 +44,8 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
         {
             if (!_incompleteQueue.Any())
             {
-                throw new IotHubException("Nothing to complete.", isTransient: false);
+                Debug.Assert(_incompleteQueue.Any(), "Nothing to complete.");
```
drwill-ms, line 47:
```suggestion
                Debug.Fail("Nothing to complete.");
```

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -43,7 +44,8 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
         {
             if (!_incompleteQueue.Any())
             {
```
drwill-ms, line 46: Add a call to the logs to warn us this happened.

drwill-ms (approved):
We'll probably need an update in our latest LTS as well as in v2.

abhipsaMisra (approved):

tmahmood-microsoft:
> We'll probably need an update in our latest LTS as well as in v2.

Testing and adding changes to v2

patilsnr (approved):

timtay-microsoft:
I'm actually still a little confused about this bug. Was the previous behavior that our SDK would ack a QoS 1 message (like a direct method), or was it something else?

tmahmood-microsoft:
For a M2M message we would Ack QoS 1 message but not a direct method. 
https://github.com/Azure/azure-iot-sdk-csharp/blob/e9c06e8f23f0be7445f9b2ccd4cdf10b14b34dca/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs#L185

My recent change makes sure we Ack message Direct method with QoS 1

timtay-microsoft:
> For a M2M message we would Ack QoS 1 message but not a direct method.
> 
> https://github.com/Azure/azure-iot-sdk-csharp/blob/e9c06e8f23f0be7445f9b2ccd4cdf10b14b34dca/iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs#L185
> 
> My recent change makes sure we Ack Direct method with QoS 1

This fix feels a bit odd, then. Why would we need a special case like this if we are already sending acks for other QoS 1 messages like M2M messages?

tmahmood-microsoft:
Ack for M2M message is triggered by CompleteAsync() call. For Direct method we need to do it explicitly here for QoS 1

timtay-microsoft:
Okay, I can live with this approach, then.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -719,6 +719,12 @@ private async void ProcessMessageAsync(IChannelHandlerContext context, Packet pa
 
                     case PacketType.PUBLISH:
                         ProcessPublish(context, (PublishPacket)packet);
+                        // Checking if acknowledgement needs to be sent after receiving method from Hub with QoS = 1
```
timtay-microsoft, line 722: You need to make it more clear here why we are only sending the ack if it was a direct method message with QoS 1. Explain that the other QoS messages are ack'd elsewhere

timtay-microsoft (approved):

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -43,7 +45,11 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
...
-                throw new IotHubException("Nothing to complete.", isTransient: false);
+                if (Logging.IsEnabled)
+                    Logging.Error(context, "Nothing to complete.", nameof(CompleteWorkAsync));
```
drwill-ms, line 49:
```suggestion
                    Logging.Error(context, "{nameof(CompleteWorkAsync)} called but there are no items in the queue to complete!", nameof(CompleteWorkAsync));
```

PR #2934, opened 2022-11-04
Remove tests that validate service behavior, and fix another failing …

drwill-ms:
…test


patilsnr (approved):

brycewang-microsoft (approved):

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -817,7 +817,8 @@ public async Task DPS_Registration_AmqpWs_X509_IndividualEnrollment_InvalidGloba
                     auth,
                     clientOptions);
 
-                using var cts = new CancellationTokenSource(FailingTimeoutMiliseconds);
+                // Needs enough time for the transport to timeout
+                using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));
```
schoims, line 821: FailingTimeoutMiliseconds initialized can be removed?

schoims (approved):

e2e/test/provisioning/ProvisioningE2ETests.cs
abhipsaMisra, line 821: Do we know where the bottleneck is for the transport layer timeout? Is there a transport setting we can modify to force the transport layer to timeout sooner?

drwill-ms: Good question. I'll try that.

Although, we didn't appear to get rid of the default timeout in DPS for v2. We should do that.

drwill-ms: My mistake. It is IdleTimeout we still have. That is fine.

I think this is the relevant code:

```csharp
            _mqttClientOptionsBuilder
                .WithTls(tlsParameters)
                .WithProtocolVersion(MQTTnet.Formatter.MqttProtocolVersion.V311) // 3.1.1
                .WithCleanSession(_mqttTransportSettings.CleanSession)
                .WithKeepAlivePeriod(_mqttTransportSettings.IdleTimeout)
                .WithTimeout(TimeSpan.FromMilliseconds(-1)); // MQTTNet will only time out if the cancellation token requests cancellation.
```

We don't let the user influence the timeout.

We don't want it to timeout, though. We want it to realize the endpoint doesn't existing.

abhipsaMisra: > We don't want it to timeout, though. We want it to realize the endpoint doesn't existing.

Right; does MqttNet take over a couple of seconds to realize that the endpoint it wants to connect to doesn't exist? Or does it keep on retrying the operation?

Issue #2351, opened 2022-04-19
[Bug Report] Unexpected ObjectDisposedException when IoT Hub is throttling


danigian:
# Context

- **OS, version, SKU and CPU architecture used:** arm32/x86 debian bullseye based docker image
- **Application's .NET Target Framework :** net6
- **Device:** laptop/rpi
- **SDK version used:** 1.40.0 (not sure about IoT Hub C# SDK version used underneath by IoT Edge)

## Description of the issue
Cross-posting from https://github.com/Azure/iotedge/issues/6042

Assuming a IoT Hub with 100 devices, one edge gateway and one module with the following code

``` csharp
// Creating registry manager in order to get 50 connection strings
var registry = RegistryManager.CreateFromConnectionString(iothubownerstring);

var deviceTwins = await registry.CreateQuery("select * from devices", 50).GetNextAsTwinAsync();

// Getting 50 devices
var listDeviceTasks = new List<Task<Device>>();
foreach (var element in deviceTwins)
{
	listDeviceTasks.Add(registry.GetDeviceAsync(element.DeviceId));
}
var devices = await Task.WhenAll(listDeviceTasks);

// Building up the connection string manually
var connectionStrings = new List<string>();
foreach (var d in devices)
{
	connectionStrings.Add($"HostName={iothubname}.azure-devices.net;DeviceId={d.Id};SharedAccessKey={d.Authentication.SymmetricKey.PrimaryKey};GatewayHostName={gwhostname}");
}

// Setting AMQP Transport Settings in such a way that Operation Timeout is 10 seconds instead of default 60secs
var amqp = new AmqpTransportSettings(Microsoft.Azure.Devices.Client.TransportType.Amqp_Tcp_Only)
{
	AmqpConnectionPoolSettings = new AmqpConnectionPoolSettings()
	{
		Pooling = true,
		MaxPoolSize = 1
	},
	OperationTimeout = TimeSpan.FromSeconds(3)
};
amqp.RemoteCertificateValidationCallback += (_, _, _, _) => true;

// Constructing list of GetTwin Task
var tasks = new List<Task<Twin>>();
foreach (var element in connectionStrings)
{
	var client = DeviceClient.CreateFromConnectionString(element, new[] { amqp });

	client.SetRetryPolicy(new ExponentialBackoff(int.MaxValue,
				minBackoff: TimeSpan.FromMilliseconds(100),
				maxBackoff: TimeSpan.FromSeconds(10),
[...]

Issue #2339, opened 2022-04-08
[Bug Report] AmqpAuthenticationRefresher: Odd disposal/cancel behavior (ObjectDisposedException)


rgrace-puck:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 10 x64
- **Application's .NET Target Framework :** net48
- **Device:** Azure App Service
- **SDK version used:** 1.40.0 (Microsoft.Azure.Devices.Client)

## Description of the issue
To preface this, the usage pattern described in this issue is unusual and we're aware of that but I believe this issue still affects other consumers of this SDK.

We have a web app which creates a short lived `DeviceClient` (<=60s alive) and sends messages back to our field equipment via an API controller. When deployed, we start to see `ObjectDisposedException` (thousands per hour) in our Application Insights logs after an extended period of time (usually ~20-30 minutes). We do have a try/catch block around the IoT Hub operations but this exception is not caught there. 

After doing some investigation/testing, using `ReceiveAsync(CancellationToken)` (instead of the TimeSpan equivalent) and manually cancelling before disposal of the client yielded no errors. 

## Additional notes
Looking at the internals of the library, it seems the behavior of using `TimeoutHelper` to generate a cancellation token is flawed with `AmqpTransportHandler` in combination with the refresh loop behavior in `AmqpAuthenticationRefresher`. 

For example, there is a method in `AmqpAuthenticationRefresher` called `StopLoop` but this method doesn't seem to do anything except log that it was called. 

These are very basic findings from glancing at the repository so it may be lacking context / inaccurate.

## Code sample exhibiting the issue
NOTE: This is a simplified version and not the actual code we're using but is effectively the same operation.
```cs
using Microsoft.Azure.Devices.Client;
using System;
using System.IO;
using System.Threading.Tasks;

namespace iot_test
{
    public class TestController
    {
        private static AmqpTransportSettings GetAmqpTransportSettings()
        {
            return new AmqpTransportSettings(TransportType.Amqp_Tcp_Only)
            {
                AmqpConnectionPoolSettings = new AmqpConnectionPoolSettings
                {
                    Pooling = true,
                    MaxPoolSize = 1
                },
                OpenTimeout = TimeSpan.FromSeconds(5),
                OperationTimeout = TimeSpan.FromSeconds(5)
            };
        }

        private static readonly ITransportSettings[] TransportSettings = { GetAmqpTransportSettings() };

[...]

PR #2935, opened 2022-11-04
Cancel amqp token refresher as a part of transport cleanup

abhipsaMisra:
The amqp sas token refresher runs as a background unmonitored task. We need to ensure that the task is cancelled as a part of transport layer cleanup.

Related: #2339 , maybe #2351

### DIFF
--- a/iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
+++ b/iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Amqp
 
             // Create a linked cancellation token source which can be signaled for cancellation by both the SDK
             // and the supplied cancellation token.
-            _refresherCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            _refresherCancellationTokenSource ??= CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
 
             if (refreshOn < DateTime.MaxValue)
             {
@@ -129,7 +129,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Amqp
             if (Logging.IsEnabled)
                 Logging.Enter(this, nameof(StopLoop));
 
-            _refresherCancellationTokenSource.Cancel();
+            _refresherCancellationTokenSource?.Cancel();
 
             if (Logging.IsEnabled)
                 Logging.Exit(this, nameof(StopLoop));
@@ -156,6 +156,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Amqp
                     {
                         StopLoop();
                         _refresherCancellationTokenSource?.Dispose();
+                        _refresherCancellationTokenSource = null;
+
                         _amqpIotCbsTokenProvider?.Dispose();
                     }
 
--- a/iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
+++ b/iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
@@ -51,9 +51,10 @@ namespace Microsoft.Azure.Devices.Client.Transport.Amqp
                     cancellationToken)
                 .ConfigureAwait(false);
 
-            // Create a linked cancellation token source which can be signaled for cancellation by both the SDK
-            // and the supplied cancellation token.
-            _refresherCancellationTokenSource ??= CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            // InitLoopAsync is not re-entrant, so we can initialize a new cancellation token source for each
+            // instance of AmqpAuthenticationRefresher initialized.
+            // This cancellation token source is disposed when the authentication refresher is disposed.
+            _refresherCancellationTokenSource = new CancellationTokenSource();
 
             if (refreshOn < DateTime.MaxValue)
             {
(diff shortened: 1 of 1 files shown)


patilsnr (approved):

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -120,7 +127,12 @@ private async Task RefreshLoopAsync(DateTime refreshesOn, CancellationToken canc
         public void StopLoop()
         {
             if (Logging.IsEnabled)
-                Logging.Info(this, nameof(StopLoop));
+                Logging.Enter(this, nameof(StopLoop));
+
+            _refresherCancellationTokenSource.Cancel();
```
drwill-ms, line 132: Protect against null?

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -50,9 +51,13 @@ public async Task InitLoopAsync(CancellationToken cancellationToken)
                     cancellationToken)
                 .ConfigureAwait(false);
 
+            // Create a linked cancellation token source which can be signaled for cancellation by both the SDK
+            // and the supplied cancellation token.
+            _refresherCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```
drwill-ms, line 56: Is this method re-entrant? If so, should we dispose before overwriting?

drwill-ms (approved):

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
abhipsaMisra: The current implementation has upstream caller protection against multiple threads calling it in parallel. However, this method is defined in an interface, so I'll update it to check for an existing source and reuse it if it exists. I can't think of a scenario where this method would be called by concurrently with different cancellation tokens, so us reusing the previous token source should be fine.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -50,9 +51,14 @@ public async Task InitLoopAsync(CancellationToken cancellationToken)
                     cancellationToken)
                 .ConfigureAwait(false);
 
+            // InitLoopAsync is not re-entrant, so we can initialize a new cancellation token source for each
+            // instance of AmqpAuthenticationRefresher initialized.
+            // This cancellation token source is disposed when the authentication refresher is disposed.
+            _refresherCancellationTokenSource = new CancellationTokenSource();
```
abhipsaMisra, line 57: the original cancellation token passed into this API depends on the cancellation token of the upstream call, i.e. the cancellation token passed into OpenAsync() etc. I'd like to have that decoupled since the token refresher should ideally run on a separate thread all the way until it is canceled.

patilsnr (approved):

patilsnr:
Optimistic sign off, would recommend waiting for another approval from someone more experienced.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -51,9 +51,10 @@ public async Task InitLoopAsync(CancellationToken cancellationToken)
                     cancellationToken)
                 .ConfigureAwait(false);
 
-            // Create a linked cancellation token source which can be signaled for cancellation by both the SDK
-            // and the supplied cancellation token.
-            _refresherCancellationTokenSource ??= CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            // InitLoopAsync is not re-entrant, so we can initialize a new cancellation token source for each
```
drwill-ms, line 54: Can we detect if it isn't and assert?

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -78,10 +86,13 @@ private async Task RefreshLoopAsync(DateTime refreshesOn, CancellationToken canc
             while (!cancellationToken.IsCancellationRequested)
             {
                 if (Logging.IsEnabled)
-                    Logging.Info(this, refreshesOn, $"Before {nameof(RefreshLoopAsync)}");
+                    Logging.Info(this, refreshesOn, $"{_amqpIotCbsTokenProvider} Before {nameof(RefreshLoopAsync)}");
```
drwill-ms, line 89:
```suggestion
                    Logging.Info(this, refreshesOn, $"{_amqpIotCbsTokenProvider} before {nameof(RefreshLoopAsync)}");
```

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -99,17 +110,20 @@ private async Task RefreshLoopAsync(DateTime refreshesOn, CancellationToken canc
...
+                    catch (Exception ex) when (ex is IotHubCommunicationException || ex is OperationCanceledException)
                     {
+                        // In case the token refresh is not successful either due to a communication exception or cancellation token cancellation
+                        // then log the exception and continue.
+                        // This task runs on an unmonitored thread so there is no point throwing these exceptions.
                         if (Logging.IsEnabled)
                         {
-                            Logging.Error(this, refreshesOn, $"Refresh token failed {ex}");
+                            Logging.Error(this, refreshesOn, $"{_amqpIotCbsTokenProvider} Refresh token failed {ex}");
                         }
```
drwill-ms, line 121: Braces can be removed.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
abhipsaMisra: Yeah. The instance hash was useful to me when debugging multiplexed devices, so I thought it would be useful to add it into the log.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -143,6 +162,9 @@ private void Dispose(bool disposing)
                     if (disposing)
                     {
                         StopLoop();
+                        _refresherCancellationTokenSource?.Dispose();
+                        _refresherCancellationTokenSource = null;
```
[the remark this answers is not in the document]
abhipsaMisra: I can remove this, sure

drwill-ms (approved):

PR #2938, opened 2022-11-07
Fixed M2M message (MQTT) failure after receiving Twin response or Direct method

tmahmood-microsoft:
Module-to-module message (MQTT) failed after receiving a twin response from EdgeHub. The issue was occurring due to mis-matched Ack for the M2M message after twin response. The Mqtt client was not processing the Ack for twin response properly that caused sending incorrect Ack for each M2M message after receiving twin response.

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -107,12 +107,11 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         private readonly string _deviceboundMessagePrefix;
         private readonly string _hostName;
         private readonly Func<IPAddress[], int, Task<IChannel>> _channelFactory;
-        private readonly Queue<string> _completionQueue;
+        private readonly ConcurrentQueue<string> _completionQueue;
         private readonly MqttIotHubAdapterFactory _mqttIotHubAdapterFactory;
         private readonly QualityOfService _qosSendPacketToService;
         private readonly QualityOfService _qosReceivePacketFromService;
         private readonly bool _retainMessagesAcrossSessions;
-        private readonly object _syncRoot = new object();
         private readonly RetryPolicy _closeRetryPolicy;
         private readonly ConcurrentQueue<Message> _messageQueue;
         private readonly TaskCompletionSource _connectCompletion = new TaskCompletionSource();
@@ -161,7 +160,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         {
             _mqttIotHubAdapterFactory = new MqttIotHubAdapterFactory(settings);
             _messageQueue = new ConcurrentQueue<Message>();
-            _completionQueue = new Queue<string>();
+            _completionQueue = new ConcurrentQueue<string>();
 
             _serverAddresses = null; // this will be resolved asynchronously in OpenAsync
             _hostName = iotHubConnectionString.HostName;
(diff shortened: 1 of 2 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -64,7 +64,7 @@ internal sealed class MqttTransportHandler : TransportHandler, IMqttIotHubEventH
         // The client first subscribes to "$iothub/twin/res/#", to receive the operation's responses.
         // It then sends an empty message to the topic "$iothub/twin/GET/?$rid={request id}, with a populated value for request Id.
         // The service then sends a response message containing the device twin data on topic "$iothub/twin/res/{status}/?$rid={request id}", using the same request Id as the request.
-
+
```
drwill-ms, line 67: extra whitespace can be removed

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -653,6 +657,26 @@ public async void OnMessageReceived(Message message)
             }
         }
 
+        private async Task CompleteIncomingMessageAsync(Message message)
+        {
+            try
+            {
+                if (_qosReceivePacketFromService == QualityOfService.AtLeastOnce)
+                {
+                    lock (_syncRoot)
```
drwill-ms, line 666: What if we made this a ConcurrentQueue, so we don't have to lock around it in all cases?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -584,6 +586,7 @@ private async Task HandleIncomingMessagesAsync()
             // This is a user-supplied callback that isn't required to be awaited by us. We can simply invoke it and continue.
             _ = _deviceMessageReceivedListener?.Invoke(message);
             await TaskHelpers.CompletedTask.ConfigureAwait(false);
+            await CompleteIncomingMessageAsync(message).ConfigureAwait(false);
```
abhipsaMisra, line 589: Since we're now awaiting a task inside this method, you can remove the `await TaskHelpers.CompletedTask.ConfigureAwait(false);` call. 
That was previously added since this async method expected a Task in return but we weren't awaiting any tasks in here.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -653,6 +657,26 @@ public async void OnMessageReceived(Message message)
...
+        {
+            try
+            {
+                if (_qosReceivePacketFromService == QualityOfService.AtLeastOnce)
+                {
+                    lock (_syncRoot)
+                    {
+                        _completionQueue.Enqueue(message.LockToken);
+                    }
+                    await CompleteAsync(_generationId + message.LockToken, CancellationToken.None).ConfigureAwait(false);
```
tmahmood-microsoft, line 670: Can we use `CancellationToken.None` here?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -546,6 +546,7 @@ private async Task HandleIncomingTwinPatchAsync(Message message)
                     string patch = reader.ReadToEnd();
                     TwinCollection props = JsonConvert.DeserializeObject<TwinCollection>(patch);
                     await Task.Run(() => _onDesiredStatePatchListener(props)).ConfigureAwait(false);
+                    await CompleteIncomingMessageAsync(message).ConfigureAwait(false);
```
drwill-ms, line 549: I think these could use code comments explaining why they are being completed right away.

drwill-ms (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra, line 666: Remind me - was there a reason why this is a queue and not a concurrent queue?

abhipsaMisra (approved):
I like this implementation.

PR #2939, opened 2022-11-08
Add more doc comments for handling user-supplied callbacks for cloud-to-device communication

brycewang-microsoft:
Also add internal exception handling for those user-supplied callbacks.

I don't feel confident about good places for putting these doc comments to share with users, so open to any suggestions.

### DIFF
--- a/iothub/device/src/IotHubBaseClient.cs
+++ b/iothub/device/src/IotHubBaseClient.cs
@@ -80,9 +80,8 @@ namespace Microsoft.Azure.Devices.Client
         /// The callback to be executed each time connection status change notification is received.
         /// </summary>
         /// <remarks>
-        /// This user-supplied callback is "fire-and-forget" and the SDK doesn't wait on it.
-        /// All of requests will be processed as they arrive. Exceptions thrown within the
-        /// callback will be caught and logged by the SDK internally.
+        /// All of requests will be processed as they arrive. If you put async code within this
+        /// callback, you'll need to handle exceptions that could originate in there.
         /// </remarks>
         /// <example>
         /// deviceClient.ConnectionStatusChangeCallback = OnConnectionStatusChanged;
@@ -198,12 +197,10 @@ namespace Microsoft.Azure.Devices.Client
         /// <remarks>
         /// Calling this API more than once will result in the callback set last overwriting any previously set callback.
         /// A method callback can be unset by setting <paramref name="messageCallback"/> to null.
-        /// </remarks>
-        /// <param name="messageCallback">
-        /// The callback to be invoked when a cloud-to-device message is received by the client.
         /// This user-supplied callback is awaited by the SDK. All of requests will be processed as they arrive.
         /// Exceptions thrown within the callback will be caught and logged by the SDK internally.
-        /// </param>
+        /// </remarks>
+        /// <param name="messageCallback">The callback to be invoked when a cloud-to-device message is received by the client.</param>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <exception cref="InvalidOperationException">Thrown if instance is not opened already.</exception>
         /// <exception cref="OperationCanceledException">Thrown when the operation has been canceled.</exception>
@@ -253,12 +250,10 @@ namespace Microsoft.Azure.Devices.Client
         /// <remarks>
         /// Calling this API more than once will result in the callback set last overwriting any previously set callback.
         /// A method callback can be unset by setting <paramref name="directMethodCallback"/> to null.
-        /// </remarks>
-        /// <param name="directMethodCallback">
-        /// The callback to be invoked when any method is invoked by the cloud service.
         /// This user-supplied callback is awaited by the SDK. All of requests will be processed as they arrive.
         /// Exceptions thrown within the callback will be caught and logged by the SDK internally.
-        /// </param>
+        /// </remarks>
+        /// <param name="directMethodCallback">The callback to be invoked when any method is invoked by the cloud service.</param>
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
         /// <exception cref="OperationCanceledException">Thrown when the operation has been canceled.</exception>
         public async Task SetDirectMethodCallbackAsync(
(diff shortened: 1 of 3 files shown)


iothub/device/src/IotHubBaseClient.cs
```diff
@@ -194,7 +199,12 @@ public async Task SendTelemetryBatchAsync(IEnumerable<TelemetryMessage> messages
         /// Calling this API more than once will result in the callback set last overwriting any previously set callback.
         /// A method callback can be unset by setting <paramref name="messageCallback"/> to null.
         /// </remarks>
-        /// <param name="messageCallback">The callback to be invoked when a cloud-to-device message is received by the client.</param>
+        /// <param name="messageCallback">
+        /// The callback to be invoked when a cloud-to-device message is received by the client.
+        /// This user-supplied callback is awaited by the SDK. All of requests will be processed as they arrive.
+        /// Exceptions will be thrown to the users and the SDK won't log them internally, so the user is responsible
```
drwill-ms, line 205: What does this mean: "Exceptions will be thrown to the users"?

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -194,7 +199,12 @@ public async Task SendTelemetryBatchAsync(IEnumerable<TelemetryMessage> messages
...
+        /// to catch and handle exceptions explicitly within their codes.
```
drwill-ms, line 206:
```suggestion
        /// to catch and handle exceptions explicitly within their code.
```

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -328,7 +342,12 @@ public async Task<long> UpdateReportedPropertiesAsync(ReportedProperties reporte
         /// This has the side-effect of subscribing to the PATCH topic on the service.
         ///  </para>
         /// </remarks>
-        /// <param name="callback">The callback to be invoked when a desired property update is received from the service.</param>
+        /// <param name="callback">
+        /// The callback to be invoked when a desired property update is received from the service.
+        /// This user-supplied callback is "fire-and-forget" and the SDK doesn't wait on it. All of requests will be processed as they arrive.
+        /// Exceptions will be thrown to the users and the SDK won't log them internally, so the user is responsible
```
drwill-ms, line 348: Same question here about the meaning of "Exceptions will be thrown to the users".

iothub/device/src/IotHubBaseClient.cs
brycewang-microsoft: we don't catch any exceptions in this method of subscribing incoming message callback `SetIncomingMessageCallbackAsync()`

brycewang-microsoft: we don't catch or handle any exceptions in the method `OnMessageReceivedAsync()`

iothub/device/src/IotHubBaseClient.cs
brycewang-microsoft: we don't catch or handle any exceptions in the method `OnDesiredStatePatchReceived()`

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -79,6 +79,11 @@ public abstract class IotHubBaseClient : IDisposable
         /// <summary>
         /// The callback to be executed each time connection status change notification is received.
         /// </summary>
+        /// <remarks>
+        /// This user-supplied callback is "fire-and-forget" and the SDK doesn't wait on it.
```
andyk-ms, line 83: Do we have reference or document to point out what are the exceptions the user can expect?
What about updating the example with 'try/catch' to explicitly remind the user about handling exceptions?

iothub/device/src/IotHubBaseClient.cs
drwill-ms: Okay, the original text sounds like we'll throw an exception that the user must catch. Is that the case? Because, if this is a callback, there's no way they can catch it. Exceptions won't be thrown **to** them, right?

drwill-ms: Also, thanks for documenting, but can we change the behavior to catch exceptions and then update the docs to indicate that? I'd rather fix the behavior than just document bad behavior.

brycewang-microsoft: Yeah, sure, we should be able to apply the same try-catch logic from `OnMethodCalledAsync()` into this one.

iothub/device/src/IotHubBaseClient.cs
drwill-ms: We call the user's callback, therefore we can't throw exceptions that would impact them here. However, if they throw an exception, we need to handle it so as to not crash the process. We're just fixing and documenting that behavior.

patilsnr (approved):

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -579,6 +609,13 @@ internal async Task<MessageAcknowledgement> OnMessageReceivedAsync(IncomingMessa
 
                 return MessageAcknowledgement.Abandon;
             }
+            catch (Exception ex)
+            {
+                if (Logging.IsEnabled)
+                    Logging.Error(this, $"User code threw exception: {ex}\nAbandoning message with message Id: {message.MessageId}.", nameof(OnMessageReceivedAsync));
```
drwill-ms, line 615:
```suggestion
                    Logging.Error(this, $"Abandoning message with message Id: {message.MessageId} because user code threw exception: {ex}.", nameof(OnMessageReceivedAsync));
```

iothub/device/src/IotHubBaseClient.cs
abhipsaMisra: This specific callback doesn't return anything (`Action<ConnectionStatusInfo>`), so instead of calling this fire-and-forget (which would mean that it returns a task but we don't await on it) why don't we call out that only if you put async code within this callback, you'll need to handle exceptions that could originate in there?

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -194,7 +199,11 @@ public async Task SendTelemetryBatchAsync(IEnumerable<TelemetryMessage> messages
         /// Calling this API more than once will result in the callback set last overwriting any previously set callback.
         /// A method callback can be unset by setting <paramref name="messageCallback"/> to null.
         /// </remarks>
-        /// <param name="messageCallback">The callback to be invoked when a cloud-to-device message is received by the client.</param>
+        /// <param name="messageCallback">
+        /// The callback to be invoked when a cloud-to-device message is received by the client.
+        /// This user-supplied callback is awaited by the SDK. All of requests will be processed as they arrive.
```
abhipsaMisra, line 204: What do you think about putting this explanation in the remarks section for this method definition? Having a whole lot of text in the param name field might be overwhelming to some people.

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -328,7 +341,11 @@ public async Task<long> UpdateReportedPropertiesAsync(ReportedProperties reporte
         /// This has the side-effect of subscribing to the PATCH topic on the service.
         ///  </para>
         /// </remarks>
-        /// <param name="callback">The callback to be invoked when a desired property update is received from the service.</param>
+        /// <param name="callback">
+        /// The callback to be invoked when a desired property update is received from the service.
+        /// This user-supplied callback is "fire-and-forget" and the SDK doesn't wait on it. All of requests will be processed as they arrive.
+        /// Exceptions thrown within the callback will be caught and logged by the SDK internally.
```
abhipsaMisra, line 347: If it is a fire-and-forget then the SDK won't be handling the exceptions, right? We'll need to ask users to handle exceptions within their callback implementation.

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -489,7 +511,15 @@ internal void OnDesiredStatePatchReceived(DesiredProperties patch)
...
                 Logging.Info(this, patch.GetSerializedString(), nameof(OnDesiredStatePatchReceived));
 
-            _ = _desiredPropertyUpdateCallback.Invoke(patch);
+            try
+            {
+                _ = _desiredPropertyUpdateCallback.Invoke(patch);
+            }
+            catch (Exception ex)
+            {
+                if (Logging.IsEnabled)
```
abhipsaMisra, line 520: We're not awaiting on this task, so this exception block wouldn't actually catch anything. The code above is simply queuing up a task that runs on an unmonitored thread.

PR #2942, opened 2022-11-08
Convert tests to Moq and remove Nsubstitute

ngastelum-ms:


iothub/device/tests/Pipeline/RetryDelegatingHandlerImplicitOpenTests.cs
```diff
@@ -21,26 +20,26 @@ public class RetryDelegatingHandlerImplicitOpenTests
...
+            nextHandlerMock.Setup(x => x.OpenAsync(It.IsAny<CancellationToken>())).Returns(() => Task.CompletedTask);
 
             var cancellationToken = new CancellationToken();
-            var sut = new RetryDelegatingHandler(contextMock, nextHandlerMock);
+            var sut = new RetryDelegatingHandler(contextMock, nextHandlerMock.Object);
             await sut.OpenAsync(cancellationToken).ConfigureAwait(false);
         }
 
         [TestMethod]
         public async Task ImplicitOpenSubjWasNotOpenSubjIsOpen()
```
drwill-ms, line 34: @brycewang-microsoft does this unit test even make sense in V2 with explicit open required?

iothub/device/tests/Pipeline/RetryDelegatingHandlerImplicitOpenTests.cs
```diff
@@ -21,26 +20,26 @@ public class RetryDelegatingHandlerImplicitOpenTests
...
-            nextHandlerMock.SendTelemetryAsync(Arg.Any<TelemetryMessage>(), Arg.Any<CancellationToken>()).Returns(t => Task.CompletedTask);
-            nextHandlerMock.SendTelemetryBatchAsync(Arg.Any<IEnumerable<TelemetryMessage>>(), Arg.Any<CancellationToken>()).Returns(t => Task.CompletedTask);
-            nextHandlerMock.WaitForTransportClosedAsync().Returns(Task.Delay(TimeSpan.FromSeconds(10)));
+            var nextHandlerMock = new Mock<IDelegatingHandler>();
+            nextHandlerMock.Setup(x => x.OpenAsync(It.IsAny<CancellationToken>())).Returns(() => Task.CompletedTask);
+            nextHandlerMock.Setup(x => x.SendTelemetryAsync(It.IsAny<TelemetryMessage>(), It.IsAny<CancellationToken>())).Returns(() => Task.CompletedTask);
+            nextHandlerMock.Setup(x => x.SendTelemetryBatchAsync(It.IsAny<IEnumerable<TelemetryMessage>>(), It.IsAny<CancellationToken>())).Returns(() => Task.CompletedTask);
+            nextHandlerMock.Setup(x => x.WaitForTransportClosedAsync()).Returns(() => Task.Delay(TimeSpan.FromSeconds(10)));
             var cancellationToken = new CancellationToken();
```
drwill-ms, line 43: Remove this and pass CancellationToken.None.

iothub/device/tests/Pipeline/RetryDelegatingHandlerImplicitOpenTests.cs
drwill-ms: I think the test was updated and the test name wasn't fixed to match.

drwill-ms (approved):

iothub/device/tests/Pipeline/RetryDelegatingHandlerImplicitOpenTests.cs
brycewang-microsoft: > `NoImplicitOpenByClientMethods`?
"No implicit open" seems not what this test is doing here as it calls `OpenAsync()`, `WaitForTransportClosedAsync` and then `OpenAsync()` again. 

I notice this [comment ](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/device/src/Pipeline/RetryDelegatingHandler.cs#L386) so this test might be for an unsupported behavior?

drwill-ms: I think that comment is out of date. I think @timtay-microsoft did the work to enable open/close/open.

In the test here, we call open, then we call 2 methods, and in the end ensure that openasync was called only once (not 3 times). That seems to validate that there is no implicit open.

PR #2947, opened 2022-11-09
Update doc comments, specifically looking at web docs generated

drwill-ms:
Looking at generated web docs at https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.client?view=azure-dotnet-preview

### DIFF
--- a/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
+++ b/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
@@ -26,23 +26,23 @@ namespace Microsoft.Azure.Devices.Client
         /// The authentication method that is used. It includes <see cref="ClientAuthenticationWithSharedAccessKeyRefresh"/>, <see cref="ClientAuthenticationWithSharedAccessSignature"/>,
         /// <see cref="ClientAuthenticationWithX509Certificate"/> or <see cref="EdgeModuleAuthenticationWithHsm"/>.
         /// </param>
-        /// <param name="iotHubHostName">The fully-qualified DNS host name of IoT hub.</param>
+        /// <param name="hostName">The fully-qualified DNS host name of IoT hub.</param>
         /// <param name="gatewayHostName">The fully-qualified DNS host name of the gateway (optional).</param>
         /// <returns>A new instance of the <c>IotHubConnectionCredentials</c> class with a populated connection string.</returns>
-        /// <exception cref="ArgumentNullException"><paramref name="iotHubHostName"/>, device Id or <paramref name="authenticationMethod"/> is null.</exception>
-        /// <exception cref="ArgumentException"><paramref name="iotHubHostName"/> or device Id are an empty string or consist only of white-space characters.</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="hostName"/>, device Id or <paramref name="authenticationMethod"/> is null.</exception>
+        /// <exception cref="ArgumentException"><paramref name="hostName"/> or device Id are an empty string or consist only of white-space characters.</exception>
         /// <exception cref="ArgumentException"><see cref="ClientAuthenticationWithX509Certificate.CertificateChain"/> is used over a protocol other than MQTT over TCP or AMQP over TCP.</exception>
         /// <exception cref="FormatException">Neither shared access key, shared access signature or X509 certificates were presented for authentication.</exception>
         /// <exception cref="FormatException">Either shared access key or shared access signature where presented together with X509 certificates for authentication.</exception>
-        /// <exception cref="IotHubClientException"><see cref="ClientAuthenticationWithX509Certificate.CertificateChain"/> could not be installed.</exception>
-        public IotHubConnectionCredentials(IAuthenticationMethod authenticationMethod, string iotHubHostName, string gatewayHostName = null)
+        /// <exception cref="IotHubClientException">When <see cref="ClientAuthenticationWithX509Certificate.CertificateChain"/> could not be installed.</exception>
+        public IotHubConnectionCredentials(IAuthenticationMethod authenticationMethod, string hostName, string gatewayHostName = null)
         {
             Argument.AssertNotNull(authenticationMethod, nameof(authenticationMethod));
-            Argument.AssertNotNullOrWhiteSpace(iotHubHostName, nameof(iotHubHostName));
+            Argument.AssertNotNullOrWhiteSpace(hostName, nameof(hostName));
 
-            IotHubHostName = iotHubHostName;
+            IotHubHostName = hostName;
             GatewayHostName = gatewayHostName;
-            HostName = gatewayHostName ?? iotHubHostName;
+            HostName = gatewayHostName ?? hostName;
 
             AuthenticationMethod = authenticationMethod;
             IotHubConnectionCredentials iotHubConnectionCredentials = this;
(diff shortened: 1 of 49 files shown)


iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -26,23 +26,23 @@ internal IotHubConnectionCredentials()
         /// The authentication method that is used. It includes <see cref="ClientAuthenticationWithSharedAccessKeyRefresh"/>, <see cref="ClientAuthenticationWithSharedAccessSignature"/>,
         /// <see cref="ClientAuthenticationWithX509Certificate"/> or <see cref="EdgeModuleAuthenticationWithHsm"/>.
         /// </param>
-        /// <param name="iotHubHostName">The fully-qualified DNS host name of IoT hub.</param>
+        /// <param name="hostName">The fully-qualified DNS host name of IoT hub.</param>
```
drwill-ms, line 29: We're throwing an exception here based on this name, if null or whitespace. It bubbles up to the client ctor, though, which uses `hostName` as the param name.

We really shouldn't be doing this, because the caller doesn't know about `iotHubHostName` as a parameter as the API surface they used doesn't include that parameter.

Aside from fixing this callstack, I can just make the parameters the same. I don't love it, but I don't have time to do the full fix.

iothub/device/src/ConnectionStatusInfo.cs
```diff
@@ -27,34 +27,23 @@ internal ConnectionStatusInfo(ConnectionStatus status, ConnectionStatusChangeRea
         /// <summary>
         /// The current connection status of the device or module.
         /// </summary>
-        /// <remark>
```
drwill-ms, line 30: What it defaults to is of no concern to a user.

iothub/device/src/Exceptions/IotHubClientErrorCode.cs
```diff
@@ -31,10 +31,13 @@ public enum IotHubClientErrorCode
         /// client connection was lost and regained while receiving the message but before you could reject/abandon/complete it.
         /// </summary>
         /// <remarks>
-        /// An abandoned message will be re-enqueued in the per-device/module queue, and the <see cref="IotHubDeviceClient"/> or <see cref="IotHubModuleClient"/> instance will receive it again.
+        /// An abandoned message will be re-enqueued in the per-device/module queue, and the
```
drwill-ms, line 34: All these changes are line length fixes

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -84,13 +84,15 @@ public abstract class IotHubBaseClient : IDisposable
         /// callback, you'll need to handle exceptions that could originate in there.
         /// </remarks>
         /// <example>
+        /// <code language="csharp">
```
drwill-ms, line 87: Requires a code tag to render the code properly, and a csharp language property to colorize it right.

iothub/device/src/Messaging/IncomingMessage.cs
```diff
@@ -26,19 +26,22 @@ protected internal IncomingMessage(byte[] payload)
         }
 
         /// <summary>
-        /// An identifier for the message used for request-reply patterns.
-        /// Format: A case-sensitive string ( up to 128 char long) of ASCII 7-bit alphanumeric chars
-        /// + {'-', ':', '/', '\', '.', '+', '%', '_', '#', '*', '?', '!', '(', ')', ',', '=', '@', ';', '$', '''}.
```
drwill-ms, line 31: The plus sign was rendering as a bullet.

iothub/device/src/TransportSettings/IotHubClientTransportSettings.cs
```diff
@@ -28,7 +28,7 @@ public abstract class IotHubClientTransportSettings
         /// <example>
         /// To set a proxy you must instantiate an instance of the <see cref="WebProxy"/> class--or any class that derives from <see cref="IWebProxy"/>.
         /// The snippet below shows a method that returns a device using a proxy that connects to localhost on port 8888.
-        /// <c>
```
drwill-ms, line 31: We learned that multi-line code must use the `<code>` tag.

provisioning/service/src/Twin/ProvisioningTwinState.cs
```diff
@@ -63,29 +63,29 @@ public class ProvisioningTwinState
         /// <remarks>
         /// This constructor creates an instance of the TwinState with the provided twin collection tags and desired properties.
         /// </remarks>
+        /// <param name="tags">The twin collection with the initial tags state. It can be null.</param>
+        /// <param name="desiredProperties">The twin collection with the initial desired properties. It can be null.</param>
         /// <example>
         /// When serialized, this class will looks like the following example:
-        /// <code>
+        /// <code language="csharp">
         /// {
```
schoims, line 71: language="json"?

iothub/device/src/ConnectionStatusInfo.cs
```diff
@@ -27,34 +27,23 @@ internal ConnectionStatusInfo(ConnectionStatus status, ConnectionStatusChangeRea
...
 
         /// <summary>
-        /// Timestamp in UTC when the last connection status was changed.
+        /// When the last connection status was changed, in UTC.
         /// </summary>
         public DateTimeOffset StatusLastChangedOnUtc { get; }
 
         /// <summary>
-        /// Recommended actions for users to take upon different ConnectionStatus and ConnectionStatusChangeReason.
+        /// Recommended actions for users to take upon different connectio status and change reasons.
```
schoims, line 43: connection status

schoims (approved):

iothub/device/src/ConnectionStatusInfo.cs
abhipsaMisra: Is this combination (Disconnected, Client_Close) the first status, reason that is returned by this class? If yes, then it might be worth calling that out somewhere so that users know to expect it as the first reported status.

iothub/device/src/IotHubClientOptions.cs
```diff
@@ -66,25 +67,24 @@ public IotHubClientOptions(IotHubClientTransportSettings transportSettings)
...
         /// <summary>
         /// Sets the retry policy used in the operation retries.
         /// </summary>
         /// <remarks>
-        /// Defaults to a nearly infinite exponential backoff. If set to null, will use <see cref="IotHubClientNoRetry"/> to perform no retries.
-        /// Can be set to any of the built in retry policies such as <see cref="IotHubClientFixedDelayRetryPolicy"/> or <see cref="IotHubClientIncrementalDelayRetryPolicy"/> 
+        /// Defaults to a nearly infinite exponential backoff.
+        /// If set to null, will use <see cref="IotHubClientNoRetry"/> to perform no retries.
+        /// It can be set to any of the built-in retry policies such as <see cref="IotHubClientExponentialBackoffRetryPolicy"/>,
+        /// <see cref="IotHubClientFixedDelayRetryPolicy"/>, or <see cref="IotHubClientIncrementalDelayRetryPolicy"/>,
```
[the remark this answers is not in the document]
drwill-ms: I don't know about quickly going out of date, but just like `IAuthenticationMethod`, I think we need to tell someone what types are available to them to use, otherwise they have to hunt (not as easy when you aren't in azureiot.sln) or find a sample.

abhipsaMisra (approved):

iothub/device/src/ConnectionStatusInfo.cs
drwill-ms: Fair enough. I see your point.

PR #2948, opened 2022-11-09
Moq test conversion

ngastelum-ms:


iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -267,19 +267,17 @@ await deviceClient
                 .ConfigureAwait(false);
 
             // assert
-            await innerHandler
-                .Received()
-                .DisableMethodsAsync(Arg.Any<CancellationToken>())
-                .ConfigureAwait(false);
+            innerHandler
```
drwill-ms, line 270: I'm not sure what Verify does.

drwill-ms: Shouldn't there be a `Times` parameter or something?

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -1141,15 +1139,15 @@ public async Task IotHubDeviceClient_SendTelemetryAsync_Cancelled_ThrowsOperatio
...
             // We will setup the main handler which can be either MQTT or AMQP or HTTP handler to throw
             // a cancellation token expiry exception (OperationCancelledException) to ensure that we mimic when a token expires.
             mainProtocolHandler
-                .When(x => x.SendTelemetryAsync(Arg.Any<TelemetryMessage>(), Arg.Any<CancellationToken>()))
-                .Do(x => { throw new OperationCanceledException(); });
-
-            ErrorDelegatingHandler errorHandler = new ErrorDelegatingHandler(null, mainProtocolHandler);
+                .When(x => x.SendTelemetryAsync(It.IsAny<TelemetryMessage>(), It.IsAny<CancellationToken>()))
+                .Invoking(x => { throw new OperationCanceledException(); });
+
```
drwill-ms, line 1149: extra whitespace here

iothub/device/tests/Pipeline/ErrorDelegatingHandlerTests.cs
```diff
@@ -131,97 +130,78 @@ public async Task ErrorHandler_NonTransientErrorOccured_ChannelIsRecreated()
...
             //initial OpenAsync to emulate Gatekeeper behavior
             var cancellationToken = new CancellationToken();
             await sut.OpenAsync(cancellationToken).ConfigureAwait(false);
 
             //set initial operation result that throws
 
+            var message = new TelemetryMessage(new byte[0]);
             bool[] setup = { false };
-            mockSetup(innerHandler).Returns(ci =>
+            innerHandler.Setup(x => x.SendTelemetryAsync(message, It.IsAny<CancellationToken>())).Returns(() =>
```
drwill-ms, line 165:
```suggestion
            innerHandler
                .Setup(x => x.SendTelemetryAsync(message, It.IsAny<CancellationToken>()))
                .Returns(() =>
```

iothub/device/tests/IotHubDeviceClientTests.cs
ngastelum-ms: I think Verify defaults to at least once if no Times parameter is passed in

drwill-ms: I'd rather the code be explicit.

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -1141,15 +1182,15 @@ public async Task IotHubDeviceClient_SendTelemetryAsync_Cancelled_ThrowsOperatio
...
 
-            var mainProtocolHandler = Substitute.For<IDelegatingHandler>();
+            var mainProtocolHandler = new Mock<IDelegatingHandler>();
 
             // We will setup the main handler which can be either MQTT or AMQP or HTTP handler to throw
             // a cancellation token expiry exception (OperationCancelledException) to ensure that we mimic when a token expires.
             mainProtocolHandler
-                .When(x => x.SendTelemetryAsync(Arg.Any<TelemetryMessage>(), Arg.Any<CancellationToken>()))
-                .Do(x => { throw new OperationCanceledException(); });
+                .Setup(x => x.SendTelemetryAsync(It.IsAny<TelemetryMessage>(), It.IsAny<CancellationToken>()))
```
ngastelum-ms, line 1190: Not super confident that this change is correct

drwill-ms: Why `Invoking` and not `Returns`? What's the difference?

drwill-ms: https://nsubstitute.github.io/help/callbacks/

drwill-ms: I wonder if this works the same if you use `Returns` instead.

`SendTelemetryAsync` returns something (a Task), but more importantly, we're just throwing an exception here. Seems that the callback/no-return-value-method doesn't really apply here.

PR #2953, opened 2022-11-10
Ignoring E2E tests that fail with timeout

tmahmood-microsoft:


e2e/test/iothub/CombinedClientOperationsPoolAmqpTests.cs
```diff
@@ -48,6 +48,7 @@ public async Task DeviceCombinedClientOperations_SingleConnection_AmqpWs()
                 .ConfigureAwait(false);
         }
 
+        [Ignore]
```
abhipsaMisra, line 51: You'll want to add a comment calling out why this test was ignored.

tmahmood-microsoft: added

patilsnr (approved):

PR #2955, opened 2022-11-14
refactor(prov-client): Make CRL inspection configurable

abhipsaMisra:
Aside from the provisioning clients, all of our clients use the CRL property in the shared package to specify if the certificate revocation list should be inspected.
The behavior for provisioning clients today is:
Http -> does not set either CRL check or remote certificate validation callback
Amqp -> does not set the CRL check but allows the user to specify a server certificate validation callback where any checks can be overridden
Mqtt -> hardcodes the CRL check to be true for X509 auth based TCP connections, hardcodes the CRL check to be false for symmetric key based TCP connections. CRL check isn't set for websocket connections.

This PR will remove the hardcoded checks for mqtt connections over tcp, but will instead rely on the values configured in the shared package.
This will introduce a breaking change for X509 auth based TCP connections, since the check will now be turned off by default. Checking in with the DPS team to see if there is any concern with this change.

### DIFF
--- a/provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
+++ b/provisioning/transport/mqtt/src/ProvisioningTransportHandlerMqtt.cs
@@ -207,7 +207,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
 
             var tlsSettings = new ClientTlsSettings(
                 TlsVersions.Instance.Preferred,
-                true,
+                TlsVersions.Instance.CertificateRevocationCheck,
                 new List<X509Certificate> { clientCertificate },
                 message.GlobalDeviceEndpoint);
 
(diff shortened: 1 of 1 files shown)


timtay-microsoft:
Making a security breaking change seems a bit unsafe here. Is there a compelling reason to do this in main? Could it wait until v2?

abhipsaMisra:
Certificate revocation lists are quickly becoming a thing of the past since CRLs can grow to be very large and these checks can quickly become resource intensive. As an alternative, most services are now moving to OCSP Stapling. Keeping this in mind, it is a good idea for us to not make certificate revocation list checks mandatory, but instead to allow this to be configurable by the user.

OCSP Stapling - https://knowledge.digicert.com/quovadis/ssl-certificates/ssl-general-topics/what-is-ocsp-stapling.html

abhipsaMisra:
CertificateRevocationCheck on TlsVersions current defaults to false. In order for us to make this check configurable for the provisioning client without breaking the behavior, we'd need to introduce another property on our transport settings. Even though this change would be useful to users, we'll hold off on making this change until we have a request for this.
In the meanwhile, we'll make this configurable for all our clients in the preview package: #3022 .

PR #2959, opened 2022-11-14
Add back DisposeAsync support to device clients

drwill-ms:
Users don't know to call CloseAsync and that can cause them problems? How to solve?

Today, I floated the idea of throwing an exception in Dispose() if close wasn't called. I also said I'd check with the Azure SDK team for their guidance.

Krzysztof suggested using IAsyncDisposable and reminded me we can bring in this type to a library that targets .netstandard2.0 by using a NuGet reference for the type we need. He also pointed me to the track2 ServiceBus library that does this. I've copied their choices (only implement IAsyncDisposable and not IDisposable, version number of the library containing this type).

Thoughts?

If we like this, we can also implement it in IoT hub service client, if we wish.


iothub/device/tests/ClientSettings/TransportSettingsTests.cs
```diff
@@ -125,20 +126,20 @@ public void ConnectionPoolSettings_PoolingOff()
 
         [TestMethod]
         [ExpectedException(typeof(ArgumentException))]
-        public void IotHubDeviceClient_NullX509Certificate()
+        public async Task IotHubDeviceClient_NullX509Certificate()
```
patilsnr, line 129: Possible noob question, but do async test methods need to have the word async in them? Or is that just for normal SDK class methods?

iothub/device/tests/ClientSettings/TransportSettingsTests.cs
```diff
@@ -125,20 +126,20 @@ public void ConnectionPoolSettings_PoolingOff()
...
         {
             // arrange
             const string hostName = "acme.azure-devices.net";
             var authMethod = new ClientAuthenticationWithX509Certificate(null, "device1");
             var options = new IotHubClientOptions(new IotHubClientAmqpSettings { PrefetchCount = 100 });
 
             // act
-            using var dc = new IotHubDeviceClient(hostName, authMethod, options);
+            await using var dc = new IotHubDeviceClient(hostName, authMethod, options);
```
patilsnr, line 137: Consider using `deviceClient` for consistency and readability across the SDK

iothub/device/tests/ClientSettings/TransportSettingsTests.cs
```diff
@@ -148,7 +149,7 @@ public void IotHubDeviceClient_NullX509CertificateChain()
...
```
patilsnr, line 152: Same here

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -119,25 +119,25 @@ public void DeviceAuthenticationWithX509Certificate_ChainCertsMqtttTcp_DoesNotTh
             var options = new IotHubClientOptions(new IotHubClientMqttSettings(IotHubClientTransportProtocol.Tcp));
 
             // act
-            using var dc = new IotHubDeviceClient(hostName, authMethod, options);
+            await using var dc = new IotHubDeviceClient(hostName, authMethod, options);
```
patilsnr, line 122: Another case here -- we use `dc` and `deviceClient` as variable names in the same file to mean the same type. It would be cleaner if we had a consistent name, imho

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -1077,7 +1077,7 @@ public void IotHubDeviceClient_CreateWithConnectionString_SasTokenTimeToLiveRene
 
             // act
             DateTime startTime = DateTime.UtcNow;
-            using IotHubDeviceClient deviceClient = new IotHubDeviceClient(FakeHostName, auth, options);
+            await using IotHubDeviceClient deviceClient = new IotHubDeviceClient(FakeHostName, auth, options);
```
patilsnr, line 1080: possible use of `var`?

iothub/device/tests/IotHubModuleClientTests.cs
```diff
@@ -50,34 +51,34 @@ public class IotHubModuleClientTests
 
         [TestMethod]
         [ExpectedException(typeof(ArgumentNullException))]
-        public void ModuleClient_CreateFromConnectionString_NullConnectionStringThrows()
+        public async Task ModuleClient_CreateFromConnectionString_NullConnectionStringThrows()
         {
-            using var mc = new IotHubModuleClient(null);
+            await using var mc = new IotHubModuleClient(null);
```
patilsnr, line 56: `moduleClient` instead of `mc` would be better imho.

patilsnr (approved):
Left some minorish comments. Overall I think this is a neat change!

SDK v2 migration guide.md
```diff
@@ -138,6 +138,11 @@ Find a client you currently use below, read the table of API name changes and us
   - The supported workaround is to provide a client web socket instance in the client options.
 - The authentication classes for devices and modules have been consolidated.
 - Reduced access levels to classes and methods that were never intended to be public where possible.
+- The device and module clients now only support `IAsyncDisposable`, which will ensure `CloseAsync()` is called before disposing.
+  - The syntax for this might be quite new for some and feel awkward. You can choose from the following options:
+    1. Manually call `await client.DisposeAsync();`.
+    1. Initialize the client with the **await** keyword: `await using client = new IotHubDeviceClient(...);`. The client will be disposed when it goes out of scope.
```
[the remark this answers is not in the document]
drwill-ms: In markdown, you can label each numbered bullet with 1 and it will autoincrement for you.

<https://www.markdownguide.org/basic-syntax/#ordered-lists>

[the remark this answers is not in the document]
drwill-ms: Nope, with IAsyncDisposable, you use both `await using` as a prefix to your disposable type.

e2e/test/iothub/DeviceTokenRefreshE2ETests.cs
```diff
@@ -144,7 +144,7 @@ public async Task IotHubDeviceClient_CreateFromConnectionString_TokenIsRefreshed
 
             var options = new IotHubClientOptions(new IotHubClientMqttSettings());
 
-            using IotHubDeviceClient deviceClient = new IotHubDeviceClient(testDevice.IotHubHostName, auth, options);
+            await using IotHubDeviceClient deviceClient = new IotHubDeviceClient(testDevice.IotHubHostName, auth, options);
```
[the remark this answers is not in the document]
drwill-ms: I'm all for it, but realize when someone changes a ton of files, they aren't necessarily taking on the task of fixing all the code in all those files. This is a great comment for new code, though.

iothub/device/tests/ClientSettings/TransportSettingsTests.cs
drwill-ms: If the method awaits other calls, it must have the keyword `async` in it, so yes.

SDK v2 migration guide.md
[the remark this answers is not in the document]
patilsnr: Ah, I misunderstood the second BP. Thanks!

patilsnr (approved):

iothub/device/tests/IotHubModuleClientTests.cs
```diff
@@ -86,14 +87,14 @@ public void ModuleClient_CreateFromConnectionStringWithClientOptions_DoesNotThro
...
-            using var moduleClient = new IotHubModuleClient(FakeConnectionString, clientOptions);
+            await using var moduleClient = new IotHubModuleClient(FakeConnectionString, clientOptions);
         }
 
         [TestMethod]
         public async Task ModuleClient_SetReceiveCallbackAsync_SetCallback_Mqtt()
         {
             var options = new IotHubClientOptions(new IotHubClientMqttSettings());
-            using var moduleClient = new IotHubModuleClient(FakeConnectionString, options);
+            await using var moduleClient = new IotHubModuleClient(FakeConnectionString, options);
```
andyk-ms, line 97: Not an issue with this directly.
Looking at all the existing tests and samples, it appears they are short-running and the impact of this change maybe hard to observe within the scope of the test or the sample, right?

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -401,14 +401,10 @@ public async Task CloseAsync(CancellationToken cancellationToken = default)
...
-        /// Releases the unmanaged resources used by the client and optionally disposes of the managed resources.
-        /// </summary>
-        /// <remarks>
-        /// The method <see cref="CloseAsync(CancellationToken)"/> should be called before disposing.
-        /// </remarks>
-        public void Dispose()
+        /// <inheritdoc/>
+        public async ValueTask DisposeAsync()
         {
+            await CloseAsync(CancellationToken.None).ConfigureAwait(false);
```
abhipsaMisra, line 407: What if we encounter an exception as a part of the CloseAsync() call. Should we wrap this with a try block and add the Dipose() call in the finally block?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -585,7 +586,7 @@ private async Task HandleDisconnectAsync()
 
                     await base.OpenAsync(cancellationToken).ConfigureAwait(false);
 
-                    var tasks = new List<Task>(4);
+                    var tasks = new List<Task>(3);
```
abhipsaMisra, line 589: nit - q: is there a reason why we need the size to be defined in advance?

abhipsaMisra (approved):

iothub/device/tests/IotHubModuleClientTests.cs
drwill-ms: There will be no real impact to this change. It'll still be closed and disposed, if it was before. If it wasn't opened, close will do nothing. If was opened before but not closed, this will close it, which we should have been doing anyway.

It won't make the test run measurably longer.

iothub/device/src/IotHubBaseClient.cs
drwill-ms: Good question. I'm not sure that's what I'd do, but it is an interesting idea.

Anyone else want to weigh in?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: So it is pre-allocated for the size we know it needs to be. It just reduces memory reallocations or overallocations. Do we need it? No. Is it better for memory usage? Yes.

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -1196,63 +1165,14 @@ public async Task IotHubDeviceClient_SendTelemetryAsync_WithoutExplicitOpenAsync
...
-                await deviceClient.SendTelemetryAsync(new TelemetryMessage());
-                await deviceClient.OpenAsync();
-            };
-
-            // assert
-            await act.Should().ThrowAsync<InvalidOperationException>();
-        }
-
-        [TestMethod]
-        public void IotHubDeviceClient_SendTelemetryAsync_AfterExplicitOpenAsync_DoesNotThrow()
```
drwill-ms, line 1217: This wasn't an actual unit test (it tries to connect and fails), and we aren't even observing/awaiting the action. I moved it to E2E.

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -1196,63 +1165,14 @@ public async Task IotHubDeviceClient_SendTelemetryAsync_WithoutExplicitOpenAsync
             await act.Should().ThrowAsync<InvalidOperationException>();
         }
 
-        [TestMethod]
-        public async Task IotHubDeviceClient_SendTelemetryAsync_BeforeExplicitOpenAsync_ThrowsInvalidOperationException()
```
drwill-ms, line 1200: This test is not different than the previous one, because OpenAsync will never get called due to the throw from SendTelemetryAsync.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -442,13 +440,17 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
...
+            {
+                // Already closed so gracefully exit, instead of throw.
+                return;
+            }
 
             await _handlerSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
 
             try
             {
-                if (!_wasOpenCalled)
```
drwill-ms, line 451: We were tracking whether open was called, only to customize the error message, which seemed overly complicated (Abhipsa agreed), so we got rid of the notion.

drwill-ms: See old line 477

iothub/device/tests/Microsoft.Azure.Devices.Client.Tests.csproj
```diff
@@ -35,7 +30,6 @@
 
   <!--Conditional package dependencies-->
   <ItemGroup Condition=" '$(TargetFramework)' == 'net472' ">
-    <Reference Include="System.Web" />
```
drwill-ms, line 38: Removing this didn't seem to alter the build.

iothub/device/src/IotHubBaseClient.cs
drwill-ms: The SB client does what you suggest: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs#L85

andyk-ms: My initial thought is yes, we should have try/catch. But then I looked second round, this is on DisposeAsync() when we try to CloseAsync(). If there is actual exception that we can handle it differently, we should have try/catch. But it seems if catching an exception, we will ignore it and move on, right?

drwill-ms: Yes, it would silently fail. That's why I didn't originally like the idea, but that is what the SB client does.

drwill-ms: Oh, I take that back. They catch it only to log it, but then rethrow.

abhipsaMisra: I like the idea of us catching the exception, logging it and throwing it back to the user. Seems fair to me that since "close" fail, the resources also didn't get disposed.

patilsnr (approved):

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -401,16 +401,28 @@ public async Task CloseAsync(CancellationToken cancellationToken = default)
...
+        public async ValueTask DisposeAsync()
         {
-            Dispose(true);
-            GC.SuppressFinalize(this);
+            if (Logging.IsEnabled)
+                Logging.Enter(this, "Closing and disposing", nameof(DisposeAsync));
+
+            try
+            {
+                await CloseAsync(CancellationToken.None).ConfigureAwait(false);
```
timtay-microsoft, line 412: This seems like an appropriate time to pass in a cancellation token that the SDK creates in order to avoid any sort of timeout error causing a deadlock. Maybe a minute at most?

timtay-microsoft: This pattern seems fine to me otherwise, though. (Signed off)

PR #2960, opened 2022-11-14
Fix certificate validation E2E tests for the hub and provisioning clients

brycewang-microsoft:
Also fix issue that we didn't pass in the user-supplied `RetryPolicy` value into `IotHubClientOptions` instance.


iothub/device/src/IotHubClientOptions.cs
```diff
@@ -112,6 +112,7 @@ internal IotHubClientOptions Clone()
                 ModelId = ModelId,
                 SdkAssignsMessageId = SdkAssignsMessageId,
                 AdditionalUserAgentInfo = AdditionalUserAgentInfo,
+                RetryPolicy = RetryPolicy,
```
brycewang-microsoft, line 115: Previously we didn't pass in the user-supplied `RetryPolicy` here, which caused the hub device/module client can only use the default retry policy rather than setting with their own choices.

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
```diff
@@ -34,73 +37,113 @@ public static void TestClassSetup(TestContext _)
...
-            using var provisioningServiceClient = new ProvisioningServiceClient(
-                TestConfiguration.Provisioning.ConnectionStringInvalidServiceCertificate);
+            // arrange
+            using var provisioningServiceClient =
+                new ProvisioningServiceClient(
+                    TestConfiguration.Provisioning.ConnectionStringInvalidServiceCertificate,
+                    new ProvisioningServiceClientOptions()
+                    {
+                        RetryPolicy = new ProvisioningServiceNoRetry(),
+                    });
```
drwill-ms, line 47:
```suggestion
            using var provisioningServiceClient = new ProvisioningServiceClient(
                TestConfiguration.Provisioning.ConnectionStringInvalidServiceCertificate,
                new ProvisioningServiceClientOptions
                {
                    RetryPolicy = new ProvisioningServiceNoRetry(),
                });
```

drwill-ms: No reason to make newlines quite so aggressively. Plenty of line length remaining for the initial call to new.

e2e/test/provisioning/ProvisioningCertificateValidationE2ETest.cs
```diff
@@ -34,73 +37,113 @@ public static void TestClassSetup(TestContext _)
...
         public async Task ProvisioningDeviceClient_RegisterAsyncInvalidServiceCertificateAmqpTcp_Fails()
         {
-            var clientOptions = new ProvisioningClientOptions(new ProvisioningClientAmqpSettings(ProvisioningClientTransportProtocol.Tcp));
-            Func<Task> act = async () => await TestInvalidServiceCertificate(clientOptions);
+            // arrange
+            var clientOptions =
+                new ProvisioningClientOptions(new ProvisioningClientAmqpSettings(ProvisioningClientTransportProtocol.Tcp))
+                {
+                    RetryPolicy = new ProvisioningClientNoRetry(),
+                };
```
drwill-ms, line 70: Same here.

```suggestion
            var clientOptions = new ProvisioningClientOptions(new ProvisioningClientAmqpSettings(ProvisioningClientTransportProtocol.Tcp))
            {
                RetryPolicy = new ProvisioningClientNoRetry(),
            };
```

iothub/device/src/IotHubClientOptions.cs
drwill-ms: Our test for this may need to use recursion to find properties that weren't cloned, otherwise we rely only on PRs to validate.

drwill-ms (approved):

patilsnr (approved):
Optimistic sign off

e2e/test/iothub/service/IoTHubCertificateValidationE2ETest.cs
```diff
@@ -18,42 +21,62 @@ public class IotHubCertificateValidationE2ETest : E2EMsTestBase
...
-#endif
+
+            // act
+            Func<Task> act = async () => await sc.Query.CreateAsync<ClientTwin>("select * from devices").ConfigureAwait(false);
+
+            // assert
+            var error = await act.Should().ThrowAsync<IotHubServiceException>();
+            error.And.StatusCode.Should().Be(System.Net.HttpStatusCode.RequestTimeout);
+            error.And.ErrorCode.Should().Be(IotHubServiceErrorCode.Unknown);
+            error.And.IsTransient.Should().BeTrue();
```
andyk-ms, line 34: Should this be 'BeFals()'?
This is not a transient issue when the cert is invalid, right?

brycewang-microsoft: This is sortof tricky as our invalid cert testing would throw out `HttpRequestException`, and the cause of such an exception could be transient or intransient, in different cases. Also we have similar problems when the inner exception is a `SocketException`, `WebsocketException`, or `MqttCommunicationTimedOutException`. 

I will bring this question to our standup for more insights and discussions.

PR #2961, opened 2022-11-15
Add samples to v2 pipeline

ngastelum-ms:


iothub/service/samples/how to guides/AutomaticDeviceManagementSample/AutomaticDeviceManagementSample.cs
```diff
@@ -83,6 +83,7 @@ private static void CreateDeviceContent(Configuration configuration, string conf
 
         private static void CreateMetricsAndTargetCondition(Configuration configuration)
         {
+            configuration.Metrics = new ConfigurationMetrics();
```
ngastelum-ms, line 86: Added to prevent NullReferenceException

PR #2966, opened 2022-11-16
Microsoft Azure IoT Hub SDK for .NET Release 2022-11-16

tmahmood-microsoft:
## Microsoft.Azure.Devices.Client 1.41.3

- Fixed M2M message (MQTT) failure after receiving Twin response or Direct method [#2938](https://github.com/Azure/azure-iot-sdk-csharp/pull/2938)
- Cancel amqp token refresher as a part of transport cleanup [#2935](https://github.com/Azure/azure-iot-sdk-csharp/pull/2935)

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -37,7 +37,7 @@
     <Description>Device SDK for Azure IoT Hub</Description>
     <PackageIcon>nugetIcon.png</PackageIcon>
     <PackageProjectUrl>https://github.com/Azure/azure-iot-sdk-csharp</PackageProjectUrl>
-    <Copyright>Â© Microsoft Corporation. All rights reserved.</Copyright>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageTags>IoT Microsoft Azure IoTHub Device Client .NET AMQP MQTT HTTP</PackageTags>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


versions.csv
```diff
@@ -1,5 +1,5 @@
 AssemblyPath, Version
-iothub\device\src\Microsoft.Azure.Devices.Client.csproj, 1.41.2
```
abhipsaMisra, line 2: The nuget push step would have failed in this attempt, right? (attempting to push bits with a version number that already exists)

abhipsaMisra (approved):

versions.csv
tmahmood-microsoft: I pushed updated version for Microsoft.Azure.Devices instead of Devices.Client by mistake and it worked as expected

abhipsaMisra: We might want to unlist it from nuget if Microsoft.Azure.Devices was pushed out without any code changes. Or else, we might want to add a note to our release notes that this version was pushed out but had no changes. This is just to help with user confusion when they look at the package but don't see any associated release notes.

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -37,7 +37,7 @@
     <Description>Device SDK for Azure IoT Hub</Description>
     <PackageIcon>nugetIcon.png</PackageIcon>
     <PackageProjectUrl>https://github.com/Azure/azure-iot-sdk-csharp</PackageProjectUrl>
-    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <Copyright>Â© Microsoft Corporation. All rights reserved.</Copyright>
```
drwill-ms, line 40: Fix

drwill-ms, line 40:
```suggestion
    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
```

drwill-ms (approved):

Issue #2958, opened 2022-11-14
[Bug Report] - ulong values in object cause UpdateReportedPropertiesAsync to hang


tank104:
# Context

- Windows 11 Professional x64,
- .NET 6 Framework 
- SDK "previews/v2" branch (latest)
- Device is Laptop

if the json contains a ulong value UpdateReportedPropertiesAsync hangs, and doesn't report any errors that I can see (tried turning on Trace too).  Setting the ulong to 0 works.

## Code sample exhibiting the issue
```
   var _deviceClient = new IotHubDeviceClient(ConnecitonString);
    var deviceInformationProperties = new Dictionary<string, object>
    {
      { "serialNumber", 5283920058631410000  }
    };
    var reportedProperties = new ReportedProperties
    {
      { "test", properties }
    };
    await _deviceClient.UpdateReportedPropertiesAsync(reportedProperties);

```

## Console log of the issue
Log attached (renamed from ETL to LOG so I can attach)

[iot_000001.log](https://github.com/Azure/azure-iot-sdk-csharp/files/10007403/iot_000001.log)

abhipsaMisra:
Thanks for bringing this to our attention.
It looks like the SDK is incorrectly handling this "bad request" and is attempting to retry the operation, which is why the application looks like it is hanging. We'll put a fix in for this.

For reference, twin property integer values need to be between -4503599627370496 and 4503599627370495.
Ref: https://learn.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-device-twins#tags-and-properties-format

The actual error message: 
{"errorCode":400004,"trackingId":"ca254def40b049bd8fd468483787a8f4-G:0-TimeStamp:11/14/2022 22:52:27","message":"Argument 'serialNumber' value must be between '-4503599627370496' and '4503599627370495'. The provided value was '5283920058631410000'.","timestampUtc":"2022-11-14T22:52:27.7262459Z"}

tank104:
Thanks @abhipsaMisra - error would be perfect.  For now I am converting those to string.

abhipsaMisra:
This has been released in https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2023-02-09-v2preview

PR #2967, opened 2022-11-16
Parse out service returned error message

abhipsaMisra:
Our transport layer currently isn't inspecting and parsing out the service returned error message over AMQP and MQTT. Instead, the entire error "message" block containing the error code, description, tracking Id etc. is put into IotHubClientException.Message. This leaves the parsing responsibility onto the customer.
This PR proposes a change where we inspect the received error message and parse it out.

Another related issue is that as a result of us not parsing out the error message, we currently mark all unsuccessful MQTT twin operations as retryable. This results in the actual error being hidden from the user and the SDK unnecessarily retying non-retryable error scenarios.

Related: https://github.com/Azure/azure-iot-sdk-csharp/issues/2958

### DIFF
--- a/iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
+++ b/iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
@@ -54,7 +54,7 @@ namespace Microsoft.Azure.Devices.Client.HsmAuthentication.Transport
 
                 if (length == 0)
                 {
-                    throw new IotHubClientException("Unexpected end of stream.", IotHubClientErrorCode.NetworkErrors);
+                    throw new IotHubClientException("Unexpected end of stream.", errorCode: IotHubClientErrorCode.NetworkErrors);
                 }
 
                 if (crFound && (char)buffer[position] == LF)
--- a/iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
+++ b/iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
@@ -54,10 +54,9 @@ namespace Microsoft.Azure.Devices.Client.HsmAuthentication.Transport
 
                 if (length == 0)
                 {
-                    throw new IotHubClientException("Unexpected end of stream.")
-                    {
-                        ErrorCode = IotHubClientErrorCode.NetworkErrors,
-                    };
+                    throw new IotHubClientException(
+                        "Unexpected end of stream.",
+                        IotHubClientErrorCode.NetworkErrors);
                 }
 
                 if (crFound && (char)buffer[position] == LF)
--- a/iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
+++ b/iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
@@ -54,7 +54,10 @@ namespace Microsoft.Azure.Devices.Client.HsmAuthentication.Transport
 
                 if (length == 0)
                 {
-                    throw new IotHubClientException("Unexpected end of stream.", errorCode: IotHubClientErrorCode.NetworkErrors);
+                    throw new IotHubClientException("Unexpected end of stream.")
+                    {
+                        ErrorCode = IotHubClientErrorCode.NetworkErrors,
+                    };
                 }
 
                 if (crFound && (char)buffer[position] == LF)
--- a/iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
+++ b/iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
@@ -54,9 +54,7 @@ namespace Microsoft.Azure.Devices.Client.HsmAuthentication.Transport
 
                 if (length == 0)
                 {
-                    throw new IotHubClientException(
-                        "Unexpected end of stream.",
-                        IotHubClientErrorCode.NetworkErrors);
+                    throw new IotHubClientException("Unexpected end of stream.", IotHubClientErrorCode.NetworkErrors);
                 }
 
                 if (crFound && (char)buffer[position] == LF)
--- a/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
+++ b/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
@@ -386,7 +386,10 @@ namespace Microsoft.Azure.Devices.Client
                         if (Logging.IsEnabled)
                             Logging.Error(null, $"{nameof(CertificateInstaller)} failed to read or write to cert store due to: {ex}");
 
-                        throw new IotHubClientException($"Failed to provide certificates in the chain - {ex.Message}", errorCode: IotHubClientErrorCode.Unauthorized, innerException: ex);
+                        throw new IotHubClientException($"Failed to provide certificates in the chain - {ex.Message}", ex)
+                        {
+                            ErrorCode = IotHubClientErrorCode.Unauthorized,
+                        };
                     }
                 }
             }
--- a/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
+++ b/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
@@ -388,7 +388,8 @@ namespace Microsoft.Azure.Devices.Client
 
                         throw new IotHubClientException(
                             $"Failed to provide certificates in the chain - {ex.Message}",
-                            IotHubClientErrorCode.Unauthorized, ex);
+                            IotHubClientErrorCode.Unauthorized,
+                            ex);
                     }
                 }
             }
--- a/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
+++ b/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
@@ -386,7 +386,7 @@ namespace Microsoft.Azure.Devices.Client
                         if (Logging.IsEnabled)
                             Logging.Error(null, $"{nameof(CertificateInstaller)} failed to read or write to cert store due to: {ex}");
 
-                        throw new IotHubClientException($"Failed to provide certificates in the chain - {ex.Message}", IotHubClientErrorCode.Unauthorized, ex);
+                        throw new IotHubClientException($"Failed to provide certificates in the chain - {ex.Message}", errorCode: IotHubClientErrorCode.Unauthorized, innerException: ex);
                     }
                 }
             }
--- a/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
+++ b/iothub/device/src/Authentication/IotHubConnectionCredentials.cs
@@ -386,10 +386,9 @@ namespace Microsoft.Azure.Devices.Client
                         if (Logging.IsEnabled)
                             Logging.Error(null, $"{nameof(CertificateInstaller)} failed to read or write to cert store due to: {ex}");
 
-                        throw new IotHubClientException($"Failed to provide certificates in the chain - {ex.Message}", ex)
-                        {
-                            ErrorCode = IotHubClientErrorCode.Unauthorized,
-                        };
+                        throw new IotHubClientException(
+                            $"Failed to provide certificates in the chain - {ex.Message}",
+                            IotHubClientErrorCode.Unauthorized, ex);
                     }
                 }
             }
--- a/iothub/device/src/Exceptions/ExceptionHandlingHelper.cs
+++ b/iothub/device/src/Exceptions/ExceptionHandlingHelper.cs
@@ -15,21 +15,19 @@ namespace Microsoft.Azure.Devices.Client
         {
             var mappings = new Dictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>
             {
-                { 
+                {
                     HttpStatusCode.NoContent,
                     async (response) =>
-                        new IotHubClientException(CreateMessageWhenDeviceNotFound(await GetExceptionMessageAsync(response).ConfigureAwait(false)))
-                        {
-                            ErrorCode = IotHubClientErrorCode.DeviceNotFound,
-                        }
+                        new IotHubClientException(
+                            CreateMessageWhenDeviceNotFound(await GetExceptionMessageAsync(response).ConfigureAwait(false)),
+                            IotHubClientErrorCode.DeviceNotFound)
                 },
                 { 
                     HttpStatusCode.NotFound,
                     async (response) =>
-                        new IotHubClientException(CreateMessageWhenDeviceNotFound(await GetExceptionMessageAsync(response).ConfigureAwait(false)))
-                        {
-                            ErrorCode = IotHubClientErrorCode.DeviceNotFound,
-                        }
+                        new IotHubClientException(
+                            CreateMessageWhenDeviceNotFound(await GetExceptionMessageAsync(response).ConfigureAwait(false)),
+                            IotHubClientErrorCode.DeviceNotFound)
                 },
                 { 
                     HttpStatusCode.BadRequest,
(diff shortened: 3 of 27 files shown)


iothub/device/src/Exceptions/IotHubClientErrorCode.cs
```diff
@@ -9,9 +9,9 @@ namespace Microsoft.Azure.Devices.Client
     public enum IotHubClientErrorCode
     {
         /// <summary>
-        /// The request completed without exception.
+        /// Used when the error code returned by the hub is unrecognized. If encountered, please report the issue so it can be added here.
         /// </summary>
-        Ok,
+        Unknown,
```
abhipsaMisra, line 14: Changing the default case from "Ok" to "Unknown". 
IotHubClientErrorCode being used only via IotHubClientException, a value of "Ok" didn't seem right.

iothub/device/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -11,7 +11,21 @@ namespace Microsoft.Azure.Devices.Client
...
+        /// <summary>
+        /// Mapper to map IoT hub service returned 6 digit error codes to <see cref="IotHubClientException.ErrorCode"/>.
+        /// </summary>
+        /// <param name="errorCode">The 6 digit error code returned by IoT hub service.</param>
+        /// <returns>The IotHubClientErrorCode that is returned as a part of IotHubClientException.</returns>
+        internal static IotHubClientErrorCode GetIotHubClientErrorCode(string errorCode)
+        {
+            return errorCode switch
+            {
+                "400004" => IotHubClientErrorCode.ArgumentInvalid,
```
abhipsaMisra, line 23: I have just this one error code mapped for now. We can add on to this as we encounter other cases.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1030,7 +1033,7 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
                     {
                         Status = status,
                         Version = version,
-                        Message = Encoding.UTF8.GetString(payloadBytes), // This will only ever contain an error message which is encoded based on service contract (UTF-8).
+                        ErrorResponseMessage = JsonConvert.DeserializeObject<IotHubClientErrorResponseMessage>(Encoding.UTF8.GetString(payloadBytes)), // This will only ever contain an error message which is encoded based on service contract (UTF-8).
```
abhipsaMisra, line 1036: I've made this change only for the mqtt-> patch twin flow. However, this will need to go into the amqp flow as well.

iothub/device/src/Exceptions/ExceptionHandlingHelper.cs
```diff
@@ -11,7 +11,21 @@ namespace Microsoft.Azure.Devices.Client
 {
     internal class ExceptionHandlingHelper
     {
-        public static IDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>> GetDefaultErrorMapping()
+        /// <summary>
+        /// Mapper to map IoT hub service returned 6 digit error codes to <see cref="IotHubClientException.ErrorCode"/>.
+        /// </summary>
+        /// <param name="errorCode">The 6 digit error code returned by IoT hub service.</param>
```
brycewang-microsoft, line 17: I am not sure if this field would always be 6 digit numeric codes. We do have cases of non-numeric codes returned by the hub on the service client side. Error message is like [this](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/service/tests/ExceptionHandlingHelperTests.cs#L107).

iothub/device/src/Exceptions/IotHubClientErrorCode.cs
```diff
@@ -142,5 +142,15 @@ public enum IotHubClientErrorCode
         /// device certificates and ensure they are up-to-date.
         /// </remarks>
         TlsAuthenticationError,
+
+        /// <summary>
+        /// Something in the request payload is invalid. Check the error message for more information about what
+        /// is invalid.
+        /// </summary>
+        /// <remarks>
+        /// One example when this error is encountered is when the twin property patch does not meet the specified restrictions.
```
brycewang-microsoft, line 151:
```suggestion
        /// One example of encountering with this error is when the twin property patch does not meet the specified restrictions.
```

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
brycewang-microsoft: We may need to wrap this with a try-catch block for `JsonReaderException` as the payload returned from the hub may not always be what we are expecting to receive. Similar handling on the service client side is like [this](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/service/src/Exceptions/ExceptionHandlingHelper.cs#L29).

iothub/device/src/Exceptions/ExceptionHandlingHelper.cs
drwill-ms, line 23: Why not add an enum with a name of ArgumentInvalid and a value of 400004? Then Enum.TryParse?

iothub/device/src/Exceptions/IotHubClientErrorCode.cs
drwill-ms: Unless the Hub actually returns the value 'Ok', but I don't know why it would for an error.

iothub/device/src/Authentication/HsmAuthentication/Transport/HttpBufferedStream.cs
```diff
@@ -54,7 +54,7 @@ internal async Task<string> ReadLineAsync(CancellationToken cancellationToken)
 
                 if (length == 0)
                 {
-                    throw new IotHubClientException("Unexpected end of stream.", IotHubClientErrorCode.NetworkErrors);
+                    throw new IotHubClientException("Unexpected end of stream.", errorCode: IotHubClientErrorCode.NetworkErrors);
```
drwill-ms, line 57: Why are we naming parameters?

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -37,65 +37,20 @@ protected internal IotHubClientException() : base()
         {
         }
 
-        /// <summary>
-        /// Creates an instance of this class.
-        /// </summary>
-        /// <param name="message">The message that describes the error.</param>
-        /// <param name="innerException">The exception that is the cause of the current exception.</param>
-        protected internal IotHubClientException(string message, Exception innerException = null)
```
drwill-ms, line 45: I see, because you removed overloads.

Well, I didn't love the many ctor overloads either. I'd just rather we use settable properties so the instantiator can set what they want, and make the required values be ctor params (probably message and inner exception).

abhipsaMisra: I don't mind that approach since the multiple optional params on the constructor didn't feel very user friendly to me. Having settable internal properties seems like a better option.

iothub/device/src/Exceptions/ExceptionHandlingHelper.cs
abhipsaMisra: Interesting, yeah.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: That is a good point. 
Since we don't know (don't have any documentation) what all formats the error response could be in, how about this - for now we parse out the error message based on the format we know. If we encounter an exception we don't parse and set the ErrorCode to "Unknown". Once a customer reaches out to us about the "Unknown" use-case we can look at the error response and add that case in.

iothub/device/src/Exceptions/ExceptionHandlingHelper.cs
abhipsaMisra: That is fair. However, since we don't know of any such case on the client side yet, we'll have to wait until we encounter this scenario and then address it.

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -113,26 +65,32 @@ protected internal IotHubClientException(SerializationInfo info, StreamingContex
...
+        public string TrackingId { get; protected internal set; } = string.Empty;
 
         /// <summary>
-        /// The service returned tracking Id associated with this particular error.
+        /// Indicates if the error is transient and should be retried.
         /// </summary>
-        public string TrackingId { get; protected internal set; }
+        public bool IsTransient
+        {
+            get => _isTransient != null && (bool)_isTransient;
```
drwill-ms, line 78:
```suggestion
            get => _isTransient.HasValue && _isTransient.Value;
```

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -180,26 +180,24 @@ private static void AddCustomHeaders(HttpRequestMessage requestMessage, IDiction
...
-                throw new IotHubClientException(ex.Message, IotHubClientErrorCode.NetworkErrors, ex);
+                throw new IotHubClientException(ex.Message, ex)
+                {
+                    ErrorCode = IotHubClientErrorCode.NetworkErrors,
+                };
             }
             catch (Exception ex) when (!Fx.IsFatal(ex) && ex is not OperationCanceledException)
             {
-                throw new IotHubClientException(ex.Message, ex);
+                throw new IotHubClientException(ex.Message, innerException: ex);
```
drwill-ms, line 200: is the named parameter needed?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1029,12 +1122,34 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
...
+                        try
+                        {
+                            errorResponse = JsonConvert.DeserializeObject<IotHubClientErrorResponseMessage>(errorResponseString);
+                        }
+                        catch (JsonException ex)
+                        {
+                            if (Logging.IsEnabled)
+                                Logging.Error(this, $"Failed to parse twin patch error response JSON: {ex}. Message body: '{errorResponseString}'");
+
+                            errorResponse = new IotHubClientErrorResponseMessage()
```
drwill-ms, line 1140:
```suggestion
                            errorResponse = new IotHubClientErrorResponseMessage
```

iothub/device/tests/RetryPolicies/IotHubClientExponentialBackoffRetryPolicyTests.cs
```diff
@@ -46,8 +50,13 @@ public void ExponentialBackoffRetryPolicy_IsExponential(uint retryCount)
             var exponentialBackoff = new IotHubClientExponentialBackoffRetryPolicy(uint.MaxValue, TimeSpan.FromDays(30), false);
             TimeSpan previousDelay = TimeSpan.Zero;
             uint exponent = retryCount + 6; // starts at 7
+            var exception = new IotHubClientException()
```
drwill-ms, line 53:
```suggestion
            var exception = new IotHubClientException
```

iothub/device/tests/RetryPolicies/IotHubClientFixedDelayRetryPolicyTests.cs
```diff
@@ -22,9 +22,13 @@ public void FixedDelayRetryPolicy_IsFixedDelay(uint retryCount)
             // arrange
             var expected = TimeSpan.FromSeconds(10);
             var retryPolicy = new IotHubClientFixedDelayRetryPolicy(0, expected, false);
+            var exception = new IotHubClientException()
```
drwill-ms, line 25:
```suggestion
            var exception = new IotHubClientException
```

iothub/device/tests/RetryPolicies/IotHubClientRetryPolicyBaseTests.cs
```diff
@@ -22,7 +22,10 @@ public void RetryPolicyBase_ObservesMax()
             // arrange
             const uint maxRetries = 2;
             var retryPolicy = new IotHubClientTestRetryPolicy(maxRetries);
-            var ex = new IotHubClientException("", true);
+            var ex = new IotHubClientException()
```
drwill-ms, line 25:
```suggestion
            var ex = new IotHubClientException
```

iothub/device/tests/RetryPolicies/IotHubClientRetryPolicyBaseTests.cs
```diff
@@ -34,7 +37,10 @@ public void RetryPolicyBase_ObservesInifiniteRetries()
         {
             // arrange
             var retryPolicy = new IotHubClientTestRetryPolicy(0);
-            var ex = new IotHubClientException("", true);
+            var ex = new IotHubClientException()
```
drwill-ms, line 40:
```suggestion
            var ex = new IotHubClientException
```

iothub/device/tests/RetryPolicies/IotHubClientRetryPolicyBaseTests.cs
```diff
@@ -47,7 +53,10 @@ public void RetryPolicyBase_IotHubException_ReturnsTrueWhenTransient(bool isTran
         {
             // arrange
             var retryPolicy = new IotHubClientTestRetryPolicy(0);
-            var ex = new IotHubClientException("", isTransient);
+            var ex = new IotHubClientException()
```
drwill-ms, line 56:
```suggestion
            var ex = new IotHubClientException
```

drwill-ms (approved):

iothub/device/src/Exceptions/IotHubClientException.cs
```diff
@@ -113,26 +65,20 @@ protected internal IotHubClientException(SerializationInfo info, StreamingContex
...
+        /// Indicates if the error is transient and should be retried.
         /// </summary>
-        public string TrackingId { get; protected internal set; }
+        public bool IsTransient { get; protected internal set; }
 
         /// <summary>
         /// The specific error code.
         /// </summary>
-        public IotHubClientErrorCode ErrorCode { get; protected internal set; }
+        public IotHubClientErrorCode ErrorCode { get; private set; }
```
abhipsaMisra, line 81: Settable only through constructor

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -386,7 +386,10 @@ private void Validate()
                         if (Logging.IsEnabled)
                             Logging.Error(null, $"{nameof(CertificateInstaller)} failed to read or write to cert store due to: {ex}");
 
-                        throw new IotHubClientException($"Failed to provide certificates in the chain - {ex.Message}", IotHubClientErrorCode.Unauthorized, ex);
+                        throw new IotHubClientException(
```
abhipsaMisra, line 389: minor - split large message into separate lines for readability

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -318,12 +318,20 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
                     case MqttClientConnectResultCode.BadUserNameOrPassword:
                     case MqttClientConnectResultCode.NotAuthorized:
                     case MqttClientConnectResultCode.ClientIdentifierNotValid:
-                        throw new IotHubClientException("Failed to open the MQTT connection due to incorrect or unauthorized credentials", IotHubClientErrorCode.Unauthorized, ex);
+                        throw new IotHubClientException(
```
abhipsaMisra, line 321: minor - split large message into separate lines for readability

abhipsaMisra:
This PR addresses this only for the update twin flow over MQTT. I'll put out a separate PR for updating twin over AMQP (and other operations, if applicable).

drwill-ms (approved):

abhipsaMisra:
This has been released in https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2023-02-09-v2preview

PR #2971, opened 2022-11-17
Stop throwing error code DeviceMaximumQueueDepthExceeded from device client

brycewang-microsoft:
With `AmqpErrorCode.ResourceLimitExceeded`, we are supposed to throw `IotHubClientException` with `QuotaExceeded` other than `DeviceMaximumQueueDepthExceeded `. Now that we are shipping v2, it's a good time to correct this behavior as a breaking change.


e2e/test/iothub/device/MessageSendFaultInjectionTests.cs
```diff
@@ -237,46 +236,28 @@ public async Task Message_ThrottledConnectionLongTimeNoRecovery_AmqpWs()
...
+                    new IotHubClientAmqpSettings(),
+                    FaultInjectionConstants.FaultType_QuotaExceeded,
+                    FaultInjectionConstants.FaultCloseReason_Boom)
                 .ConfigureAwait(false);
-            };
-
-            // assert
-            var error = await act.Should().ThrowAsync<IotHubClientException>();
-            error.And.ErrorCode.Should().Be(IotHubClientErrorCode.DeviceMaximumQueueDepthExceeded);
-            error.And.IsTransient.Should().BeFalse();
```
brycewang-microsoft, line 253: As `QuotaExceeded` is transient, the "SendMessageRecoveryAsync" call will succeed with retries. We won't expect `IotHubClientException` thrown out here.

e2e/test/iothub/device/MessageSendFaultInjectionTests.cs
```diff
@@ -237,46 +236,28 @@ public async Task Message_ThrottledConnectionLongTimeNoRecovery_AmqpWs()
...
         }
 
         [TestMethod]
         [Timeout(TestTimeoutMilliseconds)]
-        public async Task Message_AuthenticationRecovery_Amqp()
+        public async Task Message_AuthenticationNoRecovery_Amqp()
```
brycewang-microsoft, line 260: `Unauthorized` error code is intransient, so the method call won't succeed or be retried. Correct the test name here.

e2e/test/iothub/device/MessageSendFaultInjectionTests.cs
abhipsaMisra: Should we mark quota exceeded as non-transient? Until the quota thresholds are refreshed retrying this operation won't help.

abhipsaMisra (approved):
Only comment is whether we should mark quota exceeded exception as non-retryable.

e2e/test/iothub/device/MessageSendFaultInjectionTests.cs
brycewang-microsoft: This is a good question. I notice we are currently marking quota exceeded as transient for both hub device and service clients. I forgot why we are doing so, but let me check within our team if any objection of changing it to non-transient.

drwill-ms: It should be transient. Also remember that a retry policy can give up at some point (still retryable) and someone may have NoRetryPolicy in use.

drwill-ms (approved):

andyk-ms (approved):

schoims (approved):

e2e/test/iothub/device/MessageSendFaultInjectionTests.cs
drwill-ms: All quota I know of are time-based, meaning at some point the operation can be successful. It might be < a minute, or could be until the next day.

brycewang-microsoft: I see, I believe this also explained why retrying for quota exceed error succeeded. I will leave it as transient then.

PR #2974, opened 2022-11-18
Remove Required properties, use property names where possible and rem…

drwill-ms:
System.Text.Json doesn't support the Required property, and our thought is that we don't need to use it. The service should be sending stuff and we don't need to validate, and for user's data that gets serialized, we can let the service validate.

I found a case where we weren't using JsonPropertyName to decorate a property, and used a converter. I feel we should be able to avoid using the converter.


iothub/service/src/Registry/Models/ExportImportDevice.cs
```diff
@@ -168,7 +168,7 @@ public ExportImportDevice(Device device, ImportMode importmode)
         /// For more information, see
         /// <see href="https://docs.microsoft.com/azure/iot-edge/iot-edge-as-gateway?view=iotedge-2020-11#parent-and-child-relationships"/>.
         /// </remarks>
-        [JsonPropertyName("deviceScope", NullValueHandling = NullValueHandling.Include)]
```
abhipsaMisra, line 171: Does System.Text.Json support NullValueHandling.Ignore? That seems like a case that we might want to use in some cases.

iothub/service/src/Twin/Models/ClientTwinJsonConverter.cs
```diff
@@ -1,422 +0,0 @@
...
-using Azure;
-using System.Text.Json;
-using System.Text.Json.Serialization;
-
-namespace Microsoft.Azure.Devices
-{
-    /// <summary>
-    /// Converts a twin to JSON.
-    /// </summary>
-    internal sealed class ClientTwinJsonConverter : JsonConverter<ClientTwin>
```
abhipsaMisra, line 18: Nice that we can get rid of all this serialization/deserialization code.

provisioning/service/src/Twin/ProvisioningClientCapabilities.cs
```diff
@@ -14,6 +14,6 @@ public class ProvisioningClientCapabilities
         /// IoT Edge capability.
         /// </summary>
         [JsonPropertyName("iotEdge")]
-        public bool IotEdge { get; set; }
+        public bool IsIotEdge { get; set; }
```
abhipsaMisra, line 17: 👍🏽

abhipsaMisra (approved):
LGTM

iothub/service/src/Registry/Models/ExportImportDevice.cs
drwill-ms: No, but you can get it by using `JsonSerializerOptions`, however, I think we probably shouldn't use it as it is how a user can explicitly unset a value.

iothub/service/src/Twin/Models/ClientTwinJsonConverter.cs
drwill-ms: I hope so.

Issue #2921, opened 2022-10-27
[Technical Question] AuthenticationException: The remote certificate is invalid because of errors in the certificate chain: RevocationStatusUnknown, OfflineRevocation


Marusyk:
The issue is in the China region when I try to connect my device to DPS.
For other regions it works as expected, even when I connect from Europe to China DPS it works. The problem is only when connecting from China to China  ("global.azure-devices-provisioning.cn").

Could you please point me to where it can be an issue?
Please let me know what information should I provide to investigate the issue.

StackTrace:

```
Microsoft.Azure.Devices.Provisioning.Client.ProvisioningTransportException: MQTT transport exception
 ---> System.Security.Authentication.AuthenticationException: The remote certificate is invalid because of errors in the certificate chain: RevocationStatusUnknown, OfflineRevocation
   at System.Net.Security.SslStream.SendAuthResetSignal(ProtocolToken message, ExceptionDispatchInfo exception)
   at System.Net.Security.SslStream.CompleteHandshake(SslAuthenticationOptions sslAuthenticationOptions)
   at System.Net.Security.SslStream.ForceAuthenticationAsync[TIOAdapter](TIOAdapter adapter, Boolean receiveFirst, Byte[] reAuthenticationData, Boolean isApm)
   at Microsoft.Azure.Devices.Provisioning.Client.Transport.ProvisioningTransportHandlerMqtt.ProvisionOverTcpCommonAsync(ProvisioningTransportRegisterMessage message, ClientTlsSettings tlsSettings, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.Provisioning.Client.Transport.ProvisioningTransportHandlerMqtt.RegisterAsync(ProvisioningTransportRegisterMessage message, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at Microsoft.Azure.Devices.Provisioning.Client.Transport.ProvisioningTransportHandlerMqtt.RegisterAsync(ProvisioningTransportRegisterMessage message, CancellationToken cancellationToken)
   at AzureIoTHubAdapter.AzureIoTHubAdapter.RegisterDeviceAsync(SecurityProviderX509Certificate security, String dpsIdScope)
Microsoft.Azure.Devices.Provisioning.Client.ProvisioningTransportException: MQTT transport exception
[...]

PR #2975, opened 2022-11-18
fix - v2 GatewayHostname ignored for MQTT

tmahmood-microsoft:
1. Updated implementation to use GatewayHostname when provided instead of always using IotHubHostname (MQTT).
2. For [AMQP implementation](https://github.com/Azure/azure-amqp/blob/6ae2104a3497dbe71113941c003b5ae0d94a6414/src/Transport/TlsTransport.cs#L189), if certificate information is not provided, certificate revocation checks are set to false. MQTT implementation in v2 was missing this change. To fix this, for cases when certificate information is not provided, boolean 'IgnoreCertificateRevocationErrors' needs to be set to true, based on this [MQTT net implementation](https://github.com/dotnet/MQTTnet/blob/e88b280eb0fac04267487555399bff7241af4349/Source/MQTTnet/Implementations/MqttTcpChannel.cs#L113).

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -217,14 +217,12 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             }
 
             var tlsParameters = new MqttClientOptionsBuilderTlsParameters();
-            if (_connectionCredentials.ClientCertificate != null)
-            {
-                tlsParameters.Certificates = new List<X509Certificate> { _connectionCredentials.ClientCertificate };
-            }
-            else
-            {
-                tlsParameters.IgnoreCertificateRevocationErrors = true;
-            }
+            List<X509Certificate> certs = _connectionCredentials.ClientCertificate == null
+                ? new List<X509Certificate>(0)
+                : new List<X509Certificate> { _connectionCredentials.ClientCertificate };
+
+            tlsParameters.Certificates = certs;
+            tlsParameters.IgnoreCertificateRevocationErrors = !settings.CertificateRevocationCheck;
 
             if (_mqttTransportSettings?.RemoteCertificateValidationCallback != null)
             {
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -183,7 +183,7 @@ internal MqttTransportHandler(PipelineContext context, IotHubClientMqttSettings
             _mqttClient = mqttFactory.CreateMqttClient();
             _mqttClientOptionsBuilder = new MqttClientOptionsBuilder();
 
-            _hostName = context.IotHubConnectionCredentials.IotHubHostName;
+            _hostName = context.IotHubConnectionCredentials.HostName;
```
abhipsaMisra, line 186: 👍🏽 
This can be either the hub name directly or a gateway name, right?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -217,12 +217,14 @@ internal MqttTransportHandler(PipelineContext context, IotHubClientMqttSettings
...
-                : new List<X509Certificate> { _connectionCredentials.ClientCertificate };
-
-            tlsParameters.Certificates = certs;
+            if (_connectionCredentials.ClientCertificate != null)
+            {
+                tlsParameters.Certificates = new List<X509Certificate> { _connectionCredentials.ClientCertificate };
+            }
+            else
+            {
+                tlsParameters.IgnoreCertificateRevocationErrors = true;
```
abhipsaMisra, line 226: Is there a spec that defines the relationship between client certificates being set and the certificate revocation check being set in TLS settings? If not, then can we expose this setting to the user so that they can set it however they want. That way we can address issues such as https://github.com/Azure/azure-iot-sdk-csharp/issues/2921.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
tmahmood-microsoft: correct.

`HostName = gatewayHostName ?? hostName;`

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
drwill-ms: Yeah, this seems like a bad assumption to me.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1108,12 +1110,12 @@ private void RemoveOldOperations(object _)
             _ = _twinResponseTimeouts
                 .Where(x => DateTimeOffset.UtcNow - x.Value > s_twinResponseTimeout)
                 .Select(x =>
-                    {
```
drwill-ms, line 1111: I prefer this tabbing. The method body is a parameter, and as such should be indented under the call to `Select`.

drwill-ms (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
tmahmood-microsoft: Now using the 'CertificateRevocationCheck' in IotHubClientTransportSettings to control it.

abhipsaMisra (approved):

PR #2976, opened 2022-11-18
Remove out-of-support targets from test matrix

abhipsaMisra:
With this new version we can update out test matrix to only test against supported targets. Targets that have reached EOL will no longer be tested against.

Support matrix: https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

Note: .NET 3.1 is also set to reach EOL in Dec 22. Should we remove this target from our testing matrix?

<html>
<body>

Version | Original release date | Latest patch version | Patch release date | Release type | Support phase | End of support
-- | -- | -- | -- | -- | -- | --
.NET Core 3.1 | December 3, 2019 | 3.1.31 | November 8, 2022 | LTS | Maintenance | December 13, 2022

</body>
</html>


e2e/test/E2ETests.csproj
```diff
@@ -1,17 +1,15 @@
...
     <LangVersion>9.0</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <IsPackable>false</IsPackable>
     <RootNamespace>Microsoft.Azure.Devices.E2ETests</RootNamespace>
     <AssemblyName>Microsoft.Azure.Devices.E2ETests</AssemblyName>
     <RootDir>$(MSBuildProjectDirectory)\..\..</RootDir>
+    <!-- We will turn on this suppression once we are ready to GA-->
     <!-- Suppress framework end of life warnings as we have to keep supporting these frameworks for our customers -->
-    <CheckEolTargetFramework>false</CheckEolTargetFramework>
+    <!-- <CheckEolTargetFramework>false</CheckEolTargetFramework> -->
```
abhipsaMisra, line 12: We'll add this suppression back once we are ready to GA

vsts/vsts.yaml
```diff
@@ -43,9 +40,7 @@ variables:
   testNetcore31: ${{ or(eq(variables['minMatrix'], 'True'), contains(parameters['testTargets'], 'netcore3.1')) }}
 
 # The remaining build/test targets.
-  testNet50: ${{ or(eq(variables['allTestTargets'], 'True'), contains(parameters['testTargets'], 'net5.0')) }}
-  testNetcore21: ${{ or(eq(variables['allTestTargets'], 'True'), contains(parameters['testTargets'], 'netcore2.1')) }}
-  testNet472: ${{ or(eq(variables['allTestTargets'], 'True'), contains(parameters['testTargets'], 'net4.7.2')) }}
+  # Add remaining targets in here
```
abhipsaMisra, line 43: I've left this template in here for future reference

e2e/test/E2ETests.csproj
drwill-ms: Why not turn it on now?

drwill-ms: Oh, in case something else goes out of support, and we want to change. Got it.

vsts/vsts.yaml
drwill-ms: Shall we add .net 7.0 testing?

drwill-ms (approved):

tmahmood-microsoft (approved):

drwill-ms:
Need to update the supported platforms file, and probably the migraiton guide.

vsts/vsts.yaml
abhipsaMisra: How about we remove net3.1 and add net 7 (net 3.1 is set to go out of support end of this year).

schoims (approved):

vsts/vsts.yaml
drwill-ms: k

vsts/vsts.yaml
```diff
@@ -101,13 +91,18 @@ jobs:
             Write-Host "minMatrix: ${{ variables.minMatrix }}"
             Write-Host "testNet60: ${{ variables.testNet60 }}"
             Write-Host "testNetcore31: ${{ variables.testNetcore31 }}"
```
andyk-ms, line 93: Do we need to remove this?

vsts/vsts.yaml
```diff
@@ -398,50 +357,25 @@ jobs:
...
```
andyk-ms, line 359: Do we need to remove this?

andyk-ms (approved):
Double check on the variable for 3.1 which we are removing.

vsts/vsts.yaml
abhipsaMisra: Retaining these logging statements

supported_platforms.md
```diff
@@ -20,6 +20,8 @@ Nightly test platform details:
 .NET versions tested on
 - .NET 7.0
 - .NET 6.0
+- .NET Core 3.1
+- .NET Framework 4.6.1 (only IoT Hub SDKs)
```
drwill-ms, line 24: Why only Hub SDKs? Is net461 not .netstandard2.0 compliant?

drwill-ms (approved):

supported_platforms.md
abhipsaMisra: net461 isn't fully compliant with ns2.0. For this reason I think we should test against net472 instead (which is what we have been doing). What do you think? @drwill-ms 

From https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0#select-net-standard-version:
The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

drwill-ms: Wow, I wonder why track2 tests with it then.

I'd say stick with 472 for now, but ask the azure sdk team about it.

PR #2982, opened 2022-11-22
Drwill/merge previews v2

drwill-ms:


provisioning/service/src/Http/ContractApiHttp.cs
```diff
@@ -189,6 +194,13 @@ internal class ContractApiHttp : IContractApiHttp
             return response;
         }
 
+        private static bool ContainsAuthenticationException(Exception ex)
+        {
+            return ex != null
```
drwill-ms, line 199: @brycewang-microsoft is this an okay change?

PR #2983, opened 2022-11-22
User shouldn't need to access a single child of a child, so we can

drwill-ms:
flatten

### DIFF
--- a/provisioning/service/src/Twin/InitialTwinState.cs
+++ b/provisioning/service/src/Twin/InitialTwinState.cs
@@ -15,6 +15,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
         /// Specifies the initial desired properties on a twin.
         /// </summary>
         [JsonIgnore]
+        // This is here to flatten the hierarchy to make it simpler for users. The API needn't be as clumsy to use as the JSON hierachy is.
         public InitialTwinPropertyCollection DesiredProperties
         {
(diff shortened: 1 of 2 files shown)


provisioning/service/src/Twin/InitialTwinState.cs
```diff
@@ -11,11 +11,21 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
...
+            get => Properties.Desired;
+            set => Properties.Desired = value;
+        }
+
         /// <summary>
         /// Getter and setter the desired properties.
         /// </summary>
         [JsonPropertyName("properties")]
-        public InitialTwinProperties Properties { get; set; } = new();
+        internal InitialTwinProperties Properties { get; set; } = new();
```
drwill-ms, line 28: We can hide this from the user, but still use it for serialization. They instead access its child directly from `DesiredProperties` above.

provisioning/service/tests/Config/TwinStateTests.cs
```diff
@@ -75,15 +75,12 @@ public void TwinStateSucceedOnNull()
             var initialTwin = new InitialTwinState
             {
                 Tags = tags,
-                Properties =
```
drwill-ms, line 78: Basically fixing this nonsense. It is correct for serialization, but creates an undue burden on the user.

PR #2985, opened 2022-11-22
Delete duplicate service getting started sample directory

ngastelum-ms:


patilsnr:
Did we rebuild the solution file after these were deleted? I can't remember if that makes a difference.

abhipsaMisra (approved):
This also needs to go
https://github.com/Azure/azure-iot-sdk-csharp/tree/previews/v2/provisioning/device/samples/Getting%20Started

patilsnr (approved):

ngastelum-ms:
> Did we rebuild the solution file after these were deleted? I can't remember if that makes a difference.

The solution file is already updated, there's no reference to the uppercase directory so should be fine.

drwill-ms (approved):
Does the solution reference this by case, at all?

PR #2986, opened 2022-11-22
Delete duplicate provisioning/device/samples/Getting Started directory

ngastelum-ms:


abhipsaMisra (approved):

abhipsaMisra:
Confirm that build scripts refer to the collect folder when they try to build the samples in our nightly run

drwill-ms (approved):

drwill-ms:
Does the solution reference this by case at all?

ngastelum-ms:
> Does the solution reference this by case at all?

I double checked and the solution only refers to the lowercase folder, it used to reference the uppercase one before I made the change in a previous PR

PR #2990, opened 2022-11-23
Update hub service direct method JSON usage

drwill-ms:


iothub/service/src/DirectMethod/Models/DirectMethodServiceRequest.cs
```diff
@@ -12,13 +12,17 @@ namespace Microsoft.Azure.Devices
     /// </summary>
     public class DirectMethodServiceRequest
     {
-        private object _payload;
-
         /// <summary>
-        /// Initialize an instance of this class.
+        /// Creates an instance of this class with the specified payload to be serialized.
```
drwill-ms, line 16: I converted this file as close as I could to what we had before, meaning we take their object and serialize it and stuff it in a JSON type. From there we can serialize it (again!?) or offer it back to the user if they want (but why would they want it?).

This is a mostly true update, but I'd rather we change it up and just store the user payload a native object to serialize.

Thoughts?

iothub/service/src/DirectMethod/Models/DirectMethodClientResponse.cs
```diff
@@ -28,15 +28,43 @@ protected internal DirectMethodClientResponse()
...
+            value = default;
+
+            if (JsonPayload == null)
+            {
+                return false;
+            }
+
+            try
+            {
+                value = JsonSerializer.Deserialize<T>(JsonPayload.Value.GetRawText());
```
abhipsaMisra, line 61: Q - what's the difference between `JsonPayload.Value.ToString()` and `JsonPayload.Value.GetRawText()`?

iothub/service/src/DirectMethod/Models/DirectMethodServiceRequest.cs
```diff
@@ -59,35 +63,19 @@ public DirectMethodServiceRequest()
...
-        public string PayloadAsJsonString { get; internal set; }
+        public string PayloadAsJsonString => JsonPayload?.GetRawText();
 
         /// <summary>
-        /// The JSON payload in JRaw type.
+        /// The deserialized payload.
         /// </summary>
         [JsonPropertyName("payload")]
-        internal JRaw JsonPayload { get; set; }
+        internal JsonElement? JsonPayload { get; set; }
```
abhipsaMisra, line 78: Is this the property that eventually gets serialized and sent over in the transport layer?

iothub/service/src/DirectMethod/Models/DirectMethodServiceRequest.cs
abhipsaMisra: > just store the user payload a native object to serialize.

This is what this PR does, right?

abhipsaMisra (approved):

PR #2992, opened 2022-11-28
Deserialize stream

abhipsaMisra:


iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -236,10 +236,9 @@ private static async Task<T> ReadAsAsync<T>(HttpContent content, CancellationTok
         {
             token.ThrowIfCancellationRequested();
 
+            var jsonSerializerOptions = new JsonSerializerOptions();
```
drwill-ms, line 239: Do we need options?

drwill-ms (approved):

iothub/device/src/Transport/Http/HttpClientHelper.cs
abhipsaMisra: Unfortunately, the method signature requires options to be passed in: https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserializeasync?view=net-7.0.
I can make it inline. 
(This is the http layer, so our client isn't using it directly today. If we decide to revert the removal of support in the future, we can make passing in this options configurable).

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -236,10 +236,9 @@ private static async Task<T> ReadAsAsync<T>(HttpContent content, CancellationTok
...
             using Stream stream = await content.ReadAsStreamAsync().ConfigureAwait(false);
-            using var reader = new StreamReader(stream);
-            using var jsonReader = new JsonTextReader(reader);
-            return new JsonSerializer().Deserialize<T>(jsonReader);
+            return await JsonSerializer.DeserializeAsync<T>(stream, jsonSerializerOptions, token).ConfigureAwait(false);
```
abhipsaMisra, line 241:
```suggestion
            return await JsonSerializer.DeserializeAsync<T>(stream, new JsonSerializerOptions(), token).ConfigureAwait(false);
```

iothub/device/src/Transport/Http/HttpClientHelper.cs
abhipsaMisra, line 239:
```suggestion
            
```

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -236,10 +236,9 @@ private static async Task<T> ReadAsAsync<T>(HttpContent content, CancellationTok
         {
             token.ThrowIfCancellationRequested();
 
+
```
abhipsaMisra, line 239:
```suggestion
```

patilsnr (approved):

PR #2994, opened 2022-11-29
Update hub service models ctor for deserialization

drwill-ms:
https://learn.microsoft.com/en-us/dotnet/core/compatibility/serialization/5.0/non-public-parameterless-constructors-not-used-for-deserialization


iothub/service/src/Registry/Models/ExportImportDevice.cs
```diff
@@ -13,38 +13,6 @@ namespace Microsoft.Azure.Devices
...
     public sealed class ExportImportDevice
     {
-        /// <summary>
-        /// The desired and reported properties of the twin.
-        /// </summary>
-        /// <remarks>
-        /// Type definition for the <see cref="Properties"/> property.
-        /// The maximum depth of the object is 10.
-        /// </remarks>
-        public sealed class PropertyContainer
```
drwill-ms, line 23: Moved to another file.

iothub/service/src/DirectMethod/Models/DirectMethodClientResponse.cs
```diff
@@ -18,7 +19,8 @@ public class DirectMethodClientResponse
         /// To unit test methods that use this type as a response, inherit from this class and give it a constructor
         /// that can set the properties you want.
         /// </remarks>
-        protected internal DirectMethodClientResponse()
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public DirectMethodClientResponse()
```
abhipsaMisra, line 23: This doesn't have a parameterized constructor defined, so we should be able to exclude this.

iothub/service/src/Feedback/Models/FeedbackBatch.cs
```diff
@@ -21,7 +22,8 @@ public class FeedbackBatch
         /// To unit test methods that use this type as a response, inherit from this class and give it a constructor
         /// that can set the properties you want.
         /// </remarks>
-        protected internal FeedbackBatch()
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public FeedbackBatch()
```
abhipsaMisra, line 26: No parameterized constructor, so this shouldn't be reqiured, right?

iothub/service/src/DirectMethod/Models/DirectMethodClientResponse.cs
drwill-ms: We had it there for unit testing, but wanted to hide it because normally a user shouldn't be creating this object. I think in this case declaring the ctor but hiding it is the right thing to do.

iothub/service/src/Feedback/Models/FeedbackBatch.cs
drwill-ms: Same as above.

iothub/service/src/Jobs/Models/JobStatus.cs
```diff
@@ -15,49 +14,41 @@ public enum JobStatus
         /// <summary>
         /// Unknown status.
         /// </summary>
-        [EnumMember(Value = "unknown")]
```
abhipsaMisra, line 18: 👍🏽

iothub/service/src/Registry/Models/ExportImportDevice.cs
```diff
@@ -128,7 +94,7 @@ public ExportImportDevice(Device device, ImportMode importmode)
         /// access keys are auto-generated.
         /// </remarks>
         [JsonPropertyName("authentication")]
-        public AuthenticationMechanism Authentication { get; set; }
+        public AuthenticationMechanism Authentication { get; set; } = new();
```
abhipsaMisra, line 97: I noticed that we do this in a bunch of places; any reason why we want this default to empty object and not null?

drwill-ms: To avoid NRE for users when they explore an object.

iothub/service/src/Registry/Models/StorageAuthenticationType.cs
```diff
@@ -16,13 +15,11 @@ public enum StorageAuthenticationType
         /// Use a shared access key for authentication.
         /// </summary>
         /// <remarks>This means authentication must be supplied in the storage URI(s).</remarks>
-        [EnumMember(Value = "keyBased")]
-        KeyBased = 0,
+        KeyBased,
```
abhipsaMisra, line 18: Q - should all enums require an "unknown/none" option that they should default to?

drwill-ms: I like that idea, but I don't feel compelled to do that in this PR.

iothub/service/src/Twin/TwinsClient.cs
```diff
@@ -549,7 +549,6 @@ private static IEnumerable<ExportImportDevice> GenerateExportImportDeviceListFor
                     ImportMode = importMode,
                     TwinETag = twin.ETag,
                     Tags = twin.Tags,
-                    Properties = new ExportImportDevice.PropertyContainer(),
```
abhipsaMisra, line 552: removed?

drwill-ms: The model gives a default now.

abhipsaMisra (approved):

PR #2996, opened 2022-11-29
Fix the logic of checking AuthenticationException as inner exceptions

brycewang-microsoft:


tmahmood-microsoft (approved):

patilsnr (approved):
Out of curiosity, why did we have `while` statements in the first place?

iothub/service/src/Utilities/Fx.cs
```diff
@@ -57,7 +57,7 @@ internal static bool IsFatal(Exception ex)
 
         internal static bool ContainsAuthenticationException(Exception ex)
         {
-            while (ex != null)
+            if (ex != null)
```
drwill-ms, line 60: This can be simplified now:

```csharp
            return ex != null
                && (ex is AuthenticationException
                    || ContainsAuthenticationException(ex.InnerException));
```

iothub/service/src/Utilities/Fx.cs
```diff
@@ -57,7 +57,7 @@ internal static bool IsFatal(Exception ex)
...
             {
                 return ex is AuthenticationException || ContainsAuthenticationException(ex.InnerException);
```
abhipsaMisra, line 62: The recursion is happening in here first, so it doesn't need the `while` loop at the top.

abhipsaMisra (approved):

brycewang-microsoft:
> Out of curiosity, why did we have `while` statements in the first place?

That was my mistake. The inner exception will never be changed, so it is not supposed to be `while` in the first place.

andyk-ms (approved):

PR #2999, opened 2022-12-01
Parse service error fot get twin failures - mqtt

abhipsaMisra:
Related #2967 

Our transport layer currently isn't inspecting and parsing out the service returned error message over AMQP and MQTT. Instead, the entire error "message" block containing the error code, description, tracking Id etc. is put into IotHubClientException.Message. This leaves the parsing responsibility onto the customer.
This PR proposes a change where we inspect the received error message and parse it out.

Another related issue is that as a result of us not parsing out the error message, we currently mark all unsuccessful MQTT twin operations as retryable. This results in the actual error being hidden from the user and the SDK unnecessarily retying non-retryable error scenarios.

This PR also adds back the error parsing logic from here: https://github.com/Azure/azure-iot-sdk-csharp/blob/main/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs#L1130-L1143

### DIFF
--- a/iothub/device/src/Exceptions/IotHubClientErrorCode.cs
+++ b/iothub/device/src/Exceptions/IotHubClientErrorCode.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Azure.Devices.Client
         /// Retrying with exponential back-off could resolve this error. For information on the IoT hub quotas and
         /// throttling, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-quotas-throttling"/>.
         /// </remark>
-        Throttled,
+        Throttled = 429,
 
         /// <summary>
         /// The ETag in the request does not match the ETag of the existing resource.
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -642,8 +642,9 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
                     // Check if we have an int to string error code translation for the service returned error code.
                     // The error code could be a part of the service returned error message, or it can be a part of the topic string.
                     // We will check with the error code in the error message first (if present) since that is the more specific error code returned.
-                    if (Enum.TryParse(getTwinResponse.ErrorResponseMessage.ErrorCode.ToString(), out IotHubClientErrorCode errorCode)
+                    if ((Enum.TryParse(getTwinResponse.ErrorResponseMessage.ErrorCode.ToString(), out IotHubClientErrorCode errorCode)
                         || Enum.TryParse(getTwinResponse.Status.ToString(), out errorCode))
+                        && Enum.IsDefined(typeof(IotHubClientErrorCode), errorCode))
                     {
                         throw new IotHubClientException(getTwinResponse.ErrorResponseMessage.Message, errorCode)
                         {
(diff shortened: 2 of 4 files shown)


tmahmood-microsoft (approved):

schoims (approved):

iothub/device/src/Exceptions/IotHubClientErrorCode.cs
```diff
@@ -72,7 +72,7 @@ public enum IotHubClientErrorCode
         /// Retrying with exponential back-off could resolve this error. For information on the IoT hub quotas and
         /// throttling, see <see href="https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-quotas-throttling"/>.
         /// </remark>
-        Throttled,
+        Throttled = 429,
```
drwill-ms, line 75: Does setting this value do anything?

abhipsaMisra: It'll help in the Enum.TryParse below

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -639,9 +639,22 @@ public override async Task<TwinProperties> GetTwinAsync(CancellationToken cancel
...
-                    throw new IotHubClientException(
-                        getTwinResponse.Message,
-                        IotHubClientErrorCode.NetworkErrors);
+                    // Check if we have an int to string error code translation for the service returned error code.
+                    // The error code could be a part of the service returned error message, or it can be a part of the topic string.
+                    // We will check with the error code in the error message first (if present) since that is the more specific error code returned.
+                    if (Enum.TryParse(getTwinResponse.ErrorResponseMessage.ErrorCode.ToString(), out IotHubClientErrorCode errorCode)
+                        || Enum.TryParse(getTwinResponse.Status.ToString(), out errorCode))
+                    {
+                        throw new IotHubClientException(getTwinResponse.ErrorResponseMessage.Message, errorCode)
```
drwill-ms, line 648: Any idea on whether we can make it retriable or not?

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1061,10 +1078,33 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
...
+                        {
+                            string errorResponseString = Encoding.UTF8.GetString(payloadBytes);
+                            try
+                            {
+                                errorResponse = JsonConvert.DeserializeObject<IotHubClientErrorResponseMessage>(errorResponseString);
+                            }
+                            catch (JsonException ex)
+                            {
+                                if (Logging.IsEnabled)
+                                    Logging.Error(this, $"Failed to parse twin patch error response JSON: {ex}. Message body: '{errorResponseString}'");
```
drwill-ms, line 1094: Consider swapping the order as an exception will be many, many lines and the message body probably not nearly as long.

drwill-ms (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
abhipsaMisra: Depending on the error code returned IotHubClientException will mark it as transient. That is the only way for us to mark a scenario as retriable.
Note that networking issues won't fit into this scenario as they will not result in a `GetTwinResponse`. This logic only looks at the Status within GetTwinResponse to parse out the error code.

Networking issues will still be handled here:
https://github.com/Azure/azure-iot-sdk-csharp/blob/5508c160d584e77fdf8d3613123d4b9dfbb5767e/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs#L649-L670

PR #3003, opened 2022-12-05
Add in unrelated changes from STJ conversion attempt

drwill-ms:

PR #3005, opened 2022-12-05
Fix DPS device client MQTT layer not returning the MQTT client as the sender in the certificate validation callback

timtay-microsoft:
The certificate validation callback should include the "sender" object (which is the mqtt client in this case) but we weren't providing it previously

We already did this in the MQTT layer for the IoT hub device client:

https://github.com/Azure/azure-iot-sdk-csharp/blob/eb20ec22df8b6c19ae483fe3277b81c63077333f/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs#L827


timtay-microsoft:
Closing this because #3003 will make this same change

PR #3006, opened 2022-12-06
Add M2M message failure fixes to lts_2021_03

schoims:
Bringing in 
- Fixed M2M message (MQTT) failure after receiving Twin response or Direct method (https://github.com/Azure/azure-iot-sdk-csharp/pull/2938)

### DIFF
--- a/iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
+++ b/iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
@@ -2,11 +2,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using DotNetty.Transport.Channels;
 using Microsoft.Azure.Devices.Client.Exceptions;
+using Microsoft.Azure.Devices.Shared;
 
 namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 {
@@ -27,7 +30,8 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
 
         private readonly Func<TWork, TWorkId> _getWorkId;
         private readonly Func<IChannelHandlerContext, TWork, Task> _completeWorkAsync;
-        private readonly Queue<IncompleteWorkItem> _incompleteQueue = new Queue<IncompleteWorkItem>();
+
+        private readonly ConcurrentQueue<IncompleteWorkItem> _incompleteQueue = new ConcurrentQueue<IncompleteWorkItem>();
 
         public OrderedTwoPhaseWorkQueue(
             Func<IChannelHandlerContext, TWork, Task> workerAsync,
@@ -43,19 +47,25 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
         {
             if (!_incompleteQueue.Any())
             {
-                throw new IotHubException("Nothing to complete.", isTransient: false);
+                if (Logging.IsEnabled)
+                    Logging.Error(context, $"{nameof(CompleteWorkAsync)} called but there are no items in the queue to complete.", nameof(CompleteWorkAsync));
+
+                Debug.Fail($"{nameof(CompleteWorkAsync)} called but there are no items in the queue to complete.");
+                return TaskHelpers.CompletedTask;
             }
 
-            IncompleteWorkItem incompleteWorkItem = _incompleteQueue.Peek();
-            if (incompleteWorkItem.Id.Equals(workId))
+            if (_incompleteQueue.TryDequeue(out IncompleteWorkItem incompleteWorkItem))
             {
-                _incompleteQueue.Dequeue();
-                return _completeWorkAsync(context, incompleteWorkItem.WorkItem);
-            }
+                if (incompleteWorkItem.Id.Equals(workId))
+                {
+                    return _completeWorkAsync(context, incompleteWorkItem.WorkItem);
+                }
 
-            throw new IotHubException(
-                $"Work must be complete in the same order as it was started. Expected work id: '{incompleteWorkItem.Id}', actual work id: '{workId}'",
-                isTransient: false);
+                throw new IotHubException(
+                    $"Work must be complete in the same order as it was started. Expected work id: '{incompleteWorkItem.Id}', actual work id: '{workId}'",
+                    isTransient: false);
+            }
+            return TaskHelpers.CompletedTask;
         }
 
         protected override async Task DoWorkAsync(IChannelHandlerContext context, TWork work)
@@ -77,17 +87,17 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             if (stateBefore != State
                 && State == States.Aborted)
             {
-                while (_incompleteQueue.Any())
+                while (_incompleteQueue.TryDequeue(out IncompleteWorkItem workItem))
                 {
-                    var workItem = _incompleteQueue.Dequeue().WorkItem as ICancellable;
+                    var cancellableWorkItem = workItem.WorkItem as ICancellable;
 
                     if (exception == null)
                     {
-                        workItem?.Cancel();
+                        cancellableWorkItem?.Cancel();
                     }
                     else
                     {
-                        workItem?.Abort(exception);
+                        cancellableWorkItem?.Abort(exception);
                     }
                 }
             }
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -27,6 +27,7 @@ using DotNetty.Handlers.Tls;
 using DotNetty.Transport.Bootstrapping;
 using DotNetty.Transport.Channels;
 using DotNetty.Transport.Channels.Sockets;
+using Microsoft.Azure.Amqp.Encoding;
 using Microsoft.Azure.Devices.Client.Exceptions;
 using Microsoft.Azure.Devices.Client.Extensions;
 using Microsoft.Azure.Devices.Client.TransientFaultHandling;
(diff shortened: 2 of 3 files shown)


iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -43,19 +45,29 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
...
+                }
+
+                Debug.Fail($"{nameof(CompleteWorkAsync)} called but there are no items in the queue to complete.");
+                return TaskHelpers.CompletedTask;
             }
 
             IncompleteWorkItem incompleteWorkItem = _incompleteQueue.Peek();
-            if (incompleteWorkItem.Id.Equals(workId))
+            if (incompleteWorkItem != null)
             {
```
schoims, line 59: TryDequeue method is not available to all supported platforms, hence added this check

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -710,6 +710,14 @@ private async void ProcessMessage(IChannelHandlerContext context, Packet packet)
 
                     case PacketType.PUBLISH:
                         ProcessPublish(context, (PublishPacket)packet);
+                        // Checking if acknowledgement is needed after receiving direct method from Hub with QoS = 1,
+                        // otherwise its not acknowledged because direct method response does not require CompleteAsync() call.
+                        // Other messages with QoS = 1 are acknowledged from WriteAsync() after CompleteAsync() call.
+                        if (((PublishPacket)packet).QualityOfService == QualityOfService.AtLeastOnce &&
+                            ((PublishPacket)packet).TopicName.StartsWith(MqttTransportHandler.MethodPostTopicPrefix, StringComparison.OrdinalIgnoreCase))
+                        {
+                            await AcknowledgeAsync(context, ((PublishPacket)packet).PacketId.ToString(CultureInfo.InvariantCulture)).ConfigureAwait(false);
```
schoims, line 719: ToString method required IFormatProvider to be specified, so added CultureInfo.InvariantCulture

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -64,9 +65,28 @@ public async Task InitLoopAsync(TimeSpan timeout)
                     timeout)
                 .ConfigureAwait(false);
 
+            // This cancellation token source is disposed when the authentication refresher is disposed
+            // or if this code block is executed more than once per instance of AmqpAuthenticationRefresher (not expected).
```
drwill-ms, line 69: If it isn't expected you can instead opt for a `Debug.Assert(_refresherCancellationTokenSource == null);` instead. The idea being that we catch it in testing.

iothub/device/src/Transport/Mqtt/MqttIotHubAdapter.cs
```diff
@@ -710,6 +710,14 @@ private async void ProcessMessage(IChannelHandlerContext context, Packet packet)
 
                     case PacketType.PUBLISH:
                         ProcessPublish(context, (PublishPacket)packet);
+                        // Checking if acknowledgement is needed after receiving direct method from Hub with QoS = 1,
+                        // otherwise its not acknowledged because direct method response does not require CompleteAsync() call.
+                        // Other messages with QoS = 1 are acknowledged from WriteAsync() after CompleteAsync() call.
+                        if (((PublishPacket)packet).QualityOfService == QualityOfService.AtLeastOnce &&
+                            ((PublishPacket)packet).TopicName.StartsWith(MqttTransportHandler.MethodPostTopicPrefix, StringComparison.OrdinalIgnoreCase))
```
drwill-ms, line 717:
```suggestion
                        if (((PublishPacket)packet).QualityOfService == QualityOfService.AtLeastOnce
                          && ((PublishPacket)packet).TopicName.StartsWith(MqttTransportHandler.MethodPostTopicPrefix, StringComparison.OrdinalIgnoreCase))
```

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -43,19 +45,29 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
         {
             if (!_incompleteQueue.Any())
             {
-                throw new IotHubException("Nothing to complete.", isTransient: false);
+                if (Logging.IsEnabled)
+                {
+                    Logging.Error(context, $"{nameof(CompleteWorkAsync)} called but there are no items in the queue to complete.", nameof(CompleteWorkAsync));
+                }
```
drwill-ms, line 51: We omit braces for logging only, when it is just 1 statement.

```suggestion
                    Logging.Error(context, $"{nameof(CompleteWorkAsync)} called but there are no items in the queue to complete.", nameof(CompleteWorkAsync));
```

iothub/device/src/Transport/Mqtt/OrderedTwoPhaseWorkQueue.cs
```diff
@@ -43,19 +45,29 @@ public Task CompleteWorkAsync(IChannelHandlerContext context, TWorkId workId)
...
+
+                Debug.Fail($"{nameof(CompleteWorkAsync)} called but there are no items in the queue to complete.");
+                return TaskHelpers.CompletedTask;
             }
 
             IncompleteWorkItem incompleteWorkItem = _incompleteQueue.Peek();
```
drwill-ms, line 57: Can we guarantee that if we peek and then dequeue, we'll get the item we peeked?

If we have not full platform support, can we use `#if` preprocessing directives to use the best implementation possible for the more modern ones, and only take that risk on the older ones?

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -143,11 +174,20 @@ private async Task RefreshLoopAsync(DateTime refreshesOn, CancellationToken canc
         public void StopLoop()
         {
             if (Logging.IsEnabled)
+                Logging.Enter(this, nameof(StopLoop));
+
+            try
             {
-                Logging.Info(this, $"{nameof(StopLoop)}");
+                _refresherCancellationTokenSource?.Cancel();
+            }
```
drwill-ms, line 182: should it wait for the loop to end? We have cached the Task from it, right?

drwill-ms (approved):

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
schoims: We cannot wait for the loop to end since this is an infinite loop. Control will exit the loop only when the cancellation token is signaled for cancellation.

schoims: This an unmonitored loop running on a separate thread.

drwill-ms: Right, but you just canceled it on line 181, right?

abhipsaMisra: I thought awaiting on the loop in here might help us catch an OperationCancelledException in the RefreshLoopAsync running unmonitored, but it turns out the only place we refer to the cancellation token is when we check it manually for cancellation. So, the API should not encounter OperationCancellationException.

Would there be any advantage to us awaiting this loop then?

drwill-ms: Awaiting for any exception and logging it.

If we don't await it, let's make sure that method does not throw, and logs anything it catches.

schoims: Sure, updated accordingly.

abhipsaMisra (approved):

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -140,14 +173,43 @@ private async Task RefreshLoopAsync(DateTime refreshesOn, CancellationToken canc
             }
         }
 
-        public void StopLoop()
+        // StopLoop ensures that the resources related to token refresher are cancelled and ready for cleanup.
+        // It will cancel the associated cancellation token source and will await completion of the refresh loop task.
+        // Any exceptions thrown as a result of these operations are caught and logged.
+        // StopLoop will not generate any exceptions so it safe for us to declare this as async void.
```
drwill-ms, line 179: Still don't do this. It hurts nothing to return a Task.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -140,14 +173,43 @@ private async Task RefreshLoopAsync(DateTime refreshesOn, CancellationToken canc
...
+            {
+                try
+                {
+                    await _refreshLoop.ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    if (Logging.IsEnabled)
+                        Logging.Error(this, $"Exception caught while canceling SAS token refresh loop: {ex}", nameof(StopLoop));
+                }
```
drwill-ms, line 207: This try/catch should cover the entire method, including logging and the CTS cancel.

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -140,14 +173,43 @@ private async Task RefreshLoopAsync(DateTime refreshesOn, CancellationToken canc
...
+            if (_refreshLoop != null)
+            {
+                try
+                {
+                    await _refreshLoop.ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    if (Logging.IsEnabled)
+                        Logging.Error(this, $"Exception caught while canceling SAS token refresh loop: {ex}", nameof(StopLoop));
```
drwill-ms, line 206:
```suggestion
                        Logging.Error(this, $"Exception caught while waiting for SAS token refresh loop to exit: {ex}", nameof(StopLoop));
```

drwill-ms (approved):

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
drwill-ms: The call chain doesn't prevent this from returning a task. It just prevents the callers from awaiting it.

Also, we can choose to use IAsyncDisposable, if we wish.

PR #3008, opened 2022-12-06
Disable invalid certificate test setup

abhipsaMisra:


build.ps1
```diff
@@ -395,6 +395,7 @@ try
         # test categories to exclude
         $testCategory += "&TestCategory!=LongRunning"
         $testCategory += "&TestCategory!=Flaky"
+        $testCategory += "&TestCategory!=InvalidServiceCertificate"
```
drwill-ms, line 398: Add a comment to indicate this is temporary until the pipeline is fixed for docker commands on Linux?

Can we only disable it for Linux?

vsts/vsts.yaml
```diff
@@ -167,84 +167,84 @@ jobs:
...
-            cat /etc/hosts
-            echo "====================="
-            echo "Ping containers (URL)"
-            echo "====================="
-            docker ps -a
-            route
-            ping -c 2 invalidcertgde1.westus.cloudapp.azure.com
-            ping -c 2 invalidcertdps1.westus.cloudapp.azure.com
-            ping -c 2 invalidcertiothub1.westus.cloudapp.azure.com
+      # - task: Bash@3
```
drwill-ms, line 170: Same, add comment why.

abhipsaMisra: Sure

build.ps1
abhipsaMisra: sure

PR #3010, opened 2022-12-06
Use ETag type with DigitalTwinClient

drwill-ms:


patilsnr (approved):

iothub/service/src/DigitalTwin/Models/UpdateDigitalTwinOptions.cs
```diff
@@ -15,13 +17,12 @@ public class UpdateDigitalTwinOptions
         /// The request's operation is performed only if this ETag matches the value maintained by the server,
         /// indicating that the entity has not been modified since it was last retrieved.
         /// <para>
-        /// To perform the operation only if the entity exists, set the ETag to the wildcard character <c>"*"</c>.
-        /// To perform the operation unconditionally, leave it the default value of <c>null</c>.
```
abhipsaMisra, line 19: q - are we losing this feature?
`To perform the operation only if the entity exists, set the ETag to the wildcard character <c>"*"</c>`. 
If yes, then it might be worth documenting this in the migration doc.

abhipsaMisra (approved):

iothub/service/src/DigitalTwin/Models/UpdateDigitalTwinOptions.cs
drwill-ms: Tim said it didn't make any sense. You can't patch an item that doesn't already exist.

abhipsaMisra: Was it a relic of the "CreateOrUpdate" style APIs?

drwill-ms: Not sure. Probably a copy pasta.

PR #3012, opened 2022-12-07
Cleanup more models

drwill-ms:


schoims (approved):

iothub/service/src/Feedback/Models/FeedbackRecord.cs
```diff
@@ -20,7 +20,7 @@ public class FeedbackRecord
         /// <remarks>
         /// Feedback messages are delivery acknowledgments for messages sent to a device from IoT hub.
         /// </remarks>
-        [JsonProperty("originalMessageId", DefaultValueHandling = DefaultValueHandling.Ignore)]
```
abhipsaMisra, line 23: q - Is this worth retaining, only because NewtonSoft does let us do this, and it might save someone some bandwidth?

iothub/service/src/Jobs/Models/JobRequest.cs
```diff
@@ -15,13 +15,13 @@ internal class JobRequest
         /// <summary>
         /// Job identifier.
         /// </summary>
-        [JsonProperty("jobId", Required = Required.Always)]
```
abhipsaMisra, line 18: 👍🏽

abhipsaMisra (commented):
Just the one question about DefaultValueHandling

iothub/service/src/Jobs/Models/JobRequest.cs
abhipsaMisra: We should call this out in the doc comment - that this field is required.

abhipsaMisra (commented):
If not already done - we'll need to update the migration doc to include instructions for users to inspect the doc comments of models to figure out which properties are required.

iothub/service/src/Feedback/Models/FeedbackRecord.cs
drwill-ms: Bandwidth is an interesting reason.

I was thinking more about, across the board and not specifically on this property, enabling users to remove a property by nulling it out. We should be including their nulls. But how do we distinguish between null being default and a model formerly having a value that is now null?

drwill-ms:
> If not already done - we'll need to update the migration doc to include instructions for users to inspect the doc comments of models to figure out which properties are required.

Where a field was required, I made it required in the ctor.

abhipsaMisra:
> > If not already done - we'll need to update the migration doc to include instructions for users to inspect the doc comments of models to figure out which properties are required.
> 
> Where a field was required, I made it required in the ctor.

Nice!

iothub/service/src/Feedback/Models/FeedbackRecord.cs
abhipsaMisra: iirc the only place where service supports removing properties based on "null" value are the desired and reported "properties" within a twin. I don't think other model properties support this behavior. So, for the rest of the properties, as long as the property is optional excluding it should be acceptable.

PR #3013, opened 2022-12-08
Update JSON for Hub service client twins

drwill-ms:


iothub/service/src/Twin/Models/ClientTwin.cs
```diff
@@ -33,6 +30,7 @@ public ClientTwin(string deviceId)
         /// <summary>
         /// Gets and sets the twin Id.
         /// </summary>
+        [JsonProperty("deviceId")]
```
drwill-ms, line 33: We can ditch the converter and use attributes.

patilsnr (approved):

andyk-ms (approved):

abhipsaMisra (approved):

PR #3014, opened 2022-12-08
Fix unit test case that does not match its name

schoims:


iothub/service/tests/Registry/DevicesClientTests.cs
```diff
@@ -310,10 +310,10 @@ public async Task DevicesClient_SetAsync_WithOnlyIfUnchangedTrue()
                 s_retryHandler);
 
             // act
-            Func<Task> act = async () => await devicesClient.SetAsync(new List<Device> { goodDevice1, goodDevice2 }, false).ConfigureAwait(false);
+            Func<Task> act = async () => await devicesClient.SetAsync(new List<Device> { badDevice1, badDevice2 }, true).ConfigureAwait(false);
```
schoims, line 313: OnlyIfUnchanged parameter was incorrectly set to false

iothub/service/tests/Registry/DevicesClientTests.cs
```diff
@@ -285,8 +285,8 @@ public async Task DevicesClient_SetAsync_EmptyListThrows()
         public async Task DevicesClient_SetAsync_WithOnlyIfUnchangedTrue()
         {
             // arrange
-            var goodDevice1 = new Device("123") { ConnectionState = ClientConnectionState.Connected };
-            var goodDevice2 = new Device("234") { ConnectionState = ClientConnectionState.Connected };
+            var badDevice1 = new Device("123") { ConnectionState = ClientConnectionState.Connected };
+            var badDevice2 = new Device("234") { ConnectionState = ClientConnectionState.Connected };
```
schoims, line 289: They are not good devices because they are missing ETag properties

tmahmood-microsoft (approved):

iothub/service/tests/Registry/DevicesClientTests.cs
abhipsaMisra: You could add this comment to the test file so that it explains why these devices are "bad"!
Alternatively, 
var deviceWithoutETag1 = new Device("123") { ConnectionState = ClientConnectionState.Connected };

abhipsaMisra (approved):

iothub/service/tests/Registry/DevicesClientTests.cs
schoims: Sure

drwill-ms (approved):

PR #3016, opened 2022-12-09
Remove converter for twin metadata

drwill-ms:


iothub/service/src/Registry/Models/ExportImportDevice.cs
```diff
@@ -134,7 +134,7 @@ public ExportImportDevice(string deviceId, ImportMode importMode)
         /// The desired and reported properties for the device or module.
         /// </summary>
         [JsonProperty("properties")]
-        public PropertyContainer Properties { get; set; } = new();
```
drwill-ms, line 137: We can retire this other type and reuse the one we use for Twin.

iothub/service/src/Twin/Models/ClientTwinMetadata.cs
```diff
@@ -10,28 +13,49 @@ namespace Microsoft.Azure.Devices
...
+        [JsonExtensionData]
+        internal IDictionary<string, JToken> Properties { get; set; } = new Dictionary<string, JToken>();
+
+        /// <summary>
+        /// Gets the specified property's metadata by name as another metadata object.
+        /// </summary>
+        /// <param name="propertyName">The name of the property.</param>
+        /// <param name="propertyValue">The property's metadata.</param>
+        /// <returns>True if the property exists and could be converted, otherwise false.</returns>
+        public bool TryGetPropertyMetadata(string propertyName, out ClientTwinMetadata propertyValue)
```
drwill-ms, line 40: This should only ever be ClientTwinMetadata, so we could consider changing to `GetPropertyMetadata` which returns the property, but I figured we might as well follow the same pattern as elsewhere in the Twin code.

Thoughts?

iothub/service/src/Twin/Models/ClientTwinProperties.cs
```diff
@@ -12,102 +12,28 @@ namespace Microsoft.Azure.Devices
...
-        /// Creates an instance of this class using the given JSON fragments for the body and metadata.
+        /// Gets the version of the collection.
         /// </summary>
-        /// <param name="twinJson">JSON fragment containing the twin data.</param>
-        /// <param name="metadataJson">JSON fragment containing the metadata.</param>
-        public ClientTwinProperties(string twinJson, string metadataJson)
-            : this(JObject.Parse(twinJson), JObject.Parse(metadataJson))
-        {
-        }
+        [JsonProperty("$version", DefaultValueHandling = DefaultValueHandling.Ignore)]
```
drwill-ms, line 20: Instead of a converter, we now use attributes.

iothub/service/src/Twin/Models/ClientTwinProperties.cs
```diff
@@ -12,102 +12,28 @@ namespace Microsoft.Azure.Devices
...
-            _metadata = metadataJson;
-        }
-
-        /// <summary>
-        /// Gets the version of the collection.
-        /// </summary>
-        public long Version => !TwinData.TryGetValue(VersionName, out JToken versionToken)
-            ? default
-            : (long)versionToken;
+        [JsonExtensionData]
```
drwill-ms, line 29: This gets all the user-defined properties.

iothub/service/src/Twin/Models/ClientTwinMetadata.cs
abhipsaMisra: I like TryGet here, for the reason you mentioned about maintaining the same pattern

abhipsaMisra: For nested metadata, TryGet would be useful, for example - someone tries to get a metadata for a property update that doesn't exist

drwill-ms: > For nested metadata, TryGet would be useful, for example - someone tries to get a metadata for a property update that doesn't exist

True!

iothub/service/src/Twin/Models/ClientTwinProperties.cs
```diff
@@ -117,81 +43,12 @@ public int Count
...
-
-        /// <summary>
-        /// Clears all metadata out of the twin collection as well as the base metadata object.
-        /// </summary>
-        /// <remarks>
-        /// This will remove all metadata from the base metadata object as well as the metadata for the twin collection. 
-        /// This will also clear the underlying metadata object which will affect methods such as
-        /// <see cref="GetMetadata"/> and <see cref="GetLastUpdatedVersion"/>.
-        /// </remarks>
-        public void ClearMetadata()
```
abhipsaMisra, line 160: Do we still have this?

abhipsaMisra (approved):

iothub/service/src/Twin/Models/ClientTwinProperties.cs
drwill-ms: No, I assumed since the metadata is now accessible by properties, no one needs method access to it.

Is there a case for it?

PR #3022, opened 2022-12-14
Make CRL inspection configurable

abhipsaMisra:
Certificate revocation lists are quickly becoming a thing of the past since CRLs can grow to be very large and these checks can quickly become resource intensive. As an alternative, most services are now moving to OCSP Stapling. Keeping this in mind, it is a good idea for us to not make certificate revocation list checks mandatory, but instead to allow this to be configurable by the user.

OCSP Stapling - https://knowledge.digicert.com/quovadis/ssl-certificates/ssl-general-topics/what-is-ocsp-stapling.html

Current behavior - 
IoT hub client and provisioning device client:
-> Amqp => does not set the CRL check directly on the TLS settings but allows the user to specify a server certificate validation callback where any checks can be overridden
-> Mqtt => sets it on the TLS settings

### DIFF
--- a/e2e/test/helpers/TestDeviceCallbackHandler.cs
+++ b/e2e/test/helpers/TestDeviceCallbackHandler.cs
@@ -175,7 +175,6 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
             _methodCallbackSemaphore?.Dispose();
             _twinCallbackSemaphore?.Dispose();
             _receivedMessageCallbackSemaphore?.Dispose();
-            GC.SuppressFinalize(this);
         }
     }
 }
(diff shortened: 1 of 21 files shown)


timtay-microsoft (approved):

abhipsaMisra:
From [#2955 ](https://github.com/Azure/azure-iot-sdk-csharp/pull/2955#issuecomment-1350123631):

In main - CertificateRevocationCheck on TlsVersions current defaults to false. In order for us to make this check configurable for the provisioning client without breaking the behavior, we'd need to introduce another property on our transport settings. Even though this change would be useful to users, we'll hold off on making this change until we have a request for this.

PR #3024, opened 2022-12-14
Fix build errors in previews/v2

abhipsaMisra:
Issues:

- ~Fix param name matching in base class method and derived class overridden method~ - #3023 
- ~String comparison errors -> String.Format, String.Equals, String.StartsWith... etc~ - #3031 
- ~using reserved keyword~ #3025
- ~Add missing constructors to exception classes~ #3026 
- ~CA1010: Collections should implement generic interface~ #3028 
- ~CA2201: Exception type System.Exception is not sufficiently specific~ #3029 
- ~CA1822: Mark members as static~ #3032

### DIFF
--- a/iothub/device/tests/Transport/Amqp/AmqpConnectionPoolTests.cs
+++ b/iothub/device/tests/Transport/Amqp/AmqpConnectionPoolTests.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.Devices.Client.Tests.Amqp
         {
             string sharedAccessKeyName = "HubOwner";
             uint poolSize = 10;
-            IConnectionCredentials testDevice = AmqpConnectionPoolTests.CreatePooledSasGroupedClientIdentity(sharedAccessKeyName);
+            IConnectionCredentials testDevice = CreatePooledSasGroupedClientIdentity(sharedAccessKeyName);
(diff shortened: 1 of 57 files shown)


schoims (approved):

timtay-microsoft (approved):

iothub/device/tests/Transport/Amqp/AmqpConnectionPoolTests.cs
```diff
@@ -38,7 +38,7 @@ public void AmqpConnectionPool_Add_Remove_ConnectionHolderIsRemoved()
         {
             string sharedAccessKeyName = "HubOwner";
             uint poolSize = 10;
-            IConnectionCredentials testDevice = CreatePooledSasGroupedClientIdentity(sharedAccessKeyName);
+            IConnectionCredentials testDevice = AmqpConnectionPoolTests.CreatePooledSasGroupedClientIdentity(sharedAccessKeyName);
```
ngastelum-ms, line 41:
```suggestion
            IConnectionCredentials testDevice = CreatePooledSasGroupedClientIdentity(sharedAccessKeyName);
```

ngastelum-ms (approved):

PR #3025, opened 2022-12-14
Fix reserved keyword error

ngastelum-ms:


iothub/service/src/Registry/ModulesClient.cs
```diff
@@ -61,18 +60,18 @@ protected ModulesClient()
         /// For a complete list of possible error cases, see <see cref="IotHubServiceErrorCode"/>.
         /// </exception>
         /// <exception cref="OperationCanceledException">If the provided cancellation token has requested cancellation.</exception>
-        public virtual async Task<Module> CreateAsync(Module module, CancellationToken cancellationToken = default)
+        public virtual async Task<Module> CreateAsync(Module moduleIdentity, CancellationToken cancellationToken = default)
```
abhipsaMisra, line 63: I like this renaming. I think we should check what we call the device counterpart - it might be worth calling this "deviceIdentity".

abhipsaMisra (approved):

PR #3027, opened 2022-12-14
Fix for Specify IFormatProvider errors

ngastelum-ms:
[CA1305: Specify IFormatProvider](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1305)

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttLogger.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttLogger.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics.Tracing;
+using System.Globalization;
 using MQTTnet.Diagnostics;
 
 namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
@@ -21,7 +22,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             string formattedMessage = message;
             if (parameters != null && parameters.Length > 0)
             {
-                formattedMessage = string.Format(message, parameters);
+                formattedMessage = string.Format(CultureInfo.CurrentCulture, message, parameters);
             }
 
             Logging.Log.Write(formattedMessage, options);
--- a/iothub/device/src/Transport/Mqtt/MqttLogger.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttLogger.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             string formattedMessage = message;
             if (parameters != null && parameters.Length > 0)
             {
-                formattedMessage = string.Format(CultureInfo.CurrentCulture, message, parameters);
+                formattedMessage = string.Format(CultureInfo.InvariantCulture, message, parameters);
             }
 
             Logging.Log.Write(formattedMessage, options);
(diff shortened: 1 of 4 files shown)


iothub/device/src/Transport/Mqtt/MqttLogger.cs
```diff
@@ -21,7 +22,7 @@ public void Publish(MqttNetLogLevel logLevel, string source, string message, obj
             string formattedMessage = message;
             if (parameters != null && parameters.Length > 0)
             {
-                formattedMessage = string.Format(message, parameters);
+                formattedMessage = string.Format(CultureInfo.CurrentCulture, message, parameters);
```
abhipsaMisra, line 25: I can see the point that a log statement should be formatted based on current culture; however, I wanted to get this confirmed from what the track 2 SDKs do.
Turns out, they format based on InvariantCulture, possibly to preserve the logging behavior across regions.

I'd recommend that we do the same.

https://github.com/Azure/azure-sdk-for-net/blob/e59553b4fb61b37c9c13681b4b117cc58de17e6d/sdk/core/Azure.Core/src/Diagnostics/AzureEventSourceListener.cs#L96

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1239,7 +1239,7 @@ private bool ParseResponseTopic(string topicName, out string rid, out int status
                 {
                     // This query string key-value pair is only expected in a successful patch twin response message.
                     // Get twin requests will contain the twin version in the payload instead.
-                    version = int.Parse(queryStringKeyValuePairs.Get(VersionTopicKey));
+                    version = int.Parse(queryStringKeyValuePairs.Get(VersionTopicKey), CultureInfo.CurrentCulture);
```
abhipsaMisra, line 1242: InvariantCulture

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -275,7 +275,7 @@ await _internalRetryHandler
                 // No need to deserialize here since the user will deserialize this into their expected type
                 // after this function returns.
                 string responsePayload = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
-                int responseStatusCode = int.Parse(response.Headers.GetValues(StatusCodeHeaderKey).FirstOrDefault());
+                int responseStatusCode = int.Parse(response.Headers.GetValues(StatusCodeHeaderKey).FirstOrDefault(), CultureInfo.CurrentCulture);
```
abhipsaMisra, line 278: InvariantCulture

iothub/service/src/DigitalTwin/DigitalTwinsClient.cs
```diff
@@ -364,7 +364,7 @@ await _internalRetryHandler
...
```
abhipsaMisra, line 367: InvariantCulture

provisioning/device/src/Transports/Mqtt/MqttLogger.cs
```diff
@@ -21,7 +22,7 @@ public void Publish(MqttNetLogLevel logLevel, string source, string message, obj
             string formattedMessage = message;
             if (parameters != null && parameters.Length > 0)
             {
-                formattedMessage = string.Format(message, parameters);
+                formattedMessage = string.Format(CultureInfo.CurrentCulture, message, parameters);
```
abhipsaMisra, line 25: InvariantCulture

iothub/device/src/Transport/Mqtt/MqttLogger.cs
ngastelum-ms: Makes sense, updated

PR #3028, opened 2022-12-14
Collections should implement generic interface

abhipsaMisra:


provisioning/service/src/Twin/InitialTwinPropertyCollection.cs
```diff
@@ -205,6 +205,12 @@ public IEnumerator GetEnumerator()
             }
         }
 
+        /// <inheritdoc/>
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
```
timtay-microsoft, line 211: Is this method just calling itself?

timtay-microsoft: And why do we need two "GetEnumerator()" functions in this class?

abhipsaMisra: It seems that way, isn't it!

It's calling https://github.com/Azure/azure-iot-sdk-csharp/blob/78138673c18535408c0321d7aab82e0acbce8f02/provisioning/service/src/Twin/InitialTwinPropertyCollection.cs#L195-L206

abhipsaMisra: We need this one because InitialTwinPropertyCollection an `IEnumerator GetEnumerator()` iterator.

abhipsaMisra: The interface `IEnumerable<T>` requires the method `IEnumerator<T> GetEnumerator();` to be implemented. At the same time, `IEnumerable<T>` implements `IEnumerable`, which requires the method `IEnumerator GetEnumerator();` to be implemented.

provisioning/service/src/Twin/InitialTwinPropertyCollection.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
     /// Represents a collection of properties for device twin.
     /// </summary>
     [JsonConverter(typeof(InitialTwinPropertiesJsonConverter))]
-    public class InitialTwinPropertyCollection : IEnumerable
+    public class InitialTwinPropertyCollection : IEnumerable<KeyValuePair<string, object>>
```
timtay-microsoft, line 16: We have similar classes in the 2 IoT hub clients as well. Do they need this change as well?

timtay-microsoft (approved):

PR #3029, opened 2022-12-14
Throw InvalidOperationException instead of System.Exception

abhipsaMisra:


schoims (approved):

iothub/service/src/Feedback/MessageFeedbackProcessorClient.cs
```diff
@@ -111,7 +111,7 @@ public virtual async Task OpenAsync(CancellationToken cancellationToken = defaul
             {
                 if (MessageFeedbackProcessor == null)
                 {
-                    throw new Exception("Callback for message feedback must be set before opening the connection.");
+                    throw new InvalidOperationException("Callback for message feedback must be set before opening the connection.");
```
brycewang-microsoft, line 114: Do we need to call out `InvalidOperationException` to be expected within the doc comments above?

brycewang-microsoft (approved):

iothub/service/src/Feedback/MessageFeedbackProcessorClient.cs
abhipsaMisra: good point, sure

PR #3035, opened 2022-12-15
IDisposable fixes

ngastelum-ms:
[CA1001: Types that own disposable fields should be disposable](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1001)

[CA1816: Call GC.SuppressFinalize correctly](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1816)

### DIFF
--- a/iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
+++ b/iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
@@ -10,7 +10,7 @@ using Microsoft.Azure.Devices.Client.Transport.AmqpIot;
 
 namespace Microsoft.Azure.Devices.Client.Transport.Amqp
 {
-    internal class AmqpAuthenticationRefresher : IAmqpAuthenticationRefresher
+    internal class AmqpAuthenticationRefresher : IAmqpAuthenticationRefresher, IDisposable
     {
         private static readonly string[] s_accessRightsStringArray = new[] { "DeviceConnect" };
         private readonly Uri _amqpEndpoint;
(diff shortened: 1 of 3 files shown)


iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -237,5 +237,10 @@ private static async Task<T> ReadAsAsync<T>(HttpContent content, CancellationTok
             using var jsonReader = new JsonTextReader(reader);
             return new JsonSerializer().Deserialize<T>(jsonReader);
         }
+
+        public void Dispose()
+        {
+            throw new NotImplementedException();
```
timtay-microsoft, line 243: The reason this class is flagged with CA1001 is because it creates a disposable object (HttpClient) and needs to dispose it. The expected implementation here is to call dispose on all the disposable instances this class owns, like:

```csharp
_httpClientObj?.Dispose();
```

iothub/device/src/Transport/Amqp/AmqpAuthenticationRefresher.cs
```diff
@@ -170,5 +170,10 @@ private static string CreateAmqpCbsAudience(IConnectionCredentials connectionCre
             // SAS tokens granted to a group SAS authenticated client will scoped to the IoT hub-level; for example, myHub.azure-devices.net
             return connectionCredentials.HostName;
         }
+
+        public void Dispose()
+        {
+            throw new NotImplementedException();
```
timtay-microsoft, line 176: Same comment here. You need to dispose the _loopCancellationTokenSource instance here, not throw NotImplementedException

iothub/device/src/Transport/Http/HttpClientHelper.cs
timtay-microsoft: Generally, ```NotImplementedException``` is reserved for cases where a developer hasn't written the body of the method yet but wants to be able to build/run tests that don't use this method.

We really wouldn't ever check in something like that since our codebase isn't that early on in its development.

ngastelum-ms: Makes sense, fixed

PR #3036, opened 2022-12-15
Create best practice solution sample using X.509 certificate authentication

schoims:
- Created best practice solution using X.509 certificate readme.md doc for .NET framework
- Created a script that generate test certificates, do proof-of-possession in DPS, and create a group enrollment in DPS.
- Created a script that cleans up a created device and group enrollment. 
- Updated parameters in DeviceReconnectionSample to take certificate-related parameters.
- Improve X509Sample with exception handling, sdk stack trace log. Include an example how to set a RetryPolicy.


provisioning/device/samples/getting started/X509Sample/ProvisioningDeviceClientSample.cs
```diff
@@ -16,76 +18,110 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Samples
...
+                _logger.LogWarning("Sample execution cancellation requested; will exit.");
+            };
+
+            _logger.LogInformation("Loading the certificate...");
             using X509Certificate2 certificate = LoadProvisioningCertificate();
             var security = new AuthenticationProviderX509(certificate);
 
-            Console.WriteLine($"Initializing the device provisioning client...");
+            // Perform retry up to 10 times every 2 seconds
+            s_clientOptions.RetryPolicy = new ProvisioningClientFixedDelayRetryPolicy(10, TimeSpan.FromSeconds(2));
```
andyk-ms, line 60: Should we have parameters for the retry policy values instead?

provisioning/device/samples/solutions/BestPracticeSampleX509/readme.md
```diff
@@ -0,0 +1,139 @@
...
+        dotnet build
+    ```
+
+5. Set the IoT hub hostname, device Id, device certificate password and path and run the sample.
+
+```powershell
+    dotnet run --h <assigned IoT hub hostname> --d <device Id> --p <your password> --n <path to device certificate pfx file>
+```
+
+![x509-bootsequence](media/bootsequence.png)
```
andyk-ms, line 117: auth flow and the boot sequence diagram seems to have different info. Let's double check these:
1) the password: is that send to the DPS or within the SDK?
2) do we need to separate the device process created by the users vs the SDK processes?
3) between #2 and #3, is there an additional process to establish connection with IoTHub?

provisioning/device/samples/solutions/BestPracticeSampleX509/readme.md
```diff
@@ -0,0 +1,139 @@
...
+The business needs of Contoso are:
+
+- Provision devices securely at scale.
+- Group devices by production line.
+- Collect data from devices.
+- Manage devices remotely (e.g., controlling the temperature of a thermostat).
+
+## Recommended Solution
+
+![solution](media/auth_flow_diagram.png)
```
andyk-ms, line 20: auth flow and the boot sequence diagram seems to have different info. Let's double check the items in the comments for the boot sequence. For the Auth flow, is it targeted to identify how the scripts are creating the authentication certs and uploads that will be used in the boot sequence?

Maybe using #A, B. C etc for authentication setup, and #1,2,3 for boot sequence?

provisioning/device/samples/getting started/X509Sample/ProvisioningDeviceClientSample.cs
andyk-ms: Make sense for getting started to use a simple fixed delay retry policy. But as an example, maybe instead of commenting "Perform retry frequency", it would be more insightful to comment there are more advanced exponential backoff retry policy in the SDK.

provisioning/device/samples/solutions/BestPracticeSampleX509/readme.md
```diff
@@ -0,0 +1,139 @@
...
+
+4. From terminal, navigate to the [device reconnection sample folder](</iothub/device/samples/how to guides/DeviceReconnectionSample>). Read [readme.md](</iothub/device/samples/how to guides/DeviceReconnectionSample/readme.md>) to learn how to initialize the device client, send device to cloud telemetry, receive cloud to device message, receive twin desired property update notifications, and update device twin's reported properties.
+
+    Open [DeviceReconnectionSample.cs](</iothub/device/samples/how to guides/DeviceReconnectionSample/DeviceReconnectionSample.cs>) in code editor. Find
+
+    ```csharp
+    s_deviceClient = new IotHubDeviceClient(_deviceConnectionStrings.First(), _clientOptions);
+    ```
+
+    and replace it with
```
andyk-ms, line 97: What about having these part of the code within the other sample with retry policy in a block or a function, and this replace take that complete section for completeness and 'best practice' implementation?

provisioning/device/samples/getting started/X509Sample/ProvisioningDeviceClientSample.cs
schoims: Sure, I can make them configurable.

provisioning/device/samples/solutions/BestPracticeSampleX509/readme.md
schoims: 1. Certificate password is used within the SDK, not sent to DPS. I will make an update in the diagram.
2. X509Sample + DeviceReconnection samples combined is the device process. 
3. Nope.

provisioning/device/samples/getting started/X509Sample/ProvisioningDeviceClientSample.cs
schoims: Exponential backoff retry policy is our default retry policy. Should I use exponential backoff retry policy instead?

provisioning/device/samples/solutions/BestPracticeSampleX509/readme.md
schoims: Sure, sounds good!

provisioning/device/samples/getting started/X509Sample/ProvisioningDeviceClientSample.cs
andyk-ms: I would think there are 2 approaches:
1) use fixed retry policy (as custom) to show how to do custom, but note that this is to show how to do custom retry and not recommending fix delay retry for production.
2) use default exp retry policy to show best practice, but not that customer can create custom policy to accommodate specific needs.

Either way, I was thinking there should be additional comment to explain this in the sample.

provisioning/device/samples/solutions/BestPracticeSampleX509/readme.md
andyk-ms: Agree on the #1 for the diagram update.
For #2 and #3 points, there are 2 different clients in the SDK(s) that we have used, maybe we just need to have 2 different box to indicate a DPS client and a IoTHub client?

andyk-ms (approved):
discussed about the suggestion in details and assume they are updated.

patilsnr (approved):

PR #3045, opened 2022-12-27
Setting MaxDepth parameter in the JsonSerializerSettings as safe defaults

schoims:
Reference: https://github.com/advisories/GHSA-5crp-9r3c-p9vr 
- Setting the MaxDepth parameter to 128 in the JsonSerializerSettings in clients

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -12,6 +12,7 @@ using Microsoft.Azure.Devices.Shared;
 using System.Security.Cryptography.X509Certificates;
 using System.IO;
 using Microsoft.Azure.Devices.Client.Exceptions;
+using Newtonsoft.Json;
 
 #if NET451
 
@@ -155,6 +156,8 @@ namespace Microsoft.Azure.Devices.Client
             {
                 Logging.Enter(this, transportSettings, pipelineBuilder, nameof(InternalClient) + "_ctor");
             }
+
+            JsonConvert.DefaultSettings = JsonSerializerSettingsInitializer.GetDefaultJsonSerializerSettings();
 
             TlsVersions.Instance.SetLegacyAcceptableVersions();
 
@@ -989,7 +992,7 @@ namespace Microsoft.Azure.Devices.Client
                     methodResponseInternal?.Dispose();
                 }
                 finally
-                { 
+                {
                     // Need to release this semaphore even if the above dispose call fails
                     _methodsDictionarySemaphore.Release();
                 }
@@ -1405,7 +1408,7 @@ namespace Microsoft.Azure.Devices.Client
                 return;
             }
 
-            // Grab this semaphore so that there is no chance that the _deviceReceiveMessageCallback instance is set in between the read of the 
+            // Grab this semaphore so that there is no chance that the _deviceReceiveMessageCallback instance is set in between the read of the
             // item1 and the read of the item2
             await _deviceReceiveMessageSemaphore.WaitAsync().ConfigureAwait(false);
             ReceiveMessageCallback callback = _deviceReceiveMessageCallback?.Item1;
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -157,7 +157,7 @@ namespace Microsoft.Azure.Devices.Client
                 Logging.Enter(this, transportSettings, pipelineBuilder, nameof(InternalClient) + "_ctor");
             }
 
-            JsonConvert.DefaultSettings = JsonSerializerSettingsInitializer.GetDefaultJsonSerializerSettings();
+            JsonConvert.DefaultSettings = JsonSerializerSettingsInitializer.GetDefaultJsonSerializerSettingsDelegate();
 
             TlsVersions.Instance.SetLegacyAcceptableVersions();
 
--- a/iothub/service/tests/QueryTests.cs
+++ b/iothub/service/tests/QueryTests.cs
@@ -297,26 +297,27 @@ namespace Microsoft.Azure.Devices.Api.Test
         [TestMethod]
         public void QueryResult_OverrideDefaultJsonSerializer_ExceedMaxDepthThrows()
         {
-            // simulate json serialize/deserialize
-            var serverQueryResult = new QueryResult
-            {
-                Type = QueryResultType.Twin,
-                Items = new List<Twin>
-                {
-                    new Twin
-                    {
-                        DeviceId = "test",
-                    }
-                },
-                ContinuationToken = "GYUVJDBJFKJ"
-            };
-
             var settings = new JsonSerializerSettings { MaxDepth = 2 };
-            // serialize
-            var jsonQueryResult = JsonConvert.SerializeObject(serverQueryResult, settings);
-            jsonQueryResult.Should().Be("{\"type\":\"twin\",\"items\":[{\"deviceId\":\"test\",\"etag\":null,\"version\":null,\"properties\":{\"desired\":{},\"reported\":{}}}],\"continuationToken\":\"GYUVJDBJFKJ\"}");            // deserialize
+            // simulate json deserialize
+            const string jsonString = @"
+{
+""type"":""twin"",
+""items"":
+    [{
+        ""deviceId"": ""test"",
+        ""etag"":null,
+        ""version"":null,
+        ""properties"":
+            {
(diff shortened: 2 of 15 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -156,6 +157,8 @@ public InternalClient(IotHubConnectionString iotHubConnectionString, ITransportS
                 Logging.Enter(this, transportSettings, pipelineBuilder, nameof(InternalClient) + "_ctor");
             }
 
+            JsonConvert.DefaultSettings = () => new JsonSerializerSettings { MaxDepth = 128 };
```
abhipsaMisra, line 160: You'll want to add some tests that validate that this value can be overridden by the user.

drwill-ms, line 160: Every time this is access, a new settings object will be created. We'd rather access a shared instance.

If we're lucky, Newtonsoft will get it and save the instance, but I wouldn't count on it.

drwill-ms: https://github.com/JamesNK/Newtonsoft.Json/issues/78#issuecomment-17460093

He made it a delegate intentionally. The case he mentions may not matter to us, though.

schoims: I updated it to use a shared instance of JsonSerializerSettings. Could you take a look at the implementation again? Thank you in advance!

schoims: I added a few test cases. Could you take a look at them? Thank you in advance!

iothub/service/tests/QueryTests.cs
```diff
@@ -292,5 +293,59 @@ public void QueryResultUserSuppliedContinuationTest()
             Assert.AreEqual("test", r.ElementAt(0).DeviceId);
             Assert.IsTrue(q.HasMoreResults);
         }
+
+        [TestMethod]
+        public void QueryResult_OverrideDefaultJsonSerializer_Ok()
+        {
+            // simulate json serialize/deserialize
+            var serverQueryResult = new QueryResult()
```
drwill-ms, line 301:
```suggestion
            var serverQueryResult = new QueryResult
```

iothub/service/tests/QueryTests.cs
```diff
@@ -292,5 +293,59 @@ public void QueryResultUserSuppliedContinuationTest()
...
+            {
+                Type = QueryResultType.Twin,
+                Items = new List<Twin>()
```
drwill-ms, line 304:
```suggestion
                Items = new List<Twin>
```

iothub/service/tests/QueryTests.cs
```diff
@@ -292,5 +293,59 @@ public void QueryResultUserSuppliedContinuationTest()
...
+                {
+                    new Twin()
```
drwill-ms, line 306:
```suggestion
                    new Twin
```

iothub/service/tests/QueryTests.cs
```diff
@@ -292,5 +293,59 @@ public void QueryResultUserSuppliedContinuationTest()
...
+                        DeviceId = "test",
+                    }
+                },
+                ContinuationToken = "GYUVJDBJFKJ"
+            };
+
+            var settings = new JsonSerializerSettings { MaxDepth = 3 };
+            // serialize
+            var jsonQueryResult = JsonConvert.SerializeObject(serverQueryResult, settings);
+            Assert.AreEqual("{\"type\":\"twin\",\"items\":[{\"deviceId\":\"test\",\"etag\":null,\"version\":null,\"properties\":{\"desired\":{},\"reported\":{}}}],\"continuationToken\":\"GYUVJDBJFKJ\"}", jsonQueryResult);
```
drwill-ms, line 317: Use FluentAssertions instead.

```suggestion
            jsonQueryResult.Should().Be("{\"type\":\"twin\",\"items\":[{\"deviceId\":\"test\",\"etag\":null,\"version\":null,\"properties\":{\"desired\":{},\"reported\":{}}}],\"continuationToken\":\"GYUVJDBJFKJ\"}");
```

iothub/service/tests/SerializationTests.cs
```diff
@@ -68,5 +69,53 @@ public void Configuration_TestNullSchema_Ok()
...
+ ""connectionState"": ""Connected"",
+ ""lastActivityTime"": ""2018-06-29T21:17:08.7759733"",
+}";
+
+            var settings = new JsonSerializerSettings { MaxDepth = 2 };
+            var twin = JsonConvert.DeserializeObject<Twin>(jsonString, settings);
+
+            Assert.IsNotNull(twin);
+            Assert.AreEqual(twin.DeviceId, "test");
+            Assert.AreEqual(twin.ETag, "AAAAAAAAAAM=");
```
drwill-ms, line 92:
```suggestion
            twin.Should().NotBeNull();
            twin.DeviceId.Should().Be("test");
            twin.ETag.Should().Be("AAAAAAAAAAM=");
```

drwill-ms (approved):

iothub/service/tests/QueryTests.cs
```diff
@@ -292,5 +293,58 @@ public void QueryResultUserSuppliedContinuationTest()
...
+                {
+                    new Twin
+                    {
+                        DeviceId = "test",
+                    }
+                },
+                ContinuationToken = "GYUVJDBJFKJ"
+            };
+
+            var settings = new JsonSerializerSettings { MaxDepth = 3 };
```
abhipsaMisra, line 314: I understand the idea behind this test, but I think it might be superfluous. Since our default value is greater than 3, this test doesn't conclusively assert that the default did get overridden. 
We'd want to use a value greater than our default value in order to test this scenario.

Having said that, I'd be ok with us removing this test altogether and only testing the assertion via the "ExceedMaxDepthThrows" test case. 
The "ExceedMaxDepthThrows" test case also tests that the default is overridden, it does it with a much lighter string value, and it does that more conclusively.

iothub/service/tests/QueryTests.cs
```diff
@@ -292,5 +293,58 @@ public void QueryResultUserSuppliedContinuationTest()
...
+                    {
+                        DeviceId = "test",
+                    }
+                },
+                ContinuationToken = "GYUVJDBJFKJ"
+            };
+
+            var settings = new JsonSerializerSettings { MaxDepth = 2 };
+            // serialize
+            var jsonQueryResult = JsonConvert.SerializeObject(serverQueryResult, settings);
```
abhipsaMisra, line 341: Isn't the serialize call expected to throw an exception as well?

iothub/service/tests/Registry/DevicesClientTests.cs
```diff
@@ -0,0 +1,1176 @@
...
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using static System.Net.WebRequestMethods;
+
+namespace Microsoft.Azure.Devices.Tests
+{
+    [TestClass]
+    [TestCategory("Unit")]
+    public class DevicesClientTests
```
abhipsaMisra, line 22: Is this a new test class that was added?

iothub/service/tests/SerializationTests.cs
```diff
@@ -68,5 +69,53 @@ public void Configuration_TestNullSchema_Ok()
 
             JsonConvert.DeserializeObject<Configuration>(jsonString);
         }
+
+        [TestMethod]
+        public void Twin_Json_OverrideDefaultJsonSerializer_Ok()
```
abhipsaMisra, line 74: You could remove this test case as well.

iothub/service/tests/QueryTests.cs
schoims: Looks like the max depth limit is only enforced when reading JSON data, not when writing it.

abhipsaMisra: You're right about that: https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializerSettings_MaxDepth.htm
"Gets or sets the maximum depth allowed when reading JSON. Reading past this depth will throw a [JsonReaderException](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonReaderException.htm)."

You can remove the serialization bit from the test case.

schoims: Sure!

PR #3048, opened 2023-01-03
Provisioning device unit tests coverage - part 1

brycewang-microsoft:


provisioning/device/src/Utilities/Logging.cs
```diff
@@ -4,13 +4,15 @@
...
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Tracing;
 using System.Globalization;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 namespace Microsoft.Azure.Devices.Provisioning.Client
 {
+    [ExcludeFromCodeCoverage]
```
abhipsaMisra, line 15: 👍🏽

provisioning/device/src/Utilities/Argument.cs
```diff
@@ -53,19 +53,20 @@ public static void AssertNotNullOrWhiteSpace(string value, string name)
         /// <summary>
         /// Checks an argument to ensure that its value isn't negative.
         /// </summary>
-        /// <param name="argumentValue">The value of the argument.</param>
-        /// <param name="argumentName">The name of the argument for diagnostic purposes.</param>
-        internal static void AssertNotNegativeValue<T>(T argumentValue, string argumentName)
+        /// <param name="value">The value of the argument.</param>
+        /// <param name="name">The name of the argument for diagnostic purposes.</param>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="value"/> is negative.</exception>
+        internal static void AssertNotNegativeValue<T>(T value, string name)
```
abhipsaMisra, line 59: I like what you've changed it to but I'm curious - did the tool highlight this change or was it personal preference?

provisioning/device/src/Utilities/ProductInfo.cs
```diff
@@ -1,11 +1,13 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using System.Runtime.InteropServices;
 
 namespace Microsoft.Azure.Devices.Provisioning.Client
 {
+    [ExcludeFromCodeCoverage]
```
abhipsaMisra, line 10: Just to be thorough., these could use a comment explaining why they were excluded from code coverage analysis.

provisioning/device/tests/RetryPolicies/ProvisioningClientExponentialBackoffRetryPolicyTests.cs
```diff
@@ -50,5 +50,18 @@ public void ExponentialBackoffRetryPolicy_IsExponential(uint retryCount)
...
+        [DataRow(true)]
+        [DataRow(false)]
+        public void ExponentialBackoffRetryPolicy_ProvisioningClientException_ReturnsTrueWhenTransient(bool isTransient)
+        {
+            // arrange
+            var retryPolicy = new ProvisioningClientExponentialBackoffRetryPolicy(0, TimeSpan.FromMinutes(100), false);
+            var ex = new ProvisioningClientException("", isTransient);
+
+            // act - assert
+            retryPolicy.ShouldRetry(1, ex, out _).Should().Be(isTransient);
```
abhipsaMisra, line 64: Since these retry policy tests would be needed for all clients, maybe someone could take the lead and make sure we're covering all scenarios for all of our clients.

provisioning/device/tests/mqtt/ProvisioningErrorDetailsMqttTests.cs
```diff
@@ -19,55 +19,55 @@ public class ProvisioningErrorDetailsMqttTests
         private static readonly TimeSpan s_defaultInterval = TimeSpan.FromSeconds(2);
 
         [TestMethod]
-        public void TestRetryAfterValidThrottled()
+        public void Test_RetryAfter_ValidThrottled()
```
abhipsaMisra, line 22: We probably don't need the "test" prefix here, right?

abhipsaMisra (approved):

provisioning/device/src/Utilities/Argument.cs
brycewang-microsoft: It would be great if the tools are capable to highlight this, but unfortunately no, I had to manually align and update it.

provisioning/device/tests/RetryPolicies/ProvisioningClientExponentialBackoffRetryPolicyTests.cs
brycewang-microsoft: Yeah, sure, I would love to take care of that!

provisioning/device/tests/ArgumentTests.cs
```diff
@@ -0,0 +1,70 @@
...
+    [TestClass]
+    [TestCategory("Unit")]
+    public class ArgumentTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void AssertNotNull_IfNull_Throws()
+        {
+            string argument = null;
+            Argument.AssertNotNull<string>(argument, "");
```
schoims, line 19: Just curious why we are not using FluentAssertions?

schoims (approved):

provisioning/device/tests/ArgumentTests.cs
ngastelum-ms: Curious about this too. We have both ways of doing it present in our tests, but is this something where we should be sticking with one style?

drwill-ms: Argument.AssertNotNull is not a unit test assertion framework. We use that for validating parameters in our public API.

provisioning/device/src/TransportSettings/ProvisioningClientTransportSettings.cs
```diff
@@ -90,6 +90,7 @@ public abstract class ProvisioningClientTransportSettings
 
         // The default remote certificate validation callback. It returns false if any SSL level exceptions occurred
         // during the handshake.
+        [ExcludeFromCodeCoverage]
```
drwill-ms, line 93: Why exclude?

provisioning/device/src/Utilities/Logging.cs
drwill-ms: We should be really careful about excluding things from cc. I'd like to see it called out explicitly and a case made in a code comment about why it is not needed.

We talked about one in an earlier PR for a logging class, which I agree we can omit from cc, so I'm fine with this one too.

Still needs a comment. I'd suggest we cover exclusions in stand up with a parking lot convo.

provisioning/device/tests/ArgumentTests.cs
drwill-ms, line 19:
```suggestion
        public void AssertNotNull_IfNull_Throws()
        {
            Action act = () => Argument.AssertNotNull<string>(null, "");
            act.Should().Throw<ArgumentNullException>();
```

provisioning/device/tests/ProvisioningTransportRegisterRequestTests.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.UnitTests
+{
+    [TestClass]
+    [TestCategory("Unit")]
+    public class ProvisioningTransportRegisterRequestTests
+    {
+
```
drwill-ms, line 19:
```suggestion
    {
```

drwill-ms (approved):

provisioning/device/src/TransportSettings/ProvisioningClientTransportSettings.cs
drwill-ms: Why private methods? They include code that is exercised as part of our SDK running, and could have bugs.

We can't unit test it a private method but we always have the option of making it internal or pulling out business logic from a class and making it an internal class that can be exercised independently.

brycewang-microsoft: That's true

PR #3050, opened 2023-01-03
Add test for recovery after quota being exceeded

abhipsaMisra:
Test to validate that twin operations recover after client recovery on exceeding quotas over AMQP.

Linked workitem (internal): https://msazure.visualstudio.com/One/_workitems/edit/9956630

### DIFF
--- a/e2e/test/iothub/device/TwinFaultInjectionTests.cs
+++ b/e2e/test/iothub/device/TwinFaultInjectionTests.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Twins
         [Timeout(TestTimeoutMilliseconds)]
         public async Task Twin_DeviceReportedPropertiesTcpConnRecovery_Mqtt()
         {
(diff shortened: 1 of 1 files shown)


e2e/test/iothub/device/TwinFaultInjectionTests.cs
```diff
@@ -235,6 +282,7 @@ async Task InitAsync(IotHubDeviceClient deviceClient, TestDevice testDevice)
 
             async Task TestOperationAsync(IotHubDeviceClient deviceClient, TestDevice testDevice)
             {
+                string propName = Guid.NewGuid().ToString();
```
abhipsaMisra, line 285: https://github.com/Azure/azure-iot-sdk-csharp/blob/69f53aa08d8ee723cc3cef92415c373c01161671/iothub/device/src/Twin/ReportedProperties.cs#L51

abhipsaMisra: Property key now needs to be unique

drwill-ms (approved):

PR #3054, opened 2023-01-05
JSON-related changes for enums and required property

drwill-ms:


iothub/service/src/Messaging/Models/PurgeMessageQueueResult.cs
```diff
@@ -23,13 +23,13 @@ protected internal PurgeMessageQueueResult()
         /// <summary>
         /// The Id of the device whose messages are being purged.
         /// </summary>
-        [JsonProperty("deviceId", Required = Required.Always)]
+        [JsonProperty("deviceId")]
```
drwill-ms, line 26: We don't need to validate items are in the payload when de/serializing. We can trust the service included it, or if the user didn't include a required item, the service should return an error.

iothub/service/src/Registry/Models/AuthenticationMechanism.cs
```diff
@@ -61,7 +60,6 @@ public X509Thumbprint X509Thumbprint
         /// <summary>
         /// Gets or sets the authentication type.
         /// </summary>
-        [DefaultValue(ClientAuthenticationType.Sas)]
```
drwill-ms, line 64: Not sure this even worked, but we don't want SAS to be the default, anyway.

provisioning/service/src/Models/CustomAllocationDefinition.cs
```diff
@@ -13,15 +14,13 @@ public class CustomAllocationDefinition
         /// <summary>
         /// The webhook URL used for allocation requests.
         /// </summary>
-        [JsonProperty("webhookUrl", Required = Required.Always)]
-#pragma warning disable CA1056 // Uri properties should not be strings
-        public string WebhookUrl { get; set; }
-#pragma warning restore CA1056 // Uri properties should not be strings
+        [JsonProperty("webhookUrl")]
+        public Uri WebhookUrl { get; set; }
```
drwill-ms, line 18: Made this a URI and got rid of the suppression.

timtay-microsoft (approved):

abhipsaMisra (approved):

andyk-ms (approved):

schoims (approved):

PR #3055, opened 2023-01-06
fix docker version config and enable invalid cert tests

andyk-ms:
Pipeline agent now include Azure specific docker CLI package and no need to install on the go
- Removed script for Docker CLI setup
- Undo commented out scripts for the invalid certificate tests
- Revised and clean up the docker setup script

### DIFF
--- a/vsts/build-test-nuget.yml
+++ b/vsts/build-test-nuget.yml
@@ -299,11 +299,6 @@ jobs:
           command: "login"
           containerRegistry: "Azure IoT ACR"
 
-      - task: DockerInstaller@0
-        displayName: "Cert Validation - Install Docker CLI"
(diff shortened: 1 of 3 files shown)


vsts/build-test-nuget.yml
```diff
@@ -299,11 +299,6 @@ jobs:
           command: "login"
           containerRegistry: "Azure IoT ACR"
 
-      - task: DockerInstaller@0
```
andyk-ms, line 302: This is the part caused the conflicts on the script. No longer need to install this as it is already configured in the latest agent image.

timtay-microsoft (approved):

schoims (approved):

PR #3057, opened 2023-01-09
Misc code cleanup

drwill-ms:


timtay-microsoft (approved):

iothub/device/src/Transport/Amqp/ClientWebSocketTransport.cs
```diff
@@ -281,7 +280,7 @@ private static bool ReadTaskDone(Task<int> taskResult, TransportAsyncCallbackArg
                 args.CompletedSynchronously = ((IAsyncResult)taskResult).CompletedSynchronously;
                 return true;
             }
-            else if (taskResult.IsCanceled) // This should not happen since TaskCanceledException is handled in ReadAsyncCore.
+            else if (taskResult.IsCanceled)
```
andyk-ms, line 283: If this should not happen, should we log this if it ever get here?

abhipsaMisra, line 283: Not moving this to before the "completed" block?

drwill-ms: The comment was out of date so I just removed it.

The code block will never be hit for another reason because if it is canceled it is also completed, and in the previous block observing taskResult.Result will throw.

We could choose to remove this block all together. Thoughts?

drwill-ms: No, because otherwise the user wouldn't get an exception that the task was canceled.

andyk-ms: I was thinking this: if it is a dead branch of code that should never reach, it should not exist. But if it exists, that is unexpected and should have warning or logging of such unexpected execution. Your new commit removing it is fine.

PR #3059, opened 2023-01-10
Fix DPS registration over HTTP with escaped characters in the registr…

drwill-ms:
…ation Id

### DIFF
--- a/provisioning/transport/http/src/Generated/RuntimeRegistration.cs
+++ b/provisioning/transport/http/src/Generated/RuntimeRegistration.cs
@@ -1,11 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.Azure.Devices.Provisioning.Client.Transport.Models;
-using Microsoft.Azure.Devices.Shared;
-using Microsoft.Rest;
-using Microsoft.Rest.Serialization;
-using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
@@ -15,14 +10,18 @@ using System.Net.Http.Headers;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.Devices.Provisioning.Client.Transport.Models;
+using Microsoft.Azure.Devices.Shared;
+using Microsoft.Rest;
+using Microsoft.Rest.Serialization;
+using Newtonsoft.Json;
 
 namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
 {
     /// <summary>
     /// RuntimeRegistration operations.
     /// </summary>
-    internal partial class RuntimeRegistration
-        : IServiceOperations<DeviceProvisioningServiceRuntimeClient>, IRuntimeRegistration
+    internal partial class RuntimeRegistration : IServiceOperations<DeviceProvisioningServiceRuntimeClient>, IRuntimeRegistration
     {
         /// <summary>
         /// Initializes a new instance of the RuntimeRegistration class.
@@ -80,7 +79,7 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
             string operationId,
             string idScope,
             Dictionary<string, List<string>> customHeaders = null,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = default)
         {
             if (registrationId == null)
             {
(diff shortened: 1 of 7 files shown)


provisioning/transport/http/src/Generated/RuntimeRegistration.cs
```diff
@@ -1,11 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Microsoft.Azure.Devices.Provisioning.Client.Transport.Models;
```
drwill-ms, line 4: Most of these changes are just code cleanup (like variable renames, etc.) but I'll point out more interesting changes.

Nothing is necessary for this bug fix, though.

provisioning/transport/http/src/Generated/RuntimeRegistration.cs
```diff
@@ -94,149 +93,133 @@ public RuntimeRegistration(DeviceProvisioningServiceRuntimeClient client)
...
-            _url = _url.Replace("{idScope}", Uri.EscapeDataString(idScope), StringComparison.Ordinal);
-#else
-            _url = _url.Replace("{registrationId}", Uri.EscapeDataString(registrationId));
-            _url = _url.Replace("{operationId}", Uri.EscapeDataString(operationId));
-            _url = _url.Replace("{idScope}", Uri.EscapeDataString(idScope));
-#endif
+            string baseUrl = Client.BaseUri.AbsoluteUri;
+            string url = new Uri(
+                new Uri(baseUrl + (baseUrl.EndsWith("/", StringComparison.Ordinal) ? "" : "/")),
+                $"{Uri.EscapeDataString(idScope)}/registrations/{Uri.EscapeDataString(registrationId)}/operations/{Uri.EscapeDataString(operationId)}").ToString();
```
drwill-ms, line 118: We can do this inline with string interpolation instead of string replacing. Odd original choice.

provisioning/transport/http/src/SymmetricKeyCredentials.cs
```diff
@@ -12,30 +13,35 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
 {
     internal class SymmetricKeyCredentials : ServiceClientCredentials
     {
-        private const string SASHeaderName = "SharedAccessSignature";
+        private const string SasHeaderName = "SharedAccessSignature";
         private const string Registration = "registration";
-        private readonly string SymmetricKey;
-        private volatile string _sasToken;
```
drwill-ms, line 18: No reason to cache it if we don't read from the cache.

provisioning/transport/http/src/Generated/RuntimeRegistration.cs
```diff
@@ -484,135 +456,121 @@ public RuntimeRegistration(DeviceProvisioningServiceRuntimeClient client)
...
-                ServiceClientTracing.Enter(_invocationId, this, "RegisterDevice", tracingParameters);
+                ServiceClientTracing.Enter(invocationId, this, "RegisterDevice", tracingParameters);
             }
+
             // Construct URL
             string baseUrl = Client.BaseUri.AbsoluteUri;
             string url = new Uri(
                     new Uri(baseUrl + (baseUrl.EndsWith("/", StringComparison.Ordinal) ? "" : "/")),
-                    "{idScope}/registrations/{registrationId}/register")
+                    $"{WebUtility.UrlEncode(idScope)}/registrations/{WebUtility.UrlEncode(registrationId)}/register")
```
timtay-microsoft, line 466: I'm assuming this is the bug fix, right? Everything else is just code style fixes?

timtay-microsoft: And why do we need to URL encode the id scope? It is alphanumeric, right?

timtay-microsoft (approved):

provisioning/transport/http/src/SymmetricKeyCredentials.cs
```diff
@@ -12,30 +13,35 @@ namespace Microsoft.Azure.Devices.Provisioning.Client.Transport
...
         {
             string audience = request.RequestUri.AbsolutePath.Trim('/');
             string[] segments = audience.Split('/');
 
-            _sasToken = ProvisioningSasBuilder.BuildSasSignature(Registration, this.SymmetricKey, string.Concat(segments[0], '/', segments[1], '/', segments[2]), TimeSpan.FromDays(1));
-            SetAuthorizationHeader(request, _sasToken);
+            string sasToken = ProvisioningSasBuilder.BuildSasSignature(
+                Registration,
+                _symmetricKey,
+                $"{WebUtility.UrlDecode(segments[0])}/{WebUtility.UrlDecode(segments[1])}/{WebUtility.UrlDecode(segments[2])}",
```
abhipsaMisra, line 35: nit: could use a comment here

abhipsaMisra (approved):

provisioning/transport/http/src/Generated/RuntimeRegistration.cs
drwill-ms: I retained the behavior of encoding it. I agree we could make the assumption that ID scopes don't include escapable characters, but that doesn't mean a customer couldn't pass bad data. I figured it would be better to continue to encode it (and decode for this path) so the error the user gets is perhaps more meaningful.

PR #2970, opened 2022-11-17
Update DotNetty versions

schoims:

PR #3060, opened 2023-01-10
Bump versions of remaining packages since shared package was updated

schoims:
## Microsoft.Azure.Devices.Client 1.36.7
- Update DotNetty versions (https://github.com/Azure/azure-iot-sdk-csharp/pull/2970)
- Bring amqp sas token refresh cleanup logic to LTS (https://github.com/Azure/azure-iot-sdk-csharp/pull/3039)
- Setting MaxDepth parameter in the JsonSerializerSettings as safe default #3045
- Updated reference to `Microsoft.Azure.Devices.Shared` nuget

## Microsoft.Azure.Devices  1.31.4
- Setting MaxDepth parameter in the JsonSerializerSettings as safe default #3045
- Updated reference to `Microsoft.Azure.Devices.Shared` nuget

## Microsoft.Azure.Devices.Shared 1.27.2
- Setting MaxDepth parameter in the JsonSerializerSettings as safe default #3045

## Microsoft.Azure.Devices.Provisioning.Transport.Mqtt 1.14.3
- Update DotNetty versions (https://github.com/Azure/azure-iot-sdk-csharp/pull/2970)
- Add M2M message failure fix (https://github.com/Azure/azure-iot-sdk-csharp/pull/3006)
- Updated reference to `Microsoft.Azure.Devices.Shared` nuget

## Microsoft.Azure.Devices.Provisioning.Service 1.16.5
- Setting MaxDepth parameter in the JsonSerializerSettings as safe default #3045
- Updated reference to `Microsoft.Azure.Devices.Shared` nuget

## Microsoft.Azure.Devices.Provisioning.Client 1.16.5
- Updated reference to `Microsoft.Azure.Devices.Shared` nuget

## Microsoft.Azure.Devices.Provisioning.Transport.Amqp 1.13.8
- Updated reference to `Microsoft.Azure.Devices.Shared` nuget

## Microsoft.Azure.Devices.Provisioning.Transport.Http 1.12.5
- Updated reference to `Microsoft.Azure.Devices.Shared` nuget
- Fix DPS device client group enrollment provisioning over symmetric key when registration Id contains invalid URI characters (#3061)

## Microsoft.Azure.Devices.Provisioning.Security.Tpm 1.12.5
- Updated reference to `Microsoft.Azure.Devices.Shared` nuget

### DIFF
--- a/shared/src/Microsoft.Azure.Devices.Shared.csproj
+++ b/shared/src/Microsoft.Azure.Devices.Shared.csproj
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.27.2</Version>
+    <Version>1.27.3</Version>
     <Title>Microsoft Azure IoT Devices Shared</Title>
     <IncludeSource>True</IncludeSource>
     <IncludeSymbols>True</IncludeSymbols>
--- a/shared/src/Microsoft.Azure.Devices.Shared.csproj
+++ b/shared/src/Microsoft.Azure.Devices.Shared.csproj
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.27.3</Version>
+    <Version>1.27.2</Version>
     <Title>Microsoft Azure IoT Devices Shared</Title>
     <IncludeSource>True</IncludeSource>
     <IncludeSymbols>True</IncludeSymbols>
--- a/versions.csv
+++ b/versions.csv
@@ -1,10 +1,10 @@
 AssemblyPath, Version
 iothub\device\src\Microsoft.Azure.Devices.Client.csproj, 1.36.7
 iothub\service\src\Microsoft.Azure.Devices.csproj, 1.31.4
-shared\src\Microsoft.Azure.Devices.Shared.csproj, 1.27.2
-provisioning\device\src\Microsoft.Azure.Devices.Provisioning.Client.csproj, 1.16.4
-provisioning\transport\amqp\src\Microsoft.Azure.Devices.Provisioning.Transport.Amqp.csproj, 1.13.7
+shared\src\Microsoft.Azure.Devices.Shared.csproj, 1.27.3
+provisioning\device\src\Microsoft.Azure.Devices.Provisioning.Client.csproj, 1.16.5
+provisioning\transport\amqp\src\Microsoft.Azure.Devices.Provisioning.Transport.Amqp.csproj, 1.13.8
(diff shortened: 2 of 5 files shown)


versions.csv
```diff
@@ -1,10 +1,10 @@
 AssemblyPath, Version
 iothub\device\src\Microsoft.Azure.Devices.Client.csproj, 1.36.7
 iothub\service\src\Microsoft.Azure.Devices.csproj, 1.31.4
-shared\src\Microsoft.Azure.Devices.Shared.csproj, 1.27.2
-provisioning\device\src\Microsoft.Azure.Devices.Provisioning.Client.csproj, 1.16.4
-provisioning\transport\amqp\src\Microsoft.Azure.Devices.Provisioning.Transport.Amqp.csproj, 1.13.7
-provisioning\transport\http\src\Microsoft.Azure.Devices.Provisioning.Transport.Http.csproj, 1.12.4
+shared\src\Microsoft.Azure.Devices.Shared.csproj, 1.31.0
```
abhipsaMisra, line 4: I know this is adding an API and therefore should be a minor version upgrade, but I think there might be some value in retaining this as a patch version upgrade. LTS patches are supposed to be security/critical bug fixes and thus customers wouldn't generally be expecting new APIs to be added.
For this Newtonsoft fix, a case could be made that even though a public API was added, we don't expect users to call that API. (Usage-wise, it is an `internal` API but has been declared as `public` since it will be referenced in other packages).
Bumping the minor version might confuse some users into expecting (or being wary of) additional new APIs.

This is just a thought I had and I'm curious to know what others think.

abhipsaMisra (approved):

versions.csv
schoims: That sounds reasonable! I have no problem with that either.

drwill-ms: Yes, patch version only for LTS.

shared/src/Microsoft.Azure.Devices.Shared.csproj
```diff
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.27.2</Version>
+    <Version>1.31.0</Version>
```
drwill-ms, line 21: Agreed, this should be 1.27.3 because of how we handle versioning with LTS.

drwill-ms (approved):

shared/src/Microsoft.Azure.Devices.Shared.csproj
schoims: yup, I will update to the next patch version which is 1.27.2 since this has been already updated in the previous pr

PR #3066, opened 2023-01-12
Fix ETag handling for serialization into http request header and body

abhipsaMisra:
Azure.Core ETag is formatted differently based on if it is to be placed into an HTTP request header or into the request body - https://github.com/Azure/azure-sdk-for-net/blob/9c6238e0f0dd403d6583b56ec7902c77c64a2e37/sdk/core/Azure.Core/src/ETag.cs#L87-L114
This PR uses the logic defined in Azure.Core to perform the formatting.

This currently works for all scenarios other than `ETag.All` -> "*". I have a [question](https://github.com/Azure/azure-sdk-for-net/issues/33435) open with the Azure.Core team to confirm the expected behavior for `ETag.All`.

image

With the workaround for `ETag.All`:

image


iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -92,7 +92,7 @@ internal static void ConditionallyInsertETag(HttpRequestMessage requestMessage,
             {
                 // "Perform this operation only if the entity is unchanged"
                 // Sends the If-Match header with a value of the ETag.
-                string escapedETag = EscapeETag(eTag.ToString());
+                string escapedETag = eTag.ToString("H");
```
abhipsaMisra, line 95: From Azure.Core:
"Valid values are "G" for standard format and "H" for header format."

https://github.com/Azure/azure-sdk-for-net/blob/9c6238e0f0dd403d6583b56ec7902c77c64a2e37/sdk/core/Azure.Core/src/ETag.cs#L87-L114

iothub/service/src/Utilities/NewtonsoftJsonETagConverter.cs
```diff
@@ -44,7 +44,7 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
             }
             else
             {
-                writer.WriteValue(eTag.ToString("H"));
+                writer.WriteValue(eTag.ToString());
```
abhipsaMisra, line 47: The json etag converter is used for serializing the object to be used in the request body. The "H" format tag is not valid for this scenario. The default tag is "G" -> standard formatting.

build.ps1
```diff
@@ -355,6 +355,21 @@ try
         $env:AZURE_IOT_LOCALPACKAGES = $packageTempPath
     }
 
+    if ($localPackagesAvailableForTesting)
```
abhipsaMisra, line 358: Moving this up the confirm the local packages directory immediately after the solution has been built.

build.ps1
```diff
@@ -484,6 +476,14 @@ try
         RunApp e2e\stress\MemoryLeakTest "MemoryLeakTest test"
     }
 
+    if ($package)
```
abhipsaMisra, line 479: Moving this below to be after the tests section

provisioning/service/samples/getting started/CleanupEnrollmentsSample/CleanupEnrollmentsSample.cs
```diff
@@ -88,7 +89,14 @@ private async Task QueryAndDeleteEnrollmentGroupsAsync()
                     {
                         Console.WriteLine($"Enrollment group to be deleted: {enrollment.EnrollmentGroupId}");
                         s_enrollmentGroupsDeleted++;
-                        await _provisioningServiceClient.EnrollmentGroups.DeleteAsync(enrollment.EnrollmentGroupId);
+                        try
+                        {
+                            await _provisioningServiceClient.EnrollmentGroups.DeleteAsync($"{enrollment.EnrollmentGroupId}_1");
+                        }
+                        catch (ProvisioningServiceException ex) when (ex.StatusCode == HttpStatusCode.NotFound)
```
abhipsaMisra, line 96: This is added to address pipeline failures. The different frameworks could be attempting the delete the same enrollments, resulting in this error.
Once the tests are updated to clean up their respective resources, we can limit the pipeline cleanup sample to a single target framework run.

provisioning/service/src/Http/ContractApiHttp.cs
```diff
@@ -218,7 +218,7 @@ private static void ValidateHttpResponse(ContractApiResponse response)
                     if (response.StatusCode >= HttpStatusCode.Ambiguous)
                     {
                         throw new ProvisioningServiceException(
-                            $"{response.ErrorMessage}:{response.Body}",
+                            $"{response.ErrorMessage}:{responseBody.Message}",
```
abhipsaMisra, line 221: Using the message from the response for the exception message

timtay-microsoft: What's the difference here?

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -102,26 +102,5 @@ internal static void ConditionallyInsertETag(HttpRequestMessage requestMessage,
                 requestMessage.Headers.IfMatch.Add(new EntityTagHeaderValue(s_eTagForce, true));
             }
         }
-
-        // ETag values other than "*" need to be wrapped in escaped quotes if they are not
-        // already.
-        private static string EscapeETag(string eTag)
```
timtay-microsoft, line 108: I love that we can get rid of this

timtay-microsoft (approved):

iothub/service/src/Utilities/NewtonsoftJsonETagConverter.cs
drwill-ms: Should we be explicit?

provisioning/service/src/Http/ContractApiHttp.cs
drwill-ms:
```suggestion
                            $"{response.ErrorMessage}: {responseBody.Message}",
```

drwill-ms (approved):

provisioning/service/src/Http/ContractApiHttp.cs
abhipsaMisra: `response.Body` is the complete unparsed response body, whereas `responseBody.Message` is the message body. 
Since we've parsed out the rest of the response into status, error code, tracking Id etc., it makes sense for the exception message to contain only the error message.

iothub/service/src/Http/HttpMessageHelper.cs
```diff
@@ -80,48 +75,30 @@ internal static async Task<T> DeserializeResponseAsync<T>(HttpResponseMessage re
...
             {
-                escapedETagBuilder.Append('"');
+                // Azure.Core.ETag expects the format "H" for serializing ETags that go into the header.
+                // https://github.com/Azure/azure-sdk-for-net/blob/9c6238e0f0dd403d6583b56ec7902c77c64a2e37/sdk/core/Azure.Core/src/ETag.cs#L87-L114
+                // Also, System.Net.Http.Headers does not allow ETag.All (*) as a valid value even though RFC allows it.
+                // For this reason, we'll add the ETag value without additional validation.
+                // System.Net.Http.Headers validation: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/tests/UnitTests/Headers/EntityTagHeaderValueTest.cs#L214,
+                // https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/Headers/GenericHeaderParser.cs#L98
+                // RFC specification: https://www.rfc-editor.org/rfc/rfc7232#section-3.1
+                _ = requestMessage.Headers.TryAddWithoutValidation("If-Match", eTag.ToString("H"));
```
abhipsaMisra, line 100: Should we throw an exception if this "TryAdd" fails? Previously this threw IEnumerable.Add exceptions or an exception that HttpRequestHeaders defined, eg. FormatException.

schoims (approved):

iothub/service/src/Http/HttpMessageHelper.cs
drwill-ms: That's a tough one. Should a user know if we can't put their eTag in and it just succeeds anyway? Maybe so.

Will that be a scenario blocker though? Which is worse?

drwill-ms (approved):

iothub/service/src/Http/HttpMessageHelper.cs
abhipsaMisra: This adds the etag without any validation. So the only scenario I'd except this to fail over is if we try to add the etag twice - key already exists scenario. Since our SDK doesn't do that, I'm ok with us not inspecting the result of this call.

PR #3081, opened 2023-01-18
Add test cases for provisioning service client

abhipsaMisra:
Test run with current code:
image

~Fix is in [Fix ETag handling for serialization into http request header and body](https://github.com/Azure/azure-iot-sdk-csharp/pull/3066)~ Fix has been merged in


e2e/test/helpers/RetryOperationHelper.cs
```diff
@@ -38,7 +38,7 @@ public class RetryOperationHelper
                     await asyncOperation().ConfigureAwait(false);
                     return;
                 }
-                catch (Exception ex) when (!retryPolicy.ShouldRetry(counter, ex, out retryInterval))
+                catch (Exception ex) when (retryPolicy.ShouldRetry(counter, ex, out retryInterval))
```
abhipsaMisra, line 41: Exception should be retried only if the retry policy states so. Else the exception should be thrown back to the caller test.

timtay-microsoft (approved):

e2e/test/provisioning/ProvisioningServiceClientE2ETests.EnrollmentGroupsClient.cs
```diff
@@ -0,0 +1,347 @@
...
+            string registrationId2 = $"{s_devicePrefix}_{Guid.NewGuid()}";
+
+            // create two enrollments in bulk
+            var enrollmentGroup1 = new EnrollmentGroup(registrationId1, new SymmetricKeyAttestation());
+            var enrollmentGroup2 = new EnrollmentGroup(registrationId2, new SymmetricKeyAttestation());
+            var createEnrollmentGroupList = new List<EnrollmentGroup> { enrollmentGroup1, enrollmentGroup2 };
+
+            BulkEnrollmentOperationResult createBulkEnrollmentResult = await provisioningServiceClient
+                .EnrollmentGroups
+                .RunBulkOperationAsync(BulkOperationMode.Create, createEnrollmentGroupList);
```
drwill-ms, line 98:
```suggestion
                .RunBulkOperationAsync(BulkOperationMode.Create, createEnrollmentGroupList)
                .ConfigureAwait(false);
```

drwill-ms: below too

e2e/test/provisioning/ProvisioningServiceClientE2ETests.IndividualEnrollmentsClient.cs
```diff
@@ -0,0 +1,392 @@
...
+using Azure;
+using FluentAssertions;
+using FluentAssertions.Specialized;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.Azure.Devices.Provisioning.Service;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests.Provisioning
+{
+    public partial class ProvisioningServiceClientE2ETests : E2EMsTestBase
```
drwill-ms, line 19: Why partial classes?

drwill-ms: I mean, why not just give each their own class with a name that is appropriate, e.g. IndividualEnrollmentsClientsTests.

drwill-ms (approved):

e2e/test/provisioning/ProvisioningServiceClientE2ETests.IndividualEnrollmentsClient.cs
abhipsaMisra: The test file was getting too large to include test cases for both individual enrollments and enrollment groups. I could have split them into separate files, but I wanted to make use of some common methods for both. So, I decided to use partial classes instead.

drwill-ms: I don't see a good argument for partial classes. I agree we don't want files with too many tests and in splitting them up.

I see partial classes as a solution to a specific set of problems, not something to be used because it feels convenient. They come at a conceptual cost for code readers. Why is this class partial? Where is the other implementation? Navigating to definition is harder, etc.

In this case, can we just make them different class names?

e2e/test/provisioning/ProvisioningServiceIndividualEnrollmentTests.cs
```diff
@@ -0,0 +1,431 @@
...
+        private static async Task ProvisioningServiceClient_GetIndividualEnrollmentAttestation(AttestationMechanismType attestationType)
+        {
+            using var provisioningServiceClient = new ProvisioningServiceClient(TestConfiguration.Provisioning.ConnectionString);
+            string registrationId = AttestationTypeToString(attestationType) + "-" + Guid.NewGuid();
+
+            IndividualEnrollment individualEnrollment = null;
+
+            try
+            {
+                individualEnrollment = await CreateIndividualEnrollmentAsync(
```
schoims, line 161: nit: tabbing looks a bit off
```suggestion
                individualEnrollment = await CreateIndividualEnrollmentAsync(
                    provisioningServiceClient,
                    registrationId,
                    attestationType,
                    null,
                    null,
                    AllocationPolicy.Static,
                    null).ConfigureAwait(false);
```

schoims (approved):

e2e/test/provisioning/ProvisioningServiceIndividualEnrollmentTests.cs
drwill-ms: Nope, that is intended. Too many parameters so each get their own line, and we have ConfigureAwait after, which we'd prefer on the next line. So, the vars get an extra tab.

If this was more fluent in style, you'd do another newline on the method, so something like:

```csharp
var foo = MyObj
    .MethodAsync(
        param1,
        param2)
    .ConfigureAwait(false);
```

schoims: Thank you for clarifying!

PR #3082, opened 2023-01-18
Remove ClientRuntime dependency and add Newtonsoft.Json package

schoims:
As part of the v2 migration, HTTP has been removed as a transport option, resulting in Microsoft.Rest.ClientRuntime no longer being required as a dependency of Provisioning Client SDK. However, its dependency is still utilized, so this change has been introduced.

### DIFF
--- a/provisioning/device/src/Microsoft.Azure.Devices.Provisioning.Client.csproj
+++ b/provisioning/device/src/Microsoft.Azure.Devices.Provisioning.Client.csproj
@@ -64,10 +64,11 @@
     <PackageReference Include="Azure.Core" Version="1.26.0" />
     <PackageReference Include="Microsoft.Azure.Amqp" Version="2.5.12" />
     <PackageReference Include="MQTTnet" Version="4.1.2.350" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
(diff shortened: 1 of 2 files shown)


provisioning/device/src/Microsoft.Azure.Devices.Provisioning.Client.csproj
```diff
@@ -64,10 +64,11 @@
     <PackageReference Include="Azure.Core" Version="1.26.0" />
     <PackageReference Include="Microsoft.Azure.Amqp" Version="2.5.12" />
     <PackageReference Include="MQTTnet" Version="4.1.2.350" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
drwill-ms, line 67: Not 13.02?

drwill-ms (approved):

provisioning/device/src/Microsoft.Azure.Devices.Provisioning.Client.csproj
schoims: I can update it to 13.0.2. I chose 13.0.1 for consistency only since other SDKs use this version.

abhipsaMisra (approved):

PR #3084, opened 2023-01-19
{AzureIOT} fixes Azure/azure-sdk-for-net#33498 Add the deprecation info

navba-MSFT:
fixes Azure/azure-sdk-for-net#33498 Add the deprecation info about ServiceClient

In the Java SDK we see this info. But in .NET SDK for IOT this info seems to be missing.

More Info here: https://learn.microsoft.com/en-us/java/api/com.microsoft.azure.sdk.iot.service.serviceclient?view=azure-java-stable

## Checklist

### DIFF
--- a/iothub/service/src/Messaging/ServiceClient.cs
+++ b/iothub/service/src/Messaging/ServiceClient.cs
@@ -123,7 +123,8 @@ namespace Microsoft.Azure.Devices
         }
 
         /// <summary>
-        /// Creates ServiceClient from an IoT hub connection string.
+        /// Creates ServiceClient from an IoT hub connection string. This method is deprecated because this method declares a thrown IOException
+        /// even though it never throws an IOException.
         /// </summary>
         /// <param name="connectionString">Connection string for the IoT hub.</param>
         /// <param name="options">The <see cref="ServiceClientOptions"/> that allow configuration of the service client instance during initialization.</param>
@@ -246,7 +247,8 @@ namespace Microsoft.Azure.Devices
         }
 
         /// <summary>
-        /// Create an instance of ServiceClient from the specified IoT hub connection string using specified Transport Type.
+        /// Create an instance of ServiceClient from the specified IoT hub connection string using specified Transport Type. This method is deprecated
+        /// because this method declares a thrown IOException even though it never throws an IOException.
         /// </summary>
         /// <param name="connectionString">Connection string for the IoT hub.</param>
         /// <param name="transportType">The <see cref="TransportType"/> used (Amqp or Amqp_WebSocket_Only).</param>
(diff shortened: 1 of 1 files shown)


abhipsaMisra:
@navba-MSFT This PR doesn't seem relevant to the .NET SDK. The comment regarding API depreciation was added in the linked Java SDK because it had a public method with an incorrectly defined exception that it did not throw.
This is not true for the .NET SDK, and hence no API has been marked as deprecated in here.

Could you share some more details about what the intended change is?

navba-MSFT:
> @navba-MSFT This PR doesn't seem relevant to the .NET SDK. The comment regarding API depreciation was added in the linked Java SDK because it had a public method with an incorrectly defined exception that it did not throw. This is not true for the .NET SDK, and hence no API has been marked as deprecated in here.
> 
> Could you share some more details about what the intended change is?

@abhipsaMisra Thanks for reviewing. We have got an issue reported by the customer here: https://github.com/Azure/azure-sdk-for-net/issues/33498  I am sharing the issue verbatim here if that helps. Hence this PR was created. So are you saying that the [.NET serviceclient](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.serviceclient?view=azure-dotnet) is not deprecated yet ?

> _There are services which are outdated in Azure IOT SDK such as serviceclient, but the documentation still exists. This is confusing for the both existing users as well as new users. Request you to remove such documents or clearly put a header note mentioning that the service has outdated and kept for reference purpose only?._

navba-MSFT:
@abhipsaMisra Could you please review this once you get a chance ?

abhipsaMisra:
> @abhipsaMisra Thanks for reviewing. We have got an issue reported by the customer here: [Azure/azure-sdk-for-net#33498](https://github.com/Azure/azure-sdk-for-net/issues/33498) I am sharing the issue verbatim below if that helps. Hence this PR was created. So are you saying that the [.NET serviceclient](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.serviceclient?view=azure-dotnet) is not deprecated yet ?
> 
> > _There are services which are outdated in Azure IOT SDK such as serviceclient, but the documentation still exists. This is confusing for the both existing users as well as new users. Request you to remove such documents or clearly put a header note mentioning that the service has outdated and kept for reference purpose only?._

Yes, that is correct - the [IoT .NET ServiceClient]( https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.devices.serviceclient?view=azure-dotnet) has not been deprecated yet. 
Could it be that the original post is looking for the ServiceClient under azure-dotnet-legacy, and being unable to find it thinks that the ServiceClient has been deprecated? Our GA'd APIs can be found under azure-dotnet (latest).

image

navba-MSFT:
Thanks @abhipsaMisra .

I am closing this PR for now.

PR #3091, opened 2023-01-26
Make CertificateInstaller unit testable by turning X509Store into an injectable dependency

drwill-ms:
In a 1:1 with Nate talking about how to unit test some methods that seem hard, this is an example of dependency injection.

### DIFF
--- a/provisioning/device/src/Utilities/CertificateInstaller.cs
+++ b/provisioning/device/src/Utilities/CertificateInstaller.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Security.Cryptography.X509Certificates;
+using Microsoft.Azure.Devices.Provisioning.Client.Utilities;
 
 namespace Microsoft.Azure.Devices.Provisioning.Client
 {
@@ -20,7 +21,10 @@ namespace Microsoft.Azure.Devices.Provisioning.Client
         /// for more information.
         /// </remarks>
         /// <param name="certificates">The certificate chain to ensure is installed.</param>
-        public static void EnsureChainIsInstalled(X509Certificate2Collection certificates)
+        /// <param name="certStore">For unit testing to allow for mocking.</param>
+        internal static void EnsureChainIsInstalled(
+            X509Certificate2Collection certificates,
+            ICertificiateStore certStore = default)
         {
             if (certificates == null
                 || certificates.Count == 0)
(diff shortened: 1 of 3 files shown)


provisioning/device/src/Utilities/CertificateInstaller.cs
```diff
@@ -20,7 +21,10 @@ internal static class CertificateInstaller
         /// for more information.
         /// </remarks>
         /// <param name="certificates">The certificate chain to ensure is installed.</param>
-        public static void EnsureChainIsInstalled(X509Certificate2Collection certificates)
+        /// <param name="certStore">For unit testing to allow for mocking.</param>
+        internal static void EnsureChainIsInstalled(
+            X509Certificate2Collection certificates,
+            ICertificiateStore certStore = default)
```
drwill-ms, line 27: The client won't pass this parameter, but a unit test can, with a mock of `ICertificateStore`. By doing so it can pass different certificates and control whether the code will try to add it to the store or not.

provisioning/device/src/Utilities/ICertficiateStore.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Utilities
+{
+    internal interface ICertificiateStore
+    {
+        internal bool Find(X509Certificate2 certificate);
```
drwill-ms, line 10: Probably needs a better set of names.

provisioning/device/src/Utilities/X509CertificateStore.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Azure.Devices.Provisioning.Client.Utilities
+{
+    internal class X509CertificateStore : ICertificiateStore, IDisposable
```
drwill-ms, line 9: Could use doc comments to explain what this class is and why it exists, like, to enable mocking of `CertfiicateInstaller` so unit tests don't need to mess with the system cert store.

drwill-ms, line 9: Maybe instead of an interface, we make the class methods virtual and a unit test just overrides them. How would that look in a unit test? That would rid us of the interface definition.

PR #3092, opened 2023-01-30
Cleanup and add more unit tests

ngastelum-ms:


provisioning/device/tests/AuthenticationProviderX509Tests.cs
```diff
@@ -1,16 +1,21 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+﻿using Microsoft.Azure.Devices.Provisioning.Client;
+// Copyright (c) Microsoft. All rights reserved.
```
schoims, line 2: Move this using statement below

provisioning/device/src/Utilities/CertificateInstaller.cs
```diff
@@ -31,27 +35,25 @@ public static void EnsureChainIsInstalled(X509Certificate2Collection certificate
                 return;
             }
 
+
```
brycewang-microsoft, line 38: nit: extra line

provisioning/device/tests/AuthenticationProviderX509Tests.cs
schoims: just curious, are we actually using Microsoft.Azure.Devices.Provisioning.Client in this test file?

provisioning/device/src/Utilities/CertificateInstaller.cs
```diff
@@ -60,7 +62,14 @@ public static void EnsureChainIsInstalled(X509Certificate2Collection certificate
...
+                    if (certStore is IDisposable disposableCertStore)
+                    {
+                        disposableCertStore.Dispose();
+                    }
+                }
             }
         }
     }
-}
+}
```
brycewang-microsoft, line 75: nit: need one extra line at end of file

provisioning/device/tests/AuthenticationProviderX509Tests.cs
ngastelum-ms: No we're not, not sure how that got up there. Removed it

schoims (approved):

provisioning/device/tests/AuthenticationProviderX509Tests.cs
```diff
@@ -1,16 +1,21 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+﻿using Microsoft.Azure.Devices.Provisioning.Client;
```
brycewang-microsoft, line 1: Is this needed?

provisioning/device/tests/AuthenticationProviderX509Tests.cs
```diff
@@ -20,5 +25,23 @@ public void AuthenticationProviderX509_ThrowsWhenMissingCert()
             // assert
             act.Should().Throw<ArgumentException>();
         }
+
+        [TestMethod]
+        public void AuthenticationProviderX509_Works()
+        {
+            // arrange
```
brycewang-microsoft, line 32: We can leave an extra line below this comment if we have a code block instead of a single line in this section.

provisioning/device/tests/CertificateInstallerTests.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Microsoft.Azure.Devices.Provisioning.Client.UnitTests
+{
+    [TestClass]
+    [TestCategory("Unit")]
+    public class CertificateInstallerTests
+    {
+        [TestMethod]
+        public void CertificateInstaller_EnsureChainIsInstalled_Works()
+        {
+            // arrange
```
brycewang-microsoft, line 22: leave an extra line when we have a code block below, since we want to follow with the same standard here and there

PR #3095, opened 2023-02-03
Remove backdoor cert validation bypass

drwill-ms:
We don't have any code using it and Edge verified they don't.


ngastelum-ms (approved):

timtay-microsoft (approved):
Since Edge isn't using this anymore, I'm fine with removing this.

schoims (approved):

Issue #3093, opened 2023-01-31
ISO 8601 date strings are modified in cloud to device method results


pihai:
# Context

- **SDK version used:** Microsoft.Azure.Devices 1.38.2

## Description of the issue
If a `DateTimeOffest` formatted as ISO 8601 string is used in a `CloudToDeviceMethodResult` then the SDK messes with the date, like removing trailing zeros.

## Code sample exhibiting the issue
```csharp
using System.Globalization;
using System.Text;
using Microsoft.Azure.Devices;
using Microsoft.Azure.Devices.Client;
using Newtonsoft.Json;

const string serviceConnectionString = "<TODO>";
const string deviceConnectionString = "<TODO>";

var deviceId = Microsoft.Azure.Devices.Client.IotHubConnectionStringBuilder.Create(deviceConnectionString).DeviceId;
const string methodName = "test";
var deviceClient = DeviceClient.CreateFromConnectionString(deviceConnectionString, Microsoft.Azure.Devices.Client.TransportType.Amqp_WebSocket_Only);

var date = new DateTimeOffset(638107582284599400, TimeSpan.FromHours(1));
var responseJson = JsonConvert.SerializeObject(new MyDto { MyIso8601Date = date.ToString("o", CultureInfo.InvariantCulture) });
var responseBytes = Encoding.UTF8.GetBytes(responseJson);

await deviceClient.SetMethodHandlerAsync(
    methodName, 
    async (request, context) => new MethodResponse(responseBytes, 0),
    null);

var serviceClient = ServiceClient.CreateFromConnectionString(serviceConnectionString, Microsoft.Azure.Devices.TransportType.Amqp_WebSocket_Only);

var result = await serviceClient.InvokeDeviceMethodAsync(deviceId, new CloudToDeviceMethod(methodName).SetPayloadJson("{}"));

var actualResponse = result.GetPayloadAsJson();

Console.WriteLine(responseJson);
Console.WriteLine(actualResponse);

public class MyDto
{
    public string MyIso8601Date { get; set; }
}
```

Outputs:
```
{"MyIso8601Date":"2023-01-31T10:37:08.4599400+01:00"}
{"MyIso8601Date":"2023-01-31T10:37:08.45994+01:00"}
```

Notice that the trailing zeros are gone. This is problematic when the string is converted back into a `DateTimeOffset` using the same format identifier because a `FormatException` is thrown:

```csharp
[...]

PR #3097, opened 2023-02-06
Fix bug in date deserialization of device method response 

brycewang-microsoft:
For https://github.com/Azure/azure-iot-sdk-csharp/issues/3093,

Newtonsoft.Json has a known [issue ](https://github.com/JamesNK/Newtonsoft.Json/issues/1511)where date serialization/deserialization drops trailing zeros in the microseconds date portion by default. As a result, a method payload containing date type will be deserialized into an improperly formatted ISO 8601 string and cannot be parsed correctly. 

```c#
{"MyIso8601Date":"2023-01-31T01:37:08.45994-08:00"}

Unhandled exception. System.FormatException: String '2023-01-31T01:37:08.45994-08:00' was not recognized as a valid DateTime.
   at System.DateTimeParse.ParseExact(ReadOnlySpan`1 s, ReadOnlySpan`1 format, DateTimeFormatInfo dtfi, DateTimeStyles style, TimeSpan& offset)
   at System.DateTimeOffset.ParseExact(String input, String format, IFormatProvider formatProvider, DateTimeStyles styles)
```

### DIFF
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -307,12 +307,14 @@ namespace Microsoft.Azure.Devices
                 return (T)(object)message;
             }
 
-#if NET451
-            T entity = await message.Content.ReadAsAsync<T>(token).ConfigureAwait(false);
-#else
+            /*#if NET451
+                        T entity = await message.Content.ReadAsAsync<T>(token).ConfigureAwait(false);
+            #else
+                        string str = await message.Content.ReadHttpContentAsStringAsync(token).ConfigureAwait(false);
+                        T entity = JsonConvert.DeserializeObject<T>(str);
+            #endif*/
             string str = await message.Content.ReadHttpContentAsStringAsync(token).ConfigureAwait(false);
             T entity = JsonConvert.DeserializeObject<T>(str, new JsonSerializerSettings { DateParseHandling = DateParseHandling.None });
-#endif
 
             // Etag in the header is considered authoritative
             var eTagHolder = entity as IETagHolder;
--- a/common/src/service/HttpClientHelper.cs
+++ b/common/src/service/HttpClientHelper.cs
@@ -48,6 +48,8 @@ namespace Microsoft.Azure.Devices
             _defaultErrorMapping = defaultErrorMapping;
             _defaultOperationTimeout = timeout;
 
+            JsonConvert.DefaultSettings = JsonSerializerSettingsInitializer.GetJsonSerializerSettingsDelegate();
+
             // We need two types of HttpClients, one with our default operation timeout, and one without. The one without will rely on
             // a cancellation token.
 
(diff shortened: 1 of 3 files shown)


common/src/service/HttpClientHelper.cs
```diff
@@ -307,12 +307,8 @@ private static async Task<T> ReadResponseMessageAsync<T>(HttpResponseMessage mes
                 return (T)(object)message;
             }
 
-#if NET451
-            T entity = await message.Content.ReadAsAsync<T>(token).ConfigureAwait(false);
-#else
             string str = await message.Content.ReadHttpContentAsStringAsync(token).ConfigureAwait(false);
-            T entity = JsonConvert.DeserializeObject<T>(str);
-#endif
+            T entity = JsonConvert.DeserializeObject<T>(str, new JsonSerializerSettings { DateParseHandling = DateParseHandling.None });
```
brycewang-microsoft, line 311: By default, serializing/deserializing with Newtonsoft.Json will try to parse date-formatted strings to a date type, which drops trailing zeros in the microseconds date portion ("2023-01-31T10:37:08.45994**00**+01:00" -> "2023-01-31T10:37:08.45994+01:00"). By specifying `DateParseHandling` with `None`, the original string will be read as-is.

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -617,5 +678,10 @@ await Task
 
             await moduleClient.CloseAsync().ConfigureAwait(false);
         }
+
+        private class DateTime
```
drwill-ms, line 682: Don't cause a naming conflict with a built-in type.

drwill-ms (approved):

e2e/test/iothub/method/MethodE2ETests.cs
```diff
@@ -301,6 +303,65 @@ await deviceClient
             }
         }
 
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task Method_ServiceInvokeDeviceMethodWithDateTimePayload_DoesNotThrow()
+        {
+            // arrange
+
```
schoims, line 311: nit: extra line

tmahmood-microsoft (approved):

common/src/service/HttpClientHelper.cs
abhipsaMisra: Should we set the serializer settings at a library level so that all dates are handled as expected? (similar to how we set the max json depth value).

abhipsaMisra (approved):

e2e/test/iothub/method/MethodE2ETests.cs
brycewang-microsoft: this extra line is actually left intentionally as we have a code block below other than a single line in the "arrange" section.

common/src/service/HttpClientHelper.cs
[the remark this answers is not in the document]
brycewang-microsoft: Added a new class `JsonSerializerSettingsInitializer` to set up the global serializer settings.

PR #3098, opened 2023-02-06
Fix bug in date deserialization of device method response

brycewang-microsoft:
Similar to PR #3097 but for V2.

### DIFF
--- a/iothub/service/src/IotHubServiceClient.cs
+++ b/iothub/service/src/IotHubServiceClient.cs
@@ -6,6 +6,7 @@ using System.Net.Http;
 using Azure;
 using Azure.Core;
 using Microsoft.Azure.Amqp;
+using Newtonsoft.Json;
 
 namespace Microsoft.Azure.Devices
 {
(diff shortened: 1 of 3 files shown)


schoims (approved):

iothub/service/src/IotHubServiceClient.cs
```diff
@@ -233,6 +234,9 @@ private void InitializeSubclients(IotHubServiceClientOptions options)
             MessageFeedback = new MessageFeedbackProcessorClient(_hostName, _credentialProvider, options, _retryHandler);
             FileUploadNotifications = new FileUploadNotificationProcessorClient(_hostName, _credentialProvider, options, _retryHandler);
 
+            // Specify the JsonSerializerSettings for subclients
+            JsonConvert.DefaultSettings = JsonSerializerSettingsInitializer.GetJsonSerializerSettingsDelegate();
```
abhipsaMisra, line 238: What about the other clients? Would you need to do this for provisioning clients as well? The device client might need a slightly different approach since the serializer there is customizable.

abhipsaMisra (approved):

iothub/service/src/IotHubServiceClient.cs
brycewang-microsoft: Good point! I will look into other clients within a new PR since I missed this for dps clients in the previous PR (commit to "main").

iothub/service/src/JsonSerializerSettingsInitializer.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// A class to initialize JsonSerializerSettings which can be applied to the project.
+    /// </summary>
+    public static class JsonSerializerSettingsInitializer
```
drwill-ms, line 12: Why public? We don't want this in our public API, do we?

PR #3106, opened 2023-02-08
Fix build errors in release flavor

drwill-ms:
Mostly static analysis

### DIFF
--- a/iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
+++ b/iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
@@ -9,11 +9,14 @@ using Microsoft.Azure.Devices.Client.Transport.AmqpIot;
 
 namespace Microsoft.Azure.Devices.Client.Transport.Amqp
 {
-    internal sealed class AmqpConnectionPool : IAmqpUnitManager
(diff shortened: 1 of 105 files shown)


iothub/device/src/Transport/Amqp/AmqpConnectionPool.cs
```diff
@@ -9,17 +9,12 @@
...
 namespace Microsoft.Azure.Devices.Client.Transport.Amqp
 {
-    internal class AmqpConnectionPool : IAmqpUnitManager
+    internal sealed class AmqpConnectionPool : IAmqpUnitManager
     {
         private AmqpConnectionHolder[] _amqpSasIndividualPool;
         private readonly Dictionary<string, AmqpConnectionHolder[]> _amqpSasGroupedPool = new();
         private readonly object _lock = new();
 
-        protected virtual IDictionary<string, AmqpConnectionHolder[]> GetAmqpSasGroupedPoolDictionary()
```
drwill-ms, line 18: If not protected, don't need this method, but even if we did, a property with private setter and protected getter would be simpler.

brycewang-microsoft (approved):

schoims (approved):

PR #3107, opened 2023-02-09
Apply JsonSerializerSettings to hub device

brycewang-microsoft:
Also bring "JsonSerializerSettingsInitializer" from the "shared" folder to both the hub device and service clients separately.

### DIFF
--- a/e2e/test/iothub/twin/TwinE2ETests.cs
+++ b/e2e/test/iothub/twin/TwinE2ETests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Devices.Client;
@@ -11,6 +12,7 @@ using Microsoft.Azure.Devices.E2ETests.Helpers;
 using Microsoft.Azure.Devices.Shared;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.Devices.E2ETests.Twins
 {
@@ -34,6 +36,13 @@ namespace Microsoft.Azure.Devices.E2ETests.Twins
                 Name = "someName"
             }
         };
+
+        private static readonly TestDateTime s_dateTimeProperty = new()
+        {
+            Iso8601String = new DateTimeOffset(638107582284599400, TimeSpan.FromHours(1)).ToString("o", CultureInfo.InvariantCulture)
+        };
+
+        private static readonly string s_iso8601DateTimeString = "2023-01-31T10:37:08.4599400+01:00";
 
         [TestMethod]
         [Timeout(TestTimeoutMilliseconds)]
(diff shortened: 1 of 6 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -161,6 +162,8 @@ internal class InternalClient : IDisposable
             if (Logging.IsEnabled)
                 Logging.Enter(this, transportSettings, pipelineBuilder, nameof(InternalClient) + "_ctor");
 
+            JsonConvert.DefaultSettings = JsonSerializerSettingsInitializer.GetJsonSerializerSettingsDelegate();
```
abhipsaMisra, line 165: Tests for hub device settings?

abhipsaMisra (approved):

drwill-ms (approved):
Just needs comments about the significance of test data and links to the bug we're working around. Also, in the JsonSerializerSettingsInitializer class.

e2e/test/iothub/twin/TwinE2ETests.cs
```diff
@@ -656,6 +701,35 @@ private async Task Twin_ServiceSetsDesiredPropertyAndDeviceReceivesItOnNextGetAs
...
+
+            var twinPatch = new Twin();
+
+            JObject jobjectProperty = JsonConvert.DeserializeObject<JObject>(
+                jsonString,
+                new JsonSerializerSettings
+                {
+                    // This ensures the date-formatted string will not drop trailing zeros
+                    // in the microseconds date portion while parsing the json properties.
+                    DateParseHandling = DateParseHandling.None
```
abhipsaMisra, line 720: This is set in the SDK, right? Does the user need to set this again while setting the properties?

brycewang-microsoft: The setting we just added is for SDK only and will only take effect after the (internal) client has been initiated. 

This setting here will only apply to the property itself, so if the users want to keep the trailing zero while setting the property, they would need to go with this.

brycewang-microsoft: I can add more comments here to clarify this which can hopefully be helpful for us or anyone checking this e2e test.

schoims: Kindly let me know if I am mistaken, even if the user didn't pass in a new JsonSerializerSettings here, deviceTwin.Properties.Desired["Iso8601String"] should have the trailing zeros because of the default setting we introduced in earlier PR. Right?

schoims: I deleted my earlier comment. This setting was added to make sure the datetime has the trailing zeros and we return desired property with the trailing zeros intact, right?

schoims (approved):

e2e/test/iothub/twin/TwinE2ETests.cs
brycewang-microsoft: > I deleted my earlier comment. This setting was added to make sure the datetime has the trailing zeros and this test is verifying that returned desired property has the trailing zeros intact, right?

Yes, that's correct. FWIW, users don't have to use the property this way, but if they want, here's the example.

PR #3109, opened 2023-02-10
Make Location property from DigitalTwinUpdateResponse internal

schoims:
Location property from DigitialTwinUpdateResponse is marked internal as there is no known use case from users and it was added to be complete with the guidance.

### DIFF
--- a/iothub/service/src/DigitalTwin/Models/DigitalTwinUpdateResponse.cs
+++ b/iothub/service/src/DigitalTwin/Models/DigitalTwinUpdateResponse.cs
@@ -34,6 +34,6 @@ namespace Microsoft.Azure.Devices
         /// <summary>
         /// Gets the URI of the digital twin.
         /// </summary>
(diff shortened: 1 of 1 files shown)


tmahmood-microsoft (approved):

iothub/service/src/DigitalTwin/Models/DigitalTwinUpdateResponse.cs
```diff
@@ -34,6 +34,6 @@ protected internal DigitalTwinUpdateResponse(ETag eTag = default, string locatio
         /// <summary>
         /// Gets the URI of the digital twin.
         /// </summary>
-        public string Location { get; }
+        internal string Location { get; }
```
drwill-ms, line 37: Please add a comment as to why we're hiding this.

1. No known use case for users.
2. Only added to the service to be complete with guidance.

drwill-ms (approved):

PR #3111, opened 2023-02-10
Test only with project references

drwill-ms:
Simplify our testing process by removing step to test against built NuGet packges as we have not seen this step produce any value in the last several years (at least) and it often costs us some trouble.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -31,9 +31,6 @@ Build will automatically detect if the machine is Windows vs Unix. On Windows de
 
 The following environment variables can tune the build behavior:
     - AZURE_IOT_DONOTSIGN: disables delay-signing if set to 'TRUE'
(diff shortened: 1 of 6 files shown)


patilsnr (approved):

abhipsaMisra (approved):
It might be worth adding a note somewhere that the only remaining reason for maining separate release pipelines for build+sign and test is so that we can share the signed nugets with the partner team for testing.

drwill-ms:
> It might be worth adding a note somewhere that the only remaining reason for maining separate release pipelines for build+sign and test is so that we can share the signed nugets with the partner team for testing.

I think we should just always make nugets (perhaps unsigned) and push them to our own ADO NuGet feed, so then we won't need to maintain separate pipelines. Signing and packing nugets could even be a parameterized step.

andyk-ms (approved):

PR #3112, opened 2023-02-13
Fix device export deserialization of Id

drwill-ms:


iothub/service/src/Registry/Models/ExportImportDevice.cs
```diff
@@ -58,7 +58,7 @@ public ExportImportDevice(string deviceId, ImportMode importMode)
         /// The unique identifier of the device.
         /// </summary>
         [JsonProperty("id")]
-        public string Id { get; }
+        public string Id { get; set; }
```
drwill-ms, line 61: If only settable via ctor then can't be set by deserialization.

iothub/service/src/Registry/Models/ExportImportDevice.cs
```diff
@@ -82,7 +82,7 @@ public ExportImportDevice(string deviceId, ImportMode importMode)
         /// The type of registry operation and ETag preferences.
         /// </summary>
         [JsonProperty("importMode")]
-        public ImportMode ImportMode { get; set; }
+        public ImportMode? ImportMode { get; set; }
```
drwill-ms, line 85: Allow this to be null so when deserialized, if no value we don't default to one that isn't true.

schoims (approved):

abhipsaMisra (approved):

PR #3116, opened 2023-02-15
BugFix(MQTT): send an event in ReceiveMessageHandler before completin…

tmahmood-microsoft:
For MQTT, after receiving C2D message, the implementation was making an additional call to CompleteAsync(). This did not cause an issue if message handler was eventually only making call to Complete/Abandon/RejectAsync() but if an event was sent within message handler, the ACKs within CompleteAsync() get mismatched. The mismatch was due to awaited event call causing delay in ACK for the first lock token. i.e:

**_Flow with no event within message handler:_**
- Add message lock token to queue.
- Remove lock token from queue and complete ACK.
- Add message lock token **again** to the queue.
- Remove lock token from queue and complete ACK.

**_Flow with event within message handler:_**
- Add message lock token to queue.
- Add message lock token **again** to the queue.
- Remove first lock token from queue but expects the second lock token.
- Remove second lock token from queue but expects the first lock token.

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -578,9 +578,7 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             // We are intentionally not awaiting _deviceMessageReceivedListener callback.
             // This is a user-supplied callback that isn't required to be awaited by us. We can simply invoke it and continue.
             _ = _deviceMessageReceivedListener?.Invoke(message);
-            // Messages with QoS = 1 need to be Acknowledged otherwise it results in mismatched Ack to IoT Hub
-            // causing next message being replayed and all subsequent messages being queued.
-            await CompleteIncomingMessageAsync(message).ConfigureAwait(false);
+            await TaskHelpers.CompletedTask.ConfigureAwait(false);
 
             if (Logging.IsEnabled)
                 Logging.Exit(this, "Process C2D message via callback", nameof(HandleIncomingMessagesAsync));
--- a/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
+++ b/e2e/test/iothub/messaging/MessageReceiveE2ETests.cs
@@ -938,6 +938,9 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
                 async (message, context) =>
                 {
                     VerboseTestLogger.WriteLine($"Received message over the first message handler: MessageId={message.MessageId}");
+                    var messageD2C = new Client.Message(Encoding.UTF8.GetBytes("DeviceToCloud"));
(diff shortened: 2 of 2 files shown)


patilsnr (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -578,9 +578,7 @@ private async Task HandleIncomingMessagesAsync()
             // We are intentionally not awaiting _deviceMessageReceivedListener callback.
             // This is a user-supplied callback that isn't required to be awaited by us. We can simply invoke it and continue.
             _ = _deviceMessageReceivedListener?.Invoke(message);
-            // Messages with QoS = 1 need to be Acknowledged otherwise it results in mismatched Ack to IoT Hub
-            // causing next message being replayed and all subsequent messages being queued.
-            await CompleteIncomingMessageAsync(message).ConfigureAwait(false);
```
abhipsaMisra, line 583: The comment states that this completion is required for QoS 1 messages which would otherwise cause messages to keep being replayed (if, for instance, the user doesn't explicitly mark the message as complete). Is the issue stemming from the fact that both the SDK and user are trying to complete the same message?

abhipsaMisra: Also, doesn't the "complete" API take in the message/lock token as the input? Does the internal queue synchronization not take into account the passed in lock token?

tmahmood-microsoft: Yes, I added this change a few months ago to fix M2M message failure after direct method. So, this change was specifically for HandleIncomingMethodPostAsync() and HandleIncomingTwinPatchAsync()

abhipsaMisra: So, my question is - is the previous comment added here invalid? Is the QoS 1 ack required only for methods and twin?

tmahmood-microsoft: Yes, the previous comment here is incorrect. The SDK trying to complete QoS 1 message internally is not needed.

abhipsaMisra: Alright. I'd recommend making a note to add the above module client scenarios so that we can have the tests validate the behavior for us.

abhipsaMisra (approved):

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -578,9 +578,7 @@ private async Task HandleIncomingMessagesAsync()
...
+            await TaskHelpers.CompletedTask.ConfigureAwait(false);
```
drwill-ms, line 581: Why await this instead of removing the `async` keyword from the method?

Issue #3115, opened 2023-02-15
[Bug Report] Invalid cast on CreationTimeUtc


nikoraes:
# Context

- **Application's .NET Target Framework :** net6
- **SDK version used:** **v2.0.0-preview002**
- **Transport:** AMQP (TCP) 

## Description of the issue
CreationTimeUtc type in Microsoft.Azure.Devices.Client.TelemetryMessage is of DateTimeOffset type.
[UpdateAmqpMessageHeadersAndProperties](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/device/src/Transport/AmqpIot/AmqpIotMessageConverter.cs) tries to cast the CreationTimeUtc system property to DateTime (before converting it to an ISO string), which doesn't work.
Would be better to use the typed properties of the TelemetryMessage instead of casting from the SystemProperties.

## Code sample exhibiting the issue
```
DateTimeOffset creationTimeUtc = ...
var message = new TelemetryMessage(payload)
{
     CreatedOnUtc = creationTimeUtc
};
deviceClient.SendTelemetryAsync(message);
```

## Console log of the issue
Follow the instructions [here](https://github.com/Azure/azure-iot-sdk-csharp/tree/main/tools/CaptureLogs) to capture SDK logs.
Don't forget to remove any connection string information!

```
Exception has occurred: CLR/System.InvalidCastException
Exception thrown: 'System.InvalidCastException' in System.Private.CoreLib.dll: 'Unable to cast object of type 'System.DateTimeOffset' to type 'System.DateTime'.'
   at Microsoft.Azure.Devices.Client.Transport.AmqpIot.AmqpIotMessageConverter.UpdateAmqpMessageHeadersAndProperties(AmqpMessage amqpMessage, TelemetryMessage data, Boolean copyUserProperties)
   at Microsoft.Azure.Devices.Client.Transport.AmqpIot.AmqpIotMessageConverter.OutgoingMessageToAmqpMessage(TelemetryMessage message)
   at Microsoft.Azure.Devices.Client.Transport.AmqpIot.AmqpIotSendingLink.<SendMessageAsync>d__9.MoveNext()
   at Microsoft.Azure.Devices.Client.Transport.AmqpIot.AmqpUnit.<SendMessageAsync>d__39.MoveNext()
   at Microsoft.Azure.Devices.Client.Transport.AmqpIot.AmqpUnit.<SendTelemetryAsync>d__48.MoveNext()
[...]

PR #3117, opened 2023-02-16
Fix bug for CreatedOnUtc DateTimeOffset in TelemetryMessage

abhipsaMisra:
Fix for #3115 : Invalid cast on CreationTimeUtc

### DIFF
--- a/e2e/test/helpers/RetryOperationHelper.cs
+++ b/e2e/test/helpers/RetryOperationHelper.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
                     await asyncOperation().ConfigureAwait(false);
                     return;
                 }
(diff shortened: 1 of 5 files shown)


schoims (approved):

brycewang-microsoft (approved):

tmahmood-microsoft (approved):

e2e/test/helpers/RetryOperationHelper.cs
```diff
@@ -78,7 +78,7 @@ public class RetryOperationHelper
                     await asyncOperation().ConfigureAwait(false);
                     return;
                 }
-                catch (Exception ex) when (!retryPolicy.ShouldRetry(counter, ex, out retryInterval))
+                catch (Exception ex) when (retryPolicy.ShouldRetry(counter, ex, out retryInterval))
```
drwill-ms, line 81: This doesn't seem right to me.

If we should retry, then we should not let the exception be thrown.

PR #3130, opened 2023-02-24
Fix test retry

drwill-ms:
Bad logic causing retry not to happen when device not found due to slow device registry.

### DIFF
--- a/e2e/test/helpers/HubServiceTestRetryPolicy.cs
+++ b/e2e/test/helpers/HubServiceTestRetryPolicy.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Helpers
         {
             retryInterval = TimeSpan.Zero;
 
-            if (currentRetryCount < MaxRetries)
+            if (currentRetryCount > MaxRetries)
             {
                 VerboseTestLogger.WriteLine($"{nameof(HubServiceTestRetryPolicy)}: Exhausted {currentRetryCount}/{MaxRetries} retries and failing due to {lastException}");
                 return false;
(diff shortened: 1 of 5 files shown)


e2e/test/helpers/HubServiceTestRetryPolicy.cs
```diff
@@ -22,7 +22,7 @@ public bool ShouldRetry(uint currentRetryCount, Exception lastException, out Tim
         {
             retryInterval = TimeSpan.Zero;
 
-            if (currentRetryCount < MaxRetries)
+            if (currentRetryCount > MaxRetries)
```
drwill-ms, line 25: This was the oopsie.

e2e/test/helpers/TestDevice.cs
```diff
@@ -52,25 +48,14 @@ private TestDevice(Device device, IAuthenticationMethod authenticationMethod)
         /// <param name="type">The way the device will authenticate</param>
         public static async Task<TestDevice> GetTestDeviceAsync(string namePrefix, TestDeviceType type = TestDeviceType.Sasl)
         {
-            string prefix = namePrefix + type + "_";
-
-            try
-            {
-                await s_semaphore.WaitAsync().ConfigureAwait(false);
```
drwill-ms, line 59: Why do we have a semaphore here??

timtay-microsoft: The service client operations needed to create a new test device should already be thread safe, so I agree with removing this semaphore

timtay-microsoft (approved):

e2e/test/helpers/TestDevice.cs
timtay-microsoft: It's just an HTTP request, after all

brycewang-microsoft (approved):

abhipsaMisra (approved):

andyk-ms (approved):

PR #3134, opened 2023-02-27
Correct test job creation error handling

drwill-ms:


abhipsaMisra (approved):

brycewang-microsoft (approved):

iothub/service/src/Exceptions/IotHubServiceErrorCode.cs
```diff
@@ -188,10 +196,20 @@ public enum IotHubServiceErrorCode
...
         /// </summary>
         ThrottlingException = 429001,
 
+        /// <summary>
+        /// IoT hub throttling limits have been exceeded for the requested operation.
+        /// <para>
+        /// For more information, <see href="https://aka.ms/iothubthrottling">IoT hub quotas and throttling</see>.
+        /// </para>
+        /// </summary>
+        ThrottlingBacklogTimeout = 429003,
```
abhipsaMisra, line 211: Oh wow, service has multiple error subcodes for throttling! I wish they had added some more context in their swagger files.

iothub/service/src/Exceptions/IotHubServiceException.cs
```diff
@@ -109,10 +109,12 @@ private static bool DetermineIfTransient(HttpStatusCode statusCode, IotHubServic
                 IotHubServiceErrorCode.IotHubQuotaExceeded
                     or IotHubServiceErrorCode.DeviceNotOnline
                     or IotHubServiceErrorCode.ThrottlingException
+                    or IotHubServiceErrorCode.ThrottlingBacklogTimeout
                     or IotHubServiceErrorCode.ServerError
                     or IotHubServiceErrorCode.ServiceUnavailable
                     => true,
-                IotHubServiceErrorCode.Unknown => statusCode == HttpStatusCode.RequestTimeout,
+                IotHubServiceErrorCode.Unknown => statusCode == HttpStatusCode.RequestTimeout
+                    || statusCode == (HttpStatusCode)429,
```
abhipsaMisra, line 117: Isn't this getting mapped to "IotHubServiceErrorCode.ThrottlingException"? If it isn't already then should we?

iothub/service/src/Exceptions/IotHubServiceErrorCode.cs
brycewang-microsoft, line 211: Are we aware of any difference between these two?

drwill-ms: I asked Andrei. He's not sure.

iothub/service/src/Exceptions/IotHubServiceException.cs
drwill-ms: It might be, but it occurred to me we wouldn't have had this issue if we had this line.

iothub/service/src/Exceptions/IotHubServiceErrorCode.cs
drwill-ms: Totally

iothub/service/src/Exceptions/IotHubServiceException.cs
drwill-ms: Sorry 429 is not associated. We parse the error code (string, not the numeric status code). That is 429001.

iothub/service/src/Exceptions/IotHubServiceErrorCode.cs
drwill-ms: I figured, for now, I'd duplicate the doc comments. It is upsetting that we have 2 error codes for being throttled, but I can't change what the service returns, unless we want to find a way to remap it on the client (which I'm wary to do).

PR #3135, opened 2023-02-27
Decouple client open-close semaphore from callback subscription semaphore

abhipsaMisra:
The sepmaphores associated with enabling and disabling subscriptions for callbacks should not be dependent on the Semaphore used for opening or closing the client.
The only purpose of the enable-disable events semaphore is to ensure that requests to enable and disable the subscription invoked by parallel threads are executed in a thread-safe manner.

### DIFF
--- a/iothub/device/src/Pipeline/DefaultDelegatingHandler.cs
+++ b/iothub/device/src/Pipeline/DefaultDelegatingHandler.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
 {
     internal abstract class DefaultDelegatingHandler : IDelegatingHandler
     {
-        protected const string ClientDisposedMessage = "The client has already been disposed and is no longer usable.";
+        protected const string ClientDisposedMessage = "The client has been disposed and is no longer usable.";
         protected volatile bool _isDisposed;
         private volatile IDelegatingHandler _innerHandler;
 
--- a/iothub/device/src/Pipeline/DefaultDelegatingHandler.cs
+++ b/iothub/device/src/Pipeline/DefaultDelegatingHandler.cs
@@ -11,8 +11,9 @@ namespace Microsoft.Azure.Devices.Client.Transport
 {
     internal abstract class DefaultDelegatingHandler : IDelegatingHandler
     {
-        private volatile IDelegatingHandler _innerHandler;
+        protected const string ClientDisposedMessage = "The client has already been disposed and is no longer usable.";
         protected volatile bool _isDisposed;
+        private volatile IDelegatingHandler _innerHandler;
 
         protected DefaultDelegatingHandler(PipelineContext context, IDelegatingHandler innerHandler)
         {
@@ -209,7 +210,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
         {
             if (_isDisposed)
             {
-                throw new ObjectDisposedException("IoT hub client");
+                throw new ObjectDisposedException("IoT client", ClientDisposedMessage);
             }
         }
 
--- a/iothub/device/src/Pipeline/RetryDelegatingHandler.cs
+++ b/iothub/device/src/Pipeline/RetryDelegatingHandler.cs
@@ -788,7 +788,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
                     Logging.Enter(this, cancellationToken, nameof(CloseAsync));
 
                 _handleDisconnectCts.Cancel();
-                _cancelPendingOperationsCts.Cancel();
+                _cancelPendingOperationsCts.Cancel(); // Is it possible that this is already disposed by the time close was still waiting to acquire the semaphore?
                 await base.CloseAsync(cancellationToken).ConfigureAwait(false);
             }
             finally
--- a/iothub/device/src/Pipeline/RetryDelegatingHandler.cs
+++ b/iothub/device/src/Pipeline/RetryDelegatingHandler.cs
@@ -32,6 +32,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
         private bool _twinEnabled;
         private bool _eventsEnabled;
         private bool _deviceReceiveMessageEnabled;
+        private bool _isDisposing;
         private bool _isAnEdgeModule = true;
 
         private Task _transportClosedTask;
@@ -246,7 +247,16 @@ namespace Microsoft.Azure.Devices.Client.Transport
                             }
                             finally
                             {
-                                _cloudToDeviceMessageSubscriptionSemaphore?.Release();
+                                try
+                                {
+                                    _cloudToDeviceMessageSubscriptionSemaphore?.Release();
+                                }
+                                catch (ObjectDisposedException) when (_isDisposing)
+                                {
+                                    if (Logging.IsEnabled)
+                                        Logging.Error(this, "Tried releasing cloud-to-device message subscription semaphore but it has already been disposed by client disposal on a separate thread." +
+                                            "Ignoring this exception and continuing with client cleanup.");
+                                }
                             }
                         },
                         operationCts.Token)
@@ -287,7 +297,16 @@ namespace Microsoft.Azure.Devices.Client.Transport
                             }
                             finally
                             {
-                                _cloudToDeviceMessageSubscriptionSemaphore?.Release();
+                                try
+                                {
+                                    _cloudToDeviceMessageSubscriptionSemaphore?.Release();
+                                }
+                                catch (ObjectDisposedException) when (_isDisposing)
+                                {
+                                    if (Logging.IsEnabled)
+                                        Logging.Error(this, "Tried releasing cloud-to-device message subscription semaphore but it has already been disposed by client disposal on a separate thread." +
+                                            "Ignoring this exception and continuing with client cleanup.");
+                                }
                             }
                         },
                         operationCts.Token)
@@ -326,7 +345,16 @@ namespace Microsoft.Azure.Devices.Client.Transport
                             }
                             finally
                             {
-                                _cloudToDeviceMessageSubscriptionSemaphore?.Release();
+                                try
+                                {
+                                    _cloudToDeviceMessageSubscriptionSemaphore?.Release();
+                                }
+                                catch (ObjectDisposedException) when (_isDisposing)
+                                {
+                                    if (Logging.IsEnabled)
+                                        Logging.Error(this, "Tried releasing cloud-to-device message subscription semaphore but it has already been disposed by client disposal on a separate thread." +
+                                            "Ignoring this exception and continuing with client cleanup.");
+                                }
                             }
                         },
                         operationCts.Token)
@@ -362,7 +390,16 @@ namespace Microsoft.Azure.Devices.Client.Transport
                             }
                             finally
                             {
-                                _directMethodSubscriptionSemaphore?.Release();
+                                try
+                                {
+                                    _directMethodSubscriptionSemaphore?.Release();
+                                }
+                                catch (ObjectDisposedException) when (_isDisposing)
+                                {
+                                    if (Logging.IsEnabled)
+                                        Logging.Error(this, "Tried releasing direct method subscription semaphore but it has already been disposed by client disposal on a separate thread." +
+                                            "Ignoring this exception and continuing with client cleanup.");
+                                }
                             }
                         },
                         operationCts.Token)
@@ -398,7 +435,16 @@ namespace Microsoft.Azure.Devices.Client.Transport
                             }
                             finally
                             {
-                                _directMethodSubscriptionSemaphore?.Release();
+                                try
+                                {
+                                    _directMethodSubscriptionSemaphore?.Release();
+                                }
+                                catch (ObjectDisposedException) when (_isDisposing)
+                                {
+                                    if (Logging.IsEnabled)
+                                        Logging.Error(this, "Tried releasing direct method subscription semaphore but it has already been disposed by client disposal on a separate thread." +
+                                            "Ignoring this exception and continuing with client cleanup.");
+                                }
                             }
                         },
                         operationCts.Token)
@@ -435,7 +481,16 @@ namespace Microsoft.Azure.Devices.Client.Transport
                             }
                             finally
                             {
-                                _cloudToDeviceEventSubscriptionSemaphore?.Release();
+                                try
+                                {
+                                    _cloudToDeviceEventSubscriptionSemaphore?.Release();
+                                }
+                                catch (ObjectDisposedException) when (_isDisposing)
+                                {
+                                    if (Logging.IsEnabled)
+                                        Logging.Error(this, "Tried releasing cloud-to-device event subscription semaphore but it has already been disposed by client disposal on a separate thread." +
+                                            "Ignoring this exception and continuing with client cleanup.");
+                                }
                             }
                         },
                         operationCts.Token)
@@ -472,7 +527,16 @@ namespace Microsoft.Azure.Devices.Client.Transport
                             }
                             finally
                             {
-                                _cloudToDeviceEventSubscriptionSemaphore?.Release();
+                                try
+                                {
+                                    _cloudToDeviceEventSubscriptionSemaphore?.Release();
+                                }
+                                catch (ObjectDisposedException) when (_isDisposing)
+                                {
+                                    if (Logging.IsEnabled)
+                                        Logging.Error(this, "Tried releasing cloud-to-device event subscription semaphore but it has already been disposed by client disposal on a separate thread." +
+                                            "Ignoring this exception and continuing with client cleanup.");
+                                }
                             }
                         },
                         operationCts.Token)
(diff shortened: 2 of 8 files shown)


iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -21,7 +21,11 @@ internal class RetryDelegatingHandler : DefaultDelegatingHandler
 
         private RetryPolicy _internalRetryPolicy;
 
-        private SemaphoreSlim _handlerSemaphore = new SemaphoreSlim(1, 1);
+        private SemaphoreSlim _clientOpenCloseSemaphore = new SemaphoreSlim(1, 1);
```
drwill-ms, line 24: readonly?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -219,8 +223,8 @@ await _internalRetryPolicy
                             // Ensure that the connection has been opened, before enabling the callback for receiving messages.
                             await EnsureOpenedAsync(false, cancellationToken).ConfigureAwait(false);
 
-                            // Wait to acquire the _handlerSemaphore. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
-                            await _handlerSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                            // Wait to acquire the _cloudToDeviceSubscriptionSemaphore. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
+                            await _cloudToDeviceMessageSubscriptionSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
drwill-ms, line 227: Why null check on 237 but not here? Do we ever expect this to be null?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -259,8 +263,8 @@ await _internalRetryPolicy
                             // Ensure that the connection has been opened before returning pending messages to the callback.
                             await EnsureOpenedAsync(false, cancellationToken).ConfigureAwait(false);
 
-                            // Wait to acquire the _handlerSemaphore. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
-                            await _handlerSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                            // Wait to acquire the _cloudToDeviceSubscriptionSemaphore. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
```
drwill-ms, line 266:
```suggestion
                            // Wait to acquire the _cloudToDeviceMessageSubscriptionSemaphore. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
```

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -297,8 +301,8 @@ await _internalRetryPolicy
                             // Ensure that the connection has been opened, before disabling the callback for receiving messages.
                             await EnsureOpenedAsync(false, cancellationToken).ConfigureAwait(false);
 
-                            // Wait to acquire the _handlerSemaphore. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
-                            await _handlerSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                            // Wait to acquire the _cloudToDeviceSubscriptionSemaphore. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
```
drwill-ms, line 304:
```suggestion
                            // Wait to acquire the _cloudToDeviceMessageSubscriptionSemaphore. This ensures that concurrently invoked API calls are invoked in a thread-safe manner.
```

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -1090,12 +1091,41 @@ protected override void Dispose(bool disposing)
...
                         _handleDisconnectCts?.Cancel();
                         _handleDisconnectCts?.Dispose();
-                        if (_handlerSemaphore != null && _handlerSemaphore.CurrentCount == 0)
+
+                        if (_clientOpenCloseSemaphore != null && _clientOpenCloseSemaphore.CurrentCount == 0)
+                        {
+                            _clientOpenCloseSemaphore.Release();
+                        }
+                        _clientOpenCloseSemaphore?.Dispose();
+                        _clientOpenCloseSemaphore = null;
```
drwill-ms, line 1100: Maybe we shouldn't assign to null so we get an object disposed exception instead of NRE.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -1090,12 +1091,41 @@ protected override void Dispose(bool disposing)
                     {
                         _handleDisconnectCts?.Cancel();
                         _handleDisconnectCts?.Dispose();
-                        if (_handlerSemaphore != null && _handlerSemaphore.CurrentCount == 0)
+
+                        if (_clientOpenCloseSemaphore != null && _clientOpenCloseSemaphore.CurrentCount == 0)
+                        {
+                            _clientOpenCloseSemaphore.Release();
```
drwill-ms, line 1097: We've seen timing issues recently where we get an error here. I'd say wrap in a catch, maybe?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: yeah sure, it can be readonly

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Dispose() will set the semaphores to null and I don't expect this code to be hit asfter Dispose() has been called.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: I like that, sure. We're not relying on this field being null to indicate any special scenario, so I see no issues in us disposing this semaphore and then leaving it to the garbage collector to eventually clear up the references.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: This release is attempted only if the above expression indicates that the semaphore is currently held by a thread. This shouldn't give us any error. What do you think?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: I don't love the assumption, but I'll leave it to you.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: I'm saying we've seen errors in our pipeline that it couldn't release because it would exceed the count. Meaning the code sections above that had the semaphore were still in it, our if statement confirmed they still had it, we tried to release it for them (suspect anyway), but before we could they released it.

I'm thinking we should either not release it here (just dispose), or if we do, we should do each of them inside a try/catch.

abhipsaMisra: I can add some additional checks around this cleanup code so that failures here don't crash the application, sure.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -76,20 +82,22 @@ public override async Task SendEventAsync(Message message, CancellationToken can
                 if (Logging.IsEnabled)
                     Logging.Enter(this, message, cancellationToken, nameof(SendEventAsync));
 
+                using var operationCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _cancelPendingOperations.Token);
+
                 await _internalRetryPolicy
                     .RunWithRetryAsync(
                         async () =>
                         {
```
drwill-ms, line 90: Should we consider checking if the cancellation token is signaled and throw before retry attempt?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -1085,20 +1241,25 @@ protected override void Dispose(bool disposing)
 
                 if (!_isDisposed)
                 {
+                    _isDisposing = true;
+
                     base.Dispose(disposing);
                     if (disposing)
                     {
                         _handleDisconnectCts?.Cancel();
```
drwill-ms, line 1249: We have 2 places where we cancel, which means one of them will get an exception, which we should gracefully handle.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -1085,20 +1241,25 @@ protected override void Dispose(bool disposing)
...
                         _handleDisconnectCts?.Dispose();
-                        if (_handlerSemaphore != null && _handlerSemaphore.CurrentCount == 0)
-                        {
-                            _handlerSemaphore.Release();
-                        }
-                        _handlerSemaphore?.Dispose();
-                        _handlerSemaphore = null;
+
+                        _cancelPendingOperations?.Cancel();
```
drwill-ms, line 1252: Is there any chance one of these could throw, and we'd fail to dispose the remaining set?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -1085,20 +1241,25 @@ protected override void Dispose(bool disposing)
 
                 if (!_isDisposed)
                 {
+                    _isDisposing = true;
+
                     base.Dispose(disposing);
                     if (disposing)
                     {
                         _handleDisconnectCts?.Cancel();
                         _handleDisconnectCts?.Dispose();
```
drwill-ms, line 1250: If the CTS is disposed, will the cancellation token still be valid (used by the async tasks that are watching it, or might they throw when the token from it is accessed)?

drwill-ms: It works fine.

```csharp
namespace CtsDisposed
{
    public static class Program
    {
        public static async Task Main()
        {
            var cts = new CancellationTokenSource();
            Task work = DoWorkAsync(cts.Token);
            await Task.Delay(5000).ConfigureAwait(false);
            cts.Cancel();
            try
            {
                await work.ConfigureAwait(false);
            }
            catch (Exception ex)
            {
                await Console.Out.WriteLineAsync($"Caught {ex}");
            }
        }

        private static async Task DoWorkAsync(CancellationToken ct)
        {
            try
            {
                await Task.Delay(-1, ct).ConfigureAwait(false);
            }
            catch (OperationCanceledException)
            {
                await Console.Out.WriteLineAsync("DoWork cancelled");
            }
        }
    }
}
```

drwill-ms (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Looks like `CancellationTokenSource` gracefully handles multiple calls to the `Cancel()` method. From the [sourcecode](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs):

```c#
public void Cancel() => Cancel(false);

public void Cancel(bool throwOnFirstException)
{
	ThrowIfDisposed();
	NotifyCancellation(throwOnFirstException);
}

private void NotifyCancellation(bool throwOnFirstException)
{
	if (TransitionToCancellationRequested())
	{
		// If we're the first to signal cancellation, do the main extra work.
		ExecuteCallbackHandlers(throwOnFirstException);
		Debug.Assert(IsCancellationCompleted, "Expected cancellation to have finished");
	}
}

private bool TransitionToCancellationRequested()
{
	if (!IsCancellationRequested &&
		Interlocked.CompareExchange(ref _state, NotifyingState, NotCanceledState) == NotCanceledState)
	{
		// Dispose of the timer, if any.  Dispose may be running concurrently here, but TimerQueueTimer.Close is thread-safe.
		TimerQueueTimer? timer = _timer;
		if (timer != null)
		{
			_timer = null;
			timer.Close();
		}

		// Set the event if it's been lazily initialized and hasn't yet been disposed of.  Dispose may
		// be running concurrently, in which case either it'll have set m_kernelEvent back to null and
		// we won't see it here, or it'll see that we've transitioned to NOTIFYING and will skip disposing it,
		// leaving cleanup to finalization.
		_kernelEvent?.Set();

		return true;
	}

	return false;
}


```

The [first check](https://github.com/dotnet/runtime/blob/0a15c3b97dabaf94e5d4399aef2fd48c2b3f446a/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs#L676) in `TransitionToCancellationRequested` returns `false`, which will result in each loop being gracefully exited immediately.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: I can wrap these calls within try-catch block

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: The code comments in the CancellationTokenSource [sourcecode](https://github.com/dotnet/runtime/blob/0a15c3b97dabaf94e5d4399aef2fd48c2b3f446a/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs#L505-L510) supports your test:

```
// And, we tolerate Dispose being used concurrently with Cancel.  This is necessary
// to properly support, e.g., LinkedCancellationTokenSource, where, due to common usage patterns,
// it's possible for this pairing to occur with valid usage (e.g. a component accepts
// an external CancellationToken and uses CreateLinkedTokenSource to combine it with an
// internal source of cancellation, then Disposes of that linked source, which could
// happen at the same time the external entity is requesting cancellation).
```

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: We have this custom logic in our retry policy implementation that makes checking for cancellation token validity prior to entering the retry block a bit cumbersome to add; i.e. we'd need to add a similar translation of exception in order to preserve the behavior.
https://github.com/Azure/azure-iot-sdk-csharp/blob/480a27ba6a5a4486d2514b48cfefd5d42c30a6c1/iothub/device/src/TransientFaultHandling/RetryPolicy.cs#L281-L294

@drwill-ms: I inclined to leave the code in `main` as it is and introduce this change in `previews/v2`. What do you think?

drwill-ms: I'm fine with leaving main mostly the same.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -828,7 +819,7 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
                 return;
             }
 
-            await _clientOpenCloseSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+            await _clientOpenCloseSemaphore.WaitAsync(operationCts.Token).ConfigureAwait(false);
```
drwill-ms, line 822: Rename to `_clientOpenSemaphore`.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -882,17 +873,24 @@ private async Task EnsureOpenedAsync(bool withRetry, CancellationToken cancellat
...
+            // The current behavior does not support open->close->open
+            if (_isDisposed)
             {
-                return;
+                throw new ObjectDisposedException(nameof(RetryDelegatingHandler));
             }
 
-            bool gain = await _clientOpenCloseSemaphore.WaitAsync(timeoutHelper.GetRemainingTime()).ConfigureAwait(false);
-            if (!gain)
+            if (Volatile.Read(ref _opened))
```
drwill-ms, line 887: Consider what we read about Eric Lippert's take on `volatile` and instead use `Interlocked.Exchange`.

drwill-ms (approved):

iothub/device/src/Pipeline/DefaultDelegatingHandler.cs
```diff
@@ -11,8 +11,9 @@ namespace Microsoft.Azure.Devices.Client.Transport
 {
     internal abstract class DefaultDelegatingHandler : IDelegatingHandler
     {
-        private volatile IDelegatingHandler _innerHandler;
+        protected const string ClientDisposedMessage = "The client has already been disposed and is no longer usable.";
```
drwill-ms, line 14:
```suggestion
protected const string ClientDisposedMessage = "The client has been disposed and is no longer usable.";
```

drwill-ms (approved):

PR #3141, opened 2023-03-01
Name change for edge module client method to send message(s)

drwill-ms:
Feedback from Rido about making these method names clear that they only pertain to Edge modules for module-to-module messages


iothub/device/src/IotHubModuleClient.cs
```diff
@@ -170,10 +170,10 @@ public async Task SendTelemetryAsync(string outputName, TelemetryMessage message
         /// <returns>The task containing the event</returns>
         /// <exception cref="InvalidOperationException">Thrown if IotHubModuleClient instance is not opened already.</exception>
         /// <exception cref="OperationCanceledException">Thrown when the operation has been canceled.</exception>
-        public async Task SendTelemetryBatchAsync(string outputName, IEnumerable<TelemetryMessage> messages, CancellationToken cancellationToken = default)
+        public async Task SendMessagesToRouteAsync(string outputName, IEnumerable<TelemetryMessage> messages, CancellationToken cancellationToken = default)
```
abhipsaMisra, line 173: Are we deliberately keeping the name the same for single message and batch message?

abhipsaMisra: Ah, I missed the "s" at the end of "message"

abhipsaMisra: The API to send TelemetryMessage to IoT Hub directly is called "SendTelemetryAsync". Should we use the term "Telemetry" instead of "Message" in here as well? The data type involved is a "TelemetryMessage" in both cases.

abhipsaMisra (approved):

iothub/device/src/IotHubModuleClient.cs
drwill-ms: @rido-min doesn't see these messages as telemetry

rido-min: My understanding of Telemetry (in the context of IoTHub), is for MQTT messages published (D2C) to `devices/{clientId}/messages/events/` that can be routed to a configured BuiltIn-Endpoint/Route by using a Data source of type `Device Telemetry Messages`

My suggestion of changing the name of this overload is based on the assumption that D2C messages sent with this overload won't be able to be routed as Telemetry.. Let me run a quick test and I'll get back to this thread.

rido-min: just checked the behavior in hub with a device (non-edge) module.

The value used in `outputName` is part of the message metadata as:

```json
 "properties": {
            "application": {
                "$.on": "myOutput"
            }
        },
```

let me review the full edge module scenario, including routing.. maybe SendTelemetry is not so bad..

rido-min: also wondering if DeviceClient should also accept an output name

drwill-ms: What is the user scenario for adding `outputName` parameter to IotHubDeviceClient telemetry API?

rido-min: I dont know the use case, just trying to reflect the IoTHub features wrt messaging to have consistency in Device/Module clients.

drwill-ms: I've added the property to `TelemetryMessage` to allow someone to set it when sending from a device client, but I wouldn't go so far as to add a parameter for it as it is not a mainline scenario. Does that work for you?

iothub/device/src/Messaging/TelemetryMessage.cs
```diff
@@ -167,6 +167,18 @@ public string InputName
...
+        /// <summary>
+        /// Specifies the output name on which the message will be sent, if applicable.
+        /// </summary>
+        /// <remarks>
+        /// Used for message routes with IoT Edge.
+        /// </remarks>
+        public string OutputName
+        {
+            get => GetSystemProperty<string>(MessageSystemPropertyNames.OutputName);
+            protected internal set => SystemProperties[MessageSystemPropertyNames.OutputName] = value;
```
rido-min, line 179: public set

rido-min (approved):
public set in msg.OutputName

PR #3145, opened 2023-03-03
Make DPS tests more resilient to device registry delays

drwill-ms:


e2e/test/helpers/TestDevice.cs
```diff
@@ -34,7 +34,7 @@ public sealed class TestDevice : IDisposable
             });
 
         private X509Certificate2 _authCertificate;
-        private static readonly IotHubServiceClient _client = new(TestConfiguration.IotHub.ConnectionString);
+        internal static IotHubServiceClient ServiceClient { get; } = new(TestConfiguration.IotHub.ConnectionString);
```
drwill-ms, line 37: Make this public so we can use the same instance from tests.

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -896,44 +896,34 @@ public static ProvisioningClientOptions CreateProvisioningClientOptionsFromName(
...
+                {
+                    await iotClient.OpenAsync().ConfigureAwait(false);
+                    break;
+                }
+                catch (Exception ex) when (attempt < 4)
                 {
-                    [new Guid().ToString()] = new Guid().ToString(),
-                };
-                await iotClient.UpdateReportedPropertiesAsync(propertiesToReport).ConfigureAwait(false);
+                    VerboseTestLogger.WriteLine($"Attempt #{attempt} failed to open device {result.DeviceId} connection due to {ex}");
```
drwill-ms, line 913: Try a few times to open the client. We keep seeing an error in some PR builds where the device creds are rejected (so OpenAsync doesn't retry).

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -896,44 +896,34 @@ public static ProvisioningClientOptions CreateProvisioningClientOptionsFromName(
...
         {
-            if (capabilities != null && capabilities.IsIotEdge)
+            if (capabilities != null)
             {
-                //If device is edge device, it should be able to connect to iot hub as its edgehub module identity
-                var iotHubConnectionCredentials = new IotHubConnectionCredentials(auth, result.AssignedHub);
-                string edgehubConnectionString = iotHubConnectionCredentials.GetIotHubConnectionString() + ";ModuleId=$edgeHub";
-                await using var moduleClient = new IotHubModuleClient(edgehubConnectionString);
-                await moduleClient.OpenAsync().ConfigureAwait(false);
+                ClientTwin twin = await TestDevice.ServiceClient.Twins.GetAsync(result.DeviceId).ConfigureAwait(false);
```
drwill-ms, line 925: To validate, all we need to do is check that the Hub's twin matches what was specified in the DPS registration initial twin.

Is this testing the service? Yes. But this is also testing that we've serialized the initial twin, so I've left the validation in place.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -550,7 +550,10 @@ public override async Task StopSasTokenLoopAsync()
 
                 // Await the completion of _refreshLoop.
                 // This will ensure that when StopLoopAsync has been exited then no more token refresh attempts are in-progress.
-                await _refreshLoop.ConfigureAwait(false);
+                if (_refreshLoop != null)
+                {
+                    await _refreshLoop.ConfigureAwait(false);
```
drwill-ms, line 555: Through debugging these tests, I found this sometimes failed with an NRE.

provisioning/device/src/Transports/Amqp/ClientWebSocketTransport.cs
```diff
@@ -326,8 +326,7 @@ private void ThrowIfNotOpen()
 
         public void Dispose()
         {
-            _writeCancellationTokenSource?.Dispose();
-            _writeCancellationTokenSource = null;
```
drwill-ms, line 330: Avoid an NRE.

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -896,44 +896,34 @@ public static ProvisioningClientOptions CreateProvisioningClientOptionsFromName(
...
+                    await iotClient.OpenAsync().ConfigureAwait(false);
+                    break;
+                }
+                catch (Exception ex) when (attempt < 4)
                 {
-                    [new Guid().ToString()] = new Guid().ToString(),
-                };
-                await iotClient.UpdateReportedPropertiesAsync(propertiesToReport).ConfigureAwait(false);
+                    VerboseTestLogger.WriteLine($"Attempt #{attempt} failed to open device {result.DeviceId} connection due to {ex}");
+                    await Task.Delay(1000).ConfigureAwait(false);
```
timtay-microsoft, line 914: Missing an ```attempt++``` somewhere in this loop?

timtay-microsoft (approved):

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -896,44 +896,34 @@ public static ProvisioningClientOptions CreateProvisioningClientOptionsFromName(
...
             {
-                VerboseTestLogger.WriteLine("DeviceClient updating desired properties.");
-                TwinProperties twin = await iotClient.GetTwinPropertiesAsync().ConfigureAwait(false);
-                var propertiesToReport = new ReportedProperties
+                try
+                {
+                    await iotClient.OpenAsync().ConfigureAwait(false);
+                    break;
+                }
+                catch (Exception ex) when (attempt++ < 4)
```
abhipsaMisra, line 911: Do we want to retry all scenarios or just the specific auth error that we've been seeing? I think it'd be good to add the comment you have in the PR below into the code comments here.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Oh, I wonder why.

abhipsaMisra (approved):

e2e/test/provisioning/ProvisioningE2ETests.cs
drwill-ms: I couldn't identify the auth error from the callstack.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: User calls OpenAsync, but it fails, so the call to set the SasTokenRefreshesOn doesn't get called. But the RetryDelegatingHandler still exists, so if the client is closed, we call StopSasTokenLoopAsync on the inner handler.

e2e/test/provisioning/ProvisioningE2ETests.cs
drwill-ms: We don't print the error code on ToString :(

e2e/test/provisioning/ProvisioningE2ETests.cs
```diff
@@ -896,44 +896,34 @@ public static ProvisioningClientOptions CreateProvisioningClientOptionsFromName(
         private static async Task ConfirmRegisteredDeviceWorksAsync(
             DeviceRegistrationResult result,
             IAuthenticationMethod auth,
-            IotHubClientTransportSettings transportSettings,
-            bool sendReportedPropertiesUpdate)
+            IotHubClientTransportSettings transportSettings)
         {
             await using var iotClient = new IotHubDeviceClient(result.AssignedHub, auth, new IotHubClientOptions(transportSettings));
```
brycewang-microsoft, line 901: can we rename `iotClient` to something else, like `deviceClient`?

e2e/test/provisioning/ProvisioningE2ETests.cs
drwill-ms: Can we guess what it was? Maybe `IotHubClientErrorCode.Unauthorized`?

brycewang-microsoft (approved):

e2e/test/provisioning/ProvisioningE2ETests.cs
drwill-ms: I'll change it, but in a different PR unless I need to restart gates.

e2e/test/provisioning/ProvisioningE2ETests.cs
abhipsaMisra: Yes, it should be IotHubClientErrorCode.Unauthorized.

https://github.com/Azure/azure-iot-sdk-csharp/blob/317b9a3e4558f871ee31b9088c8b9222bde360e0/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs#L317-L323

e2e/test/helpers/TestDevice.cs
```diff
@@ -140,32 +81,70 @@ public string ConnectionString
...
+                await OpenWithRetryAsync().ConfigureAwait(false);
+            }
+
+            return DeviceClient;
+        }
+
+        /// Sometimes the just newly created device can fail to connect if there is lag in the hub for the identity existing.
+        /// Retry a few times to prevent test failures.
+        /// </summary>
+        public async Task OpenWithRetryAsync()
```
drwill-ms, line 120: I moved this open with retry logic to TestDevice so it can be used by more tests.

e2e/test/helpers/TestDevice.cs
```diff
@@ -140,32 +81,70 @@ public string ConnectionString
         /// </summary>
         public Device Device { get; private set; }
 
+        public IotHubDeviceClient DeviceClient { get; private set; }
+
         public IAuthenticationMethod AuthenticationMethod { get; private set; }
 
-        public IotHubDeviceClient CreateDeviceClient(IotHubClientOptions options = default, ConnectionStringAuthScope authScope = ConnectionStringAuthScope.Device)
+        public async Task<IotHubDeviceClient> CreateDeviceClientAsync(IotHubClientOptions options = default, ConnectionStringAuthScope authScope = ConnectionStringAuthScope.Device, bool openClient = false)
```
abhipsaMisra, line 88: nit: should we rename this so that the method name calls out that the returned client might optionally be already open?

e2e/test/helpers/TestDevice.cs
abhipsaMisra, line 120: Since this method is public, is there any value in adding the `openClient` block to the `Create` method above?

abhipsaMisra: imo keeping them separate would also drive home the point in our tests that open needs to be called explicitly.

abhipsaMisra (approved):

brycewang-microsoft (approved):

PR #3146, opened 2023-03-04
Using IAsyncDisposable to cleanup test devices

drwill-ms:
Instead of each test method having to have a try/finally to remove the test device, we can use IAsyncDisposable to remove it when the variable goes out of scope.

### DIFF
--- a/e2e/test/helpers/TestModule.cs
+++ b/e2e/test/helpers/TestModule.cs
@@ -6,56 +6,51 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Azure.Devices.E2ETests.Helpers
 {
-    public class TestModule
+    internal class TestModule : IAsyncDisposable
     {
-        private readonly Module _module;
+        private Module _module;
+        private TestDevice _testDevice;
 
-        private TestModule(Module module)
-        {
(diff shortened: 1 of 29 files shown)


timtay-microsoft (approved):
I like it!

e2e/test/helpers/TestModule.cs
```diff
@@ -3,61 +3,54 @@
...
-        /// </summary>
-        public string ConnectionString
+        public async ValueTask DisposeAsync()
         {
-            get
-            {
-                string iotHubHostName = GetHostName(TestConfiguration.IotHub.ConnectionString);
-                return $"HostName={iotHubHostName};DeviceId={_module.DeviceId};ModuleId={_module.Id};SharedAccessKey={_module.Authentication.SymmetricKey.PrimaryKey}";
-            }
+            await _testDevice.DisposeAsync().ConfigureAwait(false);
```
abhipsaMisra, line 52: small nit: we could add a comment here stating that TestDevice.DisposeAsync() does in fact delete the associated device as well (since the deleteion code is in a different file).

abhipsaMisra (approved):

PR #3149, opened 2023-03-08
Set state to 'closed' when CloseAsync() is called

abhipsaMisra:


iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -33,7 +33,7 @@ internal class RetryDelegatingHandler : DefaultDelegatingHandler
         private bool _deviceReceiveMessageEnabled;
         private bool _isDisposing;
         private bool _isAnEdgeModule = true;
-        private long _isOpened; // store the opened status in an int which can be accessed via Interlocked class. opened = 1, closed = 0.
+        private long _isOpen; // store the open/closed status in an int which can be accessed via Interlocked class. open = 1, closed = 0.
```
abhipsaMisra, line 36: Renamed "isOpened" -> "isOpen"

timtay-microsoft (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
timtay-microsoft, line 36: Is there any chance this could be made to be a boolean? I'm not that familiar with the ```Interlocked``` class that uses this

abhipsaMisra: Interlocked is similar to AtomicInteger in Java, wherein it allows for atomic operations to be executed on the variable location supplied. This works better in C# than using the "volatile" keyword to access the latest value of the variable from memory.
Interlocked, unfortunately, doesn't work with bools so this needs to be defined as a long.

patilsnr (approved):

drwill-ms (commented):
I'd still encourage you to make a helper class that encapsulates the long and hard coded values 0/1, exposes a bool get/set via method.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -1243,8 +1269,10 @@ protected override void Dispose(bool disposing)
                 if (!_isDisposed)
                 {
                     _isDisposing = true;
+                    SetClientTransportStatus(ClientTransportStatus.Closed);
```
abhipsaMisra, line 1272: Set the client status to "closed" when Dispose is called

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -1231,6 +1238,25 @@ private void HandleConnectionStatusExceptions(Exception exception, bool retryAtt
...
         }
 
+        private bool IsClientTransportOpen()
+        {
+            return Interlocked.Read(ref _isOpen) == 1;
+        }
+
+        private void SetClientTransportStatus(ClientTransportStatus transportStatus)
+        {
+            switch (transportStatus)
```
timtay-microsoft, line 1248: Any reason why you are using a switch here? If/Else is usually what I go for unless there are a lot of possible cases or if I need fall through.

timtay-microsoft (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -14,6 +14,12 @@
 
 namespace Microsoft.Azure.Devices.Client.Transport
 {
+    internal enum ClientTransportStatus
```
drwill-ms, line 17: Can we put this is its own file, unless it should be part of the class only (in which case it'd also be private)?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -1231,6 +1238,25 @@ private void HandleConnectionStatusExceptions(Exception exception, bool retryAtt
                     nameof(HandleConnectionStatusExceptions));
         }
 
+        private bool IsClientTransportOpen()
+        {
+            return Interlocked.Read(ref _isOpen) == 1;
```
drwill-ms, line 1243: Save 1 and 0 as class constants.

drwill-ms (approved):

patilsnr (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: I agree. I left this as a switch case since I think I might need to add another case here to facilitate the open-close-open flow that we have in previews/v2. I left the switch here as well, just for keeping the code the same. That said, if I don't end up using 2+ cases on previews/v2 I'll update both to use simpler if-else instead.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: It is an enum defined on the namespace, so unfortunately we cannot make it private. I did place it within this file since this is the only place it is used. Would you prefer I move it to its own file?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: How about I assign the values to the enum directly and then the helpers can be modified to:

```c#
        private bool IsClientTransportOpen()
        {
            return Interlocked.Read(ref _isOpen) == (int)ClientTransportStatus.Open;
        }

        private void SetClientTransportStatus(ClientTransportStatus transportStatus)
        {
            _ = Interlocked.Exchange(ref _isOpen, (int)transportStatus);
        }
```

PR #3153, opened 2023-03-13
Fix release build and prep for new version

drwill-ms:


iothub/service/src/Amqp/AmqpCbsSessionHandler.cs
```diff
@@ -12,7 +12,9 @@ namespace Microsoft.Azure.Devices.Amqp
     /// Handles a single CBS session (for SAS token renewal) including the inital authentication and scheduling all subsequent
     /// authentication attempts.
     /// </summary>
+#pragma warning disable CA1852 // used in debug for unit test mocking
```
drwill-ms, line 15: We need to remember to do this when we add unit tests that access internals, otherwise it'll fail the release build due to static analysis.

timtay-microsoft (approved):

abhipsaMisra (commented):
lgtm

PR #3154, opened 2023-03-13
For c2d, take payload as object, similar to other operations

drwill-ms:
And rename Message to `OutgoingMessage` (symmetry with the `IncomingMessage` object in the device SDK).

### DIFF
--- a/iothub/service/src/Messaging/Models/OutgoingMessage.cs
+++ b/iothub/service/src/Messaging/Models/OutgoingMessage.cs
@@ -3,20 +3,21 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text;
+using Newtonsoft.Json;
 
 namespace Microsoft.Azure.Devices
 {
     /// <summary>
-    /// The data structure represent the message that is used for interacting with IoT hub.
+    /// An outgoing message to be sent to a device or module using cloud-to-device messaging.
     /// </summary>
-    public sealed class Message
+    /// <seealso href="https://learn.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-c2d"/>
+    public sealed class OutgoingMessage
     {
-        private readonly byte[] _payload;
-
         /// <summary>
         /// Default constructor with no body data.
         /// </summary>
-        public Message()
+        public OutgoingMessage()
         {
         }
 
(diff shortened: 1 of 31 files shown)


iothub/service/src/Messaging/Models/OutgoingMessage.cs
```diff
@@ -3,31 +3,33 @@
...
 
 namespace Microsoft.Azure.Devices
 {
     /// <summary>
-    /// The data structure represent the message that is used for interacting with IoT hub.
+    /// An outgoing message to be sent to a device or module using cloud-to-device messaging.
     /// </summary>
-    public sealed class Message
+    /// <seealso href="https://learn.microsoft.com/azure/iot-hub/iot-hub-devguide-messages-c2d"/>
+    public sealed class OutgoingMessage
```
abhipsaMisra, line 15: Unrelated to this specific PR but do you remember if we review all the properties exposed on this class and removed the ones that weren't relevant for OutgoingMessage? I remember that we did that for the device/module client's view of the "message" but I don't remember if we did it for the service client's view of the "message" as well.

abhipsaMisra (approved):

iothub/service/src/Messaging/Models/OutgoingMessage.cs
drwill-ms: I'm not sure if we did.

drwill-ms: Taking a glance to see if any seem unrelated. How about this one?

```
        /// <summary>
        /// [Required in feedback messages] Used to specify the origin of messages generated by device hub.
        /// Possible value: “{hub name}/”
        /// </summary>
        public string UserId
```

drwill-ms: The others seem fine to me.

abhipsaMisra: It looks like UserId is a valid field for c2d messages: https://learn.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct#system-properties-of-c2d-iot-hub-messages

patilsnr (approved):

PR #3159, opened 2023-03-13
Throw IotHubClientException for file uploading with invalid correlation id

brycewang-microsoft:


patilsnr (approved):

timtay-microsoft (approved):

e2e/test/iothub/device/FileUploadE2ETests.cs
```diff
@@ -146,16 +145,15 @@ await using (deviceClient)
...
+                    catch (IotHubClientException ex) when (ex.ErrorCode is IotHubClientErrorCode.ServerError)
                     {
                         // Gateway V1 flow
-                        ex.ErrorCode.Should().Be(IotHubClientErrorCode.ServerError);
                     }
-                    catch (ArgumentException ex)
+                    catch (IotHubClientException ex) when (ex.ErrorCode is IotHubClientErrorCode.BadRequest)
                     {
                         // Gateway V2 flow
                         ex.Message.Should().Contain("400006");
```
drwill-ms, line 159:
```suggestion
                        ex.ErrorCode.Should().Be(IotHubClientErrorCode.IotHubFormatError);
```

drwill-ms: Well, that suggestion doesn't make sense given the `when` clause above.

Is there a reason we can't get the error code out?

drwill-ms (changes requested):
Comment inline. We should parse the payload and make a proper exception out of it.

e2e/test/iothub/device/FileUploadE2ETests.cs
brycewang-microsoft: That's reasonable. Sure, I will add that.

drwill-ms (approved):

e2e/test/iothub/device/FileUploadE2ETests.cs
```diff
@@ -146,19 +145,20 @@ await using (deviceClient)
 
                     try
                     {
-                        // TODO: the HTTP layer handles errors differently and does not produce the right kind of exceptions.
-                        // It should be updated to throw the same kind of exceptions as MQTT and AMQP.
+                        // TODO: Gateway V1 wrongly returns 500 for uploading file with invalid correlation id, while Gateway V2 corrects this behavior.
```
drwill-ms, line 148: Remove the TODO

PR #3161, opened 2023-03-14
Fix synchronization of semaphore in retry delegating handler

abhipsaMisra:
Changes from #3135, #3149

The sepmaphores associated with enabling and disabling subscriptions for callbacks should not be dependent on the Semaphore used for opening or closing the client.
The only purpose of the enable-disable events semaphore is to ensure that requests to enable and disable the subscription invoked by parallel threads are executed in a thread-safe manner.

* Decouple client open-close semaphore from callback subscription semaphore
* Cancel pending operations when CloseAsync() is invoked


drwill-ms (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -10,26 +10,37 @@
...
         private readonly RetryHandler _internalRetryHandler;
         private IIotHubClientRetryPolicy _retryPolicy;
 
-        private bool _isOpen;
-        private SemaphoreSlim _handlerSemaphore = new(1, 1);
         private bool _methodsEnabled;
         private bool _twinEnabled;
         private bool _deviceReceiveMessageEnabled;
+        private bool _isDisposing;
+        private long _clientTransportState; // references the current client transport status as the int value of ClientTransportStatus
```
brycewang-microsoft, line 36: rename to align with other places?

```suggestion
        private long _clientTransportStatus; // references the current client transport status as the int value of ClientTransportStatus
```

brycewang-microsoft (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -58,16 +71,19 @@ public override async Task SendTelemetryAsync(TelemetryMessage message, Cancella
             if (Logging.IsEnabled)
                 Logging.Enter(this, message, cancellationToken, nameof(SendTelemetryAsync));
 
+            await VerifyIsOpenAsync(cancellationToken).ConfigureAwait(false);
+            using var operationCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _cancelPendingOperationsCts.Token);
```
abhipsaMisra, line 75: I've added a pre-check operation to each API call - `VerifyIsOpenAsync()` once outside the retry block.

If the verification that the client is open is always linked with the retry block, then closing a client and then calling `SendTelemetryAsync()` will result in the SDK complaining that the linked cancellation token source (the one that signals in-progress operations for closure) has been canceled already, instead of complaining that the client is closed.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -82,16 +98,19 @@ public override async Task SendTelemetryBatchAsync(IEnumerable<TelemetryMessage>
             if (Logging.IsEnabled)
                 Logging.Enter(this, messages, cancellationToken, nameof(SendTelemetryAsync));
 
+            await VerifyIsOpenAsync(cancellationToken).ConfigureAwait(false);
+            using var operationCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _cancelPendingOperationsCts.Token);
```
abhipsaMisra, line 102: This is the linked cancellation token that will be signaled for closure if `CloseAsync()` is called

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -388,53 +484,68 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
-            await _handlerSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
-            try
-            {
-                if (!_isOpen)
-                {
-                    if (Logging.IsEnabled)
-                        Logging.Info(this, "Opening connection", nameof(OpenAsync));
+                case ClientTransportStatus.Open:
+                case ClientTransportStatus.Opening:
+                    return;
```
abhipsaMisra, line 493: If the client is currently open or opening, simply exit.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -388,53 +484,68 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
-                    {
-                        await OpenInternalAsync(cancellationToken).ConfigureAwait(false);
-                    }
-                    catch (Exception ex) when (!Fx.IsFatal(ex))
-                    {
-                        HandleConnectionStatusExceptions(ex, true);
-                        throw;
-                    }
+                case ClientTransportStatus.Closing:
+                    throw new InvalidOperationException($"The client is currently closing. Wait until {nameof(CloseAsync)} completes and then invoke {nameof(OpenAsync)} again.");
```
abhipsaMisra, line 496: If `Openasync()` is called while the client is closing, throw an InvalidOperationException.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -388,53 +484,68 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
+                                // This is to ensure that if OpenInternalAsync() fails on retry expiration with a custom retry policy,
+                                // we are returning the corresponding connection status change event => disconnected: retry_expired.
+                                try
+                                {
+                                    await OpenInternalAsync(operationCts.Token).ConfigureAwait(false);
+                                }
+                                catch (Exception ex) when (!Fx.IsFatal(ex))
+                                {
+                                    HandleConnectionStatusExceptions(ex, true);
+                                    SetClientTransportStatus(ClientTransportStatus.Closed);
```
abhipsaMisra, line 523: If `OpenInternalAsync()` fails, change the status from opening -> closed.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -443,33 +554,26 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
...
-                _handleDisconnectCts.Cancel();
+                _cancelPendingOperationsCts?.Cancel();
+                _handleDisconnectCts?.Cancel();
                 await base.CloseAsync(cancellationToken).ConfigureAwait(false);
             }
             finally
             {
-                _isOpen = false;
+                _cancelPendingOperationsCts?.Dispose();
+                SetClientTransportStatus(ClientTransportStatus.Closed);
```
abhipsaMisra, line 573: @drwill-ms We had discussed encapsulating the `CloseAync()` within a try-catch block so that close failures are suppressed. However, as I was testing this flow I had another thought. Since we do allow open-close-open flow, I wanted to avoid suppressing any exceptions in here as failure to cleanly close the transport layer might result in issues down the line if subsequently `OpenAsync()` is called. For this reason, I wanted to leave this as-is and evaluate this in the future if we get feedback from customers on this flow.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -504,26 +610,26 @@ public override void SetSasTokenRefreshesOn()
 
             if (_refreshLoop == null)
             {
-                if (_loopCancellationTokenSource != null)
+                if (_sasRefreshLoopCancellationCts != null)
```
abhipsaMisra, line 613: This is just a rename

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -621,24 +727,37 @@ private async Task RefreshSasTokenLoopAsync(DateTime refreshesOn, CancellationTo
...
         private async Task VerifyIsOpenAsync(CancellationToken cancellationToken)
         {
-            await _handlerSemaphore.WaitAsync(cancellationToken);
+            ThrowIfDisposed();
+            await _clientOpenSemaphore.WaitAsync(cancellationToken);
 
             try
             {
-                if (!_isOpen)
+                if (GetClientTransportStatus() != ClientTransportStatus.Open)
```
abhipsaMisra, line 735: This will result in operations invoked when the client is "opening" to also get rejected. It seemed like a reasonable expectation since we do want the client behavior to be deterministic.

iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs
```diff
@@ -163,48 +162,6 @@ public async Task RetryOneMessageHasBeenTouchedTransientExceptionOccuredSuccess(
...
-            // act
-            await retryDelegatingHandler.OpenAsync(CancellationToken.None).ConfigureAwait(false);
-            await retryDelegatingHandler.SendTelemetryBatchAsync(messages, CancellationToken.None).ConfigureAwait(false);
-
-            // assert
-            callCounter.Should().Be(2);
-        }
-
-        [TestMethod]
-        public async Task RetryMessageWithSeekableStreamHasBeenReadTransientExceptionOccuredThrows()
```
abhipsaMisra, line 189: This test was repeated twice. I've removed one instance.

iothub/device/tests/ExceptionAssertions.cs
abhipsaMisra, line 1: I cannot find a way to fluently assert the state of an in-progress Task. So, I ended up adding this custom assertion helper. Let me know if you know of a way to do that using `FluentAssertions` directly.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -60,14 +73,17 @@ public override async Task SendTelemetryAsync(TelemetryMessage message, Cancella
 
             try
             {
+                await VerifyIsOpenAsync(cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 76: I've added a pre-check operation to each API call - `VerifyIsOpenAsync()` once outside the retry block.

If the verification that the client is open is always linked with the retry block, then closing a client and then calling `SendTelemetryAsync()` will result in the SDK complaining that the linked cancellation token source (the one that signals in-progress operations for closure) has been canceled already, instead of complaining that the client is closed.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Updating the formatting marked this comment as outdated. Resolving this and adding a new one.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -60,14 +73,17 @@ public override async Task SendTelemetryAsync(TelemetryMessage message, Cancella
...
+                using var operationCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _cancelPendingOperationsCts.Token);
```
abhipsaMisra, line 77: This is the linked cancellation token that will be signaled for closure if `CloseAsync()` is called

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Updating the formatting marked this comment as outdated. Resolving this and adding a new one.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -383,58 +479,86 @@ public override async Task<long> UpdateReportedPropertiesAsync(ReportedPropertie
...
-                    catch (Exception ex) when (!Fx.IsFatal(ex))
-                    {
-                        HandleConnectionStatusExceptions(ex, true);
-                        throw;
-                    }
+                switch (GetClientTransportStatus())
+                {
+                    case ClientTransportStatus.Open:
+                    case ClientTransportStatus.Opening:
+                        return;
```
abhipsaMisra, line 499: If the client is currently open or opening, simply exit.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Updating the formatting marked this comment as outdated. Resolving this and adding a new one.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -383,58 +479,86 @@ public override async Task<long> UpdateReportedPropertiesAsync(ReportedPropertie
...
 
-                    if (!_isDisposed)
-                    {
-                        _isOpen = true;
+                    case ClientTransportStatus.Closing:
+                        throw new InvalidOperationException($"The client is currently closing. Wait until {nameof(CloseAsync)} completes and then invoke {nameof(OpenAsync)} again.");
```
abhipsaMisra, line 502: If `Openasync()` is called while the client is closing, throw an InvalidOperationException.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Updating the formatting marked this comment as outdated. Resolving this and adding a new one.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -383,58 +479,86 @@ public override async Task<long> UpdateReportedPropertiesAsync(ReportedPropertie
...
+                                    // This is to ensure that if OpenInternalAsync() fails on retry expiration with a custom retry policy,
+                                    // we are returning the corresponding connection status change event => disconnected: retry_expired.
+                                    try
+                                    {
+                                        await OpenInternalAsync(operationCts.Token).ConfigureAwait(false);
+                                    }
+                                    catch (Exception ex) when (!Fx.IsFatal(ex))
+                                    {
+                                        HandleConnectionStatusExceptions(ex, true);
+                                        SetClientTransportStatus(ClientTransportStatus.Closed);
```
abhipsaMisra, line 530: If `OpenInternalAsync()` fails, change the status from opening -> closed.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -443,33 +567,31 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
...
                 }
+                SetClientTransportStatus(ClientTransportStatus.Closing);
 
-                _handleDisconnectCts.Cancel();
-                await base.CloseAsync(cancellationToken).ConfigureAwait(false);
+                try
+                {
+                    _cancelPendingOperationsCts?.Cancel();
+                    _handleDisconnectCts?.Cancel();
+                    await base.CloseAsync(cancellationToken).ConfigureAwait(false);
```
abhipsaMisra, line 583: @drwill-ms We had discussed encapsulating the `CloseAync()` within a try-catch block so that close failures are suppressed. However, as I was testing this flow I had another thought. Since we do allow open-close-open flow, I wanted to avoid suppressing any exceptions in here as failure to cleanly close the transport layer might result in issues down the line if subsequently `OpenAsync()` is called. For this reason, I wanted to leave this as-is and evaluate this in the future if we get feedback from customers on this flow.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -621,170 +745,222 @@ private async Task RefreshSasTokenLoopAsync(DateTime refreshesOn, CancellationTo
...
             try
             {
-                if (!_isOpen)
+                ThrowIfDisposed();
+                await _clientOpenSemaphore.WaitAsync(cancellationToken);
+
+                try
                 {
-                    throw new InvalidOperationException($"The client connection must be opened before operations can begin. Call '{nameof(OpenAsync)}' and try again.");
+                    if (GetClientTransportStatus() != ClientTransportStatus.Open)
```
abhipsaMisra, line 758: This will result in operations invoked when the client is "opening" to also get rejected. It seemed like a reasonable expectation since we do want the client behavior to be deterministic.

iothub/device/tests/ExceptionAssertions.cs
abhipsaMisra, line 1: I cannot find a way to fluently assert the state of an in-progress Task. So, I ended up adding this custom assertion helper. Let me know if you know of a way to do that using `FluentAssertions` directly.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms: Is this documented on the client's OpenAsync call?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms, line 758: consider checking status, and if opening, then wait the semaphore. That way during happy path, these methods don't wait for the semaphore and slow each other down.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -621,170 +745,222 @@ private async Task RefreshSasTokenLoopAsync(DateTime refreshesOn, CancellationTo
...
+                            tasks.Add(base.EnableReceiveMessageAsync(cancellationToken));
+                        }
 
+                        if (tasks.Any())
+                        {
+                            await Task.WhenAll(tasks).ConfigureAwait(false);
+                        }
+
+                        // Send the request for transport close notification.
+                        _transportClosedTask = HandleDisconnectAsync();
```
drwill-ms, line 937: I'm worried about the lifetime of this handle disconnect code. It is called by someone else the first time and then by itself the second time.

Is any given instance ever pointing to the wrong token that it captures off of `_handleDisconnectCts`? Are we sure none of these calls duplicate over time?

I'd suggest diagraming it to be sure, and honestly this pattern is probably too fragile that anyone who touches this code and doesn't thoroughly understand this lifetime could break it.

I know you didn't introduce it in this PR, but perhaps we should look for a simpler impl like a callback.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -10,26 +10,39 @@
 
 namespace Microsoft.Azure.Devices.Client.Transport
 {
+    internal enum ClientTransportStatus
+    {
+        Closed,
```
drwill-ms, line 15: Let's add comments for each state, what it actually means in this class, and what behavior is expected in each state.

drwill-ms: Consider moving this and the code that changes states to its own class (state management class) so it can be unit tested.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
drwill-ms, line 1: It occurs to me that this file which is supposed to be about retrying, and it looks like it does a lot more than that, especially with the addition of reconnection and overall connection (sort of) management.

Should checking the connection, reconnection attempts, etc. be a separate delegating handler, so this class doesn't get quite so full of mixed concerns?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Added the comments.

abhipsaMisra: We will address the separation of state management in the subsequent "separate-delegating-handlers" PR.

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Can be modified further in the "separate-delegating-handlers" PR.

abhipsaMisra:
Closing in favor of #3251

PR #3164, opened 2023-03-14
Update migration doc by adding IotHubServiceException with more error codes

brycewang-microsoft:


iothub/service/src/Exceptions/IotHubServiceErrorCode.cs
```diff
@@ -180,7 +185,7 @@ public enum IotHubServiceErrorCode
         // RequestEntityTooLarge - 413
 
         /// <summary>
-        /// When the message is too large for IoT hub you will receive this error.'
+        /// When the message is too large for IoT hub you will receive this error.
```
drwill-ms, line 188:
```suggestion
        /// The message is <see href="https://learn.microsoft.com/azure/iot-hub/iot-hub-devguide-quotas-throttling#other-limits">too large</see> for IoT hub.
```

drwill-ms (approved):

PR #3165, opened 2023-03-14
Utilize AsyncPageable for query return types

timtay-microsoft:
The new APIs look like:
```csharp
public virtual AsyncPageable<T> CreateAsync<T>(string query, CancellationToken cancellationToken = default)
```

A couple points worth discussing:
- Unlike in any of our other APIs, the AsyncPageable APIs contain the Response<T> type that allows users to see the http request contents. 
  - This is generally a good thing to let users see, but it may beg the question as to why the other APIs don't allow this same thing.
- Some code (```PageableHelpers```) has to be copied from Azure.Core since constructing an AsyncPageable instance is non-trivial
  -  The public facing return types are the Azure.Core defined classes, though.
- The page size and continuation token can be set through the AsyncPageable itself, so we can remove the ```QueryOptions``` class that we had that served the same purpose.
  - The only real loss here is that users cannot set page size or continuation token unless they are iterating page by page. Users iterating item by item won't actually be capable of getting the continuation token, but would they still care about page size?
```csharp
// Old pagination configuration using our defined QueryOptions
var options = new QueryOptions()
{
    ContinuationToken = "someToken",
    PageSize = 4
};
AsyncPageable<ClientTwin> twinQuery = serviceClient.Query.CreateAsync("some twin query", queryOptions);
await foreach (Page<ClientTwin> queriedTwinPage in twinQuery.AsPages())
{
    // some application logic
}
```

```csharp
// Pagination configuration using AsyncEnumerable methods
int pageSize = 5;
AsyncPageable<ClientTwin> twinQuery = serviceClient.Query.CreateAsync("some twin query");
await foreach (Page<ClientTwin> queriedTwinPage in twinQuery.AsPages("continuationToken", pageSize))
{
    // some application logic
}
```

Once this PR is merged in, I'll apply the same changes to the DPS service client since it also has a query API


iothub/service/src/Query/QueryResponse.cs
```diff
@@ -3,194 +3,79 @@
...
 {
-    /// <summary>
-    /// An iterable set of queried items.
-    /// </summary>
-    /// <typeparam name="T">
-    /// The type of the queried items. For instance, when using a query such as "SELECT * FROM devices",
-    /// this type should be type <see cref="ClientTwin"/>. When using a query such as "SELECT * FROM devices.jobs",
-    /// this type should be type <see cref="ScheduledJob"/>.
-    /// </typeparam>
-    public class QueryResponse<T>
```
timtay-microsoft, line 20: This class used to be the "AsyncPageable" implementation we had locally, but now it is the "Response<T>" implementation since the Azure.Core AsyncPageable pages all contain the Response<T> type embedded in them

e2e/test/iothub/service/QueryClientE2ETests.cs
```diff
@@ -44,19 +45,20 @@ public async Task TwinQuery_Works()
 
             await WaitForDevicesToBeQueryableAsync(serviceClient.Query, queryText, 2).ConfigureAwait(false);
 
-            QueryResponse<ClientTwin> queryResponse = await serviceClient.Query.CreateAsync<ClientTwin>(queryText).ConfigureAwait(false);
+            AsyncPageable<ClientTwin> queryResponse = serviceClient.Query
+                .CreateAsync<ClientTwin>(queryText);
```
drwill-ms, line 49:
```suggestion
            AsyncPageable<ClientTwin> queryResponse = serviceClient.Query.CreateAsync<ClientTwin>(queryText);
```

iothub/service/src/Query/PageableHelpers.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Threading.Tasks;
+using Azure;
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Copy of a subset of the helper functions defined in the Azure.Core class by the same name:
+    /// https://github.com/Azure/autorest.csharp/blob/main/src/assets/Generator.Shared/PageableHelpers.cs
+    /// </summary>
+    internal class PageableHelpers
```
drwill-ms, line 15: static?

iothub/service/src/Query/QueryClient.cs
```diff
@@ -170,41 +154,94 @@ await _internalRetryHandler
...
+                throw;
+            }
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, $"Creating query with jobType: {options?.JobType}, jobStatus: {options?.JobStatus}", nameof(CreateAsync));
+            }
+        }
+
+        private async Task<Page<T>> BuildAndSendRequest<T>(HttpRequestMessage request, string continuationToken, int? pageSizeHint, CancellationToken cancellationToken)
```
drwill-ms, line 221:
```suggestion
        private async Task<Page<T>> BuildAndSendRequestAsync<T>(HttpRequestMessage request, string continuationToken, int? pageSizeHint, CancellationToken cancellationToken)
```

iothub/service/src/Query/QueryClient.cs
```diff
@@ -104,45 +102,31 @@ public virtual async Task<QueryResponse<T>> CreateAsync<T>(string query, QueryOp
...
-                    s_queryUri,
-                    _credentialProvider,
-                    new QuerySpecification { Sql = query });
-                request.Content.Headers.ContentType = new MediaTypeHeaderValue("application/json") { CharSet = "utf-8" };
-                if (!string.IsNullOrWhiteSpace(options?.ContinuationToken))
+                async Task<Page<T>> nextPageFunc(string continuationToken, int? pageSizeHint)
                 {
-                    request.Headers.Add(ContinuationTokenHeader, options?.ContinuationToken);
-                }
+                    cancellationToken.ThrowIfCancellationRequested();
```
drwill-ms, line 107: Shouldn't this be before the try?

iothub/service/src/Query/QueryClient.cs
```diff
@@ -170,41 +154,94 @@ await _internalRetryHandler
...
-                    BuildQueryJobQueryString(options));
+                async Task<Page<ScheduledJob>> nextPageFunc(string continuationToken, int? pageSizeHint)
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+                    using HttpRequestMessage request = _httpRequestMessageFactory.CreateRequest(
+                                        HttpMethod.Get,
+                                        s_jobsQueryFormat,
+                                        _credentialProvider,
+                                        null,
+                                        BuildQueryJobQueryString(options));
```
drwill-ms, line 181:
```suggestion
                    using HttpRequestMessage request = _httpRequestMessageFactory.CreateRequest(
                         HttpMethod.Get,
                         s_jobsQueryFormat,
                         _credentialProvider,
                         null,
                         BuildQueryJobQueryString(options));
```

iothub/service/src/Query/QueryClient.cs
```diff
@@ -170,41 +154,94 @@ await _internalRetryHandler
...
-                    request.Headers.Add(ContinuationTokenHeader, options?.ContinuationToken);
+                    cancellationToken.ThrowIfCancellationRequested();
+                    using HttpRequestMessage request = _httpRequestMessageFactory.CreateRequest(
+                                        HttpMethod.Get,
+                                        s_jobsQueryFormat,
+                                        _credentialProvider,
+                                        null,
+                                        BuildQueryJobQueryString(options));
+
+                    return await BuildAndSendRequest<ScheduledJob>(request, null, pageSizeHint, cancellationToken);
```
drwill-ms, line 196:
```suggestion
                    return await BuildAndSendRequest<ScheduledJob>(request, null, pageSizeHint, cancellationToken).ConfigureAwait(false);
```

iothub/service/src/Query/QueryClient.cs
```diff
@@ -170,41 +154,94 @@ await _internalRetryHandler
...
+            }
+        }
+
+        private async Task<Page<T>> BuildAndSendRequest<T>(HttpRequestMessage request, string continuationToken, int? pageSizeHint, CancellationToken cancellationToken)
+        {
+            if (!string.IsNullOrWhiteSpace(continuationToken))
+            { 
+                request.Headers.Add(ContinuationTokenHeader, continuationToken);
+            }
+
```
drwill-ms, line 227: check cancellation token?

iothub/service/src/Query/QueryClient.cs
```diff
@@ -170,41 +154,94 @@ await _internalRetryHandler
...
+                request.Headers.Add(PageSizeHeader, pageSizeHint.ToString());
+            }
+
+            if (request.Content != null)
+            { 
+                request.Content.Headers.ContentType = new MediaTypeHeaderValue("application/json") 
+                { 
+                    CharSet = "utf-8" 
+                };
+            }
```
drwill-ms, line 239:
```suggestion
            (request.Content?.Headers.?ContentType ??= new MediaTypeHeaderValue("application/json") 
            { 
                CharSet = "utf-8" 
            };
            }
```

iothub/service/src/Query/QueryResponse.cs
```diff
@@ -3,194 +3,79 @@
...
-        ///
-        /// This function is async because it may make a service request to fetch the next page of results if the current page
-        /// of results has been advanced through already. Note that this function will return True even if it is at the end
-        /// of a particular page of items as long as there is at least one more page to be fetched.
-        /// </remarks>
-        public async Task<bool> MoveNextAsync(QueryOptions queryOptions = default, CancellationToken cancellationToken = default)
+        public override Stream ContentStream 
         {
-            cancellationToken.ThrowIfCancellationRequested();
+            get =>  _httpResponse.Content.ReadAsStreamAsync().Result;
```
drwill-ms, line 36: I don't like this for potential deadlock. Is this how it is done in track 2?

iothub/service/src/Query/QueryResponse.cs
```diff
@@ -3,194 +3,79 @@
...
+        protected override IEnumerable<HttpHeader> EnumerateHeaders()
+        {
+            return _httpHeaders;
+        }
 
-            if (!string.IsNullOrEmpty(_originalQuery))
+        protected override bool TryGetHeader(string name, out string value)
+        {
+            IEnumerable<string> outVariableHeaders = new List<string>();
+            bool found = _httpResponse.Headers.TryGetValues(name, out outVariableHeaders);
```
drwill-ms, line 63:
```suggestion
            value = _httpResponse.Headers.TryGetValues(name, out outVariableHeaders)
                ? outVariableHeaders.First()
                : string.Empty;
```

iothub/service/src/Query/QueryResponse.cs
```diff
@@ -3,194 +3,79 @@
...
-                Current = CurrentPage.GetEnumerator().Current;
-                ContinuationToken = response.ContinuationToken;
+                value = "";
             }
 
-            return true;
+            return found;
+        }
+
+        protected override bool TryGetHeaderValues(string name, out IEnumerable<string> values)
```
drwill-ms, line 76: See `HttpHeadersExtensions.SafeGetValue)`

iothub/service/tests/QueryClientTests.cs
```diff
@@ -15,6 +15,7 @@
 using FluentAssertions;
 using System.Collections;
 using Newtonsoft.Json.Linq;
+using Azure;
```
drwill-ms, line 18: sort

drwill-ms (changes requested):
Needs an update to migration guide too.

iothub/service/src/Query/QueryClient.cs
timtay-microsoft: This cancellation token check is within the "get next page" func that is called several times over the course of the query. There is a cancellation token check outside the try as well for when the query is being constructed.

iothub/service/src/Query/QueryResponse.cs
timtay-microsoft: Nice

iothub/service/src/Query/QueryClient.cs
timtay-microsoft: This actually didn't work for me. Even after replacing ```Headers.?``` with ```Headers?.```

iothub/service/src/Query/QueryResponse.cs
timtay-microsoft: I'm not sure what you are going for here, but the ```HttpHeadersExtensions.SafeGetValue``` call trivializes this so I'll go with it instead

iothub/service/src/Query/QueryClient.cs
drwill-ms: Oh well.

iothub/service/src/Query/QueryResponse.cs
drwill-ms: Yeah, I left this comment before I saw the opportunity to use a helper we already had.

iothub/service/src/Query/PageableHelpers.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Microsoft.Azure.Devices
+{
+    /// <summary>
+    /// Copy of a subset of the helper functions defined in the Azure.Core class by the same name:
+    /// https://github.com/Azure/autorest.csharp/blob/main/src/assets/Generator.Shared/PageableHelpers.cs
+    /// </summary>
+    internal static class PageableHelpers
+    {
+        public static AsyncPageable<T> CreateAsyncEnumerable<T>(Func<int?, Task<Page<T>>> firstPageFunc, Func<string, int?, Task<Page<T>>> nextPageFunc, int? pageSize = default) where T : notnull
```
drwill-ms, line 17:
```suggestion
        internal static AsyncPageable<T> CreateAsyncEnumerable<T>(Func<int?, Task<Page<T>>> firstPageFunc, Func<string, int?, Task<Page<T>>> nextPageFunc, int? pageSize = default) where T : notnull
```

iothub/service/src/Query/QueryResponse.cs
```diff
@@ -3,194 +3,77 @@
...
+        }
 
-            if (ContinuationToken == null && queryOptions?.ContinuationToken == null)
-            {
-                // The query has no more pages of results to return and the last page has been
-                // completely exhausted.
-                return false;
-            }
+        public override void Dispose()
+        {
```
drwill-ms, line 54: Dispose `_bodyStream`?

iothub/service/src/Query/QueryResponse.cs
```diff
@@ -3,194 +3,77 @@
...
-            var queryOptionsClone = new JobQueryOptions
-            {
-                ContinuationToken = queryOptions?.ContinuationToken ?? ContinuationToken,
-                PageSize = queryOptions?.PageSize ?? _defaultPageSize,
-                JobType = _jobType,
-                JobStatus = _jobStatus,
-            };
+        protected override bool ContainsHeader(string name)
+        {
+            return _httpResponse.Headers.Contains(name);
```
drwill-ms, line 60: Any issue if `name` is `""` or `null`?

iothub/service/src/Query/QueryResponse.cs
drwill-ms: Does this method still have callers?

drwill-ms (approved):
Still needs an update to migration guide.

PR #3167, opened 2023-03-15
Update migration guide and update SDK for Edge

drwill-ms:


patilsnr (approved):

iothub/device/src/ConnectionStatusInfo.cs
```diff
@@ -10,13 +10,19 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     public class ConnectionStatusInfo
     {
-        internal ConnectionStatusInfo()
+        /// <summary>
+        /// Generally only created by the SDK or for unit testing, using defaults of disconnected and client closed.
+        /// </summary>
+        public ConnectionStatusInfo()
```
abhipsaMisra, line 16: Are we making this public so that Edge can use the constructor?

abhipsaMisra (approved):

iothub/device/src/ConnectionStatusInfo.cs
drwill-ms: Yes, they use it to feed into their connection status handler.

We could make it protected internal for unit testing, but that seems unnecessarily complicated for a non-unit test case.

PR #3168, opened 2023-03-16
Add tests for RetryDelegatingHandler CloseAsync task cancellation

abhipsaMisra:


iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs
```diff
@@ -129,7 +129,7 @@ public async Task RetryOneMessageHasBeenTouchedTransientExceptionOccuredSuccess(
 
             var contextMock = Substitute.For<PipelineContext>();
             contextMock.ConnectionStatusChangesHandler = new ConnectionStatusChangesHandler(delegate (ConnectionStatus status, ConnectionStatusChangeReason reason) { });
-            var innerHandlerMock = Substitute.For<IDelegatingHandler>();
+            using var innerHandlerMock = Substitute.For<IDelegatingHandler>();
```
drwill-ms, line 132: Why would this require a `using` statement?

iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs
```diff
@@ -414,18 +414,150 @@ await sut
...
+            using var sut = new RetryDelegatingHandler(contextMock, innerHandlerMock);
             using var cts = new CancellationTokenSource();
             cts.Cancel();
 
             // act and assert
             await sut.RejectAsync(Arg.Any<string>(), cts.Token).ExpectedAsync<TaskCanceledException>().ConfigureAwait(false);
         }
 
+        [TestMethod]
+        public async Task CloseAsyncCancelPendingOpenAsync()
```
drwill-ms, line 430:
```suggestion
        public async Task RetryDelegatingHandler_CloseAsync_CancelsPendingOpenAsync()
```

drwill-ms (approved):

iothub/device/tests/Pipeline/RetryDelegatingHandlerTests.cs
abhipsaMisra: I had been running into the "ambiguous arguments" exception that we've seen on NSubstitute, so I made this change in an effort to clean up all involved resources in a test.
Do you think there'd be a problem in marking the mocked handlers to be disposed at the end of each test?

PR #3171, opened 2023-03-17
Upgrade APIs in long tests from V1 to V2

brycewang-microsoft:


e2e/LongHaul/device/ExceptionHelper.cs
```diff
@@ -18,28 +11,6 @@ namespace Microsoft.Azure.IoT.Thief.Device
     // Sample exception handler class - this class should be modified based on your application's logic
     internal class ExceptionHelper
     {
-        private static readonly HashSet<Type> s_networkExceptions = new HashSet<Type>
```
brycewang-microsoft, line 21: We should be able to handle this internally in V2 SDK.

e2e/LongHaul/device/IIotHub.cs
```diff
@@ -21,6 +21,6 @@ internal interface IIotHub
         /// </summary>
         /// <param name="properties">Twin properties to set.</param>
         /// <param name="cancellationToken">A cancellation token.</param>
-        Task SetPropertiesAsync(object properties, CancellationToken cancellationToken = default);
+        Task SetPropertiesAsync(SystemProperties properties, CancellationToken cancellationToken = default);
```
brycewang-microsoft, line 24: This is used for `SystemProperties` only, so refactoring the parameter type.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -118,35 +115,13 @@ public async Task RunAsync(CancellationToken ct)
...
-                    }
-                    catch (TaskCanceledException)
-                    {
-                        // App is signalled to exit
-                        _logger.Trace($"Exit signal encountered. Terminating telemetry message pump.");
-                        return;
-                    }
-                    catch (Exception ex)
-                    {
-                        _logger.Trace($"Unknown error sending telemetry: {ex}", TraceSeverity.Critical);
```
brycewang-microsoft, line 148: Now we should be able to handle this internally in V2 SDK.

e2e/LongHaul/device/IotHubConnectionStringHelper.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Azure.IoT.Thief.Device
+{
+    internal class IotHubConnectionStringHelper
```
brycewang-microsoft, line 8: Added this class since the long-haul tests' logger needs some info from connection string provided in setting files.

e2e/LongHaul/device/MessageHelpers.cs
```diff
@@ -11,27 +11,24 @@ internal static class MessageHelpers
         /// </summary>
         public const int MaxMessagePayloadBytes = (256 * 1024) - 1024;
 
-        private const int Iso8601Length = 24;
+        public static int GetMessagePayloadSize(TelemetryMessage message)
```
brycewang-microsoft, line 14: It seems that we don't have any references in this class, but upgraded to V2 contexts just in case.

e2e/LongHaul/device/Program.cs
```diff
@@ -104,13 +105,36 @@ private static Logger InitializeLogging(string deviceConnectionString, string ai
...
+            var iotHub = new IotHub(logger, _settings.DeviceConnectionString, GetTransportSettings(_settings.TransportProtocolType));
             foreach (var prop in _commonProperties)
             {
                 iotHub.IotProperties.Add(prop.Key, prop.Value);
             }
 
             return iotHub;
         }
+
+        private static IotHubClientTransportSettings GetTransportSettings(string transportProtocolType)
```
brycewang-microsoft, line 117: Previously the setting file takes the property "Transport" in type of `TransportType`. Now that we don't have this type in V2 any more, change the property type into `string` and map it to the specific `IotHubClientTransportSettings` object here.

e2e/LongHaul/device/IIotHub.cs
drwill-ms: Rename the method?

drwill-ms: Actually, I think this is meant to be twin properties, so it should be ReportedProperties.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -158,21 +133,9 @@ public async Task RunAsync(CancellationToken ct)
             Debug.Assert(_deviceClient != null);
             Debug.Assert(telemetryObject != null);
 
-            // Save off the event time, or use "now" if not specified
-            var creationTimeUtc = telemetryObject.EventDateTimeUtc ?? DateTime.UtcNow;
-            // Remove it so it does not get serialized in the message
-            telemetryObject.EventDateTimeUtc = null;
```
drwill-ms, line 164: We still want to do this.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -158,21 +133,9 @@ public async Task RunAsync(CancellationToken ct)
...
-
-            string message = JsonSerializer.Serialize(telemetryObject, _jsonSerializerOptions);
-            Debug.Assert(!string.IsNullOrWhiteSpace(message));
-
-            var iotMessage = new Message(Encoding.UTF8.GetBytes(message))
+            var iotMessage = new TelemetryMessage(telemetryObject)
             {
-                ContentEncoding = _contentEncoding,
-                ContentType = _contentType,
```
drwill-ms, line 172: Does this get set by the serialization and encoding layer?

e2e/LongHaul/device/IotHub.cs
```diff
@@ -158,21 +133,9 @@ public async Task RunAsync(CancellationToken ct)
...
                 MessageId = Guid.NewGuid().ToString(),
-                // Add the event time to the system property
-                CreationTimeUtc = creationTimeUtc,
```
drwill-ms, line 175: Still want to set this.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -198,22 +161,26 @@ public async Task RunAsync(CancellationToken ct)
...
+        public async Task SetPropertiesAsync(SystemProperties properties, CancellationToken cancellationToken)
         {
             Debug.Assert(_deviceClient != null);
             Debug.Assert(properties != null);
 
-            string propertiesPayload = JsonSerializer.Serialize(properties, _jsonSerializerOptions);
-            Debug.Assert(!string.IsNullOrWhiteSpace(propertiesPayload));
+            var reportedProperties = new ReportedProperties
+            {
+                { "systemArchitecture", properties.SystemArchitecture },
```
drwill-ms, line 171: Ah, this is something @abhipsaMisra hasn't added support for yet. In our current API, one can only set an object if they have a key for it. I want to be able to set a payload with a single parameter that has all the key/value pairs we want to set.

Anyway, we shouldn't have to hardcode these keys and this method shouldn't be hardcoded to only `SystemProperties`. Let's talk and work out how we enable this.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -198,22 +161,26 @@ public async Task RunAsync(CancellationToken ct)
...
             await _deviceClient
                 .UpdateReportedPropertiesAsync(
-                    new TwinCollection(propertiesPayload),
+                    reportedProperties,
                     cancellationToken)
                 .ConfigureAwait(false);
         }
 
-        public void Dispose()
+        public async void Dispose()
```
drwill-ms, line 183: Change to `ValueTask DisposeAsync()` and update class to inherit from `IAsyncDisposable`.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -246,25 +213,32 @@ private bool ResetClient(bool force = false)
...
+                {
+                    _disconnectedTimer.Stop();
+                    _logger.Metric(
+                        DisconnectedDurationSeconds,
+                        _disconnectedTimer.Elapsed.TotalSeconds,
+                        new Dictionary<string, string>
+                        {
                         { DisconnectedStatus, _disconnectedStatus.ToString() },
                         { DisconnectedReason, _disconnectedReason.ToString() },
                         { ConnectionStatusChangeCount, _connectionStatusChangeCount.ToString() },
```
drwill-ms, line 239:
```suggestion
                            { DisconnectedStatus, _disconnectedStatus.ToString() },
                            { DisconnectedReason, _disconnectedReason.ToString() },
                            { ConnectionStatusChangeCount, _connectionStatusChangeCount.ToString() },
```

e2e/LongHaul/device/IotHubConnectionStringHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public string HostName { get; private set; }
+
+        public string DeviceId { get; private set; }
+
+        public void GetValuesFromConnectionString(string connectionString)
```
drwill-ms, line 14: I suggest making this static and have it return an instance, or perhaps better, turn it into a ctor.

e2e/LongHaul/device/MessageHelpers.cs
drwill-ms: The original purpose of this was to enable sending messages in bulk.

We should only use it for AMQP, because MQTT doesn't do true bulk sending.

e2e/LongHaul/device/Program.cs
drwill-ms: Or make an enum like we do for many of our samples. It'll parse better that way.

drwill-ms: https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/device/samples/how%20to%20guides/DeviceReconnectionSample/Parameters.cs#L9

And change to 2 params, one for transport and one for transport protocol.

e2e/LongHaul/device/IIotHub.cs
drwill-ms: Nope, I was wrong there too.

The idea was to set a twin property where the contents of the object were the key/value pairs. The v2 API doesn't allow that.

e2e/LongHaul/device/IotHub.cs
brycewang-microsoft: Yes, they will be set while calling `[SendTelemetryAsync()](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/device/src/IotHubBaseClient.cs#L153-L154)` .

e2e/LongHaul/device/IotHubConnectionStringHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Microsoft.Azure.IoT.Thief.Device
+{
+    internal class IotHubConnectionStringHelper
+    {
+        public string HostName { get; private set; }
+
+        public string DeviceId { get; private set; }
+
+        public IotHubConnectionStringHelper(string connectionString)
```
drwill-ms, line 14: Not sure where my comment went, but this should be the ctor.

drwill-ms (approved):

PR #3176, opened 2023-03-20
Utilize AsyncPageable for DPS query return types

timtay-microsoft:
see #3165, but for the DPS service client this time.

This PR also removes some superfluous classes like ```ContractApiResponse``` since they were needless abstractions that made it harder to implement this change

I also change all the ```CreateAsync``` Query APIs to just ```Create``` since the function itself is not async. The first service request isn't made until the iteration begins.

I also removed the ```IContractApiHttp``` interface since it only had one implementation and we don't currently allow users to override this implementation in any way.


SDK v2 migration guide.md
```diff
@@ -284,8 +284,9 @@ These span across all service clients.
 #### Notable breaking changes
 
 - Operations that offer concurrency protection using `ETag`s, now take a parameter `onlyIfUnchanged` that relies on the ETag property of the submitted entity.
-- `IotHubServiceClient.Query.CreateAsync<T>(...)` is now async.
```
timtay-microsoft, line 287: This function is no longer async since it just builds the query (the first service request isn't made until the iteration begins) so I'm stripping the ```async``` suffix here and elsewhere

provisioning/service/src/Utilities/QueryBuilder.cs
```diff
@@ -0,0 +1,67 @@
...
+        private const string ContinuationTokenHeaderKey = "x-ms-continuation";
+        private const string ItemTypeHeaderKey = "x-ms-item-type";
+        private const string PageSizeHeaderKey = "x-ms-max-item-count";
+        private const string QueryUriFormat = "{0}/query";
+
+        internal static async Task<Page<T>> BuildAndSendRequestAsync<T>(ContractApiHttp contractApiHttp, RetryHandler retryHandler, string query, Uri path, string continuationToken, int? pageSizeHint, CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            Dictionary<string, string> headers = new Dictionary<string, string>();
```
patilsnr, line 26: nit: `var` or `Dictionary<string, string> headers = new();`

patilsnr (approved):

provisioning/service/samples/getting started/CleanupEnrollmentsSample/CleanupEnrollmentsSample.cs
```diff
@@ -48,48 +46,39 @@ public async Task RunSampleAsync()
         private async Task QueryAndDeleteIndividualEnrollmentsAsync()
         {
             Console.WriteLine("Creating a query for enrollments...");
-            Query query = _provisioningServiceClient.IndividualEnrollments.CreateQuery("SELECT * FROM enrollments", QueryPageSize);
-            while (query.HasNext())
+            AsyncPageable<IndividualEnrollment> query = _provisioningServiceClient.IndividualEnrollments.CreateQuery("SELECT * FROM enrollments");
+            List<IndividualEnrollment> individualEnrollments = new List<IndividualEnrollment>();
```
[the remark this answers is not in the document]
drwill-ms: var. new() only for class members.

provisioning/service/src/Models/QueriedPage.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        // Payload is taken separately from http response because reading the payload should only be done
+        // in an async function.
+        internal QueriedPage(HttpResponseMessage response, string payload)
+        {
+            Items = JsonConvert.DeserializeObject<IReadOnlyList<T>>(payload);
+            ContinuationToken = response.Headers.SafeGetValue(ContinuationTokenHeader);
+        }
+
+        [JsonProperty("items")]
```
drwill-ms, line 25: For these to work to deserialize, you'll need a parameterless ctor. Is this class only used for serializing?

provisioning/service/src/Models/QueryResponse.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                _httpHeaders.Add(new HttpHeader(header.Key, header.Value.First()));
+            }
+        }
+
+        public override int Status => (int)_httpResponse.StatusCode;
+
+        public override string ReasonPhrase => _httpResponse.ReasonPhrase;
+
+        public override Stream ContentStream
```
drwill-ms, line 41: Can we use an autoproperty?

provisioning/service/src/Models/QueriedPage.cs
timtay-microsoft: This class isn't used for serialization, so I'll get rid of these Json property tags

provisioning/service/src/Utilities/QueryBuilder.cs
```diff
@@ -0,0 +1,67 @@
...
+                                null,
+                                JsonConvert.SerializeObject(new QuerySpecification(query)),
+                                new ETag(),
+                                cancellationToken)
+                            .ConfigureAwait(false);
+                    },
+                    cancellationToken)
+                .ConfigureAwait(false);
+
+            Stream responseStream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
```
drwill-ms, line 58: Does this need a `using`?

drwill-ms (approved):

provisioning/service/src/Utilities/QueryBuilder.cs
timtay-microsoft: No because it is given to the user to view as part of the ```Response<T>``` pattern that AsyncEnumerable uses. It is the user's responsibility to dispose the stream here.

provisioning/service/src/Models/QueryResponse.cs
timtay-microsoft: Yes

provisioning/service/src/Utilities/QueryBuilder.cs
drwill-ms: Because it is part of QueryResponse, which is also IDisposable?

drwill-ms: Is there an E2E test or a sample that shows how to dispose it?

timtay-microsoft:
Add samples in xml docs for disposing when iterating by page, make sure we are following that pattern in our samples/e2e tests

timtay-microsoft:
Have the sample xml docs not print out the raw individual enrollment type class. Print out something interesting instead

PR #3178, opened 2023-03-20
Drop dependency on the hub service client from the device client

brycewang-microsoft:


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -55,13 +55,9 @@
...
   </ItemGroup>
 
-  <!--Project reference dependencies-->
-  <ItemGroup>
-    <ProjectReference Include="..\..\service\src\Microsoft.Azure.Devices.csproj" />
-  </ItemGroup>
-
   <!--Nuget package dependencies-->
   <ItemGroup>
+    <PackageReference Include="Azure.Core" Version="1.26.0" />
```
brycewang-microsoft, line 60: It seems that we have dependency on some libraries like "System.Text.Json" taken from "Azure.Core". That's why I am not directly referring to "System.Text.Json" here.

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -55,13 +55,9 @@
     </None>
   </ItemGroup>
 
-  <!--Project reference dependencies-->
-  <ItemGroup>
-    <ProjectReference Include="..\..\service\src\Microsoft.Azure.Devices.csproj" />
```
abhipsaMisra, line 60: Why did we even have this dependency to begin with?

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
abhipsaMisra: Do we use any other type from Azure.Core?

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
brycewang-microsoft: Seems that we were fixing some merge conflict [errors](https://github.com/Azure/azure-iot-sdk-csharp/commit/cb85816a2fdad9dce0ec2c528ea16772b1702917#diff-1667ae85c249af1b35977a22138d469b46fe76aa7859a5739a15f44610f4a8fc). 

But I think it makes no sense to correlate two clients together, so I would delete it anyway.

iothub/device/src/Microsoft.Azure.Devices.Client.csproj
brycewang-microsoft: I found this only so far. Another reason to use "Azure.Core" is that it could be hard to decide the proper version of "System.Text.Json" to depend on. I tried with the latest version of "System.Text.Json" but it gave me lots of build errors regarding version conflicts or package downgrades. So I just refer to how we depend on "Azure.Core" on the [service ](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/service/src/Microsoft.Azure.Devices.csproj#L63)client.

drwill-ms: We get ETag from Azure.Core, IIRC. Also, the JsonPatchDocument, but that isn't in our API surface, just as a recommendation, I think.

drwill-ms (approved):

PR #3180, opened 2023-03-21
Npatilsen/e2e coverage

patilsnr:


timtay-microsoft (approved):

e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -122,6 +123,59 @@ public async Task MessagingClient_CanReopenClosedClient(IotHubTransportProtocol
...
+
+            // assert
+            await act.Should().NotThrowAsync();
+        }
+
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        [DataRow(IotHubTransportProtocol.Tcp)]
+        [DataRow (IotHubTransportProtocol.WebSocket)]
+        public async Task MessagingClient_SendMessageOnClosedClient_ThrowsIotHubServiceException(IotHubTransportProtocol protocol)
```
abhipsaMisra, line 154: This seems to also be testing the client's internal behavior, independent of service's behavior. How about we refactor this into a unit test as well?

e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -122,6 +123,59 @@ public async Task MessagingClient_CanReopenClosedClient(IotHubTransportProtocol
             await sender.Messages.CloseAsync().ConfigureAwait(false);
         }
 
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        [DataRow(IotHubTransportProtocol.Tcp)]
+        [DataRow(IotHubTransportProtocol.WebSocket)]
+        public async Task MessagingClient_OpeningAlreadyOpenClient_DoesNotThrow(IotHubTransportProtocol protocol)
```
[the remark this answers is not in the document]
patilsnr: This test class has the following comment:


        // Unfortunately, the way AmqpServiceClient is implemented, it makes mocking the required amqp types difficult
        // (the amqp types are private members of the class, and cannot be set from any public/ internal API).
        // For this reason the following test is tested in the E2E flow, even though this is a unit test scenario.


It's possible that some of the changes Sophia made now make these classes mockable. I can look into it.

e2e/test/iothub/service/MessagingClientE2ETests.cs
abhipsaMisra: Actually, it is interesting that we are expecting an IotHubServiceException and not an InvalidOperationException. Do you know what the underlying error message is? Since we've already closed the client it should be SDK generated.
I wonder if we should be throwing an InvalidOperationException instead.

e2e/test/iothub/service/MessagingClientE2ETests.cs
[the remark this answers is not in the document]
abhipsaMisra: Ah, I see. That explains, thanks!

abhipsaMisra (approved):
Question if an operation on a closed client should throw IotHubServiceException or InvalidOperationException.

e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -122,6 +123,59 @@ public async Task MessagingClient_CanReopenClosedClient(IotHubTransportProtocol
...
+            {
+                Protocol = protocol
+            };
+
+            await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(DevicePrefix).ConfigureAwait(false);
+            using var serviceClient = new IotHubServiceClient(TestConfiguration.IotHub.ConnectionString, options);
+
+            // act
+            // Call OpenAsync on already open client
+            await serviceClient.Messages.OpenAsync().ConfigureAwait(false);
```
drwill-ms, line 143: needs a try/finally with a call to close

e2e/test/iothub/service/MessagingClientE2ETests.cs
```diff
@@ -122,6 +123,59 @@ public async Task MessagingClient_CanReopenClosedClient(IotHubTransportProtocol
...
+                Protocol = protocol
+            };
+
+            await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync (DevicePrefix).ConfigureAwait(false);
+            using var serviceClient = new IotHubServiceClient(TestConfiguration.IotHub.ConnectionString, options);
+
+            // act
+            var message = new Message(new byte[10]);
+            await serviceClient.Messages.OpenAsync().ConfigureAwait(false);
+            await serviceClient.Messages.SendAsync(testDevice.Id, message).ConfigureAwait(false);
```
drwill-ms, line 168: Is this step necessary?

e2e/test/iothub/service/PurgeMessageQueueE2ETests.cs
```diff
@@ -42,5 +47,28 @@ public async Task PurgeMessageQueueOperation()
...
+
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task PurgeMessageQueueOperation_ThrowsIotHubServiceExeception()
+        {
+            // arrange
+            await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(_devicePrefix).ConfigureAwait(false);
+            string faultyId = "some wrong Id";
+            string expectedDeviceId = testDevice.Device.Id;
+            using var sc = new IotHubServiceClient(TestConfiguration.IotHub.ConnectionString);
```
drwill-ms, line 59:
```suggestion
            IotHubServiceClient sc = TestDevice.ServiceClient;
```

drwill-ms: Because you don't need it initialized with any specific options like protocol.

drwill-ms (approved):

e2e/test/iothub/service/MessagingClientE2ETests.cs
patilsnr: I agree `InvalidOperation` would be better. Changing it in `AmqpSesionHandler`

PR #3189, opened 2023-03-22
Simplify PayloadConvention to enable non-JSON payloads

drwill-ms:


iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -82,12 +83,6 @@ public DirectMethodRequest(string methodName)
         [JsonIgnore]
         public string RequestId { get; protected internal set; }
 
-        /// <summary>
-        /// The JSON payload in JRaw type.
-        /// </summary>
-        [JsonProperty("payload", NullValueHandling = NullValueHandling.Include)]
-        protected internal JRaw JsonPayload { get; set; }
```
drwill-ms, line 89: We didn't seem to use this from the client, nor even use the JsonProperty attributes.

iothub/device/src/Serialization/DefaultPayloadConvention.cs
```diff
@@ -1,19 +1,40 @@
...
+        /// <summary>
+        /// A static instance of JsonSerializerSettings which sets DateParseHandling to None.
+        /// </summary>
+        /// <remarks>
+        /// By default, serializing/deserializing with Newtonsoft.Json will try to parse date-formatted
+        /// strings to a date type, which drops trailing zeros in the microseconds portion. By
+        /// specifying DateParseHandling with None, the original string will be read as-is. For more details
+        /// about the known issue, see https://github.com/JamesNK/Newtonsoft.Json/issues/1511.
+        /// </remarks>
+        private static readonly JsonSerializerSettings s_settings = new()
```
drwill-ms, line 30: Imports all the relevant functionality from the NewtonsoftJsonPayloadSerializer and Utf8PayloadEncoder types.

iothub/device/src/Serialization/PayloadConvention.cs
```diff
@@ -1,35 +1,60 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// The payload convention class. It is used to define a specific serializer as well as a specific content encoding.
+    /// The payload convention class.
     /// </summary>
     public abstract class PayloadConvention
```
drwill-ms, line 13: Imports all the relevant API surface from the PayloadSerializer and PayloadEncoder types.

iothub/device/src/Serialization/SystemTestJsonPayloadConvention.cs
```diff
@@ -1,22 +1,56 @@
...
+
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// A <see cref="PayloadConvention"/> that uses <see cref="SystemTextJsonPayloadSerializer"/>.
+    /// An implementation of the <see cref="PayloadConvention"/> class for System.Text.Json".
     /// </summary>
+    /// <remarks>
+    /// This class makes use of the <see cref="JsonSerializer"/> serializer and the <see cref="Encoding.UTF8"/> encoder.
+    /// </remarks>
```
drwill-ms, line 16: Imports all the relevant functionality from the SystemTextJsonPayloadSerializer and Utf8PayloadEncoder types.

iothub/device/src/Transport/Amqp/AmqpConnectionHolder.cs
```diff
@@ -38,7 +38,7 @@ public AmqpConnectionHolder(IConnectionCredentials connectionCredentials, IotHub
             AdditionalClientInformation additionalClientInformation,
             IotHubClientAmqpSettings amqpSettings,
             Func<DirectMethodRequest, Task> onMethodCallback,
-            Action<AmqpMessage, string, IotHubClientException> twinMessageListener,
+            Func<AmqpMessage, string, IotHubClientException, Task> twinMessageListener,
```
drwill-ms, line 41: We need the twinMessageListener to be async so we can call the DeserializeAsync method that takes a stream

iothub/device/src/Transport/AmqpIot/AmqpIotReceivingLink.cs
```diff
@@ -266,12 +266,12 @@ private void OnTwinChangesReceived(AmqpMessage amqpMessage)
                             };
                         }
 
-                        _onTwinMessageReceived.Invoke(null, correlationId, twinException);
+                        _onTwinMessageReceived.Invoke(null, correlationId, twinException).GetAwaiter().GetResult();
```
drwill-ms, line 269: This is a callback defined by AMQP and it doesn't support async so we risk using `GetAwaiter().GetResult()`.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -1105,15 +1104,10 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
...
+                            // Use the encoder that has been agreed to between the client and service to decode the byte[] response
                             // The response is deserialized into an SDK-defined type based on service-defined NewtonSoft.Json-based json property name.
                             // For this reason, we use NewtonSoft Json serializer for this deserialization.
-                            TwinDocument clientTwinProperties = JsonConvert
-                                .DeserializeObject<TwinDocument>(
-                                    _payloadConvention
-                                    .PayloadEncoder
-                                    .ContentEncoding
-                                    .GetString(payloadBytes));
+                            TwinDocument clientTwinProperties = DefaultPayloadConvention.Instance.GetObject<TwinDocument>(payloadBytes);
```
drwill-ms, line 1110: We hardcode to use Newtonsoft.Json by using our DefaultPayloadConvention class.

patilsnr (approved):
optimistic sign off

iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -128,7 +123,7 @@ public bool TryGetPayload<T>(out T payload)
 
             try
             {
-                payload = PayloadConvention.PayloadSerializer.DeserializeToType<T>(GetPayloadAsJsonString());
+                payload = PayloadConvention.GetObject<T>(Payload);
                 return true;
             }
             catch (Exception)
```
rido-min, line 129: this seems like exception swallowing. Why catch if we are not doing anything with it? At least, it should be reported via Logs, or let the exception flow and be captured later.

rido-min (changes requested):
1. As suggested in https://github.com/Azure/azure-iot-sdk-csharp/discussions/2977 the _convention_ should be an interface (or abstract class) to serialized from/to bytes, there is no need to encode any string. Which encoding should I use when using Protobuf?
2. The serializer should be configured at operation (Telemetry, Property or Command) level. If I want to use AVRO for Telemetry, then I cannot use Twins or Commands.
3. Have consistent API surface using Generics<T> instead of `object`, providing an overload in TelemetryMessage to use the same `T`

iothub/device/src/Serialization/PayloadConvention.cs
```diff
@@ -1,35 +1,60 @@
...
         /// </summary>
-        /// <remarks>This will use the <see cref="PayloadSerializer"/> and <see cref="PayloadEncoder"/> to create the byte array.</remarks>
-        /// <param name="objectToSendWithConvention">The convention-based message to be sent.</param>
         /// <returns>The correctly encoded object for this convention.</returns>
-        public virtual byte[] GetObjectBytes(object objectToSendWithConvention)
-        {
-            string serializedString = PayloadSerializer.SerializeToString(objectToSendWithConvention);
-            return PayloadEncoder.EncodeStringToByteArray(serializedString);
-        }
+        public abstract byte[] GetObjectBytes(object objectToSendWithConvention);
```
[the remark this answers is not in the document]
drwill-ms: Because JSON doesn't care what the T is when serializing. Will Protobuff or AVRO care?

iothub/device/src/Serialization/PayloadConvention.cs
```diff
@@ -1,35 +1,60 @@
...
+        /// <value>A string representing the content type to use when sending a payload.</value>
+        public abstract string ContentType { get; }
 
         /// <summary>
-        /// Gets the encoder used for the payload to be serialized.
+        /// The encoding used for the payload.
         /// </summary>
-        /// <value>An encoder that will be used to convert the serialized string to a byte array.</value>
-        public abstract PayloadEncoder PayloadEncoder { get; }
+        public abstract Encoding ContentEncoding { get; }
```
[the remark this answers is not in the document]
drwill-ms: This probably should just be a string of what the encoding should be set to, not a .NET object that is limited in the options available.

iothub/device/src/DirectMethod/DirectMethodRequest.cs
drwill-ms: It is a "Try" .NET method pattern like `int.TryParse(...)`, meaning it will not throw.

I don't see huge value in logging out the exception.

drwill-ms:
> 1. As suggested in #2977 the _convention_ should be an interface (or abstract class) to serialized from/to bytes

The base class is an abstract class.

> there is no need to encode any string. Which encoding should I use when using Protobuf?

I'll change it to a string. It is used when sending telemetry to set the content encoding property.

> 2. The serializer should be configured at operation (Telemetry, Property or Command) level. If I want to use AVRO for Telemetry, then I cannot use Twins or Commands.

We don't have the design sufficiently complicated to enable someone to change their convention per hub primitive nor per operation. While we can see someone maybe wanting to do that, we don't see it as mainstream enough to warrant the cost to implement/maintain nor the added complexity to the API.

> 3. Have consistent API surface using Generics instead of `object`, providing an overload in TelemetryMessage to use the same `T`

In JSON, serialization doesn't require the type to serialize. All classes inherit from System.Object, so that is sufficient. If you see a need to do that for binary conventions, let us know.

rido-min:
1. The comment is not about interface vs abstract class, is about the signatures, there is no need to serialize from/to string, only byte[]. Content Encoding should be optional, only used by string-based serializers.
2. I got it, then I would limit the usage of custom PayloadSerializers for just Telemetry, and always use JSONUtf8 for DirectMethods and Twins.
3. Using `T` will enable the compiler to catch if the type used in new `TelemetryMessage` matches the type registered in the PayloadSerializer

Why is it called _convention_? I dont see any convention here, just a way to provide custom serializers. 
Maybe rename `PayloadConvention`  to `TelemetryPayloadSerializer`

drwill-ms:
> The comment is not about interface vs abstract class, is about the signatures, there is no need to serialize from/to string, only byte[]. Content Encoding should be optional, only used by string-based serializers.

Okay, I didn't know what part of your sentence the key point was, then. If your only complaint is that there is a method that takes a string and returns a type specified by the user, I found that is required for getting a type from a value on a twin. The payload convention class is used for that too and must be JSON. When we're performing twin operations, the client is configurable to customize how that is deserialized.

> I got it, then I would limit the usage of custom PayloadSerializers for just Telemetry, and always use JSONUtf8 for DirectMethods and Twins.

What? Why? We can use it for telemetry, direct methods, C2D, and twins. They all have an opportunity de/serialize a custom payload.

> Using T will enable the compiler to catch if the type used in new TelemetryMessage matches the type registered in the PayloadSerializer

I don't understand. This isn't required for JSON serialization. The type passed by the user is a real type, but the method doesn't need to know it, though. That's why `JsonConvert.SerializeObject()` isn't a generic.

However, I accept it might matter for binary serialization. Do you know that it does?

> Why is it called convention? I dont see any convention here, just a way to provide custom serializers.

A service and device pass payloads back and forth in a serialized state by a convention they choose. The Hub dictates JSON for Twins and promotes JSON for everything else.

> Maybe rename PayloadConvention to TelemetryPayloadSerializer

It isn't just for telemetry.

iothub/device/src/Serialization/DefaultPayloadConvention.cs
```diff
@@ -22,9 +44,48 @@ private DefaultPayloadConvention()
...
+            string payloadString = s_encoding.GetString(objectToConvert);
+            return GetObject<T>(payloadString);
+        }
+
+        /// <inheritdoc/>
+        public override Task<T> GetObjectAsync<T>(Stream streamToConvert)
+        {
+            using var sw = new StreamReader(streamToConvert);
+            using var jtr = new JsonTextReader(sw);
+            return Task.FromResult(s_jsonSerializer.Deserialize<T>(jtr));
```
abhipsaMisra, line 76: Is there an implementation where we expect this to be async? I'm wondering why we have this declared as async if we have to do a Task.FromResult anyway?

iothub/device/src/Serialization/PayloadConvention.cs
```diff
@@ -1,35 +1,63 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
-    /// The payload convention class. It is used to define a specific serializer as well as a specific content encoding.
+    /// The payload convention class.
     /// </summary>
     public abstract class PayloadConvention
```
abhipsaMisra, line 13: I think an example of how one would use PayloadConvention would be helpful here. What do you think?

iothub/device/src/Twin/PropertyCollection.cs
```diff
@@ -100,7 +101,7 @@ public bool TryGetValue<T>(string propertyKey, out T propertyValue)
         /// </summary>
         public string GetSerializedString()
         {
-            return PayloadConvention.PayloadSerializer.SerializeToString(_properties);
+            return JsonConvert.SerializeObject(_properties);
```
abhipsaMisra, line 104: can we do this by referencing the default convention?

abhipsaMisra (approved):
The changes overall look good to me. I have some small questions here and there, but nothing major.

My apologies again on the delay it took getting to this PR. 😑

iothub/device/src/Serialization/DefaultPayloadConvention.cs
drwill-ms: STJ requires this be async so I changed PayloadConvention interface to be Task<T>.

I could revert that and simply use `GetAwaiter().GetResult()` here. At the time, I tried to make it async all the way up the stack, until I found the AMQP layer isn't async where I need it to be.

However, it eventually will be. We need the new major version to come out.

Thoughts?

iothub/device/src/Transport/Amqp/AmqpUnit.cs
```diff
@@ -1013,19 +1013,22 @@ private async Task OpenTwinSenderLinkAsync(AmqpIotSession amqpIotSession, string
...
+        private async Task OnDesiredPropertyReceivedAsync(AmqpMessage responseFromService, string correlationId, IotHubClientException ex = default)
         {
             if (Logging.IsEnabled)
-                Logging.Enter(this, responseFromService, nameof(OnDesiredPropertyReceived));
+                Logging.Enter(this, responseFromService, nameof(OnDesiredPropertyReceivedAsync));
 
             try
             {
-                _twinMessageListener?.Invoke(responseFromService, correlationId, ex);
+                if (_twinMessageListener != null)
```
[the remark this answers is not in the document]
drwill-ms: I mean, we didn't choose to before. You make a good case for it, but I'd probably go with a Debug.Assert instead so we only find it during testing.

Not really related to this PR, though.

I had to change this because you can't use the null conditional operator on an async call.

iothub/device/src/Transport/AmqpIot/AmqpIotReceivingLink.cs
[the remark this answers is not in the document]
drwill-ms: Not awaiting tasks is potentially dangerous so I wouldn't suggest doing it lightly.

First, note that the finally block will fire before the task is complete and the log to exit will show up in the logs, which would be misleading.

Also, if that task throws, then it couldn't be caught here nor up the stack anywhere.

iothub/device/src/Twin/PropertyCollection.cs
drwill-ms: We lost the half step of going from an object to a string, and not straight to bytes.

I could expose that on the default one, though.

tmahmood-microsoft (approved):

PR #3196, opened 2023-03-27
Fix bug where reconnection failed if client had been previously closed

abhipsaMisra:
There are two updates that this PR brings in:
- the code to open a client was duplicated in two places:
    - client is opened explicitly
    - client is opened as a result of a reconnection attempt
  This PR merges both of those flows
- if a client was opened after it had been previously closed, any reconnection attempt as a result of transport disconnection would fail. This is because the cancellation token source influencing subsequent reconnection attempt was canceled when CloseAsync() was called, and it was never reinitialized.

### DIFF
--- a/iothub/device/src/Pipeline/RetryDelegatingHandler.cs
+++ b/iothub/device/src/Pipeline/RetryDelegatingHandler.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
         private bool _deviceReceiveMessageEnabled;
 
         private Task _transportClosedTask;
-        private readonly CancellationTokenSource _handleDisconnectCts = new();
+        private CancellationTokenSource _handleDisconnectCts;
 
         private readonly Action<ConnectionStatusInfo> _onConnectionStatusChanged;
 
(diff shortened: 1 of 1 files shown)


iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -636,10 +641,12 @@ private async Task VerifyIsOpenAsync(CancellationToken cancellationToken)
             }
         }
 
-        private async Task OpenInternalAsync(CancellationToken cancellationToken)
+        private async Task OpenInternalAsync(bool recoverSubscriptions, CancellationToken cancellationToken)
         {
             var connectionStatusInfo = new ConnectionStatusInfo();
 
+            _handleDisconnectCts ??= new CancellationTokenSource();
```
abhipsaMisra, line 648: The cancellation token guiding reconnection attempts is initialized when the client is opened. This cancellation token source is canceled, disposed and set to null when the client is either closed, it is disposed or the connection attempt fails.

brycewang-microsoft (approved):

PR #3197, opened 2023-03-27
new direct method service client tests

patilsnr:
adds two supplemental e2e tests that cover direct method to supplement those in `MethodE2ETests` which is part of the device client e2e namespace.


e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                Payload = "10",
+                ResponseTimeout = TimeSpan.FromSeconds(30),
+            };
+
+            // act
+            Func<Task> act1 = async () => await serviceClient.DirectMethods.InvokeAsync(testDevice.Id, methodInvocation, cts);
+
+            // assert
+            var errorContext = await act1.Should().ThrowAsync<IotHubServiceException>();
```
abhipsaMisra, line 45: nit: the type isn't obvious here so Id' recommend replacing `var` by the explicit type.

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+            errorContext.And.StatusCode.Should().Be(System.Net.HttpStatusCode.NotFound);
+            errorContext.And.IsTransient.Should().BeTrue();
```
abhipsaMisra, line 47: This reads to me as a non-transient error, i.e. retrying this operation isn't going to change the result unless some external operation (setting a handler on the client) has been taken to rectify the situation. Can you think of a scenario where we'd want this to be marked transient?

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+                actualRequest = request;
+                return Task.FromResult(new DirectMethodResponse(200));
+            }).ConfigureAwait(false);
+
+            // act
+            Func<Task> act2 = async () => await serviceClient.DirectMethods.InvokeAsync(testDevice.Id, methodInvocation, cts);
+
+            // assert
+            var response = await act2.Should().NotThrowAsync();
+            actualRequest.Should().NotBeNull();
```
abhipsaMisra, line 65: You could additionally assert the status code returned by the client's response.

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    public class DirectMethodE2ETests : E2EMsTestBase
+    {
+        private readonly string _devicePrefix = $"{nameof(DirectMethodE2ETests)}_";
+        private readonly string _modulePrefix = $"{nameof(DirectMethodE2ETests)}_Module_";
+
+
```
drwill-ms, line 26: nit: collapse multiple empty lines to 1
```suggestion

```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task DirectMethodsClient_InvokeAsync_MethodDoesNotExistAtFirst()
```
drwill-ms, line 29:
```suggestion
        public async Task DirectMethodsClient_InvokeAsync_DeviceDoesNotExist()
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+            }).ConfigureAwait(false);
+
+            // act
+            Func<Task> act2 = async () => await serviceClient.DirectMethods.InvokeAsync(testDevice.Id, methodInvocation, cts);
+
+            // assert
+            var response = await act2.Should().NotThrowAsync();
+            actualRequest.Should().NotBeNull();
+
+            await deviceClient.CloseAsync();
```
drwill-ms, line 67:
```suggestion
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+            var cts = new CancellationToken();
+
+            var methodInvocation = new DirectMethodServiceRequest("SetTelemetryInterval")
+            {
+                Payload = "10",
+                ResponseTimeout = TimeSpan.FromSeconds(30),
+            };
+
+            // act
+            Func<Task> act1 = async () => await serviceClient.DirectMethods.InvokeAsync("some nonexistent device", methodInvocation, cts);
```
brycewang-microsoft, line 42: With this, this test seems more correlated to non-existing **device** instead of not registered **method**. I assume the testing purpose is to compare 2 scenarios - invoking methods before and after registering the method callback on the device client. If so, I would recommend to initiate a device client before this line, and call `InvokeAsync` with it.

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            // arrange
+            using IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+            var cts = new CancellationToken();
+
+            var methodInvocation = new DirectMethodServiceRequest("SetTelemetryInterval")
+            {
+                Payload = "10",
+                ResponseTimeout = TimeSpan.FromSeconds(30),
+            };
```
drwill-ms, line 39: In this case, the payload is not important.
```suggestion
            var methodInvocation = new DirectMethodServiceRequest("SomeMethod");
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+            // assert
+            var response = await act2.Should().NotThrowAsync();
+            actualRequest.Should().NotBeNull();
+
+            await deviceClient.CloseAsync();
+        }
+
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task DirectMethodsClient_InvokeAsycn_MethodDoesNotExist_ModuleId()
```
drwill-ms, line 72:
```suggestion
        public async Task DirectMethodsClient_InvokeAsync_ModuleDoesNotExist()
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            // arrange
+            using IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+            await using TestModule testModule = await TestModule.GetTestModuleAsync(_devicePrefix, _modulePrefix);
```
drwill-ms, line 76:
```suggestion
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            await deviceClient.CloseAsync();
+        }
+
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task DirectMethodsClient_InvokeAsycn_MethodDoesNotExist_ModuleId()
+        {
+            // arrange
+            using IotHubServiceClient serviceClient = TestDevice.ServiceClient;
```
drwill-ms, line 75: Don't do this or this test will dispose the shared instance.
```suggestion
            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,94 @@
...
+            var cts = new CancellationToken();
+
+            var methodInvocation = new DirectMethodServiceRequest("SetTelemetryInterval")
+            {
+                Payload = "10",
+                ResponseTimeout = TimeSpan.FromSeconds(30),
+            };
+
+            // act
+            Func<Task> act = async () => await serviceClient.DirectMethods.InvokeAsync(testModule.Id, methodInvocation, cts);
```
drwill-ms, line 86:
```suggestion
            var methodInvocation = new DirectMethodServiceRequest("SomeMethod");

            // act
            Func<Task> act = async () => await serviceClient.DirectMethods.InvokeAsync(testModule.Id, methodInvocation);
```

brycewang-microsoft, line 86: I don't think we have `InvokeAsync` calls with "moduleId" as the 1st parameter. The closed one would be using both of the "deviceId" and "moduleId" together.

https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/service/src/DirectMethod/DirectMethodsClient.cs

patilsnr: yup, mistake. Thanks!

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,136 @@
...
+        private readonly string _modulePrefix = $"{nameof(DirectMethodE2ETests)}_Module_";
+
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task DirectMethodsClient_InvokeAsync_DeviceDoesNotExist()
+        {
+            // arrange
+            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+
+            var methodInvocation = new DirectMethodServiceRequest("someDirectMethod");
```
brycewang-microsoft, line 36: nit: add an extra line after comment when it applies for a code block.
```suggestion
            // arrange
            
            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
            var methodInvocation = new DirectMethodServiceRequest("someDirectMethod");
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            // act
+            Func<Task> act = async () => await serviceClient.DirectMethods.InvokeAsync("someNonexistentDevice", methodInvocation);
+
+            // assert
+            ExceptionAssertions<IotHubServiceException> errorContext = await act.Should().ThrowAsync<IotHubServiceException>();
+            errorContext.And.StatusCode.Should().Be(System.Net.HttpStatusCode.NotFound);
+            errorContext.And.IsTransient.Should().BeFalse();
```
brycewang-microsoft, line 44: nit: one extra line between comment and code block
```suggestion
            // assert
            
            ExceptionAssertions<IotHubServiceException> errorContext = await act.Should().ThrowAsync<IotHubServiceException>();
            errorContext.And.StatusCode.Should().Be(System.Net.HttpStatusCode.NotFound);
            errorContext.And.IsTransient.Should().BeFalse();
```

brycewang-microsoft: Do we need to check the `ErrorCode` [property ](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/service/src/Exceptions/IotHubServiceException.cs#L91)here as well?

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,136 @@
...
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task DirectMethodsClient_InvokeAsycn_ModuleDoesNotExist()
+        {
+            // arrange
+            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+
+            var methodInvocation = new DirectMethodServiceRequest("someDirectMethod");
+
+            // act
+            Func<Task> act = async () => await serviceClient.DirectMethods.InvokeAsync("someNonexistentDevice", "someNonexistentModule", methodInvocation);
```
brycewang-microsoft, line 57: For a test with "ModuleDoesNotExist", I'd assume we are verifying the case when the device exists but the module doesn't. Otherwise, the assertion would just "fail" due to non-existence of device.

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,136 @@
...
+            // arrange
+            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+            await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(_devicePrefix);
+
+            IotHubDeviceClient deviceClient = testDevice.CreateDeviceClient();
+            await testDevice.OpenWithRetryAsync().ConfigureAwait(false);
+            var methodInvocation = new DirectMethodServiceRequest("someDirectMethod");
+
+            // ensure device client is closed
+            await deviceClient.CloseAsync();
```
brycewang-microsoft, line 78: How about we add a comment stating something like "We don't explicitly call open on the device client for testing this particular case." instead?

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,136 @@
...
+            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+
+            var methodInvocation = new DirectMethodServiceRequest("someDirectMethod");
+
+            // act
+            Func<Task> act = async () => await serviceClient.DirectMethods.InvokeAsync("someNonexistentDevice", methodInvocation);
+
+            // assert
+            ExceptionAssertions<IotHubServiceException> errorContext = await act.Should().ThrowAsync<IotHubServiceException>();
+            errorContext.And.StatusCode.Should().Be(System.Net.HttpStatusCode.NotFound);
```
drwill-ms, line 43: Is there an ErrorCode we can check against?

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,136 @@
...
+            // arrange
+            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+            await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(_devicePrefix);
+            var methodInvocation = new DirectMethodServiceRequest("someDirectMethod")
+            {
+                ConnectionTimeout = TimeSpan.FromSeconds(3),
+                ResponseTimeout = TimeSpan.FromSeconds(3),
+            };
+            IotHubDeviceClient deviceClient = testDevice.CreateDeviceClient();
+            await deviceClient.OpenAsync();
```
drwill-ms, line 108:
```suggestion
            await testDevice.OpenWithRetryAsync().ConfigureAwait(false);
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,136 @@
...
+            await testDevice.OpenWithRetryAsync().ConfigureAwait(false);
```
drwill-ms, line 109: OpenWithRetryAsync calls OpenAsync with a retry policy that includes bad credentials, which is sometimes needed for brand new created devices, so you don't need to call both.

```suggestion
            await testDevice.OpenWithRetryAsync().ConfigureAwait(false);
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,136 @@
...
+                .ConfigureAwait(false);
+
+                // assert
+                Func<Task> act = async() => await serviceClient.DirectMethods.InvokeAsync(testDevice.Id, methodInvocation);
+                ExceptionAssertions<IotHubServiceException> response = await act.Should().ThrowAsync<IotHubServiceException>();
+                response.And.StatusCode.Should().Be(System.Net.HttpStatusCode.BadRequest);
+            }
+            finally
+            {
+                await deviceClient.CloseAsync();
```
drwill-ms, line 132: await using on the TestDevice will close and dispose the device client, so you can get rid of the try/finally block.

e2e/test/iothub/service/DirectMethodE2ETests.cs
drwill-ms: Good point.

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,127 @@
...
+            errorContext.And.IsTransient.Should().BeFalse();
+            errorContext.And.ErrorCode.Should().Be(IotHubServiceErrorCode.ModuleNotFound);
+        }
+
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task DirectMethodsClient_DeviceOnline_DeviceNotSubscribedToDirectMethods()
+        {
+            // arrange
+            // device is online but device client is not explicitly opened
```
patilsnr, line 75: device client not explicitly opened

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            // act
+            Func<Task> act = async () => await serviceClient.DirectMethods.InvokeAsync("someNonexistentDevice", methodInvocation);
+
+            // assert
+
+            ExceptionAssertions<IotHubServiceException> errorContext = await act.Should().ThrowAsync<IotHubServiceException>();
+            errorContext.And.StatusCode.Should().Be(System.Net.HttpStatusCode.NotFound);
+            errorContext.And.IsTransient.Should().BeFalse();
+            errorContext.And.ErrorCode.Should().Be(IotHubServiceErrorCode.DeviceNotFound);
```
patilsnr, line 46: error code validations are now part of all asserts in tests

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,127 @@
...
+        [TestMethod]
+        [Timeout(TestTimeoutMilliseconds)]
+        public async Task DirectMethodsClient_DeviceOnline_DeviceNotSubscribedToDirectMethods()
+        {
+            // arrange
+            // device is online but device client is not explicitly opened
+            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+            await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(_devicePrefix);
+            
+            //IotHubDeviceClient deviceClient = testDevice.CreateDeviceClient();
```
drwill-ms, line 79:
```suggestion
```

e2e/test/iothub/service/DirectMethodE2ETests.cs
drwill-ms, line 75: Device is created, but not online, let alone subscribed.

Do you want to change the comments and test method name, or make the device online but not subscribed?

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,127 @@
...
+                    var response = new DirectMethodResponse(200);
+                    return response;
+                })
+            .ConfigureAwait(false);
+
+            // assert
+            Func<Task> act = async() => await serviceClient.DirectMethods.InvokeAsync(testDevice.Id, methodInvocation);
+            ExceptionAssertions<IotHubServiceException> response = await act.Should().ThrowAsync<IotHubServiceException>();
+            response.And.StatusCode.Should().Be(System.Net.HttpStatusCode.BadRequest);
+            response.And.ErrorCode.Should().Be(IotHubServiceErrorCode.ArgumentInvalid);
```
drwill-ms, line 124: This is surprising. Do we get the same error over AMQP?

e2e/test/iothub/service/DirectMethodE2ETests.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.Azure.Devices.Client;
+using System.Reflection;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Azure.Devices.E2ETests.iothub.service
+{
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
```
abhipsaMisra, line 23:
```suggestion
    [TestCategory("IoTHub-Service")]
```

nit: I split the "IoTHub" test category into "IoTHub-Client" and "IoTHub-Service" in the last PR.

e2e/test/iothub/service/DirectMethodE2ETests.cs
abhipsaMisra, line 75: Did you mean that the device exists but isn't explicitly opened = isn't online? A device being online would mean that it is open = connected to IoT Hub service.

e2e/test/iothub/service/DirectMethodE2ETests.cs
patilsnr: It looks like it. I can ping a service team member to see if they have insight, but it might not be something we can change.

drwill-ms (approved):

PR #3198, opened 2023-03-27
Complete received C2D message within semaphore - Device Client sample

tmahmood-microsoft:
C2D messages require to ack'ed according to the order that they were sent in. If multiple messages are in queue, and up receiving, device client sends separate events before completing the message, there is a possibility of ack'ing messages in incorrect order.

The current change would make sure a sequential completion of messages.


patilsnr (approved):

patilsnr:
Does this also need to happen in previews/v2?

iothub/device/samples/getting started/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -104,13 +110,15 @@ private async Task ReceiveC2dMessagesPollingAndCompleteAsync(CancellationToken c
 
         private async Task OnC2dMessageReceivedAsync(Message receivedMessage, object _)
         {
+            await _processMessageSemaphore.WaitAsync().ConfigureAwait(false);
```
abhipsaMisra, line 113: I'd recommend passing in a cancellation token to this `WaitAsync()` so that we aren't waiting indefinitely.

Semaphores and locks can cause deadlocks if not implemented correctly, i.e. we need to be careful how we expect them to be implemented in both the happy path and also corner cases where unexcepted exceptions are encountered. That does not mean that we should never use them, but it is definitely worth investigating alternate approaches. 

Are there any alternative approaches to solve this issue?

tmahmood-microsoft:
So far, I have not been able to reproduce this in v2/preview so dont need it in v2

iothub/device/samples/getting started/MessageReceiveSample/MessageReceiveSample.cs
tmahmood-microsoft: Discussed it with David in the morning and he suggested using semaphores as its a smaller change compared to having concurrent queue.

iothub/device/samples/getting started/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -70,6 +76,8 @@ public async Task RunSampleAsync()
 
                 // Now unsubscibe from receiving the callback.
                 await _deviceClient.SetReceiveMessageHandlerAsync(null, null);
+                _processMessageSemaphore.Dispose();
+                _cts.Dispose();
```
drwill-ms, line 80: this should be in a try/finally to ensure it gets called

iothub/device/samples/getting started/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -32,21 +34,25 @@ public MessageReceiveSample(DeviceClient deviceClient, TransportType transportTy
 
         public async Task RunSampleAsync()
         {
-            using var cts = _maxRunTime.HasValue
+            _cts = _maxRunTime.HasValue
                 ? new CancellationTokenSource(_maxRunTime.Value)
                 : new CancellationTokenSource();
+
+            // Semaphore to ensure sequential execution of received C2D messages.
+            _processMessageSemaphore = new SemaphoreSlim(1, 1);
```
drwill-ms, line 42: This needs disposal.

iothub/device/samples/getting started/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -104,13 +112,15 @@ private async Task ReceiveC2dMessagesPollingAndCompleteAsync(CancellationToken c
...
         {
+            await _processMessageSemaphore.WaitAsync(_cts.Token).ConfigureAwait(false);
             Console.WriteLine($"{DateTime.Now}> C2D message callback - message received with Id={receivedMessage.MessageId}.");
             PrintMessage(receivedMessage);
 
             await _deviceClient.CompleteAsync(receivedMessage);
             Console.WriteLine($"{DateTime.Now}> Completed C2D message with Id={receivedMessage.MessageId}.");
 
             receivedMessage.Dispose();
+            _processMessageSemaphore.Release();
```
drwill-ms, line 123: This should be in a try/finally so it is always released, even if the code above crashes.

iothub/device/samples/getting started/MessageReceiveSample/MessageReceiveSample.cs
```diff
@@ -104,13 +112,15 @@ private async Task ReceiveC2dMessagesPollingAndCompleteAsync(CancellationToken c
 
         private async Task OnC2dMessageReceivedAsync(Message receivedMessage, object _)
         {
+            await _processMessageSemaphore.WaitAsync(_cts.Token).ConfigureAwait(false);
```
drwill-ms, line 115:
```suggestion
           // Use a semaphore to ensure C2D messages are processed in order - a requirement of IoT hub
            await _processMessageSemaphore.WaitAsync(_cts.Token).ConfigureAwait(false);
```

drwill-ms (approved):

drwill-ms (approved):

PR #3200, opened 2023-03-27
Split test categories for IoT hub service and IoT hub client

abhipsaMisra:


iothub/device/tests/RetryPolicies/IotHubClientExponentialBackoffRetryPolicyTests.cs
```diff
@@ -2,15 +2,13 @@
...
 using System;
-using System.Threading;
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Azure.Devices.Client.Test
 {
     [TestClass]
     [TestCategory("Unit")]
-    [TestCategory("IoTHub")]
```
abhipsaMisra, line 13: I don't see the advantage of tagging unit tests as IoTHub. They have minimal running time and don't provide any substantial benefit if we exclude them from our runs.

tmahmood-microsoft (approved):

iothub/device/tests/RetryPolicies/IotHubClientExponentialBackoffRetryPolicyTests.cs
drwill-ms: The goal was to separate hub from dps tests.

drwill-ms (approved):

PR #3202, opened 2023-03-27
Set desired property updates callback on device app and also report device twins frequently

brycewang-microsoft:


e2e/LongHaul/device/IotHub.cs
```diff
@@ -287,22 +289,25 @@ private Task<DirectMethodResponse> DirectMethodCallback(DirectMethodRequest meth
                                 C2dDirectMethodDelaySeconds,
                                 (DateTimeOffset.UtcNow - methodPayload.CurrentTimeUtc).TotalSeconds);
 
+                            // Set "methodCallsCount" obtained from the method payload as one of the reported properties.
+                            await SetPropertiesAsync("methodCallsCount", methodPayload.MethodCallsCount).ConfigureAwait(false);
```
drwill-ms, line 293: This isn't quite what I had in mind.

The service client will send a twin property request to the device. The device just needs to respond to property update requests and set them. This will be independent of direct methods working.

We'll also want a separate running task (like telemetry) that sets a twin property on a frequency.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -128,6 +129,31 @@ public async Task RunAsync(CancellationToken ct)
...
 
+        public async Task CheckDesiredPropertiesAndUpdateAsync(CancellationToken ct)
+        {
+            while (!ct.IsCancellationRequested)
+            {
+                // If not connected, skip the work below this round
+                if (!IsConnected)
+                {
+                    _logger.Trace($"Waiting for connection before any other operations.", TraceSeverity.Warning);
+                    continue;
```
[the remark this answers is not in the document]
drwill-ms: Irrelevant based on my next comment.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -128,6 +129,31 @@ public async Task RunAsync(CancellationToken ct)
             }
         }
 
+        public async Task CheckDesiredPropertiesAndUpdateAsync(CancellationToken ct)
```
drwill-ms, line 132: This should just be about setting a read-only property in a loop. It can be based on something, like how many telemetry messages have been sent or something, or it could be gibberish, like a Guid.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -53,7 +53,7 @@ public async Task RunAsync(CancellationToken ct)
             while (!ct.IsCancellationRequested)
             {
                 await InvokeDirectMethodAsync(++methodCallsCount, ct).ConfigureAwait(false);
-                await SetDesiredPropertiesAsync("methodCallsCount", methodCallsCount, ct).ConfigureAwait(false);
+                await SetDesiredPropertiesAsync("methodCallsCount", methodCallsCount.ToString(), ct).ConfigureAwait(false);
```
brycewang-microsoft: I am still calling both methods from `RunAsync()` here as we need to pass in the `methodCallsCount` into both methods and increase its value per each loop.

brycewang-microsoft: `methodCallsCount.ToString()` as we assume all properties are string only.

e2e/LongHaul/service/Program.cs
```diff
@@ -68,9 +68,15 @@ private static async Task Main(string[] args)
...
             };
 
+            var systemHealthMonitor = new SystemHealthMonitor(s_logger.Clone());
+
             try
             {
-                await iotHub.RunAsync(cancellationTokenSource.Token).ConfigureAwait(false);
+                await Task
+                    .WhenAll(
+                        systemHealthMonitor.RunAsync(cancellationTokenSource.Token),
```
brycewang-microsoft, line 77: Frequently record system health metrics.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -52,32 +52,8 @@ public async Task RunAsync(CancellationToken ct)
...
-                if (response.TryGetPayload(out CustomDirectMethodPayload responsePayload))
-                {
-                    _logger.Metric(
-                        D2cDirectMethodDelaySeconds,
-                        (DateTimeOffset.UtcNow - responsePayload.CurrentTimeUtc).TotalSeconds);
-                }
-
-                _logger.Trace($"Response status: {response.Status}, payload:\n\t{JsonConvert.SerializeObject(response.PayloadAsString)}", TraceSeverity.Information);
+                await InvokeDirectMethodAsync(++methodCallsCount, ct).ConfigureAwait(false);
+                await SetDesiredPropertiesAsync("methodCallsCount", methodCallsCount.ToString(), ct).ConfigureAwait(false);
```
brycewang-microsoft, line 56: I am still calling both methods from `RunAsync()` here as we need to pass in the `methodCallsCount` into both methods and increase its value per each loop.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -128,6 +130,29 @@ public async Task RunAsync(CancellationToken ct)
...
+            while (!ct.IsCancellationRequested)
+            {
+                // If not connected, skip the work below this round
+                if (IsConnected)
+                {
+                    var reported = new ReportedProperties
+                    {
+                        { "TotalMessagesSent", _totalMessagesSent },
+                    };
+                    await _deviceClient.UpdateReportedPropertiesAsync(reported ,ct).ConfigureAwait(false);
```
drwill-ms, line 144: Add some logging.

drwill-ms: Maybe keep track of how many twin updates we've made and log a metric with it?

e2e/LongHaul/device/IotHub.cs
```diff
@@ -304,5 +331,27 @@ private Task<DirectMethodResponse> DirectMethodCallback(DirectMethodRequest meth
...
+                    continue;
+                }
+
+                _logger.Trace($"Got request for {prop.Key} with [{propertyValue}]", TraceSeverity.Information);
+                reported.Add(prop.Key, propertyValue);
+            }
+
+            if (reported.Any())
+            {
+                await _deviceClient.UpdateReportedPropertiesAsync(reported).ConfigureAwait(false);
```
drwill-ms, line 353: metric for # of twin callbacks handled, and how many desired properties handled?

e2e/LongHaul/service/IotHub.cs
drwill-ms: Let's just send a guid as a string and not tie the operations together.

drwill-ms (approved):

Issue #3201, opened 2023-03-27
[Bug Report] Send message to multiple outputs using ModuleClient throws


Sdelausnay:
# Context

- **OS, version, SKU and CPU architecture used:**  Raspbian GNU/Linux 9 (stretch), armv7l
- **Application's .NET Target Framework :** net6.0
- **Device:** Raspberry PI3
- **SDK version used:** Microsoft.Azure.Devices.Client 1.41.3

## Description of the issue
I have a message in a custom module which I want to send to 2 different output depending on the message content. Sending the message to the first output is no issue, but when targeting the second output gives and exception of  `An item with the same key has already been added. Key: iothub-outputname` . I have check the code but I'm sure that I'm sending the message to a different output name. 

## Code sample exhibiting the issue
```csharp

var messageJson = JsonConvert.SerializeObject(mqttMsg, JsonSerializerSettings);

var message = new Message(Encoding.UTF8.GetBytes(messageJson))
{
  ContentEncoding = "UTF-8",
  ContentType = "application/JSON"
};

_logger.LogTrace("Send to main output: {0}", Endpoints.MqttBrokerOutputName);

await _moduleClient.SendEventAsync(Endpoints.MqttBrokerOutputName, message, stoppingToken);

if (mqttMessages.FirstOrDefault().Topic.Contains("boxstatus"))
{
  _logger.LogTrace("Send to Events output: {0}", Endpoints.EventsModuleOutputName);
  await _moduleClient.SendEventAsync(Endpoints.EventsModuleOutputName, message, stoppingToken);
}

_logger.LogTrace("Messages send to all endpoints.");                 
```

## Console log of the issue 

Line 153 refers to the `await _moduleClient.SendEventAsync(Endpoints.EventsModuleOutputName, message, stoppingToken);`

<details><summary>Logs</summary>
<p>

```
<7>2023-03-27 22:08:53.293 +00:00 MqttBroker.Services.MqttBrokerService[0] Send to main output: mqtt
<7>2023-03-27 22:08:53.303 +00:00 MqttBroker.Services.MqttBrokerService[0] Send to Events output: events_output
[...]

PR #3203, opened 2023-03-28
Allow a message to be sent to multiple outputs

drwill-ms:
#3201

### DIFF
--- a/iothub/device/src/InternalClient.cs
+++ b/iothub/device/src/InternalClient.cs
@@ -707,10 +707,7 @@ namespace Microsoft.Azure.Devices.Client
             {
                 foreach (Message message in messages)
                 {
-                    if (message.MessageId == null)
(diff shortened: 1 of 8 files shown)


iothub/device/src/InternalClient.cs
```diff
@@ -1502,7 +1499,7 @@ public Task SendEventAsync(string outputName, Message message, CancellationToken
                     throw new ArgumentNullException(nameof(message));
                 }
 
-                message.SystemProperties.Add(MessageSystemPropertyNames.OutputName, outputName);
+                message.SystemProperties[MessageSystemPropertyNames.OutputName] = outputName;
```
drwill-ms, line 1502: Using the `Add()` method will fail if the key already exists. Using the array indexer will add or overwrite the value.

PR #3208, opened 2023-03-28
Update analyzers in LTS

drwill-ms:


vsts/vsts.yaml
```diff
@@ -419,12 +421,15 @@ jobs:
           choco install -y squid
 
         displayName: "Install Squid"
+        condition: eq(variables.SHOULD_RUN, 'True')
 
       - powershell: ./vsts/start_tpm_windows.ps1
```
patilsnr, line 426: nit: if this is the windows build, should we use backslashes?

patilsnr (approved):

vsts/vsts.yaml
drwill-ms: Nope, it works either way.

PR #3210, opened 2023-03-28
Remove duplicate logs

abhipsaMisra:


e2e/test/helpers/logging/VerboseTestLogger.cs
```diff
@@ -9,7 +9,6 @@ public static class VerboseTestLogger
     {
         public static void WriteLine(string message)
         {
-            Console.WriteLine(message);
```
drwill-ms, line 12: If we don't log to console and event logging isn't enabled, will we get any logs?

abhipsaMisra: If event logging isn't enabled then we won't capture any logs => nothing would be written to console.

patilsnr (approved):

PR #3212, opened 2023-03-29
Ensure device client methods throw an object disposed exception

drwill-ms:


iothub/device/src/ClientFactory.cs
```diff
@@ -535,6 +535,7 @@ private static IDeviceClientPipelineBuilder BuildPipeline()
         {
             var transporthandlerFactory = new TransportHandlerFactory();
             IDeviceClientPipelineBuilder pipelineBuilder = new DeviceClientPipelineBuilder()
+                .With((ctx, innerHandler) => new DefaultDelegatingHandler(ctx, innerHandler))
```
drwill-ms, line 538: This ensures we check if the client is disposed before doing retry logic

iothub/device/src/InternalClient.cs
```diff
@@ -1378,13 +1379,23 @@ private async Task OnDeviceMessageReceivedAsync(Message message)
             FileUploadSasUriRequest request,
             CancellationToken cancellationToken = default)
         {
+            if (_isDisposed)
```
drwill-ms, line 1382: For the methods over HTTP that don't use the regular pipeline

brycewang-microsoft (approved):

tmahmood-microsoft (approved):

patilsnr (approved):

PR #3216, opened 2023-03-30
Docs and tests for disposed module client operations

drwill-ms:


tmahmood-microsoft (approved):

iothub/device/src/InternalClient.cs
```diff
@@ -200,6 +199,7 @@ internal class InternalClient : IDisposable
             if (Logging.IsEnabled)
                 Logging.Exit(this, transportSettings, pipelineBuilder, nameof(InternalClient) + "_ctor");
         }
+        internal bool IsDisposed { get; private set; }
```
abhipsaMisra, line 202: Do you think we could make this public or would that go against IDisposable pattern (given that .NET doesn't expose such a property on IDisposable types)?

abhipsaMisra (approved):
lgtm

iothub/device/src/InternalClient.cs
drwill-ms: I'm thinking about doing it for v2 but I don't know about adding new stuff to v1.

PR #3221, opened 2023-03-30
Replace System.Text.Json with Newtonsoft.Json in HubEvents

brycewang-microsoft:
`partitionEvent.Data.SystemProperties` is type of "IReadOnlyDictionary". It appears that serializing/deserializing the `ReadOnlyMemory` type of object with `System.Text.Json` would complain the following:

Untitled

`Newtonsoft.Json` doesn't cause such a problem, so replaced.

### DIFF
--- a/e2e/LongHaul/service/HubEvents.cs
+++ b/e2e/LongHaul/service/HubEvents.cs
@@ -41,8 +41,8 @@ namespace Microsoft.Azure.Devices.LongHaul.Service
 
                     try
                     {
-                        var testProperties = JsonConvert.SerializeObject(partitionEvent.Data.SystemProperties);
-                        DeviceEventSystemProperties eventMetadata = JsonConvert.DeserializeObject<DeviceEventSystemProperties>(testProperties);
+                        DeviceEventSystemProperties eventMetadata = JsonConvert.DeserializeObject<DeviceEventSystemProperties>(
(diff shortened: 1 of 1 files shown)


e2e/LongHaul/service/HubEvents.cs
```diff
@@ -58,12 +59,20 @@ await foreach (PartitionEvent partitionEvent in eventHubClient.ReadEventsAsync(r
...
+                            {
+                                _logger.Trace(
+                                    $"Failed to convert event [{JsonConvert.SerializeObject(partitionEvent.Data.Properties)}] to DeviceEventProperties due to {ex}",
+                                    TraceSeverity.Warning);
+                            }
                         }
                     }
                     catch (Exception ex)
                     {
-                        _logger.Trace($"Failed to convert event [{JsonSerializer.Serialize(partitionEvent.Data.Properties)}] to DeviceEventProperties due to {ex}", TraceSeverity.Warning);
```
brycewang-microsoft, line 66: The exception caught here is not necessarily caused by failed serial-/deserialization of `partitionEvent.Data.Properties`. It can be possibly caused by failures on `partitionEvent.Data.SystemProperties` as well. So fixed logging contents here.

tmahmood-microsoft (approved):

brycewang-microsoft:
Close this PR as we have a workaround to keep using `System.Text.Json` for long-haul. There will be another PR to address that.

PR #3234, opened 2023-03-31
Rename and cleanup ErrorDelegatingHandler

drwill-ms:


iothub/device/src/Fx/Fx.cs
```diff
@@ -2,53 +2,48 @@
...
+            typeof(SEHException),
+            typeof(NullReferenceException),
+        };
+
         internal static bool IsFatal(Exception ex)
         {
-            while (ex != null)
+            if (s_fatalExceptionTypes.Any(fatalType => fatalType.IsAssignableFrom(ex.GetType())))
             {
-                if (ex is OutOfMemoryException
```
drwill-ms, line 17: Rather than have a list here, I made a hashset above.

iothub/device/src/Fx/Fx.cs
```diff
@@ -2,53 +2,48 @@
...
+            typeof(InvalidOperationException),
+            typeof(OutOfMemoryException),
+            typeof(SEHException),
+            typeof(NullReferenceException),
+        };
+
         internal static bool IsFatal(Exception ex)
         {
-            while (ex != null)
+            if (s_fatalExceptionTypes.Any(fatalType => fatalType.IsAssignableFrom(ex.GetType())))
```
drwill-ms, line 28: IsAssignableFrom would include an exception like TaskCanceledException which inherits from OperationCanceledException (which is in the list above).

This is like using `if (ex is OperationCanceledException)`.

iothub/device/src/Fx/Fx.cs
```diff
@@ -2,53 +2,48 @@
...
-                    return true;
-                }
-
-                // These exceptions aren't themselves fatal, but since the CLR uses them to wrap other exceptions,
-                // we want to check to see whether they've been used to wrap a fatal exception. If so, then they
-                // count as fatal.
-                if (ex is TypeInitializationException
-                    || ex is TargetInvocationException)
-                {
-                    ex = ex.InnerException;
```
drwill-ms, line 30: This was the only thing using while loop, so if instead we just use recursion as the other blocks do.

iothub/device/src/Fx/Fx.cs
```diff
@@ -2,53 +2,48 @@
...
+                || ex is TargetInvocationException)
+            {
+                return IsFatal(ex.InnerException);
+            }
+            else if (ex is AggregateException aggEx)
+            {
+                // AggregateExceptions have a collection of inner exceptions, which may themselves be other
+                // wrapping exceptions (including nested AggregateExceptions). Recursively walk this
+                // hierarchy. The (singular) InnerException is included in the collection.
+                return aggEx.InnerExceptions.Any(ie => IsFatal(ie));
```
drwill-ms, line 46: Pretty sure this does everything the previous block did. 8D

iothub/device/src/Pipeline/ExceptionRemappingHandler.cs
```diff
@@ -128,56 +130,50 @@ private static bool IsNetwork(Exception singleException)
...
+        private async Task<T> RunWithExceptionRemappingAsync<T>(Func<Task<T>> asyncOperation)
         {
+            if (Logging.IsEnabled)
+                Logging.Enter(this, $"{nameof(ExceptionRemappingHandler)}.{nameof(ExecuteWithExceptionRemappingAsync)}");
+
             try
+            {
+                return await asyncOperation().ConfigureAwait(false);
+            }
+            catch (Exception ex) when (ex is not IotHubClientException && !Fx.IsFatal(ex))
```
drwill-ms, line 151: Adding this exception (pun intended) to the catch makes sense to me and eliminates the small block on former lines 162-166. What do you think?

If the job here is to not allow most exceptions that aren't IotHubClientException, why would we catch when it isn't?

iothub/device/src/Pipeline/ExceptionRemappingHandler.cs
```diff
@@ -128,56 +130,50 @@ private static bool IsNetwork(Exception singleException)
...
-                    {
-                        throw new IotHubClientException("Transient network error occurred; please retry.", IotHubClientErrorCode.NetworkErrors, ex);
-                    }
-                    else
-                    {
-                        throw;
-                    }
+                        Logging.Error(this, $"Unmapped exception {ex.GetType()}");
+
+                    throw new IotHubClientException("An unexpected exception occurred. See the inner exception for more details.", IotHubClientErrorCode.Unknown, ex);
```
drwill-ms, line 170: I figured we should default to creating an IotHubClientException with the real one as the inner, rather than let it go through and log the error.

Thoughts?

patilsnr (approved):
optimistic sign off

iothub/device/src/Pipeline/ClientPipelineBuilder.cs
```diff
@@ -6,6 +6,11 @@
 
 namespace Microsoft.Azure.Devices.Client
 {
+    /// <summary>
+    /// The client uses a decorator pattern of tasks that act on AMQP and MQTT requests
+    /// which bring functionality such as checking for and throwing if the client is disposed,
+    /// retry logic, etc.
```
abhipsaMisra, line 12: 👍🏽

PR #3235, opened 2023-03-31
Long-haul: service app pair with online devices

brycewang-microsoft:


e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+
+        public void Dispose()
+        {
+            _logger.Trace("Disposing", TraceSeverity.Verbose);
+
+            _serviceClient?.Dispose();
+
+            _logger.Trace($"IotHub instance disposed", TraceSeverity.Verbose);
```
drwill-ms, line 135:
```suggestion
            _logger.Trace($"IotHub instance disposed for {_deviceId}", TraceSeverity.Verbose);
```

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+            finally
+            {
+                await _serviceClient.Messages.CloseAsync().ConfigureAwait(false);
+            }
+        }
+
+        public void Dispose()
+        {
+            _logger.Trace("Disposing", TraceSeverity.Verbose);
```
drwill-ms, line 131:
```suggestion
            _logger.Trace("Disposing DeviceOperations for {_deviceId}.", TraceSeverity.Verbose);
```

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -0,0 +1,138 @@
...
+                    _logger.Metric(TotalC2dMessagesSentCount, _totalC2dMessagesSentCount);
+
+                    await _serviceClient.Messages.SendAsync(_deviceId, message, ct).ConfigureAwait(false);
+
+                    await Task.Delay(s_c2dMessagesSentInterval, ct).ConfigureAwait(false);
+                }
+            }
+            finally
+            {
+                await _serviceClient.Messages.CloseAsync().ConfigureAwait(false);
```
drwill-ms, line 125: If this is a shared service client instance, you won't want to close this sub client.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -124,78 +71,37 @@ public async Task MonitorConnectedDevicesAsync(CancellationToken ct)
...
                     }
-                    else if (!s_activeDeviceSet.Contains(device.DeviceId) && device.ConnectionState == ClientConnectionState.Connected)
+                    else if (!s_onlineDeviceOperations.ContainsKey(deviceId) && device.ConnectionState is ClientConnectionState.Connected)
                     {
-                        s_activeDeviceSet.Add(device.DeviceId);
+                        s_onlineDeviceOperations.Add(
+                            deviceId,
+                            async () =>
+                            {
+                                using var deviceOperations = new DeviceOperations(s_serviceClient, deviceId, _logger);
```
drwill-ms, line 86: Clone the logger, add the device Id as a custom dimension, and pass that instance to this class.
```suggestion
                                using var deviceOperations = new DeviceOperations(s_serviceClient, deviceId, _logger);
```

e2e/LongHaul/service/IotHub.cs
```diff
@@ -124,78 +71,37 @@ public async Task MonitorConnectedDevicesAsync(CancellationToken ct)
...
                     {
-                        s_activeDeviceSet.Remove(device.DeviceId);
+                        s_onlineDeviceOperations.Remove(deviceId);
                     }
-                    else if (!s_activeDeviceSet.Contains(device.DeviceId) && device.ConnectionState == ClientConnectionState.Connected)
+                    else if (!s_onlineDeviceOperations.ContainsKey(deviceId) && device.ConnectionState is ClientConnectionState.Connected)
                     {
-                        s_activeDeviceSet.Add(device.DeviceId);
+                        s_onlineDeviceOperations.Add(
+                            deviceId,
```
drwill-ms, line 83: How do you cancel the running of these tasks for this device but not the rest of the app?

I'd suggest having device-specific cancellation token sources, so when a device disconnects, you can signal it and remove it from this list.

e2e/LongHaul/service/Parameters.cs
```diff
@@ -30,13 +30,6 @@ internal class Parameters
             HelpText = "The instrumentation key string for application insights.")]
         public string InstrumentationKey { get; set; } = Environment.GetEnvironmentVariable("APPLICATION_INSIGHTS_INSTRUMENTATION_KEY");
 
-        [Option(
-            'd',
-            "DeviceId",
-            Required = false,
-            HelpText = "The Id of the device to receive the direct method.")]
-        public string DeviceId { get; set; } = "LongHaulDevice1";
```
brycewang-microsoft, line 38: The service app will continuously query to get online devices, so we don't need this arg any more.

e2e/LongHaul/service/Program.cs
```diff
@@ -81,9 +79,6 @@ await Task
                     .WhenAll(
                         systemHealthMonitor.RunAsync(cancellationTokenSource.Token),
                         iotHub.MonitorConnectedDevicesAsync(cancellationTokenSource.Token),
-                        iotHub.InvokeDirectMethodAsync(cancellationTokenSource.Token),
```
brycewang-microsoft, line 84: Moved to IotHub.cs, will be called from `MonitorConnectedDevicesAsync()`

e2e/LongHaul/service/Program.cs
```diff
@@ -81,9 +79,6 @@ await Task
...
-                        iotHub.SetDesiredPropertiesAsync("guidValue", Guid.NewGuid().ToString(), cancellationTokenSource.Token),
-                        iotHub.SendC2dMessagesAsync(cancellationTokenSource.Token),
```
brycewang-microsoft, line 86: Moved to "service/IotHub.cs", will be called from `MonitorConnectedDevicesAsync()`

e2e/LongHaul/service/IotHub.cs
```diff
@@ -20,29 +19,24 @@ internal class IotHub : IDisposable
...
         private int _totalFileUploadNotificationsReceived;
 
         private static IotHubServiceClient s_serviceClient;
-        private static HashSet<string> s_activeDeviceSet;
         private BlobContainerClient _blobContainerClient;
 
-        private long _totalMethodCallsCount = 0;
-        private long _totalDesiredPropertiesUpdatesCount = 0;
-        private long _totalC2dMessagesSentCount = 0;
+        private static volatile Dictionary<string, Action> s_onlineDeviceOperations;
```
brycewang-microsoft, line 31: cache `Action` of service app operations per "deviceId"

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                await _serviceClient.Messages.SendAsync(_deviceId, message, ct).ConfigureAwait(false);
+
+                await Task.Delay(s_c2dMessagesSentInterval, ct).ConfigureAwait(false);
+            }
+        }
+
+        public void Dispose()
+        {
+            _logger.Trace($"{nameof(DeviceOperations)} instance disposed for the device [{_deviceId}].", TraceSeverity.Verbose);
```
drwill-ms, line 125: Why is this class disposable if there is nothing to dispose?

e2e/LongHaul/service/IotHub.cs
```diff
@@ -124,78 +75,56 @@ public async Task MonitorConnectedDevicesAsync(CancellationToken ct)
...
+                        && device.ConnectionState is ClientConnectionState.Connected)
                     {
-                        s_activeDeviceSet.Add(device.DeviceId);
+                        // For each online device, initiate a new cancellation token source.
+                        // Once the device goes offline, cancel all operations on this device.
+                        var source = new CancellationTokenSource();
+                        CancellationToken token = source.Token;
+                        s_onlineDeviceCancellationTokenSources.Add(deviceId, source);
+
+                        s_onlineDeviceOperations.Add(
```
drwill-ms, line 99: I think this would read better if you instead got a task from this and then put it in the dictionary on line 118. The way it is requires both actions to access the dictionary, which is an opportunity for a bug.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -20,29 +19,24 @@ internal class IotHub : IDisposable
...
 
         private static IotHubServiceClient s_serviceClient;
-        private static HashSet<string> s_activeDeviceSet;
         private BlobContainerClient _blobContainerClient;
 
-        private long _totalMethodCallsCount = 0;
-        private long _totalDesiredPropertiesUpdatesCount = 0;
-        private long _totalC2dMessagesSentCount = 0;
+        private static volatile Dictionary<string, Action> s_onlineDeviceOperations;
+        private static volatile Dictionary<string, CancellationTokenSource> s_onlineDeviceCancellationTokenSources;
```
drwill-ms, line 32: Consider caching both CTS and the task together in a tuple or a custom class. Then you don't need two dictionaries and risk them getting out of sync from each other.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -124,78 +75,56 @@ public async Task MonitorConnectedDevicesAsync(CancellationToken ct)
...
+                    string deviceId = device.DeviceId;
+
+                    if (s_onlineDeviceCancellationTokenSources.ContainsKey(deviceId)
+                        && s_onlineDeviceOperations.ContainsKey(deviceId)
+                        && device.ConnectionState is ClientConnectionState.Disconnected)
                     {
-                        s_activeDeviceSet.Remove(device.DeviceId);
+                        CancellationTokenSource source = s_onlineDeviceCancellationTokenSources[deviceId];
+                        source.Cancel(); // Signal cancellation to all tasks on the particular device.
+                        s_onlineDeviceCancellationTokenSources.Remove(deviceId);
```
drwill-ms, line 86: remove and dispose right?

e2e/LongHaul/service/DeviceOperations.cs
brycewang-microsoft: Oops, previously I wrongly disposed the shared service client in this class. Corrected.

tmahmood-microsoft (approved):
much better design

e2e/LongHaul/service/IotHub.cs
```diff
@@ -20,29 +19,23 @@ internal class IotHub : IDisposable
...
         private int _totalFileUploadNotificationsReceived;
 
         private static IotHubServiceClient s_serviceClient;
-        private static HashSet<string> s_activeDeviceSet;
         private BlobContainerClient _blobContainerClient;
 
-        private long _totalMethodCallsCount = 0;
-        private long _totalDesiredPropertiesUpdatesCount = 0;
-        private long _totalC2dMessagesSentCount = 0;
+        private static volatile Dictionary<string, Tuple<Action, CancellationTokenSource>> s_onlineDeviceOperations;
```
abhipsaMisra, line 31: I'm looking at only the code and I might not have the entire context of your change, so take this with a pinch of salt - is there a reason you're saving off an Action (and thus no return value) vs a Func (that returns a task). Since there is a cancellation token associated, Iwas expecting a Task in return that would signal how the Task ended.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -124,75 +73,53 @@ public async Task MonitorConnectedDevicesAsync(CancellationToken ct)
...
+                            _logger.Trace($"Creating {nameof(DeviceOperations)} on the device [{deviceId}]", TraceSeverity.Verbose);
+
+                            try
+                            {
+                                await Task
+                                .WhenAll(
+                                    deviceOperations.InvokeDirectMethodAsync(token),
+                                    deviceOperations.SetDesiredPropertiesAsync("guidValue", Guid.NewGuid().ToString(), token),
+                                    deviceOperations.SendC2dMessagesAsync(token))
+                                .ConfigureAwait(false);
```
drwill-ms, line 109:
```suggestion
                                    .WhenAll(
                                        deviceOperations.InvokeDirectMethodAsync(token),
                                        deviceOperations.SetDesiredPropertiesAsync("guidValue", Guid.NewGuid().ToString(), token),
                                        deviceOperations.SendC2dMessagesAsync(token))
                                    .ConfigureAwait(false);
```

e2e/LongHaul/service/IotHub.cs
```diff
@@ -124,75 +73,53 @@ public async Task MonitorConnectedDevicesAsync(CancellationToken ct)
...
-                    if (s_activeDeviceSet.Contains(_deviceId))
-                    {
-                        var payload = new CustomC2dMessagePayload
-                        {
-                            RandomId = Guid.NewGuid(),
-                            CurrentTimeUtc = DateTime.UtcNow,
-                            MessagesSentCount = ++_totalC2dMessagesSentCount,
-                        };
-                        var message = new OutgoingMessage(payload)
+                        async void Operations()
```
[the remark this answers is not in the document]
brycewang-microsoft: I take it as an `Action`, however, with the comments below, I agree it'd be better to take this as `Func<Task>`.

e2e/LongHaul/service/IotHub.cs
brycewang-microsoft: Yeah, that would make more sense.

drwill-ms (approved):

PR #3224, opened 2023-03-30
Make file upload notification callback async and update doc comments

drwill-ms:

PR #3237, opened 2023-04-03
Unsubscribe from connection loss events when closing service client AMQP connection

timtay-microsoft:
There is no reason to notify users that the connection was lost when they are actively trying to close the connection.

I also made the other callbacks in these AMQP using clients work asynchronously just like #3224 did for the file upload notification processor callbacks.


drwill-ms (commented):
You'll need to update the E2E test that I added to ensure the error processor gets called. I'd suggest changing it to ensuring it does NOT get called after a call to CloseAsync, and consider adding another test to ensure it does get called, maybe if the user code throws in the FileUploadNotificationProcessor callback?

timtay-microsoft:
> You'll need to update the E2E test that I added to ensure the error processor gets called. I'd suggest changing it to ensuring it does NOT get called after a call to CloseAsync, and consider adding another test to ensure it does get called, maybe if the user code throws in the FileUploadNotificationProcessor callback?

Sure, I'll do that, too. I'm also working on a proper "reconnection" sample for these AMQP service clients.

e2e/Tests/iothub/service/FileUploadNotificationE2ETest.cs
```diff
@@ -119,48 +119,34 @@ async Task<AcknowledgementType> OnFileUploadNotificationReceived(FileUploadNotif
...
 
         [TestMethod]
-        [DataRow(IotHubTransportProtocol.Tcp)]
-        [DataRow(IotHubTransportProtocol.WebSocket)]
-        public async Task FileUploadNotification_ErrorProcessor_ReceivesNotifications(IotHubTransportProtocol protocol)
+        public async Task FileUploadNotification_CloseGracefully_DoesNotExecuteConnectionLoss()
         {
-            var options = new IotHubServiceClientOptions
+            // arrange
+            using var sender = new IotHubServiceClient(TestConfiguration.IotHub.ConnectionString);
```
drwill-ms, line 125: Use the cached one since we're not initializing with options. Do not dispose.
```suggestion
            var sender = TestDevice.ServiceClient;
```

e2e/Tests/iothub/service/FileUploadNotificationE2ETest.cs
```diff
@@ -119,48 +119,34 @@ async Task<AcknowledgementType> OnFileUploadNotificationReceived(FileUploadNotif
...
-                await errorProcessorNotified.WaitAsync(cts.Token).ConfigureAwait(false);
-                errorProcessorNotified.Task.IsCompleted.Should().BeTrue();
-            }
-            finally
-            {
-                serviceClient.FileUploadNotifications.ErrorProcessor = null;
-                await serviceClient.FileUploadNotifications.CloseAsync().ConfigureAwait(false);
+            Task<AcknowledgementType> OnFileUploadNotificationReceivedAsync(FileUploadNotification fileUploadNotification) 
+            { 
+                return Task.FromResult(AcknowledgementType.Complete);
```
drwill-ms, line 139: If you accept them here, other tests that are expecting one may not get it. Better to reject?

iothub/service/src/Messaging/MessagesClient.cs
```diff
@@ -88,11 +88,16 @@ protected MessagesClient()
         /// public void OnConnectionLost(ErrorContext errorContext)
         /// {
         ///    // Add reconnection logic as needed
-        ///    Console.WriteLine("Messaging client connection lost")
+        ///    Console.WriteLine("Messaging client connection lost");
```
drwill-ms, line 91: How about just call openasync as an example?

drwill-ms (approved):

e2e/Tests/iothub/service/FileUploadNotificationE2ETest.cs
timtay-microsoft: Good point.

iothub/service/src/Messaging/MessagesClient.cs
timtay-microsoft: Sure, that works

PR #3238, opened 2023-04-04
Fix file upload notifications in long-haul tests of service app

brycewang-microsoft:


e2e/LongHaul/service/IotHub.cs
```diff
@@ -165,32 +165,41 @@ AcknowledgementType OnC2dMessageAck(FeedbackBatch feedbackMessages)
 
         public async Task ReceiveFileUploadAsync(CancellationToken ct)
         {
-            while (!ct.IsCancellationRequested)
```
brycewang-microsoft, line 168: We are registering file upload notification callback here, so we shouldn't do this in a loop.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -165,32 +165,41 @@ AcknowledgementType OnC2dMessageAck(FeedbackBatch feedbackMessages)
...
+
+                s_blobContainerClient.DeleteBlobIfExists(fileUploadNotification.BlobName);
+                return Task.FromResult(ackType);
+            }
+
+            s_serviceClient.FileUploadNotifications.FileUploadNotificationProcessor = FileUploadNotificationCallback;
+
+            try
             {
                 await s_serviceClient.FileUploadNotifications.OpenAsync(ct).ConfigureAwait(false);
```
brycewang-microsoft, line 192: Callback should be registered before opening the connection.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -165,32 +165,41 @@ AcknowledgementType OnC2dMessageAck(FeedbackBatch feedbackMessages)
...
-                s_serviceClient.FileUploadNotifications.FileUploadNotificationProcessor = FileUploadNotificationCallback;
-                _logger.Metric("TotalFileUploadNotificiationsReceived", _totalFileUploadNotificationsReceived);
-
-                await Task.Delay(s_receiveFileUploadInterval, ct).ConfigureAwait(false);
+                catch (OperationCanceledException) { }
+            }
+            finally
+            {
+                await s_serviceClient.FileUploadNotifications.CloseAsync().ConfigureAwait(false);
             }
```
brycewang-microsoft, line 203: Similar to what we are doing in [`ReceiveMessageFeedbacksAsync()`](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/e2e/LongHaul/service/IotHub.cs#L154-L163).

e2e/LongHaul/service/IotHub.cs
drwill-ms: Correct, but we need to have something wait on the cancellation token before the method exits.

drwill-ms (approved):

PR #3239, opened 2023-04-04
Remove some extra test cases that don't test new scenarios

abhipsaMisra:


e2e/Tests/iothub/device/TelemetryMessageE2eTests.cs
```diff
@@ -66,10 +64,8 @@ public async Task Message_DeviceSendSingleMessage_MqttWs_WithProxy()
         }
 
         [DataTestMethod]
-        [DataRow(TestDeviceType.Sasl, IotHubClientTransportProtocol.Tcp)]
         [DataRow(TestDeviceType.Sasl, IotHubClientTransportProtocol.WebSocket)]
         [DataRow(TestDeviceType.X509, IotHubClientTransportProtocol.Tcp)]
-        [DataRow(TestDeviceType.X509, IotHubClientTransportProtocol.WebSocket)]
```
brycewang-microsoft, line 72: Any reason why we remove Websocket for this test only instead of TCP like anywhere else in this PR?

brycewang-microsoft (approved):

e2e/Tests/iothub/device/TelemetryMessageE2eTests.cs
abhipsaMisra: No reason, they should behave the same way. I just realized that I had been including the websocket variant in most of the tests so I decided to leave a random tcp variant in there.

In general, I have been favoring the websocket variant since that has historically been the more commonly used protocol.

PR #3241, opened 2023-04-04
Fix serialization/deserialization in long-haul tests

brycewang-microsoft:


e2e/LongHaul/device/CustomC2dMessagePayload.cs
brycewang-microsoft, line 1: This type of payload should work for both NewtonSoft and System.Text.Json, so added here.

e2e/LongHaul/device/CustomDirectMethodPayload.cs
brycewang-microsoft, line 1: This type of payload should work for both NewtonSoft and System.Text.Json, so added here.

e2e/LongHaul/service/CustomC2dMessagePayload.cs
brycewang-microsoft, line 1: We only use NewtonSoft internally for service client, so correcting the serialization lib we use for service app.

e2e/LongHaul/service/CustomDirectMethodPayload.cs
brycewang-microsoft, line 1: We only use NewtonSoft internally for service client, so correcting the serialization lib we use for service app.

iothub/device/src/Utilities/ObjectConversionHelper.cs
```diff
@@ -20,8 +20,7 @@ internal static bool TryCastOrConvert<T>(object objectToCastOrConvert, PayloadCo
             {
                 // If the cannot be cast to <T> directly we need to try to convert it using the serializer.
                 // If it can be successfully converted, go ahead and return it.
```
brycewang-microsoft, line 22: I believe this comment needs an update as well, any advice? @drwill-ms

drwill-ms: Yes

drwill-ms, line 22:
```suggestion
                // We'll serialize the object back to JSON using the user's configured payload convention
                // and then to the type of their choosing with the same payload convention.
```

iothub/device/src/Utilities/ObjectConversionHelper.cs
```diff
@@ -71,12 +70,12 @@ private static bool TryGetNumeric(object expression)
             }
 
             return double.TryParse(
```
drwill-ms, line 72:
```suggestion
            // This call requires parameter labels to avoid confusion with an overload.
            return double.TryParse(
```

drwill-ms (approved):

PR #3245, opened 2023-04-05
Cleanup more device sdk handlers

drwill-ms:


iothub/device/src/Pipeline/MqttTransportHandler.cs
drwill-ms, line 1: Moved to Pipeline folder because it feels more natural to have all the of the handlers in the same folder, even though they also have association with the other transport-specific files in a different folder.

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -265,109 +267,124 @@ internal MqttTransportHandler(PipelineContext context, IotHubClientMqttSettings
 
         public override async Task OpenAsync(CancellationToken cancellationToken)
         {
-            string clientId = string.IsNullOrWhiteSpace(_moduleId) ? _deviceId : $"{_deviceId}/{_moduleId}";
-            _mqttClientOptionsBuilder.WithClientId(clientId);
-
-            string username = $"{_hostName}/{clientId}/?{ClientApiVersionHelper.ApiVersionQueryStringLatest}&DeviceClientType={Uri.EscapeDataString(_productInfo.ToString())}";
+            if (Logging.IsEnabled)
```
drwill-ms, line 270: Not much changed here except adding try/finally for logging.

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1221,26 +1329,26 @@ private bool ParseResponseTopic(string topicName, out string rid, out int status
             // "$iothub/twin" "res/204" "$rid=efc34c73-79ce-4054-9985-0cdf40a3c794&$version=2"
             // Then the third group is parsed for key value pairs such as "$version=2"
             Match match = _twinResponseTopicRegex.Match(topicName);
-            if (match.Success)
+            if (!match.Success)
```
drwill-ms, line 1332: swapped to fail quicker, and remaining execution is success path

iothub/device/src/Pipeline/TransportDelegatingHandler.cs
```diff
@@ -27,47 +27,27 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
 
             cancellationToken.ThrowIfCancellationRequested();
 
+            await _handlerLock.WaitAsync(cancellationToken).ConfigureAwait(false);
             try
             {
-                await _handlerLock.WaitAsync(cancellationToken).ConfigureAwait(false);
```
drwill-ms, line 32: semaphore waits should happen outside the try/finally

iothub/device/src/Pipeline/TransportDelegatingHandler.cs
```diff
@@ -27,47 +27,27 @@ public override async Task OpenAsync(CancellationToken cancellationToken)
...
             }
             finally
             {
+                _handlerLock.Release();
                 if (Logging.IsEnabled)
                     Logging.Exit(this, cancellationToken, $"{nameof(TransportDelegatingHandler)}.{nameof(OpenAsync)}");
             }
         }
 
-        private void CreateNewTransportIfNotReady()
```
drwill-ms, line 58: Moved lower

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -1256,7 +1256,7 @@ public async Task MessageIdDefaultNotSet_SendEventBatchDoesNotSetMessageId()
         {
             // arrange
             string messageId = Guid.NewGuid().ToString();
-            await using var deviceClient = new IotHubDeviceClient(fakeConnectionString);
+            await using var deviceClient = new IotHubDeviceClient(fakeConnectionString, new IotHubClientOptions(new IotHubClientAmqpSettings()));
```
drwill-ms, line 1259: Avoid exception over MQTT where it is not supported.

iothub/device/src/Pipeline/TransportDelegatingHandler.cs
abhipsaMisra: If we move it outside the try then the release on finally won't happen on exception. Is that what we want?

drwill-ms: That's not how it works. If we successfully get the lock, then we enter the try. If we fail to get the lock then you don't want to execute release, so we don't put it inside the try.

abhipsaMisra (approved):

PR #3249, opened 2023-04-05
Separate out semaphores for subscription and client open-close

abhipsaMisra:


iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -137,18 +141,17 @@ await _internalRetryHandler
                         async () =>
                         {
                             await VerifyIsOpenAsync(cancellationToken).ConfigureAwait(false);
```
drwill-ms, line 143: I'll renew my dislike for the semaphore wait to evaluate _isOpen. I don't see what it buys us, and it costs each operation an extra semaphore get.

Did you have a future plan to refactor that?

drwill-ms (approved):

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: Yess, we'll get rid of the semaphores once we introduce the atomic variable for the state variable

PR #3251, opened 2023-04-06
Connection state management

abhipsaMisra:


iothub/device/src/Pipeline/ClientTransportStateMachine.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        Closed,
+        Opening,
+        Open,
+        Closing,
+    }
+
+    internal enum ClientStateAction
+    {
+        Open_Start,
```
drwill-ms, line 19: OpenStart

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
```diff
@@ -0,0 +1,90 @@
...
+                { new StateTransition(ClientTransportState.Opening, ClientStateAction.Open_Success), ClientTransportState.Open },
+                { new StateTransition(ClientTransportState.Opening, ClientStateAction.Open_Failure), ClientTransportState.Closed },
+                { new StateTransition(ClientTransportState.Opening, ClientStateAction.Close_Start), ClientTransportState.Closing },
+                { new StateTransition(ClientTransportState.Open, ClientStateAction.Close_Start), ClientTransportState.Closing },
+                { new StateTransition(ClientTransportState.Open, ClientStateAction.Connection_Lost), ClientTransportState.Closed },
+                { new StateTransition(ClientTransportState.Closing, ClientStateAction.Close_Complete), ClientTransportState.Closed },
+            };
+        }
+
+        internal ClientTransportState CurrentState { get; private set; }
```
drwill-ms, line 70: get under lock?

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
```diff
@@ -0,0 +1,90 @@
...
+        private ClientTransportState GetNextState(ClientStateAction action)
+        {
+            var transition = new StateTransition(CurrentState, action);
+
+            return _transitions.TryGetValue(transition, out ClientTransportState nextState)
+                ? nextState
+                : throw new InvalidOperationException("Invalid transition: " + CurrentState + " -> " + action);
+        }
+
+        internal ClientTransportState MoveNext(ClientStateAction action)
```
drwill-ms, line 81: void return

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            public override bool Equals(object obj)
+            {
+                return obj is StateTransition other
+                    && other._currentState == _currentState && other._nextAction == _nextAction;
+            }
+
+            public override int GetHashCode()
+            {
+                throw new NotImplementedException();
```
abhipsaMisra, line 48: TODO

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -453,15 +453,22 @@ public async ValueTask DisposeAsync()
...
 
-                await CloseAsync(CancellationToken.None).ConfigureAwait(false);
+                try
+                {
+                    await CloseAsync(CancellationToken.None).ConfigureAwait(false);
+                }
+                catch (Exception ex) when (Logging.IsEnabled)
+                {
+                    Logging.Error(this, $"Exception observed while closing the client: {ex}", nameof(DisposeAsync));
+                }
```
drwill-ms, line 463:
```suggestion
                catch (Exception ex)
                {
                    if (Logging.IsEnabled)
                        Logging.Error(this, $"Exception observed while closing the client: {ex}", nameof(DisposeAsync));
                        
                    // fail gracefully
                }
```

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -453,15 +453,22 @@ public async ValueTask DisposeAsync()
...
+
                 Dispose(true);
                 GC.SuppressFinalize(this);
                 _isDisposed = true;
             }
             catch (Exception ex) when (Logging.IsEnabled)
             {
                 Logging.Error(this, ex, nameof(DisposeAsync));
```
drwill-ms, line 471:
```suggestion
            catch (Exception ex)
            {
                if (Logging.IsEnabled)
                    Logging.Error(this, ex, nameof(DisposeAsync));
                    
                // fail silently
```

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
```diff
@@ -0,0 +1,96 @@
...
+            internal StateTransition(ClientTransportState state, ClientStateAction action)
+            {
+                _currentState = state;
+                _nextAction = action;
+            }
+
+            public override bool Equals(object obj)
+            {
+                return obj is StateTransition other
+                    && other._currentState == _currentState && other._nextAction == _nextAction;
```
drwill-ms, line 43:
```suggestion
                    && other._currentState == _currentState
                    && other._nextAction == _nextAction;
```

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            public override int GetHashCode()
+            {
+                return _currentState.GetHashCode() + _nextAction.GetHashCode();
```
drwill-ms, line 48: https://learn.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-8.0#examples

```suggestion
                return _currentState.GetHashCode() ^ _nextAction.GetHashCode();
```

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        private readonly Dictionary<StateTransition, ClientTransportState> _transitions;
+        private readonly object _stateTransitionLock = new();
+        private ClientTransportState _currentState;
+
+        internal ClientTransportStateMachine()
+        {
+            _currentState = ClientTransportState.Closed;
+            _transitions = new Dictionary<StateTransition, ClientTransportState>
```
drwill-ms, line 59: I think this could be static. It doesn't change. It is a set of facts. Right?

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        OpenStart,
+        OpenSuccess,
+        OpenFailure,
+        CloseStart,
+        CloseComplete,
+        ConnectionLost,
+    }
+
+    internal class ClientTransportStateMachine
```
drwill-ms, line 27: Obviously you aren't there yet, but you'll be adding copious code comments and some unit tests, yes?

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
```diff
@@ -0,0 +1,382 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Client.Transport
+{
+    internal sealed class ConnectionStateDelegatingHandler : DefaultDelegatingHandler
+    {
+
```
drwill-ms, line 13: extra newline here

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
```diff
@@ -0,0 +1,382 @@
...
+                    case ClientTransportState.Open:
+                        return;
+
+                    case ClientTransportState.Closing:
+                        throw new InvalidOperationException($"The client is currently closing. To reopen the client wait until {nameof(CloseAsync)} completes" +
+                            $" and then invoke {nameof(OpenAsync)}.");
+
+                    case ClientTransportState.Opening:
+                    case ClientTransportState.Closed:
+                        {
```
drwill-ms, line 51: This block is big enough, it could warrant being in a private method to improve readability of this method.

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
```diff
@@ -0,0 +1,382 @@
...
+                            if (_clientTransportStateMachine.GetCurrentState() != ClientTransportState.Open)
+                            {
+                                throw new InvalidOperationException($"The client connection must be opened before operations can begin. Call '{nameof(OpenAsync)}' and try again.");
+                            }
+                        }
+                        finally
+                        {
+                            _clientOpenSemaphore?.Release();
+                        }
```
drwill-ms, line 140: I suspect this would be a reusable snippet for other override methods too, so prime candidate for moving to a private method?

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -57,6 +47,78 @@ internal void SetRetryPolicy(IIotHubClientRetryPolicy retryPolicy)
...
+                                Logging.Enter(this, cancellationToken, nameof(OpenAsync));
+
+                            try
+                            {
+                                // Will throw on error.
+                                await base.OpenAsync(cancellationToken).ConfigureAwait(false);
+
+                                if (_recoverSubscriptions)
+                                {
+
```
drwill-ms, line 71: extra newline

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -57,6 +47,78 @@ internal void SetRetryPolicy(IIotHubClientRetryPolicy retryPolicy)
...
+                                    if (tasks.Any())
+                                    {
+                                        await Task.WhenAll(tasks).ConfigureAwait(false);
+                                    }
+
+                                    if (Logging.IsEnabled)
+                                        Logging.Info(this, "Subscriptions recovered.", nameof(OpenAsync));
+                                }
+
+                                _recoverSubscriptions = true;
```
drwill-ms, line 104: Why is this set to true, if already in an if block confirming it was already true?

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
abhipsaMisra: sure

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
abhipsaMisra: yes, that's correct

iothub/device/src/Pipeline/RetryDelegatingHandler.cs
abhipsaMisra: It is set to true after the if block. This is to signal the method that if we have ever successfully opened the connection, we will need to recover the subscriptions with any future reconnection attempt.

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
abhipsaMisra: I'm planning to use the prime number approach to evaluate the hash. That seems to be the most commonly recommended approach, and has been recommended by Jon Skeet as well: https://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-overriding-gethashcode/263416#263416.

It is easy to understand, doesn't have memory overhead and is fairly good with collisions.

The bit operations example in MS docs requires some shifting and wrapping of bits, so it'll end up requiring more lines of code that multiplication by prime numbers.

What do you think?

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
abhipsaMisra: On second thought, I think I'm going to leave this block as it is. If I split it out to its own method then we'll need to add documentation &/or checks to check for the other switch cases that we have above so that it doesn't execute in cases that it shouldn't. 
Since it is expected to be called from only a single location, it doesn't seem worth it to split it out.

iothub/device/src/Pipeline/ClientTransportStateMachine.cs
drwill-ms: Your hash set is very small. You likely don't have to worry about collisions. I'd pick a fast and cheap option. Bitwise operations are even faster than addition or multiplication.

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
drwill-ms: I think you could document on the private method that it should only be called from one place, and you get better readability out of it. But that's just my opinion and current take.

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
```diff
@@ -0,0 +1,374 @@
...
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, nameof(HandleDisconnectAsync));
+            }
+        }
+
+        // The retryAttemptsExhausted flag differentiates between calling this method while still retrying
+        // vs calling this when no more retry attempts are being made.
+        private void HandleConnectionStatusExceptions(Exception exception, bool retryAttemptsExhausted = false)
```
abhipsaMisra, line 266: "retryAttemptsExhausted" is never being passed as false right now. Should it?

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
```diff
@@ -0,0 +1,625 @@
...
+                    operationCts.Token);
+            }
+            finally
+            {
+                if (Logging.IsEnabled)
+                    Logging.Exit(this, reportedProperties, cancellationToken, nameof(UpdateReportedPropertiesAsync));
+            }
+        }
+
+        public override async Task<DateTime> RefreshSasTokenAsync(CancellationToken cancellationToken)
```
abhipsaMisra, line 361: Is this chained properly and does it get cancelled if CloseAsync() is called?

iothub/device/src/Pipeline/ConnectionStateDelegatingHandler.cs
```diff
@@ -0,0 +1,625 @@
...
+        private CancellationTokenSource _handleDisconnectCts;
+        private CancellationTokenSource _cancelPendingOperationsCts;
+
+        internal ConnectionStateDelegatingHandler(PipelineContext context, IDelegatingHandler innerHandler)
+            : base(context, innerHandler)
+        {
+            _onConnectionStatusChanged = context.ConnectionStatusChangeHandler;
+
+            // This retry policy is saved only to handle the client state when the retry policy informs that no more retry attempts are to be made
+            _retryPolicy = context.RetryPolicy;
```
abhipsaMisra, line 29: Todo - check exp backoff retry policy. Possible that we increase the wait time way quicker than "required"

iothub/device/tests/Pipeline/ConnectionStateDelegatingHandlerTests.cs
```diff
@@ -0,0 +1,336 @@
...
+                .Returns(() =>
+                {
+                    callCounter++;
+                    return Task.CompletedTask;
+                });
+            // Artificial wait to simulate the transport protocol level async operations.
+            // Without this, the transport layer will immediately transition to Closed, which will cause the "Open" flow to be executed again.
+            // The internal thread pool will determine how the parallel OpenAsync() calls in act are scheduled.
+            nextHandlerMock
+                .Setup(x => x.WaitForTransportClosedAsync())
```
abhipsaMisra, line 57: Review and remove wherever not required

iothub/device/tests/Pipeline/ClientTransportStateMachineTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseStart, ClientTransportState.Closing)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseStart, ClientTransportState.Closed)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseComplete, ClientTransportState.Opening)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseComplete, ClientTransportState.Open)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseComplete, ClientTransportState.Closing)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.ConnectionLost, ClientTransportState.Opening)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.ConnectionLost, ClientTransportState.Open)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.ConnectionLost, ClientTransportState.Closing)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.ConnectionLost, ClientTransportState.Closed)]
+        public void ClientTransportStateMachine_InvalidStateTransition(object initialState, object action, object nextState)
```
abhipsaMisra, line 126: These are all of the invalid state transitions.

1. Let me know if you can think of a better way to test this.
2. Let me know if a state transition listed here looks valid to you.

iothub/device/tests/Pipeline/ClientTransportStateMachineTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseStart, ClientTransportState.Opening)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseStart, ClientTransportState.Open)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseStart, ClientTransportState.Closing)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseStart, ClientTransportState.Closed)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseComplete, ClientTransportState.Opening)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseComplete, ClientTransportState.Open)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.CloseComplete, ClientTransportState.Closing)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.ConnectionLost, ClientTransportState.Opening)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.ConnectionLost, ClientTransportState.Open)]
+        [DataRow(ClientTransportState.Closing, ClientStateAction.ConnectionLost, ClientTransportState.Closing)]
```
tmahmood-microsoft, line 124: what would be a valid state here?

[DataRow(ClientTransportState.Closing, ClientStateAction.ConnectionLost, ClientTransportState.Closing)]

tmahmood-microsoft (approved):

iothub/device/tests/Pipeline/ClientTransportStateMachineTests.cs
patilsnr: Aside from breaking this up into three categories, `startFromOpen`, `startFromClosing` and `startFromClosed` or something like that, I can't think of another way to make this read better.

patilsnr (approved):
Looks good to me so far!

iothub/device/tests/Pipeline/ClientTransportStateMachineTests.cs
abhipsaMisra: I like that idea. I'll split these in a different PR.

iothub/device/tests/Pipeline/ConnectionStateDelegatingHandlerTests.cs
andyk-ms: Is this the same as openAsync on already opened device client? Or that would be a separate test?

andyk-ms (approved):

iothub/device/tests/Pipeline/ConnectionStateDelegatingHandlerTests.cs
abhipsaMisra: This test is for specifically testing that parallel openasync calls are synchronized in the connection state management class. I'm not sure if we have a test for calling open on an already open client, but I can add one.

PR #3254, opened 2023-04-06
Fix some ErrorCode in IotHubServiceException

brycewang-microsoft:
In the case of mapping `HttpRequestException` into `IotHubServiceException`, we manually put "RequestTimeout" as `StatusCode` and "Unknown" as `ErrorCode` as there's nothing actually returned from the service. Now we add a new ErrorCode "GenericRequestTimeout" and use that instead. 

Also addressing a few more fixes in this PR.

### DIFF
--- a/iothub/device/src/Fx/Fx.cs
+++ b/iothub/device/src/Fx/Fx.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Azure.Devices.Client
 
             if (ex is IotHubClientException hex)
             {
-                return hex.IsTransient;
+                return !hex.IsTransient;
             }
 
             // These exceptions aren't themselves fatal, but since the CLR uses them to wrap other exceptions,
--- a/iothub/service/src/Exceptions/IotHubServiceErrorCode.cs
+++ b/iothub/service/src/Exceptions/IotHubServiceErrorCode.cs
@@ -159,6 +159,13 @@ namespace Microsoft.Azure.Devices
         /// </summary>
         ConfigurationNotFound = 404301,
 
+        // Request Timeout - 408
+
+        /// <summary>
+        /// The client did not send a request within the time the server was expecting the request.
+        /// </summary>
+        GenericRequestTimeout,
+
         // Conflict - 409
 
         /// <summary>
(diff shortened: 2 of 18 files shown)


iothub/device/src/Fx/Fx.cs
```diff
@@ -27,7 +27,7 @@ internal static bool IsFatal(Exception ex)
 
             if (ex is IotHubClientException hex)
             {
-                return hex.IsTransient;
+                return !hex.IsTransient;
```
brycewang-microsoft, line 30: `IotHubClientException` whose `IsTransient` flag is "true" should **not** be marked as Fatal.

iothub/service/tests/Exceptions/IotHubServiceExceptionTests.cs
```diff
@@ -76,42 +75,6 @@ public void IotHubServiceException_Ctor_WithTransientErrorCode_Ok(IotHubServiceE
...
-                Message,
-                statusCode,
-                IotHubServiceErrorCode.Unknown);
-
-            // assert
-            exception.StatusCode.Should().Be(statusCode);
-            exception.Message.Should().Be(Message);
-            exception.IsTransient.Should().BeTrue();
-            exception.TrackingId.Should().BeNull();
-        }
```
brycewang-microsoft, line 113: We don't these 2 tests as we won't have a transient "IotHubServiceException" with "RequestTimeout" as `StatusCode` and "Unknown" as `ErrorCode`.

tmahmood-microsoft (approved):

patilsnr (approved):

iothub/service/src/Exceptions/IotHubServiceErrorCode.cs
```diff
@@ -159,6 +159,13 @@ public enum IotHubServiceErrorCode
         /// </summary>
         ConfigurationNotFound = 404301,
 
+        // Request Timeout - 408
+
+        /// <summary>
+        /// The client did not send a request within the time the server was expecting the request.
+        /// </summary>
+        GenericRequestTimeout = 408000,
```
drwill-ms, line 167: I'd prefer we just name this `RequestTimeout`. I don't find that the word "Generic" means much to me as a user.

drwill-ms (approved):

PR #3255, opened 2023-04-06
Simplify/rename error class given to IoT hub service client error processors

timtay-microsoft:
Instead of provided either an IOException or an IotHubServiceException, we now just provide a single Exception that can be either of those types or more. Also create some facade classes around ErrorContext so that the error processors of each of the AMQP clients reads more cohesively


iothub/service/src/Messaging/Models/ErrorContext.cs
```diff
@@ -9,52 +10,20 @@ namespace Microsoft.Azure.Devices
     /// The context for a given connection loss event for <see cref="MessageFeedbackProcessorClient"/>,
     /// <see cref="FileUploadNotificationProcessorClient"/>, and <see cref="MessagesClient"/>.
     /// </summary>
-    /// <remarks>
-    /// The context includes the cause of the connection loss and makes a distinction between network
-    /// level issues(no internet) and IoT Hub level issues(resource not found, throttling, internal server error, etc.).
-    /// </remarks>
     public class ErrorContext
```
drwill-ms, line 13: abstract?

I'd prefer we not have an extra class as part of our public API. Maybe we choose not to have a base class?

iothub/service/src/Messaging/MessagesClient.cs
```diff
@@ -391,17 +392,16 @@ private void OnConnectionClosed(object sender, EventArgs e)
         {
             if (((AmqpObject)sender).TerminalException is AmqpException exception)
             {
-                ErrorContext errorContext = AmqpClientHelper.GetErrorContextFromException(exception);
-                ErrorProcessor?.Invoke(errorContext);
-                Exception exceptionToLog = errorContext.IotHubServiceException;
+                IotHubServiceException mappedException = AmqpClientHelper.GetIotHubExceptionFromAmqpException(exception);
+                ErrorProcessor?.Invoke(new MessagesClientError(mappedException));
```
drwill-ms, line 396: Shouldn't this be awaited?

iothub/service/src/Messaging/MessagesClient.cs
```diff
@@ -391,17 +392,16 @@ private void OnConnectionClosed(object sender, EventArgs e)
...
+                    Logging.Error(this, $"{nameof(sender)}.{nameof(OnConnectionClosed)} threw an exception: {mappedException}", nameof(OnConnectionClosed));
             }
             else
             {
-                var defaultException = new IotHubServiceException("AMQP connection was lost", ((AmqpObject)sender).TerminalException);
-                var errorContext = new ErrorContext(defaultException);
-                ErrorProcessor?.Invoke(errorContext);
+                var defaultException = new IOException("AMQP connection was lost", ((AmqpObject)sender).TerminalException);
+                var error = new MessagesClientError(defaultException);
+                ErrorProcessor?.Invoke(error);
```
drwill-ms, line 404: Same, awaited?

iothub/service/src/Messaging/MessagesClient.cs
```diff
@@ -85,9 +86,9 @@ protected MessagesClient()
         ///
         /// //...
         ///
-        /// public void OnConnectionLost(ErrorContext errorContext)
+        /// public void OnConnectionLost(MessagingError error)
```
drwill-ms, line 89:
```suggestion
        /// public async Task OnConnectionLost(MessagingError error)
```

iothub/service/src/Messaging/Models/ErrorContext.cs
timtay-microsoft: It'll be a small DRY violation, but sure

iothub/service/src/Messaging/MessagesClient.cs
timtay-microsoft: This callback method is sync since the AMQP library demands it. We can either fire-and-forget here or we can block on this call to make it sync. Thoughts?

timtay-microsoft: It's worth noting that this callback doesn't return anything, so we don't _need_ to wait on it.

iothub/service/src/Messaging/Models/ErrorContext.cs
timtay-microsoft: I'll just have 3 concrete classes with no abstract class

iothub/service/src/Messaging/MessagesClient.cs
drwill-ms: If that processor throws? Then what?

drwill-ms (approved):

iothub/service/src/Messaging/MessagesClient.cs
timtay-microsoft: If we block and wait for this invocation, we'd need to wrap it in a try/catch so it doesn't crash the SDK. If we don't wait for it, then it doesn't matter whether their callback code throws or not.

timtay-microsoft: I'm inclined to not wait for the callback here since it risks deadlock

drwill-ms: k

iothub/service/src/FileUpload/FileUploadNotificationProcessorClient.cs
```diff
@@ -239,14 +239,7 @@ private async void OnNotificationMessageReceivedAsync(AmqpMessage amqpMessage)
...
                     {
-                        if (ex is IotHubServiceException hubEx)
-                        {
-                            await ErrorProcessor.Invoke(new ErrorContext(hubEx)).ConfigureAwait(false);
-                        }
-                        else if (ex is IOException ioEx)
-                        {
-                            await ErrorProcessor.Invoke(new ErrorContext(ioEx)).ConfigureAwait(false);
-                        }
+                        await ErrorProcessor.Invoke(new FileUploadNotificationProcessorError(ex)).ConfigureAwait(false);
```
drwill-ms, line 242: So why await here and not for C2D?

PR #3257, opened 2023-04-07
Alternate sending bulk telemetry over AMQP and parameterize payload convention

drwill-ms:


e2e/LongHaul/device/Parameters.cs
```diff
@@ -34,7 +40,7 @@ internal class Parameters
             "Transport",
             Default = TransportType.Mqtt,
             Required = false,
-            HelpText = "The transport to use for the connection.")]
+            HelpText = "The transport to use for the connection (i.e. Mqtt, Amqp).")]
```
patilsnr, line 43: very nit but perhaps `e.g.` is more gramatically correct

e2e/LongHaul/device/Parameters.cs
```diff
@@ -44,5 +50,33 @@ internal class Parameters
             Required = false,
             HelpText = "The protocol over which a transport (i.e., MQTT, AMQP) communicates.")]
         public IotHubClientTransportProtocol TransportProtocol { get; set; }
+
+        [Option(
+            'j',
+            "JsonSerializer",
+            Default = JsonSerializingLibrary.SystemTextJson,
+            Required = false,
+            HelpText = "Which JSON serialization library for the device app to use (i.e., SystemTextJson, NewtonsoftJson)")]
```
patilsnr, line 59: same here, also this `i.e.,` has a comma afterward while the previous one doesn't. I think either work but we should be consistent in theory

e2e/LongHaul/device/Parameters.cs
patilsnr: actually, if `amqp` and `mqtt` are the only options then `i.e.` works

patilsnr (approved):

e2e/LongHaul/device/Parameters.cs
drwill-ms: They are the only options.

e2e/LongHaul/device/Parameters.cs
drwill-ms: I'll fix the missing comma

e2e/LongHaul/device/IotHub.cs
```diff
@@ -135,23 +140,48 @@ public async Task SendTelemetryMessagesAsync(Logger logger, CancellationToken ct
...
+                    }
                     sw.Stop();
 
-                    _logger.Metric(TotalTelemetryMessagesSent, ++_totalTelemetryMessagesSent);
+                    _totalTelemetryMessagesSent += pendingMessages.Count;
+                    _logger.Metric(TotalTelemetryMessagesSent, _totalTelemetryMessagesSent);
                     _logger.Metric(TelemetryMessageDelaySeconds, sw.Elapsed.TotalSeconds);
 
-                    pendingMessage = null;
+                    sendSingle = !sendSingle;
```
brycewang-microsoft, line 183: why flip `sendSingle` after sending all pending messages?

brycewang-microsoft (approved):

e2e/LongHaul/device/IotHub.cs
drwill-ms: To alternate between sending in bulk or single so AMQP gets both paths tested. I can add a code comment.

brycewang-microsoft: got it. Sure, that could help!

e2e/LongHaul/device/IotHub.cs
```diff
@@ -101,9 +100,19 @@ public async Task InitializeAsync()
...
+                ? 10
+                : 1;
+
+            // We want to test both paths for AMQP so we'll use a boolean to alternate between bulk and single.
+            bool sendSingle = true;
+
+            var pendingMessages = new List<TelemetryMessage>(maxBulkMessages);
             bool loggedDisconnection = false;
             logger.LoggerContext.Add(OperationName, LoggingConstants.TelemetryMessage);
+            var sw = new Stopwatch();
```
drwill-ms, line 114: @tmahmood-microsoft here's an example of stop watch usage I'm talking about.

PR #3258, opened 2023-04-07
Long haul - elapsed time metric for operations

tmahmood-microsoft:


e2e/LongHaul/device/IotHub.cs
```diff
@@ -169,10 +169,13 @@ public async Task ReportReadOnlyPropertiesAsync(Logger logger, CancellationToken
                     {
                         { "TotalTelemetryMessagesSent", _totalTelemetryMessagesSent },
                     };
+                    var sw = Stopwatch.StartNew();
```
drwill-ms, line 172: You can create the stopwatch above, and then use Reset and Stop in the loop.

That way the object isn't created over and over again.

e2e/LongHaul/device/LoggingConstants.cs
```diff
@@ -18,6 +18,8 @@ internal static class LoggingConstants
         public const string TotalC2dMessagesCompleted = "TotalC2dMessagesCompleted";
         public const string TotalC2dMessagesRejected = "TotalC2dMessagesRejected";
         public const string C2dMessageDelaySeconds = "C2dMessageDelaySeconds";
+        public const string ReportedTwinUpdateDelaySeconds = "ReportedTwinUpdateDelaySeconds";
```
drwill-ms, line 21: We talked about using the term `OperationSeconds` right?

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -59,10 +60,13 @@ public async Task InvokeDirectMethodAsync(Logger logger, CancellationToken ct)
                 {
                     try
                     {
+                        var sw = Stopwatch.StartNew();
```
patilsnr, line 63: could the stopwatch also be created above here and reused each time we need it?

patilsnr (approved):

drwill-ms (approved):

PR #3259, opened 2023-04-07
Map MQTT connect failure to network error and mark a flaky test

drwill-ms:


e2e/Tests/iothub/device/TwinE2ETests.cs
```diff
@@ -190,29 +190,29 @@ public async Task Twin_ServiceSetsDesiredPropertyAndDeviceReceivesEvent_Amqp()
 
         // This operation behaves the same irrespective of if the client is initialized over tcp or websocket.
         [TestMethod]
-        public async Task Twin_ServiceSetsDesiredPropertyArrayAndDeviceReceivesEvent_MqttWs()
+        public async Task Twin_ServiceSetsDesiredPropertyArrayAndDeviceReceivesEvent_Mqtt()
```
drwill-ms, line 193: If it doesn't matter, I suggest we use TCP by default

abhipsaMisra: Most of our customers seem to be connecting through websockets in their apps, so this seemed like a good option for us to test against. Is there a reason we'd like to test against tcp instead?

timtay-microsoft (approved):

patilsnr (approved):

PR #3260, opened 2023-04-07
Long-haul: Fix exception handling in service app

brycewang-microsoft:


SDK v2 migration guide.md
```diff
@@ -91,12 +91,12 @@ The v2 strategy can be grouped into 3 categories.
...
     - When some runtime validated user input is invalid in some other way, for example, importing or exporting devices where there is a null value in the list.
     - `ObjectDisposedException` when the client has been disposed but an operation has been issued.
-1. Handle canceled operations.
+2. Handle canceled operations.
   - `OperationCanceledException` when a user's cancelation token was signaled and the operation was stopped prematurely.
     > All async operations accept an optional `CancellationToken` parameter.
     > It is recommended to use cancelation tokens with a timed expiration as the SDK no longer has a default operation timeout.
     > Without that, an operation could continue indefinitely, depending on retry policy.
-1. When an operation fails due to a service error or network issues prevent issuance.
+3. When an operation fails due to a service error or network issues prevent issuance.
```
brycewang-microsoft, line 99: Correcting the number

e2e/LongHaul/service/IotHub.cs
```diff
@@ -96,16 +96,19 @@ await foreach (ClientTwin device in allDevices)
 
                         async Task Operations()
                         {
-                            var deviceOperations = new DeviceOperations(s_serviceClient, deviceId, _logger.Clone());
+                            var deviceOperations = new DeviceOperations(s_serviceClient, deviceId);
                             _logger.Trace($"Creating {nameof(DeviceOperations)} on the device [{deviceId}]", TraceSeverity.Verbose);
 
+                            Logger loggerPerDevice = _logger.Clone();
+                            loggerPerDevice.LoggerContext.Add("DeviceId", deviceId);
```
brycewang-microsoft, line 103: Create new logger clone per device, add custom dimension of "DeviceId", pass its clone to each operation.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -150,18 +153,18 @@ Task<AcknowledgementType> OnC2dMessageAck(FeedbackBatch feedbackMessages)
                 return Task.FromResult(AcknowledgementType.Complete);
             }
 
-            async Task OnC2dError(ErrorContext errorContext)
+            async Task OnC2dFeedbackError(ErrorContext errorContext)
```
brycewang-microsoft, line 156: Rename as this is for handling errors in c2d **feedback**.

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -140,14 +156,32 @@ public async Task SendC2dMessagesAsync(Logger logger, CancellationToken ct)
...
+                        TraceSeverity.Warning);
+                    // retry
+                }
+                catch (InvalidOperationException ex) when (ex.Message.Contains("Must open client"))
+                {
+                    logger.Trace(
+                        $"{ex.Message}\nOpening client to send C2D messsage...",
+                        TraceSeverity.Warning);
+                    await _serviceClient.Messages.OpenAsync(ct).ConfigureAwait(false);
+                }
```
brycewang-microsoft, line 179: in case of disconnect, the `Message` subclient would be closed and throw `InvalidOperationException` after reconnect. Add handling here.

SDK v2 migration guide.md
abhipsaMisra: md files actually handle the numbering automatically for you

abhipsaMisra: https://carpentries-incubator.github.io/markdown-intro/06-lists/index.html

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -26,12 +25,10 @@ internal class DeviceOperations
 
         private static readonly TimeSpan s_retryInterval = TimeSpan.FromSeconds(1);
 
-        public DeviceOperations(IotHubServiceClient serviceClient, string deviceId, Logger logger)
+        public DeviceOperations(IotHubServiceClient serviceClient, string deviceId)
```
abhipsaMisra, line 28: Is there an advantage to passing the logger to each method instead of initializing it on the constructor itself?

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -71,17 +70,23 @@ public async Task InvokeDirectMethodAsync(Logger logger, CancellationToken ct)
...
+                        logger.Trace(
+                            $"Response status: {response.Status}, payload:\n\t{JsonConvert.SerializeObject(response.PayloadAsString)}",
+                            TraceSeverity.Information);
                         break;
                     }
-                    catch (IotHubServiceException ex) when (ex.ErrorCode == IotHubServiceErrorCode.DeviceNotOnline)
+                    catch (IotHubServiceException ex) when (ex.IsTransient is true)
                     {
-                        logger.Trace($"Caught exception invoking direct method.\n{ex}", TraceSeverity.Warning);
+                        logger.Trace(
```
abhipsaMisra, line 80: Instead of having a single Trace method with a parameter accepting the severity, we could create severity-specific Trace methods. Just an observation.

abhipsaMisra: Just for readability

e2e/LongHaul/service/DeviceOperations.cs
abhipsaMisra: Shouldn't the SDK's retry policy handle the reopening after disconnect? Why would the user need to make this call?

abhipsaMisra (approved):

e2e/LongHaul/service/DeviceOperations.cs
brycewang-microsoft: There will be custom dimension "OperationName" added in for each operation method, and will override its value if we use the same logger object.

abhipsaMisra: Maybe add this as a comment for context so that it is helpful when we look at the code later?

patilsnr (approved):

e2e/LongHaul/service/DeviceOperations.cs
drwill-ms: We wouldn't need a comment if we cloned here and added the new property then passed it.

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -71,17 +70,23 @@ public async Task InvokeDirectMethodAsync(Logger logger, CancellationToken ct)
...
                         }
 
-                        logger.Trace($"Response status: {response.Status}, payload:\n\t{JsonConvert.SerializeObject(response.PayloadAsString)}", TraceSeverity.Information);
+                        logger.Trace(
+                            $"Response status: {response.Status}, payload:\n\t{JsonConvert.SerializeObject(response.PayloadAsString)}",
+                            TraceSeverity.Information);
                         break;
                     }
-                    catch (IotHubServiceException ex) when (ex.ErrorCode == IotHubServiceErrorCode.DeviceNotOnline)
+                    catch (IotHubServiceException ex) when (ex.IsTransient is true)
```
drwill-ms, line 78:
```suggestion
                    catch (IotHubServiceException ex) when (ex.IsTransient)
```

e2e/LongHaul/service/DeviceOperations.cs
```diff
@@ -105,14 +110,25 @@ public async Task SetDesiredPropertiesAsync(string keyName, string properties, L
...
                     ++_totalDesiredPropertiesUpdatesCount;
-                    logger.Trace($"Updating the desired properties for device: {_deviceId}", TraceSeverity.Information);
+                    logger.Trace(
+                        $"Updating the desired properties for device: {_deviceId}",
+                        TraceSeverity.Information);
                     logger.Metric(TotalDesiredPropertiesUpdatesCount, _totalDesiredPropertiesUpdatesCount);
 
                     await _serviceClient.Twins.UpdateAsync(_deviceId, twin, false, ct).ConfigureAwait(false);
                 }
+                catch (IotHubServiceException ex) when (ex.IsTransient is true)
```
drwill-ms, line 120:
```suggestion
                catch (IotHubServiceException ex) when (ex.IsTransient)
```

e2e/LongHaul/service/DeviceOperations.cs
drwill-ms: If these exceptions are transient, wouldn't they be retried by the retry policy?

The case we had here was one that wasn't retriable, but we choose to retry anyway.

e2e/LongHaul/service/DeviceOperations.cs
drwill-ms: Yeah, this behavior seems wrong.

The app called open but then the client got disconnected. So we shouldn't throw InvalidOperationException, we should throw a service exception with a NetworkError error code.

drwill-ms: Which would be retryable, and automatically retried.

e2e/LongHaul/service/IotHub.cs
drwill-ms: We're using camelCase for properties, right? Also, this should be put in a logging constants file.

drwill-ms (changes requested):
comments inline

SDK v2 migration guide.md
brycewang-microsoft: Weird, I am seeing [this](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/SDK%20v2%20migration%20guide.md#exception-handling) 

image

abhipsaMisra: You're right. I wonder if that's because we have a bulleted list between the numbered list.

e2e/LongHaul/service/DeviceOperations.cs
brycewang-microsoft: Yeah, the case here is just for logging a retriable/transient exception with the "Warning" severity trace, while other exceptions would be logged as "Error" severity trace. I don't mind to remove this if we don't care about logging differnently.

e2e/LongHaul/service/DeviceOperations.cs
brycewang-microsoft: `InvalidOperationException` was thrown [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/service/src/Messaging/MessagesClient.cs#L414), so just throwing a retriable `IotHubServiceException` would not help and I also validated this. I am wondering calling open within "MessagesClient" instead of throwing exceptions.

drwill-ms: 1. We never want to force customers to match exception text to know what's up and act on it in a meaningful way programmatically.
2. If the user called open, and later we are disconnected, we must not throw InvalidOperationException. Throwing that is for helping someone get their app written right, not for dealing with disconnects. Let's fix that disconnect experience to either retry connection or throw an IotHubServiceException with a NetworkError code.

@abhipsaMisra you can probably weigh in here on how to refactor this considering what you're doing for device client.

andyk-ms (approved):

iothub/service/src/Messaging/MessagesClient.cs
```diff
@@ -407,9 +410,28 @@ private void OnConnectionClosed(object sender, EventArgs e)
...
+                {
+                    await _internalRetryHandler
+                        .RunWithRetryAsync(
+                            async () =>
+                            {
+                                await _amqpConnection.OpenAsync(cancellationToken).ConfigureAwait(false);
+                            },
+                            cancellationToken)
+                        .ConfigureAwait(false);
+                    _opened = true;
```
brycewang-microsoft, line 427: Maybe we don't need L427 as it has been `true` already.

iothub/service/src/Messaging/MessagesClient.cs
```diff
@@ -407,9 +410,28 @@ private void OnConnectionClosed(object sender, EventArgs e)
             }
         }
 
-        private void CheckConnectionIsOpen()
+        private async Task CheckConnectionIsOpenAsync(CancellationToken cancellationToken = default)
         {
-            if (!_amqpConnection.IsOpen)
+            if (_opened)
```
brycewang-microsoft, line 415: I d recommend to add a comment here and state that this is to make sure the `OpenAsync` call was made before but the client was disconnected afterwards.

iothub/service/src/Messaging/MessagesClient.cs
```diff
@@ -407,9 +410,28 @@ private void OnConnectionClosed(object sender, EventArgs e)
...
+                            {
+                                await _amqpConnection.OpenAsync(cancellationToken).ConfigureAwait(false);
+                            },
+                            cancellationToken)
+                        .ConfigureAwait(false);
+                    _opened = true;
+                }
+                catch (Exception ex)
+                {
+                    throw new IotHubServiceException(ex.Message, ex);
```
brycewang-microsoft, line 430: Specify the `StatusCode` and `ErrorCode`

brycewang-microsoft: `throw new IotHubServiceException(ex.Message, HttpStatusCode.RequestTimeout, IotHubServiceErrorCode.RequestTimeout, null, ex);`

e2e/LongHaul/service/DeviceOperations.cs
drwill-ms: The exceptions won't be surfaced because they'll be retried. The SDK doesn't let the exception get to the user code in that case.

drwill-ms (approved):

PR #3261, opened 2023-04-10
Service client samples, bug fix

timtay-microsoft:
- Add messaging client sample
- Add retry logic to file upload notification sample
- Fix bug where error processor would execute 3 times on a connection loss event
  - The same event handler was hooked up to the connection drop, session drop, and link drop event. Now the link drop event just triggers the session to close and the session drop event just triggers the connection to close. Only the connection layer can trigger the "Connection lost" event


iothub/service/src/Amqp/AmqpCbsSessionHandler.cs
```diff
@@ -20,7 +20,6 @@ internal class AmqpCbsSessionHandler : IDisposable
         private readonly IotHubConnectionProperties _credential;
 
         private AmqpCbsLink _cbsLink;
-        private readonly EventHandler _connectionLossHandler;
```
timtay-microsoft, line 23: Unused. CBS links don't allow for subscribing to link drop events for some reason.

patilsnr (approved):

iothub/service/samples/how to guides/MessagingClientSample/MessagingClientSample.csproj
```diff
@@ -0,0 +1,31 @@
...
+    can't target a main release yet.
+    -->
+    <NoWarn>NU5104</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.9.1" />
+  </ItemGroup>
+
+  <!--<ItemGroup Condition="'$(Configuration)' == 'Release'">
```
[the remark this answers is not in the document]
timtay-microsoft: I see this same block in all the other samples, so I believe David has something in mind with this commented out code.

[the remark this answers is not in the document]
drwill-ms: We can put it in after we GA, so customers will get the shipped nuget tested if they run Release.

We can choose to only use project reference tho. If we add new functionality to an SDK and use it in a sample, it'd otherwise be broken in Release until after we release, so this may not have been my brightest idea.

iothub/service/samples/how to guides/MessagingClientSample/MessagingClientSample.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>9.0</LangVersion>
```
drwill-ms, line 6: Make not packable

iothub/service/src/Amqp/AmqpSessionHandler.cs
```diff
@@ -35,25 +36,25 @@ protected internal AmqpSessionHandler()
         /// upload notifications or receiving feedback messages.
         /// </summary>
         /// <param name="linkAddress">The link address to receive from/send to.</param>
-        /// <param name="connectionLossHandler">The handler to invoke if this session or its links are dropped.</param>
+        /// <param name="sessionLossHandler">The handler to invoke if this session or its links are dropped.</param>
```
drwill-ms, line 39: Should the comment still mention dropped links?

timtay-microsoft: Yes because this handler is still eventually invoked if its links are dropped. A link drop triggers the session to close.

PR #3264, opened 2023-04-13
Miscellaneous DPS service client cleanup

timtay-microsoft:
- Move JsonSerializerSettingsInitializer and PageableHelpers to the Utilities folder
- Rename ResponseBody to ErrorResponse
- Remove unused fields in QueryBuilder


provisioning/service/src/Http/ErrorResponse.cs
```diff
@@ -3,9 +3,9 @@
...
 {
     /// <summary>
-    /// A class used as a model to deserialize response body object received from DPS.
+    /// A class used as a model to deserialize response body object received from DPS in error cases.
     /// </summary>
-    internal sealed class ResponseBody
+    internal sealed class ErrorResponse
     {
         [JsonProperty("errorCode")]
         internal int ErrorCode { get; set; }
```
brycewang-microsoft, line 11: Oh, so this property is not for general "status code" (2XX) as well, right?

brycewang-microsoft: Hmm, it's interesting that we have another property called "Code" and getting from this property.

timtay-microsoft: This class is only used to deserialize the HTTP response in cases where the HTTP status is not 200 or 204, but I suspect this field is still the same value as the HTTP status code

Issue #3263, opened 2023-04-12
[Technical Question] Possible DNS issue - File Upload limiting connetion lifetime


jmasek:
Hi,
we are using your SDK on approx 20 thousand devices all around the world. However, we are seeing that some devices (approx 1000 devices out of 20k) have problems with File Upload (they got stuck on the method `DeviceClient.GetFileUploadSasUriAsync` ). After some digging, we suspect that the DNS change of IoT Hub occurred but the underlying HttpClient keeps using the old IP address. 
I discovered that in your SDK, there are two classes `HttpClientHelper`. The one in the namespace `Microsoft.Azure.Devices` allows specifying `connectionLeaseTimeoutMilliseconds` by passing proper options to the caller but the one that is used for file uploads - in the namespace `Microsoft.Azure.Devices.Client.Transport` does not have such a parameter and is created with default `PooledConnectionLifetime` set to infinite. Therefore, I have the following questions:

- Are my findings correct or am I missing something?
- Is it possible to limit the connection lifetime to force the connection to refresh the IP Address?

Thank you,
Jan

timtay-microsoft:
Hi @jmasek, this does appear to be a bug in the SDK. We'll get a fix out soon that sets the default connection lease timeout to 5 minutes on the HTTP client used for file upload operations.

timtay-microsoft:
#3266 should fix this issue

jmasek:
Hi @timtay-microsoft, thanks for the quick fix! 
One last question, do you happen to have any estimate of when the fix could be released?

timtay-microsoft:
It may be a week or two before we put out a release. If you need this fix now, I recommend making the same change that the SDK made [here](https://github.com/Azure/azure-iot-sdk-csharp/blob/be0a5a524f6d4495132469ab65d8edee6b98a3a0/iothub/device/src/Transport/Http/HttpClientHelper.cs#L78) in your code. This change uses a static class that all HttpClients will reference, so it is safe to make this change outside of the SDK.

```csharp
ServicePoint servicePoint = ServicePointManager.FindServicePoint("<your hub name>.azure-devices.net");
servicePoint.ConnectionLeaseTimeout = <your preferred timeout>;
```

jmasek:
Hi @timtay-microsoft,
[...]

PR #3266, opened 2023-04-13
Fix file upload HTTP client not handling DNS changes

timtay-microsoft:
By setting the connection lease timeout to 5 minutes (defaults to infinite), HTTP operations will perform DNS lookups once every 5 minutes. This allows a client to continue performing file upload operations even if the service changes DNS due to failover.

We already set this value in the HTTP clients used elsewhere in this library.

#3263 

Note that this is already fixed in v2

### DIFF
--- a/iothub/device/src/Transport/Http/HttpClientHelper.cs
+++ b/iothub/device/src/Transport/Http/HttpClientHelper.cs
@@ -43,6 +43,22 @@ namespace Microsoft.Azure.Devices.Client.Transport
         private readonly ProductInfo _productInfo;
         private readonly bool _isClientPrimaryTransportHandler;
 
+        // These default values are consistent with Azure.Core default values:
+        // https://github.com/Azure/azure-sdk-for-net/blob/7e3cf643977591e9041f4c628fd4d28237398e0b/sdk/core/Azure.Core/src/Pipeline/ServicePointHelpers.cs#L28
+        private const int DefaultMaxConnectionsPerServer = 50;
+
+        // How long, in milliseconds, a given cached TCP connection created by this client's HTTP layer will live before being closed.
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -59,6 +75,9 @@ internal sealed class HttpClientHelper : IHttpClientHelper
             _authenticationHeaderProvider = authenticationHeaderProvider;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
 
+            ServicePoint servicePoint = ServicePointManager.FindServicePoint(_baseAddress);
+            servicePoint.ConnectionLeaseTimeout = s_defaultConnectionLeaseTimeout.Milliseconds;
```
abhipsaMisra, line 79: Any advantage in making this configurable?

timtay-microsoft: It isn't configurable anywhere else in the SDK yet, but we could add that feature later

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -59,6 +75,9 @@ internal sealed class HttpClientHelper : IHttpClientHelper
             _authenticationHeaderProvider = authenticationHeaderProvider;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
 
+            ServicePoint servicePoint = ServicePointManager.FindServicePoint(_baseAddress);
```
drwill-ms, line 78: Is this also a work around a user could make in their own app if they are using an SDK that doesn't do this?

PR #3279, opened 2023-04-18
Properly set connection lease timeout regardless of .NET target framework

timtay-microsoft:
Previous fix only worked for .NET framework, not .NET core

### DIFF
--- a/iothub/device/src/Transport/Http/HttpClientHelper.cs
+++ b/iothub/device/src/Transport/Http/HttpClientHelper.cs
@@ -75,9 +75,6 @@ namespace Microsoft.Azure.Devices.Client.Transport
             _authenticationHeaderProvider = authenticationHeaderProvider;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
 
-            ServicePoint servicePoint = ServicePointManager.FindServicePoint(_baseAddress);
-            servicePoint.ConnectionLeaseTimeout = s_defaultConnectionLeaseTimeout.Milliseconds;
-
 #if NET451
             TlsVersions.Instance.SetLegacyAcceptableVersions();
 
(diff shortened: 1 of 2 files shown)


iothub/device/src/Transport/Http/ServicePointHelpers.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    // This type manages changing HttpClient defaults to more appropriate values
+    // There are two limits we target:
+    // - Per Server Connection Limit
+    // - Keep Alive Connection Timeout
+    // On .NET Core 2.1+ the HttpClient defaults to using the HttpSocketHandler so we adjust both limits on the client handler
+    //
+    // On .NET Standard & NET 4.51+ the HttpClient defaults to using the HttpClientHandler
+    //   and there is no easy way to set Keep Alive Connection Timeout but it's mitigated by setting the service point's connection lease timeout
+    internal static class ServicePointHelpers
```
timtay-microsoft, line 18: Note that this class is copied from the IoT hub service client which already handles this scenario correctly

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -123,11 +118,11 @@ internal sealed class HttpClientHelper : IHttpClientHelper
...
                 _httpClientHandler.Proxy = proxy;
             }
+#endif
 
-            _httpClientHandler.MaxConnectionsPerServer = DefaultMaxConnectionsPerServer;
+            ServicePointHelpers.SetLimits(_httpClientHandler, _baseAddress);
 
-            _httpClientObj = new HttpClient(_httpClientHandler);
-#endif
+            _httpClientObj = _httpClientHandler != null ? new HttpClient(_httpClientHandler) : new HttpClient();
```
drwill-ms, line 125: _httpClientHandler can be null, but the call to SetLimits doesn't seem to handle that well.

drwill-ms: I don't think it can be null, per other code above.

drwill-ms (approved):

iothub/device/src/Transport/Http/HttpClientHelper.cs
timtay-microsoft: I'll make the service point helpers method handle null cases here just in case

abhipsaMisra (approved):

PR #3280, opened 2023-04-18
Handle connection lease timeout for all target frameworks

timtay-microsoft:
Setting the connection lease timeout via the ServicePointManager doesn't work for .NET core, so this adds the appropriate handling.

This is based heavily on the ServicePointHelper class defined in Azure.Core (with some small modifications)

### DIFF
--- a/iothub/device/src/Transport/Http/HttpClientHelper.cs
+++ b/iothub/device/src/Transport/Http/HttpClientHelper.cs
@@ -55,9 +55,24 @@ namespace Microsoft.Azure.Devices.Client.Transport
             _baseAddress = baseAddress;
             _connectionCredentials = connectionCredentials;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
-            _httpClientHandler = httpClientHandler ?? new HttpClientHandler();
-            _httpClientHandler.SslProtocols = iotHubClientHttpSettings.SslProtocols;
-            _httpClientHandler.CheckCertificateRevocationList = iotHubClientHttpSettings.CertificateRevocationCheck;
+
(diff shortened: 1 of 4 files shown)


iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -55,9 +55,24 @@ internal sealed class HttpClientHelper : IDisposable
             _baseAddress = baseAddress;
             _connectionCredentials = connectionCredentials;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
-            _httpClientHandler = httpClientHandler ?? new HttpClientHandler();
-            _httpClientHandler.SslProtocols = iotHubClientHttpSettings.SslProtocols;
-            _httpClientHandler.CheckCertificateRevocationList = iotHubClientHttpSettings.CertificateRevocationCheck;
+
+#if NETCOREAPP
```
timtay-microsoft, line 59: Even though we no longer target .NET core versions, my understanding is that this block will still execute if a user with a .NET core app takes a dependency on our SDK. Please correct me if I'm wrong on this one, though

abhipsaMisra: A preprocessor directive exists only in sourcecode = pre-compiled code. When we compile our code and build a dll, the #if-defs should already have been compiled based on the condition and the targets defined in the project.
So, if I understand this correctly, our v2 dll (which is ns2.0 only) will not have this if-def code compiled, so this will essentially be unreachable code.

You can do a local build (dll) + pack (nuget) and confirm this.

timtay-microsoft: Interesting. I'll need to rethink this strategy a bit then.

PR #3281, opened 2023-04-18
Fix bug where device operations not recovering after reconnect

brycewang-microsoft:
Using Amqp (with TCP), when device client disconnects for a duration like 10 sec and then re-connects, some operations will not recover by retries. It appears that an `OperationCanceledException` would be thrown from SendAmqpMessageAsync() call in "AmqpIotSendingLink" [class](https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/device/src/Transport/AmqpIot/AmqpIotSendingLink.cs#L125), even if cancellation has not been requested yet.

### DIFF
--- a/iothub/device/src/Transport/AmqpIot/AmqpIotSendingLink.cs
+++ b/iothub/device/src/Transport/AmqpIot/AmqpIotSendingLink.cs
@@ -130,16 +130,18 @@ namespace Microsoft.Azure.Devices.Client.Transport.AmqpIot
                         cancellationToken)
                     .ConfigureAwait(false);
             }
-            catch (OperationCanceledException oce) when (!cancellationToken.IsCancellationRequested)
-            {
-                // OperationCanceledException may be thrown here when there is networking disconnect
-                // even if cancellation has not been requested yet. This case is treated as a transient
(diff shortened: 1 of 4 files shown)


drwill-ms (approved):

patilsnr (approved):

iothub/device/src/Transport/AmqpIot/AmqpIotSendingLink.cs
```diff
@@ -130,6 +130,13 @@ private async Task<Outcome> SendAmqpMessageAsync(AmqpMessage amqpMessage, Cancel
                         cancellationToken)
                     .ConfigureAwait(false);
             }
+            catch (OperationCanceledException oce) when (!cancellationToken.IsCancellationRequested)
+            {
+                // OperationCanceledException may be thrown here when there is networking disconnect
+                // even if cancellation has not been requested yet. This case is treated as a transient
```
abhipsaMisra, line 136: Interesting. 
Fyi - Mqtt had a similar behavior that they seem to have fixed in their latest update.

tmahmood-microsoft (approved):

PR #3283, opened 2023-04-18
Create cleanup ADO pipeline, remove cleanup from nightly build

timtay-microsoft:
By running our cleanup job immediately after each test framework, it is possible that one test framework is still running when cleanup begins. Now we'll have a separate pipeline to run just cleanup and it will run nightly.


vsts/cleanup.ps1
```diff
@@ -0,0 +1,39 @@
...
+
+if (IsWindows)
+{
+	Write-Host Start ETL logging
+	logman create trace IotTrace -o iot.etl -pf tools/CaptureLogs/iot_providers.txt
+	logman start IotTrace
+}
+
+Write-Host List installed .NET SDK versions
+dotnet --list-sdks
```
abhipsaMisra, line 17: we shouldn't need this block for our samples run

vsts/cleanup.ps1
```diff
@@ -0,0 +1,39 @@
...
+
+.\build.ps1 -runCleanupSamples
```
abhipsaMisra, line 19: This is probably the only line of code required. We might be able to put that in the yaml itself.

vsts/cleanup.yaml
```diff
@@ -0,0 +1,21 @@
...
+      vmImage: windows-2022
+    steps:
+      - powershell: ./vsts/cleanup.ps1
+        displayName: build
+        env:
+          IOTHUB_CONNECTION_STRING: $(IOTHUB-CONNECTION-STRING)
+          PROVISIONING_CONNECTION_STRING: $(PROVISIONING-CONNECTION-STRING)
+          STORAGE_ACCOUNT_CONNECTION_STRING: $(STORAGE-ACCOUNT-CONNECTION-STRING)
+
+      - task: ComponentGovernanceComponentDetection@0
```
abhipsaMisra, line 16: Does every pipeline require this check? If we could remove this then it would make our maintenance cost even lower. Our cleanup samples wouldn't give us any beneficial component governance results.

abhipsaMisra (approved):
Some small comments but nothing major

vsts/cleanup.yaml
timtay-microsoft: It should be auto-injected in anyways, so I can delete this block

vsts/cleanup.yaml
```diff
@@ -0,0 +1,16 @@
+name: $(BuildID)_$(BuildDefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
+
+jobs:
+  - job: WINDOWS
+    displayName: Windows
+    pool:
+      vmImage: windows-2022
```
drwill-ms, line 7: Would be faster on Linux, no?

drwill-ms (approved):

PR #3284, opened 2023-04-18
Make device long haul client operation loops resilient to disconnection

drwill-ms:


e2e/LongHaul/device/IotHub.cs
```diff
@@ -120,16 +122,7 @@ public async Task SendTelemetryMessagesAsync(Logger logger, CancellationToken ct
                 if (!IsConnected
                     || !_messagesToSend.Any())
                 {
-                    try
```
drwill-ms, line 123: We can let it throw to the top level like we do elsewhere.

PR #3287, opened 2023-04-19
Cancel twin ops on disconnect

drwill-ms:


e2e/LongHaul/device/IotHub.cs
```diff
@@ -66,8 +67,13 @@ public IotHub(Logger logger, Parameters parameters)
         /// <summary>
         /// Initializes the connection to IoT Hub.
         /// </summary>
-        public async Task InitializeAsync()
+        public async Task InitializeAsync(CancellationToken? ct = null)
         {
+            if (ct != null)
```
drwill-ms, line 72: Share it when the program inits so we can reference it from the connection status callback.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -185,14 +191,8 @@ public async Task ReportReadOnlyPropertiesAsync(Logger logger, CancellationToken
...
                 {
-                    var reported = new ReportedProperties
-                    {
-                        { "TotalTelemetryMessagesSent", _totalTelemetryMessagesSent },
-                    };
-
-                    logger.Trace($"Updating reported properties.", TraceSeverity.Information);
                     sw.Restart();
-                    await _deviceClient.UpdateReportedPropertiesAsync(reported, ct).ConfigureAwait(false);
+                    await SetPropertiesAsync("totalTelemetryMessagesSent", _totalTelemetryMessagesSent, logger, ct).ConfigureAwait(false);
```
drwill-ms, line 195: Use the method we have for setting properties.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -244,6 +218,25 @@ public async Task ReportReadOnlyPropertiesAsync(Logger logger, CancellationToken
...
 
+        public async Task ReportReadOnlyPropertiesAsync(Logger logger, CancellationToken ct)
+        {
+            logger.LoggerContext.Add(OperationName, ReportTwinProperties);
+            var sw = new Stopwatch();
+
+            while (!ct.IsCancellationRequested)
+            {
+                sw.Restart();
+                await SetPropertiesAsync("totalTelemetryMessagesSent", _totalTelemetryMessagesSent, logger, ct).ConfigureAwait(false);
```
drwill-ms, line 229: Use the helper method we have for setting properties which will guard against exceptions.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -257,11 +250,11 @@ public async Task SetPropertiesAsync(string keyName, object properties, Logger l
             {
                 try
                 {
-                    await _deviceClient
+                    long version = await _deviceClient
```
drwill-ms, line 253: Capture and print the version.

e2e/LongHaul/device/IotHub.cs
```diff
@@ -410,7 +403,19 @@ private async void ConnectionStatusChangesHandlerAsync(ConnectionStatusInfo conn
             {
                 case RecommendedAction.OpenConnection:
                     _logger.Trace($"Following recommended action of reinitializing the client.", TraceSeverity.Information);
-                    await InitializeAsync().ConfigureAwait(false);
+                    while (!_ct.IsCancellationRequested)
+                    {
+                        try
+                        {
+                            await InitializeAsync().ConfigureAwait(false);
```
drwill-ms, line 410: Ensure the device doesn't fail on reconnect and then never retry.

e2e/LongHaul/device/Program.cs
```diff
@@ -110,6 +110,7 @@ await Task
                     .ConfigureAwait(false);
             }
             catch (OperationCanceledException) { } // user signalled an exit
+            catch (AggregateException ex) when (ex.InnerException is OperationCanceledException) { } // user signaled an exit
```
brycewang-microsoft, line 113: Any particular cases while seeing this exception?

e2e/LongHaul/service/IotHub.cs
```diff
@@ -68,125 +68,129 @@ public async Task MonitorConnectedDevicesAsync(CancellationToken ct)
...
-                AsyncPageable<ClientTwin> allDevices = s_serviceClient.Query.Create<ClientTwin>(
-                    "SELECT deviceId, connectionState FROM devices where is_defined(properties.reported.runId)",
-                    ct);
-
-                AsyncPageable<ClientTwin> allModules = s_serviceClient.Query.Create<ClientTwin>(
-                    "SELECT deviceId, moduleId, connectionState FROM devices.modules where is_defined(properties.reported.runId)",
-                    ct);
-
-                await foreach (ClientTwin device in allDevices)
+                try
```
drwill-ms, line 71: Add a try/catch as I found this failed if I killed the connection and ended the task.

iothub/device/src/Pipeline/AmqpTransportHandler.cs
```diff
@@ -22,8 +22,7 @@ internal class AmqpTransportHandler : TransportHandlerBase
         protected AmqpUnit _amqpUnit;
         private readonly Action<DesiredProperties> _onDesiredStatePatchListener;
         private readonly object _lock = new();
-        private readonly ConcurrentDictionary<string, TaskCompletionSource<AmqpMessage>> _twinResponseCompletions = new();
-        private readonly ConcurrentDictionary<string, DateTimeOffset> _twinResponseTimeouts = new();
+        private readonly ConcurrentDictionary<string, PendingAmqpTwinOperation> _pendingTwinOperations = new();
```
drwill-ms, line 25: Use just one concurrent dictionary for tracking twin operations and the associated time.

iothub/device/src/Pipeline/AmqpTransportHandler.cs
```diff
@@ -32,7 +31,7 @@ internal class AmqpTransportHandler : TransportHandlerBase
 
         internal IotHubConnectionCredentials _connectionCredentials;
 
-        private static readonly TimeSpan s_twinResponseTimeout = TimeSpan.FromMinutes(60);
+        private static readonly TimeSpan s_twinResponseTimeout = TimeSpan.FromMinutes(2);
```
drwill-ms, line 34: 60 minutes was fine for cleaning up memory, but just realized we need to cancel the operation so waiting clients can time out outside of whatever cancellation token they use.

2 minutes seems like long enough for a twin operation to have completed.

Thoughts?

iothub/device/src/Pipeline/AmqpTransportHandler.cs
```diff
@@ -492,43 +496,60 @@ private void TwinMessageListener(AmqpMessage responseFromService, string correla
...
+
+                            pendingOperation.CompletionTask.TrySetException(new IotHubClientException("Did not receive twin response from service.", IotHubClientErrorCode.NetworkErrors));
+                        }
                         return true;
-                    });
+                    })
+                .Count();
+
+            if (Logging.IsEnabled)
+                Logging.Error(this, $"Remnoved {canceledOperations} twin responses", nameof(RemoveOldOperations));
```
brycewang-microsoft, line 552:
```suggestion
                Logging.Error(this, $"Removed {canceledOperations} twin responses", nameof(RemoveOldOperations));
```

iothub/device/src/Pipeline/AmqpTransportHandler.cs
```diff
@@ -492,43 +496,60 @@ private void TwinMessageListener(AmqpMessage responseFromService, string correla
...
-                            task.TrySetException(ex);
+                            pendingOperation.CompletionTask.TrySetException(ex);
                         }
                     }
                     else
                     {
                         // This can happen if we received a message from service with correlation Id that was not set by SDK or does not exist in dictionary.
                         if (Logging.IsEnabled)
-                            Logging.Info("Could not remove correlation Id to complete the task awaiter for a twin operation.", nameof(TwinMessageListener));
+                            Logging.Info(this, "Could not remove correlation Id to complete the task awaiter for a twin operation.", nameof(TwinMessageListener));
```
drwill-ms, line 514: Fixing all this broken logging.

iothub/device/src/Pipeline/AmqpTransportHandler.cs
```diff
@@ -492,43 +496,60 @@ private void TwinMessageListener(AmqpMessage responseFromService, string correla
...
                     // This is an acknowledgement received from service for subscribing to desired property updates
                     if (Logging.IsEnabled)
-                        Logging.Info("Subscribed for twin desired property updates successfully", nameof(TwinMessageListener));
+                        Logging.Info(this, "Subscribed for twin desired property updates successfully", nameof(TwinMessageListener));
                 }
             }
         }
 
-        private void RemoveOldOperations(object _)
+        private void RemoveOldOperations(object state)
```
drwill-ms, line 526: The timer won't send a state, but when we call it directly on disconnect we can. We'll make it the duration, which defaults to the previous value (but now changed to 2 minutes).

iothub/device/src/Pipeline/AmqpTransportHandler.cs
```diff
@@ -492,43 +496,60 @@ private void TwinMessageListener(AmqpMessage responseFromService, string correla
...
+                        {
+                            if (Logging.IsEnabled)
+                                Logging.Error(this, $"Removing twin response for {x.Key}", nameof(RemoveOldOperations));
+
+                            pendingOperation.CompletionTask.TrySetException(new IotHubClientException("Did not receive twin response from service.", IotHubClientErrorCode.NetworkErrors));
+                        }
                         return true;
-                    });
+                    })
+                .Count();
```
drwill-ms, line 549: The code in select wasn't executing because the IQueryable wasn't finalized. So we'll get the count and print it, and it forces execution of the lambda.

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1181,10 +1178,10 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
             {
                 byte[] payloadBytes = receivedEventArgs.ApplicationMessage.Payload ?? Array.Empty<byte>();
 
-                if (_getTwinResponseCompletions.TryRemove(receivedRequestId, out TaskCompletionSource<GetTwinResponse> getTwinCompletion))
+                if (_pendingTwinOperations.TryRemove(receivedRequestId, out PendingMqttTwinOperation patchTwinOperation))
```
drwill-ms, line 1181: This path was semi broken, written part as "get twin" and other part as "patch twin". I updated it to be patch twin. The difference is that the version is returned.

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1355,17 +1351,35 @@ private bool ParseResponseTopic(string topicName, out string rid, out int status
...
+                            pendingOperation.TwinResponseTask?.TrySetException(new IotHubClientException(exceptionMessage, IotHubClientErrorCode.NetworkErrors));
+                            pendingOperation.TwinPatchTask?.TrySetException(new IotHubClientException(exceptionMessage, IotHubClientErrorCode.NetworkErrors));
+                        }
                         return true;
-                    });
+                    })
+                .Count();
+
+            if (Logging.IsEnabled)
+                Logging.Error(this, $"Remnoved {canceledOperations} twin responses", nameof(RemoveOldOperations));
```
brycewang-microsoft, line 1382:
```suggestion
                Logging.Error(this, $"Removed {canceledOperations} twin responses", nameof(RemoveOldOperations));
```

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1211,13 +1208,14 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
                         }
 
                         // This received message is in response to an update reported properties request.
-                        var getTwinResponse = new GetTwinResponse
+                        var patchTwinResponse = new PatchTwinResponse
```
drwill-ms, line 1211: Fixing the type and including the version.

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1228,37 +1226,35 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
                             // For this reason, we use NewtonSoft Json serializer for this deserialization.
                             TwinDocument clientTwinProperties = DefaultPayloadConvention.Instance.GetObject<TwinDocument>(payloadBytes);
 
-                            var twinDesiredProperties = new DesiredProperties(clientTwinProperties.Desired)
```
drwill-ms, line 1231: inlined these

iothub/device/src/Serialization/SystemTestJsonPayloadConvention.cs
```diff
@@ -14,20 +14,20 @@ namespace Microsoft.Azure.Devices.Client
     /// </remarks>
     public class SystemTextJsonPayloadConvention : PayloadConvention
     {
-        internal static readonly Encoding s_encoding = Encoding.UTF8;
-
         private SystemTextJsonPayloadConvention() { }
 
+        internal static Encoding Encoding => Encoding.UTF8;
```
brycewang-microsoft, line 19: internal comes after public properties?

e2e/LongHaul/device/Program.cs
drwill-ms: When more than 1 task had an exception. Made the same as service client code.

iothub/device/src/Pipeline/AmqpTransportHandler.cs
brycewang-microsoft: This looks good to me!

brycewang-microsoft (approved):

iothub/device/src/Serialization/SystemTestJsonPayloadConvention.cs
```diff
@@ -14,20 +14,20 @@ namespace Microsoft.Azure.Devices.Client
...
     {
-        internal static readonly Encoding s_encoding = Encoding.UTF8;
-
         private SystemTextJsonPayloadConvention() { }
 
         /// <summary>
         /// A static instance of this class.
         /// </summary>
-        public static readonly SystemTextJsonPayloadConvention Instance = new();
+        public static SystemTextJsonPayloadConvention Instance => new();
```
abhipsaMisra, line 22: interesting syntax

abhipsaMisra (approved):

tmahmood-microsoft (approved):

iothub/device/src/Serialization/SystemTestJsonPayloadConvention.cs
drwill-ms: This makes it a property, not a field. The readonly for properties is that there is no setter. In this case, as there is only a getter, there is a lamba implementation.

However, I realize this is wrong because this will yield a new object on every call. I'll update it to be `{ get; } = new();`

iothub/device/src/Pipeline/AmqpTransportHandler.cs
drwill-ms: Reverting to 1 hour per Tim and team agreement

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1181,10 +1176,10 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
             {
                 byte[] payloadBytes = receivedEventArgs.ApplicationMessage.Payload ?? Array.Empty<byte>();
 
-                if (_getTwinResponseCompletions.TryRemove(receivedRequestId, out TaskCompletionSource<GetTwinResponse> getTwinCompletion))
+                if (_pendingTwinOperations.TryRemove(receivedRequestId, out PendingMqttTwinOperation patchTwinOperation))
                 {
                     if (Logging.IsEnabled)
-                        Logging.Info(this, $"Received response to get twin request with request id {receivedRequestId}.");
+                        Logging.Info(this, $"Received response to patch twin request with request id {receivedRequestId}.", nameof(HandleTwinResponse));
```
timtay-microsoft, line 1182:
```suggestion
                        Logging.Info(this, $"Received response to twin request with request id {receivedRequestId}.", nameof(HandleTwinResponse));
```

We don't really know anymore if this is a response to a twin patch request or a twin get request anymore, so this can be more generic

iothub/device/src/Pipeline/TransportHandlerBase.cs
```diff
@@ -17,6 +18,8 @@ protected TransportHandlerBase(PipelineContext context, IDelegatingHandler nextH
         {
         }
 
+        protected static TimeSpan TwinResponseTimeout { get; } = TimeSpan.FromHours(1);
```
timtay-microsoft, line 21: A comment explaining what this is would be helpful for future reference

iothub/device/src/Transport/Amqp/PendingAmqpTwinOperation.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.Amqp;
+
+namespace Microsoft.Azure.Devices.Client.Transport.Amqp
```
timtay-microsoft, line 8: Remind me, did we want all classes in this library to just use the ```Microsoft.Azure.Devices.Client``` namespace?

iothub/device/src/Transport/Amqp/PendingAmqpTwinOperation.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// A class for holding the task completion source of a pending operation and the date/time when that operation
+    /// was initiated. Because PubSub behavior has us send a request on one sending link and receive a response later on another
+    /// receiving link, we need a way to identify the request, when it was sent, and the task completion source to complete the
+    /// waiting user's task.
+    /// </summary>
+    internal class PendingAmqpTwinOperation
```
timtay-microsoft, line 16: As we discussed over a call, this could be made a bit more generic since AMQP and MQTT could use this class in the same way. We don't need to change it now, but please leave a comment in here somewhere to the effect of "future work: consolidate with PendingMqttTwinOperation"

timtay-microsoft: I'd be happy to do this work, FYI

timtay-microsoft: It's my fault that the MQTT implementation didn't follow the pattern set by the AMQP implementation

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1181,120 +1176,100 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
             {
                 byte[] payloadBytes = receivedEventArgs.ApplicationMessage.Payload ?? Array.Empty<byte>();
 
-                if (_getTwinResponseCompletions.TryRemove(receivedRequestId, out TaskCompletionSource<GetTwinResponse> getTwinCompletion))
+                if (_pendingTwinOperations.TryRemove(receivedRequestId, out PendingMqttTwinOperation twinOperation))
```
timtay-microsoft, line 1179: This "if" statement is the same as the "else if" statement on line 1252

timtay-microsoft: They should probably be consolidated in some way

timtay-microsoft: Something like

```csharp
                if (_pendingTwinOperations.TryRemove(receivedRequestId, out PendingMqttTwinOperation twinOperation))
{
    if (twinOperation.TwinPatchTask != null)
    {
        // handle twin patch deserialization/error handling    
    }
    else if (twinOperation.TwinResponseTask != null)
    {
        // handle get twin response deserialization/error handling    
    }
}
```

drwill-ms: Fixed

iothub/device/src/Transport/Amqp/PendingAmqpTwinOperation.cs
drwill-ms: You can do it if you like. No hurry on it as it shouldn't change behavior.

iothub/device/src/Transport/Amqp/PendingAmqpTwinOperation.cs
drwill-ms: I think for transport specific stuff we decided it was best to isolate them by namespace.

iothub/device/src/Pipeline/MqttTransportHandler.cs
drwill-ms: Fixed

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1181,120 +1176,100 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
...
                 {
                     if (Logging.IsEnabled)
-                        Logging.Info(this, $"Received response to get twin request with request id {receivedRequestId}.");
+                        Logging.Info(this, $"Received response to patch twin request with request id {receivedRequestId}.", nameof(HandleTwinResponse));
 
-                    if (status != 200)
+                    IotHubClientErrorResponseMessage ParseError(byte[] payloadBytes)
```
timtay-microsoft, line 1184: I'm not a huge fan of functions within functions like this. Thoughts on moving it out and making it static?

timtay-microsoft: And rename to "ParseTwinResponseError" to be more specific

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1181,120 +1176,100 @@ private void HandleTwinResponse(MqttApplicationMessageReceivedEventArgs received
...
-
-                            errorResponse = new IotHubClientErrorResponseMessage
+                            try
                             {
-                                Message = errorResponseString,
-                            };
+                                // Use the encoder that has been agreed to between the client and service to decode the byte[] response
+                                // The response is deserialized into an SDK-defined type based on service-defined NewtonSoft.Json-based json property name.
+                                // For this reason, we use Newtonsoft.Json serializer for this deserialization.
+                                TwinDocument clientTwinProperties = status == 200
```
timtay-microsoft, line 1244: status is always 200 here

iothub/device/src/Pipeline/MqttTransportHandler.cs
drwill-ms: I think having a small nested function is nice so it doesn't pollute the overall class API surface as this API is only used by this method.

iothub/device/src/Pipeline/MqttTransportHandler.cs
drwill-ms: Even if throttled?

timtay-microsoft: You have a check for status != 200 above and this is in the else block

timtay-microsoft: So all non 200 status code cases wouldn't reach here

drwill-ms: true that. will fix

PR #3293, opened 2023-04-20
Allow users to pass in HttpClient to use for file upload operations

timtay-microsoft:
Since we don't target any version of .NET core in our device SDK, there is no way for us to create an HttpClient that sets the connection lease timeout for the user by default. Instead, we will allow users to pass in an HttpClient that we will use. This allows .NET core users to construct an HTTP client with their desired connection lease timeout, and give it to the SDK to use.

For .NET standard, net472, and .NET 5+ cases, the SDK can set the connection lease timeout for the user by default.

On the service SDK side, .NET standard, net472, .NET 5+ and .NET core 2.1+ cases will create an HttpClient with the connection lease timeout set by default.

### DIFF
--- a/e2e/test/iothub/FileUploadE2ETests.cs
+++ b/e2e/test/iothub/FileUploadE2ETests.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.Azure.Devices.Client;
+using Microsoft.Azure.Devices.Client.Exceptions;
 using Microsoft.Azure.Devices.Client.Transport;
 using Microsoft.Azure.Devices.E2ETests.Helpers;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
@@ -159,18 +160,11 @@ namespace Microsoft.Azure.Devices.E2ETests
             {
                 BlobName = "someBlobName",
             };
+            var ex = await Assert.ThrowsExceptionAsync<IotHubException>(
+                            async () => await deviceClient.GetFileUploadSasUriAsync(request).ConfigureAwait(false));
 
-            try
-            {
-                await deviceClient.GetFileUploadSasUriAsync(request).ConfigureAwait(false);
-            }
-            catch (Exception e)
-            {
-                // The custom HttpMessageHandler throws NotImplementedException when making any Http request.
-                // So if this exception is not thrown, then the client didn't use the custom HttpMessageHandler
-                e.InnerException.Should().BeOfType(typeof(NotImplementedException),
-                    "The provided custom HttpMessageHandler throws NotImplementedException when making any HTTP request");
-            }
+            ex.InnerException.Should().BeOfType<NotImplementedException>(
+                "The provided custom HttpMessageHandler throws NotImplementedException when making any HTTP request");
         }
 
         private async Task UploadFileGranularAsync(Stream source, string filename, Http1TransportSettings fileUploadTransportSettings, bool useX509auth = false)
@@ -315,7 +309,7 @@ namespace Microsoft.Azure.Devices.E2ETests
         {
             protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
             {
-                throw new NotImplementedException("Deliberately  not implemented for test purposes");
+                throw new NotImplementedException("Deliberately not implemented for test purposes");
             }
         }
 
--- a/e2e/test/iothub/FileUploadE2ETests.cs
+++ b/e2e/test/iothub/FileUploadE2ETests.cs
@@ -4,8 +4,11 @@
 using System;
 using System.IO;
 using System.Net;
+using System.Net.Http;
 using System.Security.Cryptography.X509Certificates;
+using System.Threading;
 using System.Threading.Tasks;
+using FluentAssertions;
 using Microsoft.Azure.Devices.Client;
 using Microsoft.Azure.Devices.Client.Transport;
 using Microsoft.Azure.Devices.E2ETests.Helpers;
@@ -125,6 +128,49 @@ namespace Microsoft.Azure.Devices.E2ETests
             };
 
             await UploadFileGranularAsync(fileStreamSource, filename, fileUploadTransportSettings).ConfigureAwait(false);
+        }
+
+        // File upload requests can be configured to use a user-provided HttpClient
+        [TestMethod]
+        public async Task FileUpload_UsesCustomHttpClient()
+        {
+            using TestDevice testDevice =
+                await TestDevice.GetTestDeviceAsync(_devicePrefix, TestDeviceType.Sasl).ConfigureAwait(false);
+
+            using var CustomHttpMessageHandler = new CustomHttpMessageHandler();
+            var fileUploadSettings = new Http1TransportSettings()
+            {
+                // This HttpClient should throw a NotImplementedException whenever it makes an HTTP
+                // request
+                HttpClient = new HttpClient(CustomHttpMessageHandler),
+            };
+
+            var clientOptions = new ClientOptions()
+            {
+                FileUploadTransportSettings = fileUploadSettings,
+            };
+
+            using var deviceClient =
+                DeviceClient.CreateFromConnectionString(testDevice.ConnectionString, clientOptions);
+
+            await deviceClient.OpenAsync().ConfigureAwait(false);
+
+            var request = new FileUploadSasUriRequest()
+            {
+                BlobName = "someBlobName",
+            };
+
+            try
+            {
+                await deviceClient.GetFileUploadSasUriAsync(request).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                // The custom HttpMessageHandler throws NotImplementedException when making any Http request.
+                // So if this exception is not thrown, then the client didn't use the custom HttpMessageHandler
+                e.InnerException.Should().BeOfType(typeof(NotImplementedException),
+                    "The provided custom HttpMessageHandler throws NotImplementedException when making any HTTP request");
+            }
         }
 
         private async Task UploadFileGranularAsync(Stream source, string filename, Http1TransportSettings fileUploadTransportSettings, bool useX509auth = false)
@@ -263,6 +309,14 @@ namespace Microsoft.Azure.Devices.E2ETests
 #endif
 
             return filePath;
+        }
+
+        private class CustomHttpMessageHandler : HttpMessageHandler
+        {
+            protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+            {
+                throw new NotImplementedException("Deliberately  not implemented for test purposes");
+            }
         }
 
         [ClassCleanup]
--- a/iothub/device/src/ClientSettings/Http1TransportSettings.cs
+++ b/iothub/device/src/ClientSettings/Http1TransportSettings.cs
@@ -49,6 +49,9 @@ namespace Microsoft.Azure.Devices.Client
         /// <inheritdoc/>
         public IWebProxy Proxy { get; set; }
 
+        /// <summary>
+        /// The HTTP client to use for all HTTP operations.
+        /// </summary>
         /// <remarks>
         /// If not provided, an HTTP client will be created for you based on the other settings provided.
         /// <para>
--- a/iothub/device/src/ClientSettings/Http1TransportSettings.cs
+++ b/iothub/device/src/ClientSettings/Http1TransportSettings.cs
@@ -55,6 +55,9 @@ namespace Microsoft.Azure.Devices.Client
         /// <remarks>
         /// If not provided, an HTTP client will be created for you based on the other settings provided.
         /// <para>
+        /// This HTTP client instance will be disposed when the device/module client using it is disposed.
+        /// </para>
+        /// <para>
         /// If provided, all other HTTP-specific settings (such as proxy, SSL protocols, and certificate revocation check)
         /// on this class will be ignored and must be specified on this HttpClient instance.
         /// </para>
--- a/iothub/device/src/ClientSettings/Http1TransportSettings.cs
+++ b/iothub/device/src/ClientSettings/Http1TransportSettings.cs
@@ -49,10 +49,13 @@ namespace Microsoft.Azure.Devices.Client
         /// <inheritdoc/>
         public IWebProxy Proxy { get; set; }
 
-        /// <summary>
-        /// The handler that this client will instantiate its HttpClient with. If this value
-        /// is provided, all other options will be ignored in favor of this handler.
-        /// </summary>
-        public HttpMessageHandler HttpMessageHandler { get; set; }
+        /// <remarks>
+        /// If not provided, an HTTP client will be created for you based on the other settings provided.
+        /// <para>
+        /// If provided, all other HTTP-specific settings (such as proxy, SSL protocols, and certificate revocation check)
+        /// on this class will be ignored and must be specified on this HttpClient instance.
+        /// </para>
+        /// </remarks>
+        public HttpClient HttpClient { get; set; }
     }
 }
--- a/iothub/device/src/ClientSettings/Http1TransportSettings.cs
+++ b/iothub/device/src/ClientSettings/Http1TransportSettings.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Net;
+using System.Net.Http;
 using System.Security.Cryptography.X509Certificates;
 using Microsoft.Azure.Devices.Shared;
 
@@ -47,5 +48,11 @@ namespace Microsoft.Azure.Devices.Client
 
         /// <inheritdoc/>
         public IWebProxy Proxy { get; set; }
+
+        /// <summary>
+        /// The handler that this client will instantiate its HttpClient with. If this value
+        /// is provided, all other options will be ignored in favor of this handler.
+        /// </summary>
+        public HttpMessageHandler HttpMessageHandler { get; set; }
     }
 }
--- a/iothub/device/src/ModuleClient.cs
+++ b/iothub/device/src/ModuleClient.cs
@@ -829,7 +829,7 @@ namespace Microsoft.Azure.Devices.Client
             }
             methodRequest.ThrowIfNull(nameof(methodRequest));
 
-            HttpClientHandler httpClientHandler = null;
+            HttpMessageHandler httpMessageHandler = null;
             Func<object, X509Certificate, X509Chain, SslPolicyErrors, bool> customCertificateValidation = _certValidator.GetCustomCertificateValidation();
 
             try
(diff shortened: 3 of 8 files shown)


iothub/device/src/ModuleClient.cs
```diff
@@ -834,6 +834,21 @@ private async Task<MethodResponse> InvokeMethodAsync(Uri uri, MethodRequest meth
 
             try
             {
+                var context = new PipelineContext()
```
drwill-ms, line 837:
```suggestion
                var context = new PipelineContext
```

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -38,91 +38,104 @@ internal sealed class HttpClientHelper : IHttpClientHelper
...
+                {
+                    _usingX509ClientCert = false;
+                }
 
-            if (clientCert != null)
-            {
-                _httpClientHandler.ClientCertificates.Add(clientCert);
-                _usingX509ClientCert = true;
-            }
+                if (proxy != DefaultWebProxySettings.Instance)
```
drwill-ms, line 126: Same as line 104?

drwill-ms: Oh, that is sockets handler.

iothub/device/src/Transport/Http/ServicePointHelpers.cs
```diff
@@ -41,8 +41,11 @@ public static void SetLimits(HttpMessageHandler messageHandler, Uri baseUri, int
                     ServicePoint servicePoint = ServicePointManager.FindServicePoint(baseUri);
                     servicePoint.ConnectionLeaseTimeout = connectionLeaseTimeoutMilliseconds;
                     break;
-#if NETCOREAPP2_1_OR_GREATER || NET5_0_OR_GREATER
-                // SocketsHttpHandler is only available in netcore2.1 and onwards
+#if NET5_0_OR_GREATER
```
drwill-ms, line 44: So this includes .net core 2.1?

timtay-microsoft: It does not since we do not target .net core 2.1

iothub/device/src/ClientSettings/Http1TransportSettings.cs
```diff
@@ -47,5 +48,11 @@ public TransportType GetTransportType()
 
         /// <inheritdoc/>
         public IWebProxy Proxy { get; set; }
+
+        /// <summary>
+        /// The handler that this client will instantiate its HttpClient with. If this value
+        /// is provided, all other options will be ignored in favor of this handler.
+        /// </summary>
+        public HttpMessageHandler HttpMessageHandler { get; set; }
```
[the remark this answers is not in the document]
timtay-microsoft: I don't follow what you mean here. I have changed this to be an HttpClient instead, though.

iothub/device/src/Transport/Http/ServicePointHelpers.cs
drwill-ms: I think there's a case to be made for the preprocessing directive to be true to the code and free of what our actual targets are. If the targets were to change, and I admit it is unlikely, then one doesn't need to recheck all the preprocessing directives for assumptive logic.

iothub/device/src/Transport/Http/HttpClientHelper.cs
timtay-microsoft: Yes, but I can't factor this if block out. The ```UseProxy``` and ```Proxy``` field are present in the implementation classes (HttpClientHandler, SocketsHttpHandler, WebRequestHandler) but are not present in the HttpMessageHandler interface.

iothub/device/src/Transport/Http/ServicePointHelpers.cs
timtay-microsoft: Fair point. I can add .NET core 2.1 back in here.

iothub/device/src/ClientSettings/Http1TransportSettings.cs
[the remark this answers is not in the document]
drwill-ms: I mean, for the other method that uses this but only for some .net targets, will this setting always work for a user regardless of which .NET version they target?

iothub/device/src/ModuleClient.cs
```diff
@@ -845,30 +860,20 @@ private async Task<MethodResponse> InvokeMethodAsync(Uri uri, MethodRequest meth
...
-                        Extra = InternalClient.ProductInfo
-                    }
-                };
 
-                var transportSettings = new Http1TransportSettings();
-                //We need to add the certificate to the httpTransport if DeviceAuthenticationWithX509Certificate
-                if (InternalClient.Certificate != null)
-                {
-                    transportSettings.ClientCertificate = InternalClient.Certificate;
+                    transportSettings.HttpClient = new HttpClient(httpMessageHandler);
```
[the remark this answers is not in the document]
drwill-ms: And shouldn't this call to new be conditional on the user not having provided one?

iothub/device/src/ModuleClient.cs
```diff
@@ -834,6 +834,21 @@ private async Task<MethodResponse> InvokeMethodAsync(Uri uri, MethodRequest meth
...
+                var context = new PipelineContext
+                {
+                    ProductInfo = new ProductInfo
+                    {
+                        Extra = InternalClient.ProductInfo
+                    }
+                };
+
+                var transportSettings = new Http1TransportSettings();
+                //We need to add the certificate to the httpTransport if DeviceAuthenticationWithX509Certificate
```
drwill-ms, line 846:
```suggestion
                // We need to add the certificate to the httpTransport if DeviceAuthenticationWithX509Certificate
```

iothub/device/src/ClientSettings/Http1TransportSettings.cs
```diff
@@ -47,5 +48,14 @@ public TransportType GetTransportType()
...
         public IWebProxy Proxy { get; set; }
+
+        /// <remarks>
+        /// If not provided, an HTTP client will be created for you based on the other settings provided.
+        /// <para>
+        /// If provided, all other HTTP-specific settings (such as proxy, SSL protocols, and certificate revocation check)
+        /// on this class will be ignored and must be specified on this HttpClient instance.
+        /// </para>
+        /// </remarks>
+        public HttpClient HttpClient { get; set; }
```
abhipsaMisra, line 59: Seems to be missing a summary.

Can this HttpClient be used for SendEventAsync() as well?

iothub/device/src/ModuleClient.cs
```diff
@@ -834,6 +834,21 @@ private async Task<MethodResponse> InvokeMethodAsync(Uri uri, MethodRequest meth
...
             {
+                var context = new PipelineContext
+                {
+                    ProductInfo = new ProductInfo
+                    {
+                        Extra = InternalClient.ProductInfo
+                    }
+                };
+
+                var transportSettings = new Http1TransportSettings();
```
drwill-ms, line 845: Why does this unconditionally create a `Http1TransportSettings` without checking if there is one available from options?

iothub/device/src/ModuleClient.cs
```diff
@@ -845,30 +860,20 @@ private async Task<MethodResponse> InvokeMethodAsync(Uri uri, MethodRequest meth
                         SslProtocols = TlsVersions.Instance.Preferred,
                     };
 #else
-                    httpClientHandler = new WebRequestHandler();
+#pragma warning disable CA2000 // Dispose objects before losing scope
+                    // This object is given to an HTTP client to use, so it can't be disposed yet
+                    var httpMessageHandler = new WebRequestHandler();
```
abhipsaMisra, line 865: Is this being disposed when the module client is disposed?

iothub/device/src/ClientSettings/Http1TransportSettings.cs
```diff
@@ -47,5 +48,14 @@ public TransportType GetTransportType()
...
         /// <inheritdoc/>
         public IWebProxy Proxy { get; set; }
+
+        /// <remarks>
+        /// If not provided, an HTTP client will be created for you based on the other settings provided.
+        /// <para>
+        /// If provided, all other HTTP-specific settings (such as proxy, SSL protocols, and certificate revocation check)
+        /// on this class will be ignored and must be specified on this HttpClient instance.
+        /// </para>
+        /// </remarks>
```
drwill-ms, line 58: Note that we'll dispose it?

iothub/device/src/ModuleClient.cs
abhipsaMisra, line 872: Isn't this going to use the client passed in through transport settings?

iothub/device/src/ClientSettings/Http1TransportSettings.cs
timtay-microsoft: It can, and I'm adding an e2e test for that as well

iothub/device/src/ModuleClient.cs
timtay-microsoft: Good catch

timtay-microsoft: Actually, this is a bit of an odd case. We already don't check if the user passed in an HttpTransportSettings class here because modules only use MQTT/AMQP except for this operation. We could create a ModuleDirectMethodTransportSettings class like we do with FileUploadTransportSettings, but that strikes me as more work than necessary for v1 since it will soon be replaced. 

I'm inclined to just not use the custom HTTP client for this particular case to keep v1 simple. Thoughts?

iothub/device/src/ModuleClient.cs
timtay-microsoft: Module clients can only use AMQP/MQTT for most operations. This one operation uses HTTP, though. This is kind of similar to how we have a separate FileUploadTransportSettings so that AMQP device clients can still configure file upload HTTP operations. Let's discuss this a bit more in this thread with Abhipsa:

https://github.com/Azure/azure-iot-sdk-csharp/pull/3293#discussion_r1174091015

iothub/device/src/ClientSettings/Http1TransportSettings.cs
timtay-microsoft: I have added the summary

iothub/device/src/ModuleClient.cs
abhipsaMisra: So, this case essentially doesn't even expect a transport setting since the client is initialized over either MQTT or AMQP. Since we already have the precedent set for not looking at transport settings for this case, I agree with your approach.

On a related note - I thought we did allow module clients over HTTP, but I guess if we do it's only IoT device modules and not IoT Edge modules (which uses this method). Is that right?

iothub/device/src/ModuleClient.cs
timtay-microsoft: It is now. Good catch

iothub/device/src/ClientSettings/Http1TransportSettings.cs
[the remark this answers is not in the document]
timtay-microsoft: Yes, this HttpClient will be used regardless of target framework. We won't do anything to this client's settings or handler. We just make HTTP requests with it.

iothub/device/src/ModuleClient.cs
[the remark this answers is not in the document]
timtay-microsoft: Resolving this thread, but leaving the other thread open to discuss whether we should be checking for a provided HttpClient here

abhipsaMisra: The v2 module client is also set up similarly to ignore http settings. We didn't have a comment for that, so I added it in: https://github.com/Azure/azure-iot-sdk-csharp/pull/3297

We should update the FileUploadTransportSettings to be more generic on our v2 client and have both device and module clients reference it for their HTTP operations.

abhipsaMisra: PS: I don't plan to merge #3297 in. I created that to highlight the behavior as you work on your changes.

timtay-microsoft: I tried removing that Java SDK HTTP check and running a telemetry test, but the hub threw a 404 error, so it looks like IoT hub doesn't support HTTP module operations

e2e/test/iothub/FileUploadE2ETests.cs
```diff
@@ -127,6 +130,49 @@ public async Task FileUpload_SmallFile_Http_GranularSteps_Proxy()
...
+            {
+                await deviceClient.GetFileUploadSasUriAsync(request).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                // The custom HttpMessageHandler throws NotImplementedException when making any Http request.
+                // So if this exception is not thrown, then the client didn't use the custom HttpMessageHandler
+                e.InnerException.Should().BeOfType(typeof(NotImplementedException),
+                    "The provided custom HttpMessageHandler throws NotImplementedException when making any HTTP request");
+            }
```
drwill-ms, line 173:
```suggestion
            Func<Task<string> act = (param) => await deviceClient.GetFileUploadSasUriAsync(request).ConfigureAwait(false);
            await act.Should()ThrowAsync...
```

iothub/device/src/ModuleClient.cs
drwill-ms: IOW connection lease timeout is not a concert on edge module

iothub/device/src/ModuleClient.cs
```diff
@@ -829,54 +829,56 @@ private async Task<MethodResponse> InvokeMethodAsync(Uri uri, MethodRequest meth
...
-                        Extra = InternalClient.ProductInfo
-                    }
-                };
-
-                var transportSettings = new Http1TransportSettings();
-                //We need to add the certificate to the httpTransport if DeviceAuthenticationWithX509Certificate
-                if (InternalClient.Certificate != null)
-                {
-                    transportSettings.ClientCertificate = InternalClient.Certificate;
+                    transportSettings.HttpClient = new HttpClient(httpMessageHandler, true);
```
drwill-ms, line 869: Disposed on 881
```suggestion
                    transportSettings.HttpClient = new HttpClient(httpMessageHandler);
```

iothub/device/src/ModuleClient.cs
```diff
@@ -829,54 +829,56 @@ private async Task<MethodResponse> InvokeMethodAsync(Uri uri, MethodRequest meth
...
+                using var httpTransport = new HttpTransportHandler(context, InternalClient.IotHubConnectionString, transportSettings);
                 var methodInvokeRequest = new MethodInvokeRequest(methodRequest.Name, methodRequest.DataAsJson, methodRequest.ResponseTimeout, methodRequest.ConnectionTimeout);
                 MethodInvokeResponse result = await httpTransport.InvokeMethodAsync(methodInvokeRequest, uri, cancellationToken).ConfigureAwait(false);
 
                 return new MethodResponse(Encoding.UTF8.GetBytes(result.GetPayloadAsJson()), result.Status);
             }
             finally
             {
-                httpClientHandler?.Dispose();
+                httpMessageHandler?.Dispose();
```
drwill-ms, line 881:
```suggestion
                transportSettings.HttpClient?.Dispose();
```

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -38,101 +38,113 @@ internal sealed class HttpClientHelper : IHttpClientHelper
...
+
+            if (proxy != DefaultWebProxySettings.Instance)
+            {
+                socketsHandler.UseProxy = proxy != null;
+                socketsHandler.Proxy = proxy;
+            }
 
-            if (clientCert != null)
+            _httpMessageHandler = socketsHandler;
+#else
```
drwill-ms, line 115:
```suggestion
#else // not 451 nor > 5.0
```

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -38,101 +38,113 @@ internal sealed class HttpClientHelper : IHttpClientHelper
...
             }
+
+            _httpMessageHandler = httpClientHandler;
 #endif
 
-            ServicePointHelpers.SetLimits(_httpClientHandler, _baseAddress);
+            ServicePointHelpers.SetLimits(_httpMessageHandler, _baseAddress);
 
-            _httpClientObj = _httpClientHandler != null ? new HttpClient(_httpClientHandler) : new HttpClient();
+            _httpClientObj = new HttpClient(_httpMessageHandler);
```
drwill-ms, line 141:
```suggestion
            _httpClientObj = new HttpClient(_httpMessageHandler, true);
```

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -558,10 +570,10 @@ public void Dispose()
 
                 // HttpClientHandler that is used to create HttpClient will automatically be disposed when HttpClient is disposed
                 // But in case the client handler didn't end up being used by the HttpClient, we explicitly dispose it here.
-                if (_httpClientHandler != null)
+                if (_httpMessageHandler != null)
                 {
-                    _httpClientHandler?.Dispose();
-                    _httpClientHandler = null;
+                    _httpMessageHandler?.Dispose();
+                    _httpMessageHandler = null;
```
drwill-ms, line 576: Remove?

drwill-ms (approved):

iothub/device/src/ModuleClient.cs
timtay-microsoft: I've made it so this client only creates the one HTTP client and disposes it when the module client is disposed

iothub/device/src/ModuleClient.cs
timtay-microsoft: I left a comment to that effect for future reference

iothub/device/src/Transport/Http/HttpClientHelper.cs
```diff
@@ -38,101 +38,113 @@ internal sealed class HttpClientHelper : IHttpClientHelper
...
             _baseAddress = baseAddress;
             _authenticationHeaderProvider = authenticationHeaderProvider;
             _defaultErrorMapping = new ReadOnlyDictionary<HttpStatusCode, Func<HttpResponseMessage, Task<Exception>>>(defaultErrorMapping);
+            _productInfo = productInfo;
+            _isClientPrimaryTransportHandler = isClientPrimaryTransportHandler;
+
+            if (transportSettings.HttpClient != null)
+            {
+                _httpClientObj = transportSettings.HttpClient;
+                return;
```
abhipsaMisra, line 65: I thought _httpMessageHandler  initializing was missing but turns out that it isn't required to be set for this case since it is used only as a local variable when creating an HttpClient. We should remove it from class variables and make it a local variable.

timtay-microsoft: Good point

PR #3294, opened 2023-04-20
Added payload convention tests and fix the conversion bug

tmahmood-microsoft:


iothub/device/src/Utilities/ObjectConversionHelper.cs
```diff
@@ -20,7 +20,7 @@ internal static bool TryCastOrConvert<T>(object objectToCastOrConvert, PayloadCo
             {
                 // We'll serialize the object back to JSON using the user's configured payload convention
                 // and then to the type of their choosing with the same payload convention.
-                value = payloadConvention.GetObject<T>(payloadConvention.GetObjectBytes(objectToCastOrConvert));
+                value = payloadConvention.GetObject<T>(DefaultPayloadConvention.Instance.GetObjectBytes(objectToCastOrConvert));
```
abhipsaMisra, line 23: The comment states that both of these operations are expected to use the user's payload convention.
Your change is updating one of the conversions to always use Newtonsoft.

Can you elaborate on this change?

drwill-ms: The comment needs updating.

The twin was deserialized with the default payload converter. So using the user's (when it is STJ) will fail to create object bytes that are compatible. Presumably the object is a JObject or something, right?

So we need to use the same payload convention used to deserialize the twin to serialize it back, and then use the user's one to deserialize.

iothub/device/src/Utilities/ObjectConversionHelper.cs
drwill-ms, line 1: It occurs to me this if this is used in any code path where the SDK didn't deserialize using the default payload convention, we may have created another bug by hard coding the Newtonsoft here.

Can you validate other operations as well with E2E tests? Direct method, Incoming message.

abhipsaMisra: That's my thought as well. If the intention is to do something that is specific to how twins are treated in our SDK then I'd recommend putting that code in the twin class and not in a "generic" helper. That way someone is less likely to reference it unintentionally in a use-case that it isn't intended for.

abhipsaMisra: If the intention is to broadly use the default convention for the first step of the conversion then I'd recommend renaming this method appropriately - TryCastOrConvertFromJObject<T>

drwill-ms: There's no intention. There's the reality we're now perceiving and realizing that the existing code wasn't working for STJ.

tmahmood-microsoft: Added more e2e tests for direct method and incoming message

iothub/device/src/Utilities/ObjectConversionHelper.cs
tmahmood-microsoft: Updated the comment

iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -121,7 +121,7 @@ public bool TryGetPayload<T>(out T payload)
 
             try
             {
-                payload = PayloadConvention.GetObject<T>(Payload);
+                payload = DefaultPayloadConvention.Instance.GetObject<T>(Payload);
```
drwill-ms, line 124: This should be done with the user's selected payload convention. Presumably you have an e2e test with STJ selected using a custom class with only STJ attributes, right? It should fail this scenario.

iothub/device/src/Messaging/IncomingMessage.cs
```diff
@@ -168,7 +168,7 @@ public bool TryGetPayload<T>(out T payload)
 
             try
             {
-                payload = PayloadConvention.GetObject<T>(_payload);
+                payload = DefaultPayloadConvention.Instance.GetObject<T>(_payload);
```
drwill-ms, line 171: Same here

e2e/Tests/iothub/device/DeviceCustomProperty.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.Azure.Devices.E2ETests.Twins
+{
+    internal class DeviceCustomProperty
+    {
+        [JsonPropertyName("customProperty")]
+        public string CustomProperty { get; set; }
+
+        [JsonPropertyName("guid")]
+        public string Guid { get; set; }
```
drwill-ms, line 14:
```suggestion
        [JsonPropertyName("string")]
        public string StringProperty { get; set; }

        [JsonPropertyName("guid")]
        public string GuidProperty { get; set; }
```

e2e/Tests/iothub/device/PayloadConventionE2eTests.cs
```diff
@@ -0,0 +1,303 @@
...
+            using var cts = new CancellationTokenSource(s_testTimeout);
+            CancellationToken ct = cts.Token;
+            PayloadConvention convention = SystemTextJsonPayloadConvention.Instance;
+            var options = new IotHubClientOptions(new IotHubClientAmqpSettings())
+            {
+                PayloadConvention = convention
+            };
+            await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(s_devicePrefix).ConfigureAwait(false);
+            await using var deviceClient = new IotHubDeviceClient(testDevice.ConnectionString, options);
+            await deviceClient.OpenAsync().ConfigureAwait(false);
```
drwill-ms, line 37: consider using `testDevice.OpenWithRetryAsync`

e2e/Tests/iothub/device/PayloadConventionE2eTests.cs
```diff
@@ -0,0 +1,303 @@
...
+                    msg.TryGetPayload(out StjCustomPayload actual).Should().BeTrue();
+                    actual.StringProperty.Should().Be(requestPayload.StringProperty);
+                    actual.GuidProperty.Should().Be(requestPayload.GuidProperty);
+                    messageReceived.TrySetResult(true);
+                    var response = new DirectMethodResponse(200)
+                    {
+                        Payload = responsePayload,
+                    };
+                    return Task.FromResult(response);
+                },
```
drwill-ms, line 144:
```suggestion
                    {
                        msg.TryGetPayload(out StjCustomPayload actual).Should().BeTrue();
                        actual.StringProperty.Should().Be(requestPayload.StringProperty);
                        actual.GuidProperty.Should().Be(requestPayload.GuidProperty);
                        messageReceived.TrySetResult(true);
                        var response = new DirectMethodResponse(200)
                        {
                            Payload = responsePayload,
                        };
                        return Task.FromResult(response);
                    },
```

e2e/Tests/iothub/device/PayloadConventionE2eTests.cs
```diff
@@ -0,0 +1,303 @@
...
+                    msg.TryGetPayload(out NjCustomPayload actual).Should().BeTrue();
+                    actual.StringProperty.Should().Be(payload.StringProperty);
+                    actual.GuidProperty.Should().Be(payload.GuidProperty);
+                    messageReceived.TrySetResult(true);
+                    var response = new DirectMethodResponse(200)
+                    {
+                        Payload = payload,
+                    };
+                    return Task.FromResult(response);
+                },
```
drwill-ms, line 199: tab in

e2e/Tests/iothub/device/PayloadConventionE2eTests.cs
```diff
@@ -0,0 +1,303 @@
...
+            var messageReceived = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+            await deviceClient
+                .SetIncomingMessageCallbackAsync((msg) =>
+                {
+                    msg.TryGetPayload(out NjCustomPayload actual).Should().BeTrue();
+                    actual.StringProperty.Should().Be(payload.StringProperty);
+                    actual.GuidProperty.Should().Be(payload.GuidProperty);
+                    messageReceived.TrySetResult(true);
+                    return Task.FromResult(MessageAcknowledgement.Complete);
+                },
```
drwill-ms, line 293: tab in

drwill-ms (approved):

PR #3295, opened 2023-04-21
Add long-haul test with Amqp pooling

brycewang-microsoft:
An Amqp pool with size of 3 for 3 devices totally.


azureiot.sln
```diff
@@ -216,6 +216,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MessagingClientSample", "io
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Azure.Devices.LongHaul.Module", "e2e\LongHaul\module\Microsoft.Azure.Devices.LongHaul.Module.csproj", "{6C880AEF-8BD9-42D8-8173-A52C328C304F}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Azure.Devices.LongHual.AmqpPooling", "e2e\LongHaul\amqpPooling\Microsoft.Azure.Devices.LongHual.AmqpPooling.csproj", "{3A60A151-0CE5-4CB9-A057-FE76E64308A3}"
```
drwill-ms, line 219:
```suggestion
Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Azure.Devices.LongHaul.AmqpPooling", "e2e\LongHaul\amqpPooling\Microsoft.Azure.Devices.LongHual.AmqpPooling.csproj", "{3A60A151-0CE5-4CB9-A057-FE76E64308A3}"
```

e2e/LongHaul/amqpPooling/IotHub.cs
```diff
@@ -0,0 +1,192 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Mash.Logging;
+using Microsoft.Azure.Devices.Client;
+using Newtonsoft.Json;
+using static Microsoft.Azure.Devices.LongHual.AmqpPooling.LoggingConstants;
+
+namespace Microsoft.Azure.Devices.LongHual.AmqpPooling
```
drwill-ms, line 15:
```suggestion
namespace Microsoft.Azure.Devices.LongHaul.AmqpPooling
```

e2e/LongHaul/amqpPooling/IotHub.cs
```diff
@@ -0,0 +1,192 @@
...
+                }
+
+                if (s_deviceClients != null)
+                {
+                    foreach (KeyValuePair<string, IotHubDeviceClient> entry in s_deviceClients)
+                    {
+                        string deviceId = entry.Key;
+                        IotHubDeviceClient deviceClient = entry.Value;
+
+                        var telemetryObject = new DeviceTelemetry()
```
drwill-ms, line 94:
```suggestion
                        var telemetryObject = new DeviceTelemetry
```

e2e/LongHaul/amqpPooling/IotHub.cs
```diff
@@ -0,0 +1,192 @@
...
+
+                _messagesToSend.TryDequeue(out SystemHealthTelemetry systemHealth);
+                if (systemHealth == null)
+                {
+                    continue;
+                }
+
+                if (s_deviceClients != null)
+                {
+                    foreach (KeyValuePair<string, IotHubDeviceClient> entry in s_deviceClients)
```
drwill-ms, line 89: I'd rather see independent tasks running for each device, sending telemetry as each one, rather than a loop for each device. This way the telemetry sends can happen at the same time.

drwill-ms: Consider the pattern we use in the service long haul client where for each device we kick off a Task and then `WhenAll` them.

e2e/LongHaul/amqpPooling/Parameters.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Microsoft.Azure.Devices.LongHual.AmqpPooling
+{
+    internal class Parameters
+    {
+        [Option(
+            'c',
+            "IoTHubConnectionString",
+            Required = false,
+            HelpText = "The connection string of the IoT hub instance to connect to.")]
```
drwill-ms, line 16:
```suggestion
            HelpText = "The service connection string of the IoT hub instance to connect to with permissions to create devices.")]
```

e2e/LongHaul/amqpPooling/RegisterManager.cs
```diff
@@ -0,0 +1,75 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Mash.Logging;
+
+namespace Microsoft.Azure.Devices.LongHual.AmqpPooling
+{
+    internal class RegisterManager : IDisposable
```
drwill-ms, line 12:
```suggestion
    internal class DeviceManager : IDisposable
```

e2e/LongHaul/amqpPooling/RegisterManager.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                string deviceId = device.Id;
+                _logger.Trace($"Deleting a device with Id {deviceId}", TraceSeverity.Verbose);
+                await s_serviceClient.Devices.DeleteAsync(deviceId).ConfigureAwait(false);
+            }
+
+            s_devices = null;
+        }
+
+        public async void Dispose()
```
drwill-ms, line 62: You can't just make Dispose async. It will not be awaited. On the first `await` in the method, the function will return to the caller and in this case, the program will exit. The devices will not get deleted.

e2e/LongHaul/amqpPooling/IotHub.cs
```diff
@@ -0,0 +1,192 @@
...
+                        var message = new TelemetryMessage(telemetryObject);
+
+                        foreach (KeyValuePair<string, string> prop in TelemetryUserProperties)
+                        {
+                            message.Properties.TryAdd(prop.Key, prop.Value);
+                        }
+
+                        try
+                        {
+                            _logger.Trace($"Sending a telemetry message from the device with Id [{deviceId}].", TraceSeverity.Information);
```
drwill-ms, line 109: You should add device-specific loggers that include their device Id and SDK version as custom properties.

drwill-ms (changes requested):
Comments inline

e2e/LongHaul/amqpPooling/IotHub.cs
brycewang-microsoft: Now I have a `List<Task>` for sending telemetry messages from each device and kick off them to [start ](https://github.com/Azure/azure-iot-sdk-csharp/blob/eea3f874c82f893562abf16e9b206fa2ea5ccf86/e2e/LongHaul/amqpPooling/IotHub.cs#L114)together with `Task.WhenAll`.

e2e/LongHaul/amqpPooling/DeviceManager.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                string deviceId = DevicePrefix + i;
+                _logger.Trace($"Creating a device with Id {deviceId}", TraceSeverity.Verbose);
+
+                var device = new Device(deviceId);
+                device = await s_serviceClient.Devices.CreateAsync(device, ct).ConfigureAwait(false);
+                s_devices.Add(device);
+            }
+
+            return s_devices;
```
drwill-ms, line 45: Might this work? Doesn't include logging, though, but you can add that in by changing the select statement to use a code block.

return _devicesCount
    .Select(id => s_serviceClient.Devices.CreateAsync(new Device(DevicePrefix + id))
    .ToList();

e2e/LongHaul/amqpPooling/DeviceManager.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            s_devices = null;
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            _logger.Trace($"Disposing {nameof(DeviceManager)} instance...", TraceSeverity.Verbose);
+
+            s_serviceClient?.Dispose();
```
drwill-ms, line 66: Don't dispose the service client before you've removed the devices!

e2e/LongHaul/amqpPooling/DeviceOperations.cs
```diff
@@ -0,0 +1,57 @@
...
+                _logger.Trace($"Sending a telemetry message from the device with Id [{_deviceId}].", TraceSeverity.Information);
+                sw.Restart();
+
+                await _deviceClient.SendTelemetryAsync(message, ct).ConfigureAwait(false);
+                sw.Stop();
+
+                _logger.Metric(TelemetryMessageDelaySeconds, sw.Elapsed.TotalSeconds);
+                _logger.Metric(TotalTelemetryMessagesSent, ++_totalTelemetryMessagesSent);
+
+                return Task.CompletedTask;
```
drwill-ms, line 47:
```suggestion
                return;
```

e2e/LongHaul/amqpPooling/DeviceOperations.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            _deviceClient = deviceClient;
+            _deviceId = deviceId;
+            logger.LoggerContext.Add("deviceId", deviceId);
+            _logger = logger;
+        }
+
+        public IotHubDeviceClient DeviceClient => _deviceClient;
+
+        public async Task<Task> SendAsync(TelemetryMessage message, CancellationToken ct)
```
drwill-ms, line 32:
```suggestion
        public async Task SendAsync(TelemetryMessage message, CancellationToken ct)
```

e2e/LongHaul/amqpPooling/DeviceOperations.cs
```diff
@@ -0,0 +1,57 @@
...
+                _logger.Metric(TelemetryMessageDelaySeconds, sw.Elapsed.TotalSeconds);
+                _logger.Metric(TotalTelemetryMessagesSent, ++_totalTelemetryMessagesSent);
+
+                return Task.CompletedTask;
+            }
+            catch (Exception ex)
+            {
+                _logger.Trace($"Exception when sending telemetry from the device with Id [{_deviceId}].\n{ex}", TraceSeverity.Warning);
+
+                return Task.FromException(ex);
```
drwill-ms, line 53:
```suggestion
                throw;
```

e2e/LongHaul/amqpPooling/IotHub.cs
```diff
@@ -0,0 +1,183 @@
...
+                {
+                    string deviceConnectionString = $"HostName={helper.HostName};DeviceId={device.Id};SharedAccessKey={device.Authentication.SymmetricKey.PrimaryKey}";
+
+                    var deviceClient = new IotHubDeviceClient(deviceConnectionString, _clientOptions);
+
+                    // Set the device client connection status change handler
+                    var amqpConnectionStatusChange = new AmqpConnectionStatusChange(_logger.Clone());
+                    deviceClient.ConnectionStatusChangeCallback = amqpConnectionStatusChange.ConnectionStatusChangeHandler;
+
+                    await deviceClient.OpenAsync().ConfigureAwait(false);
```
drwill-ms, line 65: Devices will need connection status callbacks to try to reconnect. Might as well put this call to open as well as that callback in the DeviceOperations class.

This code could simply create the devices and then kick off the operations which themselves handle all the device-specific concerns, including looping for telemetry.

e2e/LongHaul/amqpPooling/IotHub.cs
```diff
@@ -0,0 +1,183 @@
...
+                {
+                    foreach (KeyValuePair<string, DeviceOperations> entry in s_deviceOperations)
+                    {
+                        string deviceId = entry.Key;
+                        DeviceOperations deviceOperations = entry.Value;
+
+                        var telemetryObject = new DeviceTelemetry
+                        {
+                            DeviceId = deviceId,
+                            SystemHealth = systemHealth,
```
drwill-ms, line 101: Probably not much reason for each device to publish system health, as they are all in the same process.

You could gather than regularly and log it in a loop as another task to kick off.

e2e/LongHaul/amqpPooling/IotHub.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        public async Task SendTelemetryMessagesAsync(CancellationToken ct)
+        {
+            while (!ct.IsCancellationRequested)
+            {
+                await Task.Delay(s_messageLoopSleepTime, ct).ConfigureAwait(false);
+
+                List<Task> deviceOperationTasks = new();
+
+                _messagesToSend.TryDequeue(out SystemHealthTelemetry systemHealth);
```
drwill-ms, line 85: We used a queue for message sending in the normal long haul. I'd be fine if in this one you simply sleep/send telemetry from the device, so no need for queue.

This would work because we can collect system performance metrics and log them with AI in a loop that does not get paused by disconnections.

e2e/LongHaul/amqpPooling/DeviceManager.cs
drwill-ms, line 1: Casing for directory can match the others and be AmqpPooling?

e2e/LongHaul/amqpPooling/DeviceManager.cs
brycewang-microsoft: No, it's not working. I didn't find it simple to use LINQ here, so I would rather keep it as-is.

e2e/LongHaul/amqpPooling/DeviceOperations.cs
brycewang-microsoft: I tried to return a `Task` from this method which would be added into a `List<Task>` further, so I can call [this](https://github.com/Azure/azure-iot-sdk-csharp/blob/eea3f874c82f893562abf16e9b206fa2ea5ccf86/e2e/LongHaul/amqpPooling/IotHub.cs#L114) to kick off operations for all devices together.

e2e/LongHaul/amqpPooling/DeviceManager.cs
brycewang-microsoft: We are currently naming all directories like [this](https://github.com/Azure/azure-iot-sdk-csharp/tree/brycewang/long-haul-amqp-pooling/e2e/LongHaul), I can modify for all together though, "Device", "Module", "Service", "AmqpPooling", etc.

e2e/LongHaul/amqpPooling/IotHub.cs
brycewang-microsoft: With the comment below, I will just create and use another property within `DeviceTelemetry`, maybe something like "GuidValue".

e2e/LongHaul/amqpPooling/DeviceManager.cs
```diff
@@ -0,0 +1,92 @@
...
+        public async Task RemoveDevicesBeforeRunningAsync(CancellationToken ct)
+        {
+            _logger.Trace($"Clean up devices with Id prefix [{DevicePrefix}] before running the app.", TraceSeverity.Information);
+
+            AsyncPageable<ClientTwin> allDevices = s_serviceClient.Query.Create<ClientTwin>("SELECT deviceId FROM devices", ct);
+
+            await foreach (ClientTwin device in allDevices)
+            {
+                string deviceId = device.DeviceId;
+                if(deviceId.StartsWith(DevicePrefix))
```
drwill-ms, line 41:
```suggestion
                if (deviceId.StartsWith(DevicePrefix))
```

e2e/LongHaul/amqpPooling/DeviceOperations.cs
drwill-ms: Something is wrong with this implementation, though. You shouldn't need to do it.

e2e/LongHaul/amqpPooling/IotHub.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+            finally
+            {
+                _lifetimeControl.Release();
+            }
+        }
+
+        public async Task SendTelemetryMessagesAsync(CancellationToken ct)
+        {
+            while (!ct.IsCancellationRequested)
```
drwill-ms, line 72: This would be a lot cleaner if the `SendAsync` method in `DeviceOperations` had it's own loop.

Then here, you'd simply kick off the tasks for each, store the task in your array, and `Task.WhenAll` it.

drwill-ms (approved):

PR #3297, opened 2023-04-21
Add comment for HTTP client usage for module clients

abhipsaMisra:


timtay-microsoft:
Got it. #3299 renames the HTTP settings object to reflect that both the file upload operations and module method invocation methods pull from these settings. Thanks for pointing this out!

PR #3298, opened 2023-04-21
Do not install certs to cert store

drwill-ms:
We will probably want to do more here to ensure without having certs installed by the user that the requests work, but in the interest of preventing a breaking behavior change after we GA, I'm removing the offending code in the SDKs now.


provisioning/device/src/Authentication/AuthenticationProviderX509.cs
```diff
@@ -16,6 +16,13 @@ public class AuthenticationProviderX509 : AuthenticationProvider
         /// </summary>
         /// <remarks>
         /// Ensure that you dispose any supplied <see cref="X509Certificate2"/> after you are done using it to ensure there are no memory leaks.
+        /// <para>
+        /// The certificates provided are assumed to be a full chain and are not validated. The caller should
+        /// validate them independently.
+        /// </para>
+        /// <para>
+        /// If the certificates need to be installed to the local certificate store, that should be done by the user.
```
patilsnr, line 24: Do the ms docs have any articles about this? If so it could be nice to link one.

patilsnr (approved):

abhipsaMisra (approved):

abhipsaMisra:
Our tests work fine with this change, right?

timtay-microsoft (approved):

provisioning/device/src/Authentication/AuthenticationProviderX509.cs
drwill-ms: I don't know of any.

drwill-ms:
After reviewing with Cris Pop, we understand that 1) installing to this cert store does not make it a trusted CA root cert, and 2) it is required to be done by the user or by the SDK (as a convenience) for the request to succeed. 

I'll follow up with a different PR to update comments and tweak a few things I learned during this task.

PR #3301, opened 2023-04-24
Update code and doc comments about cert chains

drwill-ms:


e2e/Tests/provisioning/ProvisioningE2ETests.cs
```diff
@@ -678,7 +678,6 @@ public async Task DPS_Registration_AmqpWs_X509_IndividualEnrollment_InvalidGloba
                 auth,
                 clientOptions);
 
-            using var cts = new CancellationTokenSource(PassingTimeoutMilliseconds);
```
drwill-ms, line 681: This gave us 1 big timeout. If a single operation failed due to timeout, we would not try again. I broke it up so we'd get multiple attempts, with each attempt being a reasonable duration.

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -132,6 +132,13 @@ public IotHubConnectionCredentials(string iotHubConnectionString)
         /// <summary>
         /// The full chain of certificates from the one used to sign the client certificate to the one uploaded to the service.
         /// </summary>
+        /// <remarks>
+        /// If supplied, the certificate chain will be verified and sent to the service in the request by adding them to the current user
```
drwill-ms, line 136: User-facing doc comment to help users navigate how this is used and how they can opt-in or -out of it.

iothub/device/src/Authentication/IotHubConnectionCredentials.cs
```diff
@@ -132,6 +132,13 @@ public IotHubConnectionCredentials(string iotHubConnectionString)
...
+        /// intermediate certificate store.
```
CIPop, line 137: Nit: "Intermediate Certification Authorities" store.
Optionally we can also point to https://learn.microsoft.com/en-us/troubleshoot/developer/webapps/iis/www-authentication-authorization/configure-intermediate-certificates (or similar).

CIPop (approved):

abhipsaMisra (approved):

PR #3302, opened 2023-04-24
LongHaul - minor bug fix

tmahmood-microsoft:


e2e/LongHaul/device/SystemHealthMonitor.cs
```diff
@@ -20,7 +20,7 @@ internal class SystemHealthMonitor
         private readonly IotHub _iotHub;
         private readonly Logger _logger;
         private static int s_port;
-        private static readonly TimeSpan s_interval = TimeSpan.FromSeconds(3);
+        private static readonly TimeSpan s_interval = TimeSpan.FromSeconds(4);
```
tmahmood-microsoft, line 23: we create and send telemetry message every 3 seconds. This caused message backlog with telemetry message delay of 1-2 seconds on average.
To avoid message backlog, we check sending any pending messages every 3 seconds and create telemetry message every 4 seconds.

e2e/LongHaul/service/IotHub.cs
```diff
@@ -100,7 +100,6 @@ await foreach (ClientTwin deviceTwin in allDeviceTwins)
                             async Task OperateWithDeviceAsync()
                             {
                                 Logger deviceLogger = _logger.Clone();
-                                deviceLogger.LoggerContext.Add(DeviceId, deviceId);
```
tmahmood-microsoft, line 103: We already add logger context information in ctor() for DeviceOperations. This duplicate add() caused it to fail and deviceOperations never running.

patilsnr (approved):

e2e/LongHaul/device/SystemHealthMonitor.cs
drwill-ms: I don't think that is a problem per se. If it was, we can increase the rate at which we check for telemetry to send (reduce from 10 seconds).

drwill-ms: Note, we may want to change this to even 1/2 a second at some point, so let's fix the other bottleneck.

e2e/LongHaul/service/IotHub.cs
drwill-ms: My mistake.

drwill-ms (approved):

PR #3303, opened 2023-04-25
Fix device AMQP layer not proactively renewing tokens

timtay-microsoft:
All the logic is already in place for this to work. The only problem was that the default delegating handler never did anything when SetSasTokenRefreshesOn() is called. Now it delegates it to the retry delegating handler who spins up the proactive token renewal loop


patilsnr (approved):

drwill-ms (approved):

tmahmood-microsoft (approved):

abhipsaMisra (approved):
I'm surprised this was missed.

Issue #3199, opened 2023-03-27
[Bug Report]  NullReferenceException was thrown from RegistryManager.GetTwinAsync 


iryanzhang:
Nuget Verison 1.38.2

Class:  RegistryManager
Method: GetTwinAsync 
Exception: System.NullReferenceException 
Message: Object reference not set to an instance of an object
Source: Microsoft.Azure.Devices

from the stack trace. looks like there is an error occurred when GetExceptionCode.

at Microsoft.Azure.Devices.ExceptionHandlingHelper.GetExceptionCodeAsync(HttpResponseMessage response)
   at Microsoft.Azure.Devices.ExceptionHandlingHelper.<>c.<<-cctor>b__5_8>d.MoveNext()
--- End of stack trace from previous location ---
   at Microsoft.Azure.Devices.HttpClientHelper.MapToExceptionAsync(HttpResponseMessage response, IDictionary`2 errorMapping)
   at Microsoft.Azure.Devices.HttpClientHelper.ExecuteAsync(HttpClient httpClient, HttpMethod httpMethod, Uri requestUri, Func`3 modifyRequestMessageAsync, Func`2 isMappedToException, Func`3 processResponseMessageAsync, IDictionary`2 errorMappingOverrides, CancellationToken cancellationToken)
   at Microsoft.Azure.Devices.HttpClientHelper.GetAsync[T](Uri requestUri, TimeSpan operationTimeout, IDictionary`2 errorMappingOverrides, IDictionary`2 customHeaders, Boolean throwIfNotFound, CancellationToken cancellationToken)

brycewang-microsoft:
Hi @iryanzhang, thanks for reaching out to us! Could you also share the following info and a code snippet to reproduce this issue with us?

# Context

- **OS, version, SKU and CPU architecture used:** <VERSION> (Windows 10 Desktop x64, Ubuntu 15.04 x86, Windows 10 IoT Core arm32, etc.)
- **Application's .NET Target Framework :** <NETVERSION> (See https://docs.microsoft.com/en-us/dotnet/standard/frameworks. E.g. netcoreapp2.1, net451, uap10.0, xamarin)
- **Device:** <DEVICETYPE> (Laptop, Raspberry PI3, Android APIv25 etc.)

iryanzhang:
OS: Windows 10 x64
Framework .Net 6.0
Device One VM node in Service Fabric

[...]

PR #3304, opened 2023-04-25
Fix NRE from GetExceptionCodeAsync

brycewang-microsoft:
https://github.com/Azure/azure-iot-sdk-csharp/issues/3199

### DIFF
--- a/common/src/service/ExceptionHandlingHelper.cs
+++ b/common/src/service/ExceptionHandlingHelper.cs
@@ -103,15 +103,12 @@ namespace Microsoft.Azure.Devices
         public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage response)
         {
             string responseContentStr = "";
-            if (response.Content != null)
-            {
-                // First we will attempt to retrieve the error code from the response content.
-                responseContentStr = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
-            }
-            else
+            if (response.Content == null)
             {
                 return ErrorCode.InvalidErrorCode;
             }
+            // First we will attempt to retrieve the error code from the response content.
+            responseContentStr = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
 
             // There are two things to consider when surfacing service errors to the user, the 6-digit error code and the code description. Ideally, when a backend service
             // returns an error, both of these fields are set in the same place. However, IoT hub is returning the 6-digit code in the response content, while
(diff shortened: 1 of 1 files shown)


abhipsaMisra (approved):

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -102,8 +102,16 @@ public static Task<string> GetExceptionMessageAsync(HttpResponseMessage response
...
+            string responseContentStr = "";
+            if (response.Content != null)
+            {
+                // First we will attempt to retrieve the error code from the response content.
+                responseContentStr = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+            }
+            else
+            {
+                return ErrorCode.InvalidErrorCode;
+            }
```
drwill-ms, line 114:
```suggestion
            if (response.Content == null)
            {
                return ErrorCode.InvalidErrorCode;
            }

            // First we will attempt to retrieve the error code from the response content.
            responseContentStr = await response.Content.ReadAsStringAsync().ConfigureAwait(false);           
```

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -120,7 +128,11 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
 
                 try
                 {
-                    Dictionary<string, string> messageFields = JsonConvert.DeserializeObject<Dictionary<string, string>>(responseContent.Message);
+                    var messageFields = new Dictionary<string, string>();
+                    if (responseContent != null && responseContent.Message != null)
```
drwill-ms, line 132:
```suggestion
                    if (responseContent?.Message != null)
```

drwill-ms (approved):

common/src/service/ExceptionHandlingHelper.cs
```diff
@@ -120,7 +128,11 @@ public static async Task<ErrorCode> GetExceptionCodeAsync(HttpResponseMessage re
...
+                    {
+                        messageFields = JsonConvert.DeserializeObject<Dictionary<string, string>>(responseContent.Message);
+                    }
 
                     if (messageFields != null
```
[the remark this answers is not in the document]
brycewang-microsoft: That's what I thought before, but what I learnt by running this [test](https://github.com/Azure/azure-iot-sdk-csharp/blob/main/iothub/service/tests/ExceptionHandlingHelperTests.cs#L56) is that, in case of deserializing an empty `responseContent.Message`, the generated `messageFields` can be null, so we still need null check for this here.

PR #3305, opened 2023-04-26
LongHaul - Updated setup file and migration guide

tmahmood-microsoft:


e2e/LongHaul/module/Parameters.cs
```diff
@@ -20,12 +20,19 @@ public enum JsonSerializingLibrary
 
     internal class Parameters
     {
+        [Option(
+            'g',
+            "GatewayHostName",
+            Required = true,
+            HelpText = "The gateway edge device to connects to.")]
+        public string GatewayHostName { get; set; }
```
abhipsaMisra, line 28: How about we make this optional and call out that if gateway host name is passed in then it is treated as an edge module, whereas if the gateway host isn't passed in then it is treated as IoT device module. That way the same test app can be used to test both scenarios.

e2e/LongHaul/prerequisites/LongHaulSetup.ps1
```diff
@@ -255,7 +263,7 @@ $longhaulModuleConnectionString = az iot hub module-identity connection-string s
 $keyvaultKvps = @{
     # Environment variables for IoT Hub E2E tests
     "IOTHUB-LONG-HAUL-DEVICE-CONNECTION-STRING" = $longhaulDeviceConnectionString;
-    "IOTHUB-LONG-HAUL-MODULE-CONNECTION-STRING" = $longhaulModuleConnectionString;
```
abhipsaMisra, line 258: I'd recommend having both variations of the module identity so that we can test both flows

e2e/LongHaul/module/Parameters.cs
```diff
@@ -20,12 +20,19 @@ public enum JsonSerializingLibrary
...
+
         [Option(
             'c',
             "ConnectionString",
             Required = false,
-            HelpText = "The connection string for the module to simulate.")]
-        public string ConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_LONG_HAUL_MODULE_CONNECTION_STRING");
+            HelpText = "The connection string for the edge module to simulate.")]
```
brycewang-microsoft, line 34: if we still want this workable for both device module and edge module, 
```suggestion
            HelpText = "The connection string for the device/edge module to simulate.")]
```

SDK v2 migration guide.md
```diff
@@ -198,6 +198,7 @@ Find a client you currently use below, read the table of API name changes and us
 | `Message.CreationTimeUtc` | `TelemetryMessage.CreatedOnUtc`, `IncomingMessage.CreatedOnUtc` | Conforming to the naming guidelines by the Azure SDK team, where DateTime/Offset types have an "On" suffix (and "Utc" suffix when explicitly in UTC).³ |
 | `Message.EnqueuedTimeUtc` | `TelemetryMessage.EnqueuedtimeUtc`, `IncomingMessage.EnqueuedTimeUtc` | See³ |
 | `Message.ExpiryTimeUtc` | `TelemetryMessage.ExpiresOnUtc`, `IncomingMessage.ExpiresOnUtc` | See³ |
+| `Message.GetBytes` | `IncomingMessage.GetPayloadAsBytes` | See³ |
```
drwill-ms, line 201: This change has nothing to do with Azure SDK naming guidelines.

I'd file it under "More descriptive name."

e2e/LongHaul/module/IotHub.cs
```diff
@@ -41,7 +42,8 @@ internal sealed class IotHub : IAsyncDisposable
         public IotHub(Logger logger, Parameters parameters)
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
-            _moduleConnectionString = parameters.ConnectionString;
+            _moduleConnectionString = parameters.GatewayHostName == null ?
```
drwill-ms, line 45: Instead of messing with the connection string, can we just use the ClientOptions.GatewayHostname property?

e2e/LongHaul/module/Parameters.cs
```diff
@@ -24,16 +24,30 @@ internal class Parameters
             'c',
             "ConnectionString",
             Required = false,
-            HelpText = "The connection string for the module to simulate.")]
+            HelpText = "The connection string for the IoT Hub module to simulate.")]
         public string ConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_LONG_HAUL_MODULE_CONNECTION_STRING");
```
drwill-ms, line 28:
```suggestion
        public string DeviceModuleConnectionString { get; set; } = Environment.GetEnvironmentVariable("IOTHUB_LONG_HAUL_MODULE_CONNECTION_STRING");
```

e2e/LongHaul/module/Parameters.cs
drwill-ms: Looks like he's using 2 different parameters, one for each type, because the environment variables are different.

e2e/LongHaul/module/Parameters.cs
```diff
@@ -34,6 +41,13 @@ internal class Parameters
             HelpText = "The instrumentation key string for application insights.")]
         public string InstrumentationKey { get; set; } = Environment.GetEnvironmentVariable("APPLICATION_INSIGHTS_INSTRUMENTATION_KEY");
 
+        [Option(
+            'g',
+            "GatewayHostName",
+            Required = false,
+            HelpText = "The gateway edge device to connects to; if defined, this app will use the EdgeModuleConnectionString, otherwise it uses the DeviceModuleConnectionString.")]
```
patilsnr, line 48: nit: should it be `to connect to`?

patilsnr (approved):

PR #3308, opened 2023-04-27
Add code examples for public APIs in IotHubDeviceClient, IotHubModuleClient, IotHubBaseClient and retry policies

brycewang-microsoft:


iothub/device/src/IotHubBaseClient.cs
```diff
@@ -208,6 +218,11 @@ public async Task SendTelemetryAsync(TelemetryMessage message, CancellationToken
         /// <exception cref="OperationCanceledException">The operation has been canceled.</exception>
         /// <exception cref="IotHubClientException">An error occured when communicating with IoT hub service.</exception>
         /// <exception cref="ObjectDisposedException">The client has been disposed.</exception>
+        /// <example>
+        /// <code language="csharp">
+        /// await client.SendTelemetryAsync(new List<TelemetryMessage/> { message1, message2 }, cancellationToken);
```
tmahmood-microsoft, line 223:
```suggestion
        /// await client.SendTelemetryAsync(new List<TelemetryMessage> { message1, message2 }, cancellationToken);
```

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -251,6 +266,18 @@ public async Task SendTelemetryAsync(IEnumerable<TelemetryMessage> messages, Can
         /// <exception cref="OperationCanceledException">The operation has been canceled.</exception>
         /// <exception cref="IotHubClientException">An error occured when communicating with IoT hub service.</exception>
         /// <exception cref="ObjectDisposedException">The client has been disposed.</exception>
+        /// <example>
+        /// <code language="csharp">
+        /// await client.SetIncomingMessageCallbackAsync(OnC2dMessageReceivedAsync);
+        /// //...
+        /// 
+        /// Task<MessageAcknowledgement/> OnC2dMessageReceivedAsync(IncomingMessage receivedMessage)
```
tmahmood-microsoft, line 274:
```suggestion
        /// Task<MessageAcknowledgement> OnC2dMessageReceivedAsync(IncomingMessage receivedMessage)
```

iothub/device/src/IotHubBaseClient.cs
brycewang-microsoft, line 223: forward slash is needed for `List<TelemetryMessage/>` in XML, otherwise it will complain with [CS1570](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1570).

tmahmood-microsoft, line 223: should it be like:

/// await client.SendTelemetryAsync(new List &lt;TelemetryMessage&gt; { message1, message2 }, cancellationToken);

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -128,6 +128,11 @@ public abstract class IotHubBaseClient : IAsyncDisposable
         /// <exception cref="OperationCanceledException">The operation has been canceled.</exception>
         /// <exception cref="IotHubClientException">An error occured when communicating with IoT hub service.</exception>
         /// <exception cref="ObjectDisposedException">The client has been disposed.</exception>
+        /// <example>
+        /// <code language="csharp">
+        /// await client.OpenAsync(cancellationToken);
```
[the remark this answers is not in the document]
abhipsaMisra: Not normally, no. Our client will retry retryable errors for this API. There is no end retry count for our default policy. It will either continue with no end, or until a passed cancellation token is cancelled. So, a user shouldn't require anything additional to be done _normally_.

iothub/device/src/IotHubBaseClient.cs
brycewang-microsoft: check my comment above

tmahmood-microsoft (approved):

iothub/device/src/IotHubBaseClient.cs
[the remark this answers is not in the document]
drwill-ms: Not if the cancellation token is signalled.

I think what bothers me about this example is that it is too naive and straight forward. If the device doesn't connect then the IoT device is out of commission and may need human intervention.

Without having a whole reconnection sample here (which is way overkill), there are 2 things I think we can do:

1. Put the call to open in context. The idea is we initialize the client, subscribe to connection status changes, and then call open. Nothing else can happen before those 3 steps, and setting the connection status callback is best practice so you are notified that you are connected. I'd suggest we add all 3 statements to this example and end with a comment to begin device operations, such as telemetry.

2. Add a remark that the call to open is important and may require some amount of retry above and beyond the client retry policy (especially if it is disabled or if the cancellation token is signaled). One might choose to use a generous cancellation token but then (if possible) restart networking and try again. Later, once we're in main, we can add a link to the reconnection sample.

Does that sound reasonable?

iothub/device/src/IotHubBaseClient.cs
drwill-ms: Also, should show initializing the client with AMQP, as it is not the default, with a comment that this only works over AMQP.

brycewang-microsoft: > I think you need to replace the <> signs with < and >, actually.

Tried with `new List < TelemetryMessage > ` but still complains.

drwill-ms: Sorry, what I wrote didn't display right. Let me try again with a code block:

```
&lt;TelemetryMessage&gt;
```

drwill-ms: I could be wrong. Perhaps we reach out to Alma Jenks and ask how to format these correctly in doc comment code blocks?

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -251,6 +270,18 @@ public async Task SendTelemetryAsync(IEnumerable<TelemetryMessage> messages, Can
...
         /// <exception cref="IotHubClientException">An error occured when communicating with IoT hub service.</exception>
         /// <exception cref="ObjectDisposedException">The client has been disposed.</exception>
+        /// <example>
+        /// <code language="csharp">
+        /// await client.SetIncomingMessageCallbackAsync(OnC2dMessageReceivedAsync);
+        /// //...
+        /// 
+        /// Task<MessageAcknowledgement/> OnC2dMessageReceivedAsync(IncomingMessage receivedMessage)
+        /// {
+        ///     // Add message handling logic as needed
```
drwill-ms, line 280: Should we include example of using a semaphore to ensure ordering?

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -308,6 +339,18 @@ public async Task SendTelemetryAsync(IEnumerable<TelemetryMessage> messages, Can
...
         /// <exception cref="IotHubClientException">An error occured when communicating with IoT hub service.</exception>
         /// <exception cref="ObjectDisposedException">The client has been disposed.</exception>
+        /// <example>
+        /// <code language="csharp">
+        /// await client.SetDirectMethodCallbackAsync(OnDirectMethodCalledAsync, cancellationToken);
+        /// //...
+        /// 
+        /// Task<DirectMethodResponse/> OnDirectMethodCalledAsync(DirectMethodRequest directMethodRequest)
+        /// {
+        ///     // Add method request handling logic as needed
```
drwill-ms, line 349: Let's show inspecting the method name to compare to a known value, and then maybe getting the expected payload type.

iothub/device/src/IotHubBaseClient.cs
```diff
@@ -351,11 +394,16 @@ public async Task SendTelemetryAsync(IEnumerable<TelemetryMessage> messages, Can
...
         /// <param name="cancellationToken">A cancellation token to cancel the operation.</param>
+        /// <returns>The twin object for the current client.</returns>
         /// <exception cref="InvalidOperationException">The client instance is not already open.</exception>
         /// <exception cref="OperationCanceledException">The operation has been canceled.</exception>
         /// <exception cref="IotHubClientException">An error occured when communicating with IoT hub service.</exception>
         /// <exception cref="ObjectDisposedException">The client has been disposed.</exception>
-        /// <returns>The twin object for the current client.</returns>
+        /// <example>
+        /// <code language="csharp">
+        /// TwinProperties twin = await client.GetTwinPropertiesAsync(cancellationToken);
```
drwill-ms, line 404: Can we show how to enumerate desired and reported properties as well?

iothub/device/src/IotHubDeviceClient.cs
```diff
@@ -121,6 +141,17 @@ private IotHubDeviceClient(IotHubConnectionCredentials iotHubConnectionCredentia
         /// <exception cref="OperationCanceledException">The operation has been canceled.</exception>
         /// <exception cref="IotHubClientException">An error occured when communicating with IoT hub service.</exception>
         /// <exception cref="ObjectDisposedException">The client has been disposed.</exception>
+        /// <example>
+        /// <code language="csharp">
+        /// await client.CompleteFileUploadAsync(
+        ///     new FileUploadCompletionNotification(correlationId: correlationId, isSuccess: true)
+        ///     {
+        ///         StatusCode = 200,
```
drwill-ms, line 149:
```suggestion
        ///         // optional properties...
        ///         StatusCode = 200,
```

iothub/device/src/IotHubModuleClient.cs
```diff
@@ -87,8 +101,13 @@ internal IotHubModuleClient(IotHubConnectionCredentials iotHubConnectionCredenti
...
         /// <param name="options">The options that allow configuration of the module client instance during initialization.</param>
-        /// <exception cref="InvalidOperationException">The required environmental variables were missing. Check the exception thrown for additional details.</exception>
         /// <returns>A disposable client instance.</returns>
+        /// <exception cref="InvalidOperationException">The required environmental variables were missing. Check the exception thrown for additional details.</exception>
+        /// <example>
+        /// <code language="csharp">
+        /// await using var client = await IotHubModuleClient.CreateFromEnvironmentAsync(new IotHubClientOptions(new IotHubClientMqttSettings(IotHubClientTransportProtocol.WebSocket)));
+        /// </code>
+        /// </example>
         public static async Task<IotHubModuleClient> CreateFromEnvironmentAsync(IotHubClientOptions options = default)
```
drwill-ms, line 111: Any reason why we don't have a cancellation token in this async method?

Not for your PR, Bryce, but maybe we should add a task to add one here.

iothub/device/src/IotHubModuleClient.cs
```diff
@@ -163,6 +187,11 @@ public async Task SendMessageToRouteAsync(string outputName, TelemetryMessage me
         /// <exception cref="OperationCanceledException">The operation has been canceled.</exception>
         /// <exception cref="IotHubClientException">An error occured when communicating with IoT hub service.</exception>
         /// <exception cref="ObjectDisposedException">The client has been disposed.</exception>
+        /// <example>
+        /// <code language="csharp">
+        /// await client.SendMessagesToRouteAsync(outputName, new List<TelemetryMessage/> { message1, message2 }, cancellationToken);
```
drwill-ms, line 192: Include client init over AMQP with a comment saying it is only supported over AMQP?

iothub/device/src/RetryPolicies/IIotHubClientRetryPolicy.cs
```diff
@@ -18,6 +18,17 @@ public interface IIotHubClientRetryPolicy
         /// <param name="lastException">The exception that prompted this retry policy check.</param>
         /// <param name="retryDelay">Set this to the desired time to delay before the next attempt.</param>
         /// <returns>True if the operation should be retried; otherwise false.</returns>
+        /// <example>
+        /// <code language="csharp">
+        /// class CustomRetryPolicy : IIotHubClientRetryPolicy
+        /// {
+        ///     public bool ShouldRetry(uint currentRetryCount, Exception lastException, out TimeSpan retryDelay)
+        ///     {
+        ///         // Add custom logic as needed upon determining if it should retry
```
drwill-ms, line 27:
```suggestion
        ///         // Add custom logic as needed upon determining if it should retry and set the retryDelay out parameter
```

iothub/device/src/RetryPolicies/IotHubClientNoRetry.cs
```diff
@@ -13,6 +13,16 @@ public class IotHubClientNoRetry : IIotHubClientRetryPolicy
...
         /// Create an instance of this class.
         /// </summary>
+        /// <example>
+        /// <code language="csharp">
+        /// var noRetry = new IotHubClientNoRetry();
+        /// 
+        /// var clientOptions = new IotHubClientOptions(new IotHubClientMqttSettings(IotHubClientTransportProtocol.WebSocket))
+        /// {
+        ///     RetryPolicy = noRetry,
+        /// };
```
drwill-ms, line 23: Consider inlining it as it is pretty simple.

```suggestion
        /// var clientOptions = new IotHubClientOptions(new IotHubClientMqttSettings(IotHubClientTransportProtocol.WebSocket))
        /// {
        ///     RetryPolicy = new IotHubClientNoRetry(),
        /// };
```

drwill-ms (approved):
I may not be able to see the changes that come out of this, but I feel it is low risk as it is just doc comments. I'm signing off optimistically to help you finish it without me bottlenecking ya. :)

iothub/device/src/IotHubBaseClient.cs
brycewang-microsoft: > &lt;TelemetryMessage&gt;

Actually this works!

Issue #3310, opened 2023-04-28
[Technical Question] How to implement a manual Start/Stop mechanism on the IOT Device Client Re-Connection Sample 


gagwithgaffer:
Hello @drwill-ms , hope all is well.

Using the latest re-connection sample for the device client, I'm struggling to figure out how to re-start the connection without it throwing exceptions. In my project, I need to be able to start and stop the connection to Azure Hub on demand, the sample is based upon a console project but my project is a web app so i provide buttons on the web page to start/stop the process. 

The notes in the code sample readme say "If you want to perform more operations on the device client, you should dispose and then re-initialize the client."

I have tried this by adding another method call for closing the client within the reconnection sample class. I've then tried multiple different combinations of actions within this method, such as closing the client then disposing it. Disposing the client throws an exception, if i simply close the client connection alone, I can then start up the process again but get exceptions further down the code sample in the InitializeAndSetupClientAsync method call where it checks whether the s_deviceClient is null.

**Scenario 1 - Closing the client but not disposing it.**

`public async Task CloseConnection()
{
	try
	{
		await s_deviceClient.CloseAsync(CancellationToken.None);
	}
	catch (Exception ex)
	{
		var error = ex.ToString();
	}
}`

![image](https://user-images.githubusercontent.com/51385414/235191746-33f3a335-1629-41d2-a8c3-beadb5a5ad83.png)

![image](https://user-images.githubusercontent.com/51385414/235191808-1cd914be-9b26-43eb-bf46-b520f49ae035.png)

**Scenario 2 - Closing the client then disposing it at the same time.**

I have less luck trying this, because I get an exception when trying to dispose the client.

![image](https://user-images.githubusercontent.com/51385414/235192524-0a489a18-8fc1-4ce9-9c46-560616e4a10b.png)

I wandered whether this may be something to do with tha fact that each time I call the CloseConnection method, I'm calling this from an async Task method which is essentially running from a new thread from the task pool and is not necessarily referring to the same client instance? but the fact that the s_deviceClient is created as a static volatile in the reconnection sample class makes me think I should be referring to the same device client instance when tryng to dispose it.

Since our last teams call back in August I had noticed some significant changes to the code sample (namely being your approach on the new retry policy mechanism) which looks like you were able to remove a lot of the code used in the older sample, so was keen to adopt the newer sample but just struglling to implement a start/stop mechanism correctly, thanks

drwill-ms:
[...]

PR #3312, opened 2023-04-28
Avoid NRE and disposed exceptions when closing client twice

drwill-ms:
From GitHub issue #3310.

### DIFF
--- a/e2e/test/iothub/DeviceClientE2eTests.cs
+++ b/e2e/test/iothub/DeviceClientE2eTests.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.Azure.Devices.E2ETests.Helpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Azure.Devices.E2ETests
+{
+    [TestClass]
+    [TestCategory("E2E")]
+    [TestCategory("IoTHub")]
+    public class DeviceClientE2eTests
+    {
+        [TestMethod]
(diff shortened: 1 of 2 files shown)


iothub/device/src/Pipeline/RetryDelegatingHandler.cs
```diff
@@ -784,16 +784,17 @@ public override async Task CloseAsync(CancellationToken cancellationToken)
...
                 {
                     return;
                 }
 
                 if (Logging.IsEnabled)
                     Logging.Enter(this, cancellationToken, nameof(CloseAsync));
 
-                _handleDisconnectCts.Cancel();
-                _cancelPendingOperationsCts.Cancel();
+                _handleDisconnectCts?.Cancel();
```
drwill-ms, line 796: I don't see how these get to null, but we're getting an NRE on this line when closing twice. I'd rather the client get an ObjectDisposedException than NRE, but we should be protected from even that by the change on 788.

e2e/test/iothub/DeviceClientE2eTests.cs
```diff
@@ -0,0 +1,44 @@
...
+                // act
+                Func<Task> act = () => client.CloseAsync();
+
+                // assert
+                await act.Should().NotThrowAsync().ConfigureAwait(false);
+            }
+            finally
+            {
+                await testDevice.RemoveDeviceAsync().ConfigureAwait(false);
+                testDevice.Dispose();
```
patilsnr, line 39: do we need to call dispose if already having a `using` keyword?

patilsnr (approved):

e2e/test/iothub/DeviceClientE2eTests.cs
drwill-ms: The using is on the TestDevice, which only disposes the auth, not the device client.

drwill-ms: Actually, you are right. I don't need to call dispose explicitly on this one either, as there is a using on both.

PR #3314, opened 2023-05-01
Version bump for release on main

brycewang-microsoft:
# Microsoft Azure IoT SDKs for .NET Release 2023-05-01

## Microsoft.Azure.Devices.Client 1.42.0
- Apply JsonSerializerSettings to hub device (#3107)
- Set state to 'closed' when CloseAsync() is called (#3149)
- Ensure device client methods throw an object disposed exception (#3212)
- Allow users to pass in HttpClient to use for file upload operations (#3293)
- Update service error codes (#3315)
- Update reference to `Microsoft.Azure.Devices.Shared` nuget.

### Bug fixes
- Stop the sas token refresh loop once CloseAsync() is called (#3021)
- Send an event in ReceiveMessageHandler before completing received message (#3116)
- Fix synchronization of semaphore in retry delegating handler (#3135)
- Allow a message to be sent to multiple outputs (#3203)
- Fix file upload HTTP client not handling DNS changes (#3266)
- Properly set connection lease timeout regardless of .NET target framework (#3279)
- Avoid NRE and disposed exceptions when closing client twice (#3312)

___
## Microsoft.Azure.Devices 1.39.0
- Update error message for service reported errors (#3183)
- Update service error codes (#3315)
- Replace custom methods with string methods (#2801)
  - `public static bool IsNullOrWhiteSpace(this string value)` is removed due to duplicated functionality and users should use the native API on the string class instead - `string.IsNullOrWhiteSpace(value)`.
- Update reference to `Microsoft.Azure.Devices.Shared` nuget.

### Bug fixes
- Fix bug in date deserialization of device method response  (#3097)
- Fix where HttpClientHelper should not log when logging is disabled (#3150)
- Fix NRE from GetExceptionCodeAsync (#3304)

___

### Bug fixes
- Stop using JObject.Parse for json string in TwinCollection (#3113)

___

## Microsoft.Azure.Devices.Provisioning.Client 1.19.3
- Apply JsonSerializerSettings to provisioning device (#3113)
- Update reference to `Microsoft.Azure.Devices.Shared` nuget.
___

## Microsoft.Azure.Devices.Provisioning.Service 1.18.3
- Apply JsonSerializerSettings to provisioning service (#3113)
- Update reference to `Microsoft.Azure.Devices.Shared` nuget.

___
## Microsoft.Azure.Devices.Provisioning.Transport.Amqp 1.16.5
- Update reference to `Microsoft.Azure.Devices.Shared` nuget.
___

## Microsoft.Azure.Devices.Provisioning.Transport.Http 1.15.3
- Update reference to `Microsoft.Azure.Devices.Shared` nuget.

### Bug fixes
- Fix DPS registration over HTTP with escaped characters in the registration Id (#3059)
___
## Microsoft.Azure.Devices.Provisioning.Transport.Mqtt 1.17.3
- Update reference to `Microsoft.Azure.Devices.Shared` nuget.

___
## Microsoft.Azure.Devices.Provisioning.Security.Tpm 1.14.3
- Update reference to `Microsoft.Azure.Devices.Shared` nuget.


abhipsaMisra:
I'd remove "Internal variable rename (https://github.com/Azure/azure-iot-sdk-csharp/pull/3300)" from the release notes - it isn't relevant for customers

patilsnr (approved):

drwill-ms:
> Edge module creation variable name changes

This one has no external impact, so you can remove it from the list of reported changes.

drwill-ms (approved):

iothub/service/src/Common/Extensions/CommonExtensions.cs
```diff
@@ -244,6 +244,16 @@ public static void AppendKeyValuePairIfNotEmpty(this StringBuilder builder, stri
...
 
+        /// <summary>
+        /// Check if the value is null or empty.
+        /// </summary>
+        /// <param name="value">The value to check.</param>
+        /// <returns>Returns true if the value is null or empty, otherwise returns false.</returns>
+        public static bool IsNullOrWhiteSpace(this string value)
+        {
+            return string.IsNullOrWhiteSpace(value);
+        }
```
abhipsaMisra, line 255: Was there anything explained in the related PR on why this was removed? Was it simply unused?

brycewang-microsoft: Here is the related [PR](https://github.com/Azure/azure-iot-sdk-csharp/pull/2801). It appears that we were replacing the custom method `IsNullOrWhiteSpace()` with `string.IsNullOrWhiteSpace()` in this PR, as a result, this public API didn't have any references in SDK any more, so removed.

abhipsaMisra: I see, so this is an extension method that in reality already exists on the `string` class. 

Since the functionality does exist on the `string` class anyway, I'd be ok with us removing this and calling out in our release notes that we removed this duplicated functionality and users should use the "native" API on the string class directly.
Thoughts?

abhipsaMisra: We're removing this public API but users shouldn't be broken by that since they can call `string.IsNullOrWhiteSpace(value)` directly anyway, right?

brycewang-microsoft: Yes, that's right. I will still remove this API and call this out in release note then.

PR #3315, opened 2023-05-01
Update service error codes

drwill-ms:


abhipsaMisra (approved):
Do we need to add these to the device client error code list as well?

Issue #3318, opened 2023-05-02
Unobserved task exception when using MQTT websocket when connection cannot be established


ramondeklein:
## Context
- OS: Windows 10/11 and Linux
- Platforms: ARM and Intel platforms are affected
- .NET Core SDK: v7.0.203
- Microsoft.Azure.Devices.Client package v1.41.3 (although the latest preview is probably also affected, when I look at the code).

## Description of the issue
When using `Mqtt_WebSocket_Only` transport and a host that cannot be reached (i.e. due to DNS issues, because the device is offline) then a `WebSocketException` is thrown that doesn't seem to be catched, because it is picked up by `TaskScheduler.UnobservedTaskException`.

## Code sample exhibiting the issue
```csharp
using System.Runtime;
using Microsoft.Azure.Devices.Client;

// Ensure that task are collected to detect the unobserved task exceptions
// (no production code, but just to trigger the bug)
_ = Task.Run(async () =>
{
    while (true)
    {
        await Task.Delay(TimeSpan.FromSeconds(1)).ConfigureAwait(false);
        GCSettings.LargeObjectHeapCompactionMode = GCLargeObjectHeapCompactionMode.CompactOnce;
        GC.Collect(2, GCCollectionMode.Forced, true, true);
    }
});

// Catch unobserved task exceptions
TaskScheduler.UnobservedTaskException += (_, args) =>
{
    Console.WriteLine($"Caught {(args.Observed ? "observed" : "non-observed")} task exception");
    Console.WriteLine($"Top level [{args.Exception.GetType().Name}]: {args.Exception.Message}\n{args.Exception.StackTrace}");

    foreach (var exc in args.Exception.InnerExceptions)
        Console.WriteLine($"Inner [{exc.GetType().Name}]: {exc.Message}\n{exc.StackTrace}");
};


var deviceId = "invalid-device-id";
var primaryKey = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa=";
var auth = new DeviceAuthenticationWithRegistrySymmetricKey(deviceId, primaryKey);
var assignedHub = "example.com";

await using var deviceClient = DeviceClient.Create(assignedHub, auth, TransportType.Mqtt_WebSocket_Only);
deviceClient.SetConnectionStatusChangesHandler((status, reason) =>
    Console.WriteLine($"Status changed to {status}: {reason}"));
await deviceClient.OpenAsync();
```

[...]

PR #3319, opened 2023-05-02
Log possibly unobserved exceptions in MQTT

drwill-ms:
From #3318

### DIFF
--- a/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
+++ b/iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
@@ -188,19 +188,12 @@ namespace Microsoft.Azure.Devices.Client.Transport.Mqtt
             else
             {
                 ClientOptions options = context.ClientOptions;
-                switch (settings.GetTransportType())
+                _channelFactory = settings.GetTransportType() switch
                 {
-                    case TransportType.Mqtt_Tcp_Only:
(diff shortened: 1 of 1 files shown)


iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -277,37 +270,35 @@ public override async Task<Message> ReceiveAsync(CancellationToken cancellationT
 
                 return null;
             }
-            else
```
drwill-ms, line 280: Removed else block because previous block returns.

iothub/device/src/Transport/Mqtt/MqttTransportHandler.cs
```diff
@@ -944,11 +952,9 @@ public override async Task<Twin> SendTwinGetAsync(CancellationToken cancellation
                     Properties = JsonConvert.DeserializeObject<TwinProperties>(body),
                 };
             }
-            catch (JsonReaderException ex)
+            catch (JsonReaderException ex) when (Logging.IsEnabled)
```
drwill-ms, line 955: Don't enter catch when logging isn't enabled, just to throw again.

timtay-microsoft (approved):

abhipsaMisra (approved):

PR #3320, opened 2023-05-02
add versions.json + support in readme

ryanwinter:


readme.md
```diff
@@ -192,7 +185,25 @@ A couple of examples:
...
-## Long-Term Support (LTS)
+## Support
+
+The Azure IoT Hub Device Client supported releases is outlined in the following table.
+
+Refer to the [Azure IoT Device SDK lifecycle and support](https://learn.microsoft.com/azure/iot/iot-device-sdks-lifecycle-and-support) for details on the different supported stages.
+
+| Release | Category | End-of-life |
+|-|-|-|
+| 1.41.3 | Active | - |
```
drwill-ms, line 196: A new version just shippped.

```suggestion
| 1.42.0 | Active | - |
```

readme.md
```diff
@@ -192,7 +185,25 @@ A couple of examples:
...
+| 1.36.8 | Deprecated (LTS) | 2024-03-28 |
+| 2.0.0-preview004 | Preview | 2023-06-13
```
drwill-ms, line 198: Newer preview out and I'd say previews don't really have an end-of-life as they aren't officially supported anyway.

```suggestion
| 2.0.0-preview005 | Preview | N/A |
```

readme.md
```diff
@@ -192,7 +185,25 @@ A couple of examples:
...
+
+### Need help?
```
drwill-ms, line 200: Blank lines around headers.

```suggestion
### Need help?

```

versions.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "version" : "1.41.3"
```
drwill-ms, line 3:
```suggestion
        "version" : "1.42.0"
```

versions.json
```diff
@@ -0,0 +1,13 @@
...
+    },
+    {
+        "version" : "1.36.8",
+        "eol" : "2024-03-28"
+    },
+    {
+        "version" : "2.0.0--preview004",
```
drwill-ms, line 10:
```suggestion
        "version" : "2.0.0--preview005",
```

versions.json
```diff
@@ -0,0 +1,13 @@
...
+        "eol" : "2023-06-13"
```
drwill-ms, line 11: How do you want to treat this in JSON if not supported?

readme.md
drwill-ms: https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md022---headers-should-be-surrounded-by-blank-lines

versions.json
ryanwinter: Azure recommendations requested json

readme.md
ryanwinter: yeah been a bit back and forward on this. We do have some level on support for previews via github issues, but no production level support which I think you are talking about. But if we leave this blank, or N/A, it sounds to me like we are supporting it forever?

drwill-ms: If we want an actual date for previews, then I'd pick a shorter timespan, like 3-6 months.

readme.md
drwill-ms: Oh, is the idea that we continue to add all new versions so that ones previously in the json continue to stay there, at least until the EOL date?

readme.md
ryanwinter: I agree, I was thinking 3 months. I just updated the date to be 3 months from the preview5 release

readme.md
ryanwinter: Old versions would be deleted when they are out of support.  This would ideally just contain currently supported versions, but with eol it means that it doesnt matter if they are there because it explicitly says its no longer supported.

The eol would be used for:
1. We dont need to update the file when a version goes out of support
2. Azure recommendations can start advising people to update before while the version is in support, so say with 1.42, they could start advising 3 months before support ends.

drwill-ms (approved):

PR #3321, opened 2023-05-02
Increase timeout for registerAsync in provisioning e2e tests

timtay-microsoft:
Previously 10 seconds, now 60 seconds.

A normal provisioning request can take longer than 10 seconds. Plus, if the DPS instance is busy handling the other requests that this test suite makes, then they would take even longer than normal.

Also fix the retry logic in this test

### DIFF
--- a/e2e/Tests/provisioning/ProvisioningE2ETests.cs
+++ b/e2e/Tests/provisioning/ProvisioningE2ETests.cs
@@ -708,7 +708,7 @@ namespace Microsoft.Azure.Devices.E2ETests.Provisioning
                         VerboseTestLogger.WriteLine($"ProvisioningDeviceClient.RegisterAsync failed because: {ex.Message}");
                         await Task.Delay(TimeSpan.FromSeconds(2)).ConfigureAwait(false);
                     }
-                    catch (OperationCanceledException oce) when (overallCts.IsCancellationRequested)
+                    catch (OperationCanceledException oce) when (!overallCts.IsCancellationRequested)
                     {
                         // This catch statement shouldn't execute when the test itself is cancelled, but will
                         // execute when the registerAsync(cts) call times out 
--- a/e2e/Tests/provisioning/ProvisioningE2ETests.cs
+++ b/e2e/Tests/provisioning/ProvisioningE2ETests.cs
@@ -25,7 +25,8 @@ namespace Microsoft.Azure.Devices.E2ETests.Provisioning
     [TestCategory("DPS")]
     public class ProvisioningE2ETests : E2EMsTestBase
     {
-        private const int PassingTimeoutMilliseconds = 60 * 1000;
+        private const int RegisterWithRetryTimeoutMilliseconds = 3 * 60 * 1000;
+        private const int RegisterTimeoutMilliseconds = 60 * 1000;
         private const int FailingTimeoutMilliseconds = 10 * 1000;
         private const int MaxTryCount = 10;
         private const string InvalidIdScope = "0neFFFFFFFF";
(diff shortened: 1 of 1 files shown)


e2e/Tests/provisioning/ProvisioningE2ETests.cs
```diff
@@ -686,12 +687,12 @@ public async Task DPS_Registration_AmqpWs_X509_IndividualEnrollment_InvalidGloba
 
             try
             {
-                using var overallCts = new CancellationTokenSource(PassingTimeoutMilliseconds);
+                using var overallCts = new CancellationTokenSource(RegisterWithRetryTimeoutMilliseconds);
 
                 // Trying to register simultaneously can cause conflicts (409). Retry in those scenarios to succeed.
                 while (true)
                 {
-                    using var attemptCts = new CancellationTokenSource(FailingTimeoutMilliseconds);
```
timtay-microsoft, line 694: ```FailingTimeoutMilliseconds``` was being used in cases where the ```RegisterAsync() ``` call was meant to fail (invalid ID scope). To differentiate these two cases, I created a new field with a longer timeout (60 seconds)

e2e/Tests/provisioning/ProvisioningE2ETests.cs
```diff
@@ -686,12 +687,12 @@ public async Task DPS_Registration_AmqpWs_X509_IndividualEnrollment_InvalidGloba
...
+                    using var attemptCts = new CancellationTokenSource(RegisterTimeoutMilliseconds);
                     using var opCts = CancellationTokenSource.CreateLinkedTokenSource(overallCts.Token, attemptCts.Token);
```
abhipsaMisra, line 696: since opCts will cancel when either overallCts or attemptCts cancels, it is limited by the cancellation token with the shortest timeout. Then, since we own both cancellation tokens, why not pass the shortest duration one directly? Is someone else signalling them for cancellation?

e2e/Tests/provisioning/ProvisioningE2ETests.cs
```diff
@@ -707,7 +708,7 @@ public async Task DPS_Registration_AmqpWs_X509_IndividualEnrollment_InvalidGloba
                         VerboseTestLogger.WriteLine($"ProvisioningDeviceClient.RegisterAsync failed because: {ex.Message}");
                         await Task.Delay(TimeSpan.FromSeconds(2)).ConfigureAwait(false);
                     }
-                    catch (OperationCanceledException oce) when (overallCts.IsCancellationRequested)
+                    catch (OperationCanceledException oce) when (!overallCts.IsCancellationRequested)
```
timtay-microsoft, line 711: Previously, this would only catch TaskCanceledException/OperationCanceledException if the total retry timeout had passed. So the first timeout would not be caught, and the test would fail. Now this test will retry on a timeout up until the retry timeout of 3 minutes.

e2e/Tests/provisioning/ProvisioningE2ETests.cs
timtay-microsoft: > why not pass the shortest duration one directly?

Is there a way to compare cancellation tokens like that? If so, we could do that. If not, then this is the only way I know to cancel an operation when one or more tokens are canceled

timtay-microsoft: > Is someone else signalling them for cancellation?

No, just the timeout value it is constructed with will signal each of these tokens

drwill-ms: > since opCts will cancel when either overallCts or attemptCts cancels, it is limited by the cancellation token with the shortest timeout. Then, since we own both cancellation tokens, why not pass the shortest duration one directly? Is someone else signalling them for cancellation?

Because we can still retry if the original response failed to get back to our client for some reason. If we just have 1 cancellation token then you get one shot and that is it.

drwill-ms (approved):

PR #3322, opened 2023-05-02
Fix twin desired property payload conversion - MQTT

tmahmood-microsoft:


iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1158,7 +1158,7 @@ private void HandleReceivedDesiredPropertiesUpdateRequest(MqttApplicationMessage
             // This message is always QoS 0, so no ack will be sent.
             receivedEventArgs.AutoAcknowledge = true;
 
-            Dictionary<string, object> desiredPropertyPatchDictionary = _payloadConvention.GetObject<Dictionary<string, object>>(receivedEventArgs.ApplicationMessage.Payload);
```
timtay-microsoft, line 1161: Was ```_payloadConvention``` null here or something? Why do we want to always use the default payload convention?

tmahmood-microsoft: Since the payload from service client is always converted using default Payload.

timtay-microsoft: Sounds good then. Leave a comment to that effect here for future reference.

timtay-microsoft (approved):

iothub/device/src/Pipeline/MqttTransportHandler.cs
abhipsaMisra: +1 to adding a comment in here explaining why default convention is used. You could reference the fact that we deserialize this into a dictionary based on service defined convention of it being a utf-8 encoded json string.

abhipsaMisra (approved):

iothub/device/src/Pipeline/MqttTransportHandler.cs
drwill-ms: > Since the payload from service client is always converted using default Payload/Newtonsoft.Json

Not quite. However it was sent from the service client, it is just JSON.

The problem is that our SDK deserializes the twin using Newtonsoft.Json. Then, individual properties can be deserialized to the user's choice, but not until after we've converted them back to bytes using the default payload convention again.

drwill-ms (approved):

PR #3325, opened 2023-05-03
Add module-module method long haul tests

abhipsaMisra:
This currently invokes the method on itself (similar to our sample). We can expand it to work with a different edge module in the future.

### DIFF
--- a/e2e/LongHaul/module/IotHub.cs
+++ b/e2e/LongHaul/module/IotHub.cs
@@ -37,7 +37,9 @@ namespace Microsoft.Azure.Devices.LongHaul.Module
         private long _totalDesiredPropertiesToModuleHandled;
         private long _totalM2mMessagesCompleted;
         private long _totalM2mMessagesRejected;
-        private long _totalMethodCallsToModuleCount;
+        private long _totalMethodCallsServiceToModuleCount;
+        private long _totalMethodCallsModuleToModuleSentCount;
+        private long _totalMethodCallsModuleToModuleReceivedCount;
 
         public IotHub(Logger logger, Parameters parameters)
         {
@@ -259,6 +261,58 @@ namespace Microsoft.Azure.Devices.LongHaul.Module
             }
         }
 
+        public async Task InvokeDirectMethodOnLeafClientThroughEdgeAsync(string deviceId, string moduleId, string methodName, Logger logger, CancellationToken ct)
+        {
+            Debug.Assert(_moduleClient != null);
+            Debug.Assert(deviceId != null);
+            Debug.Assert(moduleId != null);
+
+            var sw = new Stopwatch();
+            while (!ct.IsCancellationRequested)
+            {
+                try
+                {
+                    var payload = new CustomDirectMethodPayload
+                    {
+                        RandomId = Guid.NewGuid(),
+                        CurrentTimeUtc = DateTimeOffset.UtcNow,
+                        MethodCallsCount = ++_totalMethodCallsModuleToModuleSentCount,
+                    };
+
+                    var directMethodRequest = new EdgeModuleDirectMethodRequest(methodName, payload)
+                    {
+                        ResponseTimeout = TimeSpan.FromSeconds(30),
+                    };
+
+                    logger.Trace($"Invoking direct method for edge device: {deviceId}, edge module: {moduleId}", TraceSeverity.Information);
+                    logger.Metric(TotalDirectMethodCallsModuleToModuleSentCount, _totalMethodCallsModuleToModuleSentCount);
+
+                    sw.Restart();
+
+                    // Invoke the direct method asynchronously and get the response from the simulated module.
+                    DirectMethodResponse response = await _moduleClient
+                        .InvokeMethodAsync(deviceId, moduleId, directMethodRequest, ct)
+                        .ConfigureAwait(false);
+                    logger.Metric(DirectMethodModuleToModuleRoundTripSeconds, sw.Elapsed.TotalSeconds);
+
+                    if (response.TryGetPayload(out CustomDirectMethodPayload responsePayload))
+                    {
+                        logger.Metric(
+                            DirectMethodModuleToModuleDelaySeconds,
+                            (DateTimeOffset.UtcNow - responsePayload.CurrentTimeUtc).TotalSeconds);
+                    }
+
+                    logger.Trace($"Response status: {response.Status}, method name:{methodName}", TraceSeverity.Information);
+                    break;
+                }
+                catch (Exception ex)
+                {
+                    logger.Trace($"Exception invoking direct method from an edge module\n{ex}", TraceSeverity.Warning);
+                    await Task.Delay(s_retryInterval, ct).ConfigureAwait(false);
+                }
+            }
+        }
+
         public async ValueTask DisposeAsync()
         {
             _logger.Trace($"Disposing the {nameof(IotHub)} instance", TraceSeverity.Verbose);
@@ -372,7 +426,30 @@ namespace Microsoft.Azure.Devices.LongHaul.Module
                             _logger.Metric(
                                 C2mDirectMethodDelaySeconds,
                                 (DateTimeOffset.UtcNow - methodPayload.CurrentTimeUtc).TotalSeconds);
-                            _logger.Metric(TotalDirectMethodCallsModuleToModuleCount, ++_totalMethodCallsToModuleCount);
+                            _logger.Metric(TotalDirectMethodCallsServiceToModuleCount, ++_totalMethodCallsServiceToModuleCount);
+
+                            // Log the current time again and send the response back to the service app.
+                            methodPayload.CurrentTimeUtc = DateTimeOffset.UtcNow;
+                            return Task.FromResult(new DirectMethodResponse(200) { Payload = methodPayload });
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.Trace($"Failed to parse the payload for direct method {methodRequest.MethodName} due to {ex}.", TraceSeverity.Error);
+                        return Task.FromResult(new DirectMethodResponse(400) { Payload = ex.Message });
+                    }
+                    break;
+
+                case "ModuleToItself":
+                    try
+                    {
+                        if (methodRequest.TryGetPayload(out CustomDirectMethodPayload methodPayload))
+                        {
+                            _logger.Trace($"Echoing back the payload of direct method.", TraceSeverity.Information);
+                            _logger.Metric(
+                                C2mDirectMethodDelaySeconds,
+                                (DateTimeOffset.UtcNow - methodPayload.CurrentTimeUtc).TotalSeconds);
+                            _logger.Metric(TotalDirectMethodCallsModuleToModuleReceivedCount, ++_totalMethodCallsModuleToModuleReceivedCount);
 
                             // Log the current time again and send the response back to the service app.
                             methodPayload.CurrentTimeUtc = DateTimeOffset.UtcNow;
--- a/e2e/LongHaul/module/IotHub.cs
+++ b/e2e/LongHaul/module/IotHub.cs
@@ -276,7 +276,6 @@ namespace Microsoft.Azure.Devices.LongHaul.Module
                     {
                         RandomId = Guid.NewGuid(),
                         CurrentTimeUtc = DateTimeOffset.UtcNow,
-                        MethodCallsCount = ++_totalMethodCallsModuleToModuleSentCount,
                     };
 
                     var directMethodRequest = new EdgeModuleDirectMethodRequest(methodName, payload)
@@ -285,14 +284,14 @@ namespace Microsoft.Azure.Devices.LongHaul.Module
                     };
 
                     logger.Trace($"Invoking direct method for edge device: {deviceId}, edge module: {moduleId}", TraceSeverity.Information);
-                    logger.Metric(TotalDirectMethodCallsModuleToModuleSentCount, _totalMethodCallsModuleToModuleSentCount);
-
                     sw.Restart();
 
                     // Invoke the direct method asynchronously and get the response from the simulated module.
                     DirectMethodResponse response = await _moduleClient
                         .InvokeMethodAsync(deviceId, moduleId, directMethodRequest, ct)
                         .ConfigureAwait(false);
+
+                    logger.Metric(TotalDirectMethodCallsModuleToModuleSentCount, ++_totalMethodCallsModuleToModuleSentCount);
                     logger.Metric(DirectMethodModuleToModuleRoundTripSeconds, sw.Elapsed.TotalSeconds);
 
                     if (response.TryGetPayload(out CustomDirectMethodPayload responsePayload))
--- a/e2e/LongHaul/module/IotHub.cs
+++ b/e2e/LongHaul/module/IotHub.cs
@@ -262,11 +262,12 @@ namespace Microsoft.Azure.Devices.LongHaul.Module
             }
         }
 
-        public async Task InvokeDirectMethodOnLeafClientThroughEdgeAsync(string deviceId, string moduleId, string methodName, Logger logger, CancellationToken ct)
+        public async Task InvokeDirectMethodOnLeafClientAsync(string deviceId, string moduleId, string methodName, Logger logger, CancellationToken ct)
         {
             Debug.Assert(_moduleClient != null);
             Debug.Assert(deviceId != null);
             Debug.Assert(moduleId != null);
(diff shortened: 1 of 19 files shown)


e2e/LongHaul/module/Program.cs
```diff
@@ -88,12 +88,20 @@ private static async Task Main(string[] args)
...
-                        iotHub.ReportReadOnlyPropertiesAsync(s_logger.Clone(), cts.Token))
-                    .ConfigureAwait(false);
+                var tasksToRun = new List<Task>
+                {
+                    systemHealthMonitor.RunAsync(cts.Token),
+                    iotHub.SendMessagesToRouteAsync(s_logger.Clone(), cts.Token),
+                    iotHub.ReportReadOnlyPropertiesAsync(s_logger.Clone(), cts.Token)
+                };
+
+                if (parameters.GatewayHostName != null)
```
abhipsaMisra, line 98: Conditional run for module-module method invocation tests

iothub/device/src/DirectMethod/DirectMethodResponse.cs
```diff
@@ -39,7 +41,89 @@ public DirectMethodResponse(int status)
...
+        ///     // Get the payload as bytes
+        ///     byte[] arr = directMethodRequest.GetPayloadAsBytes();
+        ///
+        ///     // deserialize as needed and do work...
+        /// }
+        ///
+        /// // ...
+        /// </code>
+        /// </example>
+        public bool TryGetPayload<T>(out T payload)
```
abhipsaMisra, line 84: Previously only the service client had `TryGet...()` for the direct method response. We had not added them on the client side since a client would create and send the direct method response, so it didn't need helpers for getting the payload out.
However, since the module client is also expected to be working with direct method response (on the receiving end), I'm adding these methods in.

patilsnr (approved):
optimistic approval

e2e/LongHaul/module/IotHub.cs
```diff
@@ -259,6 +261,58 @@ public async Task SetPropertiesAsync(string keyName, object properties, Logger l
...
+            var sw = new Stopwatch();
+            while (!ct.IsCancellationRequested)
+            {
+                try
+                {
+                    var payload = new CustomDirectMethodPayload
+                    {
+                        RandomId = Guid.NewGuid(),
+                        CurrentTimeUtc = DateTimeOffset.UtcNow,
+                        MethodCallsCount = ++_totalMethodCallsModuleToModuleSentCount,
```
drwill-ms, line 279: The increment here may be premature. If we fail due to some exception and try again, we aren't trying with the same payload and increment yet again.

e2e/LongHaul/module/IotHub.cs
```diff
@@ -259,6 +261,58 @@ public async Task SetPropertiesAsync(string keyName, object properties, Logger l
...
+                    };
+
+                    var directMethodRequest = new EdgeModuleDirectMethodRequest(methodName, payload)
+                    {
+                        ResponseTimeout = TimeSpan.FromSeconds(30),
+                    };
+
+                    logger.Trace($"Invoking direct method for edge device: {deviceId}, edge module: {moduleId}", TraceSeverity.Information);
+                    logger.Metric(TotalDirectMethodCallsModuleToModuleSentCount, _totalMethodCallsModuleToModuleSentCount);
```
drwill-ms, line 288: This is also a premature metric, given we haven't successfully sent it yet, right?

abhipsaMisra: You're right. I moved both to be logged after the method invocation has completed.

e2e/LongHaul/module/IotHub.cs
```diff
@@ -259,6 +261,58 @@ public async Task SetPropertiesAsync(string keyName, object properties, Logger l
...
+                            (DateTimeOffset.UtcNow - responsePayload.CurrentTimeUtc).TotalSeconds);
+                    }
+
+                    logger.Trace($"Response status: {response.Status}, method name:{methodName}", TraceSeverity.Information);
+                    break;
+                }
+                catch (Exception ex)
+                {
+                    logger.Trace($"Exception invoking direct method from an edge module\n{ex}", TraceSeverity.Warning);
+                    await Task.Delay(s_retryInterval, ct).ConfigureAwait(false);
```
drwill-ms, line 310: Delay should be after the catch so both happy and error paths wait before performing the operation?

e2e/LongHaul/module/IotHub.cs
```diff
@@ -259,6 +261,57 @@ public async Task SetPropertiesAsync(string keyName, object properties, Logger l
             }
         }
 
+        public async Task InvokeDirectMethodOnLeafClientThroughEdgeAsync(string deviceId, string moduleId, string methodName, Logger logger, CancellationToken ct)
+        {
+            Debug.Assert(_moduleClient != null);
+            Debug.Assert(deviceId != null);
+            Debug.Assert(moduleId != null);
+
```
drwill-ms, line 269: Should customize the logger custom properties by adding what operation this logger is for.

iothub/device/src/DirectMethod/DirectMethodResponse.cs
```diff
@@ -39,7 +41,89 @@ public DirectMethodResponse(int status)
...
+        ///     // do work
+        ///     // ...
+        /// }
+        /// else
+        /// {
+        ///     // Get the payload as bytes
+        ///     byte[] arr = directMethodRequest.GetPayloadAsBytes();
+        ///
+        ///     // deserialize as needed and do work...
+        /// }
```
drwill-ms, line 79: I don't think someone would try to deserialize as a type or bytes. I get that we want to show one can get the payload different ways, but this feels odd.

I'd suggest leaving out the GetPayloadAsBytes example, or simply comment that if the type isn't JSON, then to use this method, but as you already have that covered in remarks, I don't see the point.

iothub/device/src/DirectMethod/DirectMethodResponse.cs
```diff
@@ -39,7 +41,89 @@ public DirectMethodResponse(int status)
...
+        /// // ...
+        /// </code>
+        /// </example>
+        public bool TryGetPayload<T>(out T payload)
+        {
+            payload = default;
+
+            try
+            {
+                payload = PayloadConvention.GetObject<T>(GetPayloadAsBytes());
```
drwill-ms, line 90: Why not just:

```suggestion
                payload = PayloadConvention.GetObject<T>(Payload);
```

iothub/device/src/DirectMethod/DirectMethodResponse.cs
drwill-ms: Update the migration guide?

e2e/LongHaul/module/IotHub.cs
```diff
@@ -259,6 +261,57 @@ public async Task SetPropertiesAsync(string keyName, object properties, Logger l
...
+                    };
+
+                    logger.Trace($"Invoking direct method for edge device: {deviceId}, edge module: {moduleId}", TraceSeverity.Information);
+                    sw.Restart();
+
+                    // Invoke the direct method asynchronously and get the response from the simulated module.
+                    DirectMethodResponse response = await _moduleClient
+                        .InvokeMethodAsync(deviceId, moduleId, directMethodRequest, ct)
+                        .ConfigureAwait(false);
+
```
tmahmood-microsoft, line 293: missing timer stop statement

tmahmood-microsoft (approved):

brycewang-microsoft (approved):

iothub/device/src/DirectMethod/DirectMethodResponse.cs
abhipsaMisra: GetObject<T>() requires the input to be bytes. Payload is of type `object`.

`public abstract T GetObject<T>(byte[] objectToConvert);`

drwill-ms (approved):

iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -28,86 +29,32 @@ internal DirectMethodRequest()
...
+        protected internal DirectMethodRequest(string methodName, byte[] payload = default)
         {
             MethodName = methodName;
+            _payload = payload;
         }
 
         /// <summary>
         /// The method name to invoke.
         /// </summary>
-        [JsonProperty("methodName")]
```
abhipsaMisra, line 39: json property is not relevant for a device method

iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -28,86 +29,32 @@ internal DirectMethodRequest()
...
-        /// If the value is greater than zero, it may also occur if the cloud fails to deliver the request to
-        /// the target device.
-        /// <para>
-        /// This value is propagated to the service in terms of seconds, so this value does not have a level of
-        /// precision below seconds. For example, a value of <c>TimeSpan.FromMilliseconds(500)</c> will be
-        /// interpreted as 0 seconds (using <c>ConnectTimeout.TotalSeconds</c>).
-        /// </para>
-        /// </remarks>
-        [JsonIgnore]
-        public TimeSpan? ConnectionTimeout { get; set; }
```
abhipsaMisra, line 57: These timeouts are never set by the SDK for a device method request received from service

iothub/device/src/DirectMethod/EdgeModuleDirectMethodRequest.cs
```diff
@@ -1,13 +1,18 @@
...
 
+using System;
+using Newtonsoft.Json;
+
 namespace Microsoft.Azure.Devices.Client
 {
     /// <summary>
     /// Parameters to execute a direct method on an edge device or an edge module by an <see cref="IotHubModuleClient"/>.
     /// </summary>
-    public class EdgeModuleDirectMethodRequest : DirectMethodRequest
```
abhipsaMisra, line 9: These classes seem sufficiently different so justify separating them out.
- DirectMethodRequest is always received by the client. This can be the result of a method invocation request from service, or from an edge module
- EdgeModuleDirectMethodRequest is always sent by an edge module client. This is never received by the client.

Since one is an outgoing type and another an incoming type, they inherently require different methods.
The outgoing type is required to be able to set the payload as either an object (to be serialized by the convention during _send_. It does not require TryGet methods for fetching the payload value since the payload is set by the user themselves.
The incoming type requires the payload to be set as raw bytes since that is what we get from the transport layer. It requires TryGet methods so that the payload can be deserialized according to the convention.

Thoughts?

iothub/device/src/DirectMethod/EdgeModuleDirectMethodRequest.cs
```diff
@@ -37,11 +42,76 @@ public EdgeModuleDirectMethodRequest(string methodName, byte[] payload)
...
-                ? null
-                : PayloadConvention.GetObjectBytes(payload);
+            MethodName = methodName;
+            _payload = payload;
         }
+
+        /// <summary>
+        /// The method name to invoke.
+        /// </summary>
+        [JsonProperty("methodName")]
```
abhipsaMisra, line 53: This type is serialized into json by the transport layer.

iothub/device/src/DirectMethod/EdgeModuleDirectMethodRequest.cs
```diff
@@ -37,11 +42,76 @@ public EdgeModuleDirectMethodRequest(string methodName, byte[] payload)
...
+        [JsonIgnore]
+        internal PayloadConvention PayloadConvention { get; set; }
+
+        /// <summary>
+        /// The direct method payload.
+        /// </summary>
+        [JsonProperty("payload", NullValueHandling = NullValueHandling.Include)]
+        internal byte[] Payload => _payload is byte[] payloadAsByteArray
+            ? payloadAsByteArray
+            : PayloadConvention.GetObjectBytes(Payload);
```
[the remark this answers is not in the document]
abhipsaMisra: I have this property as internal. Let me know if you think users would benefit from having visibility to this property. (they set the payload via the constructor).

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -489,13 +489,10 @@ await deviceClient
...
                 .ConfigureAwait(false);
 
-            var DirectMethodRequest = new DirectMethodRequest("TestMethodName")
+            var DirectMethodRequest = new DirectMethodRequest("TestMethodName", Encoding.UTF8.GetBytes("Json"))
             {
                 PayloadConvention = DefaultPayloadConvention.Instance,
                 RequestId = "request",
-                Payload = Encoding.UTF8.GetBytes("Json"),
-                ResponseTimeout = TimeSpan.FromSeconds(5),
-                ConnectionTimeout = TimeSpan.FromSeconds(5),
```
abhipsaMisra, line 498: These timeouts were never set on the DirectMethodRequest we give to users

iothub/device/tests/IotHubDeviceClientTests.cs
```diff
@@ -614,16 +560,13 @@ await deviceClient
...
                         responseReceivedAsExpected = payload.TryGetPayload(out response);
-                        int? connectionTimeout = payload.ConnectionTimeoutInSeconds;
-                        int? responseTimeout = payload.ResponseTimeoutInSeconds;
                         return Task.FromResult(_directMethodResponseWithPayload);
                     })
                 .ConfigureAwait(false);
 
             const string payload = "test";
-            var directMethodRequest = new DirectMethodRequest("TestMethodName")
+            var directMethodRequest = new DirectMethodRequest("TestMethodName", DefaultPayloadConvention.Instance.GetObjectBytes(payload))
```
abhipsaMisra, line 568: Instead of referencing newtonsoft directly, using the default convention to get the bytes

iothub/device/src/DirectMethod/EdgeModuleDirectMethodRequest.cs
```diff
@@ -23,25 +26,76 @@ public EdgeModuleDirectMethodRequest(string methodName)
         /// a direct method on an edge device or an edge module connected to the same edge hub.
         /// </summary>
         /// <param name="methodName">The method name to invoke.</param>
-        /// <param name="payload">The direct method payload bytes.</param>
-        public EdgeModuleDirectMethodRequest(string methodName, byte[] payload)
-            : base(methodName)
+        /// <param name="payload">The direct method payload that will be serialized using <see cref="DefaultPayloadConvention"/>.</param>
```
abhipsaMisra, line 29: The EdgeModuleDirectMethodRequest class itself needs to be serialized using newtonsoft, so having a nested custom serializable property (Payload) within it is cumbersome to implement.
We're going to start off by restricting the usage of this to newtonsoft serializable payloads, but we can decide to expand support in the future if we like.
(This is the only type on the client side that has this kind of a "nested" serialization requirement).

iothub/device/src/DirectMethod/EdgeModuleDirectMethodRequest.cs
```diff
@@ -23,25 +26,76 @@ public EdgeModuleDirectMethodRequest(string methodName)
...
-        /// <param name="payload">The direct method payload that will be serialized and encoded per <see cref="IotHubClientOptions.PayloadConvention"/>.</param>
-        public EdgeModuleDirectMethodRequest(string methodName, object payload)
-            : base(methodName)
-        {
-            Payload = payload == null
-                ? null
-                : PayloadConvention.GetObjectBytes(payload);
-        }
+        [JsonProperty("methodName")]
+        public string MethodName { get; private set; }
```
drwill-ms, line 40: why private set and not readonly?

tmahmood-microsoft (approved):

iothub/device/src/DirectMethod/EdgeModuleDirectMethodRequest.cs
abhipsaMisra: Newtonsoft does persist readonly properties during serializtion by default; however, someone might unintentionally override that behavior by setting "JsonSerializerOptions.IgnoreReadOnlyProperties = true" on the serializer options set by their application (they might want to do that for their custom json object behavior).
That would be the only reason why I'd want us to leave this private setter as-is.

What do you think? @drwill-ms

drwill-ms: I'm not sure we employ that logic with other models, but I see your point. I'd simply suggest we be consistent across our models.

abhipsaMisra: You're right, our current approach has been to use readonly properties. I can update this to match the other models and we decide to change our approach if we run into an issue in the future. (It is a private set, so it wouldn't be a breaking change).

PR #3326, opened 2023-05-04
Use gateway hostname from options for connection string flow

abhipsaMisra:
The GatewayHostName set in IotHubClientOptions was being ignored for the connection string flow.
This PR updates our SDK to consider this value even when connection strings are used. If a gateway hostname is already specified in the connection string directly, then the value set through options is ignored.

### DIFF
--- a/iothub/device/src/IotHubClientOptions.cs
+++ b/iothub/device/src/IotHubClientOptions.cs
@@ -85,6 +85,10 @@ namespace Microsoft.Azure.Devices.Client
         /// <para>
         /// It can also be used for other, custom transparent or protocol gateways.
         /// </para>
+        /// <para>
+        /// If the client uses a connection string badsed authentication mechanism, and has a gateway hostname specified in the connection string
+        /// then the value set in options will be ignored.
+        /// </para>
         /// </remarks>
         public string GatewayHostName { get; set; }
 
--- a/iothub/device/src/IotHubClientOptions.cs
+++ b/iothub/device/src/IotHubClientOptions.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Azure.Devices.Client
         /// It can also be used for other, custom transparent or protocol gateways.
         /// </para>
         /// <para>
-        /// If the client uses a connection string based authentication mechanism, and has a gateway hostname specified in the connection string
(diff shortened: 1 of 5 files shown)


drwill-ms:
> The GatewayHostName set in IotHubClientOptions was being ignored for the connection string flow. This PR updates our SDK to consider this value even when connection strings are used. If a gateway hostname is already specified in the connection string directly, then the value set through options is ignored.

Should the connection string values override client options or the other way around?

iothub/device/src/IotHubClientOptions.cs
```diff
@@ -85,6 +85,10 @@ public IotHubClientOptions(IotHubClientTransportSettings transportSettings)
         /// <para>
         /// It can also be used for other, custom transparent or protocol gateways.
         /// </para>
+        /// <para>
+        /// If the client uses a connection string based authentication mechanism, and has a gateway hostname specified in the connection string
```
drwill-ms, line 89:
```suggestion
        /// If the client uses a connection string-based authentication mechanism, and has a gateway hostname specified in the connection string
```

patilsnr (approved):

abhipsaMisra:
> > The GatewayHostName set in IotHubClientOptions was being ignored for the connection string flow. This PR updates our SDK to consider this value even when connection strings are used. If a gateway hostname is already specified in the connection string directly, then the value set through options is ignored.
> 
> Should the connection string values override client options or the other way around?

This is the way I see it - the connection string is the most client specific argument, i.e. it is credential specific to this client instance. So, whatever value I specify in there, that should take precedence.
So, I might have a "common" options instance in my application that I might want to share across different client instances (with different identities), but the connection string I pass in will be specific to a particular client identity's instance I want to work with.

What do you think?

tmahmood-microsoft (approved):
Looks good to me.

brycewang-microsoft (approved):

drwill-ms (approved):

abhipsaMisra:
I did merge this PR in already, but I synced up with David now and we're going to take it up as a parking lot item as a team for our next scrum. 
We're going to persist the end resultant behavior, i.e. gateway hostname from client options should be considered even when the client is initialized using a connection string; however, the discussion is open on if the gateway hostname in the client options should take precedence over one in the connection string, or vice versa (for the case where both of them have different gateway hostnames specified).

PR #3329, opened 2023-05-05
Remove default remote certificate callback

timtay-microsoft:
The ClientWebSocket instance used by our MQTT/AMQP libraries already have a default remote certificate validation callback, so there is no need for us to define it as well.

In addition, now that we aren't passing in a default value for this field every time, we don't have to worry about the .NET 472 case that #3328 had to work around.

IoT hub device client already had this behavior, but now the provisioning device client and the IoT hub service client do as well. This is why we only saw these .NET 472 test failures in our DPS tests (since no Hub service client operations use MQTT)

### DIFF
--- a/iothub/service/src/IotHubServiceClientOptions.cs
+++ b/iothub/service/src/IotHubServiceClientOptions.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Azure.Devices
         /// does not support this feature.
         /// </para>
         /// </remarks>
-        public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; } = DefaultRemoteCertificateValidation;
+        public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; }
 
         /// <summary>
         /// Sets the retry policy used in the operation retries.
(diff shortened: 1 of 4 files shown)


abhipsaMisra:
Should we remove the check that was added in #3328 ?

drwill-ms:
> The ClientWebSocket instance used by our MQTT/AMQP libraries already have a default remote certificate validation callback, so there is no need for us to define it as well. Simply leaving this field as null will make these libraries use their default implementation.
> 
> In addition, now that we aren't passing in a default value for this field every time, we don't have to worry about the .NET 472 case that #3328 had to work around.
> 
> IoT hub device client already had this behavior, but now the provisioning device client and the IoT hub service client do as well. This is why we only saw these .NET 472 test failures in our DPS tests (since no Hub service client operations use MQTT)

What about HTTP calls?

drwill-ms (approved):

patilsnr (approved):

timtay-microsoft:
> > The ClientWebSocket instance used by our MQTT/AMQP libraries already have a default remote certificate validation callback, so there is no need for us to define it as well. Simply leaving this field as null will make these libraries use their default implementation.
> > In addition, now that we aren't passing in a default value for this field every time, we don't have to worry about the .NET 472 case that #3328 had to work around.
> > IoT hub device client already had this behavior, but now the provisioning device client and the IoT hub service client do as well. This is why we only saw these .NET 472 test failures in our DPS tests (since no Hub service client operations use MQTT)
> 
> What about HTTP calls?

We currently don't allow users to set this option 

> Should we remove the check that was added in #3328 ?

Yes. Looks like I forgot to commit that, so I've added that change now

Issue #3335, opened 2023-05-30
[Bug Report] [v2] Opened connection on a device consumes my message


bastyuchenko:
Hello
I implemented communication between a device and a backend server. 
I use [previews/v2](https://github.com/Azure/azure-iot-sdk-csharp/tree/previews/v2) SDK

My code for a device:
```csharp
static async Task Main(string[] args)
        {
            X509Certificate2 x509Certificate = Helper.LoadProvisioningCertificate();
            var security = new AuthenticationProviderX509(x509Certificate);

            var provClient = new ProvisioningDeviceClient(
                "global.azure-devices-provisioning.net",
                 Helper.IdScope,
                security,
                options: new ProvisioningClientOptions(new ProvisioningClientMqttSettings()));

            var result = await provClient.RegisterAsync();

            IAuthenticationMethod auth = new ClientAuthenticationWithX509Certificate(x509Certificate, security.GetRegistrationId());

            var deviceClient = new IotHubDeviceClient(result.AssignedHub, auth, new IotHubClientOptions(new IotHubClientMqttSettings()));
            deviceClient.ConnectionStatusChangeCallback = (ConnectionStatusInfo info) =>
            {
                Console.WriteLine($"Device Status {info.Status}; Change reason: {info.ChangeReason}; Recommended action {info.RecommendedAction}.");
            };
            await deviceClient!.OpenAsync(CancellationToken.None);

            await Task.Delay(-1);
        }
```

Device (uses Self-signed X509 Certificate Auth type, MQTT protocol for communication)
Backend service (uses AMQP protocol for communication)

Steps to reproduce:
1. Device: is attested in DPS
2. Device: connects (create client and open connection) to assigned IoT Hub (from step 1)
3. Backend service: sends a message
4. **Expected result:** the message is stored in IoT Hub and is waiting until the device starts a listener (deviceClient.SetIncomingMessageCallbackAsync(...))
**Actual result:** Backend service: receives message feedback.
```
{
  "EnqueuedOnUtc": "2023-05-29T20:49:52.395+00:00",
  "Records": [
    {
      "originalMessageId": "b2db3ca4-6f37-4b27-bd25-5c1d90a61943",
      "deviceGenerationId": "638156339749833913",
      "deviceId": "iothubx509device1",
      "enqueuedTimeUtc": "2023-05-29T20:49:48.2826815+00:00",
      "statusCode": "Success",
      "description": "Success"
    }
  ]
}
```

[...]

PR #3336, opened 2023-06-05
[v2 - bug fix] Queue MQTT C2D messages if callback not set

tmahmood-microsoft:
Fix for https://github.com/Azure/azure-iot-sdk-csharp/issues/3335

In v2, if incoming message callback has not been set, messages are still received because cleanSession = false (default) and the client subscribes to topic from last session. This causes messages to be received but the callback has not been set and the message is eventually completed, since MQTT messages cannot be rejected or abandoned. 
This PR adds changes where we queue the received messages if callback has not been set and processes them later when callback has been set.

### DIFF
--- a/e2e/Tests/iothub/device/IncomingMessageCallbackE2eTests.cs
+++ b/e2e/Tests/iothub/device/IncomingMessageCallbackE2eTests.cs
@@ -107,14 +107,14 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
 
             // Now, send a message to the device from the service.
             OutgoingMessage firstMsg = OutgoingMessageHelper.ComposeTestMessage(out string _, out string _);
-            deviceHandler.ExpectedOutgoingMessage = firstMsg;
             await serviceClient.Messages.SendAsync(testDevice.Id, firstMsg, ct).ConfigureAwait(false);
             VerboseTestLogger.WriteLine($"Sent C2D message from service, messageId={firstMsg.MessageId} - to be received on callback");
 
             // re-open the client under test.
             await deviceClient.OpenAsync(ct).ConfigureAwait(false);
-            Thread.Sleep(20000);
+            await Task.Delay(10000, ct).ConfigureAwait(false);
             await deviceHandler.SetIncomingMessageCallbackHandlerAndCompleteMessageAsync<string>(ct).ConfigureAwait(false);
+            deviceHandler.ExpectedOutgoingMessage = firstMsg;
 
             // Now, set a callback on the device client to receive C2D messages.
             await deviceHandler.WaitForIncomingMessageCallbackAsync(ct).ConfigureAwait(false);
--- a/e2e/Tests/iothub/device/IncomingMessageCallbackE2eTests.cs
+++ b/e2e/Tests/iothub/device/IncomingMessageCallbackE2eTests.cs
@@ -112,9 +112,8 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
 
             // re-open the client under test.
             await deviceClient.OpenAsync(ct).ConfigureAwait(false);
-            await Task.Delay(10000, ct).ConfigureAwait(false);
-            await deviceHandler.SetIncomingMessageCallbackHandlerAndCompleteMessageAsync<string>(ct).ConfigureAwait(false);
             deviceHandler.ExpectedOutgoingMessage = firstMsg;
+            await deviceHandler.SetIncomingMessageCallbackHandlerAndCompleteMessageAsync<string>(ct).ConfigureAwait(false);
 
             // Now, set a callback on the device client to receive C2D messages.
             await deviceHandler.WaitForIncomingMessageCallbackAsync(ct).ConfigureAwait(false);
--- a/e2e/Tests/iothub/device/IncomingMessageCallbackE2eTests.cs
+++ b/e2e/Tests/iothub/device/IncomingMessageCallbackE2eTests.cs
@@ -80,6 +80,46 @@ namespace Microsoft.Azure.Devices.E2ETests.Messaging
             await ReceiveMessageAfterOpenCloseOpenAsync(new IotHubClientMqttSettings(), ct).ConfigureAwait(false);
         }
 
+        [TestMethod]
+        public async Task DeviceReceiveMessageSetCallbackAfterReconnect_Mqtt()
+        {
+            // Setting up one cancellation token for the complete test flow
+            using var cts = new CancellationTokenSource(s_testTimeout);
+            CancellationToken ct = cts.Token;
+
+            await ReceiveMessageSetCallbackAfterReconnect(new IotHubClientMqttSettings(), ct).ConfigureAwait(false);
+        }
+
+        private static async Task ReceiveMessageSetCallbackAfterReconnect(IotHubClientTransportSettings transportSettings, CancellationToken ct)
+        {
+            await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(s_devicePrefix, ct: ct).ConfigureAwait(false);
+            IotHubDeviceClient deviceClient = testDevice.CreateDeviceClient(new IotHubClientOptions(transportSettings));
+            await testDevice.OpenWithRetryAsync(ct).ConfigureAwait(false);
+            using var deviceHandler = new TestDeviceCallbackHandler(deviceClient, testDevice.Id);
+
+            IotHubServiceClient serviceClient = TestDevice.ServiceClient;
+
+            await serviceClient.Messages.OpenAsync(ct).ConfigureAwait(false);
+
+            await deviceHandler.SetIncomingMessageCallbackHandlerAndCompleteMessageAsync<string>(ct).ConfigureAwait(false);
+            await deviceClient.CloseAsync(ct).ConfigureAwait(false);
+
+
+            // Now, send a message to the device from the service.
+            OutgoingMessage firstMsg = OutgoingMessageHelper.ComposeTestMessage(out string _, out string _);
+            deviceHandler.ExpectedOutgoingMessage = firstMsg;
+            await serviceClient.Messages.SendAsync(testDevice.Id, firstMsg, ct).ConfigureAwait(false);
+            VerboseTestLogger.WriteLine($"Sent C2D message from service, messageId={firstMsg.MessageId} - to be received on callback");
+
+            // re-open the client under test.
+            await deviceClient.OpenAsync(ct).ConfigureAwait(false);
+            Thread.Sleep(20000);
+            await deviceHandler.SetIncomingMessageCallbackHandlerAndCompleteMessageAsync<string>(ct).ConfigureAwait(false);
+
+            // Now, set a callback on the device client to receive C2D messages.
+            await deviceHandler.WaitForIncomingMessageCallbackAsync(ct).ConfigureAwait(false);
+        }
+
         private static async Task ReceiveMessageUsingCallbackAndUnsubscribeAsync(IotHubClientTransportSettings transportSettings, CancellationToken ct)
         {
             await using TestDevice testDevice = await TestDevice.GetTestDeviceAsync(s_devicePrefix, ct: ct).ConfigureAwait(false);
--- a/iothub/device/src/Pipeline/MqttTransportHandler.cs
+++ b/iothub/device/src/Pipeline/MqttTransportHandler.cs
@@ -1147,6 +1147,11 @@ namespace Microsoft.Azure.Devices.Client.Transport
                 while(_messageQueue.Count > MessageQueueSize)
                 {
                     _messageQueue.TryDequeue(out _);
+                    if (Logging.IsEnabled)
+                    {
+                        Logging.Info(this, $"Queue size of {MessageQueueSize} for C2D messages has been reached, removing oldest queued C2D message. " +
+                            $"To avoid losing further messages, set SetIncomingMessageCallbackAsync() to process the messages.");
+                    }
                 }
             }
             else if (Logging.IsEnabled)
--- a/iothub/device/src/Pipeline/MqttTransportHandler.cs
+++ b/iothub/device/src/Pipeline/MqttTransportHandler.cs
@@ -185,7 +185,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
 
             _hostName = context.IotHubConnectionCredentials.HostName;
             _connectionCredentials = context.IotHubConnectionCredentials;
-            _messageQueueSize = _mqttTransportSettings.MessageQueueSize;
+            _messageQueueSize = _mqttTransportSettings.IncomingMessageQueueSize;
 
             if (_mqttTransportSettings.Protocol == IotHubClientTransportProtocol.Tcp)
             {
--- a/iothub/device/src/Pipeline/MqttTransportHandler.cs
+++ b/iothub/device/src/Pipeline/MqttTransportHandler.cs
@@ -28,7 +28,6 @@ namespace Microsoft.Azure.Devices.Client.Transport
     internal sealed class MqttTransportHandler : TransportHandlerBase, IDisposable
     {
         private const int ProtocolGatewayPort = 8883;
-        private const int MessageQueueSize = 10;
 
         private const string DeviceToCloudMessagesTopicFormat = "devices/{0}/messages/events/";
         private const string ModuleToCloudMessagesTopicFormat = "devices/{0}/modules/{1}/messages/events/";
@@ -106,6 +105,7 @@ namespace Microsoft.Azure.Devices.Client.Transport
 
         private bool _isSubscribedToTwinResponses;
         private bool _isDeviceReceiveMessageCallbackSet;
+        private int _messageQueueSize;
 
         private readonly string _deviceId;
         private readonly string _moduleId;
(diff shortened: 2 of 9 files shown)


iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -28,11 +28,13 @@ namespace Microsoft.Azure.Devices.Client.Transport
     internal sealed class MqttTransportHandler : TransportHandlerBase, IDisposable
     {
         private const int ProtocolGatewayPort = 8883;
+        private const int MessageQueueSize = 10;
```
tmahmood-microsoft, line 31: Setting the default queue size to 10. The client would always keep track of latest 10 messages.

timtay-microsoft:
Could we just ask users to set the callback before calling open? Or is this to handle a case other than explicit open?

iothub/device/src/Pipeline/MqttTransportHandler.cs
abhipsaMisra: I suspect people are going to want to modify this value - why don't we expose this as a settable property?

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1054,14 +1080,29 @@ private Task HandleDisconnectionAsync(MqttClientDisconnectedEventArgs disconnect
...
         private async Task HandleReceivedMessageAsync(MqttApplicationMessageReceivedEventArgs receivedEventArgs)
         {
             receivedEventArgs.AutoAcknowledge = false;
             string topic = receivedEventArgs.ApplicationMessage.Topic;
 
             if (topic.StartsWith(_deviceBoundMessagesTopic, StringComparison.InvariantCulture))
             {
-                await HandleReceivedCloudToDeviceMessageAsync(receivedEventArgs).ConfigureAwait(false);
+                await HandleReceivedCloudToDeviceMessageAsync(ProcessC2DMessage(receivedEventArgs)).ConfigureAwait(false);
+                await receivedEventArgs.AcknowledgeAsync(CancellationToken.None).ConfigureAwait(false);
```
abhipsaMisra, line 1105: Add some comments here explaining that messages are being added to a queue for later processing, and are being ack'ed immediately

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1108,18 +1140,18 @@ private async Task HandleReceivedCloudToDeviceMessageAsync(MqttApplicationMessag
...
-                // Note that MQTT does not support Abandon or Reject, so always Complete by acknowledging the message like this.
-                try
-                {
-                    await receivedEventArgs.AcknowledgeAsync(CancellationToken.None).ConfigureAwait(false);
-                }
-                catch (Exception ex)
+            }
+            else if(!_isDeviceReceiveMessageCallbackSet)
+            {
+                _messageQueue.Enqueue(receivedCloudToDeviceMessage);
```
abhipsaMisra, line 1146: shouldn't the size be checked prior to enqueuing?

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1108,18 +1140,18 @@ private async Task HandleReceivedCloudToDeviceMessageAsync(MqttApplicationMessag
...
+                while(_messageQueue.Count > MessageQueueSize)
                 {
-                    // This likely happened because the connection was lost. The service will re-send this message so the user
-                    // can acknowledge it on the new connection.
+                    _messageQueue.TryDequeue(out _);
                     if (Logging.IsEnabled)
-                        Logging.Error(this, $"Failed to send the acknowledgement for a received cloud to device message {ex}"); ;
+                    {
+                        Logging.Info(this, $"Queue size of {MessageQueueSize} for C2D messages has been reached, removing oldest queued C2D message. " +
+                            $"To avoid losing further messages, set SetIncomingMessageCallbackAsync() to process the messages.");
```
abhipsaMisra, line 1153: Add this into doc comments. Users don't usually look at logs until something really goes wrong.

abhipsaMisra:
> Could we just ask users to set the callback before calling open? Or is this to handle a case other than explicit open?

I don't think the SDK lets you do _anything_ until the client is opened.

timtay-microsoft:
> > Could we just ask users to set the callback before calling open? Or is this to handle a case other than explicit open?
> 
> I don't think the SDK lets you do _anything_ until the client is opened.

So we would have this same problem with twin/direct method messages, right? I'd prefer if we allow users to set the callbacks before opening the client, but that may be a more involved fix than we need right now. Because of that, I won't push back if this is the approach we take.

abhipsaMisra:
This is technically a workaround for service's current behavior. Ideally, a c2d message would have been published with a QoS of 1, and the client would choose to not ack it and service would redeliver them at some frequency. Once the user sets the callback, those messages could then be ack'ed.
Since service doesn't queue and redeliver unack'ed messages currently, the client needs to do that.

Alternatively, yeah, we could ask users to set the callback before opening the clients. I think that is a reasonable ask but yeah, it will require a bigger change.

Direct methods inherently require there to be an opened client (or, soon to-be opened client), so it isn't that big of an issue there. Deliver once, if there is no ack then fail immediately. Nothing gets queued.
Twin updates aren't persisted either. Hub recommends that you call GetTwin explicitly to retrieve the last known state. So, this isn't really an issue there either.

bastyuchenko:
Hello,
As the issue #3335 creator I would like to share my thoughts:
* this PR suggests having an offline queue, but we already have IoT Hub that works as a queue. If I need to have offline queue, I can create it by myself in my solution. I expect that SDK is a wrapper for an existing Azure service but the PR suggests an implementation of its own "message broker" logic with "offline queue" that is limited by `MessageQueueSize` and is not obvious for developers who will consume this SDK.
* set callback before opening connection can be the solution but.... currently it works in the following way
** send a few c2d messages to IoT Hub
** open a connection from a device side - the messages removed
** send new c2d messages to IoT Hub - the messages removed only because the connection was opened, and no callback was specified 
 it means if I want only communicate with Device Twin, I always opened a 'black hole' that swallows all my telemetry messages  behind-the-scenes due to I've opened connection to access to Device Twin. It's not what I expects, it is not obvious.


____________________________________________________________________

> Ideally, a c2d message would have been published with a QoS of 1, and the client would choose to not ack it and service would redeliver them at some frequency. Once the user sets the callback, those messages could then be ack'ed. Since service doesn't queue and redeliver unack'ed messages currently, the client needs to do that.
I'm sorry in advance, maybe I didn't understand your point of view correctly but it sounds for me a little bit as an "overwriting of `cleanSession` property behavior". Moreover, does it mean that an implementation of message feedback callback will become mandatory on service side?

Is it possible to open a connection but do not consume messages from IoT Hub if MQTT is used?
BTW, how it works in .NET SDK v1? It looks .NET SDK v1 doesn't have such a problem. I open connection but messages do not disappear from IoT Hub (yes, I know that it is not necessary to open connection before set callback in .NET SDK v1).

Is it possible to check how it implemented in other SDKs (i.e. Python or C SDK). It would be great to have some consistency among SDKs.

iothub/device/src/Pipeline/MqttTransportHandler.cs
tmahmood-microsoft: since we are only keeping latest messages, we enqueue the new message and check if size > limit and then remove the oldest message.

abhipsaMisra:
Thanks for sharing your inputs, @bastyuchenko , we appreciate it.

> this PR suggests having an offline queue, but we already have IoT Hub that works as a queue. If I need to have offline queue, I can create it by myself in my solution. I expect that SDK is a wrapper for an existing Azure service but the PR suggests an implementation of its own "message broker" logic with "offline queue" that is limited by MessageQueueSize and is not obvious for developers who will consume this SDK.

IoT Hub is a message queue that will store messages for an offline client if the client had previously connected with `CleanSession` set to `false`. It will do so until the client reconnects again with `CleanSession` set to true. This happens as per contract today.
Now, let us consider the subsequent steps for this client:
- once the client connects, all pending messages will be delivered to it immediately. IoT Hub will not wait until it subscribes to C2D messages to deliver the offline messages to the client. This is because it has connected with `CleanSession` set to `false`.
- for the SDK, it can process these messages only if the user can a callback set. Otherwise, it doesn't have anything to do with these messages. Ideally, the client would not ack these messages, and these being sent with QoS 1 they'd eventually get redelivered. This is the bit that doesn't happen today. IoT hub will send the C2D messages immediately on reconnection, and then never resend them (irrespective of if they were ack'ed or not. This is something the hub team is aware of, and they have it on their plan to fix).
- since the messages were delivered as soon as the client connected, and they won't be redelivered by the service again, there are two options:
    - force the client to subscribe to C2D messages if it wants to be able to process messages sent when offline, OR
    - maintain an internal queue with these offline messages that can be delivered once the client subscribes to C2D messages

I agree that both approaches can have their cons:
- for approach (1) - you are forced to subscribe to and process C2D messages as soon as the client connects, even if that is not what you want. One point of view is that MQTT spec defines that when clients connect with `CleanSession` set to `false` they should be ready to process offline messages as soon as they come online.
- for approach (2) - I agree that this implements a client-side queue on top of service's queue. We'd like to avoid this as well, however, with service's current behavior we don't have many more options.

> maybe I didn't understand your point of view correctly but it sounds for me a little bit as an "override of cleanSession property behavior". Moreover, does it mean that an implementation of message feedback callback will become mandatory on service side?

Could you explain what you meant by these two statements? I'm not sure if the above explanation satisfies your question here, but if it doesn't, I'd like to understand the question a bit better.

> Is it possible to open a connection but do not consume messages from IoT Hub if MQTT is used?

The `CleanSession` property will direct hub to deliver offline messages as soon as the client reconnects. You can set `CleanSession` to `true` and forgo offline message queuing, but there isn't a way to store offline C2D messages and then deliver them after you subscribe to the topic.

> BTW, how it works in .NET SDK v1? It looks .NET SDK v1 doesn't have such a problem. I open connection but messages do not disappear from IoT Hub (yes, I know that it is not necessary to open connection before set callback in .NET SDK v1).

The .NET SDK v1 also has this internal message queue in place, similar to the implementation in this PR.

bastyuchenko:
Thanks for the detailed explanation, @abhipsaMisra 
> Could you explain what you meant by these two statements? I'm not sure if the above explanation satisfies your question here, but if it doesn't, I'd like to understand the question a bit better.

yes, above explanation fully satisfies my question here.

@tmahmood-microsoft, 
from https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-c2d
> When a device wants to receive a message, the IoT hub locks the message by setting the state to Invisible. This state allows other threads on the device to start receiving other messages.

All messages that are stored in "local offline queue" you implemented in this PR will be marked as `Completed` in IoT Hub or just stay `Invisible` (means returns after lock time-out)?

tmahmood-microsoft:
Hi @bastyuchenko, all the messages that are stored in "local offline queue" implemented in this PR will be marked as Completed in IoT Hub.

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
```diff
@@ -95,6 +95,15 @@ public IotHubClientMqttSettings(IotHubClientTransportProtocol transportProtocol
         /// </remarks>
         public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; }
 
+        /// <summary>
+        /// C2D message queue size when callback has not been set.
```
abhipsaMisra, line 99:
```suggestion
        /// Cloud-to-device message queue size for storing offline messages when the client connects with <see cref="CleanSession" /> set to <c>false</c>.
        /// These messages are delivered after the client subscribes to the callback using 
        /// <see cref="IotHubBaseClient.SetIncomingMessageCallbackAsync(Func{IncomingMessage, System.Threading.Tasks.Task{MessageAcknowledgement}}, System.Threading.CancellationToken)" />.
```

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
```diff
@@ -95,6 +95,15 @@ public IotHubClientMqttSettings(IotHubClientTransportProtocol transportProtocol
...
+        /// </summary>
+        /// <remarks>
+        /// If C2D message callback is not set, messages from service are received and stored in a queue until a callback is set. 
+        /// If the number of messages sent is greater than queue size, older messages are removed as latest messages are added.
```
abhipsaMisra, line 103:
```suggestion
        /// If the number of messages sent is greater than queue size, older messages are removed as newer messages are added. Setting this to a low value can cause offline undelivered messages to get lost.
```

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
```diff
@@ -95,6 +95,15 @@ public IotHubClientMqttSettings(IotHubClientTransportProtocol transportProtocol
         /// </remarks>
         public RemoteCertificateValidationCallback RemoteCertificateValidationCallback { get; set; }
 
+        /// <summary>
+        /// C2D message queue size when callback has not been set.
+        /// </summary>
+        /// <remarks>
+        /// If C2D message callback is not set, messages from service are received and stored in a queue until a callback is set.
```
abhipsaMisra, line 102:
```suggestion
        /// If the incoming message callback is not set, messages from service are received and stored in a queue until a callback is set. 
```

abhipsaMisra (commented):
Can you add some tests for this:
- c2d messages aren't lost when you reconnect with no callback set
- c2d messages are redelivered at a later point when the client subscribes to the callback

iothub/device/src/Pipeline/MqttTransportHandler.cs
```diff
@@ -1108,18 +1143,18 @@ private async Task HandleReceivedCloudToDeviceMessageAsync(MqttApplicationMessag
...
-                }
-                catch (Exception ex)
+            }
+            else if(!_isDeviceReceiveMessageCallbackSet)
+            {
+                _messageQueue.Enqueue(receivedCloudToDeviceMessage);
+                while(_messageQueue.Count > _messageQueueSize)
                 {
-                    // This likely happened because the connection was lost. The service will re-send this message so the user
-                    // can acknowledge it on the new connection.
```
abhipsaMisra, line 1120: Is this comment being removed because it is inaccurate?

tmahmood-microsoft: yes, the service does not re-send the message

e2e/Tests/iothub/device/IncomingMessageCallbackE2eTests.cs
```diff
@@ -80,6 +80,46 @@ public async Task DeviceReceiveMessageAfterOpenCloseOpen_Mqtt()
...
+
+            // Now, send a message to the device from the service.
+            OutgoingMessage firstMsg = OutgoingMessageHelper.ComposeTestMessage(out string _, out string _);
+            deviceHandler.ExpectedOutgoingMessage = firstMsg;
+            await serviceClient.Messages.SendAsync(testDevice.Id, firstMsg, ct).ConfigureAwait(false);
+            VerboseTestLogger.WriteLine($"Sent C2D message from service, messageId={firstMsg.MessageId} - to be received on callback");
+
+            // re-open the client under test.
+            await deviceClient.OpenAsync(ct).ConfigureAwait(false);
+            Thread.Sleep(20000);
```
abhipsaMisra, line 116: Two notes:
- Avoid using Thread.Sleep. If you want to introduce an artificial delay/wait, use Task.Delay instead.
- You can avoid introducing artificial delay by calling some other operation first. Call SendEventAsync/ GetTwinAsync instead. That will still simulate the required scenario of the client "doing something" and then subscribing to c2d callback and receiving messages.

e2e/Tests/iothub/device/IncomingMessageCallbackE2eTests.cs
```diff
@@ -80,6 +80,46 @@ public async Task DeviceReceiveMessageAfterOpenCloseOpen_Mqtt()
...
+
+            await serviceClient.Messages.OpenAsync(ct).ConfigureAwait(false);
+
+            await deviceHandler.SetIncomingMessageCallbackHandlerAndCompleteMessageAsync<string>(ct).ConfigureAwait(false);
+            await deviceClient.CloseAsync(ct).ConfigureAwait(false);
+
+
+            // Now, send a message to the device from the service.
+            OutgoingMessage firstMsg = OutgoingMessageHelper.ComposeTestMessage(out string _, out string _);
+            deviceHandler.ExpectedOutgoingMessage = firstMsg;
```
abhipsaMisra, line 110: For the sake of keeping related code together, I'd move this line to right before L120.

iothub/device/src/TransportSettings/IotHubClientMqttSettings.cs
```diff
@@ -95,6 +95,16 @@ public IotHubClientMqttSettings(IotHubClientTransportProtocol transportProtocol
...
 
+        /// <summary>
+        /// Cloud-to-device message queue size for storing offline messages when the client connects with <see cref="CleanSession" /> set to <c>false</c>.
+        /// </summary>
+        /// <remarks>
+        /// If incoming message callback is not set, messages from service are received and stored in a queue until a callback is set. 
+        /// If the number of messages sent is greater than queue size, older messages are removed as newer messages are added. 
+        /// Setting this to a low value can cause offline undelivered messages to get lost.
+        /// </remarks>
+        public int IncomingMessageQueueSize { get; set; } = 10;
```
abhipsaMisra, line 106: Can you add a reference to this property to the public `SetIncomingMessageCallbackAsync` as well. We want people to know about this property when the set the callback (even if they don't want to change the default value).

PR #3340, opened 2023-06-15
addresses supporting byte[] in direct methods and making TwinProperties able to be directly converted to json

patilsnr:
addresses supporting byte[] in direct methods and making `TwinProperties` able to be directly converted to json

API surface diff:

```diff
namespace Microsoft.Azure.Devices.Client {
	public class DirectMethodRequest {
+		public byte[] GetPayload()
-               public byte[] GetPayloadAsBytes()
	}

       public class DirectMethodResponse {
+             public byte[] Payload { get; set; }
-             public object Payload { get; set; }
+             public object PayloadAsObject { get; set; }
       }

       public class EdgeModuleDirectMethodRequest {
+             internal byte[] Payload { get; set; }
-             internal object Payload { get; set; }
+             internal object PayloadAsObject { get; set; }              
       }
}

namespace Microsoft.Azure.Devices {
      public class DirectMethodClientResponse {
+           internal byte[] PayloadAsBytes { get; set; }
      }

      public class DirectMethodServiceRequest {
+          public byte[] Payload { get; set; }     
-           public object Payload { get; set; }
+          public object PayloadAsObject { get; set; }   
    }
}

### CODE REVIEW

File: iothub/device/src/DirectMethod/EdgeModuleDirectMethodRequest.cs [hunk 1]
```diff
@@ -80,7 +80,7 @@ public EdgeModuleDirectMethodRequest(string methodName, object payload)
         /// The direct method payload.
         /// </summary>
         [JsonProperty("payload", NullValueHandling = NullValueHandling.Include)]
-        internal object Payload { get; }
+        internal byte[] Payload { get; }
```
abhipsaMisra, line 83: This field will be deserialized directly when using module-module method invocations over HTTP. Make sure that scenario is tested to ensure serialization works as expected. The long haul tests should cover this scenario.

iothub/device/src/Twin/TwinProperties.cs
```diff
@@ -23,11 +25,23 @@ protected internal TwinProperties(DesiredProperties requestsFromService, Reporte
...
         /// </summary>
+        [JsonProperty("reportedProperties")]
         public ReportedProperties Reported { get; }
+
+        /// <summary>
+        /// Gets the Twin as a JSON string
+        /// </summary>
+        /// <param name="formatting">Optional. Formatting for the output JSON string.</param>
+        /// <returns>JSON string</returns>
+        public string ToJson(Formatting formatting = Formatting.None)
```
abhipsaMisra, line 42: I'm sure you've tested this locally, but can you also add a test for this verifying that the payload is serialized as expected.

rido-min (changes requested):
We'd like to be able to set the DirectMethodResponse payload as `bytes`, in the ctor or with a property setter.

Right now the Payload only accepts `object`

https://github.com/Azure/azure-iot-sdk-csharp/blob/10e631ea827a70bf9201b87f87e929364222b542/iothub/device/src/DirectMethod/DirectMethodResponse.cs#L32

Make sure the serialized TwinProperties include the version

Add tests

iothub/device/src/DirectMethod/EdgeModuleDirectMethodRequest.cs
```diff
@@ -80,7 +81,7 @@ public EdgeModuleDirectMethodRequest(string methodName, object payload)
...
```
rido-min, line 84: any reason to not have a setter?

rido-min (approved):

e2e/LongHaul/device/IotHub.cs
```diff
@@ -472,20 +473,20 @@ private Task<DirectMethodResponse> DirectMethodCallback(DirectMethodRequest meth
 
                             // Log the current time again and send the response back to the service app.
                             methodPayload.SentOnUtc = DateTimeOffset.UtcNow;
-                            return Task.FromResult(new DirectMethodResponse(200) { Payload = methodPayload });
+                            return Task.FromResult(new DirectMethodResponse(200) { Payload = Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(methodPayload)) });
```
abhipsaMisra, line 476: Could we support both the object and bytes[] for the payload?

rido-min:
@patilsnr can you paste the Json returned by twin.GetSerializedJson? I'd like to check the reported and desired versions are properly formatted.

patilsnr:
> @patilsnr can you paste the Json returned by twin.GetSerializedJson? I'd like to check the reported and desired versions are properly formatted.

image

These are logs from the `ThermostatSample` that calls the method in question.

rido-min:
I don't see the reported/desired versions

patilsnr:
> I don't see the reported/desired versions

image

How about this?

rido-min:
Can you paste the text (json) we get from `GetTwinAsyc` ? we need the $version for reported and desired.

patilsnr:
> Can you paste the text (json) we get from `GetTwinAsyc` ? we need the $version for reported and desired.

Sure: 
image

In this case, a brand new device will return the string `expected` when we call `GetTwinPropertiesAsync` (which calls `GetTwinAsync` internally. Debugging this test shows this for the variable `twin` that we get back:
image

Let me know if this is what you're looking for, this should hopefully show that both reported and desired properties are returning the correct version numbers as well as serializing properly

rido-min:
this should be the same as V1 `Twin.ToJson()`

```json
{
  "reported" : {
      "$version" : 1
  },
  "desired" : {
     "$version" : 1
  }
}
```

patilsnr:
> this should be the same as V1 `Twin.ToJson()`
> 
> ```json
> {
>   "reported" : {
>       "$version" : 1
>   },
>   "desired" : {
>      "$version" : 1
>   }
> }
> ```

image

Okay, they seem to be the same

iothub/device/src/Twin/TwinProperties.cs
```diff
@@ -23,11 +25,23 @@ protected internal TwinProperties(DesiredProperties requestsFromService, Reporte
         /// <summary>
         /// The collection of desired property update requests received from service.
         /// </summary>
+        [JsonProperty("desiredProperties")]
```
abhipsaMisra, line 28: The json property name here should be the same as defined in the twin document, i.e. "desired", "reported".

For reference, that is what we define on the service client as well: https://github.com/Azure/azure-iot-sdk-csharp/blob/previews/v2/iothub/service/src/Twin/Models/ClientTwinDocument.cs

iothub/service/src/DirectMethod/Models/DirectMethodServiceRequest.cs
```diff
@@ -30,10 +31,15 @@ public DirectMethodServiceRequest(string methodName)
...
         /// <summary>
-        /// The payload to have serialized and send as JSON.
+        /// The payload to have serialized and sent as JSON.
         /// </summary>
         [JsonProperty("payload")]
-        public object Payload { get; set; }
+        public byte[] Payload { get; set; }
+
+        /// <summary>
+        /// An overload for the payload to have serialized and sent as JSON.
```
abhipsaMisra, line 40:
```suggestion
        /// The payload to have serialized and sent as JSON.
```

iothub/service/src/DirectMethod/Models/DirectMethodServiceRequest.cs
```diff
@@ -30,10 +31,15 @@ public DirectMethodServiceRequest(string methodName)
         public string MethodName { get; }
 
         /// <summary>
-        /// The payload to have serialized and send as JSON.
+        /// The payload to have serialized and sent as JSON.
```
abhipsaMisra, line 34:
```suggestion
        /// The serialized and encoded payload bytes.
```

abhipsaMisra: We're not going to be serializing or encoding this again.

e2e/Tests/helpers/TestDeviceCallbackHandler.cs
```diff
@@ -78,8 +79,9 @@ internal async Task SetDeviceReceiveMethodAndRespondAsync<T>(object deviceRespon
                     {
                         request.MethodName.Should().Be(ExpectedDirectMethodRequest.MethodName, "The expected method name should match what was sent from service");
 
-                        var expectedRequestPayload = (T)ExpectedDirectMethodRequest.Payload;
-                        request.TryGetPayload(out T actualRequestPayload).Should().BeTrue();
+                        byte[] expectedRequestPayload = ExpectedDirectMethodRequest.Payload;
+                        string expectedRequestPayloadAsJson = Encoding.UTF8.GetString(expectedRequestPayload);
+                        request.TryGetPayload(out byte[] actualRequestPayload).Should().BeTrue();
                         actualRequestPayload.Should().BeEquivalentTo(expectedRequestPayload, "The expected method data should match what was sent from service");
```
abhipsaMisra, line 85:
```suggestion
                        request.GetPayloadAsBytes().Should().BeEquivalentTo(ExpectedDirectMethodRequest.Payload, "The expected method data should match what was sent from service");
```

e2e/Tests/helpers/TestModuleCallbackHandler.cs
```diff
@@ -54,9 +55,11 @@ internal async Task SetModuleReceiveMethodAndRespondAsync<T>(object moduleRespon
...
                         request.MethodName.Should().Be(ExpectedDirectMethodRequest.MethodName, "The expected method name should match what was sent from service");
 
-                        var expectedRequestPayload = (T)ExpectedDirectMethodRequest.Payload;
-                        request.TryGetPayload(out T actualRequestPayload).Should().BeTrue();
-                        actualRequestPayload.Should().BeEquivalentTo(expectedRequestPayload, "The expected method data should match what was sent from service");
+                        byte[] expectedRequestPayload = ExpectedDirectMethodRequest.Payload;
+                        string expectedRequestPayloadString = Encoding.UTF8.GetString(expectedRequestPayload);
+                        request.TryGetPayload(out byte[] actualRequestPayload).Should().BeTrue();
+                        string jsonString = Encoding.UTF8.GetString(actualRequestPayload);
+                        jsonString.Should().BeEquivalentTo(expectedRequestPayloadString, "The expected method data should match what was sent from service");
```
abhipsaMisra, line 62: Similar to https://github.com/Azure/azure-iot-sdk-csharp/pull/3340/files#r1267389575

andyk-ms (approved):

iothub/device/src/DirectMethod/DirectMethodRequest.cs
```diff
@@ -131,9 +140,9 @@ public bool TryGetPayload<T>(out T payload)
         /// cancellationToken);
         /// </code>
         /// </example>
-        public byte[] GetPayloadAsBytes()
+        public byte[] GetPayload()
         {
-            return _payload == null || _payload.Length == 0 ? null : (byte[])_payload.Clone();
```
abhipsaMisra, line 136: Is there a reason this didn't work?
This class has the payload always available as bytes (only constructor present is `protected internal DirectMethodRequest(string methodName, byte[] payload = default)`). So, this should directly work.

iothub/device/src/DirectMethod/DirectMethodResponse.cs
```diff
@@ -21,24 +25,36 @@ public DirectMethodResponse(int status)
         /// <summary>
         /// The status of direct method response.
         /// </summary>
+        [JsonProperty]
```
abhipsaMisra, line 28: q - it looks like this PR is adding the json annotations - how did it work previously? Were the fields being explicitly set, as opposed to being set during deserialization?

patilsnr: if this is on the device side, it may not change anything. I'll get rid of them and see what happens.

iothub/device/src/DirectMethod/DirectMethodRequest.cs
patilsnr: Yup, this should work. Changing it back.

PR #3348, opened 2023-07-12
Add e4k connector sas utilities to v2

patilsnr:


iothub/device/src/Authentication/ClientAuthenticationWithSharedAccessKeyRefreshBehalfOf.cs
```diff
@@ -0,0 +1,60 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Client.Authentication
+{
+    /// <summary>
+    /// Authentication method that generates shared access signature (SAS) token with refresh, based on a provided shared access key (SAK).
+    /// Build for On Behalf of scenarios only.
```
abhipsaMisra, line 11: We should clarify in the comments that this is for gateway scenarios when one client authenticates on behalf of another.

iothub/device/src/Authentication/ClientAuthenticationWithSharedAccessKeyRefreshBehalfOf.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Microsoft.Azure.Devices.Client.Authentication
+{
+    /// <summary>
+    /// Authentication method that generates shared access signature (SAS) token with refresh, based on a provided shared access key (SAK).
+    /// Build for On Behalf of scenarios only. 
+    /// </summary>
+    public class ClientAuthenticationWithSharedAccessKeyRefreshBehalfOf : ClientAuthenticationWithSharedAccessKeyRefresh
+    {
+#nullable enable
```
abhipsaMisra, line 15: Why is this required?

iothub/device/src/Authentication/ClientAuthenticationWithSharedAccessKeyRefreshBehalfOf.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <summary>
+        /// Creates an instance of this class.
+        /// </summary>
+        /// <param name="sharedAccessKey">Shared access key value.</param>
+        /// <param name="parentDeviceId">Identifier of the parent device.</param>
+        /// <param name="deviceId">Device identifier.</param>
```
abhipsaMisra, line 21: I'm not very familiar with this scenario so I don't have the exact wordings to recommend, but it feels like we could use some additional description to clarify how the parent device relates to "this" device/module.

iothub/device/src/Authentication/ClientAuthenticationWithSharedAccessKeyRefreshBehalfOf.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Gets the shared access key name.
+        /// </summary>
+        public string ParentDeviceId { get; private set; }
+
+        ///<inheritdoc/>
+        protected override Task<string> SafeCreateNewTokenAsync(string iotHub, TimeSpan suggestedTimeToLive)
+        {
+            string sasToken = SharedAccessSignatureBuilder.GetDeviceToken(iotHub, ParentDeviceId, "$edgeHub", SharedAccessKey);
```
abhipsaMisra, line 56: Is there a valid use-case to make setting "$edgeHub" configurable? If not, then we should rename the class to call out that this is for edgehub scenarios only.

iothub/device/src/Authentication/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -16,6 +16,7 @@ namespace Microsoft.Azure.Devices.Client
     /// </summary>
     internal sealed class SharedAccessSignatureBuilder
     {
+        private const int DEFAULT_TOKEN_EXPIRY_MINS = 60;
```
abhipsaMisra, line 19: The SDK should already be maintaining a default ttl reference somewhere. Reuse that instead of adding a new const again.

iothub/device/src/Authentication/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -145,5 +146,38 @@ internal static string Sign(string requestString, string key)
...
+
+        internal static string GetDeviceToken(string hostname, string deviceId, string key, long expiry = 0)
+        {
+            return GetToken(GetDeviceResourceURI(hostname, deviceId), key, expiry);
+        }
+
+        internal static string GetDeviceToken(string hostname, string deviceId, string moduleId, string key, long expiry = 0)
+        {
+            return GetToken(GetDeviceResourceURI(hostname, deviceId, moduleId), key, expiry);
+        }
```
abhipsaMisra, line 171: Instead of creating additional overloads for each of these methods you can use default values for params and have a single implementation.

iothub/device/src/Authentication/ClientAuthenticationWithSharedAccessKeyRefreshBehalfOf.cs
rido-min: afaik `OnBehalfOf` only works for `$edgehub`, so I agree we can have a better class name.

How about: `ClientAuthenticationForEdgeHubOnBehalfOf` ?

iothub/device/src/Authentication/ClientAuthenticationWithSharedAccessKeyRefreshBehalfOf.cs
patilsnr: required for using a nullable operator, it's been removed and replaced by a default value in the method call

PR #3353, opened 2023-07-24
Add OnBehalfOf authentication to V2 for edgehub

patilsnr:
Adds OnBehalfOf authentication to SDK for Edge devices

### DIFF
--- a/iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
+++ b/iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
@@ -8,14 +8,14 @@ namespace Microsoft.Azure.Devices.Client.Authentication
 {
     /// <summary>
     /// Authentication method that generates shared access signature (SAS) token with refresh, based on a provided shared access key (SAK).
-    /// Build for gateway scenarios when one client authenticates on behalf of the other only.
+    /// Build for using $edgeHub in IoT Edge to authenticate on behalf of leaf devices or modules only.
     /// </summary>
     public class ClientAuthenticationForEdgeHubOnBehalfOf : ClientAuthenticationWithSharedAccessKeyRefresh
     {
         /// <summary>
         /// Creates an instance of this class.
         /// </summary>
-        /// <param name="sharedAccessKey">Shared access key value for the higher-layer parent device.</param>
+        /// <param name="sharedAccessKey">Shared access key value for the $edgehub module.</param>
         /// <param name="parentDeviceId">Identifier of the higher-layer parent device that connects directly to IoT Hub.</param>
         /// <param name="deviceId">Device identifier of the lower-layer device that authenticates through the parent IoT device.</param>
         /// <param name="moduleId">Module identifier.</param>
--- a/iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
+++ b/iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
@@ -15,9 +15,9 @@ namespace Microsoft.Azure.Devices.Client.Authentication
         /// <summary>
         /// Creates an instance of this class.
         /// </summary>
-        /// <param name="sharedAccessKey">Shared access key value.</param>
-        /// <param name="parentDeviceId">Identifier of the parent device.</param>
-        /// <param name="deviceId">Device identifier.</param>
+        /// <param name="sharedAccessKey">Shared access key value for the higher-layer parent device.</param>
+        /// <param name="parentDeviceId">Identifier of the higher-layer parent device that connects directly to IoT Hub.</param>
+        /// <param name="deviceId">Device identifier of the lower-layer device that authenticates through the parent IoT device.</param>
         /// <param name="moduleId">Module identifier.</param>
         /// <param name="sasTokenTimeToLive">
         /// The suggested time to live value for the generated SAS tokens.
--- a/iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
+++ b/iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
@@ -52,7 +52,8 @@ namespace Microsoft.Azure.Devices.Client.Authentication
         ///<inheritdoc/>
         protected override Task<string> SafeCreateNewTokenAsync(string iotHub, TimeSpan suggestedTimeToLive)
         {
-            string sasToken = SharedAccessSignatureBuilder.GetDeviceToken(iotHub, ParentDeviceId, SharedAccessKey, "$edgeHub");
+            string audience = SharedAccessSignatureBuilder.BuildAudience(iotHub, ParentDeviceId, "$edgeHub");
+            string sasToken = SharedAccessSignatureBuilder.BuildSignature(null, SharedAccessKey, null, TimeSpan.FromMinutes(60), audience, null, null);
             return Task.FromResult(sasToken);
         }
     }
--- a/iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
+++ b/iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
@@ -1,0 +1,59 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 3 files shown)


iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Authentication method that generates shared access signature (SAS) token with refresh, based on a provided shared access key (SAK).
+    /// Build for gateway scenarios when one client authenticates on behalf of the other only.
+    /// </summary>
+    public class ClientAuthenticationForEdgeHubOnBehalfOf : ClientAuthenticationWithSharedAccessKeyRefresh
+    {
+        /// <summary>
+        /// Creates an instance of this class.
+        /// </summary>
+        /// <param name="sharedAccessKey">Shared access key value for the higher-layer parent device.</param>
```
patilsnr, line 18: I'd love a double check on the correctness of these parameter descriptions, as I'm still getting familiar with this form of authentication

iothub/device/tests/Authentication/EdgeDeviceOnBehalfOfTests.cs
```diff
@@ -0,0 +1,143 @@
...
+    public class EdgeDeviceOnBehalfOfTests
+    {
+        private static string _testKey => Convert.ToBase64String(Encoding.UTF8.GetBytes(Guid.Empty.ToString("N")));
+
+        [TestMethod]
+        public async Task ConnectDeviceOnBehalfOf()
+        {
+            string leafDeviceId = "test-leaf-device";
+
+            DeviceConnectionSettings edgeDevice = new DeviceConnectionSettings("e4k-hub.azure-devices.net", "test-edge-device")
```
patilsnr, line 20: `DeviceConnectionSettings` and `EdgeHubConnectionSettingsBuilder` are classes that don't exist yet -- I'm looking for guidance on whether to add them from the e4k connector codebase or to find analogous places here in V2 to replace their functionality for these tests

iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Devices.Client.Authentication
+{
+    /// <summary>
+    /// Authentication method that generates shared access signature (SAS) token with refresh, based on a provided shared access key (SAK).
+    /// Build for gateway scenarios when one client authenticates on behalf of the other only.
```
rido-min, line 11: Can we mention in the comments that this only works for $edgeHub ?

`gateway scenarios` can be confusing, I think we mean "when using the $edgeHub in IoTEdge to authenticate on behalf of leaf devices or modules"

iothub/device/src/Authentication/ClientAuthenticationForEdgeHubOnBehalfOf.cs
rido-min, line 18: The SharedAccessKey is the key for `$edgehub` module

iothub/device/src/Authentication/Security/SharedAccessSignatureBuilder.cs
```diff
@@ -145,5 +145,30 @@ internal static string Sign(string requestString, string key)
             using var algorithm = new HMACSHA256(Convert.FromBase64String(key));
             return Convert.ToBase64String(algorithm.ComputeHash(Encoding.UTF8.GetBytes(requestString)));
         }
+
+        internal static string GetSignature(string encodedURI, string key, long expiry)
+        {
+            return Sign($"{encodedURI}\n{expiry}", key);
+        }
+
+        internal static string GetToken(string encodedURI, string key, long expiry = 0, int defaultTimeToLive = 60)
```
rido-min, line 154: Seems there is some duplication with existing methods in this class, so now there are two mechanisms to create the tokens. 

Any reason to do not use BuildSignature()?

iothub/device/tests/Authentication/EdgeDeviceOnBehalfOfTests.cs
rido-min: I would not include the DeviceConnectionSettings in V2. The class is used just to hold the connection settings, you should be able to use IotHubConnectionString or just the individual values required to establish the connection

iothub/device/tests/Authentication/EdgeDeviceOnBehalfOfTests.cs
```diff
@@ -0,0 +1,143 @@
...
+namespace Microsoft.Azure.Devices.Client.Tests.OnBehalfOf
+{
+    public class EdgeDeviceOnBehalfOfTests
+    {
+        private static string _testKey => Convert.ToBase64String(Encoding.UTF8.GetBytes(Guid.Empty.ToString("N")));
+
+        [TestMethod]
+        public async Task ConnectDeviceOnBehalfOf()
+        {
+            string leafDeviceId = "test-leaf-device";
```
rido-min, line 18: wrt test device ids: 
are you using the same approach for e2e tests? (assume the device identity has been provisioned and we have a known parent child relationship)

iothub/device/tests/Authentication/EdgeDeviceOnBehalfOfTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            string edgeModuleId = "test-leaf-device-module";
+
+            DeviceConnectionSettings edgeDevice = new DeviceConnectionSettings("e4k-hub.azure-devices.net", "test-edge-device")
+            {
+                SharedAccessKey = _testKey
+            };
+
+            var edgeHubCs = await EdgeHubConnectionSettingsBuilder.Build(edgeDevice);
```
rido-min, line 68: Where is EdgeHubConnectionSettingsBuilder? bringing this feature to V2 is very compelling (to enable provision any module, and initialize edgeHub), but I would leave it for a different PR.

@abhipsaMisra  ?

rido-min: btw.. the tests can be initialized with a _known_  $edgehub key

iothub/device/tests/Authentication/EdgeDeviceOnBehalfOfTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                edgeHubCs.SharedAccessKey!,
+                edgeHubCs.DeviceId!,
+                leafDeviceId,
+                null,
+                TimeSpan.FromMinutes(10),
+                5);
+
+            IotHubDeviceClient leafClient = new(edgeHubCs.HostName, leafAuth,
+                new IotHubClientOptions(
+                    new IotHubClientAmqpSettings
```
rido-min, line 37: should we test with MQTT as well?

we dont need to configure multiplexing in any test

rido-min (changes requested):
- Clarify some comments
- Avoid bringing more classes from SdkLite
- We need a test strategy to provision required parent/child devices

iothub/device/tests/Authentication/EdgeDeviceOnBehalfOfTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            await leafClient.DisposeAsync();
+        }
+
+        [TestMethod]
+        public async Task ConnectLeafModuleOnBehalfOf()
+        {
+            string leafDeviceId = "test-leaf-device";
+            string edgeModuleId = "test-leaf-device-module";
+
+            var edgeHubCs = new IotHubConnectionString("e4k-hub.azure-devices.net", null, "test-edge-device", "$edgeHub", null, _testKey, null);
```
andyk-ms, line 60: is there a reason why "test-leaf-device" is in a variable and "test-edge-device" is not?

iothub/device/tests/Authentication/EdgeDeviceOnBehalfOfTests.cs
```diff
@@ -0,0 +1,134 @@
...
+            twin.Should().NotBeNull();
+            twin.Reported.Should().NotBeNull();
+            twin.Reported["tick"].Should().NotBeNull();
+            tick.Should().Be((long)twin.Reported["tick"]);
+            await leafClient.CloseAsync();
+            await leafClient.DisposeAsync();
+        }
+
+        [TestMethod]
+        public async Task ConnectLeafModuleOnBehalfOf()
```
andyk-ms, line 56: Would it be better to consolidate two tests to call a single test function with different inputs for easier maintenance given both tests are similar?

Issue #3354, opened 2023-07-26
[Bug Report] [v2] Multiple calls to direct methond after reconnect


jenscski:
# Context

- **OS, version, SKU and CPU architecture used:** Windows 11
- **Application's .NET Target Framework :** net6.0
- **Device:** Windows Console app and .NET Maui (Samsung s32)
- **SDK version used:** 2.0.0-preview005

## Description of the issue
When a connection is automatically reconnected (e.g. network drop or similiar), direct method calls is called twice (possibly one more for each reconnect)

I have tested with both disabling my network, and turning "Enable connection to IoT Hub" on and off.

The reason for the error is that MqttTransportHandler.OpenAsync is called upon reconnect, but MqttTransportHandler.CloseAsync is not, so the HandleReceivedMessageAsync method is added again to the MqttClient.ApplicationMessageReceivedAsync event.

tmahmood-microsoft:
Hi @jenscski, thank you so much for evaluating previews/v2 and reporting this issue. 
I was able to reproduce and fix the issue, you can find the latest code changes here: https://github.com/Azure/azure-iot-sdk-csharp/pull/3359

Please let me know if it still does not work for you or if you face any more issues.

helynranta:
Hi @tmahmood-microsoft, we had similar issue and I was able to verify that your fix solves it. Can you publish new preview package which includes this?

tmahmood-microsoft:
@helynranta thank you for confirming the fix. Yes I am currently in process of publishing a new preview package. Should have it for you in a day or two.

tmahmood-microsoft:
Hi @helynranta, @jenscski I am attaching link to the latest release package for v2 preview containing this fix, please let me know if you face any more issues and thank you for evaluating the preview release.

https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/preview_2023-08-14

PR #3359, opened 2023-08-07
Bug fix: Direct method subscription not removed after disconnect

tmahmood-microsoft:
After a disconnect event, when the client reconnects, client subscribes to message receive event again resulting in multiple calls to direct method callback.
The root cause was missing unsubscription after a disconnect event.

github issue: https://github.com/Azure/azure-iot-sdk-csharp/issues/3354

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -60,7 +60,7 @@
     <PackageReference Include="Azure.Core" Version="1.31.0" />
     <PackageReference Include="MQTTnet" Version="4.1.4.563" />
     <PackageReference Include="Microsoft.Azure.Amqp" Version="2.6.2" />
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.3">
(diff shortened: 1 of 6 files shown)


timtay-microsoft (approved):

abhipsaMisra (changes requested):
RemoveOldOperations is called at regular intervals to remove older pending twin operation tasks from our internal collection.
Unhooking the disconnection and received message event handlers in the same function isn't what you want to do, since what will happen here is even without any disconnection event the handlers are getting unhooked after a certain interval.
Instead, you need to unhook the message processor as a part of the transport handler's disconnection detection block.

tmahmood-microsoft:
Updated and unhooking the message processor as a part of the transport handler's disconnection detection block.

abhipsaMisra:
What scenarios did you test for this?
Can we add some tests so that we don't regress in the future?
For a unit test you could set up a mock client and trigger the disconnection event handler. Then, verify that the message processor handler was unhooked. As a bonus, you can also test that no other operation was executed on the client, i.e. no other handlers were attached or detached.

patilsnr (commented):
Agree about adding a test, but otherwise makes sense.

abhipsaMisra (approved):

patilsnr (approved):

PR #3386, opened 2023-10-20
Minimize flaky test impact on the E2E tests

andyk-ms:
Added flaky/ignore labels to tests that has random timeout and network failure.
Added do not parallelize labels to reduce testing interference.


e2e/test/iothub/messaging/FaultInjectionPoolAmqpTests.MessageSendFaultInjectionPoolAmqpTests.cs
```diff
@@ -286,8 +298,11 @@ public async Task Message_AmqpConnectionLossSendRecovery_MultipleConnections_Amq
         }
 
         // TODO: #950 - Link/session faults for message send/ method/ twin operations closes the connection.
+        [Ignore]
```
timtay-microsoft, line 301: Are these multiplexing tests flakey in prod, or just against canary? If they are only flakey against canary, I'd rather we leave them enabled. The GWv2 folks do run our pipelines from time to time to see where there may be problems, and it would be nice if they saw these failures still

timtay-microsoft (approved):

PR #3388, opened 2023-10-31
Remove "key" and "value" keys from returned json

patilsnr:
`TwinProperties.ToJson()` will now return something like this:
```json
{
  "properties" : {
      "desired" : {
           "myProp" : "myValue",
           "$version" : 3
      },
     "reported" : {
            "myProp" : "myValue",
           "$version" : 4
     }
   }
}
```

### DIFF
--- a/iothub/device/src/Twin/DesiredProperties.cs
+++ b/iothub/device/src/Twin/DesiredProperties.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Azure.Devices.Client
         /// This class can be inherited from and set by unit tests for mocking purposes.
         /// </remarks>
         protected internal DesiredProperties(Dictionary<string, object> desiredProperties)
-            : base(desiredProperties, true)
+            : base(desiredProperties, false)
         {
         }
     }
--- a/iothub/device/src/Twin/TwinProperties.cs
+++ b/iothub/device/src/Twin/TwinProperties.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using Newtonsoft.Json;
 
 namespace Microsoft.Azure.Devices.Client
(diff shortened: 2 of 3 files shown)


varunpuranik:
What issue does this address, and why is the new way better than the old one?

rido-min:
Why $version is not a field? where is the $version for reported?

The Json should be something like:

```json
{
  "properties" : {
      "desired" : {
           "myProp" : "myValue",
           "$version" : 3
      },
     "reported" : {
            "myProp" : "myValue",
           "$version" : 4
     }
   }
}
```

patilsnr:
> What issue does this address, and why is the new way better than the old one?

My understanding is that this is consistent with what the V1 API returns. Without this fix an extra level of serialization exists where "key" and "value" are literally written into the json object

iothub/device/src/Twin/DesiredProperties.cs
```diff
@@ -23,7 +23,7 @@ public class DesiredProperties : PropertyCollection
         /// This class can be inherited from and set by unit tests for mocking purposes.
         /// </remarks>
         protected internal DesiredProperties(Dictionary<string, object> desiredProperties)
-            : base(desiredProperties, true)
+            : base(desiredProperties, false)
```
patilsnr, line 26: If this is a mocking ctor, false makes more sense here to just add the properties to the internal dictionary

varunpuranik:
> > What issue does this address, and why is the new way better than the old one?
> 
> My understanding is that this is consistent with what the V1 API returns. Without this fix an extra level of serialization exists where "key" and "value" are literally written into the json object

What is returned by IoT Hub?

patilsnr:
> > > What issue does this address, and why is the new way better than the old one?
> > 
> > 
> > My understanding is that this is consistent with what the V1 API returns. Without this fix an extra level of serialization exists where "key" and "value" are literally written into the json object
> 
> What is returned by IoT Hub?

Running a quick test with a device in my own Hub, I get: 

`{"properties":{"desired":"{\"$version\":\"1\"}","reported":"{\"$version\":\"1\"}"}}`

which is consistent with the desired format

andyk-ms (approved):

Issue #3379, opened 2023-10-13
Json Format Exception to get payload from Direct Method Request and Twin Json Wrongly formatted


olivakar:
# Context
- **SDK version used:** 2.0.0-preview007

## Description of the issue
1. The `GetPayload()` on `DirectMethodRequest` results in a JSON Format Exception.
2. The full twin reported seems to be different in the properties section and appears with keywords like key and value.

## Code sample exhibiting the 1st issue
```
await leafClient.SetDirectMethodCallbackAsync(async req =>
                {
                    _logger.LogInformation("c <- {cid}", deviceId!);
                    _logger.LogDebug("Method {name} called on {did} with rid: {rid}", req.MethodName, deviceId!, req.RequestId);

                    E4K.IoTHubClient.Binders.DirectMethodRequest cmdReq = new(req.MethodName)
                    {
                        CorrelationId = new Guid(req.RequestId),
                        Payload = req.GetPayload()
                    };
                    E4K.IoTHubClient.Binders.DirectMethodResponse cmdResp = await commandClient.InvokeAsync(deviceId!, cmdReq);
                    _logger.LogInformation("c -> {cid}", deviceId!);
                    return new DirectMethodResponse(cmdResp.Status)
                    {
                        Payload = cmdResp.ResponsePayloadBytes!
                    };
                }, stoppingToken).ConfigureAwait(false);
```

## Console log of the 1st issue
`Exception thrown: 'System.FormatException' in Newtonsoft.Json.dll`

Detailed logs from a console application written to identify the line:
```
Exception occurred: The input is not a valid Base-64 string as it contains a non-base 64 character, more than two padding characters, or an illegal character among the padding characters.
Stack trace
   at System.Convert.FromBase64CharPtr(Char* inputPtr, Int32 inputLength)
   at System.Convert.FromBase64CharArray(Char[] inArray, Int32 offset, Int32 length)
   at Newtonsoft.Json.JsonTextReader.ParseReadString(Char quote, ReadType readType)
   at Newtonsoft.Json.JsonTextReader.ParseString(Char quote, ReadType readType)
   at Newtonsoft.Json.JsonTextReader.ReadAsBytes()
   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
[...]

PR #3391, opened 2023-11-01
update payload convention number of serializations

patilsnr:
addresses the first part of https://github.com/Azure/azure-iot-sdk-csharp/issues/3379

### DIFF
--- a/iothub/device/src/Serialization/DefaultPayloadConvention.cs
+++ b/iothub/device/src/Serialization/DefaultPayloadConvention.cs
@@ -78,7 +78,17 @@ namespace Microsoft.Azure.Devices.Client
         [SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "Method must be instance to benefit from s_settings.")]
         internal T GetObject<T>(string jsonObjectAsText)
         {
-            return JsonConvert.DeserializeObject<T>(jsonObjectAsText);
+            T deserialized;
+
+            try
+            {
+                deserialized = JsonConvert.DeserializeObject<T>(jsonObjectAsText);
+            }
+            catch (JsonReaderException)
+            {
+                deserialized = JsonConvert.DeserializeObject<T>(JsonConvert.SerializeObject(Encoding.UTF8.GetBytes(jsonObjectAsText)));
+            }
+            return deserialized;
         }
 
         // For internal and unit testing use
--- a/iothub/device/src/Serialization/DefaultPayloadConvention.cs
+++ b/iothub/device/src/Serialization/DefaultPayloadConvention.cs
@@ -86,6 +86,8 @@ namespace Microsoft.Azure.Devices.Client
             }
             catch (JsonReaderException)
             {
+                // T should always be of type byte[] here, so this is basically a no-op.
+                // However, .NET does not allow us to simply return "payload as T" or other basic casts.
                 deserialized = JsonConvert.DeserializeObject<T>(JsonConvert.SerializeObject(Encoding.UTF8.GetBytes(jsonObjectAsText)));
             }
             return deserialized;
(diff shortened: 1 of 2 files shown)


iothub/device/src/Serialization/DefaultPayloadConvention.cs
```diff
@@ -78,7 +78,17 @@ internal T GetObject<T>(Stream streamToConvert)
...
-            return JsonConvert.DeserializeObject<T>(jsonObjectAsText);
+            T deserialized;
+
+            try
+            {
+                deserialized = JsonConvert.DeserializeObject<T>(jsonObjectAsText);
+            }
+            catch (JsonReaderException)
+            {
+                deserialized = JsonConvert.DeserializeObject<T>(JsonConvert.SerializeObject(Encoding.UTF8.GetBytes(jsonObjectAsText)));
```
timtay-microsoft, line 89: Isn't this just a no-op? You are deserializing the thing you just serialized, right?

iothub/device/src/Serialization/DefaultPayloadConvention.cs
```diff
@@ -78,7 +78,17 @@ internal T GetObject<T>(Stream streamToConvert)
         [SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "Method must be instance to benefit from s_settings.")]
         internal T GetObject<T>(string jsonObjectAsText)
         {
-            return JsonConvert.DeserializeObject<T>(jsonObjectAsText);
+            T deserialized;
+
+            try
```
timtay-microsoft, line 83: I'm a bit out of the loop on this one, so can you explain why we need to have two code paths here? I would expect a deserialize method to only handle one format of string here, not two.

iothub/device/src/Serialization/DefaultPayloadConvention.cs
patilsnr: The serialize-deserialize step ensures that we end up with type T as the final result. It's ugly, but if it were to be left out, there would be a compiler error saying that we're returning byte[] not type T (even though in the case of direct method payloads, they're the same).

iothub/device/src/Serialization/DefaultPayloadConvention.cs
patilsnr: Sure - there's an inconsistency when setting direct method payloads from the SDK vs setting them from the portal or Azure CLI. In the latter case, they're a layer of encoding behind the SDK set payloads, the method here tries to deserialize and performs that "extra layer" of encoding if it cannot be. This lack of consistency is partially my fault, since there aren't tests for when the payload is set through the portal or through AZ CLI and this bug didn't come up in the initial testing for the feature that made payloads into byte[]. Rectifying this would be a good idea in the future for SDK work, but it seemed like getting the fix out was the priority.

iothub/device/src/Serialization/DefaultPayloadConvention.cs
timtay-microsoft: Gotcha. We actually hit this same problem in the e4k sdk repo, and this is what we settled on. Feel free to use what you already have, but please do leave a note in code explaining all this for future readers.

```suggestion
            // T should always be of type byte[] here, so this is basically a no-op.
            // However, .NET does not allow us to simply return "payload as T" or other basic casts.
                deserialized = (T)Convert.ChangeType(jsonObjectAsText, typeof(T), CultureInfo.InvariantCulture);
```

iothub/device/src/Serialization/DefaultPayloadConvention.cs
timtay-microsoft: Gotcha, thanks for the context

timtay-microsoft (approved):

PR #3406, opened 2023-12-11
Update build-test-nuget.yml for Azure Pipelines

tmahmood-microsoft:


abhipsaMisra (approved):

andyk-ms (approved):
If testing fails, update the yaml to remove tests like vsts.yaml.
Older .net package testing were removed.

PR #3414, opened 2023-12-19
Using 'Azure.Storage.Blobs' instead of 'WindowsAzure.Storage'

tmahmood-microsoft:


iothub/device/src/Transport/Http/HttpTransportHandler.cs
```diff
@@ -207,14 +209,13 @@ internal async Task UploadToBlobAsync(string blobName, Stream source, Cancellati
...
-                Task uploadTask = blob.UploadFromStreamAsync(source, null, null, null, cancellationToken);
-                await uploadTask.ConfigureAwait(false);
+                var blockBlobClient = new BlockBlobClient(new Uri(putString));
+                var uploadTask = await blockBlobClient.UploadAsync(source, new BlobUploadOptions(), cancellationToken).ConfigureAwait(false);
 
                 notification.CorrelationId = fileUploadResponse.CorrelationId;
-                notification.IsSuccess = uploadTask.IsCompleted;
-                notification.StatusCode = uploadTask.IsCompleted ? 0 : -1;
-                notification.StatusDescription = uploadTask.IsCompleted ? null : "Failed to upload to storage.";
+                notification.IsSuccess = uploadTask.Value.LastModified.Subtract(DateTimeOffset.Now) <= new TimeSpan(100);
```
tmahmood-microsoft, line 216: couldn't figure out a better way to validate successful upload

timtay-microsoft: It looks like unsuccessful uploads throw:

https://github.com/Azure/azure-sdk-for-net/blob/64386052196f8f8293367c8b24dc42ae48d8a7fc/sdk/storage/Azure.Storage.Blobs/src/BlockBlobClient.cs#L589

timtay-microsoft: I'd recommend putting the upload operation in a try/catch block so that we always complete the SAS URI regardless of upload success

tmahmood-microsoft: @timtay-microsoft we already have the upload operation is try/catch and do a post operation with failed notification in case of unsuccessful upload

timtay-microsoft: You should be fine to remove this timespan check, then. If the upload call returns without throwing, you can assume it uploaded successfully.

timtay-microsoft (approved):
Looks good other than the timespan check

PR #3420, opened 2023-12-27
Updated LTS versions

tmahmood-microsoft:


iothub/service/src/Microsoft.Azure.Devices.csproj
```diff
@@ -27,7 +27,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.32.0</Version>
+    <Version>1.31.6</Version>
```
tmahmood-microsoft, line 30: reverting from 1.32.0 as 1.32.0 already exists as a version from Main.

andyk-ms (approved):

Issue #3434, opened 2024-02-23
[Bug Report] Send message in batch to multiple outputs using ModuleClient throws


wischi-chr:
# Context

_Context is very likely irrelevant_
- **OS:** Windows 10 Desktop x64
- **Application's .NET Target Framework :** .net core 6
- **Device:** Laptop
- **SDK version used:** 1.42.2 (but also inspected the source of the latest main commit)

## Description of the issue
The bug is almost exactly the same as described in #3201, but for me it happened with `SendEventBatchAsync`.

Sending a message to multiple outputs, or multiple times throws the following exception:
`An item with the same key has already been added. Key: iothub-outputname`

Looking at the source it seems that the fix (using the indexer to set the property) for #3201 was only applied to the `SendEventAsync` method. The method `SendEventBatchAsync` still uses `.Add()` and throws.

https://github.com/Azure/azure-iot-sdk-csharp/blob/9e445ce1e817c225758d7a351974848b7259788c/iothub/device/src/InternalClient.cs#L1573

Would be great to fix it there too, I'm currently using reflection to remove the key before I call the ModuleClient - which is obviously not very pretty :-/

Thanks in advance

timtay-microsoft:
#3441 should fix this for you

wischi-chr:
Was shipped with [1.42.3](https://www.nuget.org/packages/Microsoft.Azure.Devices.Client/1.42.3) (release notes [here](https://github.com/Azure/azure-iot-sdk-csharp/releases/tag/2024-03-28)) so I'm closing the issue.

Thanks for fixing it.

PR #3441, opened 2024-03-13
Fix bug where module client send batch fails if output name already set

timtay-microsoft:
#3434

The method ```Add``` on a dictionary throws if the key is already present. This change will overwrite any previously set output name without throwing.


abhipsaMisra (approved):
Let's add in a test to validate this behavior.

timtay-microsoft:
Test added

PR #3462, opened 2024-06-18
Fix bug where opening service client with Amqp_WebSocket_Only throws ArgumentOutOfRangeException

brycewang-microsoft:
Customer reported this issue via ICM
image

Also added an e2e test to cover this particular scenario in this PR.

### DIFF
--- a/iothub/service/src/Messaging/ServiceClient.cs
+++ b/iothub/service/src/Messaging/ServiceClient.cs
@@ -290,9 +290,7 @@ namespace Microsoft.Azure.Devices
             if (Logging.IsEnabled)
                 Logging.Enter(this, $"Opening AmqpServiceClient", nameof(OpenAsync));
 
-            using var ctx = new CancellationTokenSource(_openTimeout);
-
-            await _faultTolerantSendingLink.OpenAsync(ctx.Token).ConfigureAwait(false);
+            await _faultTolerantSendingLink.OpenAsync(_openTimeout).ConfigureAwait(false);
             await _feedbackReceiver.OpenAsync().ConfigureAwait(false);
 
             if (Logging.IsEnabled)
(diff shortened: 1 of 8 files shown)


iothub/service/src/Messaging/ServiceClient.cs
```diff
@@ -290,9 +289,7 @@ public virtual async Task OpenAsync()
             if (Logging.IsEnabled)
                 Logging.Enter(this, $"Opening AmqpServiceClient", nameof(OpenAsync));
 
-            using var ctx = new CancellationTokenSource(_openTimeout);
-
-            await _faultTolerantSendingLink.OpenAsync(ctx.Token).ConfigureAwait(false);
+            await _faultTolerantSendingLink.OpenAsync(_openTimeout).ConfigureAwait(false);
```
brycewang-microsoft, line 292: For the `OpenAsync()` method being invoked here, the call stacks are following - passing `TimeSpan` vs `CancellationToken`.
image
image

It's surprising that passing in `TimeSpan` would lead to `TimeSpan.MaxValue` being used in downstream no matter what's the value we are passing. This eventually caused the ArgumentOutOfRangeException.

timtay-microsoft:
It sounds like we could also just ask the user to provide a timespan or cancellation token to their ```serviceClient.OpenAsync(...)``` call to immediately unblock them, right?

timtay-microsoft (approved):

iothub/service/src/Messaging/ServiceClient.cs
abhipsaMisra: Do we have a cancellation token variant of OpenAsync() in our public API?

abhipsaMisra (approved):
lgtm

iothub/service/src/Messaging/ServiceClient.cs
brycewang-microsoft: no, unfortunately we are passing nothing
https://github.com/Azure/azure-iot-sdk-csharp/blob/main/iothub/service/src/Messaging/ServiceClient.cs#L288

PR #3476, opened 2024-11-21
feat(iot-serv): Add ability to send custom authentication scopes for RBAC auth

timtay-microsoft:
Previously, the SDK hardcoded just "https://iothubs.azure.net/.default" as the only supported scope, but this does not work for non-public clouds


patilsnr (approved):

SymbioticKilla:
@timtay-microsoft 
Are you sure?
Microsoft.Azure.Devices.Client 1.42.3 is from April 2024.

timtay-microsoft:
Client package doesn't support RBAC auth, so it doesn't need configurable RBAC authentication scopes.

SymbioticKilla:
@timtay-microsoft 
I thought that nuget updates on client side are also important changes, sorry for borrowing.

PR #3492, opened 2025-08-05
[DPS][Public Preview] Provisioniong Service SDK Update

jessHuh:
## Checklist

## Description of the changes

Adding a new field "CredentialPolicyName" in EnrollmentGruop and IndividualEnrollment. This change will be introduced as part of 2025-07-01-preview API.

### DIFF
--- a/provisioning/service/src/Config/EnrollmentGroup.cs
+++ b/provisioning/service/src/Config/EnrollmentGroup.cs
@@ -269,6 +269,12 @@ namespace Microsoft.Azure.Devices.Provisioning.Service
         public string IotHubHostName { get; set; }
 
         /// <summary>
+        /// Linked Azure Device Registry credential policy name (optional).
+        /// </summary>
+        [JsonProperty(PropertyName = "credentialPolicyName", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public string CredentialPolicyName { get; set; }
+
+        /// <summary>
         /// Initial twin state.
         /// </summary>
         [JsonProperty(PropertyName = "initialTwin", DefaultValueHandling = DefaultValueHandling.Ignore)]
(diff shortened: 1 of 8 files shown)


provisioning/service/src/Config/EnrollmentGroup.cs
```diff
@@ -268,6 +268,12 @@ public Attestation Attestation
         [JsonProperty(PropertyName = "iotHubHostName", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string IotHubHostName { get; set; }
 
+        /// <summary>
+        /// Linked Azure Device Registry credential policy name (optional).
+        /// </summary>
+        [JsonProperty(PropertyName = "credentialPolicyName", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public string CredentialPolicyName { get; set; }
```
ewertons, line 275: Is credentialPolicyName the only change expected, or there will be more? Are the signing certs also set through th DPS service API?

provisioning/service/tests/Config/EnrollmentGroupTests.cs
```diff
@@ -18,6 +18,7 @@ public class EnrollmentGroupTests
         private const string SampleLastUpdatedDateTimeUTCString = "2017-11-14T12:34:18.321Z";
         private DateTime SampleLastUpdatedDateTimeUTC = new DateTime(2017, 11, 14, 12, 34, 18, 321, DateTimeKind.Utc);
         private const string SampleEtag = "00000000-0000-0000-0000-00000000000";
+        private const string SampleCredentialPolicyName = "testCredentialPolicyName";
```
ewertons, line 21: Are you going to add any end to end tests as well?

provisioning/service/src/Config/EnrollmentGroup.cs
jessHuh: This is the only change.

provisioning/service/tests/Config/EnrollmentGroupTests.cs
jessHuh: I added samples but seems like to update the test, I need to publish the SDK first since tests are using SDK.

PR #3522, opened 2026-03-27
Fix a few release pipeline issues

timtay-microsoft:


iothub/service/src/Utilities/ETagConverter.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.Azure.Devices
 {
-    internal class ETagConverter : JsonConverter<ETag>
+    internal sealed class ETagConverter : JsonConverter<ETag>
```
timtay-microsoft, line 12: The build complained that this (and the other) class should be sealed

maximsemenov80 (approved):

PR #3523, opened 2026-03-31
Remove duplicate version property in Hub Device client project

timtay-microsoft:


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
timtay-microsoft, line 28: You can see this same property defined down here by the rest of the release metadata, so we can delete the above copy

avishekpant (approved):

PR #3524, opened 2026-04-01
Fix incorrect log statement in release pipeline

timtay-microsoft:
And change the build.buildnumber var to match just the datetime.rev


vsts/release.yaml
```diff
@@ -1,4 +1,4 @@
-name: $(BuildID)_$(BuildDefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
+name: $(Date:yyyyMMdd).$(Rev:r)
```
timtay-microsoft, line 1: This field's value becomes the $(Build.BuildNumber) variable refrerenced below, confusingly. By making it this value, we ensure that each upload of packages to the Azure SDK team's storage account can easily be browsed by the date it was uploaded

avishekpant (approved):

PR #3536, opened 2026-06-22
Edge LTS update 2

timtay-microsoft:
Disabling a few flakey tests and fixing a few MQTT layer issues

### DIFF
--- a/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
+++ b/iothub/device/src/Microsoft.Azure.Devices.Client.csproj
@@ -33,7 +33,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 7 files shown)


iothub/device/src/Microsoft.Azure.Devices.Client.csproj
```diff
@@ -33,7 +33,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Version>1.43.0-edgeLts</Version>
+    <Version>1.43.2-edgeLts</Version>
```
timtay-microsoft, line 36: Jumping two patch versions here because we already released 1.43.1-edgeLts

maximsemenov80 (approved):

